Issue #625, opened 2016-06-29
Web compatibility risk of specified RegExp lastIndex semantics


littledan:
I tried to ship a more spec-compliant implementation of RegExps in Chrome, but a user reported a web compat issue at https://bugs.chromium.org/p/chromium/issues/detail?id=624318

Seems like, historically, browsers did not throw for this test case:

``` js
x = /a/
Object.freeze(x)
"b".match(x)
```

However, ES2015 and ES5 are fairly clear that when a match isn't found, then a strict-mode write is done to set lastIndex to 0. This is tested by test262 https://github.com/tc39/test262/blob/master/test/built-ins/RegExp/prototype/exec/y-fail-lastindex-no-write.js .

Has anyone else tried shipping these semantics? I'm inclined to revert to legacy semantics for now until I can quantify the size of the breakage and know if it's OK.

[Sidebar: One argument for permanent "legacy" semantics is that it enhances the utility of frozen RegExps, but on the other hand, RegExps are already somewhat limited when frozen (e.g., in global mode, they don't make much sense).]

allenwb:
I don't think "legacy semantics" is the right term here? Perhaps legacy ES5 implementation bug would be a better characterization Do all major browsers actually have this bug?  

Some background, `Object.freeze`, strict mode, "strict writes", and the use of a strict write when updating `lastIndex` were all introduced by the ES5 spec. So, prior to ES5 there was no way to "freeze" `lastIndex`. Any code using `Object.freeze` could not be legacy code. If ES5 implementations allowed RegExp built-ins to silently update (or try to update) a non-writable, non-configurable  `lastIndex` property, the implementations were buggy. 

What is the buggy semantics you want and is this only `match` you are talking about or does the bug occur everywhere `lastIndex` is updated?  Allowing the value of a frozen `lastIndex` to be modified would violate one of the fundamental invariants and would be a security hole. Silently ignoring such a write  would be a contract violation that may hides bugs (because `lastIndex` is not updated in the manner that the spec. requires.).

Given the ES3 semantics of RegExp algorithms that update `lastIndex`, any code that is trying to freeze an RegExp is probably ill-conceived and likely buggy.  I don't think we should memorialize such bugs.

We have talked in the past about developing a functional RegExp form that didn't use any mutable internal state such as `lastIndex`. That still seems like it would be a good direction to pursue.

claudepache:
> Seems like, historically, browsers did not throw for this test case:
> ```js
> x = /a/
[...]

PR #627, opened 2016-06-30
Disregard lastIndex for non-global non-sticky regexps

claudepache:
With these changes , non-global non-sticky regexps with nonwritable
lastIndex property should just work:
- In RegExpBuiltinExec, remove any access to the lastIndex property for
  non-global, non-sticky regexp.
- In RegExp.prototype.@@search, avoid to to set the lastIndex property
  when not necessary.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -28687,17 +28687,18 @@ THH:mm:ss.sss
             1. Assert: _R_ is an initialized RegExp instance.
             1. Assert: Type(_S_) is String.
             1. Let _length_ be the number of code units in _S_.
-            1. Let _lastIndex_ be ? ToLength(? Get(_R_, `"lastIndex"`)).
             1. Let _flags_ be _R_.[[OriginalFlags]].
             1. If _flags_ contains `"g"`, let _global_ be *true*, else let _global_ be *false*.
             1. If _flags_ contains `"y"`, let _sticky_ be *true*, else let _sticky_ be *false*.
             1. If _global_ is *false* and _sticky_ is *false*, let _lastIndex_ be 0.
+            1. Else, let _lastIndex_ be ? ToLength(? Get(_R_, `"lastIndex"`)).
             1. Let _matcher_ be _R_.[[RegExpMatcher]].
             1. If _flags_ contains `"u"`, let _fullUnicode_ be *true*, else let _fullUnicode_ be *false*.
             1. Let _matchSucceeded_ be *false*.
             1. Repeat, while _matchSucceeded_ is *false*
               1. If _lastIndex_ &gt; _length_, then
-                1. Perform ? Set(_R_, `"lastIndex"`, 0, *true*).
+                1. If _global_ is *true* or _sticky_ is *true*, then
+                  1. Perform ? Set(_R_, `"lastIndex"`, 0, *true*).
                 1. Return *null*.
               1. Let _r_ be _matcher_(_S_, _lastIndex_).
               1. If _r_ is ~failure~, then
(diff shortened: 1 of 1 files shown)


claudepache:
The intention of this PR is to fix #625. I have attempted to find and remove all plausible write accesses to the lastIndex property on non-global, non-sticky regexps.

claudepache:
> I have attempted to find and remove all plausible write accesses to the lastIndex property on non-global, non-sticky regexps.

There remains the unconditional initialisation of `rx.lastIndex` that is part of the [RegExpInitialize abstract operation](https://tc39.github.io/ecma262/#sec-regexpinitialize) (step 12). However, I don’t think it is worth to "fix" it; for I think it could go wrong here only when applying the deprecated `RegExp#compile()`.

littledan:
Do you think this will be web-compatible?

claudepache:
> Do you think this will be web-compatible? Does this correspond to Firefox or Safari semantics?

Good question. First, note that setting a value to the lastIndex property of a non-global, non-writable regexp is already useless, as it is not used when performing a match (see Step 8 of [RegExpBuiltinExec](https://tc39.github.io/ecma262/#sec-regexpbuiltinexec)). The only significant difference is that that property is no longer updated to 0 after a failed match. Test cases:

```js
(function() {
    var rx = /a/
    rx.lastIndex = 3
    rx.exec('b')
    return rx.lastIndex
})()
```

* `0` in Firefox, Chrome, Edge (in accordance to current and ES5 spec);
* `3` in Safari (in accordance to the proposed changes).

```js
(function() {
    var rx = /a/
    rx.lastIndex = 3
    'b'.match(rx)
    return rx.lastIndex
})()
```

* `0` in Chrome, Edge, Firefox Nightly (in accordance to current  and ES5 spec);
* `3` in Safari and stable Firefox (in accordance to the proposed changes).

claudepache:
Also, just in case it was not clear (note the /g flag):

```js
'b'.match(Object.freeze(/a/g))
```

Throws in current Firefox, Safari, Chrome, as well as any version of spec that supports Object.freeze.

littledan:
Would we need all of the changes in this patch to get the intended effect, or just the changes in lines 28700-28701?

claudepache:
> Would we need all of the changes in this patch to get the intended effect, or just the changes in lines 28700-28701?

* The changes in lines 28690/28694 just avoids a useless read access to `rx.lastIndex`.

* The changes of RegExp.prototype.@@search in lines 28940ff are needed in order to support `String#search` on frozen non-global regexps. (More precisely, it supports the most likely case, where `lastIndex` is blocked to 0: if we want to support the other cases, it becomes slightly more complicated.)

bterlson:
This change seems good to me - no need to do puts that are useless. However, still needs-consensus. Any volunteers to lead the discussion at TC39?

leobalter:
Count me in. I don't have anything complex to bring so I'll have time to do this.

littledan:
I'd be happier with this if we made this a more minimal change, avoiding the useless yet observable changes to reading lastIndex. It sounds like the changes to search are needed, however. I will unfortunately not have data beyond the bug report to contribute to the meeting.

claudepache:
> I'd be happier with this if we made this a more minimal change, avoiding the useless yet observable changes to reading lastIndex. It sounds like the changes to search are needed, however. I will unfortunately not have data beyond the bug report to contribute to the meeting.

About the useless read of lastIndex. This is part of the [RegExpBuiltinExec](https://tc39.github.io/ecma262/#sec-regexpbuiltinexec) abstract operation, which is applied only to objects that have a [[RegExpMatcher]] internal slots; and those objects always have an own nonconfigurable data property named `lastIndex` (see [RegExpAlloc](https://tc39.github.io/ecma262/#sec-regexpalloc)). So that any useless reading of that property on such object is truly unobservable.

About the changes in RegExp.prototype.@@search: It might be possible that the change is not needed for webcompat; however, I have tried to be consistent in making frozen regexp to just work under reasonable conditions, rather than to modify the strict minimum number of line needed for resolving the particular reported issue. Consistency is important in order to avoid bugs.

arai-a:
ToLength is observable there.

  let r = /a/;
  a.lastIndex = { valueOf() { throw "hello" } };
  r.test("a"); // this throws "hello", while ToLength operation

claudepache:
> ToLength is observable there.

True. This is probably only significant for checking how scrupulously implementations follow the spec. Test case:


```js
function test(rx) {
    rx.lastIndex = { valueOf: function () { throw "ok" } }
    try {
        rx.exec('foo')
    }
    catch (e) {
        if (e === "ok")
            return true
    }
    return false
}
```

Results:

 | test(/a/g) | test(/a/)
---|---|---
ECMA262, Firefox, Chrome | true | true
Safari, IE, Edge | true | false

leobalter:
This reached consensus on the July meeting. :shipit:

littledan:
Sounds like we need test262 tests before merging this PR. Who wants to write them?

leobalter:
it's good to verify the work on tc39/test262#711 to avoid any collision.

I can help @claudepache to get the tests done.

tcare:
No collision that I could see.

bterlson:
@tcare / @leobalter we merged test262 tests for this right? Shall I take this PR?

leobalter:
yes, please

schuay:
@bterlson Any news on merging this?

bterlson:
@schuay thanks for the ping, I think it's good to go. I removed the label and then promptly lost it :-P

PR #633, opened 2016-07-06
Editorial: Remove ambiguous statement from Date.UTC

leobalter:
The word optionally might be taken as implementation defined to use
the respective arguments it refers to. Removing it makes it clear
the following steps will be taken when Date.UTC is called with 3 or
more arguments.

cc @jugglinmike

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -72,7 +72,7 @@
 <emu-clause id="sec-overview">
   <h1>Overview</h1>
   <p>This section contains a non-normative overview of the ECMAScript language.</p>
-  <p>ECMAScript is an object-oriented programming language for performing computations and manipulating computational objects within a host environment. ECMAScript as defined here is not intended to be computationally self-sufficient; indeed, there are no provisions in this specification for input of external data or output of computed results. Instead, it is expected that the computational environment of an ECMAScript program will provide not only the objects and other facilities described in this specification but also certain environment-specific objects, whose description and behavior are beyond the scope of this specification except to indicate that they may provide certain properties that can be accessed and certain functions that can be called from an ECMAScript program.</p>
+  <p>ECMAScript is an object-oriented programming language for performing computations and manipulating computational objects within a host environment. ECMAScript as defined here is not intended to be computationally self-sufficient; indeed, there are no provisions in this specification for input of external data or output of computed results. Instead, it is expected that the computational environment of an ECMAScript program will provide not only the objects and other facilities described in this specification but also certain environment-specific objects, whose description and behaviour are beyond the scope of this specification except to indicate that they may provide certain properties that can be accessed and certain functions that can be called from an ECMAScript program.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -25647,7 +25647,7 @@
       <!-- es6num="20.3.3.4" -->
       <emu-clause id="sec-date.utc">
         <h1>Date.UTC ( _year_, _month_ [ , _date_ [ , _hours_ [ , _minutes_ [ , _seconds_ [ , _ms_ ] ] ] ] ] )</h1>
-        <p>When the `UTC` function is called with fewer than two arguments, the behaviour is implementation-dependent. When the `UTC` function is called with two to seven arguments, it computes the date from _year_, _month_ and (optionally) _date_, _hours_, _minutes_, _seconds_ and _ms_. The following steps are taken:</p>
+        <p>When the `UTC` function is called with fewer than two arguments, the behaviour is implementation-dependent. When the `UTC` function is called with two to seven arguments, the following steps are taken:</p>
```
ljharb: also i think we switched to US spellings - ie "behavior" instead of "behaviour" :-)

claudepache: > i think we switched to US spellings

I think we use OED spelling (whose main difference with British spelling is verbs ending in _-ize_). See:
https://bugs.ecmascript.org/show_bug.cgi?id=2467
https://en.wikipedia.org/wiki/Oxford_spelling#Usage

allenwb: > I think we use OED spelling (whose main difference with British spelling is verbs ending in -ize). See:

Exactly.  At least that is the way it unless something changed over the last year.  ISO says standards should be written use OED English spellings.

leobalter:
@ljharb I couldn't live with inconsistency so after fixing the "behaviour" on my change, I did another commit replacing other uses of behaviour. That includes a link that will move from `sec-ordinary-and-exotic-objects-behaviours` to `sec-ordinary-and-exotic-objects-behaviors`.

@bterlson the extra commit is just a plus and please ignore and discard it if it blocks the main proposed change.

leobalter:
@ljharb I couldn't live with inconsistency so after fixing the "behaviour" on my change, I did another commit replacing other uses of behaviour. That includes a link that will move from `sec-ordinary-and-exotic-objects-behaviours` to `sec-ordinary-and-exotic-objects-behaviors`.

@bterlson the extra commit is just a plus and please ignore and discard it if it blocks the main proposed change.

On the OED spelling: I can't find any open access for it. Does Ecma has any legal form to browse it or should I subscribe it? Without consulting the OED these definitions will remain cloudy.

leobalter:
Ok, I've got word from someone with access to an OED from the 70s and _behaviour_ is the correct option.

I reverted the "behavior" commits only to avoid force push, please discard before merging.

spec.html
ljharb: Ah, thanks for clarifying. I wasn't familiar with the distinction :-)

ljharb: (That means we have some "behavior" instances to clean up tho)

ljharb:
There's no need to ever avoid force pushing - probably easiest to just rebase them out.

bterlson:
Just pushed a fix for the improper usage of "behavior".

PR #636, opened 2016-07-14
Move Array.prototype.toLocaleString's separator-computation step to just before the loop that first uses it

jswalden:
I was reviewing a SpiderMonkey patch that put the step in its current order, and I was going to tell the author to move the step out of order, then I realized I could just change the spec.  :-)  (Theoretically we could bikeshed the step even one later, but I prefer to view "Let k be 1" and the loop on k as a single logical step.)

I'm also filing a PR against ecma402 for the same issue there.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -30297,13 +30297,13 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _array_ be ? ToObject(*this* value).
           1. Let _len_ be ? ToLength(? Get(_array_, `"length"`)).
-          1. Let _separator_ be the String value for the list-separator String appropriate for the host environment's current locale (this is derived in an implementation-defined way).
           1. If _len_ is zero, return the empty String.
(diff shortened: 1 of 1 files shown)


allenwb:
Presumably, the placement of that step only matters if it is possible for it to fail in some observable way.  I can imagine that a "implementation defined way" of doing this could be self-hosted in JS code, in which case it certainly might throw for some reason.

The separator is only obtained once.  The advantage of its current  placement in the algorithms is that any such failure will occur before any array elements are processed.  With this patch, such a failure would occur after the first element is processed (including  invoking `toLocaleString` to the first element value).

In the case where there actually is an exception while retrieving the separator, it seems cleaner if the exception occurs before any element processing.

jswalden:
Hmm.  I feel like I remember seeing at least one case where an algorithm did separator-like computation after an early out for the 0/1 cases in the spec, but searching for "is zero" isn't finding it.  Given that, fair enough.

PR #642, opened 2016-07-22
Specify Date.UTC when called with fewer than two arguments

claudepache:


claudepache:
Followup of whatwg/javascript#31

bterlson:
See also: https://github.com/Microsoft/ChakraCore/issues/1318, https://github.com/whatwg/javascript/issues/31.

claudepache:
Note that, as currently specified, when called with 1 argument, step 1 may have side-effects, although the final result will always be either null or an abrupt completion:

```js
Date.UTC({ valueOf: function() { throw "hey!" } })
```

(Firefox, Chrome and Safari throw "hey!" in that testcase.)

dilijev:
@claudepache Thanks for pointing that test case out. Edge also throws "hey!" in that case.

bterlson:
Consensus is Date.UTC() returns NaN and Date.UTC(2017) === Date.UTC(2017, 0). This PR needs a slight update for this.

spec.html
```diff
@@ -25643,7 +25643,7 @@
       <!-- es6num="20.3.3.4" -->
       <emu-clause id="sec-date.utc">
         <h1>Date.UTC ( _year_, _month_ [ , _date_ [ , _hours_ [ , _minutes_ [ , _seconds_ [ , _ms_ ] ] ] ] ] )</h1>
```
dilijev: `_month_` is optional now.

claudepache: corrected

claudepache:
I've updated the PR.

spec.html
```diff
@@ -25642,11 +25642,11 @@
...
       <!-- es6num="20.3.3.4" -->
       <emu-clause id="sec-date.utc">
-        <h1>Date.UTC ( _year_, _month_ [ , _date_ [ , _hours_ [ , _minutes_ [ , _seconds_ [ , _ms_ ] ] ] ] ] )</h1>
-        <p>When the `UTC` function is called with fewer than two arguments, the behaviour is implementation-dependent. When the `UTC` function is called with two to seven arguments, the following steps are taken:</p>
+        <h1>Date.UTC ( _year_ [ , _month_ [ , _date_ [ , _hours_ [ , _minutes_ [ , _seconds_ [ , _ms_ ] ] ] ] ] ] )</h1>
+        <p>When the `UTC` function is called, the following steps are taken:</p>
         <emu-alg>
           1. Let _y_ be ? ToNumber(_year_).
-          1. Let _m_ be ? ToNumber(_month_).
+          1. If _month_ is supplied, let _m_ be ? ToNumber(_month_); else let _m_ be 0.
```
leobalter: @bterlson I just want to confirm this for clarification: we'll have a different behavior for supplied but undefined month. Like the following expected values:

```js
Date.UTC(2016) // 1451606400000
Date.UTC(2016, 0) // 1451606400000
Date.UTC(2016, undefined) // NaN
```

This will be consistent with the other parameters, but `Date.UTC` will never be perfect.

claudepache: This is also consistent with the Date constructor:
```js
new Date(2016, 0) // Fri Jan 01 2016 00:00:00 GMT+0100 (CET)
new Date(2016, undefined) // Invalid Date
```

bterlson: Yeah, this is just how it is :-P

littledan:
Is this patch ready to land? Should it go in ES2017? Seems all good to me.

evilpie:
@claudepache @bterlson can we land this?

bterlson:
@evilpie yep seems good to me. @littledan unless you object I was planning on bringing this to the es2017 branch since we have consensus on the semantics from a long while back.

bterlson:
Also merged into es2017.

Issue #357, opened 2016-02-04
More Ordinary* abstract operations


annevk:
I created #356 which refactors some existing text in a way that seems more consistent with the OrdinaryHasProperty etc. precedent that was already set. However, I would also appreciate it if we could introduce OrdinaryGetPrototypeOf, OrdinaryDelete, and OrdinaryEnumerate for use in https://github.com/annevk/html-cross-origin-objects (which will be merged into the HTML standard in due course).

Would such a PR be accepted even though the restructuring is only for an external document?

bterlson:
Mirroring my comment on #356, yep, this seems great! We should do it for all the slots!

annevk:
Closing in favor of that PR.

Issue #343, opened 2016-02-01
DefaultInternalMethod abstract operation


annevk:
To define the `Location` and `WindowProxy` objects we need to override the default behavior for various internal methods in cross-origin scenarios and in some cases for same-origin too. However, we would still like to sometimes use the default code path defined by ECMASCript. I introduced a DefaultInternalMethod abstract operation for this, defined as follows:

> DefaultInternalMethod(_internalMethod_, _O_, _arguments_...)
> 
> 1. Return the result of calling the default ordinary object _internalMethod_ internal method on _O_ passing _arguments_ as the arguments.

Invocation example: DefaultInternalMethod([[GetPrototypeOf]\], _O_).

@domenic suggested in https://github.com/annevk/html-cross-origin-objects/issues/27 to upstream this. Looking around in ECMAScript, perhaps this could be used to replace various Ordinary* abstract operations?

(See https://github.com/annevk/html-cross-origin-objects for further details. It is not yet integrated into the HTML standard, but it's getting close.)

bterlson:
I don't think abstracting out a few words of ECMASpeak behind an abstract operation helps much on the balance, personally.

jswalden:
If it were me, I'd just make every one of the OOIMs have a one-step algorithm calling an `Ordinary<name of internal method>()` method defined in the same section, i.e. normalize what we've done to `[[HasProperty]]` to everything.  Can't say I care that much how this is formatted, tho.

bterlson:
@jswalden can you give a more detailed example? Not sure I follow...

bterlson:
Also, fwiw, if we can figure out a notation to use method invocation syntax for polymorphic calls to abstract operations, it would help much here. For example, if we said we could invoke Evaluate for a particular non-terminal via `_prod_::Evaluate()` or `|Prod|::Evaluate()`, you could also imagine a dfn for OrdinaryObject combined with `OrdinaryObject::[[GetPrototypeOf]](_arguments_)`. I would much prefer we explore this route than adding convenience abstract ops to avoid a little notational overhead for individual cases.

jswalden:
Well, we have this right now:

> 9.1.3 `[[IsExtensible]] ( )`
> When the `[[IsExtensible]]` internal method of _O_ is called the following steps are taken:
> 1. Return the value of the `[[Extensible]]` internal slot of _O_.

I'd say make it:

> 9.1.3 `[[IsExtensible]] ( )`
[...]

Issue #91, opened 2015-10-19
Factor out all ordinary object internal methods into OrdinaryXyz()?


domenic:
HTML needs to specify a couple exotic objects. (Web IDL also needs to do so, if it ever gets updated to ES2015.) In many cases the behavior is "if flag X is set, throw. Otherwise, do what an ordinary object would do."

For a couple internal methods, there are pre-existing abstract operations we can use: OrdinaryGetOwnProperty and OrdinaryDefineOwnProperty. But for all others, there are not.

I see two potential ways we could spec such exotic objects:

1. Pull request ES to have OrdinaryXyz() abstract operation counterparts for all internal methods, and then use them: "If flag X is set, throw a TypeError. Otherwise, return OrdinarySetPrototypeOf(O, V)".
2. Try to come up with some phrasing that is sufficiently precise. E.g.: "If flag X is set, throw a TypeError. Otherwise, return the result of the same algorithm as the [[SetOwnProperty]]() method for ordinary objects, applied to object O with argument V."

Any advice? Is there a chance (1) could clean up parts of the ES spec that are currently duplicated?

/cc @annevk @allenwb

anba:
You could use the same language as in [9.4.4 Arguments Exotic Objects](http://tc39.github.io/ecma262/#sec-arguments-exotic-objects), for example:

> Return the result of calling the default ordinary object [[SetPrototypeOf]] internal method (9.1.2) on O passing V as the argument.

[[Get]], [[Set]] and [[Delete]] for arguments objects use this wording to call the ordinary internal methods.

domenic:
I guess that covers it; thanks!

annevk:
Is there a reason ECMAScript uses both patterns?

bterlson:
@annevk which patterns are you referring to?

annevk:
What @anba mentioned and using OrdinaryInternalMethod abstract operations.

annevk:
Also, if I say use "default ordinary object [[Get]] internal method", does that mean the [[GetOwnProperty]] call it does (and similar ones) are the normal non-overridden variants as well?

allenwb:
@annevk 
Choice of how to do this is mostly a judgement call by the editor. delegating to OrdinaryX abstraction operation adds a level of indirection that make the normal case harder to read/understand. Using ""default ordinary object..." (and note a section reference is always included) only adds complexity at the "call site".

[...]

PR #651, opened 2016-07-29
Editorial: collapse `OrdinaryPreventExtensions`/`OrdinaryIsExtensible` to their sole callsite

ljharb:
It seems silly to me to have `OrdinaryPreventExtensions` and `OrdinaryIsExtensible` when they're only used once. Thoughts on removing them?

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6335,16 +6335,8 @@
       <h1>[[IsExtensible]] ( )</h1>
       <p>When the [[IsExtensible]] internal method of _O_ is called, the following steps are taken:</p>
       <emu-alg>
-        1. Return ! OrdinaryIsExtensible(_O_).
+        1. Return _O_.[[Extensible]].
       </emu-alg>
-
-      <emu-clause id="sec-ordinaryisextensible" aoid="OrdinaryIsExtensible">
-        <h1>OrdinaryIsExtensible (_O_)</h1>
-        <p>When the abstract operation OrdinaryIsExtensible is called with Object _O_, the following steps are taken:</p>
-        <emu-alg>
-          1. Return _O_.[[Extensible]].
-        </emu-alg>
-      </emu-clause>
     </emu-clause>
 
     <!-- es6num="9.1.4" -->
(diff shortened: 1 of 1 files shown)


evilpie:
These are referenced by https://html.spec.whatwg.org, but not actually used.

ljharb:
Sounds like the original intention of all these OrdinaryFoos was so the HTML spec could use them (https://github.com/tc39/ecma262/issues/357 for example) - but if they're not actually used, could we remove them, and re-add them in the future if they're needed?

jswalden:
Yeah, these were added 1) for external access, and 2) to have a consistent pattern of specification _even if_ some OrdinaryFoo isn't referred to by an unknown external caller.  Better to do it one way everywhere, than to have oddballs that just look inconsistent.  In addition to #357, see also #91 to deliberately add all of these.  This fell out of a request from @annevk and @bzbarsky for use in DOM specs, in an issue that I'm having trouble finding just right now (and an idea of Ordinary* plus [[*]] for everything, that I think I originally suggested).

ljharb:
If we want to keep them, that's fine - instead I'll open another PR that tries to refactor a bit so the operations aren't useless in isolation, and also adds non-normative notes so it's clear why they exist.

jswalden:
My original suggestion of this was in #343; I don't know what issue/PR actually followed through on the idea everywhere.

ljharb:
@bterlson do you have any ideas for how to reduce the "OrdinaryWhatever" boilerplate, while still allowing consistent external hooks?

claudepache:
> I don't know what issue/PR actually followed through on the idea everywhere.

See https://github.com/tc39/ecma262/commit/2b61c8719c210d2381b9f62728baaf823cb48907

annevk:
We don’t need all in HTML at the moment, but a revision of IDL might. Not sure. But the reason I expanded all was editor preference.

bterlson:
All for clarifications, but it has been demonstrated that this factoring is useful for layering purposes so I intend to keep it unless I hear from HTML-land that they don't want it anymore :)

PR #656, opened 2016-08-12
Editorial: Tweak EnumerateObjectProperties informative definition

shvaikalesh:
1. Add missing semicolon.
1. Improve variable naming consistency (`protoName` --> `protoKey`, because `key`).
1. Early `continue` for `symbol`s (less indentation, more obvious what keys are skipped).
1. Use `const` instead of `let`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -16181,7 +16181,7 @@
                   }
                 }
               }
-              let proto = Reflect.getPrototypeOf(obj)
+              let proto = Reflect.getPrototypeOf(obj);
               if (proto === null) return;
               for (let protoName of EnumerateObjectProperties(proto)) {
                 if (!visited.has(protoName)) yield protoName;
--- a/spec.html
+++ b/spec.html
@@ -16173,12 +16173,11 @@
             function* EnumerateObjectProperties(obj) {
               let visited = new Set;
               for (let key of Reflect.ownKeys(obj)) {
-                if (typeof key === "string") {
-                  let desc = Reflect.getOwnPropertyDescriptor(obj, key);
-                  if (desc && !visited.has(key)) {
-                    visited.add(key);
-                    if (desc.enumerable) yield key;
-                  }
+                if (typeof key === "symbol") continue;
+                let desc = Reflect.getOwnPropertyDescriptor(obj, key);
+                if (desc && !visited.has(key)) {
+                  visited.add(key);
+                  if (desc.enumerable) yield key;
                 }
               }
               let proto = Reflect.getPrototypeOf(obj);
--- a/spec.html
+++ b/spec.html
@@ -16171,18 +16171,18 @@
           <p>The following is an informative definition of an ECMAScript generator function that conforms to these rules:</p>
           <pre><code class="javascript">
             function* EnumerateObjectProperties(obj) {
-              let visited = new Set;
-              for (let key of Reflect.ownKeys(obj)) {
+              const visited = new Set;
+              for (const key of Reflect.ownKeys(obj)) {
                 if (typeof key === "symbol") continue;
-                let desc = Reflect.getOwnPropertyDescriptor(obj, key);
+                const desc = Reflect.getOwnPropertyDescriptor(obj, key);
                 if (desc && !visited.has(key)) {
                   visited.add(key);
                   if (desc.enumerable) yield key;
                 }
               }
-              let proto = Reflect.getPrototypeOf(obj);
+              const proto = Reflect.getPrototypeOf(obj);
               if (proto === null) return;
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -16175,16 +16175,16 @@
               for (let key of Reflect.ownKeys(obj)) {
                 if (typeof key === "string") {
                   let desc = Reflect.getOwnPropertyDescriptor(obj, key);
-                  if (desc && !visited.has(key)) {
+                  if (desc &amp;&amp; !visited.has(key)) {
```
claudepache: Why? it just makes the source code less readable.

mathiasbynens: Agreed. @shvaikalesh See https://mathiasbynens.be/notes/ambiguous-ampersands for more info on why this is perfectly valid.

spec.html
```diff
@@ -16175,16 +16175,16 @@
               for (let key of Reflect.ownKeys(obj)) {
```
mathiasbynens: Shouldn’t we use `const` over `let` throughout this code sample? None of the `let`s are reassigned.

spec.html
shvaikalesh: @mathiasbynens, thanks for the article. I dropped this commit and will make another PR to unescape all unambiguous ampersands in `spec.html`.

spec.html
shvaikalesh: It is mostly a matter of code style. Some might argue that `const` is for declarations that will never ever be reassigned, like number constants or `require`s. And `let`s are for any other declarations: this keeps diffs cleaner.

Lets make them `const`s to reflect how most people write ES201* these days.

spec.html
```diff
@@ -16171,20 +16171,19 @@
...
-              let visited = new Set;
-              for (let key of Reflect.ownKeys(obj)) {
-                if (typeof key === "string") {
-                  let desc = Reflect.getOwnPropertyDescriptor(obj, key);
-                  if (desc && !visited.has(key)) {
-                    visited.add(key);
-                    if (desc.enumerable) yield key;
-                  }
+              const visited = new Set;
+              for (const key of Reflect.ownKeys(obj)) {
```
leobalter: there's no tweak here. This change is not necessary at all and for consistency it is better to remain without change.

spec.html
```diff
@@ -16171,20 +16171,19 @@
...
+                if (typeof key === "symbol") continue;
+                const desc = Reflect.getOwnPropertyDescriptor(obj, key);
+                if (desc && !visited.has(key)) {
+                  visited.add(key);
+                  if (desc.enumerable) yield key;
                 }
               }
-              let proto = Reflect.getPrototypeOf(obj)
+              const proto = Reflect.getPrototypeOf(obj);
```
leobalter: this `;` is the only part I agree we should merge. The remaining are unnecessary as TC39 does not advise using const over let or against it.

spec.html
shvaikalesh: @leobalter, thanks for the review. Is it https://github.com/tc39/ecma262/pull/656/commits/46b569dc9f9ad621c2c97a022eae3d7e8603a3e3 or https://github.com/tc39/ecma262/pull/656/commits/f1883f50a6c7df79902f01c660486b889bdd7b07 that should be dropped?

leobalter: I'm just part of TC39, @bterlson is the one able to answer that.

spec.html
shvaikalesh: Any JavaScript code contains opinionated parts: this function will be just fine without semicolons and strict equals. Does TC39 advise against `==` and ASI too?

spec.html
ljharb: I think it's ok to change it to use const, since it's never reassigned. I'd also want to see `new Set()` while we're at it :-)

spec.html
ljharb: TC39 isn't advising one over the other in either case. Since both are equivalent, and the community tends to prefer const here, I thi k that's the better choice.

spec.html
bterlson: I am fine with this change. I don't see why this change isn't consistent though - @leobalter can you clarify?

leobalter: I prefer this example being consistent (the same) along the new releases.

+ replacing let w/ const will make it valid to replace const w/ let in any near future as we don't enforce any particular use.

bterlson: We do enforce a particular use in the spec at least - whatever I feel like 😈

My feelings may change (I have in fact become far less likely to use const) but for now I think this helps and I promise I won't change it back to let any time soon.

shvaikalesh: @ljharb addressed in https://github.com/tc39/ecma262/pull/656/commits/aff0edd611d4c460225a681bf17455f2eaf3fbe7

PR #657, opened 2016-08-12
Editorial: Hoist thisBooleanValue declaration

shvaikalesh:
Move declaration to parent clause for consistency with other primitives' prototypes.

Additionally, section numbers of subsequent clauses now match with HTML `<!-- es6num=...` comments.


bterlson:
I think the others should be pulled into a subclause rather than moving this one into the parent clause. The subclause enables auto-linking of `this${PrimitiveType}Value()` calls. I don't personally care that es6num matches - I was going to remove those soon anyway :)

shvaikalesh:
Are PRs welcome on those?

bterlson:
@shvaikalesh For sure!

PR #658, opened 2016-08-12
Editorial: Rename variable (indx --> index)

shvaikalesh:
Rename variable in `Array.prototype.pop` to be consistent with naming in other algorithms.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -29832,9 +29832,9 @@ THH:mm:ss.sss
             1. Return *undefined*.
           1. Else _len_ &gt; 0,
             1. Let _newLen_ be _len_-1.
-            1. Let _indx_ be ! ToString(_newLen_).
-            1. Let _element_ be ? Get(_O_, _indx_).
-            1. Perform ? DeletePropertyOrThrow(_O_, _indx_).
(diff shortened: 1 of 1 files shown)


bterlson:
But that `e` adds one whole byte to each variable name :(

shvaikalesh:
However, proposed option is better for gzip and search.

UltCombo:
You can't read tones in the internet, but I'm pretty sure Brian's comment was not serious.

bterlson:
yeah sorry, I was being sarcastic :( There's no excuse for `indx` (other than it's been that way since ES5).

UltCombo:
Blame it on HTML/Markdown for lacking sarcastic markup. 😸

bterlson:
postfix `/s` is the defacto standard for marking sarcasm but I didn't employ it. Tsk tsk!

littledan:
I thought `/s` made `.` match newlines. I'm so confused!

UltCombo:
@littledan Not until the [dotAll proposal](https://github.com/mathiasbynens/es-regexp-dotall-flag#ecmascript-proposal-s-dotall-flag-for-regular-expressions) gets to stage 4! 😜

PR #669, opened 2016-08-15
Fix EnumerateObjectProperties definition

deian:
It seems like `Reflect.enumerate` should enumerate keys that are enumerable in the prototype but not the object itself. (As `for ... in` does.) Or maybe I am misunderstanding something?

+credit: @bengl

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -16176,8 +16176,10 @@
                 if (typeof key === "symbol") continue;
                 const desc = Reflect.getOwnPropertyDescriptor(obj, key);
                 if (desc && !visited.has(key)) {
-                  visited.add(key);
-                  if (desc.enumerable) yield key;
+                  if (desc.enumerable) {
+                    visited.add(key);
+                    yield key;
+                  }
                 }
               }
               const proto = Reflect.getPrototypeOf(obj);


ljharb:
`Reflect.enumerate` no longer exists.

The goal of `EnumerateObjectProperties` is to "[iterate] over all the String-valued keys of enumerable properties of O".

Indeed, with your change, a non-enumerable child property will not actually shadow an enumerable parent property - as-is, if the child property is non-enumerable, it doesn't matter if a parent property is enumerable or not, only the child-most property matters.

I'm pretty certain this semantic is correct, since it also follows `[[Get]]`.

In which browser/engine does `for..in` enumerate over an enumerable parent property that's shadowed by a non-enumerable child property? v8 has an open bug regarding this, for example.

deian:
Thanks for replying so quickly. I'm working with v8, so I guess the bug is in their `for-in`.
The sematics makes sense, but IMO  "[iterate] over all the String-valued keys of enumerable properties of O" is somewhat ambiguous (given the prototype lookups).

ljharb:
A shadowed parent property wouldn't be an observable property of O.

PR #673, opened 2016-08-17
Implement undefined-like exotic objects

### LINKED ISSUE #668: document.all willful violation
Other than some event loop mismatch which I think we'll be able to resolve more easily, HTML (and JavaScript engines) also contain(s) the `document.all` willful violation.

https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-all

How much appetite is there to try to formalize this exception?

bterlson:
Closes #668. Still not convinced we want this in 262 but here it is so we can discuss!

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -39907,7 +39907,7 @@ THH:mm:ss.sss
 
     <emu-annex id="sec-IsHTMLDDA-internal-slot">
       <h1>The [[IsHTMLDDA]] Internal Slot</h1>
-      <p>An <dfn>[[IsHTMLDDA]] internal slot</dfn> may exist on implementation-defined exotic objects. Objects with an [[IsHTMLDDA]] internal slot behave like undefined in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">typeof operator</emu-xref>.</p>
+      <p>An <dfn>[[IsHTMLDDA]] internal slot</dfn> may exist on implementation-defined exotic objects. Objects with an [[IsHTMLDDA]] internal slot behave like undefined in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">`typeof` operator</emu-xref>.</p>
       <emu-note>
         <p>Objects with an [[IsHTMLDDA]] internal slot are never created by this specification. However, the <a href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-all"><code>document.all</code> object</a> in web browsers is a host-created exotic object with this slot that exists for web compatibility purposes. There are no other known examples of this type of object and implementations should not create any with the exception of `document.all`.</p>
       </emu-note>
--- a/spec.html
+++ b/spec.html
@@ -39905,35 +39905,38 @@ THH:mm:ss.sss
       </emu-alg>
     </emu-annex>
 
-    <emu-annex id="sec-html-document-dot-all-exotic-object">
-      <h1>HTML Document Dot All Exotic Object</h1>
-      <p>An <dfn>HTML document dot all exotic object</dfn> is an exotic object that behaves like undefined in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">typeof operator</emu-xref>. HTML document dot all exotic objects have all the ordinary object internal methods and slots.</p>
+    <emu-annex id="sec-ishtmldda-internal-slot">
+      <h1>The [[isHTMLDDA]] Internal Slot</h1>
+      <p>An <dfn>[[isHTMLDDA]] internal slot</dfn> may exist on implementation-defined exotic objects. Objects with an [[isHTMLDDA]] intenral slot behave like undefined in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">typeof operator</emu-xref>.</p>
       <emu-note>
-        <p>These objects are never created by this specification. However, the <a href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-all"><code>document.all</code> object</a> in web browsers is a host-created HTML document dot all exotic object that exists for web compatibility purposes. There are no other known examples of this type of object.</p>
+        <p>Objects with an [[isHTMLDDA]] internal slot are never created by this specification. However, the <a href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-all"><code>document.all</code> object</a> in web browsers is a host-created exotic object with this slot that exists for web compatibility purposes. There are no other known examples of this type of object and implementations should not create any with the exception of `document.all`.</p>
       </emu-note>
 
       <emu-annex id="sec-html-document-dot-all-exotic-object-to-boolean">
         <h1>Changes to ToBoolean</h1>
         <p>The result column in <emu-xref href="#table-10"></emu-xref> for an argument type of Object is replaced with the following algorithm:</p>
         <emu-alg>
-          1. If _argument_ is an HTML document dot all exotic object, return *true*.
-          1. Return *false*.
+          1. If _argument_ has an [[isHTMLDDA]] internal slot, return *false*.
+          1. Return *true*.
         </emu-alg>
       </emu-annex>
 
       <emu-annex id="sec-html-document-dot-all-exotic-object-aec">
         <h1>Changes to Abstract Equality Comparison</h1>
         <p>The following steps are inserted after step 3 of the <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> algorithm:</p>
-        <emu-alg>
-          1. If _x_ is an undefined-like exotic object and _y_ is either *null* or *undefined*, return *true*.
-          1. If _x_ is either *null* or *undefined* and _y_ is an undefined-like exotic object, return *true*.
+        <emu-alg start=4>
+          1. If _x_ has an [[isHTMLDDA]] internal slot and _y_ is either *null* or *undefined*, return *true*.
+          1. If _x_ is either *null* or *undefined* and _y_ has an [[isHTMLDDA]] internal slot, return *true*.
         </emu-alg>
       </emu-annex>
 
       <emu-annex id="sec-html-document-dot-all-exotic-object-typeof">
-        <h1>Changes to the typeof operator</h1>
+        <h1>Changes to the `typeof` operator</h1>
         <p>The following table entry is inserted into <emu-xref href="#table-35"></emu-xref> immediately preceeding the entry for "Object (implements [[Call]])":</p>
-        <emu-table caption="Additional Typeof Table Entries">
+        <emu-table>
+          <emu-caption>
+            Additional <emu-xref href="#sec-typeof-operator">`typeof`</emu-xref> Operator Results
+          </emu-caption>
           <table>
             <tr>
               <th>
@@ -39945,7 +39948,7 @@ THH:mm:ss.sss
             </tr>
             <tr>
               <td>
-                Object (HTML dot all exotic object)
+                Object (has an [[isHTMLDDA]] internal slot)
               </td>
               <td>
                 `"undefined"`
--- a/spec.html
+++ b/spec.html
@@ -39907,12 +39907,12 @@ THH:mm:ss.sss
 
     <emu-annex id="sec-IsHTMLDDA-internal-slot">
       <h1>The [[IsHTMLDDA]] Internal Slot</h1>
-      <p>An <dfn>[[IsHTMLDDA]] internal slot</dfn> may exist on implementation-defined exotic objects. Objects with an [[IsHTMLDDA]] internal slot behave like *undefined* in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">`typeof` operator</emu-xref>.</p>
+      <p>An <dfn>[[IsHTMLDDA]] internal slot</dfn> may exist on implementation-defined objects. Objects with an [[IsHTMLDDA]] internal slot behave like *undefined* in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">`typeof` operator</emu-xref>.</p>
       <emu-note>
         <p>Objects with an [[IsHTMLDDA]] internal slot are never created by this specification. However, the <a href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-all"><code>document.all</code> object</a> in web browsers is a host-created exotic object with this slot that exists for web compatibility purposes. There are no other known examples of this type of object and implementations should not create any with the exception of `document.all`.</p>
       </emu-note>
 
-      <emu-annex id="sec-html-document-dot-all-exotic-object-to-boolean">
+      <emu-annex id="sec-IsHTMLDDA-internal-slot-to-boolean">
         <h1>Changes to ToBoolean</h1>
         <p>The result column in <emu-xref href="#table-10"></emu-xref> for an argument type of Object is replaced with the following algorithm:</p>
         <emu-alg>
@@ -39921,7 +39921,7 @@ THH:mm:ss.sss
         </emu-alg>
       </emu-annex>
 
-      <emu-annex id="sec-html-document-dot-all-exotic-object-aec">
+      <emu-annex id="sec-IsHTMLDDA-internal-slot-aec">
         <h1>Changes to Abstract Equality Comparison</h1>
         <p>The following steps are inserted after step 3 of the <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> algorithm:</p>
         <emu-alg>
@@ -39930,7 +39930,7 @@ THH:mm:ss.sss
         </emu-alg>
       </emu-annex>
 
-      <emu-annex id="sec-html-document-dot-all-exotic-object-typeof">
+      <emu-annex id="sec-IsHTMLDDA-internal-slot-typeof">
         <h1>Changes to the `typeof` operator</h1>
         <p>The following table entry is inserted into <emu-xref href="#table-35"></emu-xref> immediately preceeding the entry for "Object (implements [[Call]])":</p>
         <emu-table>
--- a/spec.html
+++ b/spec.html
@@ -39924,7 +39924,7 @@ THH:mm:ss.sss
       <emu-annex id="sec-html-document-dot-all-exotic-object-aec">
         <h1>Changes to Abstract Equality Comparison</h1>
         <p>The following steps are inserted after step 3 of the <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> algorithm:</p>
-        <emu-alg start=4>
+        <emu-alg>
           1. If _x_ has an [[IsHTMLDDA]] internal slot and _y_ is either *null* or *undefined*, return *true*.
           1. If _x_ is either *null* or *undefined* and _y_ has an [[IsHTMLDDA]] internal slot, return *true*.
         </emu-alg>
--- a/spec.html
+++ b/spec.html
@@ -13070,6 +13070,14 @@
             </tr>
             <tr>
               <td>
+                Object (undefined-like exotic object)
+              </td>
+              <td>
+                `"undefined"`
+              </td>
+            </tr>
+            <tr>
+              <td>
                 Object (implements [[Call]])
               </td>
               <td>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -4074,6 +4074,8 @@
           1. Return the result of performing Strict Equality Comparison _x_ === _y_.
         1. If _x_ is *null* and _y_ is *undefined*, return *true*.
         1. If _x_ is *undefined* and _y_ is *null*, return *true*.
+        1. If _x_ is an undefined-like exotic object and _y_ *null*, *undefined*, or an undefined-like exotic object, return *true*.
```
UltCombo: > and _y_ *null*

Missing verb ("is").

littledan:
LGTM

spec.html
```diff
@@ -4072,8 +4072,7 @@
       <emu-alg>
         1. If Type(_x_) is the same as Type(_y_), then
           1. Return the result of performing Strict Equality Comparison _x_ === _y_.
-        1. If _x_ is *null* and _y_ is *undefined*, return *true*.
-        1. If _x_ is *undefined* and _y_ is *null*, return *true*.
+        1. If _x_ is either *null*, *undefined*, or an undefined-like exotic object and _y_ is either *null*, *undefined*, or an undefined-like exotic object, return *true*.
```
ljharb: Would this benefit from a new abstract operation that emulates `== null`, and encapsulates all three of these comparisons?

chicoxyzzy:
Isn't it better to use currently inactive proposal Typed Objects (which make operators overloading possible) for this type of things?

spec.html
```diff
@@ -12472,6 +12479,14 @@
                 Implementation-defined. Must not be `"undefined"`, `"boolean"`, `"function"`, `"number"`, `"symbol"`, or `"string"`.
               </td>
             </tr>
+            <tr>
+              <td>
+                Object (undefined-like exotic object)
+              </td>
+              <td>
+                `"undefined"`
```
claudepache: That row must go before the “Object (implements [[Call]]): "function"” row, because `document.all` is callable. :-(

bterlson: Good call, thanks :-P

spec.html
claudepache: The proposed semantics seems bad to me, because it means that document.all’s from distinct frames will be reputed equal. Not unrelatedly, it breaks the invariant `x === y` iff `typeof x == typeof y && x == y`. We should optimise for sanity rather than conciseness:

1. If _x_ is undefined/null and _y_ is undefined/null, return true.
1. If _x_ is a ULEO and _y_ is undefined/null, return true.
1. If _x_ is undefined/null and _y_ is a ULEO, return true.

ljharb: I was thinking `if IsNullish(x) and IsNullish(y), return true.` and define `IsNullish` as "if x is null, undefined, or an ULEO, return true. else, return false".

claudepache: @ljharb In case it was not clear, an abstract operation doesn’t change my “document.all’s from distinct frames appear equal” issue.

evilpie: @claudepache Oh I commented on this before on IRC. Because of Step 1. different objects are always regarded as different, because they have the same type and the strict equality operation doesn't care about "undefined-like".

claudepache: > Because of Step 1. different objects are always regarded as different, because they have the same type and the strict equality operation doesn't care about "undefined-like".

Ah, ok. However, I'm not sure it is good to sacrifice clarity for conciseness. The formulation of the current spec text:
<ol start="2">
<li>If <i>x</i> is <b>null</b> and <i>y</i> is <b>undefined</b>, return <b>true</b>.  
<li>If <i>x</i> is <b>undefined</b> and <i>y</i> is <b>null</b>, return <b>true</b>.
</ol>
has the advantage of not interfering with step 1. So, just adding:
<ol start="4">
<li>If <i>x</i> is an undefined-like exotic object and <i>y</i> is <b>undefined</b> or <b>null</b>, return <b>true</b>.
<li>If <i>x</i> is <b>undefined</b> or <b>null</b> and <i>y</i> is an undefined-like exotic object, return <b>true</b>.
</ol>
seems clearer to me.

bterlson: Agree very much that this is superior given the confusion. Thanks!!

bterlson:
Fixed the issues @claudepache raised.

bterlson:
Review feedback: HTML Document Dot All Exotic Object.
Put in Annex B.

spec.html
```diff
@@ -8159,6 +8161,14 @@
         </emu-alg>
       </emu-clause>
     </emu-clause>
+
+    <emu-clause id="sec-undefined-like-exotic-objects">
+      <h1>Undefined-like Exotic Objects</h1>
+      <p>An <dfn>undefined-like exotic object</dfn> is an exotic object that behaves like undefined in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">typeof operator</emu-xref>. Undefined-like exotic objects have all the ordinary object internal methods and slots.</p>
+      <emu-note>
+        <p>Undefined-like exotic objects are never created by this specification. However, the <a href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-all">document.all object</a> in web browsers is a host-created undefined-like exotic object that exists for web compatibility purposes. There are no other known examples of this type of object.</p>
```
mathiasbynens: Nit: `document.all` → `<code>document.all</code>`

annevk:
It seems the "needs consensus" label can be removed?

littledan:
Well, we have consensus for something, but not for this patch. The consensus is described [in this comment](https://github.com/tc39/ecma262/pull/673#issuecomment-250008273).

annevk:
@bterlson could you maybe up the priority for this a bit? There's a plan to touch `HTMLAllCollection` again (to move over legacycaller from IDL to HTML) and it'd be nice to fix this at the same time.

Also, will the `<dfn>` that's being added get an ID or would that need to be added manually? It'd probably be useful to have something to point to directly.

bterlson:
@annevk I will work on this today. Was really hoping we could get away without this but alas. It'll be kinda finicky to move to Annex B (mostly in turning algorithm changes into Annex B mods) but I should be able to complete it today for your review.

The dfn will link to the parent clause ID without its own ID. Would that work for you?

annevk:
Yeah, that's probably good enough. Still need to figure out exactly how the layering will work.

bterlson:
Rebased, renamed, moved to Annex B with diff pointers. @annevk care to review?

spec.html
```diff
@@ -39904,6 +39904,57 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
...
+          1. If _argument_ is an HTML document dot all exotic object, return *true*.
+          1. Return *false*.
+        </emu-alg>
+      </emu-annex>
+
+      <emu-annex id="sec-html-document-dot-all-exotic-object-aec">
+        <h1>Changes to Abstract Equality Comparison</h1>
+        <p>The following steps are inserted after step 3 of the <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> algorithm:</p>
+        <emu-alg>
+          1. If _x_ is an undefined-like exotic object and _y_ is either *null* or *undefined*, return *true*.
```
michaelficarra: HTML document dot all exotic object

spec.html
```diff
@@ -39904,6 +39904,57 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
...
+          1. If _x_ is either *null* or *undefined* and _y_ is an undefined-like exotic object, return *true*.
```
michaelficarra: Same here.

spec.html
```diff
@@ -39904,6 +39904,57 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
         1. Return ? ForIn/OfBodyEvaluation(|BindingIdentifier|, |Statement|, _keyResult_, ~enumerate~, ~varBinding~, _labelSet_).
       </emu-alg>
     </emu-annex>
+
+    <emu-annex id="sec-html-document-dot-all-exotic-object">
+      <h1>HTML Document Dot All Exotic Object</h1>
+      <p>An <dfn>HTML document dot all exotic object</dfn> is an exotic object that behaves like undefined in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">typeof operator</emu-xref>. HTML document dot all exotic objects have all the ordinary object internal methods and slots.</p>
+      <emu-note>
+        <p>These objects are never created by this specification. However, the <a href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-all"><code>document.all</code> object</a> in web browsers is a host-created HTML document dot all exotic object that exists for web compatibility purposes. There are no other known examples of this type of object.</p>
```
annevk: Maybe add a requirement on hosts?

s/this type of object/and none must be introduced/

annevk:
@domenic do you think this is workable regardless of how we organize IDL/HTML? Would we just say that `HTMLAllCollection` is an HTML document dot all exotic object?

spec.html
```diff
@@ -39904,6 +39904,57 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
...
+      </emu-note>
+
+      <emu-annex id="sec-html-document-dot-all-exotic-object-to-boolean">
+        <h1>Changes to ToBoolean</h1>
+        <p>The result column in <emu-xref href="#table-10"></emu-xref> for an argument type of Object is replaced with the following algorithm:</p>
+        <emu-alg>
+          1. If _argument_ is an HTML document dot all exotic object, return *true*.
+          1. Return *false*.
```
claudepache: In this algorithm, switch “true” and “false”.

spec.html
```diff
@@ -39904,6 +39904,57 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
         1. Return ? ForIn/OfBodyEvaluation(|BindingIdentifier|, |Statement|, _keyResult_, ~enumerate~, ~varBinding~, _labelSet_).
       </emu-alg>
     </emu-annex>
+
+    <emu-annex id="sec-html-document-dot-all-exotic-object">
+      <h1>HTML Document Dot All Exotic Object</h1>
+      <p>An <dfn>HTML document dot all exotic object</dfn> is an exotic object that behaves like undefined in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">typeof operator</emu-xref>. HTML document dot all exotic objects have all the ordinary object internal methods and slots.</p>
```
domenic: > typeof operator

It looks like `typeof` should be `code`ified

> HTML document dot all exotic objects have all the ordinary object internal methods and slots

This isn't quite right, as HTML will define one with overridden internal methods. Not sure what the right phrasing is... "By default, ... (but other specifications may create HTML document dot all exotic objects with overridden internal methods)"?

spec.html
```diff
@@ -39904,6 +39904,57 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
         1. Return ? ForIn/OfBodyEvaluation(|BindingIdentifier|, |Statement|, _keyResult_, ~enumerate~, ~varBinding~, _labelSet_).
       </emu-alg>
     </emu-annex>
+
+    <emu-annex id="sec-html-document-dot-all-exotic-object">
+      <h1>HTML Document Dot All Exotic Object</h1>
```
domenic: "Document Dot All", or `document.all`? I feel like we may have lost something in the minutes from the TC39 meeting :).

Either way is fine, although the latter is slightly less ugly.

spec.html
```diff
@@ -39904,6 +39904,57 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
...
+        <p>The result column in <emu-xref href="#table-10"></emu-xref> for an argument type of Object is replaced with the following algorithm:</p>
+        <emu-alg>
+          1. If _argument_ is an HTML document dot all exotic object, return *true*.
+          1. Return *false*.
+        </emu-alg>
+      </emu-annex>
+
+      <emu-annex id="sec-html-document-dot-all-exotic-object-aec">
+        <h1>Changes to Abstract Equality Comparison</h1>
+        <p>The following steps are inserted after step 3 of the <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> algorithm:</p>
```
domenic: Consider labeling steps and referring to them by labels, as this is kind of fragile. However, it is how things have worked so far, so no need to fix if you'd rather keep as-is.

spec.html
```diff
@@ -39904,6 +39904,57 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
...
+        <emu-alg>
+          1. If _x_ is an undefined-like exotic object and _y_ is either *null* or *undefined*, return *true*.
+          1. If _x_ is either *null* or *undefined* and _y_ is an undefined-like exotic object, return *true*.
+        </emu-alg>
+      </emu-annex>
+
+      <emu-annex id="sec-html-document-dot-all-exotic-object-typeof">
+        <h1>Changes to the typeof operator</h1>
```
domenic: `typeof`, codeified, seems to be the precedent

spec.html
```diff
@@ -39904,6 +39904,57 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
...
+              <th>
+                Type of _val_
+              </th>
+              <th>
+                Result
+              </th>
+            </tr>
+            <tr>
+              <td>
+                Object (HTML dot all exotic object)
```
domenic: xref? I guess we're already in the section due to annex B funtimes, so maybe it doesn't need to be.

spec.html
domenic: Maybe one way to do this would be to state somehow that any object (well, one precise object) can be "marked" as a HTML document dot all exotic object, not that it "is" a HTML dot all exotic object. That way it could be a "legacy platform exotic object" (= using various internal methods defined by Web IDL) plus be marked as a HTML document dot all exotic object, at the same time.

spec.html
bterlson: Nice catch

spec.html
bterlson: This was discussed somewhat, I think there was resistance to "overreach", but I'm definitely fine with the note saying something to this effect.

spec.html
bterlson: HA, maybe! I just assumed it was a deliberately bad editorial choice to underscore the badness of the thing being specified.

spec.html
bterlson: Step labels sure would be nice, but I'll leave this as-is for now.

bterlson:
After talking with @Domenic, I made some changes. Biggie is instead of defining a separate object, I simply say that there exists a internal slot of a particular name and the annex B algorithms check for this internal slot rather than doing a "type check". LMK your thoughts. /cc @annevk

spec.html
```diff
@@ -39904,6 +39904,60 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
         1. Return ? ForIn/OfBodyEvaluation(|BindingIdentifier|, |Statement|, _keyResult_, ~enumerate~, ~varBinding~, _labelSet_).
       </emu-alg>
     </emu-annex>
+
+    <emu-annex id="sec-ishtmldda-internal-slot">
+      <h1>The [[isHTMLDDA]] Internal Slot</h1>
```
domenic: Uppercase to match other internal slot names?

spec.html
```diff
@@ -39904,6 +39904,60 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
...
+      <p>An <dfn>[[isHTMLDDA]] internal slot</dfn> may exist on implementation-defined exotic objects. Objects with an [[isHTMLDDA]] intenral slot behave like undefined in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">typeof operator</emu-xref>.</p>
```
domenic: Typo "intenrel"

typeof still not codeified. Actually just using title="" attribute might work better? For most of these xrefs? Maybe even it's not required at all for ToBoolean?

spec.html
```diff
@@ -39904,6 +39904,60 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
...
+        <emu-alg>
+          1. If _argument_ has an [[isHTMLDDA]] internal slot, return *false*.
+          1. Return *true*.
+        </emu-alg>
+      </emu-annex>
+
+      <emu-annex id="sec-html-document-dot-all-exotic-object-aec">
+        <h1>Changes to Abstract Equality Comparison</h1>
+        <p>The following steps are inserted after step 3 of the <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> algorithm:</p>
+        <emu-alg start=4>
```
bterlson: hmm I thought this did something but it doesn't. Guess I'll delete it.

bterlson:
@domenic ok nits should be fixed

spec.html
```diff
@@ -39904,6 +39904,60 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
         1. Return ? ForIn/OfBodyEvaluation(|BindingIdentifier|, |Statement|, _keyResult_, ~enumerate~, ~varBinding~, _labelSet_).
       </emu-alg>
     </emu-annex>
+
+    <emu-annex id="sec-IsHTMLDDA-internal-slot">
+      <h1>The [[IsHTMLDDA]] Internal Slot</h1>
+      <p>An <dfn>[[IsHTMLDDA]] internal slot</dfn> may exist on implementation-defined exotic objects. Objects with an [[IsHTMLDDA]] internal slot behave like undefined in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">`typeof` operator</emu-xref>.</p>
```
domenic: `*undefined*` I think?

bterlson:
@annevk seem ok to you? Seems like I can't re-request a review.

spec.html
```diff
@@ -39904,6 +39904,60 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
         1. Return ? ForIn/OfBodyEvaluation(|BindingIdentifier|, |Statement|, _keyResult_, ~enumerate~, ~varBinding~, _labelSet_).
       </emu-alg>
     </emu-annex>
+
+    <emu-annex id="sec-IsHTMLDDA-internal-slot">
+      <h1>The [[IsHTMLDDA]] Internal Slot</h1>
+      <p>An <dfn>[[IsHTMLDDA]] internal slot</dfn> may exist on implementation-defined exotic objects. Objects with an [[IsHTMLDDA]] internal slot behave like *undefined* in the <emu-xref href="#sec-toboolean">ToBoolean</emu-xref> and <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> abstract operations and when used as an operand for the <emu-xref href="#sec-typeof-operator">`typeof` operator</emu-xref>.</p>
```
claudepache: It should be “object(s)” rather than “exotic object(s)” (here and everywhere below).

Although the behaviour described in this section is very “exotic” according the common meaning of that word, it does not make an _[exotic object]_ in the sense of the definition given by the spec.

[exotic object]: https://tc39.github.io/ecma262/#sec-exotic-object

annevk: I think it's still fair, since `HTMLAllCollection` is an exotic object (due to custom `[[Call]]` and some other stuff) and it can only be used there.

spec.html
```diff
@@ -39904,6 +39904,60 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
...
+      <emu-note>
+        <p>Objects with an [[IsHTMLDDA]] internal slot are never created by this specification. However, the <a href="https://html.spec.whatwg.org/multipage/obsolete.html#dom-document-all"><code>document.all</code> object</a> in web browsers is a host-created exotic object with this slot that exists for web compatibility purposes. There are no other known examples of this type of object and implementations should not create any with the exception of `document.all`.</p>
```
annevk: Wouldn't it be better to point to `HTMLAllCollection` here? `document.all` is a property.

annevk: Also, why not must not?

annevk:
@bterlson thanks, it does.

spec.html
domenic: I think document.all is better. There's nothing special about HTMLAllCollection itself, really. Its _instances_ are the weird ones. And a good way of collectively referring to all such instances is probably "document.all".

spec.html
bterlson: I think @claudepache is right - the object may be exotic depending on host choices, but it's confusing to say its exotic when it doesn't define anything exotic. Might trigger someone to go sleuthing for hours trying to find the missing exoticness. I'll fix it.

spec.html
bterlson: @annevk I only went with "should" because this is a note and it seems odd to use strong language in a non-normative section.

domenic:
Woohoo, thank you!!

Jack-Works:
To let it evaluate to `false` in every case is harmful to web security. I have a suggestion that keeps the web compatibility meanwhile reduce the harm of falsy `document.all`.

https://github.com/Agoric/proposal-preserve-virtualizability/issues/2#issuecomment-602144216

claudepache:
@Jack-Works See [12.3.6.1](https://tc39.es/ecma262/#sec-function-calls-runtime-semantics-evaluation) step 6 for how direct eval is specced. If you want to do the same thing for `document.all`, you need to add similar logic in the runtime semantics of:

* the `typeof` operator;
* operators based on boolean logic: `||`, `&&`, `? :`, `!`;
* operator allowing comparison with null: `==`;
* statements making use of boolean logic: `if`, `while`, `do/while`, `for`.

(Fortunately, the newer `?.` and `??` operators do not have special treatment of `document.all`.)

Jack-Works:
Yes, but it will no longer be a threat to the security.

claudepache:
In fact, I’m not exactly sure that the approach I had in mind in my previous comment would work or would be the best. But more to the point:

The current PR was specifically to sanction what browsers and the HTML spec were already implementing. That is, they felt it was better that the ES spec provided the necessary hook for the HTML spec, rather than the HTML spec monkey-patched the ES spec.

What you are proposing, is modifying the semantics of document.all in a nontrivial way, which implies implementation work and BC concern. If you feel that it is important, it should be a separate issue, and probably indeed a full-fledged proposal.

PR #676, opened 2016-08-22
Annex B.3.3 use cases have a universally-shared bullet with imprecise language

jswalden:
All three use cases in Annex B.3.3 have the same first bullet point: "One or more FunctionDeclarations whose BindingIdentifier is the name f occur within the function code of an enclosing function g and that declaration is nested within a Block."

And within that bullet point, "that declaration" is somewhat ambiguous as to whether it refer to function `g` or to those `FunctionDeclaration`s.

Make the spec text apply to all three use cases, and use more-precise language to refer to those `FunctionDeclaration`s as antecedent.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -36405,17 +36405,18 @@ THH:mm:ss.sss
     <!-- es6num="B.3.3" -->
     <emu-annex id="sec-block-level-function-declarations-web-legacy-compatibility-semantics">
       <h1>Block-Level Function Declarations Web Legacy Compatibility Semantics</h1>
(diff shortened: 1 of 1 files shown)


littledan:
LGTM, @bterlson ?

spec.html
```diff
@@ -36405,18 +36405,19 @@
...
+      <p>Prior to ECMAScript 2015, the ECMAScript specification did not define the occurrence of a |FunctionDeclaration| as an element of a |Block| statement's |StatementList|. However, support for that form of |FunctionDeclaration| was an allowable extension and most browser-hosted ECMAScript implementations permitted them. Unfortunately, the semantics of such declarations differ among those implementations. Because of these semantic differences, existing web ECMAScript code that uses |Block| level function declarations is only portable among browser implementation if the usage only depends upon the semantic intersection of all of the browser implementations for such declarations.</p>
+      <p>These are the threshold requirements for the semantic intersection to potentially occur:</p>
```
bterlson: I find this language complex to parse and understand. What was wrong with the old language?

bterlson: NM I see what you're doing here.

bterlson:
Pretty sure this is saying the same thing. @allenwb can you confirm?

bterlson:
After discussion with @allenwb, we're pretty agreed that this non-normative prose is inadequate to describe the intersection semantics (which are precisely defined by B.3.3.1-B.3.3.5). A better fix here would be to just give some illustrative examples rather than trying to proseify complex algorithm steps. I can work on this.

PR #677, opened 2016-08-22
Annex B.3.3.1's monkeypatching of FunctionDeclarationInstantiation is out of date

jswalden:
It refers to "BoundNames of `argumentsList`", but `argumentsList` is now an actual value, not a grammar production.  I suspect the name used to be correct, then someone renamed `argumentsList` to `formals`, and things got skew.  Now `parameterNames` exists as the "BoundNames of `formals`", so it's the best thing to use.

I also went and added big WARNING comments next to all the monkey-patched spec algorithms, so maybe renames of stuff will remember to update Annex B as well.  There were already NOTE: steps that theoretically should have been enough reminder to the spec-modifier, but I guess they weren't.  It seems to me like a comment at the start of the algorithms is most likely to be read by someone skimming the algorithms to see where to do variable renames, so it's worth adding even atop the NOTEs.


jmdyck:
In ES6 rev33, it said "... BoundNames of FormalParameters of g".
In rev34, it changed to "... BoundNames of argumentsList".

The change appears to relate to https://bugs.ecmascript.org/show_bug.cgi?id=3490#c6
> Since we know we are dealing with a FunctinoDeclaration we know that argumentsList is a FormalParameters

littledan:
LGTM. This patch illustrates one of the hazards of having normative optional text out of line. I wonder if we could have it inline along the lines of what we agreed for 402 (but have not yet implemented) at https://github.com/bterlson/ecmarkup/issues/96 .

bterlson:
I'm not sure inline normative-optional would help much - you still have to look around your location and read everything to ensure you haven't broken anything. Comments are at least as good of a solution IMO. Perhaps there is a better way to solve this with some additional markup and/or tooling... will ponder.

PR #679, opened 2016-08-30
(10).toPrecision(1) to give "1e+1", not "1.e+1"

Yaffle:
It is strange, but web browsers do this correctly, so may be I am misinterpreting the spec text.


bterlson:
Seems fine, thanks @Yaffle!

bakkot:
In the spirit of being really pedantic about making sure all normative changes have test262 collateral, I feel the need to note for posterity that test262 was already (and [still is](https://github.com/tc39/test262/blob/6a0f1189eb00d38ef9760cb65cbc41c066876cde/test/built-ins/Number/prototype/toPrecision/exponential.js#L27)) testing for the behavior given by this PR, not what the spec said previously.

PR #682, opened 2016-08-30
Meta: Update ecmarkup version dependency

mathiasbynens:
Note that the Travis configuration has been updated accordingly.

Ref. https://github.com/bterlson/ecmarkup/pull/101.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,6 @@ env:
 language: node_js
 
 node_js:
-  - "5"
+  - "6"
 
 sudo: false
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "license": "SEE LICENSE IN https://tc39.github.io/ecma262/#sec-copyright-and-software-license",
   "homepage": "https://tc39.github.io/ecma262/",
   "dependencies": {
-    "ecmarkup": "^3.3.2"
+    "ecmarkup": "^3.4.0"
   },
   "devDependencies": {
     "@alrra/travis-scripts": "^2.0.0"


.travis.yml
```diff
@@ -8,6 +8,6 @@ env:
 language: node_js
 
 node_js:
-  - "5"
+  - "6"
```
ljharb: If you want to always point to latest, this can be "node"

michaelficarra: I prefer `stable`.

ljharb: "stable" has no meaning in semver (and thus in node/io.js >= 1.0), where everything is stable.

domenic: It is short for "latest stable", which does have meaning.

ljharb: Sure, but the latest not-stable is 0.11 - so it's mostly meaningless. Either way, nvm is going to remove "stable" at some point, so "node" is a safer long-term choice.

michaelficarra: @ljharb link?

ljharb: I'm the sole maintainer of nvm and I intend to do it.

domenic: You will break so many builds it will make left-pad look like a joke.

UltCombo: @ljharb IMHO "node" is even more ambiguous than "stable". As far I can see, "node" could read as any Node.js version.

domenic: Yeah, I definitely don't want to test against Node.js version "node" ^_^

leobalter: while `latest` is a nice terminology and maybe better than just "node", I agree `stable` is ambigous, include LTS versions to the options I would wonder for it.

If there's a chance for breaking changes, link `stable` to any stable release and do not update it anymore. It won't break anything and you'll have your deprecated alias.

ljharb: All of this is delightful discussion to have on the nvm repo when the time comes. For now, I'm suggesting "node" if we don't want to have to manually bump it to "7" etc in the future. If we don't like that, so be it.

UltCombo: Back on topic, I believe testing against "stable"/"node"/LTS is an anti-pattern.

It means CI will unsolicitedly bump the major Node version whenever it becomes available without any previous test or knowledge from the authors. Major versions indicate breaking changes, and thus should be throughout tested (e.g. via a PR bumping the major version, like this one) in order to prevent breaking the build overnight.

ljharb: Fair counterpoint.

michaelficarra: @UltCombo I agree. We can have it be in the allowed failures so that, if a new version of node has incompatibilities, we will be notified but CI will not fail.

bterlson:
Looks good - I am fine with explicitly naming version 6 to test with. Note that ecmarkup 3.4 requires version 6 now.

PR #685, opened 2016-09-08
Normative: Clarify rules around super inside eval

bterlson:
Previously, `super` was allowed inside eval if eval was called inside
any function and disallowed inside arrow functions. This semantics
likely predates the eviction of super from functions during ES6 and was
simply not updated. This PR updates the relevant semantics to align more
closely with both user expectations and implementation reality.

With this PR, SuperProperty is only allowed inside eval if the current
this environment is a function environment record with a home object.
This means SuperCall is allowed in eval (and nestings thereof) in arrow
functions in methods per implementation reality. Other usages of
SuperProperty inside eval is an early error.

Additionally, SuperCall is only allowed inside eval if the current this
environment is a function environment record with a home object and a
[[Construct]] field. Since methods do not have [[Construct]] slots, this
has the effect of only allowing SuperCall inside eval inside
constructors (or inside another eval in a constructor, or inside an
arrow function in a constructor) per implementation reality. Other
usages of SuperCall inside eval is an early error.

Lastly, neither SuperProperty or SuperCall are allowed inside eval
inside a non-method function. This is an early error. This aligns with
implementation reality with the exception of Chakra.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19460,7 +19460,7 @@
       <emu-grammar>ScriptBody : StatementList</emu-grammar>
       <ul>
         <li>
-          It is a Syntax Error if |StatementList| Contains `super` unless the source code containing `super` is eval code that is being processed by a direct eval. Additional early error conditions for `super` within direct eval are defined in <emu-xref href="#sec-performeval"></emu-xref>.
+          It is a Syntax Error if |StatementList| Contains `super` unless the source code containing `super` is eval code that is being processed by a direct eval. Additional early error rules for `super` within direct eval are defined in <emu-xref href="#sec-performeval"></emu-xref>.
         </li>
         <li>
           It is a Syntax Error if |StatementList| Contains |NewTarget| unless the source code containing |NewTarget| is eval code that is being processed by a direct eval that is contained in function code that is not the function code of an |ArrowFunction|.
--- a/spec.html
+++ b/spec.html
@@ -19460,7 +19460,7 @@
       <emu-grammar>ScriptBody : StatementList</emu-grammar>
       <ul>
         <li>
-          It is a Syntax Error if |StatementList| Contains `super` unless the source code containing `super` is eval code that is being processed by a direct eval that is contained in function code that is not the function code of an |ArrowFunction|.
+          It is a Syntax Error if |StatementList| Contains `super` unless the source code containing `super` is eval code that is being processed by a direct eval. Additional early error conditions for `super` within direct eval are defined in <emu-xref href="#sec-performeval"></emu-xref>.
         </li>
         <li>
           It is a Syntax Error if |StatementList| Contains |NewTarget| unless the source code containing |NewTarget| is eval code that is being processed by a direct eval that is contained in function code that is not the function code of an |ArrowFunction|.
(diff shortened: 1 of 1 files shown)


bterlson:
FWIW, new.target needs similar treatment.

allenwb:
except that new.target, unlike super,  is legal in all kinds of functions

ljharb:
Can you confirm what `new.target` does in generators, async, and arrow functions?

bterlson:
Indeed, I just mean that the current ScriptBody static semantics aren't quite correct at least in terms of eval in arrow functions.

littledan:
`new.target` aside, I'm a big fan of this PR. I just noticed that V8 violates the spec here ([bug](https://bugs.chromium.org/p/v8/issues/detail?id=5326)) but I like our current semantics better. There are some test262 tests that get at just this corner case. There are some tests that will need to be updated to expect a SyntaxError rather than a ReferenceError. Minor: This also gives a spec text pattern/precedent towards [banning eval in private state](https://github.com/tc39/proposal-private-fields/issues/47).

bterlson:
@ljharb new.target is allowed in all those cases. Generators and async functions it returns undefined and in arrow functions it returns the outer function's new.target.

@littledan I was planning to address new.target in a subsequent PR. Would that be ok with you?

Also do we think this needs-consensus? I'm of the opinion that this is basically a bug fix and am happy to pull this in, but I'll wait if we think it needs committee time.

allenwb:
@littledan 
It seem to me that the basic principle we should be follow is that for any parse-able string _S_ that does not include any declaration the observable semantics of a direct  eval("_S_") is exactly the same as the semantics of _S_.

Declarations are a special case because of the extra scope they introduce, but aside from scoping related issues the principle should also apply to all other aspects of such eval texts.

littledan:
@allenwb We don't follow that for `function.sent`. It is inaccessible inside an eval as it loses the [+Yield] parameter.

allenwb:
@littledan 
presumably also an issue for "yield" itself?  Also, it appears to me that the current spec (starting with ES2015) disallows `function () {eval("return 42;")}`.  If in correct, that is a breaking change (pretty sure unintentional) from previous editions.

These seems more of an issue relating to the specification mechanism (contextually parameterizing the grammar) rather than the most desirable semantics.  It suggests that step 3 of [PerformEval](https://tc39.github.io/ecma262/#sec-performeval)  need to be reworked for an direct eval to correctly parameterize the parse goal symbol with the grammar parameters that were active at the direct eval call site.  Note that the yield grammar parameter is available in [_CallExpression_ Evaluation](https://tc39.github.io/ecma262/#sec-function-calls-runtime-semantics-evaluation) which is where all direct eval invocations of [PerformEval](https://tc39.github.io/ecma262/#sec-performeval) originate. For the return grammar parameter it may be easier to dynamically determine it based upon the active environment at the direct eval site.

bterlson:
I don't recall return ever working in eval... it certainly doesn't today in all implementations so I don't think we have any compat worries here.

Though I totally agree that, at least for expressions, Expr and eval("Expr") should be identical. I would be in favor of plumbing Yield and Await through eval.

littledan:
Well, the difference with `function.sent` and `return`/`yield` is that the latter pair transfers control. IMO it'd be nice to not add more complexity for `eval` with `yield`/`await` until we have a concrete use case that would benefit. I think statement vs expression is a red herring; you can have statements inside `eval`. Anyway, there are separate [arguments](https://github.com/tc39/proposal-private-fields/issues/47) against exposing private state through to `eval`.

bterlson:
@littledan yeah I see the control transfer argument - if yield/await work, why not return (and I don't think return should work, but maybe I could be convinced)? But having a single meta-property not work inside eval seems strange and unexpected.

I'm not too comfortable with the "concrete use case" line of reasoning. What is the use case for allowing async functions or any other new feature in eval?

ajklein:
Neither `break` nor `continue` are allowed in an eval either. I don't think this is a can of worms worth opening.

bterlson:
Expanding on Allen's Principle to incorporate this notion:

For any parse-able string S that does not include any declaration <ins>or control transfer</ins> the observable semantics of a direct eval("S") is exactly the same as the semantics of S.

allenwb:
>I don't recall return ever working in eval

See ES5.1 http://ecma-international.org/ecma-262/5.1/#sec-12.9 (exact same language in ES3)

> An ECMAScript program is considered syntactically incorrect if it contains a __return__ statement that is not within a _FunctionBody_.

Note that "FunctionBody" is italic while "program" is not and the terminology "__return__ statement" is used rather than _ReturnStatement_. So, it's hard to tell whether or not this was intended as general syntactic restriction on parsing _Program_ and whether or not it applies to a direct eval that is contained in a _FunctionBody_. 

Equally vague language is used to talk about __break__ and __continue__.

I agree, that this language could be construed to make *return*, *break*, and *continue* illegal in eval code.  However, it isn't clear whether that was the actual intent of the language. 

I'll still argue that the best user conceptual model for the semantics of an _ExpressionStatement_  consisting of a direct eval is as if the eval code was substituted for the _ExpressionStatement_ (within { } for strict mode).
We should only deviated from the model for compelling reasons.

bterlson:
Got consensus on this item today.

anba:
`new class { constructor() { eval("super()"); } }` currently throws a runtime TypeError instead of an early SyntaxError. Is this intended semantics?

ljharb:
Would you expect that to throw an early error but not `new class { constructor() { eval(fnThatReturnsSuperAsString() + "()"); } }`?

anba:
I meant an early error in PerformEval.

littledan:
@anba To be clear, the issue in https://github.com/tc39/ecma262/pull/685#issuecomment-274902837 was fixed by your patch f0bd767c, right?

anba:
Yes, I think so.

PR #686, opened 2016-09-09
Normative: Clarify rules around NewTarget in eval

bterlson:
Previously, NewTarget was disallowed in nested eval contexts and in eval inside arrow functions. This change allows NewTarget inside eval anywhere it would be allowed outside of eval.

This PR is on top of #685 as it uses a similar pattern.


bterlson:
Looks like I forgot to take this with #685. Fixing merge conflicts...

Issue #683, opened 2016-09-02
OrdinarySetPrototypeOf and the HTML spec


saambarati:
Currently, OrdinarySetPrototypeOf bails from its cycle checking loop early if it sees a [[GetPrototypeOf]] which isn't the default implementation. I'm guessing this is in awareness of the Proxy's [[GetPrototypeOf]]. Is that correct? If not, what is the reasoning behind this decision?
I'm looking at: https://tc39.github.io/ecma262/#sec-ordinarysetprototypeof

The HTML spec overrides [[GetPrototypeOf]] on the window proxy. Calling [[SetPrototypeOf]] on an object who's [[Prototype]] is WindowProxy can lead to cycles now.
I'm looking at: https://html.spec.whatwg.org/#windowproxy-getprototypeof

Like so:
```
let o = {__proto__: window};
window.__proto__.__proto__.__proto__.__proto__ = o
```
(I believe this throws cycle exceptions in browsers, however, it should not according to the spec, if I'm reading it correctly, and have constructed my example correctly.)

I believe that if we don't consider the browser, it's impossible to get a cycle if you directly loop over the [[Prototype]] property. However, with the HTML spec, I believe that's no longer true.

I wonder if the function can be more restrictive, and only bail out on the loop if we encounter a ProxyObject.[[GetPrototypeOf]] internal method, instead of bailing once we see the non-default [[GetPrototypeOf]].

What are people's thoughts? I'm mostly posting here to bring awareness to this issue and to understand previous discussions of it, or to have new discussions about it. We're currently running into issues with this property inside WebKit.

domenic:
I think @jswalden also ran into this, judging by https://blog.whatwg.org/windowproxy-window-and-location#comments ...

(I have not taken the time to form an opinion on this. In general I think anything that is web-compatible is probably fine in these kind of edge cases?)

littledan:
Cc @verwaest

allenwb:
see https://bugs.ecmascript.org/show_bug.cgi?id=2437 for the background of this semantics

It seems perfectly reasonable for the HTML spec to provide modifications to   steps of 8.c.i-ii of OrdinarySetPrototypeOf  that take the WindowProxy object into account

bterlson:
The rationale wasn't entirely proxy-based - generally, exotic objects can have a [[GetPrototypeOf]] implementation that can do arbitrary things and it's impossible to enforce that there are no observable circularities in this case. Proxies are an obvious case where this can happen however.

[...]

PR #688, opened 2016-09-12
Layering: add HostObjectDefinePropertyReturnFalse

### LINKED ISSUE #672: WindowProxy and violations of internal method invariants
Basically, the problem raised by @bzbarsky in https://esdiscuss.org/topic/figuring-out-the-behavior-of-windowproxy-in-the-face-of-non-configurable-properties two years ago now was resolved, but the resolution could not be implemented per https://bugzilla.mozilla.org/show_bug.cgi?id=1197958#c4 and therefore the HTML Standard's `WindowProxy` object definition contradicts ECMAScript here after I discussed the situation with bz and @ajklein.

See https://html.spec.whatwg.org/multipage/browsers.html#windowproxy-getownproperty (step 3) and https://html.spec.whatwg.org/multipage/browsers.html#windowproxy-defineownproperty (also step 3) for the violations.

(I missed this when reporting the document.all violation as this one was not properly tagged in HTML.)

domenic:
This allows host environments which need to override the Object.defineProperty behavior, for legacy compatibility, to preserve invariants while avoiding breaking web applications that depend on not-throwing when defining non-configurable, non-writable properties on WindowProxy. This does not alter the behavior of Reflect.defineProperty or [[DefineOwnProperty]].

Closes #672 (from the ES spec side). See also https://bugzilla.mozilla.org/show_bug.cgi?id=1197958.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -22679,9 +22679,20 @@
           1. If Type(_O_) is not Object, throw a *TypeError* exception.
           1. Let _key_ be ? ToPropertyKey(_P_).
           1. Let _desc_ be ? ToPropertyDescriptor(_Attributes_).
+          1. If ! HostObjectDefinePropertyReturnFalse(_O_, _key_, _desc_) is *true*, return *false*.
           1. Perform ? DefinePropertyOrThrow(_O_, _key_, _desc_).
           1. Return _O_.
         </emu-alg>
+
+        <emu-clause id="sec-hostobjectdefinepropertyreturnfalse" aoid="HostObjectDefinePropertyReturnFalse">
+          <h1>HostObjectDefinePropertyReturnFalse( _O_, _key_, _desc_ )</h1>
+
+          <p>HostObjectDefinePropertyReturnFalse is an implementation-defined abstract operation that allows host environments to override the usual failure behavior of `Object.defineProperty` under certain circumstances necessary for legacy code.</p>
+
+          <p>An implementation of HostObjectDefinePropertyReturnFalse must complete normally, with a completion value of either *true* or *false*. An implementation of HostObjectDefinePropertyReturnFalse must only return *true* for implementation-provided exotic objects, and only when the normal behavior of using DefinePropertyOrThrow(_O_, _key_, _desc_) would cause legacy code to fail.</p>
+
+          <p>The default implementation of HostObjectDefinePropertyReturnFalse is to unconditionally return *false*.</p>
+        </emu-clause>
       </emu-clause>
 
       <!-- es6num="19.1.2.5" -->


allenwb:
Things like this exist to deal with a single specific situation and don't warrant generalization into abstract operations that might be applied in other situations.  I think we should avoid adding attractive nuisance hooks to the ES spec. that could be taken as a general invitation for hosts to do unanticipated bizarre things. In such cases some carefully chosen prose is better than adding additional abstract operations. 

I think a cleaner way to allow for the specific problem this change is intended to address (see https://github.com/tc39/ecma262/issues/672) to add a new paragraph to https://tc39.github.io/ecma262/#sec-error-handling-and-language-extensions that says something like:
 
"An implementation my define failure behavior other than throwing TypeError for Object.defineProperty when it's first argument is an implementation provided exotic object and the throwing behavior would cause legacy code to fail."

The HTML spec. does not need to algorithmicallhy  "call-back" into the ES spec. All it needs to do is defined the circumstances where it returns **false** rather than throwings.

domenic:
Right. As I stated, I think this is a clearer way of doing this, and the normative requirements on host serve to discourage the general invitation you're worried about.

spec.html
```diff
@@ -22679,9 +22679,20 @@
...
           1. Perform ? DefinePropertyOrThrow(_O_, _key_, _desc_).
           1. Return _O_.
         </emu-alg>
+
+        <emu-clause id="sec-hostobjectdefinepropertyreturnfalse" aoid="HostObjectDefinePropertyReturnFalse">
+          <h1>HostObjectDefinePropertyReturnFalse( _O_, _key_, _desc_ )</h1>
+
+          <p>HostObjectDefinePropertyReturnFalse is an implementation-defined abstract operation that allows host environments to override the usual failure behavior of `Object.defineProperty` under certain circumstances necessary for legacy code.</p>
+
+          <p>An implementation of HostObjectDefinePropertyReturnFalse must complete normally, with a completion value of either *true* or *false*. An implementation of HostObjectDefinePropertyReturnFalse must only return *true* for implementation-provided exotic objects, and only when the normal behavior of using DefinePropertyOrThrow(_O_, _key_, _desc_) would cause legacy code to fail.</p>
```
bterlson: s/for implementation-provided exotic objects/when _O_ is an implementation-provided exotic object/g.

bterlson:
I think this is needs-consensus in the vein of #673.

Also it's not purely "layering", so should be marked Normative (HostEnsureCanCompileStrings should have been as well, whoops).

"only for legacy code" prose doesn't seem like a realistic normative requirement, so in practice any ecmascript implementation may adopt this behavior. Language similar to what I propose in https://github.com/tc39/ecma262/issues/668#issuecomment-240795846 is probably better for legacy hacks.

domenic:
Hmm OK, I guess that's fair; I hadn't quite realized that this is technically a normative change. I'll update the PR and prepare a TC39 presentation for this, #673, and maybe #683 (for which I will also create a PR).

Personally I think "only for legacy code" is a pretty reasonable normative requirement; it's meant to prevent hosts from taking this as a license to do weird things. It's similar to how in Web IDL we've started renaming things "legacyX" to indicate that you shouldn't use them. I think this is a bit nicer than explicitly referencing HTML (even non-normatively). But if you prefer such an explicit reference then that works for me. Nit: I'd prefer not referencing a section number since that just creates something you need to keep updated, so I'd probably say just "HTML" or "HTML's obsolete features section".

michaelficarra:
This was [discussed at the Sept meeting](https://github.com/tc39/tc39-notes/blob/master/es7/2016-09/sept-27.md#hostobjectdefinepropertyreturnfalse). Conclusion appears to be that this goes in Annex B, mentioning HTML explicitly.

annevk:
It seems the "needs consensus" label can be removed?

littledan:
Would anyone like to do this editorial change, to put the new spec text in Annex B as suggested in TC39? It'd be great to get this patch landed and enable the fix to https://github.com/tc39/ecma262/issues/672 .

domenic:
@annevk @bzbarsky, do you remember whether this is sufficient to fix the violations? I believe we ended up needing to do a different solution (and I believe Firefox implemented a different solution) as this one was not sufficiently web-compatible.

If I recall, the violations in question are:

- https://html.spec.whatwg.org/#windowproxy-getownproperty:willful-violation
- https://html.spec.whatwg.org/#windowproxy-defineownproperty:willful-violation

domenic:
Digging around Firefox bugs my read is that nobody has yet tried to ship the proposed change (i.e., this ecma262 change + the corresponding HTML spec change which doesn't exist yet). See https://bugzilla.mozilla.org/show_bug.cgi?id=1329324 which hasn't seen any movement.

annevk:
Yeah, it seems this would require an implementer to test the waters.

bzbarsky:
Yeah, on the Firefox side I haven't had time to prod this along.   I will try to do that now.

bzbarsky:
It doesn't help that the lack of an actual spec on the ES side means https://bugzilla.mozilla.org/show_bug.cgi?id=1329321 totally didn't implement the infrastructure I _actually_ need for WindowProxy.  So I can't even experiment with this yet to see whether it's web-compatible...  I'll see what I can do to prod _that_ along.

bzbarsky:
And in particular, the actual behavior for Window still needs to be sorted out, of course.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1178638#c13 and following.

evilpie:
I wonder if returning `null` instead of `false` would be a bit more consistent with APIs like Object.getPrototypeOf that return either an object or null. I guess `false` was chosen because it matches Reflect.defineProperty?

bzbarsky:
OK, so I am actually not quite sure what exact behavior we are aiming for here.  Specifically,  consider this case:

    Object.defineProperty(window, "foo", { value: 5, configurable: false });
    Object.defineProperty(window, "foo", { value: 6 });

Should that second call succeed?  What about:

    Object.defineProperty(window, "foo", { value: 5, configurable: false });
    Object.defineProperty(window, "foo", { value: 6, configurable: true });

And what about:

     var foo = 5; // Defines a non-configurable property on the Window
     Object.defineProperty(window, "foo", { value: 6 });

and 

     var foo = 5; // Defines a non-configurable property on the Window
     Object.defineProperty(window, "foo", { value: 6, configurable: true });

Finally, what about:

    var foo; // Defines a non-configurable property on the Window
    window.foo = 5;

This last one is interesting, because https://html.spec.whatwg.org/multipage/window-object.html#windowproxy-set calls https://tc39.github.io/ecma262/#sec-ordinaryset which calls <https://tc39.github.io/ecma262/#sec-ordinarysetwithowndescriptor>.  And this does a [[DefineOwnProperty]] call which is basically identical to `Object.defineProperty(window, "foo", { value: 5 });`.  And if we force that descriptor to be non-configurable, then the assignment will fail, which is clearly not web-compatible...

@domenic @annevk

annevk:
@erights ^^

allenwb:
> OK, so I am actually not quite sure what exact behavior we are aiming for here. Specifically, consider this case:
> 
> ```
> Object.defineProperty(window, "foo", { value: 5, configurable: false });

note that writeable:false is implicit in the above descriptor (assume "foo" doesn't already exist)
> Object.defineProperty(window, "foo", { value: 6 });
> ```
> Should that second call succeed? What about:

No because, "foo" property is non-writable




> 
> ```
> Object.defineProperty(window, "foo", { value: 5, configurable: false });
> Object.defineProperty(window, "foo", { value: 6, configurable: true });
> ```

fails because configurable attribute of a property can not be changed from false to true

> And what about:
> 
> ```
>  var foo = 5; // Defines a non-configurable property on the Window
>  Object.defineProperty(window, "foo", { value: 6 });
> ```

Depends upon whether or not `window` is the global objects. If it is, succeeds because assignment to the global object creates a _writable_ non-configurable property. If it isn't (eg, b/c windowproxy) it depends upon how assignment to the actual global object is propagated to the `window` object.  So, depends upon HTML spec.

> and
> 
> ```
>  var foo = 5; // Defines a non-configurable property on the Window
>  Object.defineProperty(window, "foo", { value: 6, configurable: true });
> ```
Same as above, if `window` is the global object then would fail because can't change configurable from false to true.

> Finally, what about:
> 
> ```
> var foo; // Defines a non-configurable property on the Window
> window.foo = 5;
> ```

as above, but basically just demonstrates that windowproxy creates writable properties on `window`

> This last one is interesting, because https://html.spec.whatwg.org/multipage/window-object.html#windowproxy-set calls https://tc39.github.io/ecma262/#sec-ordinaryset which calls https://tc39.github.io/ecma262/#sec-ordinarysetwithowndescriptor. And this does a [[DefineOwnProperty]] call which is basically identical to `Object.defineProperty(window, "foo", { value: 5 });`. And if we force that descriptor to be configurable, then the assignment will fail, which is clearly not web-compatible...
> 

You need to be careful in talking about [[DefineOwnProperty]] equivalents to remember that incomplete descriptors default missing attributes to false when used to define a new property.  For existing properties missing attributes default to " no change".

bzbarsky:
> No because, "foo" property is non-writable

OK, assume both descriptors set `writable: true`.

> fails because configurable attribute of a property can not be changed from false to true

Well, does `Object.defineProperty(window, "foo", { value: 5, configurable: false });` actually create a non-configurable property?  That is one big open question here!  The property it creates is supposed to be reported as configurable via `getOwnPropertyDescriptor`.  The question is how it should behave for `defineProperty` purposes.

 > Depends upon whether or not window is the global objects

`window` is the `WindowProxy`, of course.  There is no direct access to the `Window` (which is the global object) in the web platform.   I'd really appreciate us focusing on the actual context  here instead of spending a lot of space on irrelevant hypotheticals; the issue is complicated enough as it is, without confusing it by considering cases that don't exist.

> it depends upon how assignment to the actual global object is propagated to the window object

It's not; the propagation is in the opposite directly.  The property `var` defines lives on the global object.  The `window` object forwards some operations to that global.

> So, depends upon HTML spec.

To some extent.  Part of the question here is what behavior is desired here, or in other words what exactly we want the HTML spec to say.

> You need to be careful in talking about [[DefineOwnProperty]] equivalents to remember that incomplete descriptors default missing attributes to false

Yes, I am well aware of this.

Again, to make this clear: We do not want to _report_ properties as being non-configurable on `WindowProxy`, because they can always get reconfigured during navigation.   When someone tries to define a non-configurable property on the `WindowProxy`, we want to report that we failed to do that, but still define the property on the underlying `Window`.  The open questions are:

1) Should the attempt to define on the underlying `Window` define the property as configurable or not?  Does this depend on whether an explicit `configurable` attribute was supplied in  the descriptor?
2) Does this depend on whether the `Window` already has a property of that name and on its configurability state?

My understanding of earlier discussion was that we wanted to coerce properties to be _defined_ as configurable on the `Window`, but that causes problems with the  "var followed by assignment" case, because the assignment would throw.  So I am trying to figure out what behavior we _actually_ want here that would be least-surprise.

If the answer is just "use the given descriptor as-is, return false (in which cases?), and have `getOwnPropertyDescriptor` always report configurable even if this does not match the observed behavior of defineProperty in some cases", that's fine.   But I want to make sure we're all on the same page for the behavior we're aiming to implement.

allenwb:
OK, I getting swapped back in and now understand the basis of your questions. `window` is an exotic object, but still must conform to the essential property invariants. So, cutting to the chase...

> 
> If the answer is just "use the given descriptor as-is, return false (in which cases?), and have 

You don't have to return false when defining a new property requests configurable:false even if you are going to ignore that request.  You can ignore the configurable: false and still return true. The invariant for [[DefineOwnProperty]] only requires returning false return for changing a non-configurable property to configurable.

>`getOwnPropertyDescriptor` always report configurable even if this does not match the observed behavior of defineProperty in some cases", that's fine. But I want to make sure we're all on the same page for the behavior we're aiming to implement.

Yes, that sounds fine to me.  The window exotic object internal methods should maintain a shadow copy of the configurable value passed to [[DefineOwnProperty]] and use the shadow to control what is allowed. Except that [[GetOwnPropertyDescriptor]]  always reports configurable: false. 

Also remember that `undefined`, `NaN`, and `Infinity` are specified as {configurable: false, writable: false} and [[GetOwnPropertyDescriptor]] should report that.

bzbarsky:
> You don't have to return false when defining a new property requests configurable:false even if you are going to ignore that request.

Actually, signaling that the request to create a new non-configurable property was ignored is  literally the entire basis for this whole discussion.   The people who are relying on the essential property invariants say they need to know when their attempt to define a non-configurable property didn't actually do that.

> Also remember that undefined, NaN, and Infinity are specified as {configurable: false, writable: false} and [[GetOwnPropertyDescriptor]] should report that.

Hmm.  I guess that would be OK in the sense that their property descriptors do not observably change on navigation (though they can go from returning a property descriptor to throwing on access).  But it would somewhat slow down property access if we need to check for those three special cases, and we'd need to figure out whether there are other such special cases as well...

erights:
There certainly is a lot of context to reload, and it is important to get all the simultaneous constraints right. I admit that my memory of all this is rusty and needs a refresh.
Can we arrange a conf call (hangout or something) to discuss this?

bzbarsky:
I can do that, in general...  Whom do we need in the room?

allenwb:
> Actually, signaling that the request to create a new non-configurable property was ignored is literally the entire basis for this whole discussion. The people who are relying on the essential property invariants say they need to know when their attempt to define a non-configurable property didn't actually do that.

But the specification of the essential invariants has never promised that.  The promise is that if a property has been observed to be non-configurable (and non-writable) its attributes (including value if non-writable + special exemption for writable to non-writable transactions) will never observably change.  

If a caller needs to verify that that setting configurable to true actually worked they can check it via [[GetOwnPropertyDescript]] after the define call.  The essential invariants guarantee that if [[GetOwnPropertyDescriptor]] says configurable is false then it can never observably change to true. 

> Hmm. I guess that would be OK in the sense that their property descriptors do not observably change on navigation (though they can go from returning a property descriptor to throwing on access). But it would somewhat slow down property access if we need to check for those three special cases, and we'd need to figure out whether there are other such special cases as well...

This is currently a requirement of the spec.  At least I consider the fact the the spec. says the the attributes of those properties are {writable: false, configurable: false}  is an a priori  observation of the configurable property so it may never change to configurable: true.

I would also hope that implementations already take advantage of the fact that if they aren't shadowed their constant values can be assumed.

bzbarsky:
> Sure, for barewords.

In Firefox.   I think in some other browsers even barewords go through the `WindowProxy` in some cases, in inconsistent ways.

bzbarsky:
> When a frame is navigated, the same windowProxy is also the global for the fresh JS environment of the new frame.

The global for the JS environment is what you call the _hiddenWindow_.  The _windowProxy_ itself stores no state other than what the current _hiddenWindow_ is; all other state (e.g. the names and values of global variables) is stored on the _hiddenWindow_.  The  _windowProxy_ has internal methods that call the internal methods of  _hiddenWindow_ in some cases.

> Thus, when you said "non-configurable on Window", I interpret that to mean "configurable, but refuses configuration".

That's an interesting question.   There is no way to directly query property descriptors from _hiddenWindow_, so there's no obvious way to tell these apart from script, right?

In practice in browsers, `var` defines an actual non-configurable property on _hiddenWindow_.

> We explain this as if it creates a non-configurable property on the hiddenWindow, though no one can observe this

Makes sense.  That's what browsers already do, good.

> The second call must fail with a thrown exception, but only because of the value change.

Hmm.  But we were modeling this situation as a non-configurable property on the _hiddenWindow_.  Seems to me that even without a value change this should throw an exception, because it looks like it's trying to redefine from a non-configurable to a configurable property...

> Here, the second call must succeed, because it did not ask for an attribute setting that conflicts with the current attribute setting. 

[[DefineOwnProperty]] on _windowProxy_ can just decide to throw for any reason it wants, in theory.  

But ok, if we posit that we want the behavior you describe, what is the actual behavior of [[DefineOwnProperty]]  on _windowProxy_ to accomplish this, given that there _is_ in fact a non-configurable property on the _hiddenWindow_ named "foo"?

Same concern applies to this example:

    var foo = 5; // Defines a non-configurable property on the Window
    Object.defineProperty(window, "foo", { value: 6, configurable: true });

Let me ask another question, which may tease out what you mean by "refuses to be configured".  If I do:

    var foo = 5; // Defines a non-configurable property on the Window
    Object.defineProperty(window, "foo", { get: function() {}, configurable: true });

what should happen?

> What do you mean by "basically identical" and "force"?

"basically identical" means that the way https://tc39.github.io/ecma262/#sec-ordinarysetwithowndescriptor ends up acting in the `IsDataDescriptor(ownDesc)` case when the _ownDesc_ is that it calls `Receiver.[[DefineOwnProperty]]` with a descriptor created as `{ [[Value]]: V }`.  And in strict mode that ends up with the same behavior as  the `defineProperty` call with that descriptor, afaict.

> and "force"

If the [[DefineOwnProperty]] internal method on _windowProxy_ modifies the value of `configurable` in the  descriptor to `true` before forwarding to the [[DefineOwnProperty]] internal method of the _hiddenWindow_, while the latter has a non-configurable property on it.   This comes back to what exactly you mean by "refuses to be configured" and how you propose _windowProxy_'s [[DefineOwnProperty]] be implemented.  Again, all the property storage is on the  _hiddenWindow_.

allenwb:
> > ```js
> > Object.defineProperty(window, "foo", { value: 5, configurable: false });
> > Object.defineProperty(window, "foo", { value: 6 });
> > ```
...
> * The first call must then fail by returning false, since a property with the attributes it asked for was not created. It fails by returning false, rather than by throwing, because in the absence of this special case it would have succeeded and this is the least disruptive difference that does not break legacy code. Since the property refuses configuration, it does little harm to most naive code for it to assume (by not checking the return value) that it got what it asked for. Under all behaviors _other_ than reflective examination, it did get what it asked for.

This can't be correct because `Object.defineProperty` is specified to return its first parameter if it does not fail.  eg:

```js
Object.defineProperty(obj, "foo", {value: `})
           .defineProperty(obj, "bar", {value: 2});
```
Returning false would break that use case.  It's [[DefineOwnProperty]] that returns true/false to indicate success or failure.

bzbarsky:
> This can't be correct because Object.defineProperty is specified to return its first parameter if it does not fail. 

That is one of the things we are proposing changing: that in this specific case it will return false.

> Returning false would break that use case.

Yes, it would, for the specific case of `obj` being a WindowProxy and a non-configurable property being defined.  We haven't seen instances of this pattern being used with WindowProxy and non-configurable properties so far.

raulsebastianmihaila:
> This can't be correct because `Object.defineProperty` is specified to return its first parameter if it does not fail. eg:
> 
> ```js
> Object.defineProperty(obj, "foo", {value: `})
>            .defineProperty(obj, "bar", {value: 2});
> ```
> 
> Returning false would break that use case. It's [[DefineOwnProperty]] that returns true/false to indicate success or failure.

But that doesn't work by default since `defineProperty` doesn't exist on `Object.prototype`.

erights:
> oops, right. Of course other forms of chain would have the problem:
> 
> ```js
> anyFunctionThatExpectsTheGlobalObj(Object.defineProperty(window,"foo", {value:1}));
> ```

This still works identically in the success case. In the failure case, the only surprise is that `anyFunctionThatExpectsTheGlobalObj` is still called with false, rather than the `defineProperty` throwing, thus preventing `anyFunctionThatExpectsTheGlobalObj` from being called.

bzbarsky:
@erights That sketch makes sense to me, fwiw.  If desired I can create a Firefox build with that implemented for people to experiment with...

erights:
Desired, yes, thanks!!!

allenwb:
@erights 
Also remember that the properties created for global `var` and `function` declarations (except for those created using `eval`) are non-configurable.  This requirement dates to the ES3 specification (where they were instantiated as dontDelete). In ES1&ES2 global declarations were instantiated with an empty attribute set (meaning writable, deletable, and enumerable).   Makes me wonder if this ES3 era requirement was  never implemented or if dontDelete was simply ignored when the hiddenWindow was replaced but otherwise respected (more likely). 

Also recall, ES6+ requires that a realm keep track of which global object properties were created via `var` and `function` so that state might be used in defining exotic behaviors on the windowProxy.  EG, disallowing delete even though configurable is true.

Also, after checking the HTML spec. I wish it was clearer what exactly happens when/after,  a programatic invoked navigation cases a change of global  execution context (ie current realm, hiddenWindow, the global lexical environment, etc). In particular, it isn't obvious where execution resumes,  what happens to the active job that triggered the navigation, what dependencies there might be on the actual identify of the windowProxy object, and how existing code might use the fact that such navigation preserves the identify of the windowProxy.

bzbarsky:
> or if dontDelete was simply ignored when the hiddenWindow was replaced 

Indeed.  That's the issue we are trying to solve, sort of.

> In particular, it isn't obvious where execution resumes

Resumes from what?

> what happens to the active job that triggered the navigation

Navigation is always async in the cases when the global changes.  You can run code in a global that's been navigated away from (by grabbing a function from it before the navigation and then calling it after the navigation), but you can't be in the middle of  running code on it at the moment  when the global is switched out.  At least per spec.

> what dependencies there might be on the actual identify of the windowProxy object

I'm not  sure what you mean.

> how existing code might use the fact that such navigation preserves the identify of the windowProxy

Well, most obviously by grabbing it once, then continuing to set .location.href on it and waiting for loads, then accessing the resulting things via the windowProxy.  Now this _could_ work with just having multiple windowProxy objects that all end up pointing to the same _hiddenWindow_, of course.  If the question is what depends on that not happening, I don't have a clear example offhand  (though I expect things do depend on it).  But even if we had multiple windowProxy objects, the  fact that which _hiddenWindow_ they point to needs to change is what we're dealing with here.

bzbarsky:
@erights links to builds that should work for a few weeks:

* Linux64: https://queue.taskcluster.net/v1/task/ABvT1aYzTvOy7nRj8xWUXA/runs/0/artifacts/public/build/target.tar.bz2
* Linux32: https://queue.taskcluster.net/v1/task/EOS4qgy4Q7qaZ0Qd_DN-Cg/runs/0/artifacts/public/build/target.tar.bz2
* Mac: https://queue.taskcluster.net/v1/task/bNxTmEL2SAe_utjufBFS0w/runs/0/artifacts/public/build/target.dmg
* Win64: https://queue.taskcluster.net/v1/task/JgdPBAPHTM2_I6sRivvcXw/runs/0/artifacts/public/build/target.zip
* Win32: https://queue.taskcluster.net/v1/task/I4wqxze4Q-SwCOTCaD7xrw/runs/0/artifacts/public/build/target.zip

erights:
@bzbarsky that is so awesome, thanks!!!

Confirming that it seems to do what I expected, and that both old and new SES work fine with it, without needing any mods.

attn @jfparadis @warner @dtribble 

I am on a Mac and so am only testing the Mac version. Is there anything about this that is possibly platform dependent such that I should also try some of the others?

allenwb:
> > what happens to the active job that triggered the navigation
> 

What I had in mind was:
```js
self.myGlobal = "old context"
history.back();  //assumes changes the Realm and execution context
console.log("back from back(). myGlobal: "+ myGlobal); //if this is execute what is the valueOf myGlobal
```
Does the history.back() call return; ie, does the current job run to completion before the change to Window?

> Navigation is always async in the cases when the global changes. You can run code in a global that's been navigated away from (by grabbing a function from it before the navigation and then calling it after the navigation), but you can't be in the middle of running code on it at the moment when the global is switched out. At least per spec.
> 

So, run to completion still happens and the navigation happens after the current job completes?

> ...by grabbing a function from it before the navigation...

Where can you stash away such a function?  If all globals (including global lexical bindings, etc) are completely replaced when navigation happens where do you put the reference to the function before the navigation such that it can be retrieved after the navigation?  Presumably an iframe window context could stash it in a parent but it isn't clear to me that there is any place to stash it for a outer-most window.

> > what dependencies there might be on the actual identify of the windowProxy object
> 
> I'm not sure what you mean.

I meant, are there known use cases where application code depends upon the fact that the identify of the apparent "global object" is preserved when navigation occurs?

> 
> > how existing code might use the fact that such navigation preserves the identify of the windowProxy

>Well, most obviously by grabbing it once, then continuing to set .location.href on it and waiting for loads, then accessing the resulting things via the windowProxy. 

So setting .location.href triggers a global context change, like a history navigation? By "waiting" do you mean setting up an event handler, before the navigation, that will be trigger after the context change?  If so,  I guess that implies some uber global state that survives navigations??

>Now this could work with just having multiple windowProxy objects that all end up pointing to the same hiddenWindow, of course. If the question is what depends on that not happening, I don't have a clear example offhand (though I expect things do depend on it). But even if we had multiple windowProxy objects, the fact that which hiddenWindow they point to needs to change is what we're dealing with here.

What I was really wondering was if it is unusual/hard to have cross-navigation dependencies upon the === identify of the windowProxy then perhaps it might be possible to merge the WindowProxy and hiddenWindow into a single object, hence eliminating the configurability invariant violation.

I realize this is deep browser architecture stuff that nobody is likely to want to mess with, but so are the ES essential object invariants.  Just probing for deep assumptions on either side of the integration that might prove to be unnecessary design constraints.

bzbarsky:
> Is there anything about this that is possibly platform dependent

There shouldn't be.  I just wasn't sure what OS you were on, and posting all the links was simpler than doing another round-trip. ;)

> Does the history.back() call return; ie, does the current job run to completion before the change to Window?

Yep.  `history.back()` is  async in all the cases when  it changes the Window.   It can synchronously undo anchor scrolls, but that doesn't change the Window.

> So, run to completion still happens and the navigation happens after the current job completes?

Yes.

> Where can you stash away such a function?

In a different window.   As you note, you could grab a function from an iframe, then navigate that iframe.  Or you could `window.open` a new window, grab a function from it, and navigate it.  Or  you could `window.open` a new window, it could grab a function from you (via you handing it one, or it reaching into its `window.opener`) and navigate you.

> are there known use cases where application code depends upon the fact that the identify of the apparent "global object" is preserved when navigation occurs?

Yes.  I can't link to one offhand, but I've certainly seen code on the web depending on it before.

> So setting .location.href triggers a global context change, like a history navigation?

Setting `location.href` does a navigation.  Again, this is async in all the cases that change the `Window`.

> By "waiting" do you mean setting up an event handler, before the navigation, that will be trigger after the context change?

Yes, or just waiting via a setTimeout (racy, but doable).

> If so, I guess that implies some uber global state that survives navigations??

Again, this presupposes that multiple windows are around, due to iframes or `window.open`.

> then perhaps it might be possible to merge the WindowProxy and hiddenWindow into a single object

Ah.  Grabbing `.contentWindow` from an iframe once and then using that to refer to whatever is currently loaded in the iframe as navigations happen is definitely something people do on the web.  So we can't do the merging, unfortunately.

bzbarsky:
> Grabbing .contentWindow from an iframe once and then using that to refer to whatever is currently loaded in the iframe

Or for that matter calling `window.open` and then using the return value to refer to whatever is loaded in that newly opened window.

claudepache:
Is there any value for `Object.defineProperty()` to return false in the fail-but-not-throw exceptional cases, instead of just happily return the target object?

Those interested in the truth could just use `Reflect.defineProperty()`.

As of `Object.defineProperty()`, I think that the least disruptive outcome is:
* keep the current behaviour of always either throwing or returning the target object;
* limit the exceptional cases where the operation (partially) fails but does not throw. More precisely, that should only happen when all attribute changes did succeed, except for "writable" and/or "configurable". Concretely, one could decide that this will happen precisely when the operation on _hiddenWindow_ did succeed but the operation on the WindowProxy is reputed to have failed.

bzbarsky:
OK, if people are pretty happy with the proposal in https://github.com/tc39/ecma262/pull/688#issuecomment-430648819 I'm going to try it in Firefox nightlies and see how it goes.

erights:
Is it identical to the Firefox downloads you previously provided? IIRC, they perfectly meet all my expectations. Thanks!

bzbarsky:
I believe it's identical to the builds I posted, yes.

bzbarsky:
OK, so I tried that, and in particular ran our full test suite for the first time with the getOwnPropertyDescriptor changes.   Unfortunately, I get test failures from some libraries that  define shims (for logging, I think) something like this:

1. Get the current property descriptor (for a property that was initially defined via `function foo() {}` in global scope, so is non-configurable but writable).
2. Change the value in the descriptor.
3. Try to redefine the property using that descriptor.

With the above proposal, the descriptor gotten in step 1 says `configurable: true` explicitly, so step 3 throws because the underlying property is non-configurable....

The code involved is https://github.com/sinonjs/sinon/blob/d193bb932bd03b587a661bbb74cc6772f2e6941f/lib/sinon/util/core/wrap-method.js#L68-L94 which I think is in a fairly popular library...

erights:
I suspect that there's a prettier way to write the equivalent to the following code, but anyway, here's a proposed fix.

```js
    defineProperty(_, name, desc) {
      const result = Reflect.defineProperty(hiddenWindow, name, desc);
      if (!result) {
        const hiddenDesc = Reflect.getOwnPropertyDescriptor(hiddenWindow, name);
        if (!hiddenDesc.configurable && desc.configurable) {
          result = Reflect.defineProperty(hiddenWindow, name, {...desc, configurable: false});
        }
        if (!result) {
          return false;
        }
      }
      if (desc.configurable === false && !(name in shadowTarget)) { return false; }
      return result;
    },
```

In the case of interest, this will succeed at changing the value and report that `defineProperty` succeeded. This is correct by the invariants. The call asked for a configurable property on the windowProxy. The property in question will still be reported as configurable on the windowProxy.

bzbarsky:
Yes, that seems to do the trick, good idea!

erights:
Excellent, thanks!!!

bzbarsky:
Yes, I had made that same edit locally; should have commented here...

bzbarsky:
OK, so we have tried this on nightly, and we have a web compat problem report: https://bugzilla.mozilla.org/show_bug.cgi?id=1521677

The summary is that the page is using `Object.defineProperties` to try to define explicitly non-configurable properties on window.

I suppose we could change this to return `false` like `Object.defineProperty`...  @erights, thoughts?

bzbarsky:
The other option is to have `Object.defineProperties` act as it always does: define the property, return the object.  That is, it would not communicate back the fact that we didn't "really" define the property non-configurable, but neither would it throw an exception, which is what it does right now with the proposed spec changes.

erights:
Is another option to have basecamp.com fix their site so that it works both before and after this fix?

Some previous web breakages were resolved this way. (Global `JSON` breaking facebook.com) If it can be resolved this way, that would be best.

bzbarsky:
> Is another option to have basecamp.com fix their site so that it works both before and after this fix?

I don't know.  I don't even know where they're "really" getting this code; debugging minified code is a huge pain.

More worrying to me is whether they're the only site doing this.  And given that it works in all browsers right now, whether people are likely to keep adding code like that.  And if they do, whether they will test in any non-Chrome browsers.

Or put another way, to ship any sort of behavior here that has potential for breaking sites in Firefox, I would want Chrome to be shipping it at least as early as us.

bzbarsky:
Per discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1521677 the current plan in Gecko is:

1. Switch `defineProperty` to returning `null`, not `false`, in the failure case, per https://github.com/tc39/ecma262/pull/688#issuecomment-427108126

2. Have `defineProperties` return `null` (instead of throwing) if any of the definitions failed for reasons relevant to this issue.  Such a failing definition does not prevent later ones in the `defineProperties` call from being processed.

We will keep that on in nightly for a bit and see whether there are any more compat issues.

That said, if people are happy about that setup, I would appreciate someone with the TC39 bandwidth to do so dealing with getting this into the spec and test262...

anba:
Will private fields need a similar treatment? 

@littledan

anba:
@bzbarsky @annevk Was there any discussion about the private fields issue mentioned in https://github.com/tc39/ecma262/pull/688#issuecomment-555618645, because I feel they're quite similar to non-configurable properties. I'm at the All-Hands in case you want to discuss this P2P.

annevk:
This affects WindowProxy and Location objects, at most. How would you add private fields to these objects in userland? I'm hoping that's not possible. (Happy to chat btw, seems our dinners will take place in the same restaurant in a bit.)

anba:
Class fields (public and private) in derived classes are added to the instance after `super()` returns. And classes are additionally allowed to return an arbitrary object from their `constructor` function. If both features are combined, you can write code like:
```js
class Base {
  constructor(obj) {
    return obj;
  }
}

class Derived extends Base {
  #my_field = 0;

  constructor(obj) {
    super(obj);
  }
}
```

And then call `new Derived(window)` to add the private field `#my_field` to the window object.

(Sure, we can chat at the dinner.)

erights:
> https://github.com/tc39/ecma262/pull/688#issuecomment-579860207

is correct. However, using the weakmap-model of private fields, this behavior should be neither surprising nor problematic. I don't see any analogy to the non-confugurable property issue. Can someone explain what the similarity is, or what problem we're worried about?

annevk:
As I understand it you can add fields to a `WindowProxy` that do not get cleared upon navigation (when `WindowProxy`'s underlying global object gets replaced).

ljharb:
@annevk sure, but if you retain the ability to read those fields, then you could also retain a `WeakMap` that had the WindowProxy as a key, no?

bzbarsky:
Yes, but the difference is in who can read that state.  Concretely, let's say you have a page at origin A that loads a subframe from origin B.  The script in the subframe adds some fields to the WindowProxy.  The parent page can't read them, presumably, right?  Now the parent page navigates the subframe to a different page, from origin A.  Can it not read those fields?  Why or why not?

This is not an issue with the WeakMap case, because the two different origins can never both have references to the same WeakMap.  But they _do_ have references to the same WindowProxy, with dynamic security checks on property access...

One thing that is not obvious to me is whether the fields would (should?) in fact get added to the `WindowProxy` or the underlying `Window`...

ljharb:
Only the lexical scope where the private fields are defined, though, would be able to read them, so I'm not clear on how it differs from the WeakMap case.

anba:
That's the current situation, but in the future private names may be reifiable (for example like it was proposed in the decorators proposal <https://tc39.es/proposal-decorators/#sec-private-name-objects>).

annevk:
Another concern is that it's very easy to create memory leaks this way that would be complicated to address as I understand it.

bzbarsky:
@annevk and I talked through the leak concerns, and as long as private field names cannot be reified, so we can clear private fields on a `WindowProxy` when the `Window` for their lexical scope is collected, leaks don't seem to be an issue.

If at some point private field names can escape to script and hence accesses to the private fields can outlive the relevant `Window`, then we will have a problem.  So if there are thoughts about doing that in the future, we need to figure out something about private fields on `WindowProxy` _now_ so we don't have to change their behavior later.

ljharb:
If they could be reified, there’d need to be a way for Proxy to intercept such private field symbols, so I’d assume WindowProxy could use the same mechanism. This seems like useful feedback for the https://github.com/tc39/proposal-private-declarations proposal.

erights:
I don't understand this. Reifying the private name should be equivalent to reifying a weakmap holding the windowProxy as a key. 

Perhaps this does illuminate the disconnect. It cannot be possible to associate private fields with a window, since the window is never reified. The return override technique is returning the windowProxy, not the window, and thus that's what the private fields are associated with. Navigating the frame cannot have any affect on that.

annevk:
Firefox [abandoned the plan](https://bugzilla.mozilla.org/show_bug.cgi?id=1496510) in https://github.com/tc39/ecma262/pull/688#issuecomment-457708533 as even on just Nightly it continued to cause problems for us. First due to an added test and later due to Google Analytics.

I suggest we close this PR and continue the discussion in #672 with regards to what changes to make.

PR #689, opened 2016-09-12
Normative: Remove arguments.caller poison pill

### LINKED ISSUE #670: Remove arguments.caller from the spec
_**Note.** Not to be confused with `Function#caller`._

---

The spec acknowledges that `arguments.caller` “has historically been provided as an implementation-defined extension by some ECMAScript implementations”.

However, nowadays this historical wart is not found in any browser since long ago, e.g.:
* [removed from Mozilla before Firefox was born] (https://bugzilla.mozilla.org/show_bug.cgi?id=7224) ;
* [removed from IE9 Standards mode] (https://msdn.microsoft.com/en-US/library/gg622933(v=vs.85).aspx).

I think one can clean the spec by removing all special cases around `arguments.caller`. As far as I can see, the modifications would be:

* [9.4.4. Arguments Exotic Objects] (https://tc39.github.io/ecma262/#sec-arguments-exotic-objects), Note 3: remove the mention of "caller".

* [9.4.4.1 [[GetOwnProperty]] of Arguments objects] (https://tc39.github.io/ecma262/#sec-arguments-exotic-objects-getownproperty-p): remove step 7, that special-cases `"caller"`, together with the remark _”If an implementation does not provide a built-in caller property for argument exotic objects then step 7 of this algorithm must be skipped.”_

* [9.4.4.5 [[HasProperty]] of Arguments objects] (https://tc39.github.io/ecma262/#sec-arguments-exotic-objects-hasproperty), use the same implementation as ordinary objects, removing the `"caller"` special case and the remark ”_If an implementation does not provide a built-in caller property for argument exotic objects then step 2 of this algorithm must be skipped._”

* [9.4.4.7. CreateUnmappedArgumentsObject] (https://tc39.github.io/ecma262/#sec-createunmappedargumentsobject), remove step 9, that adds a "caller" poison-pill to unmapped arguments objects.

* [16.2 Forbidden Extensions] (https://tc39.github.io/ecma262/#sec-forbidden-extensions), add: _“Arguments objects, both mapped and unmapped, must not be created with an own property named `"caller"`.”_

* [Annex C The Strict Mode of ECMAScript] (https://tc39.github.io/ecma262/#sec-strict-mode-of-ecmascript), modify item: _“Arguments objects for strict functions define <del>non-configurable accessor properties named "caller" and "callee"</del> <ins>a non-configurable accessor property named "callee"</ins> which throw<ins>s</ins> a TypeError exception on access”_ — (Also, at the time of writing, backreference to section 9.2.7 is incorrect, it should be 9.4.4.7; was already reported in #572.)

Concretely, for the implementations not supporting the arguments.caller extension, the changes resume to removing the "caller" poison-pill from their unmapped Arguments objects.

bterlson:
Implementations have not had this extension for many years so continuing
to poison pill it is needless. This removes all mention of arguments
object's "caller" property with the exception of a note that gives a bit
of history (mostly so people can still ctrl+f "caller").

Closes #670.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7547,7 +7547,10 @@
         <p>The ParameterMap object and its property values are used as a device for specifying the arguments object correspondence to argument bindings. The ParameterMap object and the objects that are the values of its properties are not directly observable from ECMAScript code. An ECMAScript implementation does not need to actually create or use such objects to implement the specified semantics.</p>
       </emu-note>
       <emu-note>
-        <p>Arguments objects for strict functions define non-configurable accessor properties named `"caller"` and `"callee"` which throw a *TypeError* exception on access. The `"callee"` property has a more specific meaning for non-strict functions and a `"caller"` property has historically been provided as an implementation-defined extension by some ECMAScript implementations. The strict mode definition of these properties exists to ensure that neither of them is defined in any other manner by conforming ECMAScript implementations.</p>
+        <p>Arguments objects for strict functions define a non-configurable accessor property named `"callee"` which throws a *TypeError* exception on access. The `"callee"` property has a more specific meaning for non-strict functions. The strict mode definition of this property exists to ensure that it is not defined in any other manner by conforming ECMAScript implementations.</p>
+      </emu-note>
+      <emu-note>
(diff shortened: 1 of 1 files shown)


littledan:
Well-done spec change--great simplification of the arguments exotic object, happy to have the historical note, and appropriate prohibited extension. Good stuff all around.

bakkot:
I [PR'd](https://github.com/tc39/test262/pull/797) a test262 update for this, but I thought we were going to stop merging normative changes until accompanying tests had gone in?

PR #697, opened 2016-09-27
Normative: make Function.prototype.toString forward-compatible and fully defined

### LINKED ISSUE #664: `Function.prototype.toString` revision
https://github.com/tc39/Function-prototype-toString-revision by @michaelficarra

Please tag this issue with the “web reality” label. Thanks!

michaelficarra:
Implements the currently stage 3 proposal at https://github.com/tc39/Function-prototype-toString-revision. This PR is part of the new stage 4 entrance criteria.

Fixes #664.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -709,7 +709,7 @@
 
     <emu-clause id=sec-algorithm-conventions-syntax-directed-operations namespace=algorithm-conventions>
       <h1>Syntax-Directed Operations</h1>
-      <p>A <dfn>syntax-directed operation</dfn> is a named operation whose definition consists of algorithms, each of which is associated with one or more productions from one of the ECMAScript grammars. A production that has multiple alternative definitions will typically have a distinct algorithm for each alternative. When an algorithm is associated with a grammar production, it may reference the terminal and nonterminal symbols of the production alternative as if they were parameters of the algorithm. When used in this manner, nonterminal symbols refer to the actual alternative definition that is matched when parsing the source text.</p>
+      <p>A <dfn>syntax-directed operation</dfn> is a named operation whose definition consists of algorithms, each of which is associated with one or more productions from one of the ECMAScript grammars. A production that has multiple alternative definitions will typically have a distinct algorithm for each alternative. When an algorithm is associated with a grammar production, it may reference the terminal and nonterminal symbols of the production alternative as if they were parameters of the algorithm. When used in this manner, nonterminal symbols refer to the actual alternative definition that is matched when parsing the source text. The <dfn>source text matched by</dfn> a grammar production is the portion of the source text that starts at the beginning of the first terminal that participated in the match and ends at the end of the last terminal that participated in the match.</p>
       <p>When an algorithm is associated with a production alternative, the alternative is typically shown without any &ldquo;[ ]&rdquo; grammar annotations. Such annotations should only affect the syntactic recognition of the alternative and have no effect on the associated semantics for the alternative.</p>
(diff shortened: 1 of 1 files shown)


littledan:
In the January 2018 TC39 meeting, the committee [decided](https://github.com/tc39/tc39-notes/blob/master/es8/2018-01/jan-23.md#13iiic-functionprototypetostring-pr-for-stage-4) to wait to advance Function.prototype.toString to Stage 4 (and land this patch) until we add a corresponding mechanism for allowing implementations to not return all of the source code. One such mechanism would be @domenic 's [Function.prototype.toString() censorship proposal](https://github.com/domenic/proposal-function-prototype-tostring-censorship).

michaelficarra:
@littledan see https://github.com/tc39/Function-prototype-toString-revision/pull/26 for the PR that addressed that concern and for a PR addressing @anba's comments https://github.com/tc39/Function-prototype-toString-revision/pull/30.

annevk:
Does HostHasSourceTextAvailable require that a host basically has a set of all its built-in functions? How does that work if the set of functions differs across agent/realm-boundaries?

littledan:
@annevk I thought that hook wouldn't be called on native functions, as their `[[SourceText]]` internal slot would be `undefined`.

annevk:
I see, this would only be for "userland" functions... I'm a little curious what https://github.com/tc39/Function-prototype-toString-revision/issues/25#issuecomment-360231682 is about as without that it's somewhat tough to evaluate whether this works.

michaelficarra:
@annevk I'll bring it up during the stage 4 presentation.

domenic:
To expand:

I think the hook provided here is fine from the perspective of the F.p.toString proposal.

If we were to try to specify a `No-Fn-ToString-Source: true` header or similar, what we'd need is the information tying each individual function to the source file in which it's found, so that when the host gets a call to HostHasSourceTextAvailable(func), it is able to tell whether func is associated with a source file for which that header is set.

Creating that association is a separate effort. I'd be tempted to go with something hand-wavey, such as prose stating "If func was originally created via source text that was present in foo.js" or whatever. You could imagine making something rigorous where we associate each function with a Script Record or Module Record and then we are able to go function -> script/module record -> source file name -> is header present. But that's a separate effort, and doesn't impact the contents or shape of this PR.

mathiasbynens:
This proposal reached stage 4 at the November 2018 meeting. Please prioritize it over other PRs. Thanks!

ljharb:
We’ll look at it first thing in next week’s editor meeting.

ljharb:
@michaelficarra i've rebased this PR on latest master; can you confirm that my conflict resolution worked out?

spec.html
```diff
@@ -24754,30 +24809,19 @@ <h1>Function.prototype.constructor</h1>
 
       <emu-clause id="sec-function.prototype.tostring">
         <h1>Function.prototype.toString ( )</h1>
-        <p>When the `toString` method is called, the following steps are taken:</p>
+        <p>When the `toString` method is called with an object _func_ as its *this* value, the following steps are taken:</p>
```
michaelficarra: This is essentially a revert of a change made in the meantime. We can probably put this line back and restore the first step of the current algorithm ("Let _func_ be the *this* value.") in front of the steps in this PR.

michaelficarra:
LGTM otherwise.

spec.html
ljharb: Thanks, will fix now.

PR #701, opened 2016-09-28
Add missing BindingInitialization production

sebmarkbage:
This only covered the empty pattern case. This is just a pass through to BindingPropertyList but in future proposals this will be an important refactoring point.

For clarity: The equivalent ObjectAssignmentPattern performs `RequireObjectCoercible(value)` but we don't do that here because it is already covered by the `BindingPattern` production.

cc @bterlson

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -14985,6 +14985,14 @@
         <emu-alg>
           1. Return NormalCompletion(~empty~).
         </emu-alg>
+        <emu-grammar>
+          ObjectBindingPattern :
+            `{` BindingPropertyList `}`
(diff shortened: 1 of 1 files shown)


michaelficarra:
This should already be defined by the [chain production rule](https://tc39.github.io/ecma262/#sec-context-free-grammars), right?

bterlson:
I believe this is already covered with the chain rule (see also: https://tc39.github.io/ecma262/#sec-algorithm-conventions, about half-way through the clause). We can add this production when it becomes required by subsequent proposals.

PR #704, opened 2016-10-02
Normative: fix `Promise#then` so `PerformPromiseThen` returns a promise

ljharb:
I noticed while working on the `Promise.prototype.finally` tests that `PerformPromiseThen` returns `resultCapability.[[Promise]]`, but that `Promise.prototype.then` never actually sets that slot.

I labeled this "normative" but if implementors weren't doing this already, `Promise#then` would be returning `undefined`.


domenic:
This doesn't seem correct... NewPromiseCapability automatically creates a promise with its [[Promise]] slot filled to the newly-created promise. What this PR does is overwrite that [[Promise]] slot to be the same as the `this` value. That would mean `promise.then(...)` always returns `promise`, which is not right.

ljharb:
whoops, you're totally right. sorry, late night spec reading :-)

PR #706, opened 2016-10-03
Apply chain rule to IsFunctionDefinition UnaryExpression: UpdateExpression

fkling:
Since a _FunctionExpression_ is a _PrimaryExpression_, I believe returning `false` in `IsFunctionDefinition` for every production of _UnaryExpression_ is incorrect. It's also not consistent with the rules for every other expression.

Otherwise I think `IsFunctionDefinition` would return `false` when evaluating initializers (e.g. `function foo(bar=function(){}) {}`), after "unrolling" all production rules.

Or am I misinterpreting the rules?

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12253,7 +12253,6 @@
       <emu-see-also-para op="IsFunctionDefinition"></emu-see-also-para>
       <emu-grammar>
         UnaryExpression :
-          UpdateExpression
           `delete` UnaryExpression
           `void` UnaryExpression
           `typeof` UnaryExpression


not-an-aardvark:
I'm not sure about the *FunctionExpression* issue, but I'm pretty sure this isn't the correct way to solve it -- that `UnaryExpression` definition is necessary for a lot of other things.

e.g. for parsing this program:

```js
5 * 4
```

The following is the bottom part of the parse tree for the leftmost token:

*MultiplicativeExpression*
*ExponentiationExpression*
***UnaryExpression***
***UpdateExpression***
*LeftHandSideExpression*
*NewExpression*
*MemberExpression*
*PrimaryExpression*
*Literal*
*NumericLiteral*
*DecimalLiteral*
*DecimalIntegerLiteral*
*NonZeroDigit*
`5`

fkling:
@not-an-aardvark: The only thing this PR changes is that `IsFunctionExpression(<UnaryExpression>)` is delegated to `IsFunctionExpression(<UpdateExpression>)` instead of returning `false`. It wouldn't have any impact on the parse tree.

allenwb:
Yes, this fix is correct and necessary.

not-an-aardvark:
Oops, I was mistaken -- I thought this fix was changing the definition for all `UnaryExpressions`. Never mind my objection, then.

fkling:
Do I have to do anything specific to get this reviewed / merged?

bterlson:
Wait for me :-P Sorry I've been AFG (away from GitHub) for a bit.

fkling:
No worries :) Thank you!

PR #720, opened 2016-10-26
Editorial: Remove characterization of JSON as a subset of ECMAScript

gibson042:
https://esdiscuss.org/topic/feature-request-make-ecma262-a-superset-of-json doesn't seem to be getting much traction, and in the meantime we should update this specification for factual correctness.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33368,7 +33368,7 @@ THH:mm:ss.sss
     <!-- es6num="24.3.1" -->
     <emu-clause id="sec-json.parse">
       <h1>JSON.parse ( _text_ [ , _reviver_ ] )</h1>
-      <p>The `parse` function parses a JSON text (a JSON-formatted String) and produces an ECMAScript value. The JSON format is a subset of the syntax for ECMAScript literals, Array Initializers and Object Initializers. After parsing, JSON objects are realized as ECMAScript objects. JSON arrays are realized as ECMAScript Array instances. JSON strings, numbers, booleans, and null are realized as ECMAScript Strings, Numbers, Booleans, and *null*.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -33368,15 +33368,15 @@
...
         1. Parse _JText_ interpreted as UTF-16 encoded Unicode points (<emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>) as a JSON text as specified in ECMA-404. Throw a *SyntaxError* exception if _JText_ is not a valid JSON text as defined in that specification.
         1. Let _scriptText_ be the result of concatenating `"("`, _JText_, and `");"`.
         1. Let _completion_ be the result of parsing and evaluating _scriptText_ as if it was the source text of an ECMAScript |Script|, but using the alternative definition of |DoubleStringCharacter| provided below. The extended PropertyDefinitionEvaluation semantics defined in <emu-xref href="#sec-__proto__-property-names-in-object-initializers"></emu-xref> must not be used during the evaluation.
         1. Let _unfiltered_ be _completion_.[[Value]].
-        1. Assert: _unfiltered_ will be either a primitive value or an object that is defined by either an |ArrayLiteral| or an |ObjectLiteral|.
+        1. Assert: _unfiltered_ will be either a String, Number, Boolean, or Null primitive value or an Object that is defined by either an |ArrayLiteral| or an |ObjectLiteral|.
```
ljharb: i think `String, Number, Boolean, the null value, or an Object` might be more typical

ljharb: Sounds good, I wasn't used to seeing "Null" anywhere.

PR #724, opened 2016-10-28
Add ValidateTypedArray in each iteration of ArrayIterator

### LINKED ISSUE #713: Inconsistency in %ArrayIteratorPrototype%.next for TypedArrays
In [%ArrayIteratorPrototype%.next](https://tc39.github.io/ecma262/#sec-%arrayiteratorprototype%.next), step 8.a., if the Array being iterated over is a TypedArray, the length used is the TypedArray's [[ArrayLength]]. However, if the TypedArray's ArrayBuffer is detached, then the `length` accessor would've returned zero ([see spec](https://tc39.github.io/ecma262/#sec-get-%typedarray%.prototype.length)). For consistency, should we do the same here?

One fix would be to make step 8.a check for a detached ArrayBuffer and use 0 for the length in that case. I'd like this one, as it leaves in the special allowance to make TypedArrays a bit more implementation-friendly by using the underlying length. Another possibility would be to always use the `length` accessor.

Thanks to @caitp and @bmeurer for raising this issue to me and discussing possible fixes.

littledan:
closes #713

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -30522,6 +30522,7 @@ THH:mm:ss.sss
             1. Let _index_ be _O_.[[ArrayIteratorNextIndex]].
             1. Let _itemKind_ be _O_.[[ArrayIterationKind]].
             1. If _a_ has a [[TypedArrayName]] internal slot, then
+              1. If IsDetachedBuffer(_a_.[[ViewedArrayBuffer]]) is *true*, throw a *TypeError* exception.
               1. Let _len_ be _a_.[[ArrayLength]].
             1. Else,
               1. Let _len_ be ? ToLength(? Get(_a_, `"length"`)).


littledan:
This blocks on discussion in #713

bterlson:
Tests: https://github.com/tc39/test262/pull/784.

spec.html
```diff
@@ -30522,6 +30522,7 @@
             1. Let _index_ be _O_.[[ArrayIteratorNextIndex]].
             1. Let _itemKind_ be _O_.[[ArrayIterationKind]].
             1. If _a_ has a [[TypedArrayName]] internal slot, then
+              1. Perform ? ValidateTypedArray(_a_).
```
jswalden: Nitpicky, but it seems to me like

    1. If IsDetachedBuffer(_a_.[[ViewedArrayBuffer]]) is *true*, throw a *TypeError* exception.

is clearer and more readable about intent, and it doesn't reuse an overbroad spec method.

littledan: Good idea, changed to that wording in the new version.

caitp: It's unlikely that implementations would perform the redundant checks, but seems fair.

leobalter: that goes all the way back to the discussion on https://github.com/tc39/ecma262/issues/713#issuecomment-256087125 where this should be performing ValidateTypedArray or not.

littledan:
Is this ready to merge? It reached consensus at TC39 in the November 2016 meeting and there are tests.

PR #731, opened 2016-11-15
Clean up definition of Immutable Prototype Exotic Objects

### LINKED ISSUE #705: Immutable Prototype Exotic Objects is underspecified
The [specification](https://tc39.github.io/ecma262/#sec-immutable-prototype-exotic-objects) for Immutable Prototype Exotic Objects is incomplete and does not fully follow the pattern used to specify other standard exotic objects.  Other than for [[GetPrototypeOf]], an implementer would have to guess at the intended definition of the essential internal methods.

This can be corrected by replacing the single paragraph of 9.4.7 with something like the following :

> _Immutable prototype exotic objects_ have the same internal slots as ordinary objects however  the  [[Prototype]] internal slot is immutable. Once the [[Prototype]] internal slot has been initialized its value must not  change.
> 
> Immutable prototype exotic objects provide alternative definitions for the following internal methods. All of the other Immutable prototype exotic object essential internal methods that are not defined below are as specified in 9.1.

littledan:
- Explicitly state that they are just like ordinary objects, and have
  the same internal slots, except for one weird trick. (closes #705)
- Factor out the body of the [[SetPrototypeOf]] method of immutable
  prototype exotic objects into an abstract operation to make it
  easier to do integration with embedders.
  (see https://github.com/whatwg/html/issues/2035)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -8156,11 +8156,21 @@
 
     <emu-clause id="sec-immutable-prototype-exotic-objects">
       <h1>Immutable Prototype Exotic Objects</h1>
-      <p>An <dfn>immutable prototype exotic object</dfn> is an exotic object that has an immutable [[Prototype]] internal slot.</p>
+      <p>An <dfn>immutable prototype exotic object</dfn> is an exotic object that has an [[Prototype]] internal slot that will not change once it is initialized.</p>
+
+      <p>Immutable prototype exotic objects have the same internal slots as ordinary objects. They are exotic only in the following internal methods. All other internal methods of immutable prototype exotic objects that are not explicitly defined below are instead defined as in <a href="#sec-ordinary-object-internal-methods-and-internal-slots">ordinary objects.</a></p>
 
       <emu-clause id="sec-immutable-prototype-exotic-objects-setprototypeof-v">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -8156,12 +8156,22 @@
...
 
       <emu-clause id="sec-immutable-prototype-exotic-objects-setprototypeof-v">
         <h1>[[SetPrototypeOf]] (_V_)</h1>
         <p>When the [[SetPrototypeOf]] internal method of an immutable prototype exotic object _O_ is called with argument _V_, the following steps are taken:</p>
         <emu-alg>
+          1. Return ? SetImmutablePrototype(O, V).
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-set-immutable-prototype">
```
domenic: Maybe leave a source file comment `<!-- This is used by host environment specs, even though it only has one consumer in this spec; don't inline it -->`

littledan:
cc @allenwb

spec.html
domenic: Not at this time, I don't think, but it's been mentioned before as a good idea we might want to start doing.

littledan: Added the comment; is this patch good to submit now?

PR #733, opened 2016-11-16
Add definition of unfolding of AbstractOperation(ReturnIfAbrupt(arg))

edgemaster:
I had trouble understanding the steps required when a ReturnIfAbrupt is nested in a call to another operation such as `ToBoolean(? Call(...))`
I could not find appropriate guidance in the spec about how to apply the "means the same thing as" macro patterns, when the macro is applied mid-line as in this case.

The attached patch gives what I think is the intended meaning (after some initial discussion on IRC).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2713,6 +2713,16 @@
           1. Else if _hygienicTemp_ is a Completion Record, let _hygienicTemp_ be _hygienicTemp_.[[Value]].
         </emu-alg>
         <p>Where _hygienicTemp_ is ephemeral and visible only in the steps pertaining to ReturnIfAbrupt.</p>
+        <p>Algorithms steps that say or are otherwise equivalent to:</p>
+        <emu-alg>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -2713,6 +2713,16 @@
...
         </emu-alg>
         <p>Where _hygienicTemp_ is ephemeral and visible only in the steps pertaining to ReturnIfAbrupt.</p>
+        <p>Algorithms steps that say or are otherwise equivalent to:</p>
+        <emu-alg>
+          1. Let _result_ be AbstractOperation(ReturnIfAbrupt(_argument_)).
+        </emu-alg>
+        <p>mean the same thing as:</p>
+        <emu-alg>
+          1. If _argument_ is an abrupt completion, return _argument_.
+          1. Else if _argument_ is a Completion Record, let _argument_ be _argument_.[[Value]].
```
edgemaster: That was a copy-paste from the first definition in the section.

PR #737, opened 2016-11-19
Editorial: Clarify how Boolean objects are coerced to Boolean values

kdex:
This addition is consistent with #sec-string-object (4.3.19) and #sec-number-object (4.3.22) where the spec contains references to how the respective coercions are carried out.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -245,7 +245,7 @@
       <h1>Boolean object</h1>
       <p>member of the Object type that is an instance of the standard built-in `Boolean` constructor</p>
       <emu-note>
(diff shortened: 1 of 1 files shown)


bakkot:
`Boolean(new Boolean(false))` gives `true`, so this comment would be misleading.

In fact I'm pretty sure that comment refers to `.valueOf()`: `(new Boolean(false)).valueOf() === false`. So `+new Boolean(false)` is 0, `+new Boolean(true)` is 1, etc.

It also comes up in JSON stringification: `JSON.stringify(new Boolean(false)) === "false"`.

kdex:
Good catch! Thanks, I'll be closing this PR.

Issue #721, opened 2016-10-27
Redundancy in the `ToLength`?


zerkms:
http://www.ecma-international.org/ecma-262/7.0/#sec-tolength

The step 3 there is an explicit comparison with the `+Infinity`.

What is the point to have it there, as long as it will be handled by the last step `Return min(len, 2^53-1).` anyway.

chozzz:
+1

allenwb:
I think this is a case where the redundancy does no harm and is actually helpful to readers of the spec. 

Readers look at the ToX abstract operations when they need verify, for themselves, how specific cases are handled. In such situations it is easier to see an explicit statement of how `+Infinity` is handled than to have to remember or lookup that max() is defined to work with `+Infinity`.

leobalter:
This step was an useful reference when I wrote tests for TypedArrays, I imagined the same applies to runtime implementors.

zerkms:
@allenwb @leobalter then for consistency it should be another step that checks for `-Infinity`.

chozzz:
@zerkms They already did that on step 2.

`If len ≤ +0, return +0.`

zerkms:
@chozzz sure.

That's my exact point: if it is fine to expect `-Infinity` less than zero, then it would be equally fine to expect `+Infinite` greater than `2^53-1`.

bterlson:
Editorially I am fine with the current spec text in this regard.

bterlson:
There is now evidence of three people being confused by this. Since the min part is fairly straight forward I guess we'll remove the redundant check and rely on implementers to understand how max/min work with +/-Infinity.

PR #738, opened 2016-11-23
tolength: remove superfluous +Infinity check

thefourtheye:
The `+Infinity` check is superfluous as the `min` function takes care
of that as well.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3791,7 +3791,6 @@
       <emu-alg>
         1. Let _len_ be ? ToInteger(_argument_).
         1. If _len_ &le; *+0*, return *+0*.
-        1. If _len_ is *+&infin;*, return 2<sup>53</sup>-1.
         1. Return min(_len_, 2<sup>53</sup>-1).
       </emu-alg>
     </emu-clause>


annevk:
See #721 for last time...

ljharb:
This looks like the same question as #721.

bterlson:
With another group of people confused by this, I'll remove the redundant check. I will entertain arguments for a note, though I don't really want to go through the entire spec adding notes about +/- infinity for every usage of min/max.

bterlson:
@thefourtheye congrats on your first contribution to ECMA262!!

thefourtheye:
Thanks @bterlson :-) To be clear, you don't want to change other places like this in the spec?

bterlson:
@thefourtheye I did a scan for other similar patterns w/ min/max and didn't see anything glaring. Which other places are you referring to?

thefourtheye:
@bterlson Oh I didn't have anything else in mind. I just wanted to understand your sentence clearly.

bterlson:
@thefourtheye Oh, my previous comment was saying that I could address the motivation to have the explicit +Infinity case by adding a note, but I can't convince myself that this usage of max/min is special enough to deserve the note and I don't want to add the note for every usage. Without a justification for ToLength being special I am ok not putting a note in anywhere.

thefourtheye:
Oh okay, got it :-) Thanks @bterlson

zerkms:
It should have been mine PR 😭

bterlson:
@zerkms I'm so sorry :( But, good news, there are boat loads of editorial issues you could fix! Just read a few clauses and I'm sure you'll find something.

zerkms:
@bterlson I'm just kidding :-) Not a real problem at all :-)

PR #740, opened 2016-11-25
Fix section number comments thrown off by ExponentiationExpression addition; pluralize Keyed Collections to match Indexed Collections.

BrendanEich:
Fix section number comments thrown off by ExponentiationExpression addition; pluralize Keyed Collections to match Indexed Collections.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12605,7 +12605,6 @@
     </emu-clause>
   </emu-clause>
 
-  <!-- es6num="12.6" -->
   <emu-clause id="sec-exp-operator">
     <h1>Exponentiation Operator</h1>
     <h2>Syntax</h2>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -32328,8 +32332,8 @@
 </emu-clause>
 
 <!-- es6num="23" -->
-<emu-clause id="sec-keyed-collection">
-  <h1>Keyed Collection</h1>
+<emu-clause id="sec-keyed-collections">
```
annevk: Should prolly not change the ID as that breaks links.

spec.html
```diff
@@ -12605,6 +12605,7 @@
     </emu-clause>
   </emu-clause>
 
+  <!-- es6num="12.6" -->
```
annevk: es6num is just a reference to the legacy section number in ES6. They should not be added or modified.

BrendanEich:
Never mind, I'll just fix the h1 text to use the plural. Thanks again @annevk -- you got me on the right path for material changes to come.

bterlson:
I'm fine leaving behind an anchor with the old id, but in this case seems fine to leave it as is.

PR #745, opened 2016-12-06
Add link to "Finished" proposals

timoxley:
I tend to find myself hitting this readme first when looking for proposal status updates, I think it's because "ecma262" is a clean, specific filter term vs "proposals".

It's currently a minor nuisance to have to click through to the "active proposals" page first, then scroll to the bottom and click "Finished Proposals" in order to navigate to the stage 4 proposals.


timoxley:
Also consider changing "Stage 1 and Above Proposals" to "Active Proposals" to mirror nomenclature on https://github.com/tc39/proposals#active-proposals.

bterlson:
Thanks!

Issue #710, opened 2016-10-12
Module namespace @@iterator seems strange


domenic:
There is no other object in JS whose iterableness expresses a list of its keys. It seems strange that for module namespace objects, we are expecting people to write

``` js
for (const x of moduleNamespace) {
  console.log(moduleNamespace[x]);
}
```

whereas for other JS objects, we are expecting people to write

``` js
for (const x of Object.keys(obj)) {
  console.log(obj[x]);
}
```

It makes more sense to me to require the latter pattern for all objects.

Is it appropriate to suggest removing the iterator at this stage?

caridy:
I totally agree! I remember we touched on this a while ago, and it is just weird. Maybe @allenwb knows better, but maybe this was just a leaking definition from the time of the loader.

littledan:
cc @GeorgNeis @ajklein

ajklein:
This should be closed, as `@@iterator` has been removed from the spec.

PR #747, opened 2016-12-06
Remove @@iterator from Module Namespace objects

### LINKED ISSUE #693: Symbols on Module Namespace Exotic Objects
Module Namespace Exotic Objects have [two (own) symbol properties](https://tc39.github.io/ecma262/#sec-module-namespace-objects).

[[[Get]]](https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver), [[GetOwnProperty]], and [[HasProperty]] for these have the ordinary semantics, but [[Delete]], [[Set]] and [[DefineProperty]] don't (e.g. [[DefineProperty]] always fails).  This seems inconsistent and I'm wondering if it is a bug.  (Also note that these symbols are set up as configurable).

ajklein:
Per consensus at the November 2016 meeting, Module Namespace exotic
objects no longer have an @@iterator method.

Also changes the [[Configurable]] property of the @@toStringTag
property to false, as per consensus in the same meeting (and
because it's impossible to reconfigure due to the exotic
methods of such objects).

This closes #693 and #710.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -35965,26 +35965,13 @@ THH:mm:ss.sss
   <emu-clause id="sec-module-namespace-objects">
     <h1>Module Namespace Objects</h1>
     <p>A Module Namespace Object is a module namespace exotic object that provides runtime property-based access to a module's exported bindings. There is no constructor function for Module Namespace Objects. Instead, such an object is created for each module that is imported by an |ImportDeclaration| that includes a |NameSpaceImport|.</p>
-    <p>In addition to the properties specified in <emu-xref href="#sec-module-namespace-exotic-objects"></emu-xref> each Module Namespace Object has the following own properties:</p>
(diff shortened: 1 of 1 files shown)


littledan:
LGTM

bakkot:
Test262 collateral: https://github.com/tc39/test262/pull/810, https://github.com/tc39/test262/pull/811.

GeorgNeis:
There's still something odd about @@toStringTag, which is that deleting it returns true, although it has no effect.  Deleting any other existing property returns false.

To change this, I suggest the following steps to be inserted between steps 2 and 3 of https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects-delete-p.

- Let symbolKeys be ! OrdinaryOwnPropertyKeys(O).
- Append all the entries of symbolKeys to the end of exports.

(There's only one symbol but I'm just reusing existing language.)

Or, alternatively:

- If Type(P) is Symbol, return OrdinaryDeleteProperty(O, P).

PR #748, opened 2016-12-06
Allow resetting a Module Namespace object's prototype to null

ajklein:
This makes its [[SetPrototypeOf]] method consistent with
other objects whose prototypes are immutable.

TC39 came to consensus on this change at the November 2016 meeting.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -8007,7 +8007,7 @@
               Null
             </td>
             <td>
-              This slot is initialized to *null* at creation time and never changes (see <emu-xref href="#sec-module-namespace-exotic-objects-setprototypeof-v"></emu-xref>).
+              This slot is always *null* (see <emu-xref href="#sec-module-namespace-exotic-objects-setprototypeof-v"></emu-xref>).
             </td>
           </tr>
           </tbody>
--- a/spec.html
+++ b/spec.html
@@ -8019,6 +8019,7 @@
         <p>When the [[SetPrototypeOf]] internal method of a module namespace exotic object _O_ is called with argument _V_, the following steps are taken:</p>
         <emu-alg>
           1. Assert: Either Type(_V_) is Object or Type(_V_) is Null.
+          1. If SameValue(_V_, *null*) is *true*, return *true*.
           1. Return *false*.
         </emu-alg>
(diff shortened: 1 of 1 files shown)


littledan:
LGTM

ajklein:
@ljharb That's actually what I had proposed to do, but it's a rather more significant diff: need to change the sentence that says that we override all the internal methods, need to add a [[Prototype]] slot to the object, need to still write a [[SetPrototypeOf]] that dispatches. Curious what @bterlson thinks.

bterlson:
I would say don't say the slot is inherited from immutable prototype exotic objects but instead invoke 9.4.7.2 SetImmutablePrototype from 9.4.6.2 [[SetPrototypeOf]].

ajklein:
@bterlson Okay, I've gone the route we originally discussed, please take a look.

bterlson:
@ajklein looks good I think, but don't update the es6ids. I should just remove those at this point.

ajklein:
@bterlson ok, amended the commit to avoid touching the es6ids.

spec.html
```diff
@@ -7999,27 +7999,28 @@
...
           </tr>
+          <tr>
+            <td>
+              [[Prototype]]
+            </td>
+            <td>
+              Null
+            </td>
+            <td>
+              This slot is initialized to *null* at creation time.
```
bterlson: Maybe "This slot is always null (see [[SetPrototypeOf]])"?

bterlson:
Sorry playing with the review options :-P Not critical but I think the clarification would be nice.

spec.html
ajklein: Added a parenthetical explanation.

ajklein: Okay now I just said exactly what you recommended :)

bakkot:
Are there test262 tests? If no, should at least open an issue over there.

ajklein:
@bakkot There exists a test for the non-null case; it would be trivial to extend it to test what happens when null is passed, though maybe the "right" way to do this in test262 is to add another test for that case?

ajklein:
@bakkot opened a pull request in tc39/test262#805

PR #755, opened 2016-12-19
Normative: Invoke superclass constructor even if its .prototype is null

### LINKED ISSUE #699: Classes extending a function with .prototype=null
[Currently](https://tc39.github.io/ecma262/#sec-runtime-semantics-classdefinitionevaluation) (see ClassDefinitionEvaluation 6.g.i, 10.a, and 15), class `B` below is considered to be a base class.

``` js
function A(){ this.x = 1; }
A.prototype = null;

class B extends A {}
```

From what I can tell, this means that `A` will never be invoked - i.e., this is identical to `class B extends null {}`, except that the `[[Prototype]]` of `B` will be `A`. This seems wrong. I believe instead that 10.a and 15 should read "if _superclass_ is not **null**" instead of referring to "_protoParent_".

bakkot:
Fixes #699: currently the below `new` does not invoke `A`; with this change it does.

```js
function A() { this.x = 1; }
A.prototype = null;

new class extends A {}
```

As mentioned there, this is a normative change, but it causes the spec to match current behavior on (at least) Chrome and Firefox and so probably isn't breaking.

At the moment it does nothing to resolve the scoping issue mentioned [here](https://github.com/tc39/ecma262/issues/699#issuecomment-254676629).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19070,7 +19070,7 @@
         1. If |ClassBody_opt| is not present, let _constructor_ be ~empty~.
         1. Else, let _constructor_ be ConstructorMethod of |ClassBody|.
         1. If _constructor_ is ~empty~, then
-          1. If |ClassHeritage_opt| is present and _protoParent_ is not *null*, then
+          1. If |ClassHeritage_opt| is present and _superclass_ is not *null*, then
             1. Let _constructor_ be the result of parsing the source text
               <pre><code class="javascript">constructor(... args){ super (...args);}</code></pre>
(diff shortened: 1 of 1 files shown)


bakkot:
test262 collateral [here](https://github.com/tc39/test262/pull/806), since this is a normative change.

bterlson:
This seems fine to me (and the semantics I intended when implementing a35f0f3). I'll think about some way to avoid the "scoping" issue. Thanks for the test262 collateral!

PR #759, opened 2016-12-29
Editorial: annotate ToObject calls with `!`

ajklein:
These calls are preceded by checks for `null` and `undefined`,
so they're guaranteed to succeed. But not all such calls were
annotated with `!`.


bterlson:
Thank you!

PR #765, opened 2017-01-12
Correct all references to "well known" Symbols

aweary:
The spec first defines them as "well-known" so this just corrects a few places where they were referred to as "well known", which could cause some problems for anyone trying to search the spec for all references to well-known symbols.


bterlson:
Very nice.

PR #767, opened 2017-01-18
Make [[Delete]] on module namespace return false for @@toStringTag.

GeorgNeis:
A module namespace exotic object has an own Symbol.toStringTag property, which
is non-configurable.  This patch changes [[Delete]] to return false rather than
true for this property, thus making its behavior consistent with other own
properties of the object.

See also: #747


bterlson:
This looks good to me. I'll tag it needs consensus just in case, since it's easy to discuss next week :)

bterlson:
We have consensus.

bterlson:
Still need tests but I'm going to merge this.

PR #768, opened 2017-01-20
SAB integration even sections

syg:
I ended up putting the memory model as its own top-level clause as the last clause.

The informative guideline sections of the proposal are still unintegrated. I'm unsure where to put them or if they should be put in.


bterlson:
I think the information is quite valuable so including the informative sections as notes seems appropriate.

PR #771, opened 2017-01-20
Add TZOffset{UTC,Local} in place of LocalTZA/DSTA

### LINKED ISSUE #725: Ambiguities in UTC()'s spec and mis-specification of DaylightSavingTA()
* [0] http://www.ecma-international.org/ecma-262/7.0/index.html#sec-localtime
* [1] http://www.ecma-international.org/ecma-262/7.0/index.html#sec-utc-t
* [2] http://www.ecma-international.org/ecma-262/7.0/index.html#sec-daylight-saving-time-adjustment

As noted at the end of [1], UTC(Localtime(t)) is not always equal to t.
Indeed, Localtime(UTC(t)) is not always equal to t, either, although neither [0] nor [1] notes this.
Nor does [1] address the question of what to do when the parameter passed to UTC is either invalid (because it represents a time that local time skipped over when starting DST, during a "spring forward") or ambiguous (because it represents a time that was reused at the end of DST - first in DST then in standard time - in a "fall back").

The spec [2] of DaylightSavingTA() neglects to say that the input parameter, t, is presumed to be a UTC time; this is, however, made clear by the way it is used in the stipulated implementations of UTC() and Localtime().  Those uses make clear that, for any t, LocalTZA + DaylightSavingTA(t) should give local time's applicable offset from UTC at time t.
The spec [2] should make clear that t is presumed to be given in UTC.

LocalTZA is specified to be local time's present (at run-time, not at time t) standard offset from UTC, without any DST correction.  If local time has ever had a change of standard offset, the tacit requirement that LocalTZA + DaylightSavingTA(t) was local time's applicable offset from UTC at time t then comes into conflict with the wording of the specification for DaylightSavingTA(); when t refers to a time the other side (from run-time) of a standard offset change in local time, DaylightSavingTA(t) needs to incorporate both the standard offset change and any DST that may have been in effect at time t, while the wording of the spec says it should only return the DST offset applicable at that time.

In the case of Pacific/Kiritimati looking at times before 1995, or Pacific/Apia before 2012, this implies a DaylightSavingTA() of 24 hours at times when DST was not in effect.  For a zone that's now given up DST in favour of setting their standard offset to what used to be their summer offset, this shall mean reporting a negative DaylightSavingTA() before the change for non-DST times, while reporting a zero offset for when DST was in force.

The cause of this is the oversimplification of using LocalTZA as base offset and treating all deviations from it as DST.  In practice, a good time offset API would have an OffsetFromUTC(t) method, taking a UTC time and returning the actual offset at that time; for legacy purposes, you can define LocalTZA as at present and DaylightSavingTA(t) as OffsetFromUTC(t) - LocalTZA.  There may be some use to also providing StandardOffsetFromUTC(t) and DaylightSavingAdjustment(t) with which to decompose OffsetFromUTC(t), but it's OffsetFromUTC(t) that most code is going to actually care about.

You can then define Localtime(t) to be t + OffsetFromUTC(t).  I encourage you to, at the same time, change the spec of UTC(); doing that right shall be contentious, but I do recommend looking at python's definitions [3] in this area.

* [3] https://www.python.org/dev/peps/pep-0495/

Some mechanism for disambiguating fall-back hours is necessary; while client code shall often now know the needed hint, requiring sensible default behaviour (and matching present behaviour shall make sense for that), it is worth providing some mechanism for callers to
* indicate when they do know whether DST was in effect, over-riding the default; and
* suppress default behaviour in some way that'll let them address ambiguity/invalidity themselves, for example by catching an exception.

Whatever mechanism disambiguates fall-backs can also be used to decide what to do with invalid local time values in a spring forward.  (My usual model for these is that someone stepped forward or backwards an hour, a day or a week from a valid time and landed in a gap.  If the prior time was in DST, that step (backwards) should have landed in the hour before the spring forward; if the prior time was in standard time, the step (forward) should land in the hour after.  So I select the time that actually contradicts the alleged DSTness, where handling of a fall-back seeks to match it.)  Actual changes in a zone's standard offset can be handled the same way as a DST transition in the same direction.

Ideally it should be the case that UTC(t) returns a u for which u + OffsetFromUTC(u) is indeed t.  Note, however, that OffsetFromUTC(u) need not be equal to OffsetFromUTC(t), so t - OffsetFromUTC(t) is not an adequately reliable answer, although it may be the best you can do, unless you arrange for your local-time values to come with an extra bit (or trit, to allow "unknown" as an option) of data, indicating their (presumed) DST-ness or some equivalent truth (such as python's "fold").

Time software would all be so much simpler if we all just stuck to UTC; I recommend doing so for all *internal* types and only meddling with the political mess of time zone variation at the very surface of software, where users are exposed to the data.  Local time may seem like a convenience, but avoiding bugs at transitions annihilates what little convenience it actually confers.

I may be reached as edward.welbourne@qt.io (and shall now fix Qt's V4 engine to be less broken than it presently is - only as broken as the spec requires it to be).

jungshik:
Currently, LocalTimezoneAdjustment is assumed to be constant across time
zone rule changes and does not take any argument. Daylight Savings Time
Adjustment does take an argument t.

DSTA(t) can 'absorb' the standard time zone offset changes (e.g. Europe/Moscow
changed the standard time zone offset multiple times in the last decade.). However,
the spec stipulates that it should only reprsent an additional
adjustment necessary when DST is in effect.

In practice, at least two implementations (Spidermonkey/Firefox and
JSC/Safari) do not follow the spec to the letter to produce results
expected by users across time zone rule changes.

This PR drops LocalTZA and DaylightSavingsTimeAdjustment(t) and
introduce TZOffsetUTC(t) and TZOffsetLocal(t_local). The former takes
't' (UTC) and gives a time zone offset in effect at t for the local
time zone (for US ET, it'd be ```4*msPerHour``` in summer and ```5*msPerHour``` in
winter). The latter takes 't_local' (local time value) and gives a time
zone offset in effect at t_local.

It's also specified that TZOffsetLocal(t_local) will return the offset
*after* the transition when t_local is wall time repeating multiple
times at a negative transition (e.g. fall backward) or skipped wall time
at a positive time zone transition (e.g. spring forward). This is in
line with the aforementioned two implementations. In the future, it
may as well be considered to introduce options to control this behavior.
[1] [2]

UTC(t_local) and Localtime(t) are reformulated with
TZOffsetLocal(t_local) and TZOffsetUTC(t).

Will fix #725 modulo an option to specify the behavior to handle
skipped or repeated wall time.

[1] ICU Calendar API has skipped wall time option (https://goo.gl/h0bP26) and
repeated wall time option (https://goo.gl/Q1VX3j).

[2] Python proposal to handle skipped/repeated time:
    https://www.python.org/dev/peps/pep-0495/

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -25946,8 +25946,8 @@
       <!-- es6num="20.3.1.8" -->
       <emu-clause id="sec-time-zone-offset-from-localtime">
         <h1>Time Zone Offset from LocalTime</h1>
-        <p>An implementation of ECMAScript is expected to determine the local time zone offset from local time TZOffsetLocal(<emu-eqn>_t_<sub>local</sub></emu-eqn>) using the best available information on time zones. The time zone offset from local time  <emu-eqn>_t_<sub>local</sub></emu-eqn> is a value <dfn>TZOffsetLocal(<emu-eqn>_t_<sub>local</sub></emu-eqn>)</dfn> measured in milliseconds which when added to the local time represents UTC.</p>
-        <p>When <emu-eqn>_t_<sub>local</sub></emu-eqn> represents wall time repeating multiple times at a negative time zone offset transition (e.g. when the daylight savings time ends or the time zone offset is decreased) or skipped wall time at a positive time zone offset transitions (e.g. when the daylight savings time starts or the time zone offset is increased), <emu-eqn>_t_<sub>local</sub></emu-eqn> should be interpreted with the time zone offset after the transition.</p>
+        <p>An implementation of ECMAScript is expected to determine the local time zone offset from local time TZOffsetLocal(<emu-eqn>_t_<sub>local</sub></emu-eqn>) using the best available information on time zones. The time zone offset at the local time <emu-eqn>_t_<sub>local</sub></emu-eqn> is a value <dfn>TZOffsetLocal(<emu-eqn>_t_<sub>local</sub></emu-eqn>)</dfn> measured in milliseconds which when added to the local time represents UTC.</p>
+        <p>When <emu-eqn>_t_<sub>local</sub></emu-eqn> represents wall time repeating multiple times at a positive time zone offset transition (e.g. when the daylight savings time ends or the time zone offset is increased) or skipped wall time at a negative time zone offset transitions (e.g. when the daylight savings time starts or the time zone offset is decreased), <emu-eqn>_t_<sub>local</sub></emu-eqn> should be interpreted with the time zone offset after the transition.</p>
         <emu-note>
           <p>It is recommended that implementations use the time zone information of the IANA Time Zone Database <a href="http://www.iana.org/time-zones/">http://www.iana.org/time-zones/</a>.</p>
           <p>1:30 AM on November 5, 2017 in America/New_York is repeated twice (fall backward), but it should be interpreted as 1:30 AM UTC-05 instead of 1:30 AM UTC-04. TZOffsetLocal should be <emu-eqn>+5 &times; msPerHour</emu-eqn>.</p>
@@ -25963,7 +25963,7 @@
           1. Return _t_ - TZOffsetUTC(_t_).
         </emu-alg>
         <emu-note>
-          <p>Two different time values (_t_ (UTC)) are converted to the same local time <emu-eqn>t<sub>local</sub></emu-eqn> at a negative time zone transition when there are repeated times (e.g. the daylight savings time ends or the time zone offset is decreased.).</p>
+          <p>Two different time values (_t_ (UTC)) are converted to the same local time <emu-eqn>t<sub>local</sub></emu-eqn> at a positive time zone offset transition when there are repeated times (e.g. the daylight savings time ends or the time zone offset is increased.).</p>
         </emu-note>
       </emu-clause>
 
--- a/spec.html
+++ b/spec.html
@@ -25935,20 +25935,23 @@
       </emu-clause>
 
       <!-- es6num="20.3.1.7" -->
-      <emu-clause id="sec-local-time-zone-adjustment">
-        <h1>Local Time Zone Adjustment</h1>
-        <p>An implementation of ECMAScript is expected to determine the local time zone adjustment. The local time zone adjustment is a value <dfn>LocalTZA</dfn> measured in milliseconds which when added to UTC represents the local <em>standard</em> time. Daylight saving time is <em>not</em> reflected by LocalTZA.</p>
+      <emu-clause id="sec-time-zone-offset-from-utc">
+        <h1>Time Zone Offset from UTC</h1>
+        <p>An implementation of ECMAScript is expected to determine the local time zone offset from UTC TZOffsetUTC(_t_) using the best available information on time zones. The time zone offset from UTC at time represented by time value _t_ (UTC) is a value <dfn>TZOffsetUTC(_t_)</dfn> measured in milliseconds which when subtracted from UTC represents the local time.</p>
         <emu-note>
           <p>It is recommended that implementations use the time zone information of the IANA Time Zone Database <a href="http://www.iana.org/time-zones/">http://www.iana.org/time-zones/</a>.</p>
         </emu-note>
       </emu-clause>
 
       <!-- es6num="20.3.1.8" -->
-      <emu-clause id="sec-daylight-saving-time-adjustment">
-        <h1>Daylight Saving Time Adjustment</h1>
-        <p>An implementation dependent algorithm using best available information on time zones to determine the local daylight saving time adjustment DaylightSavingTA(_t_), measured in milliseconds. An implementation of ECMAScript is expected to make its best effort to determine the local daylight saving time adjustment.</p>
+      <emu-clause id="sec-time-zone-offset-from-localtime">
+        <h1>Time Zone Offset from LocalTime</h1>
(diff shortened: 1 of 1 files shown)


jungshik:
@bterlson, @ediosyncratic, @littledan 

Can you take a look?  I ended up making a bit more extensive change than I thought at first. 

@bterlson, I hope it's not too late for review at a ecma262 meeting next week. 

BTW, I may not have done the right thing when making this PR in terms of what file to touch etc. Somehow, 'npm install', 'npm run build' etc didn't work on my Ubuntu (14.x). So, I just edited spec.html trying to follow the mark-up style as faithfully as I can.

jungshik:
The update is to make the sign of TZOffSetUTC and TZOffsetLocal to be the same as that of Date.prototype.getTimezoneOffset. It's unfortunate that the sign of getTimezoneOffset is  the opposite of what most people consider 'intuitive'.   

An alternative would be not to use 'Offset' in the names of two operations I'm introducing and make the sign be in line with a  more common usage (negative for timezone West of the primary meridian and positive for timezones East of the primary meridian) .  Instead of 'Offset', 'Adjustment' can be used.

Yet another alternative would be to note that the sign of offsets returned by two operations is the opposite of getTimezoneOffset() and keep the name ('offset')  in my PR.

jungshik:
I have another thought.  Instead of introducing two new abstract operations in place of LocalTZA and DSTA(t),  I can make LocalTZA take two arguments, t and is_local_time.  When is_local_time is true, 't' would be interpreted as a 'local time value' in the current time zone, while 't' would be time value in UTC when it's false.

The reason for is_local_time (boolean) is that 't' needs to be interpreted differently depending on whether 't' is UTC or local time in the current time zone. 

That way, I don't have to worry about the sign of TZOffSet{UTC,Local} and can also avoid rather clumsy name with UTC or Local at the end. 

I'll make a new PR with this update (to avoid clutter with multiple commits). I'll try to get in this afternoon (UTC-0800). Hope it's not too late for TC39 review this week.

jungshik:
I'm closing this PR in favor of #778.

spec.html
```diff
@@ -25935,20 +25935,23 @@
...
         </emu-note>
       </emu-clause>
 
       <!-- es6num="20.3.1.8" -->
-      <emu-clause id="sec-daylight-saving-time-adjustment">
-        <h1>Daylight Saving Time Adjustment</h1>
-        <p>An implementation dependent algorithm using best available information on time zones to determine the local daylight saving time adjustment DaylightSavingTA(_t_), measured in milliseconds. An implementation of ECMAScript is expected to make its best effort to determine the local daylight saving time adjustment.</p>
+      <emu-clause id="sec-time-zone-offset-from-localtime">
+        <h1>Time Zone Offset from LocalTime</h1>
+        <p>An implementation of ECMAScript is expected to determine the local time zone offset from local time TZOffsetLocal(<emu-eqn>_t_<sub>local</sub></emu-eqn>) using the best available information on time zones. The time zone offset at the local time <emu-eqn>_t_<sub>local</sub></emu-eqn> is a value <dfn>TZOffsetLocal(<emu-eqn>_t_<sub>local</sub></emu-eqn>)</dfn> measured in milliseconds which when added to the local time represents UTC.</p>
```
ediosyncratic: This function needs a second, optional parameter, by which to sanitize spring-forward gaps and disambiguate fall-back duplications.  Given that you want the default behaviour to be "use the offset that was in effect before the transition", that optional parameter needs to select between before and after; so I suggest it be called prior and default to true; it says whether to use the offset in effect before a transition, when the specified time falls into one; if given and false, it selects the offset in effect after the transition.

spec.html
```diff
@@ -25935,20 +25935,23 @@
...
+        <p>When <emu-eqn>_t_<sub>local</sub></emu-eqn> represents wall time repeating multiple times at a positive time zone offset transition (e.g. when the daylight savings time ends or the time zone offset is increased) or skipped wall time at a negative time zone offset transitions (e.g. when the daylight savings time starts or the time zone offset is decreased), <emu-eqn>_t_<sub>local</sub></emu-eqn> should be interpreted with the time zone offset after the transition.</p>
```
ediosyncratic: No need for "multiple times" - given that "repeating" implicitly says that - and explicitly saying "multiple times" rather than "twice" suggests more than twice, which does never happen and we have no sane way to cope with it if it did.

also: s/the daylight savings time/daylight-saving time/g
... and, at the end, s/after/before/, but description of the argument "prior" should cover this.

spec.html
```diff
@@ -25957,19 +25960,22 @@
...
 
       <!-- es6num="20.3.1.10" -->
       <emu-clause id="sec-utc-t" aoid="UTC">
-        <h1>UTC ( _t_ )</h1>
-        <p>The abstract operation UTC with argument _t_ converts _t_ from local time to UTC. It performs the following steps:</p>
+        <h1>UTC ( <emu-eqn>_t_<sub>local</sub></emu-eqn> )</h1>
+        <p>The abstract operation UTC with argument <emu-eqn>_t_<sub>local</sub></emu-eqn> converts <emu-eqn>_t_<sub>local</sub></emu-eqn> from local time to UTC. It performs the following steps:</p>
         <emu-alg>
-          1. Return _t_ - LocalTZA - DaylightSavingTA(_t_ - LocalTZA).
+          1. Return _t_ + TZOffsetLocal(<emu-eqn>_t_<sub>local</sub></emu-eqn>)
```
ediosyncratic: Missing subscript local on first t.

PR #398, opened 2016-02-19
Editorial: introduce %ObjProto_valueOf% for usage by the Location exotic object

annevk:
See https://github.com/whatwg/html/pull/638#issuecomment-185890302 for
context.

PR #775, opened 2017-01-23
ES2017 review

anba:
I've kept the changes in small commits so it's easier to review them.


domenic:
Maybe we should move the HostEnsureCanCompileStrings check into CreateDynamicFunction?

anba:
> Maybe we should move the HostEnsureCanCompileStrings check into CreateDynamicFunction?

That's fixed in the next commit 3f36151.

domenic:
I see, I didn't expand the commit description; sorry for the noise!

spec.html
```diff
@@ -8258,7 +8258,6 @@
 
       <emu-clause id="sec-set-immutable-prototype">
         <h1>SetImmutablePrototype ( _O_, _V_ )</h1>
-        <!-- This is used by host environment specs, even though it only has one consumer in this spec; don't inline it -->
```
ljharb: How is this stale? I don't see anywhere else in the spec that references it (unless somewhere else in this PR references it?)

Either way, it seems prudent to leave it in, in case the other uses later get factored out.

spec.html
```diff
@@ -33566,8 +33566,8 @@
 
       <!-- es6num="24.1.2.1" -->
       <emu-clause id="sec-arraybuffer-length">
-        <h1>ArrayBuffer ( [ _length_ ] )</h1>
-        <p>When the `ArrayBuffer` function is called with optional argument _length_, the following steps are taken:</p>
```
ljharb: Doesn't this only need to be changed if it's web-incompatible to change it?

spec.html
```diff
@@ -27139,7 +27139,7 @@
     <!-- es6num="21.1.3" -->
     <emu-clause id="sec-properties-of-the-string-prototype-object">
       <h1>Properties of the String Prototype Object</h1>
-      <p>The String prototype object is the intrinsic object <dfn>%StringPrototype%</dfn>. The String prototype object is an ordinary object. The String prototype is itself a String object; it has a [[StringData]] internal slot with the value *""*.</p>
+      <p>The String prototype object is the intrinsic object <dfn>%StringPrototype%</dfn>. The String prototype object is an exotic String object and has the internal methods specified for such objects. It has a [[StringData]] internal slot with the value *""*.</p>
```
ljharb: Is it exotic solely because of the [[StringData]] internal slot?

spec.html
anba: There are now two callers of SetImmutablePrototype, so it's unlikely to get inlined.

spec.html
anba: Changing the length to zero is inconsistent with other functions, e.g. we could also change DataView.prototype.setInt32's length to zero because that function also "works" when called without any arguments, but the expectation is that setInt32 is called with at least two arguments. And the same applies for ArrayBuffer, where we expect it's called with a single length argument.

spec.html
anba: An object is only exotic if it uses a different implementation for any of the internal methods. And since String.prototype is a String object instance, and String objects are exotic String objects, String.prototype is also exotic.

spec.html
ljharb: Where are those two? I was unable to find them.

ljharb: Either way tho, I think the comment is an important marker for the external dependency, and I'd expect every abstract operation that's referenced externally to be similarly noted.

spec.html
ljharb: Gotcha - and for my edification, why are String objects exotic (which internal methods does it differ for)?

anba: [[GetOwnProperty]], [[OwnPropertyKeys]], and (soon) [[DefineOwnProperty]] are overridden for exotic string objects in https://tc39.github.io/ecma262/#sec-string-exotic-objects. This is needed to support `new String("test")[0] === "t"`, because the indexed properties aren't _real_ properties but instead are defined as _virtual_ properties, similar to how this is implemented in browsers.

spec.html
anba: SetImmutablePrototype is used in 9.4.6.1 and 9.4.7.1. And we don't control which abstract operations are used by external parties, so we can't really mark every operation that's used externally. Allen's proposal in #398 seems more doable than having an incomplete coverage of operations which may or may not be used somewhere else.

spec.html
ljharb: Makes sense.

bterlson:
Reviewed most of these, working on thinking through a few of the normative-y ones. Any opinions on whether I should squash this monster or keep the individual commits?

ljharb:
The individual commits are super nice for review, and would also be nice for git blame.

michaelficarra:
Keep the individual commits. A lot of the spelling/typo commits can be merge though IMO.

bterlson:
I have a rebased branch with individual commits with updated commit messages. I'll do a rebase -i before I merge and join some like commits, but mostly leave them separate. Thanks for opinions :)

bterlson:
Pushed most of the commits individually (other than the ones I called out here). I guess I don't see a reason to collapse :)

In terms of release, none of the ones I left behind are bad to take during the RF opt-out period, so we will have some time to get them in before ES2017 is final-final.

anba:
> Pushed most of the commits individually (other than the ones I called out here).

These five commits seem to be the ones left out:
65b2c2702c97c9516b1e1962671bf873c7d97b3d
81da5a0a4fc1076ad8de841ac97ee233c613f20c
88d70d7030952bd972ef151bac10e6f7e304fb83
72aa1dec5fcbeb49573108ad2ba00004a69c31a9
0cd8543f6ae8fa187266bca6e151941c56e2ada1

The commit message for 65b2c2702c97c9516b1e1962671bf873c7d97b3d should be changed, because it's incorrect (seems like I got confused by the current use of `super` restrictions in ch14). So basically the current spec text is correct w.r.t. the super-property restrictions, but applying 65b2c2702c97c9516b1e1962671bf873c7d97b3d (with a corrected commit message!) makes the super-property restrictions more consistent with the ones in ch14.

bterlson:
65b2c27: IIUC, this has the effect of making SuperCall allowed in functions in eval. If this is correct, that doesn't align with clause 14 rules. Thoughts @anba?
81da5a0: I'm reluctant to take this change without hearing from some JSC folks (@msaboff)

88d70d7: Pulled (f608ef7aa)
72aa1de: Pulled (c934bf60)
0cd8543: Pulled (d88d05fdfc44e71f0d1f37228dc53a6b54391e1b)

bterlson:
Thinking again about 65b2c27, SuperCall is of course not allowed due to the "... outside constructors".  I'm just not sure I see the mapping to Clause 14.

I dislike this whole "... outside blah" notation after some experience with it.

bterlson:
Forgot to update to say:

65b2c27: Pulled (9c6e5b9f6a2e117073488fbf26b052c12e296846)

@anba would you be open to making a PR for 81da5a0? I'll get to it later this week if you don't :)

bakkot:
Some of these changes are normative; test262 will need updates to match.

Normative changes which have gone in include at least the following, though not all of them make sense to test:
- c934bf601df478a389745891819a4a16e04b1bf3
- f608ef7aa1cf626df4df238dd90a3a116c34c101
- 0ec1b35b90988e7af7a3ddaef8775f1656c2bb08
- bc33e1040762ad2208a1820ef5a0d8e74302d124
- d6f1b47f282f97039afa01d7351c9b9b62fbb5fa
- 18e18cfa008f851b27e66342ae1c4893d7434caa
- 56f8ac2e85483e883a27ba5dc5f65e7b5444db70
- 0e56ac5dbaf35fb3636e04d7a356ba126b5703e0
- 71a484b7adeb368326868c83f2441149ff23f52a
- c97a0b9335c11a991cc5dbde80bc01496b7f5ef9
- c023f2be43a38146e4aee942f77520b1a24a0510
- 14cc1a83d1a75126e552a5f29976c9380b3ae7c8
- 19a8bd406895a02803fad11db6f879af3cd3b182
- 505efc43fa49b4d0032ec71639eca23579fda220
- 54837352cf1bc7f74588f6b13e0c79192d612c3d
- 993e8e425a600ef3d925b5e2430fa0f687206fed
- 8271b2d810c2940a1f80faac47a5317782ede518
- 44c319f5166700e3c52b4ceea9f2a3795d09e9ff
- d007ed3ee898995602be56a9a7194e12ffdb6085
- 5eb54891e3c96c7ad25a9916d095542423ad96be
- a3b0507540c8891b32bf036efc2b90ee3f8e2435
- 38db4ff1452183ba581ec96407654171676eee2a

anba:
> @anba would you be open to making a PR for 81da5a0? I'll get to it later this week if you don't :)

Filed #810.

bterlson:
This has all been merged! Thanks @anba.

PR #778, opened 2017-01-23
Make LocalTZA take 't' and 'isUTC' and drop DSTA(t).

### LINKED ISSUE #725: Ambiguities in UTC()'s spec and mis-specification of DaylightSavingTA()

jungshik:
Currently, LocalTimezoneAdjustment is assumed to be constant across time
zone rule changes and does not take any argument. Daylight Savings Time
Adjustment does take an argument t.

DSTA(t) can 'absorb' the standard time zone offset changes
(e.g. Europe/Moscow changed the standard time zone offset multiple times
in the last decade.). However, the spec stipulates that it should only
reprsent an additional adjustment necessary when DST is in effect.

In practice, at least two implementations (Spidermonkey/Firefox and
JSC/Safari) do not follow the spec to the letter to produce results
expected by users across time zone rule changes.

This PR revises LocalTZA to accept two arguments, 't' and 'isUTC' and
drop DaylightSavingsTimeAdjustment(t). When isUTC is true, LocalTZA
interprets 't' as a time value representing UTC and gives a time zone
offset in effect at time t for the local time zone (for US ET, it'd be
```-4*msPerHour``` in summer and ```-5*msPerHour``` in winter). When isUTC is false,
't' is interpreted as local time value and gives a time zone offset in
effect at t = t_local.  

It's also specified that LocalTZA(t = t_local, false) will return the offset
*before* the transition when t_local is local time repeating multiple
times at a negative transition (e.g. fall backward) or skipped local time
at a positive time zone transition (e.g. spring forward). 

Currently, implementations do not agree with each other as to how to handle repeating
or skipped local time. This change will standardize the behavior in line with the way Java 8 does.

Finally, UTC(t) and Localtime(t) are reformulated with LocalTZA(t = t_local, false) and
LocalTZA(t = t_UTC, true).

Will fix #725.

### CODE REVIEW

**Comment by jungshik** [comment]
@bterlson, @ediosyncratic, @littledan : could you take a look?  I like this better than #771. 

Thanks.

**Comment by mj1856** [comment]
I'm very glad that there is an attempt to normalize this in the spec.  However I must say that the proposed standard is exactly backwards of what should be done (IMHO):

> 1:30 AM on November 5, 2017 in America/New_York is repeated twice (fall backward), but it must be interpreted as 1:30 AM UTC-05 instead of 1:30 AM UTC-04. 

This makes little sense.  1:30 UTC-5 is indeed the "standard" time, but it is the *second* of the two occurrences.  Time moves in a forward-only direction.  A person watching their clock would not see the first 1:30, and say to themselves "oh wait, not this one, I think I'll wait a bit for another".   They would act on the first occurrence - the "daylight" time occurrence.  Then logically when the second one came around, they would not act again because they already had done so.

>2:30 AM on March 12, 2017 in America/New_York does not exist, but it must be interpreted as 2:30 AM UTC-04.

This is also backwards.  2:30 UTC-4 does not exist in the US Eastern time zone.  Through math, it would be normalized to 1:30 UTC-5.  This is moving the wall time backward an hour.  A sane person in the local time zone would not do this.  They would not see that their regular time is 2:30 but remember to run an hour earlier because of the DST change they had not done yet!  Instead, they would at 2:00 advance their clocks to 3:00, then they would do the scheduled activity either right then, or at 3:30.  The latter is preferred, because if you had tasks at 2:05, 2:10, 2:15, 2:20, etc.. you would not do them ALL at the same time at 3:00.  Another way to think about this is: if a local time is non-existent then the local clock did not move forward.  Therefore, track the local time in place, and advance the clock forward by the amount of the gap.

These are the ONLY sensical options.  Moving backwards is nonsensical.  Time doesn't move backwards.

Note that this is exactly what Java 8 does in it's new time API, and is also what most production job scheduling systems will do.  It's also what we do in moment.js (when we have control over it with moment-timezone), and it's what NodaTime is switching to in its upcoming 2.0 release.

I strongly suggest that ECMAScript follow suit.

**Comment by mj1856** [comment]
FYI, I just checked in current versions of Chrome, IE, Edge, and Node.js, and they all follow my suggested implementation.

I also checked FireFox, and it follows the alternative implementation - the one that was originally suggested and that I am against.

I've checked this a few times over the years, and noticed that some browsers have flip-flopped on this point.  It should indeed be standardized - but towards the way that Chrome, IE, Edge and Node.js are currently implementing it.

**Comment by maggiepint** [comment]
Agree with Matt here, but have to add a couple semantics. It is properly `daylight saving time`. There is no `s`. Also, I'm generally fine with the term "wall time", but that in and of itself should be defined and codified. When I give conference talks on this matter, I will frequently use the term "local time" to mean basically the same thing, and others still will use "civil time". It is worth noting that the Java 8 documentation refers to it as local time: https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html

This is some of the best work in date/time out there, so their choices are good to model after.

**Comment by maggiepint** [comment]
One other thing of note - this is correct in not tracking a separate adjustment for daylight saving time. The 'normal' way to convert between UTC and local time in a timezone is to use a table that contains transition points for a time zone in UTC, and the new offset at that point, to determine what the correct offset is for a given point on the global timeline. This system will work for both DST, and for changes to a country's standard offset, as happened in several places in Russia in 2016.

**Comment by jungshik** [comment]
Thanks a lot for the review/feedback. 

A couple of points to clarify: 

1.  v8 is 'broken' across the timezone rule/offset  changes (e.g. multiple times in the last decade in Europe/Moscow) because it follows the current spec to its letters. 
See https://bugs.chromium.org/p/v8/issues/detail?id=3547

2. I didn't check v8's behavior when it works: that is, America/New_York. It's my mistake. Because of #1, I only checked JSC and Spidermonkey 
I didn't check Edge (I meant to), either.   I should have done more thorough investigation of the behavior of various implementations.

3. Having found that JSC and Spidermonkey behaved the way I spec'd (and the first point above about v8), I thought that it'd be better to make  the default behavior what current implementations (that handles not just ordinary DST transitions but also zone offset transitions)  do.   However, again, I should have checked v8 and Edge as well. 

With the tz set to America/New_York, Firefox does the following: 
```
(new Date(2017,10,5,1,30)).toUTCString()
"Sun, 05 Nov 2017 06:30:00 GMT"                  :  1:30 AM is interpreted as UTC-05
(new Date(2017,2,12,2,30)).toUTCString()
"Sun, 12 Mar 2017 06:30:00 GMT"                  : 2:30 AM is interpreted as UTC-04
```

v8 does the opposite, indeed.

```
V8 version 5.7.0 (candidate)
d8> (new Date(2017,10,5,1,30)).toUTCString()
"Sun, 05 Nov 2017 05:30:00 GMT"
d8> (new Date(2017,2,12,2,30)).toUTCString()
"Sun, 12 Mar 2017 07:30:00 GMT"
```



Because of the 4th point below, I thought it'd be better to consider adding an option  to control the behavior around time zone offset transition, but that is left for the future because it requires changing the public API signature. 

4. I'm not sure if we can say that one way to handle skipped and repeated wall time is better than others.  For instance, ICU offers three different ways [1] and Python's proposed spec offers two options [2].   

Anyway, I'm  perfectly fine with the default behavior as suggested by @mj1856 because having read his argument, I'm kinda persuaded that the behavior he wants seems to make sense (from the pov of a person going through the transition looking at a clock)  and I don't have any preference.  My primary goal is to get rid of the ambiguity. 

I'll revise the PR and ask you to take another look. Thanks again. 


[1] ICU Calendar API has skipped wall time option (https://goo.gl/h0bP26) and
repeated wall time option (https://goo.gl/Q1VX3j).

[2] Python proposal to handle skipped/repeated time:
https://www.python.org/dev/peps/pep-0495/

**Comment by jungshik** [comment]
@maggiepint 
>  The 'normal' way (if there is one - it it as least the way I know and have seen) to convert between UTC and local time in a timezone is to use a table that contains transition points for a time zone in UTC, and the new offset at that point, to determine what the correct offset is for a given point on the global timeline. This system will work for both DST, and for changes to a country's standard offset, as happened in several places in Russia in 2016.

Sure. I wish I'd not have to worry about converting t_local (t value calculated from y,M,d, h, m, s in local time) to t_UTC. That is,  I wish I'd not have to deal with LocalTZA(t, isUTC = false)). Unfortunately, the way Date constructor taking 'y,M,d,h,m,s' is spec'd (y,M,d,h.m,s are supposed to be interpreted in the local time), I cannot avoid it. See #725.

**Comment by mj1856** [comment]
FWIW, most APIs I've seen in other languages just pick one or the other, and don't offer much control.  Java 8, JodaTime, and NodaTime being the exception.

When I have had this discussion with library/platform authors 1 on 1, of those that chose the 1:30 UTC-5 option (in above context), they have often told me something like, "well, we weren't sure what to do, so we just went with the 'standard' time option".  Of course, there are the arguments I made earlier, and also as Maggie pointed out with Russia - that not every transition is due to DST.

I've heard a lot of use cases around this, and scheduling a daily event seems to be the one that catches on it most frequently - and in that scenario there really is no option but to pick the first ambiguous instance and shift invalid times forward.  That does push both of them into the "daylight" period, but so be it.

The only scenario I've ever personally encountered that might want the second of two instances is when a business such as a bar or movie theater might close at 1:30 AM, but for whatever reason wants to remain open until the second occurrence.  Even then - they are already thinking about this edge case and they can adjust for it manually if needed (since both 1:30's are valid).

I've *never* encountered any use case where moving an invalid time from the spring-forward gap backwards an hour is desired.  Not saying they don't exist - they're just not something I've seen.

**Comment by maggiepint** [comment]
I"m aware that the date constructor is specd for local time. I think I didn't phrase it well, but fundamentally all I am trying to convey is that the DST transition amount itself is temporal. Some countries do DST for a few years, then stop doing it, then start again. Lord Howe Island Australia does a 30 minute DST transition NOW, but I would be entirely unsurprised if they changed to one hour later. I'm not sure how this is handled right now, but I may not be reading close enough.

**Comment by jungshik** [comment]
Addressed the review comments. Can you take another look,  @mj1856  and @maggiepint ? Thanks !

**Comment by ediosyncratic** [comment]
@mj1856: "I've never encountered any use case where moving an invalid time from the spring-forward gap backwards an hour is desired." Suppose you have a flight early the morning after a transition and a moderately long journey to get to the airport.  Subtract travel time, needed to reach airport, from required arrival time.  If that lands you in the missing hour, you need to move backwards a further hour to get when to set off in good time to catch your flight.  Similar scenario with a fall-back can be a reason to want the second time round.  (Where I live, in CEST, the missing-or-repeated hour is 2 to 3; I believe there are places where it's 3 to 4.  Catching an early flight can indeed involve getting up that early.)

"The only scenario I've ever personally encountered that might want the second of two instances is when a business such as a bar or movie theater might close at 1:30 AM, but for whatever reason wants to remain open until the second occurrence. Even then - they are already thinking about this edge case and they can adjust for it manually if needed (since both 1:30's are valid)."  They're going to be paying their staff for another hour's work (possibly at elevated rates; after all, it *is* Sunday) and they may well have to get local licensing magistrates (or similar authorities) to approve their staying open an hour longer than normal; yup, they're going to know they're doing this.

I'm not a fan of making an existing numeric global variable, LocalTZA, into a function taking two parameters.  Surely that's going to require a lot of folk to change code; this cannot be good.  Web pages updated to work with the new spec shall break for users of older browsers; web pages still compatible with old browsers shall break in newer browsers.  The earlier spec, adding a new function that Does The Right Thing, looked better to me - albeit I never got round to reviewing it in detail (I've been busy, sorry).  I'll give this one a look next.

File: spec.html [hunk 1]
```diff
@@ -25936,44 +25936,44 @@
...
-        <h1>Daylight Saving Time Adjustment</h1>
-        <p>An implementation dependent algorithm using best available information on time zones to determine the local daylight saving time adjustment DaylightSavingTA(_t_), measured in milliseconds. An implementation of ECMAScript is expected to make its best effort to determine the local daylight saving time adjustment.</p>
-        <emu-note>
-          <p>It is recommended that implementations use the time zone information of the IANA Time Zone Database <a href="http://www.iana.org/time-zones/">http://www.iana.org/time-zones/</a>.</p>
+          <p>1:30 AM on November 5, 2017 in America/New_York is repeated twice (fall backward), but it must be interpreted as 1:30 AM UTC-04 instead of 1:30 AM UTC-05. LocalTZA(TimeClip(MakeDate(MakeDay(2017, 10, 5), MakeTime(1, 30, 0, 0))), false) is <emu-eqn>-4 &times; msPerHour</emu-eqn>.</p>
+          <p>2:30 AM on March 12, 2017 in America/New_York does not exist, but it must be interpreted as 2:30 AM UTC-05. LocalTZA(TimeClip(MakeDate(MakeDay(2017, 2, 12), MakeTime(2, 30, 0, 0))), false) is <emu-eqn>-5 &times; msPerHour</emu-eqn>.</p>.
         </emu-note>
```
ediosyncratic: This </emu-note> seems to be missing its opening <emu-note>

spec.html
```diff
@@ -25936,44 +25936,44 @@
 
       <!-- es6num="20.3.1.7" -->
       <emu-clause id="sec-local-time-zone-adjustment">
-        <h1>Local Time Zone Adjustment</h1>
-        <p>An implementation of ECMAScript is expected to determine the local time zone adjustment. The local time zone adjustment is a value <dfn>LocalTZA</dfn> measured in milliseconds which when added to UTC represents the local <em>standard</em> time. Daylight saving time is <em>not</em> reflected by LocalTZA.</p>
+        <h1>LocalTimeZoneAdjustment ( _t_, _isUTC_ )</h1>
```
jungshik: I do prefer 'offset' to 'adjustment', too. That's what I did in my earlier PR. However, the sign of Date.prototype.getTimezoneOffset is exactly the opposite of what's more common. That is, getTimeZoneOffset() returns a negative value for west of the primary meridian and a positive value for east of the primary meridian. 

I can use "offset' and match its sign with that of getTimeZoneOffset(), but I find it rather confusing when I talk about 'positive' offset change or 'negative' offset change. It's still a possibility, but I chose otherwise by sticking to the existing term in 'Local Timezone Adjustment'.

maggiepint:
Offset is the preferred terminology from the Moment.js team - for what it's
worth. I know that it's favored in other parts of the DateTime community as
well.

spec.html
```diff
@@ -25936,44 +25936,44 @@
...
+        <p>An implementation of ECMAScript is expected to determine the local time zone adjustment with respect to UTC using the best available information on time zones.</p>
+        <p>When _isUTC_ is true, <dfn>LocalTZA( _t_, true )</dfn> is the offset of the local time zone from UTC measured in milliseconds at time represented by time value _t_ (UTC). When it is added to _t_ (UTC), it yields the local time.</p>
+        <p>When _isUTC_ is false, <dfn>LocalTZA( _t_, false )</dfn> is the offset of the local time zone from UTC measured in milliseconds at local time represented by time value <dfn><emu-eqn>_t_<sub>local</sub> = _t_</emu-eqn></dfn>. When it is subtracted from the local time <emu-eqn>_t_<sub>local</sub>, it yields the corresponding UTC.</p>
```
jungshik: Adding a option to control the way to handle repeated and skipped time is beyond this PR. Note that my first commit message leaves it up to a consideration in the future, but I removed that responding to the review comment.  

Anyway, in the future, it can be still considered. 

This PR wants to minimize the change and get rid of the existing ambiguity by specifying the default behavior. Adding a option in the public API (note that this PR does not change the signature of any *public* API). 

> That's only used for three hours of each year, so it's more of a hint than anything else (we'll ignore it otherwise).

Is it true that it's only used for three hours of each year. As you noted and I also noted before, there are other cases where this issue of repeated or skipped time is important (namely, timezone rules changes as happened to Europe/Moscow, Pacific/Apia, etc.

spec.html
jungshik: > Changing the meaning of old names would break existing code; furthermore, names with "adjustment" in them, actually meaning "offset", are worth retiring in favour of better names.

Well,  LocalTZA is not a part of the public API. It's an internal abstract operation. Moreover, the existing code needs to be changed anyway. Spidermonkey and JSC did the exact opposite of what's specified here around a timezone offset change. v8 is broken in timezones where historic offset is different from the present tz offset, etc.

jungshik:
> Surely that's going to require a lot of folk to change code; this cannot be good. Web pages updated to work with the new spec shall break for users of older browsers; web pages still compatible with old browsers shall break in newer browsers. The earlier spec, adding a new function that Does The Right Thing, looked better to me - albeit I never got round to reviewing it in detail (I've been busy, sorry). I'll give this one a look next.

I'm afraid that you overlooked the fact that LocalTZA is **NOT a public API**. It's an internal abstract operation. So, Ecma262 implementors need to change their codes (they need to anyway because the current spec has an ambiguity and this PR gets rid of that ambiguity), but web page authors does not in general. Well, their workaround (if there's been any) or reliance on the existing implementations need to adjusted. 

Given the fact that (according to @mj1856 ) existing implementations changed the behavior over the time and different implementations have different behaviors, I don't think this change is any worse than other options.

jungshik:
> Offset is the preferred terminology from the Moment.js team - for what it's
worth. I know that it's favored in other parts of the DateTime community as
well.

See my comment 
https://github.com/tc39/ecma262/pull/778#discussion_r98055261  in response to @ediosyncratic. I do prefer offset to adjustment, too !.

jungshik:
Let me give more details on the awkwardness (at least as perceived by me) when we use "offset' and get its sign to match that of getTimeZoneOffset() (public API). 

The current PR has this (I spotted a mistake in 'increase' and 'decrease' in the previous commit and I fixed it in the latest commit. Thank you ! ).  
 
```
When <emu-eqn>_t_<sub>local</sub></emu-eqn> represents local time repeating multiple times at a negative time zone transition (e.g. when the daylight saving time ends or the time zone adjustment is decreased) or skipped local time at a positive time zone transitions (e.g. when the daylight saving time starts or the time zone adjustment is increased),
```

With the sign of offset'  to match that of getTimeZoneOffset() (public API), I'd have

```
 When <emu-eqn>_t_<sub>local</sub></emu-eqn> represents local time repeating multiple times at a positive time zone transition (e.g. when the daylight saving time ends or the time zone offset is increased) or skipped local time at a negative time zone transitions (e.g. when the daylight saving time starts or the time zone offset is decreased),
```

And, accordingly 'subtracted from' and 'added to' have to be swapped, which is fine. 

Hmm... having written it down, it does not seem that bad except that 'DST ends ==> tz offset increase' and 'DST starts ==> tz offset decrease' seems still a bit odd to me.  

Anyway, let's figure out which is the best for this term given all things I mentioned so far.

spec.html
jungshik: Thanks. I fixed in my 3rd commit.

mj1856:
WRT sign (+/-):  Yes, the `getTimezoneOffset` function is opposite of what most people expect, but that's only because we are used to working with the ISO8601 and standard and its derivatives, which place positive offsets *East* of GMT.

Quick history lesson: The `getTimezoneOffset` function, like much of the `Date` object, was copied from `java.util.Date` from Java 1.0 - which follows the POSIX standard with regard to time zones and their offsets.  If you've seen a POSIX time zone string, like `PST8PDT` or `CET-1CEST,M3.5.0/2,M10.5.0/3` - the first number (`8` or `-1` in these examples) is the standard-time offset, in hours *West* of GMT.

Anyway that absolutely cannot be changed in the existing API.  It would break all kinds of things.

mj1856:
@ediosyncratic - Thanks for the feedback - That's a good scenario!  I agree that in this case, one would want the 1:30 before the spring-forward transition, and they would also want the second 2:30 in the fall-back transition.  However - we are not landing on these times by means of addressing a local time value without offset.  We're landing on them through timeline math, as we do the subtraction operation.  We know exactly where we're starting on the timeline, and we know the rules for the time zone, so we know exactly where we'll end up.

Using `Europe/Oslo` as an example.  Let's say the flight leaves at 4:30 AM, and we have to depart two hours early to get their on time.  In the spring, the local time from 2 AM to 3 AM is skipped.

```
2017-03-26T04:30:00+02:00  // our flight time
2017-03-26T02:30:00Z       // equivalent UTC
2017-03-26T00:30:00Z       // subtract two hours
2017-03-26T01:30:00+01:00  // equivalent local time
```

We can demonstrate this directly with the `Date` object, if our computer's time zone is set for Norway.  This works in both Chrome and FireFox - which we've established are currently doing opposite behaviors:

```js
var d = new Date(2017, 2, 26, 4, 30);
d.setUTCHours(d.getUTCHours() - 2);
console.log(d.toString()); // "Sun Mar 26 2017 01:30:00 GMT+0100 (Central Europe Standard Time)"
```

Same in the fall, when the local time from 2 AM to 3 AM is repeated.
```
2017-10-29T04:30:00+01:00  // our flight time
2017-10-29T03:30:00Z       // equivalent UTC
2017-10-29T01:30:00Z       // subtract two hours
2017-10-29T02:30:00+01:00  // equivalent local time
```
We land at 2:30 CET - the second occurrence.  Again, to demonstrate:

```js
var d = new Date(2017, 9, 29, 4, 30);
d.setUTCHours(d.getUTCHours() - 2);
console.log(d.toString()); // "Sun Oct 29 2017 02:30:00 GMT+0100 (Central Europe Standard Time)"
```

Of course the mistake that is commonly made is that users use `getHours` and `setHours` instead of the `getUTCHours` and `setUTCHours` functions.  The latter are absolutely required when moving on the timeline by a duration (aka "time math" or "timeline math").  Users don't get this easily, which is one reason why we have libraries like [Moment.js](http://moment.js), and why so many folks (including @maggiepint and myself) think JavaScript needs a new datetime API.

So - long story - but there is no problem with the scenario you describe.  It doesn't invoke the condition where we we have to decide which way to adjust.   That only happens when a user gives us a *input* in terms of local time, and that input is ambiguous or invalid.  Then we go down the path with the arguments I made earlier.

Cheers. 😄

spec.html
```diff
@@ -25936,44 +25936,43 @@
...
         <emu-note>
           <p>It is recommended that implementations use the time zone information of the IANA Time Zone Database <a href="http://www.iana.org/time-zones/">http://www.iana.org/time-zones/</a>.</p>
+          <p>1:30 AM on November 5, 2017 in America/New_York is repeated twice (fall backward), but it must be interpreted as 1:30 AM UTC-04 instead of 1:30 AM UTC-05. LocalTZA(TimeClip(MakeDate(MakeDay(2017, 10, 5), MakeTime(1, 30, 0, 0))), false) is <emu-eqn>-4 &times; msPerHour</emu-eqn>.</p>
+          <p>2:30 AM on March 12, 2017 in America/New_York does not exist, but it must be interpreted as 2:30 AM UTC-05. LocalTZA(TimeClip(MakeDate(MakeDay(2017, 2, 12), MakeTime(2, 30, 0, 0))), false) is <emu-eqn>-5 &times; msPerHour</emu-eqn>.</p>.
```
mj1856: It may be useful to call out here that since the result `2:30 AM UTC-5` doesn't exist, that through normalization, the user will view this as `3:30 AM UTC-4`, since they both have the same equivalency `7:30 AM UTC`.

jungshik: Thank you for the review/LGTM.  I added '(equivalent to 3:30 AM UTC-04)'. Would it suffice?

spec.html
```diff
@@ -25944,7 +25944,7 @@
         <emu-note>
           <p>It is recommended that implementations use the time zone information of the IANA Time Zone Database <a href="http://www.iana.org/time-zones/">http://www.iana.org/time-zones/</a>.</p>
           <p>1:30 AM on November 5, 2017 in America/New_York is repeated twice (fall backward), but it must be interpreted as 1:30 AM UTC-04 instead of 1:30 AM UTC-05. LocalTZA(TimeClip(MakeDate(MakeDay(2017, 10, 5), MakeTime(1, 30, 0, 0))), false) is <emu-eqn>-4 &times; msPerHour</emu-eqn>.</p>
-          <p>2:30 AM on March 12, 2017 in America/New_York does not exist, but it must be interpreted as 2:30 AM UTC-05. LocalTZA(TimeClip(MakeDate(MakeDay(2017, 2, 12), MakeTime(2, 30, 0, 0))), false) is <emu-eqn>-5 &times; msPerHour</emu-eqn>.</p>.
+          <p>2:30 AM on March 12, 2017 in America/New_York does not exist, but it must be interpreted as 2:30 AM UTC-05 (equivalent to 3:30 AM UTC-04). LocalTZA(TimeClip(MakeDate(MakeDay(2017, 2, 12), MakeTime(2, 30, 0, 0))), false) is <emu-eqn>-5 &times; msPerHour</emu-eqn>.</p>.
```
mj1856: Looks good.  Thanks.

jungshik: Thanks a lot for the review and approval !

jungshik: I wonder what comes next.

jungshik:
I rebased the PR to the trunk to resolve a conflict. 
@bterlson,  what'd be next step for this PR?

bterlson:
We'll have to discuss in committee. Any volunteers to bring it up? Otherwise I can.

maggiepint:
I'll do it if that would be easier for you, and I'm actually there.

jungshik:
Thank you, @bterlson  and @maggiepint  !

spec.html
```diff
@@ -26211,44 +26211,43 @@
...
+        <p>When _isUTC_ is true, <dfn>LocalTZA( _t_, true )</dfn> is the offset of the local time zone from UTC measured in milliseconds at time represented by time value _t_ (UTC). When it is added to _t_ (UTC), it yields the local time.</p>
+        <p>When _isUTC_ is false, <dfn>LocalTZA( _t_, false )</dfn> is the offset of the local time zone from UTC measured in milliseconds at local time represented by time value <dfn><emu-eqn>_t_<sub>local</sub> = _t_</emu-eqn></dfn>. When it is subtracted from the local time <emu-eqn>_t_<sub>local</sub>, it yields the corresponding UTC.</p>
+        <p>When <emu-eqn>_t_<sub>local</sub></emu-eqn> represents local time repeating multiple times at a negative time zone transition (e.g. when the daylight saving time ends or the time zone adjustment is decreased due to a time zone rule change) or skipped local time at a positive time zone transitions (e.g. when the daylight saving time starts or the time zone adjustment is increased due to a time zone rule change), <emu-eqn>_t_<sub>local</sub></emu-eqn> must be interpreted with the time zone adjustment before the transition.</p>
```
littledan: To clarify, for an implementation based on ICU, it sounds like this paragraph could be implemented by calling `icu::Calendar::setRepeatedWallTimeOption(UCAL_WALLTIME_FIRST)` and `icu::Calendar::setSkippedWallTimeOption(UCAL_WALLTIME_FIRST)`; is that right?

littledan:
I'm not an expert here, but I looked at this patch, and it seems good to me. The only editorial suggestion I'd make is to take a token like `~utc~` vs `~local~` rather than a boolean, to make it easier to read, similar to the *hint* argument for ToPrimitive.

Please correct me if I'm wrong, but if we're dealing with historical date/time changes better, it seems like there are a few issues that this proposal still leaves out:
- The Julian -> Gregorian calendar change (Date extends the Gregorian calendar infinitely into the past)
- Leap seconds (all days are the exact same length in Date)

Will it be possible to address either of these with needs-consensus PRs in a similar manner, or will those have to be left for the future "Date V2"?

maggiepint:
Dan,

I think you're going to want to leave those to Date V2. Given the fact that
the current Date object centers around Unix time, there would not be great
ways to handle leap seconds.

The Julian-> gregorian exchange might be handleable, but I'm not sure if
enough users are clamoring for it to make it worth trying to jam it into
the current object.

littledan:
I was wondering about Julian -> Gregorian more out of laziness plus pedantry... Looks like ICU handles this transition correctly by default (while sharing the mistake of being based on Unix time), and if we were to just call out to it that way, it would currently be a spec violation.

spec.html
```diff
@@ -26296,7 +26295,7 @@
         </emu-alg>
       </emu-clause>
 
-      <!-- es6num="20.3.1.14" -->
+      <!-- es6num="20.3.1.13" -->
```
littledan: These numbers should not be adjusted. They are historical backreferences to a previous version of the specification.

spec.html
jungshik: However, we decided, earlier in the thread, not to add such an option. :-)

spec.html
jungshik: What should I do with section numbers? Are they never to be changed?

jungshik:
> Looks like ICU handles this transition correctly by default (while sharing the mistake of being based on Unix time), and if we were to just call out to it that way, it would currently be a spec violation.

By default, ICU's GregorianCalendar has the transition date set to 1582-10-15, but it can be changed back to the earliest possible time. So, I don't see an issue unless EcmaScript's earliest time is earlier than ICU's earliest time.  Perhaps, you meant the same thing as I'm saying here. 

http://icu-project.org/apiref/icu4c/classicu_1_1GregorianCalendar.html#af7ebc50c1164d6ed96c96466242eb9ea

spec.html
jungshik: And, earlier versions of PR description had references to those two ICU methods (as a future extension possibility), but I dropped the references later because that's deemed a not-a-good idea.

jungshik: Sorry that I misunderstood what @littledan meant. You're right that the choice made in this PR can be implemented with UCAL_WALLTIME_FIRST at fall backward. 

However, at spring forward, UCAL_WALLTIME_LAST should be used. When DST starts in America/New_York, 2:30 AM does not exist, but has to be interpreted as 2:30 AM UTC-0500 (= 3:30 Am UTC-0400). UCAL_WALLTIME_LAST has to be passed to setSkippedWallTimeOption() to get that behavior.  (see https://goo.gl/h0bP26 )

spec.html
littledan: They should not be changed. They should also not be added to new sections.

littledan:
@jungshik Thanks, I had missed that part of the documentation. It seems like a fine solution to me.

mj1856:
Keep in mind that there's a drastic difference between _calendar_ APIs and _date_ APIs.  Pretty much *all* date APIs use the [*Proleptic* Gregorian Calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar).  This is a also fundamental requirement of ISO 8601 and its derivatives.

So no, the `Date` API should not implement the Julian to Gregorian cut-over, nor would a successor API if it is meant to align with ISO8601.  Only when you get into calendar systems would you take this into account.  Then, of course, we'd also pick up other calendars like Hijiri, Thai, Hebrew, etc.

For example, in legacy Java, `java.util.Date` (which JS's `Date` was based on) does *not* model the cutover, but `java.util.Calendar` *does*.  Same with `NSDate` vs `NSCalendar` in Apple land.

WRT leap seconds, though ISO8601 does support them - almost no date/time APIs do, because almost no computer software or operating systems represent leap seconds directly.  IMHO, it would be a mistake to introduce them into JavaScript, without thinking very hard about the consequences of interoperability with other APIs.

littledan:
Thanks for explaining, @mj1856

littledan:
This PR got consensus at the March TC39 meeting, modulo

littledan:
Sorry, accidentally clicked close. I meant:

This PR got consensus at the March TC39 meeting, modulo reconsidering the following wording:

> An implementation of ECMAScript is expected to determine the local time zone adjustment with respect to UTC using the best available information on time zones.

We should make sure this is broad enough to cover embedded systems' requirements.

littledan:
I see the 'needs tests' flag on this patch. It seems a little difficult to me to write tests here--you have to set the local timezone to observe the results. This would take some not-completely-trivial harness work, and might not be possible on all OS + engine combinations. (FWIW I [did this](https://codereview.chromium.org/2724373002/) in V8 and ended up disabling the tests on Windows, but maybe there's a good way around that...)

jungshik:
Sorry that I haven't paid attention to this PR for a while. 

> An implementation of ECMAScript is expected to determine the local time zone adjustment with respect to UTC using the best available information on time zones.

littledan@ : do you have any suggestion to reword the above?  

As for testing, I agree with you that it's rather difficult to write tests.

leobalter:
In this case the spec is implementation dependent (the local time zone
adjustment) which test262 might not be able to set itself.

If complexity beats up at a very high bar, we might skip the tests for this
with this rationale.

I didn't have the chance to verify the spec changes, I'm assuming
@littledan has a valid statement I'll agree when I verify it.

littledan:
@leobalter It would take harness work to be able to set the timezone, but it should be possible somehow. It's not possible with the current test harness setup alone, I believe.

I'm happy with this spec text. Some at the March meeting expressed concern about wording that​ I mentioned above.

jungshik:
@littledan ,  do you plan to do necessary harness work? 

At least on Linux, using the environment variable to set the timezone seems to work. 

Pacific/Apia switched from UTC-10 to UTC+14 at the end of 2011-12-29 localtime. 
As a result, 2011-12-30 is skipped entirely, but any time in 2011-12-30 has to be interpreted with the timezone offset before the switch (UTC-10). 

```
TZ="Pacific/Apia" out/rel/d8 --icu-timezone-data
V8 version 6.0.0 (candidate)
d8> (new Date(2011,11,30,11,0)).toUTCString();   <== skipped time but should be interpreted as UTC-10
"Thu, 29 Dec 2011 21:00:00 GMT"
d8> (new Date(2011,11,31,1,0)).toUTCString();   <== UTC+14 now
"Fri, 30 Dec 2011 11:00:00 GMT"
```

jungshik:
Ick..  I thought the first output was '30 Dec 2011 21:00 GMT.  v8 was not yet changed to implement this spec change.  Sorry for the confusion.

Anyway, at least the above confirmed that TZ environment variable works on Linux for v8.

spec.html
jungshik: Just to make sure I understand you correctly before addressing this concern: 


20.3.1.8 was about DSTA which is being dropped.  And all other section number changes are due to 20.3.1.8 being dropped. 

Instead, I should just skip 20.3.1.8 and keep the old section numbers (20.3.1.9 and onward), shouldn't I?

jungshik:
> I'm happy with this spec text. Some at the March meeting expressed concern about wording that​ I mentioned above.

@leobalter : why does this PR still have 'needconsensus' label?   It appears that March meeting approved the change. 

In the meantime, #848 was landed before this one, which requires changing this PR because #848 refers to LocalTZA without any param that was removed by this PR. I'll update this PR. 

@littledan: as for testing, do you plan to make a necessary test-harness? Or, could we go first without test for now?

bterlson:
@Jungshik my understanding is there isn't clarity around @littledan's point above? But maybe I missed it...

Specifically the wording:

> An implementation of ECMAScript is expected to determine the local time zone adjustment with respect to UTC using the best available information on time zones

maggiepint:
Agree with Brian. The main concern is whether IoT devices will be compelled to carry all of the TZDB data with this change. I believe that some people felt that we could compel it in the spec, and just let IoT implementations be non-compliant, while others felt we should allow for the IoT concern specifically.

The testability concern is orthogonal to our ability to approve the pull.

apaprocki:
Theoretically IoT devices only need to contain the data for their configured timezone (with the current spec). This allows for a setup/configuration program to download just the binary file it needs (on the order of 1-4k) and not store all the data. If an IoT device is built with timezone support in the first place, it presumably will require a way to update/patch the timezone data anyway.

Worth mentioning that the topic of a smaller dataset was brought up on the `tz` list last month:
http://mm.icann.org/pipermail/tz/2017-May/025042.html

jungshik:
> Good call on the breakage in this PR due to #848.

The latest commit took  care of a new provision added in #848 by switching to ```LocalTZA(tv, true)``` from ```LocalTZA + DSTA(tv)```

jungshik:
@bterlson @maggiepint @apaprocki   

Thanks a lot for the explanation. 

The current spec has the following along with a note on IANA TZ DB. 

> An implementation of ECMAScript is expected to determine the local time zone adjustment.
> ....  
> An implementation-dependent algorithm using best available information on time zones to determine the local daylight saving time adjustment DaylightSavingTA(_t_), measured in milliseconds. An implementation of ECMAScript is expected to make its best effort to determine the local daylight saving time adjustment.

> It is recommended that implementations use the time zone information of the IANA Time Zone Database <a href="https://www.iana.org/time-zones/">https://www.iana.org/time-zones/

My PR has the following along with the same note on IANA TZ DB. 

> An implementation of ECMAScript is expected to determine the local time zone adjustment with respect to UTC using the best available information on time zones.

To me, it appears that both the current spec and this PR give some leeway to implementations including those on IoT devices  because not every one of them may have the latest / the most comprehensive (or as comprehensive as necessary) TZ information. 

Anyway, which direction should we move?  Compel implementations to use the (latest) IANA TZ DB (or its equivalent) or keep the current ambiguity?  Or, should it explicitly mention IoT/embedded devices?

littledan:
@jungshik and @apaprocki make good points that the specification already allows for forms of variability. It'd be good to have some more clarity from the folks who had the concern about whether this is sufficient. cc @BrendanEich @allenwb @waldemarhorwat (did I remember correctly? were there others).

jungshik:
Thanks, @littledan. 

Pinging @BrendanEich @allenwb @waldemarhorwat regarding their concern ( https://github.com/tc39/ecma262/pull/778#issuecomment-307239562 ).  It'd be great if you could respond to @littledan's comment on the issue. Thanks !

jungshik:
How could we move this forward?

allenwb:
"should" (or "SHOULD", but ECMA-262 has not generally followed that convention) is the preferred way for normative specification to express the concept "is expected to".

Bottom line, you are saying that the result is "implementation defined" within some constraints.  The best way to express such constraints is via "must" and "should" and "may" requirements.  The meaning of  "using the best available information on time zones" is unclear (as can be seen by the upthread discussion).  Does it mean that "an implementation MUST use a table of all known DST transitions which SHOULD be regularly updated to reflect political DST changes" or does it mean "an implementation SHOULD use a regularly updated table of all known political DST transitions but MAY use an incomplete or outdated table or no table at all".  If the latter, should there be an additional requirement such as "If no political DST transition information is available, an implementation MUST ...".   In other words, say what you want implementations to do in the case where  it doesn't have DST info.

littledan:
@allenwb Do you have suggested wording for this? The wording in this patch for LocalTimeZoneAdjustment is pretty similar to the wording in Daylight Saving Time Adjustment . Should they both be changed to use "should" somehow, such as this?

>  implementation of ECMAScript <del>is expected to</del><ins>should</ins> make its best effort to determine the local daylight saving time adjustment.

Do you have any ideas how to clarify what implementations should do if they don't have appropriate data?

allenwb:
Let's see how we might improve Daylight Savings Time Adjustment

 Daylight Savings Time Adjustment currently says:

>An implementation-dependent algorithm using best available information on time zones to determine the local daylight saving time adjustment DaylightSavingTA(t), measured in milliseconds. An implementation of ECMAScript is expected to make its best effort to determine the local daylight saving time adjustment.

here is how I would restate it:

>  DaylightSavingTA(t) is an implementation-defined algorithm that must return a number representing milliseconds that is suitable for adding to a Time Value. _t_ is an UTC Time Value. The result must represent a number of milliseconds (possibly negative) that is suitable for adding to a Time Value.  The result should be the number of milliseconds that converts a local Time Value from standard time to daylight saving time. The local political rules for DST in effect at _t_ should be used to determine the result. If an implementation does not support such a conversion or if political rules for time _t_ are not available within the implementation the result must be 0.

>NOTE It is recommended that implementations use the time zone information of the IANA Time Zone Database https://www.iana.org/time-zones/.

Even this has some problems, such as the fact that the term "local time" is not currently defined anywhere in ECMA-262.

Some of the awkwardness and fuzziness in the current Date spec. comes from past attempts to "improve things" while making the minimum possible changes to the existing spec structure and language. Really improving things should probably requires a comprehensive rewrite [20.3.1](https://tc39.github.io/ecma262/#sec-overview-of-date-objects-and-definitions-of-abstract-operations)

littledan:
That wording seems to be good to me, except for one nit: maybe we should remove the word 'political'. The construction of time zones itself is already political--I am not sure that current DST rules are "more" political. And lots of other things in our specs are political as well, such as the construction of locales. Maybe we should just delete the world political, as it seems to imply that other things are not political.

Anyway, it seems like the rewording you are suggesting affects existing text and should be done in a separate patch.

jungshik:
Thanks, @allenwb. 

@littledan   Because this PR drops DSTA(t), I can morph the rewording suggested by @allenwb for DSTA(t) to LocalTZA(t, is_utc) and update this PR.  That way, we can save one intermediate PR (rewording DSTA(t)). 

As for 'political', I agree that we'd better not use it.

allenwb:
I used the word "political" because I think I've heard chrono experts use it or something similar.  Perhaps "civil" would be an alternative.  My understanding is that choice of DST parameter is much more varied and politically volatile then the time zone boundaries.  

I'm more concerned about the lack of formal definition of "local" and "local time".  What does "local [political] rules for DST" actually mean.

mj1856:
I agree with the wording of "political".  Time zones are indeed *more political* than locale issues, which I would categorize as *cultural*.

It wouldn't be accurate to say that DST is more varied than any other aspect.  That may be true in some parts of the world, but at a global level, everything about time zones is politics.  Including DST, standard offset, boundaries, officialness, and observation.

The politics themselves occur at all levels, including national, state, county, and city.  They include legislative, executive, and judicial actions.  Sometimes different branches create conflicting actions and chaos ensues.  And to top it all off, sometimes they give adequate details and enough time to implement the changes, and sometimes they do not.

Lots of examples, too many to list, but here are a few:

- https://en.wikipedia.org/wiki/Time_in_Indiana
- http://boston.cbslocal.com/2017/04/28/maine-time-zone-massachusetts-new-hampshire-daylight-saving-time-change/
- http://www.reuters.com/article/us-venezuela-timezone-idUSKCN0XC2BE
- http://www.bbc.com/news/world-europe-34631326
- https://www.timeanddate.com/news/time/ukraine-crimea-adopts-moscow-time.html

From my blog:
- http://codeofmatt.com/2016/04/23/on-the-timing-of-time-zone-changes/
- http://codeofmatt.com/2016/07/01/time-zone-chaos-inevitable-in-egypt/

Yes, I think there's no more appropriate term than "political".

frutiger:
Does delineating locale as less political and DST as more political assist implementers or users of ECMA-262?

littledan:
Completely agree that timezones are politically determined. If this is the common way to phrase these things among specialists, sounds fine to include it.

&lt;small>(I'd maintain that the construction of things like national languages is extremely political and relating to the state and localities, though. It's definitely been a chaotic process. Unlike timezones, the politics/oppression here makes things easier, not harder, for localizers, since you don't have to worry about groups that don't have the right to speak their own language!)&lt;/small>

allenwb:
I think primary reason is using the word "political" in the context I suggested is to emphasize that these are non-technical issues (i.e., not defined by a technical standard) and hence implementors are warned that they may need to deal with them as such.

Also, the intent is not to imply that there is anything negative about "political".   It's just referring to a domain that is not part of the technical standards world.

frutiger:
To support @allenwb and @mj1856, I will note that at least the Python documentation does refer to "political" when referring to DST: https://docs.python.org/3/library/datetime.html?highlight=politic

ediosyncratic:
I would distinguish the **civil** rules for DST (that are part of civil law) from the _political_ process that leads to them being the (messed-up) way they are.  (There is some incidental comedy to the fact that the choice of how to describe this **is political** in its own right.)

I cannot speak for the comparative politicality of time-zone messes compared to locale messes, but it likely **is** useful to implementors to make them aware that time-zone information is apt to change at the whim of politicians; in particular, it may do so between the setting up of (say) a calendar notification and the time it comes due - so that (in a long-running application) a timer set to notify at a certain UTC time determined when the timer was set may fire at "the wrong time" if the time-zone the user was referring to when stipulating the time has been messed with in between.  Not such a huge issue for most web pages, I grant, but I've implemented things in a web-server (with a tendency to uptimes measured in years) that had to take account of that.  More importantly, for a calendar app: it is necessary to save event times in a way that represents the user's intent, as it was stated, rather than immediately mapping everything to UTC (as I **oh so wish** we could always do).  That can, in principle, involve remembering whether the user meant the first or second pass through a fall-back's repeated hour (although these are usually so timed as to make few calendar events land in them).

Of course, locale details may change at the whim of politicians (or, more usually, learned societies; for all that these, too, have their politics), which may make a localized form of something out-of-date; but it'll usually still be intelligible to denizens of the relevant locale.  So software can be out-of-date about locale without causing practical problems of note.  In contrast, if software has out-of-date time-zone information, it can lead to the user missing appointments (e.g. flights), because they turned up an hour late.
So, while locale may be just as political as time, there is a greater urgency to being up-to-date about time zones, as compared to locales; the damage potential of political whims is greater with time-zone data.

jungshik:
Sorry that I haven't followed up with this after saying that I would. I'll adjust wording per @allenwb and update the PR. Then, I agree with @littledan that merging the updated PR  can go ahead while looking for a way to add a test in parallel (instead of blocking merge.)

jungshik:
Please, take one more look and merge now that I updated the PR to reflect [@allenwb's comment](https://github.com/tc39/ecma262/pull/778#issuecomment-322935581).  Thank you !

jungshik:
What should/can I do to get this merged?

littledan:
@bterlson does this seem good to merge to you?

bterlson:
I'm happy to merge this now, but is there still no tests? I'm worried we may never see them :)

bterlson:
I also note some must/should confusion in LocalTimeZoneAdjustment now that I'm reading this again. Can we remove should from its specification text via "LocalTZA should return" --> "LocalTZA returns"? I can just make this change myself in order to get this in asap :)

littledan:
@allenwb explained the rationale for "should" in [this comment](https://github.com/tc39/ecma262/pull/778#issuecomment-322819606). I explained why it would be hard to write tests in [this comment](https://github.com/tc39/ecma262/pull/778#issuecomment-339813022),

bterlson:
@littledan must vs should confusion referred to discrepancies between the note saying must and the normative text saying should, but I see now that the note is only talking about the final must requirement not the previous should requirements. It's good, in other words, thanks :)

jungshik:
Thanks everybody !

fabriciomurta:
Issues are raising as it became released on chrome 67. It is changing seconds in the time zones and does not get carried when converted to-and-back UTC.
https://stackoverflow.com/questions/50857187/weird-seconds-offset-in-js-date-object-in-chrome

jungshik:
> does not get carried when converted to-and-back UTC

Not sure what you meant by that. Chrome 67 takes into account the historical time zone offset changes over time as recorded in the IANA time zone database.  As explained by one of answers with heading 'Not a bug', some time zones underwent offset/DST changes multiple times even after 2010 (e.g. Europe/Moscow).   In case of Sao Paolo, see https://www.timeanddate.com/time/zone/brazil/sao-paulo  (select 1900-1924). It switched to UTC-3 from LMT in 1914.

fabriciomurta:
Yes, but if you read around the discussions you'll see the complaint is not really about the -3h or -3h06m (or other in other time zones), it is about a -3h06m28s or 54s.

This is the most likely change in chrome that has introduced the above issue. Converting to-from UTC also results in a different result, e.g. the time converted back from UTC won't match the base time in a local1->UTC->local2 conversion chain. `local2` won't match `local1` by this seconds difference.

jungshik:
> local1->UTC->local2 conversion chain. local2 won't match local1 by this seconds difference

I'm not sure what you meant by the above: 
( please, ignore 'Pacific Standard Time' inside parens. That's not correct and I'm aware of that). 

```
> d2=new Date(1850,0,1)
Tue Jan 01 1850 00:00:00 GMT-0752 (Pacific Standard Time)
> d3=new Date(Date.UTC(d2.getUTCFullYear(), d2.getUTCMonth(), d2.getUTCDate(), d2.getUTCHours(), d2.getUTCMinutes(), d2.getUTCSeconds(), d2.getUTCMilliseconds()))
Tue Jan 01 1850 00:00:00 GMT-0752 (Pacific Standard Time)

> d2.getTime() == d3.getTime()
true
```

catrina4200:
> Currently, LocalTimezoneAdjustment is assumed to be constant across time
> zone rule changes and does not take any argument. Daylight Savings Time
> Adjustment does take an argument t.
> 
> DSTA(t) can 'absorb' the standard time zone offset changes
> (e.g. Europe/Moscow changed the standard time zone offset multiple times
> in the last decade.). However, the spec stipulates that it should only
> reprsent an additional adjustment necessary when DST is in effect.
> 
> In practice, at least two implementations (Spidermonkey/Firefox and
> JSC/Safari) do not follow the spec to the letter to produce results
> expected by users across time zone rule changes.
> 
> This PR revises LocalTZA to accept two arguments, 't' and 'isUTC' and
> drop DaylightSavingsTimeAdjustment(t). When isUTC is true, LocalTZA
> interprets 't' as a time value representing UTC and gives a time zone
> offset in effect at time t for the local time zone (for US ET, it'd be
> `-4*msPerHour` in summer and `-5*msPerHour` in winter). When isUTC is false,
> 't' is interpreted as local time value and gives a time zone offset in
> effect at t = t_local.
> 
> It's also specified that LocalTZA(t = t_local, false) will return the offset
> _before_ the transition when t_local is local time repeating multiple
> times at a negative transition (e.g. fall backward) or skipped local time
> at a positive time zone transition (e.g. spring forward).
> 
> Currently, implementations do not agree with each other as to how to handle repeating
> or skipped local time. This change will standardize the behavior in line with the way Java 8 does.
> 
> Finally, UTC(t) and Localtime(t) are reformulated with LocalTZA(t = t_local, false) and
> LocalTZA(t = t_UTC, true).
> 
> Will fix #725.
1648243: [owners] Make test/ owned by COMMON_OWNERS | https://chromium-review.googlesource.com/c/v8/v8/+/1648243

pschlan:
Hi,

is the following behavior, which seems to result from this spec change, expected?

```
const d1 = new Date(1635642000000);
d1.toString()
// "Sun Oct 31 2021 02:00:00 GMT+0100 (Mitteleuropäische Normalzeit)"

d1.getTime()
// 1635642000000

d1.setMinutes(0)

d1.getTime()
// 1635638400000
```

As a user of the Date API, this is not really what I would have expected, also after carefully reading the docs.

(Sorry if this is the wrong place to ask this question.)

ediosyncratic:
First: see the intro to [0] for possibly better fora in which to make this enquiry.
[0] https://www.ecma-international.org/ecma-262/11.0/index.html#title

Just to be clear (I guess you already worked this out):
You're bang on next year's CET fall-back, where 2:59 GMT+0200 has just been followed by 2:00 GMT+0100, the second 02:00 in local time that morning; the `setMinutes(0)` call has knocked you back to the first, 2:00 GMT+0200.

The `setMinutes()` spec [1] says to convert to local time, break down in terms of date, hours, minutes and seconds, build a new local time representing the same time but with the minutes (and optionally seconds, millis) revised, then convert the resulting local time back to UTC; in the present case, the reconstructed local time is surely the same local time it started with, so the whole trip amounts to just: let `v = ? thisTimeValue(this value))`, then `w = v +LocalTZA(v, true)` and `u = w -LocalTZA(w, false)` and use u in place of v as this object's time value hereafter.
[1] https://www.ecma-international.org/ecma-262/11.0/index.html#sec-date.prototype.setminutes

As the notes on `LocalTZA()` and `UTC()` point out, that round-trip isn't guaranteed to be idempotent and, indeed, *cannot* be made so, since `t +LocalTZA(t, true)` varies with t in a way that skips some values and repeats others.

I can understand why a user of this API would expect `setMinutes(getMinutes())` to be a no-op.
However: consider a user in Afghanistan, with a time-zone offset of 4:30; if they take 12:00 GMT+0430 and apply `setMinutes(0)` to it, surely they expect to get back 12:00, not 12:30 or 11:30; so `setMinutes()` has to convert to local time, set the minutes on a local time, then convert back to UTC.  That, unfortunately, means that users in whole-hour-offset zones get a surprise when their DST changes; but given a choice between always wrong behaviour, for everyone in zones with offsets that aren't  whole multiples of an hour, and very rarely wrong (one thousandth of a year is more than eight hours: zone transitions take up less of the year than this, even in the zones (Morocco, Egypt) whose periods of DST-except-during-Ramadan gave them four transitions per year, or in the one zone with a two-hour offset) for those whose governments inflict DST on them, I can see why the standard works the way it does.

Not that the situation can't be salvaged: `setMinutes(m)` could be specified to determine local time's minutes, `n`, then add `m -n` minutes' worth of milliseconds if `m > n`, else `60 +m -n` minutes' worth, to the UTC time (without converting back and forth); this will reliably (unless it crosses a time-zone transition by an amount that's *not* a multiple of an hour - these are very rare) produce a time with the desired minutes.  I trust it's obvious how to generalise this to deal with other time units, likewise.
The error case of this algorithm is much rarer than that of the present round-trip via local time (which would quite likely also produce wrong results, from a user perspective, in the event of a not-whole-hour transition close to the time being adjusted).


Incidental note (unrelated to this issue, noticed while reading the spec in hopes of getting the above right): the `TimeClip()` function's Note refers to a Step 4 - "The point of Step 4 is to ..." - of a 3-step process: there *is no* step 4 !
https://www.ecma-international.org/ecma-262/11.0/index.html#sec-timeclip

anba:
> Incidental note (unrelated to this issue, noticed while reading the spec in hopes of getting the above right): the `TimeClip()` function's Note refers to a Step 4 - "The point of Step 4 is to ..." - of a 3-step process: there _is no_ step 4 !
> https://www.ecma-international.org/ecma-262/11.0/index.html#sec-timeclip

@ljharb :arrow_up: fallout from #1827

jmdyck:
> the `TimeClip()` function's Note refers to a Step 4 - "The point of Step 4 is to ..." - of a 3-step process: there _is no_ step 4 !

That Note was removed by the merge of PR #1965.

ljharb:
The latest version of the standard, for reference: https://tc39.es/ecma262/#sec-timeclip

PR #783, opened 2017-01-25
fixes #708: removing ResolveExport's exportStarSet in favor of allowing circular valid dependencies

### LINKED ISSUE #708: ResolveExport's exportStarSet argument disallows cycles that don't seem problematic
The `exportStarSet` passed to `ResolveExport` is used to cause an error in cases like the following:

A: import {x} from "B";
B: export \* from "C";
C: export var y; export {y as x} from "B";

The sequence of ResolveExport calls here is:
1. B.ResolveExport("x", &laquo;&raquo;, &laquo;&raquo;)
2. C.ResolveExport("x", &laquo;{B, "x"}&raquo;, &laquo;B&raquo;)
3. B.ResolveExport("y", &laquo;{B, "x"}, {C, "x"}&raquo;, &laquo;B&raquo;)

where step (3) fails when hitting step 7 of ResolveExport (since B is already in the exportStarSet).

But that would have been perfectly legal without the `export *`:

A: import {x} from "B";
B: export {x, y} from "C";
C: export var y; export {y as x} from "B";

where the sequence is instead:
1. B.ResolveExport("x", &laquo;&raquo;, &laquo;&raquo;)
2. C.ResolveExport("x", &laquo;{B, "x"}&raquo;, &laquo;&raquo;)
3. B.ResolveExport("y", &laquo;{B, "x"}, {C, "x"}&raquo;, &laquo;&raquo;)
4. C.ResolveExport("y", &laquo;{B, "x"}, {C, "x"}, {B, "y"}&raquo;, &laquo;&raquo;)

and step (4) results in a a valid resolution.

What's the rationale for the former being an error?

@GeorgNeis who pointed out this oddity, and @allenwb for any history here.

caridy:
## Intuitions

The meaning of `export * from <module-specifier>` is to signal that additionally to the local exports, any additional export present in that `<module-specifier>` should also be exported. Which means that the following graph should be valid:

```js
// main.js: entry point
export * from "A";

// A.js
export * from "B";

// B.js
export * from "A";
```

Developers can start with some boilerplates, and then fill the blanks on each module in that graph, and this progression is sounded. E.g.: Add `export const a = 1;` to `B.js`, and the graph should remain valid, but now with one exported name for Main, A, and B.

## Spec

In the spec, there are two mechanisms to deal with circularity:

* `ResolveExport( exportName, resolveSet, exportStarSet )`
* `GetExportedNames( exportStarSet )`

`GetExportedNames` works fine with the scenario above when attempting to resolve the names used to create the namespace object (for the example above it will return an empty List of names). 

`ResolveExport` is responsible to resolved a specific export, including the possibility of an ambiguity, but should not care about the scenario above.

## Proposal

Remove `exportStarSet` from the `ResolveExport()` is sufficient to allow the scenario.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -8154,7 +8154,7 @@
           1. Let _exports_ be _O_.[[Exports]].
           1. If _P_ is not an element of _exports_, return *undefined*.
           1. Let _m_ be _O_.[[Module]].
-          1. Let _binding_ be ! _m_.ResolveExport(_P_, &laquo; &raquo;, &laquo; &raquo;).
+          1. Let _binding_ be ! _m_.ResolveExport(_P_, &laquo; &raquo;).
           1. Assert: _binding_ is neither *null* nor `"ambiguous"`.
           1. Let _targetModule_ be _binding_.[[Module]].
           1. Assert: _targetModule_ is not *undefined*.
@@ -21479,7 +21479,7 @@
               1. Let _requiredModule_ be ? HostResolveImportedModule(_module_, _required_).
               1. Perform ? _requiredModule_.ModuleDeclarationInstantiation().
             1. For each ExportEntry Record _e_ in _module_.[[IndirectExportEntries]], do
-              1. Let _resolution_ be ? _module_.ResolveExport(_e_.[[ExportName]], &laquo; &raquo;, &laquo; &raquo;).
+              1. Let _resolution_ be ? _module_.ResolveExport(_e_.[[ExportName]], &laquo; &raquo;).
               1. If _resolution_ is *null* or _resolution_ is `"ambiguous"`, throw a *SyntaxError* exception.
             1. Assert: All named exports from _module_ are resolvable.
             1. Let _envRec_ be _env_'s EnvironmentRecord.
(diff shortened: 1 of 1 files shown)


ajklein:
also @GeorgNeis who reported this originally

GeorgNeis:
You also need to update the uses of ResolveExport in ModuleDeclarationInstantiation, GetModuleNamespace, and module namespace [[Get]]. Looks good to me apart from that. Thanks!

caridy:
@GeorgNeis good catch. I have fixed those as well now.

bterlson:
This seems like a good change to me. Other comments from module champions? (@dherman/@wycats)?

caridy:
@bterlson I have discussed this diff with @dherman last week, he was the one suggesting this solution.

dherman:
I'm on board for the intuition. I'm trying to figure out whether the spec looks right and having trouble convincing myself. There are two things we should be confident of:

1. A cycle cannot lead to an infinite loop in ResolveExport.
2. No exports should be left out in a cycle (i.e. the ordering of the algorithm shouldn't matter).

Does that make sense? Can someone help explain the (informal) proof?

caridy:
> 1. A cycle cannot lead to an infinite loop in ResolveExport.

Invocations to ResolveExport requires the first argument `exportName` to be resolved for a module record, and this key, value pair is what we add to the `resolveSet` on every invocation. If there is cycles, the second time the same pair of module record and `exportName` matches a previous pair, it will hit step `2.a.ii`, and will break the loop by returning `null` on the resolution.

> 2. No exports should be left out in a cycle (i.e. the ordering of the algorithm shouldn't matter).

Can you elaborate more about this @dherman?

caridy:
Notes from the meeting yesterday:

* 1, and 2 were verified
* "transitive invariant for `export * from "x"`: export everything from x that hasn't been exported yet by this module record" (paraphrasing)
* `getExportedName()` was also verified.
* @dherman is ok with the changes.

@dherman can you approve this diff?

dherman:
I talked with @caridy and I'm up to speed. Summary:

* In an `export *` cycle, GetExportedNames _already_ produces the union of all names in the cycle for every module in the cycle.
* So ResolveExport was just unnecessarily failing to resolve exports that were identified as exports by GetExportedNames. Clearly a bug.
* The answer to my first question is that ResolveExport is already protected from infinite looping by the _resolveSet_.
* The answer to my second question is that GetExportedNames didn't change at all in this PR; it already produces the complete union of all names, and it already protects itself from infinite looping via the _exportStarSet_.

Summary of summary: this is a bugfix and :+1:.

bterlson:
Thank you @dherman and @Caridy.

leobalter:
tests are merged.

PR #784, opened 2017-01-25
SAB: spell out candidate execution initialization for an agent cluster

syg:
This is a normative change that spells algorithmically an initialization step for agent clusters that's been stated informally. I'd consider this a very involved bug fix.


bterlson:
After some discussion, we're going to try a smaller update to prose rather than adding an algorithm.

(Also reminder: target tc39:master!)

PR #785, opened 2017-01-25
Normative: `Array.prototype.sort`: comparefn must be fn or undefined

ljharb:
This requires implementations throw a `TypeError` when the compare function is neither undefined nor a function, which is currently vaguely implied but not explicitly required.

This reflects web reality in that every engine except v8 already throws in this case. (Safari does not, but Safari TP and webkit latest do)

eshost output of `[1, 2].sort(true)`:
```
┌─────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────┐
│ chakra              │                                                                                             │
│ chakra-es6          │ TypeError: Array.prototype.sort: argument is not a JavaScript object                        │
│ chakra-experimental │                                                                                             │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ d8                  │ 1,2                                                                                         │
│ d8 harmony          │                                                                                             │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ sm                  │                                                                                             │
│                     │ TypeError: invalid Array.prototype.sort argument                                            │
├─────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┤
│ jsc                 │                                                                                             │
│                     │ TypeError: Array.prototype.sort requires the comparsion function be a function or undefined │
└─────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────┘
```

This will likely need "web reality" and "needs consensus" labels.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -32671,6 +32671,7 @@ THH:mm:ss.sss
         <p>This function is not generic. The *this* value must be an object with a [[TypedArrayName]] internal slot.</p>
         <p>Upon entry, the following steps are performed to initialize evaluation of the `sort` function. These steps are used instead of the entry steps in <emu-xref href="#sec-array.prototype.sort"></emu-xref>:</p>
         <emu-alg>
+          1. If _comparefn_ is not *undefined* and IsCallable(_comparefn_) is *false*, throw a *TypeError* exception.
           1. Let _obj_ be the *this* value.
           1. Let _buffer_ be ? ValidateTypedArray(_obj_).
           1. Let _len_ be _obj_.[[ArrayLength]].
--- a/spec.html
+++ b/spec.html
@@ -31386,6 +31386,7 @@ THH:mm:ss.sss
         <p>The elements of this array are sorted. The sort is not necessarily stable (that is, elements that compare equal do not necessarily remain in their original order). If _comparefn_ is not *undefined*, it should be a function that accepts two arguments _x_ and _y_ and returns a negative value if _x_ &lt; _y_, zero if _x_ = _y_, or a positive value if _x_ &gt; _y_.</p>
         <p>Upon entry, the following steps are performed to initialize evaluation of the `sort` function:</p>
         <emu-alg>
+          1. If _comparefn_ is not *undefined* and IsCallable(_comparefn_) is *false*, throw a *TypeError* exception.
           1. Let _obj_ be ? ToObject(*this* value).
           1. Let _len_ be ? ToLength(? Get(_obj_, `"length"`)).
         </emu-alg>


spec.html
```diff
@@ -30780,6 +30780,7 @@
         <emu-alg>
           1. Let _obj_ be ? ToObject(*this* value).
           1. Let _len_ be ? ToLength(? Get(_obj_, `"length"`)).
+          1. If _comparefn_ is not *undefined* and IsCallable(_comparefn_) is *false*, throw a *TypeError* exception.
```
mathiasbynens: This should be step 1 instead of step 3, since the other steps aren’t needed in case we throw.

mathiasbynens: Or have you tested the order in which existing implementations (except for V8) execute these steps?

mathiasbynens: It seems SpiderMonkey checks `comparefn` first. Other engines check `ToObject(this)` first.

```
$ eshost -e 'Array.prototype.sort.call(null, null)'
#### JavaScriptCore
TypeError: Array.prototype.sort requires that |this| not be null

#### ChakraCore
TypeError: Array.prototype.sort: argument is not a JavaScript object

#### SpiderMonkey
TypeError: invalid Array.prototype.sort argument

#### V8
TypeError: Array.prototype.sort called on null or undefined

#### V8 Harmony
TypeError: Array.prototype.sort called on null or undefined
```

ljharb: Since in Firefox, `Array.prototype.sort.call(null)` throws "TypeError: can't convert null to object"  but `Array.prototype.sort.call(null, true)` throws "TypeError: invalid Array.prototype.sort argument", this matches web reality. Changing now.

mathiasbynens: ChakraCore too

ljharb: aha, thanks for the full comparison :-) I think since more engines agree on your change, that it's still the correct way to go. Updated!

claudepache: For v8, it’s quite normal that it throws on `ToObject(this)`, since it doesn’t throw on `comparefn` at all. (Still, I also prefer JSC behaviour over SpiderMonkey/Chakra one.)

ljharb: I'm ambivalent over whether the ToObject check or the IsCallable check comes first, but I think both should happen before the ToLength check.

ariya:
> escope output

Did you mean _eshost_ ?

ljharb:
@ariya thanks, fixed

bterlson:
I'm on board with this but let's wait to hear from our v8 friends.

ajklein:
This seems reasonable to me, though this seems like a fairly low-leverage tightening of `Array.prototype.sort` given what's still allowed by the spec.

ljharb:
@ajklein baby steps!

evilpie:
Btw, I just check in SpiderMonkey and we allow `[1].sort({})` so non-callable objects. Can you check how other engines behave there?

evilpie:
Turns out we optimize for arrays with length <= 1 so we never try to call the sort function in that case. We will still fail for [1, 2].sort({}). So this change is probably still fine from our side.

ljharb:
I think most engines probably choose not to call the sorting function for < 2 elements; I'll look into that in a separate PR :-)

claudepache:
Firefox seems to accept a-priori non-callable objects (but not primitives) as sort function but throws when attempting to use them. Safari TP has also an interesting behaviour:

testcase | Chrome<br>Safari 10 | Safari TP | Firefox | Edge
---------|---------|--------|-------|------
`[1].sort(_ => { throw new Error })` | `[1]` | `[1]` | `[1]` | `[1]`
`[1].sort({})` | `[1]` | `[1]` | `[1]` | TypeError
`[1].sort(2)` | `[1]` | `[1]` | TypeError | TypeError
`[1,2].sort(_ => { throw new Error })` | Error | Error | Error | Error
`[1,2].sort({})` | `[1,2]` | TypeError | TypeError | TypeError
`[1,2].sort(2)` | `[1,2]` | TypeError | TypeError | TypeError

(I vote for Edge semantics.)

ljharb:
I updated to add a commit making the same change in TypedArray.prototype.sort.

test262 tests are [here](https://github.com/tc39/test262/pull/1059)

bterlson:
Thank you @ljharb! I will pull this now that tests are pending on test262.

spec.html
leobalter: GitHub was hiding this part here. I'm also glad this is placed before the get length operation. +1 for this change.

PR #790, opened 2017-01-30
Fix step number in note 2 for RS of +

guimier:


bterlson:
Thank you @guimier, and congrats on your first ECMA262 commit!

PR #791, opened 2017-02-02
Normative: spell out initialization of agent clusters' candidate execution

syg:
Per previous discussion, use prose instead of an algorithm since the initialization is of a spec mechanism and has no meaningful, observable ordering.


spec.html
```diff
@@ -6502,6 +6502,12 @@
       <p>Examples of that type of termination are: operating systems or users terminating agents that are running in separate processes; the embedding itself terminating an agent that is running in-process with the other agents when per-agent resource accounting indicates that the agent is runaway.</p>
     </emu-note>
 
+    <p>Prior to any evaluation of any ECMAScript code by any agent in a cluster, the [[CandidateExecution]] field of the Agent Record for all agents in the cluster is set to the initial candidate execution. The initial candidate execution is an empty candidate execution whose [[EventLists]] field is a List containing, for each agent, an Agent Events Record whose [[AgentSignifier]] field is that agent's signifier and whose [[EventList]] field is an empty List.</p>
```
bterlson: I feel like ultimately this got confusing (though, I think, slightly better than the alg). It's times like these I wish we could just use TS interface definitions :)

bakkot:
(I don't think this can actually be tested, but feel free to add the 'needs tests' label if I'm mistaken.)

PR #795, opened 2017-02-06
Restrict super() in eval-scripts when in constructor methods of derived classes

anba:
Restricts `eval("super()")` to be only valid when a plain `super()` is valid.


littledan:
Thanks for this change; this behavior matches what V8 already implemented, and it seems more sensible to me.

PR #797, opened 2017-02-06
Allow duplicate function declarations in switch-statements

### LINKED ISSUE #780: Extend B.3.3.4 to cover switch-statements?
B.3.3.4 only allows `{ function f() {} function f() {} }`, but browsers also accept `switch (0) { case 0: function f() {} function f() {} }`. I think the omission to change 13.12.1 is a spec error, but before writing a PR, I'd like to get this confirmed.

anba:
Fixes #780

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38878,6 +38878,16 @@ THH:mm:ss.sss
           </li>
         </ul>
       </emu-annex>
+      <emu-annex id="sec-switch-duplicates-allowed-static-semantics">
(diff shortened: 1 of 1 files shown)


bterlson:
Ok, you've convinced me, it's just a spec bug. (I don't think anyone will object to speccing something that is already reality :))

littledan:
Oh really? I think we maybe could've gotten away with banning this one...

PR #810, opened 2017-02-09
Allow tail-calls in for-in loops

anba:
This was previously disallowed because users could add a custom return() method for enumeration iterators, cf. https://bugs.ecmascript.org/show_bug.cgi?id=3031.

Split from #775.


bterlson:
@msaboff can you (or someone on JSC) take a look at this?

msaboff:
@bterlson Sorry for the delay.  I think we are fine with the proposal.  I don't see this so much as an implementation issue but a language semantic issue.

bterlson:
I agree, though I would be concerned if the semantics didn't match what you implemented :)

Issue #807, opened 2017-02-09
Atomics operations should use ToIndex, not ToInteger


lars-t-hansen:
(Moved from https://github.com/tc39/ecmascript_sharedmem/issues/160)

Says @evilpie: We probably want to switch everything new to ToIndex instead of ToInteger. This affects at least ValidateAtomicAccess.

It appears that the current Atomics spec language is modeled on an older version of GetViewValue, which has since been changed to use ToIndex. I don't think there's any reason not to change the proposal to follow that change, as Tom says. Given the structure of the proposal text the only change should be in ValidateAtomicAccess, where steps 2-4 are replaced by a call to ToIndex and step 6 does not need to test for negative values.

Also see issue #410, which changed GetViewValue.

cc @leobalter

syg:
I audited all the ToInteger uses for the SAB+Atomics stuff and the only one I found is ValidateAtomicAccess.

PR #811, opened 2017-02-09
Normative: use ToIndex instead of ToInteger in ValidateAtomicAccess

syg:
Addresses #807


bterlson:
My belief is that this is simply fixing an editorial bug (resulting from changes to TAs happening in parallel with SAB work). Any objections to pulling this?

syg:
I'm assuming you're not asking for objections from me, since I opened the PR.

Context for Brian's belief: the SAB spec text was written at a time before TypedArray spec algorithms were changed from ToInteger to ToIndex. This instance of ToInteger in ValidateAtomicAccess is a holdover from that.

lars-t-hansen:
Needs test cases also.

Needs support in implementations.  @binji @pizlonator @evilpie

binji:
Thanks, Lars. Filed a v8 bug here: https://bugs.chromium.org/p/v8/issues/detail?id=5961

PR #813, opened 2017-02-10
Disallow multiple mappings for Shared Data Blocks

lars-t-hansen:
For background see https://github.com/tc39/ecmascript_sharedmem/issues/124.

The sentence added here creates the restriction that implementations may not create memory mappings or other situations such that parts of a Shared Data Block can be parts of other blocks that are not equal to the first block.  This is just a hygiene thing; there are currently no mechanisms in ES that would create this mapping.  However we could imagine some extensions that would take ES in that direction.  Suppose a mechanism exists to shrink a SAB in an agent, affecting the local length only.  If the SAB is then transmitted to another agent that already has a reference to the same memory but with the greater length, we have a situation where the two Shared Data Blocks can't be considered equal - they have different lengths, so equality is at least not obvious - but where they share memory anyway.  The sentence added here would arguably disallow the local shrinking API.

I think one could do more here, but given how simple implementations currently are I think it's not necessary to do more yet.  WebAssembly will add some pressure on SAB and we can let ourselves be informed by that, when it happens.


bterlson:
Should "part of" be defined in the new sentence? Is it subset only as you describe above, or any overlap?

lars-t-hansen:
I'll attempt to further clarify using some kind of set/member formalism, since "set" is already used in the previous sentence.

spec.html
```diff
@@ -3040,7 +3040,7 @@
...
+      <p>A data block that resides in memory that can be referenced from multiple agents concurrently is designated a <dfn>Shared Data Block</dfn>. A Shared Data Block has an identity (for the purposes of equality testing Shared Data Block values) that is <em>address-free</em>: it is tied not to the virtual addresses the block is mapped to in any process, but to the set of locations in memory that the block represents. Two data blocks are equal only if the sets of the locations they contain are equal.  Additionally, if two Shared Data Blocks are not equal then the intersection of the sets of locations they contain must be empty.  Finally, Shared Data Blocks can be distinguished from Data Blocks.</p>
```
bterlson: I think this is better but my concern is that it could be more clear that shared data block A and B are either equal, in which case their sets of locations are identical, or else not equal, in which case their sets of locations are completely disjoint. I'll make a slight edit to this text to make that more clear. Thanks @lars-t-hansen!

Issue #461, opened 2016-03-05
Internal List data structure semantics and its implications in section 9.5.11


saambarati:
I recently implemented section 9.5.11 of the spec inside JavaScriptCore (https://tc39.github.io/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-ownpropertykeys). 
Sections 17 and 19 perform the Remove operation on the List type.
I couldn't find documentation on the formal semantics of the List data structure.
Specifically, how is the Remove operation defined? Is it defined to just delete the first instance of something in the List or all instances?

Also, depending on the semantics, the algorithm defined in 9.5.11 allows for duplicate entries to be returned by this trap under certain circumstances. Is this intended? I think this would be the only place where [[OwnPropertyKeys]] has duplicate entries. Maybe we should guarantee the items are unique?

allenwb:
Lists can have duplicate entries and in cases where duplicate entries ae actually possible the algorithms must explicitly deal with that possibility.

It is a bug that 9.5.11 does not do so.

Note that [[OwnPropertyKeys]] does not have an invariant stating that the List it returns contains no duplicates. So,  it is possible that *trapResult* and/or *targetKeys* could contain duplicates. 

I've generally resisted adding invariant that is not essential for security reasons in order to avoid  generally unnecessary extra checking on the results of proxy traps. But in this particular case,  I think we should probably add the invariant that the List returned from [[OwnPropertyKeys]] contains no duplicates as such duplicates are generally unexpected and none of the uses of [[OwnPropertyKey]] in the spec. explicitly deals with that possibility.  This won't add much overhead to [[OwnPropertyKey]] traps as fixing the bug identified by this issue is going to require de-duping in most cases, so we might as well simplify by making "no dups" an invariant.

Here are the changes I propose:

In [6.1.7.3](https://tc39.github.io/ecma262/#sec-invariants-of-the-essential-internal-methods) under [[OwnPropertyKey]] add this following as an invariant:

> * The returned List  contains no duplicate entries.

In [9.5.11](https://tc39.github.io/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-ownpropertykeys) insert a new step immediately after step 8:

> If *trapResult* contains any duplicate entries,  throw a **TypeError** exception.

After the existing step 10 add a new step:
>Asset: *targetKeys* contains no duplicate entries.

and add to the list of invariants in the Note:

> * The returned List  contains no duplicate entries.

[...]

PR #833, opened 2017-03-03
Normative: Throw TypeError if proxy [[OwnPropertyKeys]] returns duplicate entries

bterlson:
Proxy [[OwnPropertyKeys]] now throws if the trap returns an array with duplicate entries and reverts a few of the changes made to handle duplicate property keys in a few places. See also discussion in #461. I believe this is needs-consensus, though I haven't heard arguments against this.

/cc @anba

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1516,6 +1516,9 @@
             The return value must be a List.
           </li>
           <li>
+            The returned list must not contain any duplicate entries.
+          </li>
+          <li>
             The Type of each element of the returned List is either String or Symbol.
           </li>
           <li>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -16564,7 +16572,7 @@
               for (const key of Reflect.ownKeys(obj)) {
                 if (typeof key === "symbol") continue;
                 const desc = Reflect.getOwnPropertyDescriptor(obj, key);
-                if (desc && !visited.has(key)) {
+                if (desc) {
```
ljharb: I think this polyfill code might need more tweaking - does it still need `visited` at all?

annevk:
Assuming this can only happen for proxy objects, this should be no problem from HTML's perspective.

spec.html
anba: We still need to track when the same property key is also present in the prototype chain.

anba:
Does anyone remember why we call [[GetOwnProperty]] for every property key instead of throwing an exception when we detected the first invariant violation? Was this added just to make sure the possible side-effects of [[GetOwnProperty]] are executed?

bterlson:
@annevk just to make double-sure, you mean "duplicate keys" can only happen with proxies, or making duplicate keys a type error is only for proxies? just want to make sure HTML doesn't depend on duplicate keys anywhere...

bterlson:
@anba I don't recall, but possibly @allenwb does.

annevk:
@bterlson the former. I think we already made sure that proxy-using objects never returned duplicate keys, though maybe now this change is made we should start enforcing that in IDL too more clearly.

leobalter:
We need volunteers to update the current tests on test262. I'm willing to support w/ reviews

ajklein:
What if the text said "If trapResult contains any entries a and b for which SameValue(a, b) is true, throw a TypeError". That's still one line, but at least refers to the right spec machinery.

bterlson:
Removing needs-consensus as we got consensus in March. Still need tests though I think.

rwaldron:
https://github.com/tc39/test262/commit/a775c243bc36a9a4b5656824ef210c01dc6afffd

leobalter:
Tests are up and landed into Test262

@bterlson :shipit:

bterlson:
YES!

PR #835, opened 2017-03-03
Small Atomics changes to reduce duplications

anba:
Here are possible changes to avoid some duplications in the Atomics section. 

Changes:
- Partially reverts f1239ca, because Atomics.wait actually only needs the GetValueFromBuffer bits from Atomics.load.
- I've replaced the boolean parameter in ValidateSharedIntegerTypedArray with a List, because some prefer to avoid boolean parameters, and because the List approach allows to write a more concise algorithm.
- The buffer address computation was duplicate in multiple operations, we can avoid this by moving it into the ValidateAtomicAccess operation.

Applies on top of #834, but I wanted to keep possible refactorings separate from bug fixes. And I've directly included the changes from #820 in 62ac9b4e1d7aabb01e933fd4ac00a04a860bcf70.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -34681,21 +34681,6 @@ THH:mm:ss.sss
           1. Return GetModifySetValueInBuffer(_buffer_, _indexedPosition_, _elementType_, _v_, _op_).
         </emu-alg>
       </emu-clause>
-
-      <emu-clause id="sec-atomicload" aoid="AtomicLoad">
-        <h1>AtomicLoad( _typedArray_, _index_ )</h1>
-        <p>The abstract operation AtomicLoad takes two arguments, _typedArray_, _index_. The operation atomically loads a value and returns the loaded value. It performs the following steps:</p>
-        <emu-alg>
-          1. Let _buffer_ be ? ValidateSharedIntegerTypedArray(_typedArray_).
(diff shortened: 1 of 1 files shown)


littledan:
@syg @lars-t-hansen What do you think of these editorial changes?

lars-t-hansen:
I think this PR should wait until #1127 is resolved since that will likely rearrange some of the relevant bits.  I'm not against it, but I'd probably prefer to re-evaluate changes after that PR lands.

(Personally I doubt that changing from a bool to a list of allowed types clarifies anything, but that's mostly a matter of preference and more for TC39 to decide.)

ljharb:
@lars-t-hansen @anba now that #1127 is merged, can this be updated?

lars-t-hansen:
@anba, I'll wait until you notify me about any changes.

ljharb:
@anba this can be rebased on top of master now that #2015 has landed, if you're still interested in pursuing the third commit.

bakkot:
The change to `ValidateSharedIntegerTypedArray` is pretty stale after the addition of BigInts and https://github.com/tc39/ecma262/pull/1908, and I think the `waitable` boolean which is now on `ValidateIntegerTypedArray` is better than a list of allowed types, so I'm going to close this. If there's still a change someone wants to make here, feel free to reopen (or, ideally, open a new PR).

PR #838, opened 2017-03-06
Fix CopyDataBlockBytes when copying bytes from shared data block to non-shared data block

anba:
This case is possible when CopyDataBlockBytes is called from CloneArrayBuffer as part of cloning the array buffer in 22.2.4.3, step 16.c.i.


bterlson:
Committed as 48ad268 (rebased and resolved conflicts with 2e45ce35).

PR #839, opened 2017-03-06
Change 'behaviour is undefined' to 'is implementation-defined'.

samth:
As pointed out by John Regehr on Twitter.


bterlson:
you got all 3 places too, nice! Will pull this once I'm finished cherry-picking #821.

samth:
Find-and-replace is magical like that.

PR #844, opened 2017-03-09
Call SpeciesConstructor in TypedArray constructor to reduce indirections

anba:
This allows us to remove the code path in CloneArrayBuffer when the _cloneConstructor_ argument is absent and simplifies the TypedArray constructor algorithm a bit. 
I didn't remove the extra `IsDetachedBuffer(_srcData_)` step for the same element case to avoid introducing a semantic change, but I think we should either remove this extra check or apply it for both cases.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33303,16 +33303,14 @@ THH:mm:ss.sss
           1. Let _srcByteOffset_ be _srcArray_.[[ByteOffset]].
           1. Let _elementSize_ be the Element Size value in <emu-xref href="#table-49"></emu-xref> for _constructorName_.
           1. Let _byteLength_ be _elementSize_ &times; _elementLength_.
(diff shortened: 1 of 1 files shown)


anba:
@bterlson Review ping?

bterlson:
@anba sure I can get to this tonight/tomorrow. If you beat me to it, it needs a rebase.

bterlson:
@anba nm on rebase, I just did it. Can you double-check to make sure it's correct?

anba:
> @anba nm on rebase, I just did it. Can you double-check to make sure it's correct?

Yes, looks good to me!

littledan:
Seems like this patch is purely editorial, right? LGTM.

bterlson:
@littledan that's the intent, at least. After reviewing again closely yesterday I believe it to be true. I'm glad you seem to as well. I'll pull this in.

Issue #845, opened 2017-03-10
Standardize Date.prototype.toString further?


littledan:
Across multiple JS implementations, it seems that Date.prototype.toString writes the timezone (in parens) in a long, locale-dependent form on Windows, but in a short form (2-4 letters) from the tz database on Unix-based OSes. [More details](https://bugs.chromium.org/p/v8/issues/detail?id=6076) in the V8 bug tracker.

The spec is really light on details for Date.prototype.toString:

> Return an implementation-dependent String value that represents tv as a date and time in the current time zone using a convenient, human-readable form.

Does anyone have a good memory of why this is the definition? Looks like it goes all the way back to ES1.

Fortunately, it seems that, at this point, implementations have converged on something that's almost always the same, with the exception of the timezone string.

For the timezone string, would it be a good idea to pick one of the two alternatives and standardize it across all platforms? Does anyone have evidence one way or the other whether either of the two is likely to be more web-compatible, or whether we need to preserve the variation?

littledan:
Crazy idea: what if we standardized on omitting the parenthized "readable" timezones, leaving just the preceding GMT+-... This would be analogous to IS08601. As @anba pointed out [Firefox has had a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1156728) where it omitted many of these timezones already, so maybe the web would be able to handle the loss.

jungshik:
> Crazy idea: what if we standardized on omitting the parenthized "readable" timezones, leaving just the preceding GMT+-... This would be analogous to IS08601. As @anba pointed out Firefox has had a bug where it omitted many of these timezones already, so maybe the web would be able to handle the loss.

Might work, but shouldn't we standardize to 'UTC...' instead of 'GMT...'?

jungshik:
Another alternative is 

```
UTC+-.... (Olsaon Timezone ID with "_" removed)
```

For instance, ```UTC-0400 (America/New York)```

Modern platforms would have Olson timezone ids handy so that implementation wouldn't be hard.

littledan:
@jungshik 

> Might work, but shouldn't we standardize to 'UTC...' instead of 'GMT...'?

Well, you'd think so, but it seems like all browsers are shipping printing it as GMT, so I'm not sure if it would be web-compatible to change that.

> For instance, UTC-0400 (America/New York)

[...]

PR #848, opened 2017-03-17
Normative: Specify Date string conversion functions

littledan:
Implementations seem to agree on semantics here for the most
part, modulo timezone strings, which are left implementation-defined.

Addresses part of #845

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27149,7 +27149,10 @@ THH:mm:ss.sss
           <emu-alg>
             1. Assert: Type(_tv_) is Number.
             1. Assert: _tv_ is not *NaN*.
-            1. Return String value formed by concatenating ToString(HourFromTime(_tv_)), `":"`, ToString(MinFromTime(_tv_)), `":"`, ToString(SecFromTime(_tv_)) and `" GMT"`
+            1. Let _hour_ be be the String representation of HourFromTime(_tv_), formatted as a two-digit number, padded to the left with a zero if necessary.
+            1. Let _minute_ be be the String representation of MinFromTime(_tv_), formatted as a two-digit number, padded to the left with a zero if necessary.
+            1. Let _second_ be be the String representation of SecFromTime(_tv_), formatted as a two-digit number, padded to the left with a zero if necessary.
+            1. Return String value formed by concatenating _hour_, `":"`, _minute_, `":"`, _second_, and `" GMT"`
           </emu-alg>
         </emu-clause>
 
@@ -27162,7 +27165,9 @@ THH:mm:ss.sss
             1. If _weekdaySeparator_ is not provided, let _weekdaySeparator_ be `" "`.
             1. Let _weekday_ be the Name of the entry in <emu-xref href="#sec-todatestring-day-names"></emu-xref> with the Number WeekDay(_tv_).
             1. Let _month_ be the Name of the entry in <emu-xref href="#sec-todatestring-month-names"></emu-xref> with the Number MonthFromTime(_tv_).
-            1. Let _result_ be the String value formed by concatenating _weekday_, _weekdaySeparator_, _month_, `" "`, ToString(DateFromTime(_tv_)), `" "`, and ToString(YearFromTime(_tv_)).
+            1. Let _day_ be be the String representation of DateFromTime(_tv_), formatted as a two-digit number, padded to the left with a zero if necessary.
+            1. Let _year_ be be the String representation of YearFromTime(_tv_), formatted as a number of at least four digits, padded to the left with zeroes if necessary.
+            1. Let _result_ be the String value formed by concatenating _weekday_, _weekdaySeparator_, _month_, `" "`, _day_, `" "`, and _year_.
           </emu-alg>
           <emu-table id="sec-todatestring-day-names" caption="Names of days of the week">
             <table>
@@ -27354,7 +27359,7 @@ THH:mm:ss.sss
             1. Assert: _tv_ is not *NaN*.
             1. Let _offset_ be LocalTZA + DaylightSavingTA(_tv_)
             1. If _offset_ &ge; 0, let _offsetSign_ be `"+"`; otherwise, let _offsetSign_ be `"-"`.
-            1. Let _offsetMin_  be the String representation of MinFromTime(abs(_offset_)), formatted as a two-digit number, padded to the left with a zero if necessary.
+            1. Let _offsetMin_ be the String representation of MinFromTime(abs(_offset_)), formatted as a two-digit number, padded to the left with a zero if necessary.
             1. Let _offsetHour_ be the String representation of HourFromTime(abs(_offset_)), formatted as a two-digit number, padded to the left with a zero if necessary.
             1. If _offset_ &ge; 0, let _offsetSign_ be `"+"`; otherwise, let _offsetSign_ be `"-"`.
             1. Let _tzName_ be an implementation-defined string, either `""` or a string of the form `" ("` _name_ `")"` where _name_ is an implementation-defined timezone name.
--- a/spec.html
+++ b/spec.html
@@ -27157,17 +27157,16 @@ THH:mm:ss.sss
         </emu-clause>
 
         <emu-clause id="sec-datestring" aoid="DateString">
-          <h1>Runtime Semantics: DateString( _tv_, _weekdaySeparator_ )</h1>
+          <h1>Runtime Semantics: DateString( _tv_ )</h1>
           <p>The following steps are performed:</p>
           <emu-alg>
             1. Assert: Type(_tv_) is Number.
             1. Assert: _tv_ is not *NaN*.
-            1. If _weekdaySeparator_ is not provided, let _weekdaySeparator_ be `" "`.
             1. Let _weekday_ be the Name of the entry in <emu-xref href="#sec-todatestring-day-names"></emu-xref> with the Number WeekDay(_tv_).
             1. Let _month_ be the Name of the entry in <emu-xref href="#sec-todatestring-month-names"></emu-xref> with the Number MonthFromTime(_tv_).
             1. Let _day_ be be the String representation of DateFromTime(_tv_), formatted as a two-digit number, padded to the left with a zero if necessary.
             1. Let _year_ be be the String representation of YearFromTime(_tv_), formatted as a number of at least four digits, padded to the left with zeroes if necessary.
-            1. Let _result_ be the String value formed by concatenating _weekday_, _weekdaySeparator_, _month_, `" "`, _day_, `" "`, and _year_.
+            1. Return the String value formed by concatenating _weekday_, `" "`, _month_, `" "`, _day_, `" "`, and _year_.
           </emu-alg>
           <emu-table id="sec-todatestring-day-names" caption="Names of days of the week">
             <table>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -27209,31 +27215,257 @@
...
         <!-- es6num="20.3.4.41.1" -->
         <emu-clause id="sec-todatestring" aoid="ToDateString">
           <h1>Runtime Semantics: ToDateString( _tv_ )</h1>
           <p>The following steps are performed:</p>
           <emu-alg>
             1. Assert: Type(_tv_) is Number.
             1. If _tv_ is *NaN*, return `"Invalid Date"`.
-            1. Return an implementation-dependent String value that represents _tv_ as a date and time in the current time zone using a convenient, human-readable form.
+            1. Let _t_ be LocalTime(_tv_).
+            1. Return the String value formed by concatenating DateString(_t_), `" "`, and TimeString(_t_), TimeZoneString(_tv_).
```
claudepache: ``... `" "`, TimeString(_t_), and TimeZoneString(_tv_).`` (the word _and_ is misplaced)

spec.html
```diff
@@ -27209,31 +27215,257 @@
           <p>The `toString` function is intentionally generic; it does not require that its *this* value be a Date object. Therefore, it can be transferred to other kinds of objects for use as a method.</p>
         </emu-note>
 
+        <emu-clause id="sec-timestring" aoid="TimeString">
+          <h1>Runtime Semantics: TimeString( _tv_ )</h1>
+          <p>The following steps are performed:</p>
+          <emu-alg>
+            1. Assert: Type(_tv_) is Number.
+            1. Assert: _tv_ is not *NaN*.
+            1. Return String value formed by concatenating _result_, ToString(HourFromTime(_tv_)), `":"`, ToString(MinFromTime(_tv_)), `":"`, ToString(SecFromTime(_tv_)) and `" GMT"`
```
claudepache: delete extraneous _result_ variable

spec.html
littledan: Fixed, thanks

spec.html
littledan: Fixed

spec.html
```diff
@@ -27209,31 +27215,257 @@
...
+        <emu-clause id="sec-timezoneestring" aoid="TimeZoneString">
+          <h1>Runtime Semantics: TimeZoneString( _tv_, )</h1>
+          <p>The following steps are performed:</p>
+          <emu-alg>
+            1. Assert: Type(_tv_) is Number.
+            1. Assert: _tv_ is not *NaN*.
+            1. Let _offset_ be (LocalTZA + DaylightSavingTA(_tv_)) / msPerSecond.
+            1. Assert: _offset_ is an integer, and -100000 &lt; _offset_ &lt; 100000.
+            1. If _offset_ &ge; 0, let _offsetSign_ be `"+"`; otherwise, let _offsetSign_ be `"-"`.
+            1. Let _offsetString_ be abs(_offset_) formatted as a four-digit number, padded to the left with zeros if necessary.
```
claudepache: The offset needs to be formatted as `HHMM`, not as `SSSS`.  The four preceding steps (starting with  `Let _offset_`) should be replaced with:

1. Let _offset_ be LocalTZA + DaylightSavingTA(_tv_) /* (NB: value in milliseconds) */
1. If _offset_ &ge; 0, let _offsetSign_ be `"+"`; otherwise, let _offsetSign_ be `"-"`.
1. Let _offsetMin_  be the String representation of MinFromTime(abs(_offset_)), formatted as a two-digit number, padded to the left with a zero if necessary.
1. Let _offsetHour_ be the String representation of HourFromTime(abs(_offset_)), formatted as a two-digit number, padded to the left with a zero if necessary.

And the last step of the algorithm, below:
1. Let _offsetString_ be the concatenation of _offsetSign_,  _offsetHour_, _offsetMin_, and _tzName_.

domenic:
> Do you have good suggestions for how to write tests here? I'm not sure how to do it in a way that will work in all time zones. (test262 has repeatedly had issues where tests were broken in certain timezones.)

Well, you can certainly test the Invalid Date cases.

You should be able to run regexps to make sure the patterns are met for all methods, including the time zone format restriction.

You should be able to test .toUTCString() exactly on its entire range of inputs, including extreme ones like 0, +/- Infinity, Number.(MIN/MAX)_SAFE_INTEGER, and Number.(MIN/MAX)_VALUE.

You should be able to pick times in the middle of a given month so that timezone changes can't change the month string, thus allowing you to test the month name mapping even for toString().

You might be able to pick times such that in every timezone they fall on the same day of the week, thus allowing testing the week day mapping for toString(), but I am not sure.

spec.html
```diff
@@ -27209,31 +27215,257 @@
...
         <h1>Date.prototype.toUTCString ( )</h1>
-        <p>This function returns a String value. The contents of the String are implementation-dependent, but are intended to represent this time value in a convenient, human-readable form in UTC.</p>
-        <emu-note>
-          <p>The intent is to produce a String representation of a date that is more readable than the format specified in <emu-xref href="#sec-date-time-string-format"></emu-xref>. It is not essential that the chosen format be unambiguous or easily machine parsable. If an implementation does not have a preferred human-readable format it is recommended to use the format defined in <emu-xref href="#sec-date-time-string-format"></emu-xref> but with a space rather than a `"T"` used to separate the date and time elements.</p>
-        </emu-note>
+        <emu-alg>
+          1. Let _O_ be this Date object.
+          1. Let _tv_ be ? thisTimeValue(_O_).
+          1. If _tv_ is *NaN*, return `"Invalid Date"`.
+          1. Return the String value formed by concatenating DateString(_tv_), `" "`, and TimeString(_tv_).
```
ljharb: also here

spec.html
```diff
@@ -27209,31 +27215,257 @@
...
         <!-- es6num="20.3.4.41.1" -->
         <emu-clause id="sec-todatestring" aoid="ToDateString">
           <h1>Runtime Semantics: ToDateString( _tv_ )</h1>
           <p>The following steps are performed:</p>
           <emu-alg>
             1. Assert: Type(_tv_) is Number.
             1. If _tv_ is *NaN*, return `"Invalid Date"`.
-            1. Return an implementation-dependent String value that represents _tv_ as a date and time in the current time zone using a convenient, human-readable form.
+            1. Let _t_ be LocalTime(_tv_).
+            1. Return the String value formed by concatenating DateString(_t_), `" "`, TimeString(_t_), and TimeZoneString(_tv_).
```
ljharb: for the space, the convention i've seen (and followed in padStart/padEnd) to avoid whitespace ambiguity is `" " (a String consisting of 0x0020 SPACE)`

spec.html
```diff
@@ -27209,31 +27215,257 @@
...
+          <h1>Runtime Semantics: TimeZoneString( _tv_, )</h1>
+          <p>The following steps are performed:</p>
+          <emu-alg>
+            1. Assert: Type(_tv_) is Number.
+            1. Assert: _tv_ is not *NaN*.
+            1. Let _offset_ be (LocalTZA + DaylightSavingTA(_tv_)) / msPerSecond.
+            1. Assert: _offset_ is an integer, and -100000 &lt; _offset_ &lt; 100000.
+            1. If _offset_ &ge; 0, let _offsetSign_ be `"+"`; otherwise, let _offsetSign_ be `"-"`.
+            1. Let _offsetString_ be abs(_offset_) formatted as a four-digit number, padded to the left with zeros if necessary.
+            1. Let _tzName_ be an implementation-defined string, either `""` or a string of the form `" ("` _name_ `")"` where _name_ is an implementation-defined timezone name.
```
ljharb: similarly (and throughout this PR where strings appear), this might want to specify the code units for the space and parens, to avoid ambiguity.

claudepache:
> You might be able to pick times such that in every timezone they fall on the same day of the week, thus allowing testing the week day mapping for toString(), but I am not sure.

You cannot. [UTC+14](https://en.m.wikipedia.org/wiki/UTC+14:00) will always be 1 or 2 days ahead of [UTC-12](https://en.m.wikipedia.org/wiki/UTC-12:00).

spec.html
littledan: What ambiguity would there be? There are many other places in the spec that use characters with multiple Unicode code points that render similarly or the same (e.g., every a character).

spec.html
littledan: Thanks, of course! Uploaded a new version to use your text.

spec.html
ljharb: ¯\_(ツ)_/¯ there are many characters for whitespace. Being precise about the code point ensures no confusion, even with a non-english-centric implementation.

littledan:
There's a small part of this which actually differs between browsers: When printing the year, if the number is below 1000, JSC and SpiderMonkey pad 0s to the left, V8 pads with spaces, and ChakraCore doesn't pad. This patch goes with the SpiderMonkey/JSC behavior. Before merging it, we should make sure that those engines are OK with the potential compatibility impact of this change.

claudepache:
> There's a small part of this which actually differs between browsers: When printing the year, if the number is below 1000, JSC and SpiderMonkey pad 0s to the left, V8 pads with spaces, and ChakraCore doesn't pad. 

We should definitely pad with 0s, so that years between 0 and 99 *look* unambiguous, and we will be able to correct the following glitch of `Date.parse()` (tested on Firefox, Safari, Chrome and Edge):

```js
var d = new Date;
d.setFullYear(97);
d.toString(); // variation on: Sat Mar 23 0097 16:04:11 GMT+0100 (CET)
var d2 = new Date(Date.parse(d.toString()));
d2.toString(); // variation of: Sun Mar 23 1997 16:04:11 GMT+0100 (CET) // oops, wrong year
```

spec.html
```diff
@@ -27137,31 +27143,257 @@
...
+          </emu-table>
+        </emu-clause>
+
+        <emu-clause id="sec-timezoneestring" aoid="TimeZoneString">
+          <h1>Runtime Semantics: TimeZoneString( _tv_, )</h1>
+          <p>The following steps are performed:</p>
+          <emu-alg>
+            1. Assert: Type(_tv_) is Number.
+            1. Assert: _tv_ is not *NaN*.
+            1. Let _offset_ be (LocalTZA + DaylightSavingTA(_tv_)) / msPerSecond.
```
jungshik: This will be in conflict with #778 that drops DaylightSavingTA(_tv_) and change LocalTZA to take two params.  We have to sort that out somehow.

spec.html
```diff
@@ -27137,31 +27143,257 @@
...
+            1. Assert: _offset_ is an integer, and -100000 &lt; _offset_ &lt; 100000.
+            1. If _offset_ &ge; 0, let _offsetSign_ be `"+"`; otherwise, let _offsetSign_ be `"-"`.
+            1. Let _offsetString_ be abs(_offset_) formatted as a four-digit number, padded to the left with zeros if necessary.
+            1. Let _tzName_ be an implementation-defined string, either `""` or a string of the form `" ("` _name_ `")"` where _name_ is an implementation-defined timezone name.
```
jungshik: Would it make sense to give a recommendation while leaving it still up to an implementation about _tzName_?  
If it does, it can be recommended to use the canonicalized timezone in Ecma402. Perhaps, cross-referencing is not  a good idea?   
Just my 2 cents.

spec.html
littledan: You're right; this is written against the current spec, but needs to be updated against the other one (based on whichever lands second). I guess rather than LocalTZA, we'd use LocalTimeZoneAdjustment(_tz_, true), right?

spec.html
littledan: I meant for this first patch to be purely 100% descriptive. Maybe a change like this could be in a follow-on patch?

As for making that recommendation: I don't know whether we can make it in good faith in the spec yet, as we don't know whether it's web-compatible yet. Maybe the right thing to do would be to try it out in one browser and see if it works. A good thing to know from the committee beforehand would be--would everyone be up for specifying this behavior, assuming it's web-compatible?

rwaldron:
Sorry for the label noise.

littledan:
I previously removed the needs tests label because tests are at https://github.com/tc39/test262/pull/930 (though they need a small fix).

spec.html
ljharb: @bterlson, thoughts on this?

bterlson: @ljharb I was going to clean this up myself (working on it soon). Planning on replacing all spaces with "the code unit 0x0020 (SPACE)"

spec.html
jungshik: Ok. this one was landed first. Then, the other one (#778) has to revise this part as well.

rwaldron:
@littledan I did a follow up from @leobalter's initial review and merged the test262 PR

PR #850, opened 2017-03-18
Normative: Date.prototype.toString throws on non-Date receiver

### LINKED ISSUE #849: What is the motivation for the Date.prototype.toString() non-Date receiver support?
In the spec, [`Date.prototype.toString`](https://tc39.github.io/ecma262/#sec-date.prototype.tostring) falls back to `"Invalid Date"` when the receiver is a non-Date Object. However, apparently d8 and jsc don't implement this and throw a TypeError, both for `Date.prototype` and for `{}`. ChakraCore seems to return `"Invalid Date"` for `Date.prototype` and throw for other objects, and SpiderMonkey implements the spec properly.

What was the motivation for this path in `Date.prototype.toString`? Was it web compatibility? If so, it seems like we have some evidence that it could be removed. It looks like this path wasn't present in [the ES 5.1 version](http://es5.github.io/#x15.9.5.2).

littledan:
JSC and V8 have been shipping the version of the specification
included in this patch, so it is likely to be web-compatible.
Throwing on receivers which are not Date instances is simpler
and more consistent with other methods in ECMAScript.

Closes #849

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27195,11 +27195,7 @@ THH:mm:ss.sss
         <h1>Date.prototype.toString ( )</h1>
         <p>The following steps are performed:</p>
         <emu-alg>
-          1. Let _O_ be this Date object.
-          1. If _O_ does not have a [[DateValue]] internal slot, then
-            1. Let _tv_ be *NaN*.
-          1. Else,
-            1. Let _tv_ be thisTimeValue(_O_).
+          1. Let _tv_ be ? thisTimeValue(*this* value).
           1. Return ToDateString(_tv_).
         </emu-alg>
         <emu-note>


leobalter:
+1, we might need to update test262, thou, but that's easy and fast.

leobalter:
```
20.3.4.41: Previous editions did not specify the value returned by Date.prototype.toString when this time value is NaN. ECMAScript 2015 specifies the result to be the String value is "Invalid Date".
```

From section/annex D, should this be changed too?

littledan:
Tests in https://github.com/tc39/test262/pull/924

anba:
I think the current behaviour is based on the discussion in <https://mail.mozilla.org/pipermail/es-discuss/2014-June/037630.html>.

littledan:
@anba Huh, based on the negative replies there, and the fact that RegExp was not similarly updated (until I developed evidence of web compatibility issues and patched it for ES2016), I assumed that this thread wasn't the one that led to the change.

Anyway, although V8 received bug reports about RegExp.prototype.toString() throwing, I don't see any bug reports about Date.prototype.toString() throwing. I'm not sure how much of a user issue this is. Note that many other methods on Date throw on Date.prototype, and conversion to a string throws for many other objects in the system, e.g., Symbol.prototype.

anba:
> Huh, based on the negative replies there, and the fact that RegExp was not similarly updated (until I developed evidence of web compatibility issues and patched it for ES2016), I assumed that this thread wasn't the one that led to the change.

I think this is the thread that led to the current Date.prototype.toString method, at least based on when it took place. The messages are from June 2014 and in the next ES6 draft (rev 26), @allenwb added 
```diff
+ 1. Let Q be this Date object.
+ 2. If Type(O) is Object and O does not have a [[DateValue]] internal slot, then
+   a. Let tv be NaN.
+ 3. Else,
   a. Let tv be this time value.
 4. Return ToDateString(tv)
```
to Date.prototype.toString.

> Anyway, although V8 received bug reports about RegExp.prototype.toString() throwing, I don't see any bug reports about Date.prototype.toString() throwing. I'm not sure how much of a user issue this is. 

If the change is web-compatible, we probably should try to apply it.

leobalter:
tests in https://github.com/tc39/test262/pull/924

leobalter:
@bterlson :shipit:

littledan:
IIRC we got consensus on this change in the May 2017 TC39 meeting. There are tests. Is this ready to merge?

bterlson:
Sorry I didn't see @leobalter's earlier comment. With tests, this is ready to go!

PR #852, opened 2017-03-18
Normative: Swap side effect and check in TypedArray constructor

### LINKED ISSUE #842: TypedArray constructor reads bufferByteLength too early
In 22.2.4.5 TypedArray ( buffer [ , byteOffset [ , length ] ] ):

In step 9, we store the value of the byte length of the buffer.
`9. Let bufferByteLength be buffer.[[ArrayBufferByteLength]].`

In step 11.a, we can end up in user-code via ToIndex.
`11.a. Let newLength be ? ToIndex(length).`

In step 11.c, we use the value of bufferByteLength from step 9 -- which could be totally incorrect now, as 11.a could have neutered the buffer.
`11.c. If offset + newByteLength > bufferByteLength, throw a RangeError exception.`

If an implementation followed this, it could result in (quite bad) security issues.

littledan:
Previously, in one of the TypedArray constructor paths, an ArrayBuffer
was checked for being detached, and then a ToIndex calculation is done.
Because ToIndex may detach the ArrayBuffer itself, this patch moves
the detached check to afterwards, so that it can include that possible
time when the ArrayBuffer may be detached.

Closes #842

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -32606,6 +32606,8 @@ THH:mm:ss.sss
           1. Let _elementSize_ be the Number value of the Element Size value in <emu-xref href="#table-49"></emu-xref> for _constructorName_.
           1. Let _offset_ be ? ToIndex(_byteOffset_).
           1. If _offset_ modulo _elementSize_ &ne; 0, throw a *RangeError* exception.
+          1. If _length_ is present and _length_ is not *undefined*, then
+            1. Let _newLength_ be ? ToIndex(_length_).
           1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
           1. Let _bufferByteLength_ be _buffer_.[[ArrayBufferByteLength]].
           1. If _length_ is either not present or *undefined*, then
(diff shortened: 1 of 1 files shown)


littledan:
@psmarshall Is this what you were thinking?

psmarshall:
Yes that looks good to me, thanks

leobalter:

anba:
Two questions:
- The proposed changes don't match the test case (test/built-ins/TypedArrays/buffer-arg-byteoffset-to-number-detachbuffer.js). Did you also mean to move step 7 (If offset modulo elementSize ≠ 0, throw a RangeError exception.) after step 9 (If IsDetachedBuffer(buffer) is true, throw a TypeError exception.), or is the test case incorrect?
- We should also change the DataView constructor, shouldn't we? (The DataView constructor also calls ToIndex after checking for a detached array buffer.)

littledan:
@anba 
- Oops, this seems like a test bug.
- Yes, that sounds like a good idea.

rwaldron:
This gained consensus on March 22nd

PR #853, opened 2017-03-20
Normative: Specify String.prototype.replace edge case

littledan:
In RegExp.prototype[Symbol.replace] as applied to RegExps, if a
substitution string is used with a numerical group reference which
is out of bounds, the behavior was previously implementation-defined.
This patch changes the behavior to not do a replacement.
Testing with eshost showed that V8, SpiderMonkey, JSC and ChakraCore
all agree on these semantics.


bterlson:
I guess technically needs-consensus? I can talk about it as part of my item if you like.

leobalter:

PR #856, opened 2017-03-22
Normative: Ensure that TypedArray.prototype.fill does one coercion

### LINKED ISSUE #855: Avoid repeated ToNumber calls in %TypedArray%.prototype.fill()
**Example:**
`%TypedArray%.prototype.fill(proxy)` causes a ToNumber with side-effects on every fill step.

From the usual behavior of Array.prototype functions, the argument conversion should probably happen only once and in this case would be different for TypedArrays. Obviously this only an issue with non-number fill values.

Array.prototype.fill and %TypedArray%.prototype.fill share the same spec implementation, modulo getting the array length. The TypedArray version should probably introduce an additional step performing a `ToNumber(value)` before step 3 in [22.1.3.6](https://tc39.github.io/ecma262/#sec-array.prototype.fill) to avoid the frequent conversion inside the fill step 7.b.

**Steps:**
The following steps lead to the `ToNumber` in question when calling `%TypedArray%.prototype.fill`.

- In [22.1.3.6 Array.prototype.fill](https://tc39.github.io/ecma262/#sec-array.prototype.fill):
    `7.b Perform ? Set(O, Pk, value, true).`
- In [7.3.3 Set ( O, P, V, Throw )](https://tc39.github.io/ecma262/#sec-set-o-p-v-throw):
    `4  Let success be ? O.[[Set]](P, V, O).`
- In [9.4.5.5 [[Set]] ( P, V, Receiver )](https://tc39.github.io/ecma262/#sec-integer-indexed-exotic-objects-set-p-v-receiver):
   `2.b.i Return ? IntegerIndexedElementSet(O, numericIndex, V).`
- In [9.4.5.9 IntegerIndexedElementSet(O, numericIndex, V)](https://tc39.github.io/ecma262/#sec-integerindexedelementset):
    `3. Let numValue be ? ToNumber(value).`

littledan:
Previously, TypedArray.prototype.fill failed to coerce its value
argument multiple in the beginning of the definition, instead coercing
it when writing into the TypedArray each time. This patch tweaks
the definition to perform the coercion once.

Fixes #855

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -31995,8 +31995,27 @@ THH:mm:ss.sss
       <!-- es6num="22.2.3.8" -->
       <emu-clause id="sec-%typedarray%.prototype.fill">
         <h1>%TypedArray%.prototype.fill ( _value_ [ , _start_ [ , _end_ ] ] )</h1>
-        <p>%TypedArray%`.prototype.fill` is a distinct function that implements the same algorithm as `Array.prototype.fill` as defined in <emu-xref href="#sec-array.prototype.fill"></emu-xref> except that the *this* object's [[ArrayLength]] internal slot is accessed in place of performing a [[Get]] of `"length"`. The implementation of the algorithm may be optimized with the knowledge that the *this* value is an object that has a fixed length and whose integer indexed properties are not sparse. However, such optimization must not introduce any observable changes in the specified behaviour of the algorithm.</p>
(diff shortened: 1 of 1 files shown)


leobalter:

leobalter:
@bterlson should we have a "has consensus" label for unmerged PRs/issues?

PR #857, opened 2017-03-23
Normative: Define particular precision for toFixed and friends

littledan:
Previously, the specification for Number.prototype.toFixed, Number
gave a fixed range for the fractionDigits parameter, but allowed
implementations to permit a larger range. Historically, only
SpiderMonkey accepted a larger range, and other implementations
threw a RangeError outside the range.

This patch specifies the SpiderMonkey behavior. The range checks
always permitted more decimal digits to be printed than what is
provided by the accuracy, so the limits are arbitrary. Waldemar
Horwat has mentioned that SpiderMonkey's higher limits have
been useful for him in explaining floating point numbers to
others.

An alternative would be to retain the previous limits and remove
the implementation-dependent aspects. Either way, locking down
this detail should help developers get consistent behavior
between implementations.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -24999,7 +24999,7 @@
         <emu-alg>
           1. Let _x_ be ? thisNumberValue(*this* value).
           1. Let _f_ be ? ToInteger(_fractionDigits_). (If _fractionDigits_ is *undefined*, this step produces the value 0.)
-          1. If _f_ &lt; -20 or _f_ &gt; 100, throw a *RangeError* exception.
+          1. If _f_ &lt; 0 or _f_ &gt; 100, throw a *RangeError* exception.
           1. If _x_ is *NaN*, return the String `"NaN"`.
           1. Let _s_ be the empty String.
           1. If _x_ &lt; 0, then
--- a/spec.html
+++ b/spec.html
@@ -22446,7 +22446,6 @@
       Except as restricted in <emu-xref href="#sec-forbidden-extensions"></emu-xref>, an implementation may provide additional types, values, objects, properties, and functions beyond those described in this specification. This may cause constructs (such as looking up a variable in the global scope) to have implementation-defined behaviour instead of throwing an error (such as *ReferenceError*).
     </li>
   </ul>
-  <p>An implementation may define behaviour other than throwing *RangeError* for `toFixed`, `toExponential`, and `toPrecision` when the _fractionDigits_ or _precision_ argument is outside the specified range.</p>
 
   <emu-clause id="sec-host-report-errors" aoid="HostReportErrors">
     <h1>HostReportErrors ( _errorList_ )</h1>
@@ -24956,7 +24955,7 @@
             1. Let _x_ be -_x_.
           1. If _x_ = *+&infin;*, then
             1. Return the concatenation of the Strings _s_ and `"Infinity"`.
-          1. If _f_ &lt; 0 or _f_ &gt; 20, throw a *RangeError* exception. However, an implementation is permitted to extend the behaviour of `toExponential` for values of _f_ less than 0 or greater than 20. In this case `toExponential` would not necessarily throw *RangeError* for such values.
+          1. If _f_ &lt; 0 or _f_ &gt; 100, throw a *RangeError* exception.
           1. If _x_ = 0, then
             1. Let _m_ be the String consisting of _f_+1 occurrences of the code unit 0x0030 (DIGIT ZERO).
             1. Let _e_ be 0.
@@ -25000,7 +24999,7 @@
         <emu-alg>
           1. Let _x_ be ? thisNumberValue(*this* value).
           1. Let _f_ be ? ToInteger(_fractionDigits_). (If _fractionDigits_ is *undefined*, this step produces the value 0.)
-          1. If _f_ &lt; 0 or _f_ &gt; 20, throw a *RangeError* exception. However, an implementation is permitted to extend the behaviour of `toFixed` for values of _f_ less than 0 or greater than 20. In this case `toFixed` would not necessarily throw *RangeError* for such values.
+          1. If _f_ &lt; -20 or _f_ &gt; 100, throw a *RangeError* exception.
           1. If _x_ is *NaN*, return the String `"NaN"`.
           1. Let _s_ be the empty String.
           1. If _x_ &lt; 0, then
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -25000,7 +24999,7 @@
         <emu-alg>
           1. Let _x_ be ? thisNumberValue(*this* value).
           1. Let _f_ be ? ToInteger(_fractionDigits_). (If _fractionDigits_ is *undefined*, this step produces the value 0.)
-          1. If _f_ &lt; 0 or _f_ &gt; 20, throw a *RangeError* exception. However, an implementation is permitted to extend the behaviour of `toFixed` for values of _f_ less than 0 or greater than 20. In this case `toFixed` would not necessarily throw *RangeError* for such values.
+          1. If _f_ &lt; -20 or _f_ &gt; 100, throw a *RangeError* exception.
```
ljharb: Why would there be less than zero fractionDigits, and have that not throw a RangeError? What does it mean to have -3 fraction digits?

littledan: The behavior in this patch (which Firefox has been shipping since forever) is to round off digits above the decimal point. E.g., (34).toFixed(-1) => "30". The logic only makes sense for toFixed and not toPrecision or toExponential, so that's why it's here and not there.

ljharb: Then shouldn't `(34).toFixed(-3)` throw? In Firefox right now it returns `0` but that seems both useless and silly.

I love the rest of this patch, but I feel like the lower bound here should remain 0.

littledan: Well, the definition follows mathematically pretty cleanly; I don't see why we should especially throw on 0. Anyway, I also can't think of a time when I'd want to use a negative number, and would be fine with throwing on negative offsets (and, I probably wouldn't use any of these functions, and use `Intl.NumberFormat` instead).

I'll see if we can get more from Waldemar about the motivation for these semantics in his implementation.

ljharb: I mean, if I provide a negative number that is greater than the number of nonfractional digits, it seems like it should throw, even if it works for negative numbers (which I also don't see the value of)

littledan: Currently, throwing behavior has nothing to do with the number receiver. I don't see a reason to change that--it would make things more complex for no particular user benefit that I can see.

ljharb: I agree with you - but I'm saying that since "the meaning of a negative number" actually does depend on the receiver (whereas the meaning of a positive number does not), it should probably either throw at `< 0`, or throw dependent on the receiver.

Yaffle: @littledan , just a note:
You need to change the rest of the `Number#toFixed` to work with negative fractionDigits:

-            1. If _f_ &ne; 0, then
+            1. If _f_ &gt; 0, then
              1. Let _k_ be the number of elements in _m_.
              1. If _k_ &le; _f_, then
                1. Let _z_ be the String consisting of _f_+1-_k_ occurrences of the code unit 0x0030 (DIGIT ZERO).
                1. Let _m_ be the concatenation of Strings _z_ and _m_.
                1. Let _k_ be _f_ + 1.
              1. Let _a_ be the first _k_-_f_ elements of _m_, and let _b_ be the remaining _f_ elements of _m_.
              1. Let _m_ be the concatenation of the three Strings _a_, `"."`, and _b_.
+            1. If _f_ &lt; 0 and _m_ &ne; "0", then
+               1. Let _m_ be the concatenation of the three Strings _m_, the String consisting of 0-_f_ occurrences of the code unit 0x0030 (DIGIT ZERO).
          1. Return the concatenation of the Strings _s_ and _m_.

domenic: >  but I'm saying that since "the meaning of a negative number" actually does depend on the receiver

I don't understand this perspective. A negative number in this position has a well-defined mathematical meaning and effect, as outlined in the algorithm in this PR. It's not special-cased in the mathematical operation toFixed represents, so I don't see why it should be in this API.

ljharb: I'm not familiar with a mathematical operation involving strings.

It certainly has a well-defined meaning in this PR - but I think that `(24).toFixed(-3)` being anything but an error is confusing and makes no sense, whereas I think `(24).toFixed(-2)` being `0` makes perfect sense but is useless, and `(24).toFixed(-1)` being `20` makes perfect sense and is intuitive with the behavior of `toFixed` given positive numbers (although I still don't see the value or use case of providing a negative number).

claudepache: @ljharb The mathematical interpretation of `.toFixed(n)` is:

1. round the number to the nearest multiple of 10<sup><i>-n</i></sup>;
2. format the result using a decimal representation, with _n_ digits after the decimal point (if _n_ is positive) or as an integer (if _n_ is negative).

It is debatable if we should support `x.toFixed(-3)` at all; but one the support it for a given `x`, we should support it for all `x`. Introducing special cases such as “support it only for the `x` that is easily understandable for humans” is a less useful semantics, and that criterion is very subjective.

claudepache:
I question the proposed semantics (and the Firefox implementation) of `x.toFixed(n)` for _n_ < 0 and  _x_ ≥ 10<sup>21</sup>. We have:

* `1.23456789e21.toFixed(-18)` → `"1.23456789e21"` (no rounding)
* `1.23456789e20.toFixed(-18)` → `"123000000000000000000"` (rounding to the nearest multiple of 10<sup>18</sup>).

I expect that rounding also occurs in the first case.

spec.html
ljharb: @claudepache thanks for explaining; that makes sense, but since I think "N digits + toFixed(more than N digits)" is incredibly confusing, I'd argue pretty hard to not support negative numbers whatsoever to avoid the problem.

littledan:
@claudepache What's the issue for  _x_ ≥ 10<sup>21</sup>?

littledan:
@claudepache What's the issue? The semantics in the first bullet point follow from 7.a, and the second follow from step 8.

claudepache:
> The semantics in the first bullet point follow from 7.a, and the second follow from step 8.

Yes, and that’s precisely the problem.

In more details: I expect that `x.toFixed(n)` rounds the number _x_ to _n_ digits after the decimal point; — or, using a formulation that makes sense also for negative values of _n_, to the nearest multiple of 10<sup><i>-n</i></sup>.

But the current algorithm does not perform such a rounding for _x_ ≥ 10<sup>21.

As long as _n_ is nonnegative (even, as long as _n_ ≥ -5), this lack of rounding is not observable, because implementations typically do not display more than ca. 17 significant digits anyway. But if we begin to accept negative values for _n_, that becomes an issue.

littledan:
@claudepache Oh right of course. This issue seems fixable, but the fix introduces a little complexity. Generally, unless we see particular use cases, I think the right fix is just to ban negative values here.

leobalter:
tests in https://github.com/tc39/test262/pull/932

@bterlson another one to 🚢

PR #860, opened 2017-04-01
Normative: Relax restrictions on `with` statement

jugglinmike:
@bterlson This is normative, but I don't think it's significant enough to warrant the "needs consensus" label. Probably you can just merge it immediately.

See also the requisite update for Test262: https://github.com/tc39/test262/pull/943

Commit message:

> Recent industry trends have shown an increased interest in the use of
> the `with` statement as a means for more concise code. Because class
> bodies and module code are always strict mode code, this restriction
> severely limits usage of this statement.
>
> Remove the strict mode restriction so that the `with` statement may be
> leveraged in more contexts.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -16660,9 +16660,6 @@
       <emu-grammar>WithStatement : `with` `(` Expression `)` Statement</emu-grammar>
       <ul>
         <li>
-          It is a Syntax Error if the code that matches this production is contained in strict mode code.
-        </li>
-        <li>
           It is a Syntax Error if IsLabelledFunction(|Statement|) is *true*.
         </li>
       </ul>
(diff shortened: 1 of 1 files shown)


ljharb:
Um, this is very significant' I don't want the with statement allowed anywhere else.

ljharb:
Can you highlight these "increased trends" and "interest"? I've not come across either.

syg:
What's the day today?

-- 
       shu

erights:
Yes, this was removed on purpose!! It is very important that it remain prohibited in strict code. It would certainly need to go through the whole consensus process. I doubt it would get very far.

ljharb:
(－‸ლ) that's what I get for reviewing code first thing in the morning

erights:
Doh!
Got me.

tkellen:
well played sir

bterlson:
I think I speak for the entire JavaScript when I say: closing this with prejudice :-P

Well played, though :-D

jugglinmike:
I guess I can still hope for ES2019. Maybe JavaScript will be ready for it by then. Thanks for your consideration!

bterlson:
ES2019 is possible; we can include this change along with my planned changes to rename const as `let`, `let` as `var` and `var` as `bad`.

PR #864, opened 2017-04-04
Normative: Specify RegExp malformed character class behavior

### LINKED ISSUE #863: Specify web reality RegExp character class invalid control escape semantics
Sadness: @schuay noted in [this V8 bug](https://bugs.chromium.org/p/v8/issues/detail?id=6201) that some invalid character classes are permitted in browsers. The 3/4 majority semantics (minus ChakraCore) seems to be to treat `/[\c%]/` as `/[\\c%]/`, and to treat `/[\c]/` as `[\\c]`, where currently, neither are supported in the grammar. (They are included in ChakraCore's grammar, but with different semantics.) See the bug for more details. Should we standardize the majority semantics?

littledan:
This patch specifies a previously unspecified piece of RegExp
syntax and semantics: If a character class has a control escape
(\c) which is not followed by an extended control character
[a-ZA-Z0-9_], then four engines will accept it as syntactically
valid. V8, JSC and SpiderMonkey will treat it as the character
class [\\c]; ChakraCore's behavior differs and depends on what
follows.

This patch codifies the majority behavior within the Annex B
RegExp extensions. This change is consistent with malformed \c
behavior when it takes place outside of a character class: It
behaves like \\c, rather than an identity escape.

Closes #863

Thanks @schuay for raising this issue and @anba for already specifying all the other surrounding cases I came across.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38581,7 +38581,6 @@ THH:mm:ss.sss
           [~U] `c` ClassControlLetter
           CharacterClassEscape
           CharacterEscape[?U]
-          [~U] `c`
 
         ClassControlLetter ::
           DecimalDigit
@@ -38663,12 +38662,6 @@ THH:mm:ss.sss
           1. Let _j_ be the remainder of dividing _i_ by 32.
           1. Let _d_ be the character whose character value is _j_.
           1. Return the CharSet containing the single character _d_.
-        </emu-alg>
-
-        <p>ClassEscape (<emu-xref href="#sec-classescape"></emu-xref>) includes the following additional evaluation rule:</p>
-        <p>The production <emu-grammar>ClassEscape :: `c`</emu-grammar> evaluates as follows:</p>
-        <emu-alg>
-          1. Return the CharSet containing the characters `\` and `c`.
         </emu-alg>
 
         <emu-annex id="sec-runtime-semantics-characterrangeorunion-abstract-operation" aoid="CharacterRangeOrUnion">
--- a/spec.html
+++ b/spec.html
@@ -38585,6 +38585,11 @@ THH:mm:ss.sss
         ClassControlLetter ::
           DecimalDigit
           `_`
+
+        ClassAtomNoDash[U] ::
+          SourceCharacter but not one of `\` or `]` or `-`
+          `\` ClassEscape[?U]
+          `\`
       </emu-grammar>
       <emu-note>
         <p>When the same left hand sides occurs with both [+U] and [\~U] guards it is to control the disambiguation priority.</p>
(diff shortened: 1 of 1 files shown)


ljharb:
I'm assuming this needs consensus; but I suspect if chakra is fine with it, it may not. @bterlson?

anba:
Hmm, I no longer remember why I didn't cover this case in fbdfda6f2a613f3c4813d4b34e32f5c5134cf921. But I don't think the proposed change is correct, instead we need to add `ClassAtomNoDash :: \` or alternatively change ClassAtomNoDash to:
```
ClassAtomNoDash[U]::
  \ ClassEscape[?U]
  [~U] SourceCharacter but not one of ] or -
  [+U] SourceCharacter but not one of \ or ] or -
```
to match the approach in ExtendedPatternCharacter. 

That way `/[\c-f]/.test("d")` will return `true` just like it does today in SM/JSC/V8/Chakra.

littledan:
@anba Do you think there's a way to word the specification that explicitly calls out the \c sequence? It's pretty hard to deduce that the inclusion of \ in ExtendedPatternCharacter leads to allowing \c outside of classes (and nothing else, due to the lower precedence); it'd be nice to not put in another confusing such case if it's possible.

anba:
@littledan I'm not sure there's an easy way to do this. So I think it's either adding a note to explain this special-case or copy the matcher algorithm from `Alternative :: Alternative Term`, like:

1. Let A be a one-element CharSet containing the character 'c'.
2. Let B be a one-element CharSet containing the character '\\'.
3. Let m1 be CharacterSetMatcher(A, false).
4. Let m2 be CharacterSetMatcher(B, false).
5. Return an internal Matcher closure that takes two arguments, a State x and a Continuation c, and performs the following steps when evaluated:
    a. Let d be a Continuation that takes a State argument y and returns the result of calling m2(y, c).
    b. Call m1(x, d) and return its result.

littledan:
Oof, that doesn't look so great either. Maybe keeping your approach with a note is best. Any other thoughts?

littledan:
I implemented @anba's suggestion and added another test in the test262 PR.

anba:
> Oof, that doesn't look so great either. Maybe keeping your approach with a note is best. Any other thoughts?

I've totally forgot we could use a simple lookahead restriction &rarr; #910.

littledan:
The intended semantics of these changes got consensus at the May TC39 meeting, but I need to do a non-normative rewrite on top of @anba's excellent #910 patch.

You can find tests at https://github.com/tc39/test262/pull/945 .

littledan:
OK, I added a lookahead assertion similar to @anba's other patch; this seems important for avoiding an ambiguity in the grammar, similar to the other case which @anba handled. Could someone take a look and see if this seems right?

bterlson:
With db9c53373ea2592663ac1c1e383b91f83baced4a I think this is good to go.

PR #865, opened 2017-04-04
Issue 861: Only remove a waiter from the waiter list if the wait timed out

lars-t-hansen:
@AndrewScheidecker found a bug in the wait algorithm, this should fix it.


bterlson:
@lars-t-hansen Is this bug-fix important for ES2017?

lars-t-hansen:
@bterlson, a literal implementation of the buggy spec will quickly crash in the call to RemoveWaiter since RemoveWaiter asserts that the item to be removed is in the list.  That will raise the question about whether the item should have been removed by Atomics.wake or should be left in the list for the RemoveWaiter call to remove.  It'll probably be clear (after some thought) that only the first option can lead to correct code.  So, probably not *important* for ES2017, but definitely nice to have.

PR #866, opened 2017-04-05
ES2017 review third round

anba:
Maybe it's time to update https://tc39.github.io/ecma262/#sec-algorithm-conventions to allow the `?` short-hand when invoking algorithms associated to grammar productions instead of just abstract operations.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -5583,7 +5583,7 @@
             1. Return NormalCompletion(~empty~).
           </emu-alg>
           <emu-note>
-            <p>Global function declarations are always represented as own properties of the global object. If possible, an existing own property is reconfigured to have a standard set of attribute values. Steps 10-12 are equivalent to what calling the InitializeBinding concrete method would do and if _globalObject_ is a Proxy will produce the same sequence of Proxy trap calls.</p>
+            <p>Global function declarations are always represented as own properties of the global object. If possible, an existing own property is reconfigured to have a standard set of attribute values. Steps 8-9 are equivalent to what calling the InitializeBinding concrete method would do and if _globalObject_ is a Proxy will produce the same sequence of Proxy trap calls.</p>
           </emu-note>
         </emu-clause>
       </emu-clause>
--- a/spec.html
+++ b/spec.html
@@ -4472,12 +4472,12 @@
           1. If Type(_key_) is String, then
             1. Let _desc_ be ? _O_.[[GetOwnProperty]](_key_).
             1. If _desc_ is not *undefined* and _desc_.[[Enumerable]] is *true*, then
-              1. If _kind_ is *"key"*, append _key_ to _properties_.
+              1. If _kind_ is `"key"`, append _key_ to _properties_.
               1. Else,
                 1. Let _value_ be ? Get(_O_, _key_).
-                1. If _kind_ is *"value"*, append _value_ to _properties_.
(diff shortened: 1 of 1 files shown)


bterlson:
I think it is, I didn't notice that fc1cbe3 depended on that. I wasn't too happy with the `?` operator just kind of sitting after perform but allowing the convention and keeping fc1cbe3 is my preferred approach. What do you think?

anba:
I think I'd probably revert fc1cbe3e19798e2611d3947abd77e4cdfe64f1f9 for now, and when https://tc39.github.io/ecma262/#sec-algorithm-conventions was updated, we can re-apply fc1cbe3e19798e2611d3947abd77e4cdfe64f1f9 and even get rid of many more ReturnIfAbrupt calls in the spec.

spec.html
```diff
@@ -9543,10 +9542,10 @@
         &lt;ZWJ&gt;
 
       UnicodeIDStart ::
-        &gt; any Unicode code point with the Unicode property &ldquo;ID_Start&rdquo;
+        &gt; any Unicode code point with the Unicode property “ID_Start”
```
mathiasbynens: These would be the first non-ASCII symbols in `spec.html`. Are we sure this doesn’t break anyone’s scripts? (It shouldn’t, but hey…)

FWIW, here’s the ecmarkup bug: https://github.com/bterlson/ecmarkup/issues/118. Fixing that would be preferable to this workaround IMHO.

anba: Agreed, I didn't know this was a known issue in ecmarkup. Looks like there's even an open PR to fix this bug: https://github.com/bterlson/ecmarkup/pull/117

bterlson:
I've committed most of these (minor editorial fixes also committed to es2017, sending final draft to ECMA tomorrow). There are a few I've left off:

* 685a892: I actually always intended that *'s be used for primitive values, but I see ` has taken off. Should we get rid of the distinction and also use ` for numbers and Booleans? /cc @jmdyck 
* 1e37b9b: There are more instances of this (and presumably non-strict code code should be non-strict mode code?) I'd rather leave it for now until I find where I previously discussed this.

jmdyck:
The MS Word version of the spec presented string literals pretty consistently in monospace font, so those all became backticks in ecmarkdown. (So it hasn't really "taken off", it's always been there.)

It would probably be more consistent with the current definition of ecmarkdown to use asterisks for string values, as you intended (except if/where the spec is specifically talking about a string literal as a chunk of source code, which is rare, I think).

The opposite change (using backtick for numbers and booleans) would, I think, obscure the distinction between (concrete) source code and (abstract) values.

domenic:
My preference would be stick with backticks for strings and no-special-font for number literals or other primitives, just for the smallest delta from the current spec. Then figure out our story for number literals per https://github.com/tc39/proposal-bigint/issues/10, later.

jmdyck:
For some previous discussion of "[non-]strict mode code" vs "[non-]strict code", see https://bugs.ecmascript.org/show_bug.cgi?id=4144 and https://bugs.ecmascript.org/show_bug.cgi?id=4324

bterlson:
@jmdyck Indeed you are correct, I've used *'s in a few other specs but never ecma262 (except by mistake).

@Domenic I agree we should clean up number notations and then make changes.

Issue #832, opened 2017-03-02
ES2017: allowing async method named "function" is future hostile


allenwb:
Summary:

The ES2017 draft currently allows this:
```js
class {
   async function() {}; //this creates an property named "function" whose value is an async function object;
}
```

This is future hostile because it creates issues if we every want to add lexically scoped declarations to class bodies. It would be more future proof to disallow (via a look-ahead restriction) `function` in the _PropertyName_ position of _AsyncMethod_ . It would still be possible to define an async method named "function" by quoting the property name.

We should add the look-ahead restriction in the final version of the ES2017 spec.

TL;DR

The possibility of adding lexically scoped declarations to class bodies have long been discussed and is under active discussion, particularly as a complement to private fields. However, consider the following:

```js
class {
  async function() {}   //ES2017 draft says this is an async method named "function"
  async function helper() {}  //in the future we might want this to be class-scoped async function declaration
}
```
The problem is that in parsing a _ClassElement_ whose first token is `async` we can't distinguish _AsyncMethod_ and _AsyncFunctionDeclaration_  using single token look-ahead. 

We should future-proof ES2017 to allow for the future possibility of such declarations.  This can be accomplished by changing the [definition of _AsyncMethod_](https://tc39.github.io/ecma262/#prod-AsyncMethod) to:

<pre>
<i>AsyncMethod</i><sub>[Yield, Await]</sub> :
&nbsp;&nbsp;&nbsp;<b>async</b> [no <i>LineTerminator</i> here] [lookahead ∉ {<b>function</b>}]&nbsp;<i>PropertyName</i><sub>[?Yield, ?Await]</sub> <b>(</b> <i>UniqueFormalFarameters</i><sub>[~Yield, +Await]</sub> <b>) {</b> <i>AsyncFunctionBody</i> <b>}</b>
</pre>

**Web compatibility considerations.**

Most browser are already shipping async methods.  Can we get away with making this change at this late data?  Probably. It's seems likely that very few (if any) websites have yet defined async methods named "function".  But we will never know if we don't try to make this change. If we do make this change in the final spec. one or more browser will test it in preview release.  If no compatibility problems are found, we can expect that the change will be widely adopted. If compatibility problems are found the change wouldn't be adopted by browsers and we can remove it in ES2018.  Changing it now in the ES2017 spec. is probably the only opportunity we will have to do this future proofing.

domenic:
[...]

PR #884, opened 2017-04-11
Prohibit async methods named 'function'

littledan:
Addresses #832


leobalter:
LGTM, but we might need to seek consensus in the next meeting. I'm +1 to restrict the function name.

rwaldron:
Closing per consensus

littledan:
To expand on @rwaldron 's comments, we decided at the May 2017 TC39 meeting not to do this patch, as it seemed like, because the grammar of class bodies is more restricted, two-token lookahead would not be a problem. On the other hand, it would be an unexpected lack of behavior to not be able to make async methods with this name.

PR #885, opened 2017-04-11
Normative: Prohibit ASI between 'let' and 'yield'/'await'

littledan:
Previously, @jswalden found that ASI would depend on whether the
[Yield] and [Await] flags were set. This minor change prohibits
yield and await as BindingIdentifiers based on static semantics
rules, rather than grammar parameters, in order to make ASI
invariant between the two flags.

Inspired by https://github.com/tc39/test262/pull/956

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -10724,18 +10724,18 @@
     <emu-grammar>
       IdentifierReference[Yield, Await] :
         Identifier
-        [~Yield] `yield`
-        [~Await] `await`
+        `yield`
+        `await`
 
       BindingIdentifier[Yield, Await] :
         Identifier
-        [~Yield] `yield`
-        [~Await] `await`
+        `yield`
+        `await`
 
       LabelIdentifier[Yield, Await] :
         Identifier
-        [~Yield] `yield`
-        [~Await] `await`
+        `yield`
+        `await`
 
       Identifier :
         IdentifierName but not ReservedWord
@@ -10751,27 +10751,33 @@
         </li>
       </ul>
       <emu-grammar>
-        IdentifierReference : `yield`
+        IdentifierReference[Yield, Await] : `yield`
 
-        BindingIdentifier : `yield`
+        BindingIdentifier[Yield, Await] : `yield`
 
-        LabelIdentifier : `yield`
+        LabelIdentifier[Yield, Await] : `yield`
       </emu-grammar>
       <ul>
         <li>
           It is a Syntax Error if the code matched by this production is contained in strict mode code.
         </li>
+        <li>
+          It is a Syntax Error if this production is matched with the Yield grammar parameter set.
+        </li>
       </ul>
       <emu-grammar>
-        IdentifierReference : `await`
+        IdentifierReference[Yield, Await] : `await`
 
-        BindingIdentifier : `await`
+        BindingIdentifier[Yield, Await] : `await`
 
-        LabelIdentifier : `await`
+        LabelIdentifier[Yield, Await] : `await`
       </emu-grammar>
       <ul>
         <li>
           It is a Syntax Error if the goal symbol of the syntactic grammar is |Module|.
+        </li>
+        <li>
+          It is a Syntax Error if this production is matched with the Await grammar parameter set.
         </li>
       </ul>
       <emu-grammar>
--- a/spec.html
+++ b/spec.html
@@ -10724,8 +10724,8 @@
     <emu-grammar>
       IdentifierReference[Yield, Await] :
         Identifier
-        `yield`
-        `await`
+        [~Yield] `yield`
+        [~Await] `await`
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -10724,18 +10724,18 @@
     <emu-grammar>
       IdentifierReference[Yield, Await] :
         Identifier
-        [~Yield] `yield`
-        [~Await] `await`
+        `yield`
```
anba: This change will lead to a reduce-reduce conflict.

littledan: Oh, thanks for pointing this out. It seems like the reduce-reduce conflict only really comes up for IdentifierReference, not the other two. I only threw in IdentifierReference here for symmetry, and it's not needed for the actual change here. Would backing out that part be enough?

littledan:
Test262 tests for this PR pass in V8, JSC, ChakraCore and SpiderMonkey.

caitp:
> Test262 tests for this PR pass in V8, JSC, ChakraCore and SpiderMonkey.

That might not be entirely true ever since https://chromium-review.googlesource.com/c/472568/ landed

spec.html
```diff
@@ -10751,28 +10753,34 @@
         </li>
       </ul>
       <emu-grammar>
-        IdentifierReference : `yield`
+        IdentifierReference[Yield, Await] : `yield`
```
anba: Grammar parameters are generally not repeated in static semantics.

spec.html
```diff
@@ -10751,28 +10753,34 @@
...
 
-        LabelIdentifier : `yield`
+        LabelIdentifier[Yield, Await] : `yield`
       </emu-grammar>
       <ul>
         <li>
           It is a Syntax Error if the code matched by this production is contained in strict mode code.
         </li>
+        <li>
+          It is a Syntax Error if this production is matched with the Yield grammar parameter set.
```
anba: `s/Yield/<sub>[Yield]</sub>/`

spec.html
```diff
@@ -10751,28 +10753,34 @@
...
 
-        LabelIdentifier : `await`
+        LabelIdentifier[Yield, Await] : `await`
       </emu-grammar>
       <ul>
         <li>
           It is a Syntax Error if the goal symbol of the syntactic grammar is |Module|.
         </li>
+        <li>
+          It is a Syntax Error if this production is matched with the Await grammar parameter set.
```
anba: `s/Await/<sub>[Await]</sub>/`

spec.html
```diff
@@ -10741,6 +10741,8 @@
         IdentifierName but not ReservedWord
     </emu-grammar>
 
+    <emu-note>`yield` and `await` are permitted as |BindingIdentifier| in the grammar, and prohibited with static semantics below, to prohibit ASI in cases such as `let\nawait`.</emu-note>
```
anba: s/ASI/automatic semicolon insertion/ and maybe use a `<pre>` block (I haven't checked the generator html, though!).

```html
<pre><code class="javascript">
let
await
0;
</code></pre>
```

spec.html
anba: And strictly speaking this restriction only makes sense for `BindingIdentifier` (because `IdentifierRefernce : [~Yield] yield` already ensures `yield` cannot be matched as an IdentifierReference). So maybe add a new early error restriction `<emu-grammar>` block for this case?

anba:
I think it should work, but I'd prefer to get some reviews on this change.

spec.html
littledan: Done

spec.html
littledan: Done

spec.html
littledan: Done

spec.html
littledan: Done

littledan:
What do people think of this change? cc implementers @msaboff @ajklein @syg @jswalden 

Can we gather consensus for this normative change on this GitHub thread, or should we bring it up at the next meeting?

jswalden:
Eliminating the inanity seems absolutely good to me. But I'm looking at this on a phone, on an unrendered version of the spec in the patch, and I can't say the tack is right with true confidence. Probably is, especially if others say it is, but don't look to me for the vouch - not least because I'm hiking the PCT now and won't really be around hopefully til mid-September. :-)

syg:
I think this change is great.

ajklein:
@littledan given that you implemented most of V8's current `let` parsing, I'd defer to you; but this definitely sounds like a good idea to me.

littledan:
FWIW I didn't implement this aspect of the specification because I didn't understand it at the time, until @jswalden pointed it out.

bterlson:
I can't imagine anyone objecting to this honestly. I'm gonna pull it :)

Issue #753, opened 2016-12-14
Annex B.3.3.3 should not unconditionally overwrite global bindings with `undefined` in EvalDeclarationInstantiation


syg:
André Bargull filed the following bug for Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1317373

This seems like a spec bug to me. EDI probably needs extra steps other than CanDeclareGlobalFunction to see if it should call CreateGlobalFunctionBinding.

The above bug reproduced inline for ease:

Test case:
```javascript
var f = "x";
eval("print(f); { function f(){} }")
```

Expected: Prints "undefined"
Actual: Prints "x"

V8 and Chakra print "x", JSC prints "function f() {  }".

ES2017 spec: B.3.3.3, step 7.a.i.i
https://tc39.github.io/ecma262/#sec-web-compat-evaldeclarationinstantiation

bakkot:
In particular, according to [B.3.3.3](https://tc39.github.io/ecma262/#sec-web-compat-evaldeclarationinstantiation) (`EvalDeclarationInstantiation`):

* 1.d.ii.4 returns immediately, leaving `bindingExists` as false
* `varEnvRec` is indeed a global environment record, so 1.d.ii.5 is entered
* The global environment record has no lexical declaration named `f`, so 1.d.ii.5.a is entered
* `CanDeclareGlobalFunction` will see that the existing property named `f` is a data property and is writable and enumerable, and so returns true: thus `fnDefinable` will be set to true.
* 1.d.ii.7.a.i.i then performs `CreateGlobalFunctionBinding(F, undefined, true)`, which sets the value of the `f` global variable to `undefined`.

bakkot:
Contrast the behavior in a function:

```js
function a() {
  var f = 42;
  eval("print(f); { function f(){} }");
}
a();
```

Per spec this prints 42, because [B.3.3.3](https://tc39.github.io/ecma262/#sec-web-compat-evaldeclarationinstantiation).1.d.ii.7.a.ii.i avoids clobbering the existing `f` var binding.

syg:
FWIW no implementation implements the current spec behavior of overwriting with "undefined".

littledan:
Ugh, what should the semantics be? Will it read the value of the property of the global object and rewrite it (invoking getters at the top of the eval block)? Does any browser do that?

bakkot:
It's just occurred to me that this comes up without `eval` too, in `GlobalDeclarationInstantiation`:

```html
[...]

PR #888, opened 2017-04-12
Normative: Create global properties for sloppy function hosting as vars

littledan:
This patch removes a case where a synthetic *undefined* was visible
as a result of legacy sloppy-mode function hoisting.

This patch addresses https://github.com/tc39/ecma262/issues/753


anba:
I wondered why we decided to use CanDeclareGlobalFunction/CreateGlobalFunctionBinding instead of CanDeclareGlobalVar/CreateGlobalVarBinding, and look what I found -> https://bugs.ecmascript.org/show_bug.cgi?id=4428#c1. 😄 

I think @syg should take a look at this change to confirm that this matches what he implemented in SpiderMonkey.

syg:
I'm pretty sure the PR matches my implementation. Sorry I completely forgot to respond to this.

littledan:
I think TC39 was happy with this sort of thing ultimately, right?

littledan:
I believe we [got consensus](https://github.com/tc39/tc39-notes/blob/master/es8/2017-09/sep-26.md#12ik-sloppy-function-hoisting-web-reality-tweak) here. Aside from addressing code review feedback on the tests, does anything else need to be fixed up for this patch to land?

PR #889, opened 2017-04-12
Normative: Perform legacy hoisting for `function arguments() {}`

### LINKED ISSUE #815: Annex B.3.3.1 intention with `arguments` parameter shadowing
(Credit to @anba for discovering this in [Firefox's bug 1339123](https://bugzilla.mozilla.org/show_bug.cgi?id=1339123))

Consider the following two programs:

```javascript
(function (a = arguments) {
  print(arguments);
  var arguments;
}());
```

and

```javascript
(function (a = arguments) {
  print(arguments);
  { function arguments() { } }
}());
```

According to the current spec text (ES2017), both programs should print `[object Arguments]`.
According to ES2015, the second program prints `undefined`.

PR #677 from @jswalden resulted in the semantics change, though he confirmed over IRC any semantics changes from the PR were unintended.

What is the intended behavior here? I am of the opinion both programs should print `[object Arguments]`. That is, that the current spec is correct, and ES2015 had an inconsistency bug that has since been fixed.

For completeness, SpiderMonkey prints `[object Arguments]` for both. V8 prints `undefined` for both. I don't have JSC and Chakra handy, but @anba says they both print `[object Arguments]` for both tests.

littledan:
An editorial change accidentally blocked this hoisting; this patch
restores it. Corresponding tests already pass on JSC and ChakraCore.

Closes #815

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7317,7 +7317,9 @@
           1. Else,
             1. Perform ! _envRec_.CreateMutableBinding(`"arguments"`, *false*).
           1. Call _envRec_.InitializeBinding(`"arguments"`, _ao_).
-          1. Append `"arguments"` to _parameterNames_.
+          1. Let _parameterBindings_ be a new List of _parameterNames_ with `"arguments"` appended.
+        1. Else,
+          1. Let _parameterBindings_ be _parameterNames_.
         1. Let _iteratorRecord_ be Record {[[Iterator]]: CreateListIterator(_argumentsList_), [[Done]]: *false*}.
         1. If _hasDuplicates_ is *true*, then
           1. Perform ? IteratorBindingInitialization for _formals_ with _iteratorRecord_ and *undefined* as arguments.
@@ -7325,7 +7327,7 @@
           1. Perform ? IteratorBindingInitialization for _formals_ with _iteratorRecord_ and _env_ as arguments.
         1. If _hasParameterExpressions_ is *false*, then
           1. NOTE: Only a single lexical environment is needed for the parameters and top-level vars.
-          1. Let _instantiatedVarNames_ be a copy of the List _parameterNames_.
+          1. Let _instantiatedVarNames_ be a copy of the List _parameterBindings_.
           1. For each _n_ in _varNames_, do
             1. If _n_ is not an element of _instantiatedVarNames_, then
               1. Append _n_ to _instantiatedVarNames_.
@@ -7343,7 +7345,7 @@
             1. If _n_ is not an element of _instantiatedVarNames_, then
               1. Append _n_ to _instantiatedVarNames_.
               1. Perform ! _varEnvRec_.CreateMutableBinding(_n_, *false*).
-              1. If _n_ is not an element of _parameterNames_ or if _n_ is an element of _functionNames_, let _initialValue_ be *undefined*.
+              1. If _n_ is not an element of _parameterBindings_ or if _n_ is an element of _functionNames_, let _initialValue_ be *undefined*.
               1. Else,
                 1. Let _initialValue_ be ! _envRec_.GetBindingValue(_n_, *false*).
               1. Call _varEnvRec_.InitializeBinding(_n_, _initialValue_).


anba:
This will revert the semantics back to ES2015, but I think @syg suggested to keep the current semantics. https://github.com/tc39/ecma262/issues/815#issue-207380347:
> I am of the opinion both programs should print `[object Arguments]`. That is, that the current spec is correct, and ES2015 had an inconsistency bug that has since been fixed.

littledan:
@anba No strong opinion here, just misinterpreted the thread and tried to spell out the old semantics. I'd be fine with skipping this patch.

anba:
No strong opinion from me either, let's just wait for @syg.

syg:
#815 has two separate normative questions for the following test:
```javascript
(function (a = arguments) {
  P1: print(arguments);
  { function arguments() { } }
  P2: print(arguments);
}());
```
1. Should `P1` print `undefined` or `[object Arguments]`. Printing `undefined` means that `arguments` was not considered a declared parameter name, and thus the FIB synthesized a `var arguments` binding and initialized it to undefined. This is counterintuitive because if I typed out `var arguments` in a function, it would actually have the value `[object Arguments]`. Printing `[object Arguments]` means, `arguments` was considered a declared parameter name, and no Annex B synthesizing was done.

I think the sensible thing for `P1` is to print `[object Arguments]` because typing out `var arguments` does not shadow the arguments object.

2. Suppose we consider `arguments` a declared parameter name and `P1` printed `[object Arguments]`. Then that means Annex B doesn't apply at all, so `P2` continues to print `[object Arguments]`. Suppose then the opposite, that we *don't* consider `arguments` a declared parameter name and `P1` printed `undefined`. Then `P2` prints `function arguments() {}`.

I think the sensible for `P2` is actually to print `function arguments() {}`, given the mutability of a `var arguments`. So unfortunately, given the current spec text, the sensible choice for `P2` contradicts the sensible thing for `P1`.

@littledan Double check me on this, but your normative fix would result in `P1` printing `undefined` and `P2` printing `function arguments() {}`. To get the "intuitive" result of `[object Arguments]` then `function arguments() {}` would require for Annex B.3.3.1 to probably special case `arguments`, where if *argumentsObjectNeeded* is **true** and *F* is **"arguments"**, then don't synthesize a `var` binding, but do perform the assignment.

littledan:
1. I don't think anyone's pushing for returning `undefined`. V8 currently returns `undefined` due to [this separate bug](https://bugs.chromium.org/p/v8/issues/detail?id=6255). My intention was for this patch to not return `undefined` because setting stuff to undefined is done based on *varNames* which are not included in *parameterBindings*, where *parameterBindings* is the larger List that includes `arguments`.

2. I agree, if we consider `arguments` a declared parameter, then the current spec text is WAI. My intuition here was that we would treat `arguments` as "weaker than a declared parameter" and would just go away when needed rather than becoming a conflict that would block hoisting. For another example of this principle, notice how `arguments` just silently disappears if there's a lexically declared variable named `arguments`. On the other hand, there's a contravening principle that we should not do Annex B stuff when in doubt, which the ToT spec text accomplishes. So I'm OK either way.

syg:
@littledan Wait, so what do P1 and P2 print given your normative fix?

syg:
To answer my own question, my reading of the normative change is:

1. Since *hasParameterExpressions* is **true**, step 27 of FDI is taken. **instantiatedVarNames** does not include `"arguments"`.
2. Annex B.3.3.1 comes in, and, since **instantiatedVarNames** doesn't have `"arguments"`, synthesizes a shadowing var named `arguments`.
3. `P1` prints `undefined` because of 2.
4. `P2` prints the function.

Am I reading it right?

littledan:
In Annex B.3.3.1, it takes the path of, yes, getting through the hoisting but no, not creating a new binding, just as if it'd been a local var declaration in the body.

1. *instantiatedVarNames* does include `"arguments"` because `"arguments"` on *parameterBindings*, which is the version of *parameterNames* with `"arguments"` added if appropriate. Therefore, it will print `[object Arguments]`.
2. `P2` prints the function.

This stuff is really confusing!

syg:
@littledan How does `"arguments"` get into *instantiatedVarNames* in the example from https://github.com/tc39/ecma262/pull/889#issuecomment-293716798? Since that function has parameter expressions, *instantiatedVarNames* isn't initialized as a copy of *parameterBindings*.

littledan:
@syg See line 7348

syg:
@littledan Line 7348 is what chooses the initial value while looping over *varNames* with *n*. Are you saying `arguments` is in *varNames*?

littledan:
@syg Oh right... actually, the current spec text is a pretty clean solution to this issue, by treating arguments when it's created as a parameterName and blocking the hoisting. Let's leave things as they are and just improve the test262 tests for this edge case.

syg:
@littledan Well hold on, I still don't want `P1` in my example to print `undefined`. I'll draft a normative PR.

syg:
@leobalter nooooo

allenwb:
Stepping up a level, the whole point of Annex B.3.3 is backwards computability with a few specific interoperable  legacy FiB code patterns that exist on the web.  Does anyone really think that a FiB named `arguments` that is not hoisted would be an actual backwards compat issue?

If not, treating `arguments` as a declared parameter that blocks B.3.3 FiB hoisting seems like the preferable solution.

syg:
@allenwb I'd have to confirm (actually, how do I confirm?) I feel like this is a particular case where we had intersection semantics pre-ES6 across all vendors, and that Annex B **changed** this.

allenwb:
It's only legacy intersection semantics that are actually used that we need to preserve. If there is no usage then there is no need to preserve.

Does any of the current ES6+ implementations not hoist for FiBs named `arguments` or is the hoisting behavior for `arguments` different for some of the current implementations?  If either of those are conditions are true and nobody has seen any reports of code breaking because of FiBs named `arguments` then it probably  following the normal ES6+ semantics for such functions seems possible and preferable.

littledan:
Currently, four engines *will* hoist `arguments`. As a result, we don't know whether the currently specified semantics are web-compatible.

```
$ eshost -e "(function() { { function arguments() {} } return arguments; })()"
#### chakracore
function arguments() {}

#### jsc
function arguments() {}

#### spidermonkey
function arguments() {}

#### d8
function arguments() {}
```

littledan:
@syg I don't get it, where would the `undefined` come from? The `undefined` from hoisting is blocked by "and F is not an element of *parameterNames*, then", where `"arguments"` is in *parameterNames*. `"arguments"` is not an element of *varNames* (as you point out above--no one puts it there), so it also doesn't get `undefined` from that.

Who else sets `arguments` to `undefined`?

syg:
@littledan Ugh this is so confusing! I got mixed up. :( You're right, currently, it's not `undefined`. My bad.

@allenwb Yeah, the worry is that the web reality for a while has been that Annex B assignment does happen for `arguments`.

PR #891, opened 2017-04-13
Normative: Perform legacy hoisting for `function arguments() {}` 

### LINKED ISSUE #815: Annex B.3.3.1 intention with `arguments` parameter shadowing

littledan:
An editorial change accidentally blocked this hoisting; this patch
restores it. Corresponding tests already pass on JSC and ChakraCore.

Closes #815

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7317,7 +7317,9 @@
           1. Else,
             1. Perform ! _envRec_.CreateMutableBinding(`"arguments"`, *false*).
           1. Call _envRec_.InitializeBinding(`"arguments"`, _ao_).
-          1. Append `"arguments"` to _parameterNames_.
+          1. Let _parameterBindings_ be a new List of _parameterNames_ with `"arguments"` appended.
+        1. Else,
+          1. Let _parameterBindings_ be _parameterNames_.
         1. Let _iteratorRecord_ be Record {[[Iterator]]: CreateListIterator(_argumentsList_), [[Done]]: *false*}.
         1. If _hasDuplicates_ is *true*, then
           1. Perform ? IteratorBindingInitialization for _formals_ with _iteratorRecord_ and *undefined* as arguments.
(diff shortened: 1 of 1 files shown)


littledan:
Reopening this PR with a pretty ad-hoc solution to the issue raised on the previous thread #889: Annex B 3.3.1 is augmented with skipping setting arguments to undefined. It's unfortunate to have a special case here, but arguments is already so special in FunctionDeclarationInstantiation...

spec.html
```diff
@@ -12926,6 +12928,7 @@
         <li>If _base_ is *-0* and _exponent_ &lt; 0 and _exponent_ is an odd integer, the result is *-&infin;*.</li>
         <li>If _base_ is *-0* and _exponent_ &lt; 0 and _exponent_ is not an odd integer, the result is *+&infin;*.</li>
         <li>If _base_ &lt; 0 and _base_ is finite and _exponent_ is finite and _exponent_ is not an integer, the result is *NaN*.</li>
+        <li> In the remaining cases, the result is an approximation of raising _base_ to the power _exponent_, where the difference between the result and the mathematical value is less than 1 ulp.</li>
```
bterlson: This seems like an unrelated change.

littledan: Sorry for that; thanks for landing this properly.

PR #892, opened 2017-04-15
Editorial: Use alphabetical order in the list of keywords

claudepache:


claudepache:
The alphabetical order was lost sometime between ES2015 and ES2016 when the presentation switched from columns to inline.

Issue #878, opened 2017-04-08
Notes in section Arguments Exotic Objects: “arguments object for strict function” are also used for some non-strict functions


claudepache:
Notes 1, 3 and 4 in Section [Arguments Exotic Objects](https://tc39.github.io/ecma262/#sec-arguments-exotic-objects) refer to arguments object for non-strict functions and those for strict functions. They are, in fact, Mapped arguments object and Unmapped arguments object. However, the latter is created not only for strict functions, but also for non-strict functions with non-simple parameter list.

PR #893, opened 2017-04-15
Editorial: Fix Notes in section Arguments Exotic Objects

claudepache:
The Notes assumed that arguments exotic objects are those of non-strict
functions, and the non-exotic ones are those of strict functions. This
is no longer true since non-strict functions with non-simple parameter
list do not create the exotic variant.

Fixess #878


bterlson:
I've read this section so many times and yet only now realize the current text bothers me a lot.

PR #894, opened 2017-04-16
Editorial: Clarify `instanceof` runtime semantics

shvaikalesh:
The current text implies (notes and parameter names) that:
* `lhs` is *always* an object
* `rhs` is *always* a constructor function

Both are not always true with custom `@@hasInstance`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13443,15 +13443,15 @@
 
     <!-- es6num="12.9.4" -->
     <emu-clause id="sec-instanceofoperator" aoid="InstanceofOperator">
-      <h1>Runtime Semantics: InstanceofOperator ( _O_, _C_ )</h1>
+      <h1>Runtime Semantics: InstanceofOperator ( _V_, _O_ )</h1>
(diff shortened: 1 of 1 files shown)


bterlson:
I think the variable names were fine as is - there are other places where _O_ is used where a typecheck is done (rather than an assert). Plus, it's sort of a convention. It also introduces some confusion with the names switching places in OrdinaryHasInstance. In terms of editorial text, I think it's still accurate to say it's for detecting whether an object inherits from something since a non-object can't inherit from anything.

I think the big problem the "class" verbage, but I don't like the proposal here as I don't know what "inheritance chain defined by an object" means. Probably it should be reworded to something like "The abstract operation InstanceofOperator(O, C) implements the generic algorithm for determining if an object O is an instance of C either by consulting C's @@hasInstance method or, if absent, determining whether the value of C's prototype property is present in O's prototype chain."

Thoughts?

bterlson:
I said some uninformed crap previously that you might have gotten emailed to you, but I realized you were pretty much right and deleted it. I tweaked the wording proposed here a bit and went with _target_ instead of _O_ which to me is more descriptive.

Issue #897, opened 2017-04-18
Agent Clusters note 4 seems broken


annevk:
Found by @foolip at https://github.com/whatwg/html/pull/2521#discussion_r111905976.

1. It's not a DOM SharedWorker. If you want to be vague, just say "shared worker" as defined in the HTML Standard.
2. Don't say "regular worker"; it's dedicated worker.
3. Per your own advice, a shared worker cannot share memory with a dedicated worker whose owner is a document. A shared worker can only share memory with a dedicated worker "descendant".

cc @lars-t-hansen

lars-t-hansen:
I agree that note could be rephrased to be much clearer.  Points (1) and (2) are terminology and easily rephrased, though I think we might take the opportunity to do better, I'll put something up for review.  Re your point (3), the note is expressing a hypothetical and should at least be rewritten with appropriate grammar.  I'll look into that as well.

annevk:
Fixed by 0719f44aab93215ed9a626b2f45bd34f36916834.

PR #899, opened 2017-04-18
Clarify Agent Clusters note 4: Terminology and wording

lars-t-hansen:
See issue #897 for background.  This attempts to clean up both terminology and wording.  cc @annevk


annevk:
Maybe "were to share" -> "were allowed to share". That might make it a little clearer it's a hypothetical.

bterlson:
I think the subjunctive mood is sufficient to signify the hypothetical, but I'm ok adding "allowed". Will pull this soon, though.

bterlson:
Looks good. Will integrate this into 2017 as well.

Issue #837, opened 2017-03-05
FunctionDeclarationInstantiation always returns a normal completion ?


jmdyck:
Commit c90286d9 says that FunctionDeclarationInstantiation always returns a normal completion, and inserts '!' in front of two invocations of it ([here](https://tc39.github.io/ecma262/#sec-async-function-definitions-EvaluateBody) and [here](https://tc39.github.io/ecma262/#sec-async-arrow-function-definitions-EvaluateBody)).

Presumably this assertion is not true in general, since FDI has calls to IteratorBindingInitialization that are prefixed by '?'.

If the intent was merely that *those two invocations* always return normally, then:
(a) What causes that to be true?
(b) The if-then-else following each invocation could be simplified, since it tests whether the result of FDI is abrupt.

jmdyck:
Note that simply reverting that commit will leave one of the invocations with a '?' prefix, which doesn't make sense given the if-then-else that follows. I believe the proper fix is to delete both '!' prefixes.

PR #904, opened 2017-04-27
Editorial: Remove stray ! from async function spec

littledan:
FunctionDeclarationInstantion may return an abrupt completion, and the
following lines handle that abrupt competion, so there should be no !.


jmdyck:
See also issue #837.

bterlson:
Gah, I think I meant to remove the ? and got confused...

PR #905, opened 2017-04-27
Fix typology in the spec at async expressions

vTripes:
Changed the typology of the note about Async Expression (" [...] when the Await parameter in present. " to " [...] when the Await parameter is present. ")


bterlson:
@vTripes thank you and congrats on your first ECMA262 contribution!

PR #910, opened 2017-05-04
Only allow standalone reserve solidus in Annex B regular expression when immediately followed by `c`

anba:
This makes it easier to disallow invalid `\k` escapes for https://tc39.github.io/proposal-regexp-named-groups/#annex-b-grammar and also disallows `/\/` 😰 .

cc @littledan

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -647,6 +647,7 @@
           DecimalDigit [lookahead &lt;! DecimalDigit]
       </emu-grammar>
       <p>matches either the letter `n` followed by one or more decimal digits the first of which is even, or a decimal digit not followed by another decimal digit.</p>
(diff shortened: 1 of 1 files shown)


littledan:
Intuitively, I like this way of representing it much better, but I want to defer to others on really reviewing this fully since I've gotten is so wrong before. With this formulation, do you think the definition of `\k` in the named group specification works as is, or does it need other changes?

cc @schuay @hashseed

anba:
> With this formulation, do you think the definition of \k in the named group specification works as is, or does it need other changes?

It now works as expected.

bterlson:
I like this formulation. @littledan shall I wait for others to review before pulling?

littledan:
If it looks good to you, I'm happy with this being merged. It's a nice bug fix.

bterlson:
I am probably less of a regexp expert than you are so your previous deference to experts gave me pause. But, I *think* I understand everything here so I'll merge.

bterlson:
image

Issue #496, opened 2016-03-26
abstract operations don't always return Completion Records


jmdyck:
5.2 "Algorithm Conventions" says:

> Calls to abstract operations return Completion Records.

This is not always true.

For instance, consider GetFunctionRealm: when it doesn't return an abrupt completion, it returns a Realm Record. This is not a Completion Record, nor can it be implicitly converted into one (see 6.2.2.2), because a Realm Record isn't an ECMAScript language value, and so can't be the [[Value]] of a Completion Record.

Other examples:
- CreateListFromArrayLike returns a List.
- NewDeclarativeEnvironment returns a Lexical Environment.
- PrepareForOrdinaryCall returns an Execution Context.

domenic:
It seems like this is split between two cases:

1. Cases where the abstract operation could conceivably throw, so some kind of "completion record that can contain spec types" would be useful. E.g. CreateListFromArrayLike already has a number of ?s in it that could propagate exceptions, so it's presumably already returning some kind of completion record for a List (despite that supposedly being impossible).
1. Cases where we're intentionally operating on high-level things (like execution contexts or environment records), away from ES values and potential throws, but still using the abstract operation formalism. NewDeclarativeEnvironment and maybe PrepareForOrdinaryCall fall into this category.

Very interesting.

jmdyck:
> E.g. CreateListFromArrayLike already has a number of ?s in it that could propagate exceptions, so it's presumably already returning some kind of completion record for a List (despite that supposedly being impossible).

Well, that's one possible way to resolve the inconsistency. (I.e., to say that "Calls to abstract operations return Completion Records." is always true, but the description of Completion Records is incorrect/incomplete.)

But to me, it seems like less of a stretch to say that the description of Completion Records is okay, but abstract operations *don't* always return Completion Records. Consider the sentence in 6.2.2.2 (which dates back to ES6 wd rev 6):
> Unless it is otherwise obvious from the context, an algorithm statement that returns a value that is not a Completion Record, such as `Return "Infinity"` means the same thing as `Return NormalCompletion("Infinity")`.

Without that "unless" clause, it *would* mean that every algorithm returns a Completion Record, either explicitly or implicitly. But the presence of the "unless" clause implies (and certainly allows) that sometimes, an algorithm can return something else. (And my guess would be that appearing to return a List, Realm Record, Lexical Env, etc qualifies as "otherwise obvious from context", though that's not essential to the point.)

domenic:
[...]

PR #919, opened 2017-05-17
Editorial: Refactor algorithm conventions

bterlson:
* Pull in algorithm conventions from 6.2.3 around implicit completion
  values and the like
* Define production-associated algorithms
* Define runtime semantics
* Define how production associated algorithms are invoked
* Clarify that all algorithms for runtime semantics return completion
  records (including production-associated algorithms).
* Specify that production-associated algorithms may make use of ?/!
  shorthands

I've been sitting on this for a bit. I'll probably tweak it more but this is the general idea. With this copy in place we can keep fc1cbe3 in master (although I've reverted it for 2017 as I don't want to make such a huge refactoring there at this time).


bterlson:
Here's a rendered copy of the clause in question: http://bterlson.github.io/ecma262/#sec-algorithm-conventions

spec.html
```diff
@@ -709,32 +689,171 @@
...
+        <emu-alg>
+          1. Let _val_ be ! operationName().
+        </emu-alg>
+        <p>is equivalent to the following algorithm steps:</p>
+        <emu-alg>
+          1. Let _val_ be operationName().
+          1. Assert: _val_ is never an abrupt completion.
+          1. If _val_ is a Completion Record, let _val_ be _val_.[[Value]].
+        </emu-alg>
+        <p>Production-associated algorithms for runtime semantics may also make use of this short-hand by placing `!` or `?` before the algorithm's name:</p>
```
jmdyck: It should be clearer that this para is talking about *invocations* of those algorithms, and not their definitions.

Inconsistency: "shorthand" vs "short-hand".

spec.html
```diff
@@ -709,32 +689,171 @@
...
+        </emu-alg>
+        <p>Similarly, for method application style:</p>
+        <emu-alg>
+          1. ? _someValue_.operationName().
+        </emu-alg>
+        <p>is equivalent to the following algorithm:</p>
+        <emu-alg>
+          1. ReturnIfAbrupt(_someValue_.operationName()).
+        </emu-alg>
+        <p>Similarly, prefix `!` is used to indicate that an abstract operation will never return an abrupt completion and that the resulting Completion Record's value field should be used in place of the return value of the operation. For example, the algorithm:</p>
```
jmdyck: "!" indicates that *this invocation* of the AO will never return an abrupt completion. The AO might sometimes return an abrupt completion, just not when it's invoked this way, at this point.

spec.html
```diff
@@ -709,32 +689,171 @@
...
+        <emu-alg>
+          1. Let _val_ be ! operationName().
+        </emu-alg>
+        <p>is equivalent to the following algorithm steps:</p>
+        <emu-alg>
+          1. Let _val_ be operationName().
+          1. Assert: _val_ is never an abrupt completion.
+          1. If _val_ is a Completion Record, let _val_ be _val_.[[Value]].
```
jmdyck: Maybe change `let _val_ be` to `set _val_ to`

spec.html
sarahdonofrio: ouch Brian, we even went over "shorthand" today.

spec.html
```diff
@@ -709,32 +689,171 @@
...
+        <p>Production-associated algorithms for runtime semantics may also make use of this short-hand by placing `!` or `?` before the algorithm's name:</p>
+        <emu-alg>
+          1. Perform ! ProductionAssociatedAlgorithm of |NonTerminal|.
+        </emu-alg>
+      </emu-clause>
+    </emu-clause>
+    <!-- es6num="5.3" -->
+    <emu-clause id="sec-static-semantic-rules">
+      <h1>Static Semantics</h1>
+      <p>Context-free grammars are not sufficiently powerful to express all the rules that define whether a stream of input elements form a valid ECMAScript |Script| or |Module| that may be evaluated. In some situations additional rules are needed that may be expressed using either ECMAScript algorithm conventions or prose requirements. Such rules are always associated with a production of a grammar and are called the <dfn>static semantics</dfn> of the production.</p>
```
sarahdonofrio: You know, as a readability thought.... I wonder if this document would be better for my own sanity of readability to have like a grammar/terminology section update that would separate out the definition of static semantics, for example, at the top of the page, and then you can uninterruptedly talk about the grammar here. 
I realize that's a lot of effort, but as a new set of eyes jumping into this, it was hard to sort out what was grammar changes. I had to read through this a handful of times to understand that this flow was:
this is a static semantic
static semantics will have contains
this is what contains is

and then forever to realize nothing changed here :) 
Since we're more likely to change the algorithm/the algorithm syntax over the definitions/terminology, splitting this into two chunks might make it easier to PR future changes form here.

sarahdonofrio: So I guess, TL;DR would be having definitions of terminology at the beginning of the document, then just jumping into algorithms here.

spec.html
bterlson: I got it in the other places, just missed one :(

bterlson: @jmdyck definitely yes on the confusion between invocations and the algorithms themselves. Updates coming.

spec.html
bterlson: Having all references be back-references is definitely ideal (part of this change pulls some later clauses in earlier for that reason), although it's not always possible e.g. in cases of circularity. If you come across misordered things though it's an easy fix and PR to make!

I think this PR is pretty reasonable as far as forward references go.

spec.html
leobalter: I like this suggestion.

_this invocation_ or _the following invocation_. Also, trying to keep the most from the original: "... is used to indicate __the following__ abstract operation_ ..."

jmdyck:
There seems to be some slipperiness with the term 'production-associated algorithm'.

Consider, for example, http://bterlson.github.io/ecma262/#sec-identifiers-runtime-semantics-bindinginitialization . It defines 3 distinct algorithms, each associated with a production for _BindingIdentifier_. So it seems reasonable to say that the section defines 3 production-associated algorithms. But later usage suggests that the thing named 'BindingInitialization' (whose definition comprises those 3 pairs plus a few others) is itself a production-associated algorithm.

I agree that there should be a name for the kind of thing that BindingInitialization is, but (given the above) I don't think 'production-associated algorithm' is the best choice. To me, it seems appropriate to call it some kind of 'operation', since it's fairly similar to an 'abstract operation'. Apart from that, I'm not really sure. I've tried variations on:

    {production,syntax}-{based,directed,discriminated} operation

and I'm currently using "syntax-directed operation".

bterlson:
Can you expand on why BindingInitialization is different from any other, err, production-associated algorithm*? Is the problem that it's unclear whether production-associated algorithm refers to the name (of which there are many, potentially one for each production in the grammar), or a specific "instantiation"? (My intention was for it to refer to the former).

Also I think I like "operation" better than "algorithm", but I don't see how syntax-directed and production-associated are that different.

* tentative name ;)

jmdyck:
> Can you expand on why BindingInitialization is different from any other, err, production-associated algorithms*?

It isn't. It was just an example for concreteness.

> Also I think I like "operation" better than "algorithm", but I don't see how syntax-directed and production-associated are that different.

The problem with "production-associated operation" is that it suggests that the *operation* is associated with a production, whereas the association with productions happens at a lower level. (The definition of the operation comprises algorithms associated with productions.)

I tried "production-{based,directed,discriminated} operation" for a while, but eventually left "production" for "syntax", because "production" has various senses (and we're not even using the most common), but "syntax" has pretty much just one. Also, "syntax" is shorter.

Moreover, "syntax-directed" is an existing adjective in compiler theory, and I *think* this would be an appropriate use (possibly helpful, and at least non-misleading). (But if someone objects, then there's still "syntax-based" and "syntax-discriminated", which I don't think have enough baggage to mislead.)

bterlson:
I'm convinced. Maybe @samth can confirm/deny whether this is a valid usage of "syntax-directed"? Wouldn't want to upset anyone. Otherwise, "syntax-based" is fine by me.

samth:
I promise not to be upset, but syntax-directed just means that you decide entirely what to do based on the syntax of the form you're processing, as opposed to some other criteria such as "do choice 1 if it works but choice 2 if choice 1 doesn't work".

bterlson:
Unfortunately I rebased and blew away previous comments from @jmdyck but I have addressed most of those comments some way or another. At this point I'm going to pull this in and address #496 in a subsequent commit. I think I've made some big improvements and not regressed anything so seems good to go for me.

PR #923, opened 2017-05-28
collective -> collectively

bathos:
"declarations collective define only one variable" -> "declarations collectively define only one variable"


bterlson:
Thank you @bathos, and congrats on your first contribution!

PR #930, opened 2017-06-13
Editorial: Remove "do" language from memory model relations

syg:
I just noticed that @anba added the "do" language for the "For each" lines in the memory model relations back in February.

I'd like to revert the "do", because the "algorithm" is describing conditions that must be satisfied on relations, not steps to be performed. It reads really odd to me that these conditions have a "do".

I'm open to other editorial suggestions to further clarify that these are really descriptions of relations, not algorithms.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37636,13 +37636,10 @@ THH:mm:ss.sss
     <h1>Relations of Candidate Executions</h1>
     <emu-clause id="sec-agent-order" aoid="agent-order">
       <h1>agent-order</h1>
-      <p>For a candidate execution _execution_, _execution_.[[AgentOrder]] is a Relation that satisfies the following conditions.</p>
-      <emu-alg>
-        1. Let agent-order be _execution_.[[AgentOrder]].
-        1. For each pair of events _E_ and _D_ in EventSet(_execution_),
-          1. For each Agent Events Record _aer_ in _execution_.[[EventLists]],
-            1. If _E_ and _D_ are in _aer_.[[EventList]] and _E_ is before _D_ in List order of _aer_.[[EventList]], then _E_ is agent-order before _D_.
-      </emu-alg>
+      <p>For a candidate execution _execution_, _execution_.[[AgentOrder]] is a Relation on events that satisfies the following.</p>
+      <ul>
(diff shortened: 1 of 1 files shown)


bterlson:
@syg can you expand on how "do" will lead to an incorrect interpretation? I read it like "For these things, do... assert that the following is true". There is value to reducing the different looping structures and I'm not sure removing "do" has benefits above the (minimal, granted) benefit of consistency.

bterlson:
Alternatively, if you really want to avoid interpreting these as "steps" then I would not use an algorithm. Instead, consider a bulleted list ala static semantics.

syg:
@bterlson The "algorithm" for the relations in the memory model are specese versions of math definitions. For example, agent-order is something like:

∀*E*,*D* ∈ *Events*. *E* being evaluated before *D* → *E* agent-order before *D*

I read this in my head as "for all events E and D, if E is evaluated before D, then they are related by the relation 'agents-order' and E is agent-order before D". There's no notion of "for all events E and D, perform these steps."

I'll look at rewording these a la Static Semantics.

bterlson:
I feel like prose or bullets would better express these relations. Maybe seeing it will convince me otherwise though :-P

Also, does it make sense to include the mathematical notation as well? I like its terseness and precision at least.

syg:
I'd love to include the math notation, but there's no precedence for this elsewhere in the spec. You mean including it in addition instead of instead of the prose, right?

bterlson:
In addition to, yes. There's no precedent but it seems useful. Maybe we can discuss on a future PR. Are there other places that you think call for mathematical notation? Could include those as well.

jmdyck:
In early error rules, bulleted lists generally indicate a set of independent (unordered) constraints. In a conditions-to-be-satisfied 'algorithm' (as currently written), the 'steps' are not generally independent or unordered, so I don't think bullets would be appropriate.

Superficially, we could change all occurrences of ", do" to colon, which (I think) works equally well for steps-to-be-performed and conditions-to-be-satisfied.

domenic:
+1 for getting rid of ", do". It also mismatches web specs.

syg:
@jmdyck @domenic It used to be ":", but @anba changed it to ", do".

jmdyck:
The Memory Model addition (25b02ed) used ':' on "For each" lines, but that was inconsistent with (almost) all other occurrences of "For each" in the spec. @anba (in 0802fa6c) resolved that inconsistency (and a few others) by changing the minority to match the majority. I'm suggesting the reverse, roughly speaking.

syg:
@bterlson and others, what do you think of the bullet point style in 4d77ec3?

bterlson:
Well I like this so I'm merging.

PR #932, opened 2017-06-13
Normative: Clarify that case conversion is context-sensitve

littledan:
Previously, String.prototype.toUpperCase and
String.prototype.toLowerCase were specified with wording indicating
that Unicode case conversion was to be used, but code points should
be looked up one by one. The Unicode Default Case Conversion algorithm
is context-dependent, and a note alludes to this, but the normative
text seems to prohibit using the context. However, JSC, V8 and
ChakraCore do seem to pay attention to this context, at least when
it comes to final sigma.

This patch clarifies the spec text to make an explicit reference
to the Default Case Conversion algorithm.

These semantics are already tested by @anba's test262 test https://github.com/tc39/test262/blob/master/test/built-ins/String/prototype/toLowerCase/special_casing_conditional.js

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -28417,7 +28417,7 @@ THH:mm:ss.sss
       <emu-clause id="sec-string.prototype.touppercase">
         <h1>String.prototype.toUpperCase ( )</h1>
         <p>This function interprets a String value as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>.</p>
-        <p>This function behaves in exactly the same way as `String.prototype.toLowerCase`, except that the String is mapped using toe toUppercase algorithm of Unicode Default Case Conversion.</p>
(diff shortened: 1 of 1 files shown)


bterlson:
looks good (minus a misplaced toe).

spec.html
```diff
@@ -28419,7 +28417,7 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
       <emu-clause id="sec-string.prototype.touppercase">
         <h1>String.prototype.toUpperCase ( )</h1>
         <p>This function interprets a String value as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>.</p>
-        <p>This function behaves in exactly the same way as `String.prototype.toLowerCase`, except that code points are mapped to their _uppercase_ equivalents as specified in the Unicode Character Database.</p>
+        <p>This function behaves in exactly the same way as `String.prototype.toLowerCase`, except that the String is mapped using toe toUppercase algorithm of Unicode Default Case Conversion.</p>
```
michaelficarra: FYI, `toe` is on this line.

PR #946, opened 2017-07-05
Correct an axiom of strict total order

unoduetre:
Dear Ecma-262 team,

I have noticed the following simple omission in the specification. The first axiom of the strict total order does not work when a = b. I have applied the necessary change. Please, let me know what other steps are necessary for this change to be applied. As this is a trivial change, I do not think it is necessary to pass any copyrights, but I am very happy to do so.

Yours faithfully,
Mateusz Grotek

For and on behalf of IdeaPiece Ltd.


bterlson:
Awesome, thank you much @unoduetre. Also congrats on your first time contribution!

PR #947, opened 2017-07-10
Editorial: Refer to proleptic Gregorian calendar.

apaprocki:
The spec currently refers to an "extrapolated Gregorian system",
but the term "proleptic Gregorian calendar" is more precise and is
the terminology used elsewhere in other standards and languages.
(e.g. ISO 8601:2004 3.2.1 "The use of this calendar for dates
preceding the introduction of the Gregorian calendar (also called
the proleptic Gregorian calendar) should only be by agreement of
the partners in information exchange.")

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -26225,7 +26225,7 @@
       <!-- es6num="20.3.1.3" -->
       <emu-clause id="sec-year-number">
         <h1>Year Number</h1>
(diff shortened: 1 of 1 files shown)


bterlson:
I'd say this is editorial since it's saying the same thing just with standard terminology. Otherwise, looks good.

apaprocki:
@bterlson no prob, updated the commit/PR title

bterlson:
@apaprocki thank you sir!

PR #951, opened 2017-07-12
Editorial: fix the logo to not be humongous

domenic:
This brings the logo in line with other standards organizations' logos on their specs, e.g. WHATWG (100px) or W3C (72px).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -21,14 +21,14 @@
   margin: 0; padding: 0;
 }
 
-h1.version {
-  margin-right: 100px; /* width of logo */
+h1.version, h1.title {
+  margin-right: 350px; /* width of logo */
 }
 #ecma-logo {
   position: absolute;
   right: 0;
   top: 0;
-  width: 100px;
+  width: 350px;
 }
 </style>
 <script>
--- a/spec.html
+++ b/spec.html
@@ -21,7 +21,10 @@
   margin: 0; padding: 0;
 }
 #ecma-logo {
-  width: 100%;
+  position: absolute;
+  right: 0;
+  top: 0;
+  width: 100px;
 }
 </style>
 <script>
--- a/spec.html
+++ b/spec.html
@@ -21,7 +21,7 @@
   margin: 0; padding: 0;
 }
 
-h1.title {
+h1.version {
   margin-right: 100px; /* width of logo */
 }
 #ecma-logo {
--- a/spec.html
+++ b/spec.html
@@ -20,6 +20,10 @@
   list-style-type: none;
   margin: 0; padding: 0;
 }
+
+h1.title {
+  margin-right: 100px; /* width of logo */
+}
 #ecma-logo {
   position: absolute;
   right: 0;


bterlson:
It covers the header when it doesn't break across lines.

domenic:
I thought it looked nice that way, but your wish is my command.

tabatkins:
Float it, yo.

domenic:
Before: Giant logo

After: Small logo

bterlson:
If you want the logo smaller I think just removing the "width: 100%" is sufficient. This proposal's logo is much too small.

domenic:
Can you explain your reasoning? If we remove it, then its width is 496px, and it takes up vertical space. That is ~5x wider than other specifications/standards orgs, and none of them take up vertical space.

bterlson:
Aesthetics are not reasonable :-P That said, I don't understand any arguments referencing "vertical space" when we're talking about a document that is a million pixels tall that no one reads from the top anyway.

domenic:
It's about a variety of factors, such as first impressions, flow of text, etc. As you said, aesthetics. In the absence of actual designers doing this work, I think it's best to go with precedent from other specifications, at least on contentious issues like this apparently is.

bterlson:
This is not contentious in the normal sense, everyone just loves to provide design feedback. And fwiw, I'm not the only one that likes the current logo size. If you'd like to provide evidence (not appeals to authority) that your aesthetics are right I'll listen, but in the absence of that let's work together on a compromise (such as leaving it inline and making it smaller).

domenic:
I don't think it's fair to characterize a desire to emulate established best practices as an appeal to authority.

I'd like to see the size of the logo reduced to industry standard and for it to not take up vertical space (as is industry standard). Is there any way we can reach an agreement on something within those parameters? If not, what is the escalation path?

domenic:
Would it count as evidence, or an appeal to authority, to survey the logo size on non-homepages of, say, the Alexa top 20?

michaelficarra:
FWIW, I love the gigantic logo. :heart:

bterlson:
Sorry, to clarify "appeal to authority" (probably not the right term anyway), I assumed that the existing specifications you refer to were likely not designed by professionals either and merely reflects the aesthetics of their respective editors. If this isn't the case I'd love to see any relevant discussion or design documents.

Re: Alexa top 20, I don't think such websites are relevant to this discussion. Maybe others would find it helpful?

I don't know why you're jumping to escalation. I suggest working with me rather than going around me. But if you insist, you can add this to the July agenda and I promise to follow committee consensus (and abstain from blocking).

tabatkins:
Current W3C design is a proper effort by a designer (specifically, Divya Manian) to make the specs look better.

domenic:
> were likely not designed by professionals either

My understanding is that the latest W3C stylesheet was designed by a professional, as was [the old developers.whatwg.org](http://web.archive.org/web/20170603005743/https://developers.whatwg.org/) (78px). Does that help?

> I don't know why you're jumping to escalation. I suggest working with me rather than going around me. But if you insist, you can add this to the July agenda and I promise to follow committee consensus (and abstain from blocking).

Well, I am happy to work with you, if it is possible to work on a solution within my parameters. Otherwise it seems like there are irreconcilable differences. Which is the case?

bterlson:
@tabatkins @domenic is there design rationale for the logo placement and size or is it reflective of the designer's aesthetics?

@domenic re: resolving differences, I'm looking for anything concrete relevant to specification documents and so far all I see are various subjective opinions. I'm even happy to work compromise on subjective opinion (e.g. smaller) but it seems your parameters are "my PR or nothing". If that's true, they'd seem irreconcilable unless I see a clear consensus behind your proposal.

domenic:
> is there design rationale for the logo placement and size or is it reflective of the designer's aesthetics?

That's hard for me to say. The fact that it's so universal implies to me maybe it's something taught somewhere, but a quick Google search did not turn it up. My offer of an Alexa top 20 search was meant to help make it clear how universal this is.

> I'm even happy to work compromise on subjective opinion (e.g. smaller) but it seems your parameters are "my PR or nothing". If that's true, they'd seem irreconcilable unless I see a clear consensus behind your proposal.

I don't think that's fair. I have specific parameters. Outside of that I'm happy to work on a compromise; e.g. I already changed the interaction with the h1.title at your request. I'm happy with any PR that works within those parameters.

What I'm trying to pin down are your parameters. I assume they include "there must be a logo". But I'm unsure if they include "the logo must be >=496px wide" or "the logo must take up vertical space", in which case indeed we have irreconcilable differences. But you haven't stated your parameters yet so it's not clear where we're at.

bterlson:
@domenic Google's logo (at least on many days) is bigger than the "smaller" proposal above, so presumably we're in-line with Google's own design guidelines here.

My only requirements are that the logo should not be minuscule (which also requires it to be inline I think, otherwise it takes up header room) and in general that the design looks good based on my sense of aesthetics.

domenic:
> Google's logo (at least on many days) is bigger than the "smaller" proposal above, so presumably we're in-line with Google's own design guidelines here.

Not on pages besides the homepage (google.com / ecma-international.org). E.g. on https://www.google.com/intl/en/policies/privacy/?fg=1 it is 92px by 33px.

Another comparison is w3.org / whatwg.org.

> My only requirements are that the logo should not be minuscule (which also requires it to be inline I think, otherwise it takes up header room) and in general that the design looks good based on my sense of aesthetics.

Can you be more precise about miniscule? I'm trying to get a sense of whether you think W3C/WHATWG/Google logos are miniscule.

I think I've shown an existence proof that the logo does not need to be inline, so I don't understand that point.

The aesthetics requirement we can leave until later, I guess.

bterlson:
W3C's is on the small side, but is fine since the logo contains 3 letters and so is readable. Whereas with your proposal, "International" is illegible for me.

There is existence proof on either side in this discussion.

domenic:
I'm confused. I've proposed 100px in this PR, which is bigger than the 72px W3C which you claim is OK. Is 100px OK for you, or no? Can you state at which point "International" becomes legible by your standards?

bterlson:
How about at a bare minimum no smaller than body copy text? (But probably bigger because its contrast is reduced). Looks to be ~350px wide on my screen.

domenic:
3x the size of other standards organizations' or companies' logos is not within my parameters, but I am happy to go with that for now and we can discuss further reduction in committee, if that is truly your minimum. I've updated the PR to reflect that so hopefully it can be merged as an intermediate step.

bterlson:
I've tried this and I don't like it (esp. on small screens). Inline is better IMO. I like the "natural" size better too because it lines up nicely with the header text when it breaks across two lines.

I'll commit the smaller size + inline, though my offer to defer to committee consensus still stands.

domenic:
Added an agenda item: https://github.com/tc39/agendas/commit/eec4a00601208747a0e351423830477f04553d4c

RyanCavanaugh:
Please post meeting notes. https://xkcd.com/1167/

PR #953, opened 2017-07-18
consistently use "is present" over "was supplied", "was provided", and other alternatives

michaelficarra:
There were 4 occurrences of "was provided" and 14 occurrences of "was supplied", so I went with the latter.

Unfortunately, there's still 23 occurrences of "is provided" and 13 occurrences of "is supplied". 😕 Not sure what to do there.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3257,7 +3257,7 @@
       <emu-alg>
         1. Assert: _input_ is an ECMAScript language value.
         1. If Type(_input_) is Object, then
-          1. If _PreferredType_ was not passed, let _hint_ be `"default"`.
+          1. If _PreferredType_ is not present, let _hint_ be `"default"`.
           1. Else if _PreferredType_ is hint String, let _hint_ be `"string"`.
           1. Else _PreferredType_ is hint Number, let _hint_ be `"number"`.
           1. Let _exoticToPrim_ be ? GetMethod(_input_, @@toPrimitive).
@@ -4443,7 +4443,7 @@
       <h1>Call ( _F_, _V_ [ , _argumentsList_ ] )</h1>
       <p>The abstract operation Call is used to call the [[Call]] internal method of a function object. The operation is called with arguments _F_, _V_, and optionally _argumentsList_ where _F_ is the function object, _V_ is an ECMAScript language value that is the *this* value of the [[Call]], and _argumentsList_ is the value passed to the corresponding argument of the internal method. If _argumentsList_ is not present, a new empty List is used as its value. This abstract operation performs the following steps:</p>
       <emu-alg>
-        1. If _argumentsList_ was not passed, set _argumentsList_ to a new empty List.
+        1. If _argumentsList_ is not present, set _argumentsList_ to a new empty List.
         1. If IsCallable(_F_) is *false*, throw a *TypeError* exception.
         1. Return ? _F_.[[Call]](_V_, _argumentsList_).
       </emu-alg>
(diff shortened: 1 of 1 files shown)


ljharb:
I think we should make them all be "provided" or "supplied".

Since there's 27 occurrences of each, I think we just get to make a decision. "Provided" feels more correct to me, but I don't feel strongly.

mathiasbynens:
Slight preference towards “provided” here as well.

jmdyck:
The spec also uses "passed" in many of the same contexts as "provided" and "supplied".

Also "specified" (e.g. `If _month_ is not specified` in `Date.prototype.setFullYear`).

And occasionally "present" (e.g., `If _initialValue_ is present` in `Array.prototype.reduce`).

In the programming world in general, when you're talking about a function call, I think "pass" is the verb most commonly used to describe what's happening with the arguments -- the arguments are being passed to the function. So here, I'd probably prefer "passed".

(And if anyone cares about tense ["is passed" vs. "was passed"], I'd prefer past tense within the body of an algorithm. In the preamble to an algorithm, either could work.)

Conceivably, we could distinguish between (i.e., choose to use different verbs for) spec operations and built-in functions, since the parameter-passing semantics are different, but I'm not sure that would be useful.

michaelficarra:
"is present" is used 73 times (some of which refer to properties, but I think that's okay). I strongly prefer this declarative description to all of the above forms. I'll update the PR to normalise on "is present".

michaelficarra:
Alright, done. I've left some occurrences of the other forms in notes and other prose. All occurences in algorithms use the "is present" terminology.

bterlson:
Very nice!

jmdyck:
[I started this comment before the PR was merged, hence some no-longer-appropriate tenses.] 

Currently, the spec has about 148 occurrences of 'present' as an adjective (e.g. `(is|are|were) [not] present`). I've divided them in 13 usages:

     67 optional child of node
     33 parameter of built-in function
     16 field in property descriptor
      8 parameter of abstract op
      8 element in array
      4 grammar param in production
      3 property in object
      2 substring in string
      2 object in list
      2 error in source text
      1 value in set object
      1 kv pair in map object
      1 debugger in env

This PR would increase the counts for `parameter of built-in function` and `parameter of abstract op`.

The spec has about 113 occurrences of "passed". All but one are talking about passing some kind of value to some kind of function-y thing:

     79 passed to an ECMAScript function
     31 passed to a spec operation (abstract op or syntax-directed op)
      2 passed to an internal method
      1 "hours that have passed since midnight"

Note that this PR changes only about 17 of these, so there would still be lots of uses of "passed".

So: we have a distinct concept (parameter-passing) for which we're already using a distinct term ("passed"), which is used for almost no other purpose. Given that, it seems counter-productive to change some of those uses to a term ("present") which is used for several other purposes.

We'd be better off going in the other direction: changing the parameter-passing uses of "present" (and "supplied" and "provided" and "specified") to "passed".

bterlson:
@jmdyck you make a convincing argument in favor of aligning on "passed" when talking about parameters as it leaves "present" for the other cases.

@michaelficarra would you be opposed to going in the other direction? Would also imply using past tense I think.

michaelficarra:
@bterlson The reason I felt "is present" was so obviously correct is that it is talking about the current state of things in a declarative manner. Using the "was passed" terminology makes me think about how we got to this point instead of where we are right now. I think it's fine that we talk about how values are "passed" to functions and spec operations, but once inside them, talk about the state of the world as can be seen from that invocation: "is present" or more informally "I have a".

**tl;dr** It's not the end of the world, but I still think the current text (as of merging this PR) is best.

jmdyck:
I'm not sure "declarative" is important, as the spec algorithms are fairly non-declarative, but I do have some sympathy for using terminology that talks about the current state of things, rather than what happened in the past. Even so, I personally don't think `is present` is a good choice. To my way of thinking, the parameter _per se_ is always 'present' (it's 'in scope') -- instead, the question is whether it has a value bound to it (had a value bound to it at invocation time). So, say, `is bound` or `is set` would be better on that score, though they might be somewhat problematic for other reasons.

More pragmatically, I doubt we're going to get away from using "what happened in the past" terminology, e.g. in `the number of arguments passed to this function call`.

But I'm okay to leave it as-is (for now, at least), and see if anyone else has an opinion.

michaelficarra:
> More pragmatically, I doubt we're going to get away from using "what happened in the past" terminology, e.g. in `the number of arguments passed to this function call`.

I think "the number of realised parameters" is a declarative description that works for that case.

jmdyck:
So then would you say `If _param_ is realized, ...` ?

Issue #311, opened 2016-01-22
Add a notation for static and runtime semantics attached to grammar productions.


claudepache:
It concerns the phrases of the form "the result of evaluating _production_ with argument _arg_", and "the result of performing Foo of _production_ with argument _arg_", and variations thereof.

Possible notation (to be bikeshed):

>  Evaluation{_production_}(_args_)
> Foo{_production_}(_args_)

etc.

That will allow to replace the majority of ReturnIfAbrupt steps in the spec with the `?` notation.

Issue of my proposal: we lose the editorial distinction between the uses of runtime semantics (marked with the action verbs "performing" and "evaluating") and the use of static semantics.

Examples from section [12.2.6.9](http://tc39.github.io/ecma262/#sec-object-initializer-runtime-semantics-propertydefinitionevaluation):

Replace:

> 1. Let _status_ be the result of performing PropertyDefinitionEvaluation of _PropertyDefinitionList_ with arguments _object_ and _enumerable_.
> 2. ReturnIfAbrupt(status).

with:

> 1. Perform ? PropertyDefinitionEvaluation{_PropertyDefinitionList_}(_object_, _enumerable_).

Replace:

> 1. Let propName be StringValue of _IdentifierReference_.
> 2. Let exprValue be the result of evaluating _IdentifierReference_.
> 3. Let propValue be ? GetValue(exprValue).

with:

> 1. Let propName be StringValue{_IdentifierReference_}().
> 2. Let exprValue be ? Evaluation{_IdentifierReference_}().
> 3. Let propValue be ? GetValue(exprValue).

Replace:

> 1. Let _propKey_ be the result of evaluating _PropertyName_.
> 2. ReturnIfAbrupt(_propKey_).
> 3. Let _exprValueRef_ be the result of evaluating _AssignmentExpression_.
> 4. Let _propValue_ be ? GetValue(_exprValueRef_).

with:

> 1. Let _propKey_ be ? Evaluate{_PropertyName_}().
> 3. Let _exprValueRef_ be ? Evaluate{_AssignmentExpression_}().
> 4. Let _propValue_ be ? GetValue(_exprValueRef_).

claudepache:
> Issue of my proposal: we lose the editorial distinction between the uses of runtime semantics (marked with the action verbs "performing" and "evaluating") and the use of static semantics.

Operations done at runtime should probably always be prefixed by either `?` or `!`; that will restore that distinction.

bterlson:
[...]

PR #955, opened 2017-07-24
Add aoid and update calling conventions for directed operations

bterlson:
This is the ECMA262 changes necessary to use the upcoming support for polymorphic references. I've got a rendering of this here: https://bterlson.github.io/ecma262/#sec-declarative-environment-records. Sending this in case there are any thoughts on this approach before I convert the entire document and commit the ecmarkup change as well.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7789,7 +7789,7 @@
         1. For each _d_ in _varDeclarations_, in reverse list order, do
           1. If _d_ is neither a |VariableDeclaration| nor a |ForBinding| nor a |BindingIdentifier|, then
             1. Assert: _d_ is either a |FunctionDeclaration|, a |GeneratorDeclaration|, an |AsyncFunctionDeclaration|, or an |AsyncGeneratorDeclaration|.
-            1. Let _fn_ be the sole element of the BoundNames of _d_.
+            1. Let _fn_ be the sole element of the _d_.BoundNames().
             1. If _fn_ is not an element of _functionNames_, then
               1. Insert _fn_ as the first element of _functionNames_.
               1. NOTE: If there are multiple function declarations for the same name, the last declaration is used.
@@ -15509,7 +15509,7 @@
             1. Else,
               1. Perform ! _envRec_.CreateMutableBinding(_dn_, *false*).
           1. If _d_ is a |FunctionDeclaration|, a |GeneratorDeclaration|, an |AsyncFunctionDeclaration|, or an |AsyncGeneratorDeclaration|, then
-            1. Let _fn_ be the sole element of the BoundNames of _d_.
+            1. Let _fn_ be the sole element of the _d_.BoundNames().
             1. Let _fo_ be the result of performing InstantiateFunctionObject for _d_ with argument _env_.
             1. Perform _envRec_.InitializeBinding(_fn_, _fo_).
       </emu-alg>
@@ -16038,7 +16038,7 @@
 
         <emu-grammar>BindingProperty : SingleNameBinding</emu-grammar>
         <emu-alg>
-          1. Let _name_ be the string that is the only element of BoundNames of |SingleNameBinding|.
+          1. Let _name_ be the string that is the only element of |SingleNameBinding|.BoundNames().
           1. Perform ? KeyedBindingInitialization for |SingleNameBinding| using _value_, _environment_, and _name_ as the arguments.
           1. Return a new List containing _name_.
         </emu-alg>
@@ -17189,7 +17189,7 @@
           IterationStatement : `for` `await` `(` ForDeclaration `of` AssignmentExpression `)` Statement
         </emu-grammar>
         <emu-alg>
-          1. Let _keyResult_ be the result of performing ? ForIn/OfHeadEvaluation(BoundNames of |ForDeclaration|, |AssignmentExpression|, ~async-iterate~).
+          1. Let _keyResult_ be the result of performing ? ForIn/OfHeadEvaluation(|ForDeclaration|.BoundNames(), |AssignmentExpression|, ~async-iterate~).
           1. Return ? ForIn/OfBodyEvaluation(|ForDeclaration|, |Statement|, _keyResult_, ~iterate~, ~lexicalBinding~, _labelSet_, ~async~).
         </emu-alg>
         <emu-note>
@@ -19909,7 +19909,7 @@
         <li>It is a Syntax Error if |UniqueFormalParameters| Contains |YieldExpression| is *true*.</li>
         <li>It is a Syntax Error if |UniqueFormalParameters| Contains |AwaitExpression| is *true*.</li>
         <li>It is a Syntax Error if ContainsUseStrict of |AsyncGeneratorBody| is *true* and IsSimpleParameterList of |UniqueFormalParameters| is *false*.</li>
-        <li>It is a Syntax Error if any element of the BoundNames of |UniqueFormalParameters| also occurs in the LexicallyDeclaredNames of |AsyncGeneratorBody|.</li>
+        <li>It is a Syntax Error if any element of the |UniqueFormalParameters|.BoundNames() also occurs in the LexicallyDeclaredNames of |AsyncGeneratorBody|.</li>
       </ul>
       <emu-grammar>
         AsyncGeneratorDeclaration : `async` `function` `*` BindingIdentifier `(` FormalParameters `)` `{` AsyncGeneratorBody `}`
@@ -19922,7 +19922,7 @@
         <li>If the source code matching this production is strict mode code, the Early Error rules for <emu-grammar>UniqueFormalParameters : FormalParameters</emu-grammar> are applied.</li>
         <li>If the source code matching this production is strict mode code, it is a Syntax Error if |BindingIdentifier| is the |IdentifierName| `eval` or the |IdentifierName| `arguments`.</li>
         <li>It is a Syntax Error if ContainsUseStrict of |AsyncGeneratorBody| is *true* and IsSimpleParameterList of |FormalParameters| is *false*.</li>
-        <li>It is a Syntax Error if any element of the BoundNames of |FormalParameters| also occurs in the LexicallyDeclaredNames of |AsyncGeneratorBody|.</li>
+        <li>It is a Syntax Error if any element of the |FormalParameters|.BoundNames() also occurs in the LexicallyDeclaredNames of |AsyncGeneratorBody|.</li>
         <li>It is a Syntax Error if |FormalParameters| Contains |YieldExpression| is *true*.</li>
         <li>It is a Syntax Error if |FormalParameters| Contains |AwaitExpression| is *true*.</li>
         <li>It is a Syntax Error if |FormalParameters| Contains |SuperProperty| is *true*.</li>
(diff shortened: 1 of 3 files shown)


bterlson:
@jmdyck may have impacts on #545.

jmdyck:
Is StringValue singled out because it's special, or is it just the first of many?

bterlson:
First of many.

jmdyck:
Is "polymorphic references" explained somewhere?

jmdyck:
(Also, the change to calling conventions relates back to open issue #311.)

bterlson:
"polymorphic references" is not an ECMA262 term but how I call them for Ecmarkup purposes. For ECMA262, I will be updating the algorithm conventions to include these conventions both for syntax-directed operations and type-directed operations (where type might be any ECMAScript spec or language value).

jmdyck:
> "polymorphic references" is not an ECMA262 term but how I call them for Ecmarkup purposes.

The word "references" made me wonder if they had something to do with the Reference specification type. Instead, it sounds like you're unifying (or at least bring closer together) the concepts of "concrete methods" (for Environment Records + Module Records) and syntax-directed operations. What they have in common (so far) is:
1. Each such operation has, in effect, an implicit additional parameter, which is referenced (within the op's definition) and set (within an invocation of the op) in various ways (but never in the same way as a conventional parameter).
2. Each such operation is defined by multiple separate algorithms, and which one to use (for any given invocation) is determined by the nature of the value being passed to the implicit parameter.

You could almost fit internal methods into the same scheme. Were you intending to do so?

bterlson:
Yeah my plan is to do this with internal methods as well. Can you elaborate on "almost"? 😛

My thinking in terms of algorithm conventions updates is that "directed operations" [editorial note: awkward name] are algorithms that take an implicit parameter of the receiver (but provides no way to access it). Syntax directed operations expose the receiver as parse nodes named by the receiver's non-terminals, internal methods expose it as an object _O_, and "concrete methods" (record-directed operations?) have the "Let thing be this Thing" convention.

My only concern is that "directed operations" is awkward...

jmdyck:
> Yeah my plan is to do this with internal methods as well.

Ah, then you'll need to be careful about the word "receiver", as `[[Get]]` and `[[Set]]` already have an explicit parameter by that name.

> Can you elaborate on "almost"?

I just meant that with 'concrete methods' and syntax-directed operations, the dispatching is based on an inherent/intrinsic aspect of the 'receiver', whereas with internal methods, the dispatch is just to execute whatever's in the appropriate internal slot of the 'receiver' (which isn't necessarily different in an implementation, but is different in the spec's model).  But I didn't mean that that should stop you from putting them all into the same general category.

>  Syntax directed operations expose the receiver as parse nodes named by the receiver's non-terminals,

I would say that for an SDO, the receiver *is* a parse node. (More specifically, within an algorithm for an SDO, the receiver is an instance of one of the productions that that algorithm is associated with, and it is referenced via the form `this |NT|` where `|NT|` is the lhs-nonterminal of (one of) those production(s). (In the rare case that the alg is associated with multiple productions with different lhs-nonterminals, it turns out the algorithm doesn't need to refer to the receiver.)

> internal methods expose it as an object O

Note that currently they don't all use the name `_O_` (although you could make that change, of course). E.g., [[DefineOwnProperty]] uses `_A_` for Array exotic object, `_S_` for String exotic object, and `_args_` (or `the arguments object`) for arguments exotic object.

> and "concrete methods" (record-directed operations?) have the "Let thing be this Thing" convention.

Currently, some also use the form `Let _envRec_ be the [foo] Environment Record for which the method was invoked.` (So those could be made consistent.)

> My only concern is that "directed operations" is awkward...

I don't think it would be incorrect to call them "polymorphic operations", but I'm not sure it would be clear. Maybe "dispatched operations"? "overloaded operations"?

bterlson:
> Ah, then you'll need to be careful about the word "receiver", as [[Get]] and [[Set]] already have an explicit parameter by that name.

Hmm, can you think of a different name?

> I would say that for an SDO, the receiver is a parse node.

I wanted to say this as well, but to further point out that the receiver is effectively destructured into various aliases the algorithm can access. I don't think the `this |NT|` convention is super necessary for SDOs, I like our current approach of having the NTs in scope.

> Currently, some also use the form Let _envRec_ be the [foo] ...

I want to get rid of these instances - the convention for module records is superior for terseness reasons :)

jmdyck:
>> Ah, then you'll need to be careful about the word "receiver", as [[Get]] and [[Set]] already have an explicit parameter by that name.

> Hmm, can you think of a different name?

Nothing great so far.

You might instead rename [[Get]] and [[Set]]'s `_Receiver_` parameter. Its description says:
>  `If any ECMAScript code must be executed to [verb] the property value, _Receiver_ is used as the *this* value when evaluating the code.`

So you could rename it to (say) `_contingentThisValue_`. (Probably not that, but something along those lines.)

>> I would say that for an SDO, the receiver is a parse node.

> I wanted to say this as well, but to further point out that the receiver is effectively destructured into various aliases the algorithm can access.

I definitely wouldn't use that wording (because it doesn't feel like destructuring to me, and we're not talking about aliases in the sense that's been defined), but we can certainly talk about the forms that an SDO algorithm can use to access the children of the parse node that is the receiver.

> I don't think the `this |NT|` convention is super necessary for SDOs,

What would you change such uses to?

> I like our current approach of having the NTs in scope.

Did you think I was suggesting we do away with that? (I wasn't.)

spec.html
```diff
@@ -16743,7 +16743,7 @@ <h1 id="ao-issuperreference">IsSuperReference ( _V_ )</h1>
       </emu-alg>
       <emu-grammar>ContinueStatement : `continue` LabelIdentifier `;`</emu-grammar>
       <emu-alg>
-        1. If the StringValue of |LabelIdentifier| is not an element of _iterationSet_, return *true*.
+        1. If the |LabelIdentifier|.StringValue() is not an element of _iterationSet_, return *true*.
```
claudepache: Probably “If |LabelIdentifier|.StringValue()...” (extraneous “the”).

The current spec has “StringValue of ...” in some places and “the StringValue of ...” in other places.

jmdyck:
An alternative for "receiver" might be "subject".

ljharb:
fwiw, "receiver" is an established term for the "this value" of a function; it'd be ideal to stick with that definition.

jmdyck:
Just wondering what the status of this PR is.

claudepache:
ping?

spec.html
```diff
@@ -751,29 +751,38 @@ <h1>Algorithm Conventions</h1>
...
-        1. Return the result of evaluating |StatementList|.
-      </emu-alg>
+    <emu-clause id=sec-algorithm-conventions-directed-operations>
+      <h1>Directed Operations</h1>
+      <p>Some abstract operations, called <dfn>directed operations</dfn>, are implicitly parameterized with a target and have multiple definitions that may be selected depending on the value of the target. The target is referenced by callers of a directed operation using the method-like reference form `target.Operation()`. The target value may be referenced by the algorithm in a manner defined for that type of directed operation.</p>
+      
+
+      <emu-clause id=sec-algorithm-conventions-object-directed-operations>
+        <h1>Object-Directed Operations</h1>
+        <p>Blah blah blah</p>
```
ljharb: ?

spec.html
```diff
@@ -751,29 +751,38 @@ <h1>Algorithm Conventions</h1>
...
-      </emu-grammar>
-      <p>but the Evaluation operation does not associate an algorithm with that production. In that case, the Evaluation operation implicitly includes an association of the form:</p>
-      <p><b>Runtime Semantics: Evaluation</b></p>
-      <emu-grammar type="example">Block : `{` StatementList `}`</emu-grammar>
-      <emu-alg>
-        1. Return the result of evaluating |StatementList|.
-      </emu-alg>
+    <emu-clause id=sec-algorithm-conventions-directed-operations>
+      <h1>Directed Operations</h1>
+      <p>Some abstract operations, called <dfn>directed operations</dfn>, are implicitly parameterized with a target and have multiple definitions that may be selected depending on the value of the target. The target is referenced by callers of a directed operation using the method-like reference form `target.Operation()`. The target value may be referenced by the algorithm in a manner defined for that type of directed operation.</p>
```
ljharb: tbh using something like the bind operator might work better, like `target::Operation()`, if we had that available

jmdyck: The spec is already using the syntax `target.Operation()`, and this PR generalizes its use (i.e., extends it in a compatible way).

Conversely, PR #1515 (BigInt) is proposing a `target::Operation()` syntax with different semantics from what you're suggesting. (In that PR, `target` is a *type* rather than a *value*.)

So why do you think that the `::` syntax might work better for this PR?

ljharb: I wasn’t making it as a concrete suggestion, since the bind operator isn’t a thing yet - more saying that using a dot, for a method that isn’t attached to the receiver, is confusingly non javascripty- so using an operator designed for this purpose seems nicer.

jmdyck: I'm not really following you. Can you clarify what you mean by "attached to"? E.g., give cases where you think an abstract method is/isn't attached to a spec value?

ljharb: It’s entirely possible I’m misreading this PR. It seemed to me that a standalone function was being called with an object as the implicit receiver, but perhaps these aren’t standalone functions.

jmdyck: They aren't functions at all, they're abstract operations. If that's what you meant, then I'm not sure what you mean by "standalone". Can you define it in spec terms, or give examples of standalone and non-standalone operations?

ljharb: An abstract operation is a standalone function - it isn't attached to any object. It doesn't have a receiver, just like a strict-mode non-bound JS function invoked with parens without using member access doesn't.

jmdyck: So I think you're saying that you would would like to limit the spec notation:


To clarify, if `envRec` is some kind of environment record, and some alg step invokes `envRec.HasBinding(N)`, would you say that the `HasBinding` method:
- is standalone?
- is attached to `envRec`? 
- has a receiver?

jmdyck: So I think you're saying that you'd prefer to limit the spec notation `X.Foo(args)` to cases where `Foo` is defined to be a component / attribute / "property" of `X`, and use something other that dot (e.g., `::`) when it isn't. This PR introduces notation like `|RelationalExpression|.Evaluate()`, but a syntax directed operation (like `Evaluate`) is not a component of a Parse Node, so you'd prefer the non-dot punctuation instead.

ljharb: Yes, that sounds right :-)

jmdyck: Okay. So, applying that to existing constructs would go like this?:
- `_object_.[[InternalMethod]](...)` would stick with the dot (because the spec tends to talk about internal methods as if they're "attached" to the object).
- `_envRec_.EnvRecMethod(...)` would switch to the non-dot punctuation (because specific env record methods seem more like they belong to a particular env rec *type*).
- `_moduleRec_.ModuleRecMethod(...)` ditto EnvRecMethod

claudepache: The Perl user in me is suggesting:

> _envRec_ →EnvRecMethod(...)

and:

> |ParseNode|→Evaluate()

where |ParseNode| is implicitly [blessed](https://perldoc.perl.org/functions/bless.html) into `main`.

bakkot:
@bterlson is no longer editor and the current editor group doesn't intend to pick this up, so I'm going to close this for the moment. If anyone is strongly in favor of this change and is inclined to do the work, open a new issue where we can discuss it.

jmdyck:
I made [a PR against this PR](https://github.com/bterlson/ecma262/pull/3) a couple years ago, but it never got incorporated. I suppose I'll re-mount the whole thing as a PR against master sometime.

bakkot:
@jmdyck Check with the editors before doing the work (by opening an issue); we weren't clear on whether we were in favor of the change.

jmdyck:
ok

PR #957, opened 2017-07-25
not satisfied with being simply humongous, I've made the logo tremendous

michaelficarra:
I, personally, think the size of the logo should be directly related to the greatness of the language. If anything, we should strive to make it larger over time. The fact that this logo is of exceptional size reflects on how exceptional the language is.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -21,7 +21,7 @@
   margin: 0; padding: 0;
 }
 #ecma-logo {
-  width: 500px;
+  width: 110%;
 }
 </style>
 <script>


tkellen:
🤣

spec.html
```diff
@@ -21,7 +21,7 @@
   margin: 0; padding: 0;
 }
 #ecma-logo {
-  width: 500px;
+  width: 110%;
```
ljharb: needs more zeroes

bterlson: I mean, it is an SVG. We should take advantage of that.

ljharb: 1e262%

michaelficarra: That's absurd. In a survey of other standards documents, none of them use more than 4 characters to describe the logo width. :upside_down_face:

michaelficarra:
I, personally, think the size of the logo should be directly related to the greatness of the language. If anything, we should strive to make it larger over time. The fact that this logo is of exceptional size reflects on how exceptional the language is.

claudepache:
Exceptional, indeed. Rare are the languages who have two distinct values for nil, or whose type-checking operator assimilates  half of nils to objects.

Issue #956, opened 2017-07-25
Inconsistent margin on ObjectLiteral evaluation semantics


loganfsmyth:
In https://tc39.github.io/ecma262/#sec-object-initializer-runtime-semantics-evaluation, the second runtime evaluation rule for `ObjectLiteral` is indented too much, which interrupts the alternating style of grammar rule followed by indented evaluation steps.

Looks like the others have the `collapsed` attribute, but it also looks questionable collapsed onto one line, so not sure what the right fix would be.

![screen shot 2017-07-25 at 10 45 21 am](https://user-images.githubusercontent.com/132260/28585820-6e82bd9e-7126-11e7-97ee-281f52581f52.png)

jmdyck:
(Note that the behavior is not particular to that one case. It looks like every single-line production in the spec is flush-left, and every multi-line production is indented.)

loganfsmyth:
Totally, this is just a case where some things are single-line and some aren't in the same block, which makes it look inconsistent.

tjcrowder:
This problem can be found in ES2015 and ES2016 as well as ES2017. It **significantly** impairs the readability of the spec, which isn't exactly a marvel of clarity in the first place. I know I want my time back trying to figure out what the heck it meant, viz https://esdiscuss.org/topic/reading-tail-calls-specification. :-)

With respect, I suggest this is an urgent issue that needs urgent resolution.

tjcrowder:
I'll try to carve out some time to take a stab at this this weekend. Seems like the kind of thing someone not deeply involved in the spec can do, saving the heavy lifters for more heavy lifting.

tjcrowder:
@bterlson -

Flagging you as this may require changes to [ecmarkup](https://github.com/bterlson/ecmarkup) (or [grammarkdown](https://github.com/rbuckton/grammarkdown), but I think it would be ecmarkup). Or it may not; you may have a better idea&nbsp;&mdash; in any case, would benefit from your input. :-)

I'm brand-new to this stuff, but I believe the issue (No. 1 below; "Current") is because some of the productions are collapsed in the spec's source text and some aren't. Here's a sample of the Grammarmarkdown:

```grammarmarkdown
FunctionStatementList : [empty]

StatementListItem : Declaration

Statement :
  VariableStatement
  EmptyStatement
  ExpressionStatement
  ContinueStatement
  BreakStatement
  ThrowStatement
  DebuggerStatement

Block : `{` `}`
```

[...]

PR #960, opened 2017-07-26
Adds grammarkdown checking for emu-grammar elements

rbuckton:
The ecmarkup-vscode plugin can perform some static verification of grammar during editing, but it needs to be able to distinguish between `emu-grammar` elements that add definitive grammar to the document vs. elements that are used for examples or static semantics. To support this, ecmarkup-vscode can recognize a `strict` or `relaxed` parameter in any `emu-grammar` element, as well as a `grammar: 'strict'` or `grammar: 'relaxed'` property of the document metadata. `strict` elements are checked while `relaxed` elements are not. The `grammar` metadata property sets the default for the document for any `emu-grammar` elements without either of these attributes, and the default for any spec document is `relaxed`.

This PR adds a `strict` attribute to each `emu-grammar` element that contributes syntax directly to the language. The `strict` attribute does not effect emit in any way.

~This PR also fixes several places in the grammar that were written as `[lookahead <! DecimalDigit]` (using the `<!` grammarkdown sigil indicating "not in set") that should have been written as `[lookahead != DecimalDigit]`. This inconsistency was discovered once `strict` checking was enabled for the grammar. This change _does_ affect output and should be considered carefully. If it is intended that `[lookahead <! DecimialDigit]` should be valid in grammarkdown please let me know and I can make the change there.~

*edit: support for* `[lookahead <! NonTerminal]` *at check time has been fixed in grammarkdown 1.0.8*

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3457,7 +3457,7 @@
           <p>The terminal symbols of this grammar are all composed of Unicode BMP code points so the result will be *NaN* if the string contains the UTF-16 encoding of any supplementary code points or any unpaired surrogate code points.</p>
         </emu-note>
         <h2>Syntax</h2>
-        <emu-grammar use="definition">
+        <emu-grammar>
           StringNumericLiteral :::
             StrWhiteSpace?
             StrWhiteSpace? StrNumericLiteral StrWhiteSpace?
@@ -9175,7 +9175,7 @@
   <emu-clause id="sec-source-text">
     <h1>Source Text</h1>
     <h2>Syntax</h2>
-    <emu-grammar use="definition">
+    <emu-grammar>
       SourceCharacter ::
         &gt; any Unicode code point
     </emu-grammar>
@@ -9285,7 +9285,7 @@
     </code></pre>
   </emu-note>
   <h2>Syntax</h2>
-  <emu-grammar use="definition">
+  <emu-grammar>
     InputElementDiv ::
       WhiteSpace
       LineTerminator
@@ -9495,7 +9495,7 @@
       <p>Other than for the code points listed in <emu-xref href="#table-32"></emu-xref>, ECMAScript |WhiteSpace| intentionally excludes all code points that have the Unicode &ldquo;White_Space&rdquo; property but which are not classified in category &ldquo;Space_Separator&rdquo; (&ldquo;Zs&rdquo;).</p>
     </emu-note>
     <h2>Syntax</h2>
-    <emu-grammar use="definition">
+    <emu-grammar>
       WhiteSpace ::
         &lt;TAB&gt;
         &lt;VT&gt;
@@ -9577,7 +9577,7 @@
     </emu-table>
     <p>Only the Unicode code points in <emu-xref href="#table-33"></emu-xref> are treated as line terminators. Other new line or line breaking Unicode code points are not treated as line terminators but are treated as white space if they meet the requirements listed in <emu-xref href="#table-32"></emu-xref>. The sequence &lt;CR&gt;&lt;LF&gt; is commonly used as a line terminator. It should be considered a single |SourceCharacter| for the purpose of reporting line numbers.</p>
     <h2>Syntax</h2>
-    <emu-grammar use="definition">
+    <emu-grammar>
       LineTerminator ::
         &lt;LF&gt;
         &lt;CR&gt;
@@ -9600,7 +9600,7 @@
     <p>Because a single-line comment can contain any Unicode code point except a |LineTerminator| code point, and because of the general rule that a token is always as long as possible, a single-line comment always consists of all code points from the `//` marker to the end of the line. However, the |LineTerminator| at the end of the line is not considered to be part of the single-line comment; it is recognized separately by the lexical grammar and becomes part of the stream of input elements for the syntactic grammar. This point is very important, because it implies that the presence or absence of single-line comments does not affect the process of automatic semicolon insertion (see <emu-xref href="#sec-automatic-semicolon-insertion"></emu-xref>).</p>
     <p>Comments behave like white space and are discarded except that, if a |MultiLineComment| contains a line terminator code point, then the entire comment is considered to be a |LineTerminator| for purposes of parsing by the syntactic grammar.</p>
     <h2>Syntax</h2>
-    <emu-grammar use="definition">
+    <emu-grammar>
       Comment ::
         MultiLineComment
         SingleLineComment
@@ -9637,7 +9637,7 @@
   <emu-clause id="sec-tokens">
     <h1>Tokens</h1>
     <h2>Syntax</h2>
-    <emu-grammar use="definition">
+    <emu-grammar>
       CommonToken ::
         IdentifierName
         Punctuator
@@ -9660,7 +9660,7 @@
     <p>Unicode escape sequences are permitted in an |IdentifierName|, where they contribute a single Unicode code point to the |IdentifierName|. The code point is expressed by the |HexDigits| of the |UnicodeEscapeSequence| (see <emu-xref href="#sec-literals-string-literals"></emu-xref>). The `\\` preceding the |UnicodeEscapeSequence| and the `u` and `{ }` code units, if they appear, do not contribute code points to the |IdentifierName|. A |UnicodeEscapeSequence| cannot be used to put a code point into an |IdentifierName| that would otherwise be illegal. In other words, if a `\\` |UnicodeEscapeSequence| sequence were replaced by the |SourceCharacter| it contributes, the result must still be a valid |IdentifierName| that has the exact same sequence of |SourceCharacter| elements as the original |IdentifierName|. All interpretations of |IdentifierName| within this specification are based upon their actual code points regardless of whether or not an escape sequence was used to contribute any particular code point.</p>
     <p>Two |IdentifierName|s that are canonically equivalent according to the Unicode standard are <em>not</em> equal unless, after replacement of each |UnicodeEscapeSequence|, they are represented by the exact same sequence of code points.</p>
     <h2>Syntax</h2>
-    <emu-grammar use="definition">
+    <emu-grammar>
       IdentifierName ::
         IdentifierStart
         IdentifierName IdentifierPart
@@ -9731,7 +9731,7 @@
       <h1>Reserved Words</h1>
       <p>A reserved word is an |IdentifierName| that cannot be used as an |Identifier|.</p>
       <h2>Syntax</h2>
-      <emu-grammar use="definition">
+      <emu-grammar>
         ReservedWord ::
           Keyword
           FutureReservedWord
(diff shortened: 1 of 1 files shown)


jmdyck:
Lookahead restrictions were introduced in the 3rd edition (1999), along with the text:
> For convenience, the set can also be written as a nonterminal, in which case it represents the set of all terminals to which that nonterminal could expand.

and the example:
> LookaheadExample :: DecimalDigit [lookahead ∉ DecimalDigit ]

Both have been in the spec ever since.

So it's unclear to me why you think `[lookahead <! DecimalDigit]` is an inconsistency.

`[lookahead != foo]` should only be used if `foo` is a terminal, and `DecimalDigit` is not a terminal.

rbuckton:
That's fine, and why I asked. grammarkdown currently only allows [lookahead &notin; { ... }] with `{` and `}`. It doesn't affect the emit (as it's working now) but reports an error when using the checker. I can modify grammarkdown and will amend this PR shortly.

bterlson:
I feel like 'strict' is a misnomer in this context. Does it mean something closer to 'strict' in gmd? Just looking at these changes I think I'd prefer a "definition" attribute. thoughts?

rbuckton:
I chose `strict` vs `relaxed` as it applies to how the emu-grammar nodes are checked by grammarkdown's static checker. I can consider `definition` as this is specific to how grammarkdown interacts with ecmarkup.

rbuckton:
If I use `definition` I will likely also support `strict` as interchangeable terms.

bterlson:
So "relaxed" means "don't check anything" vs. "check this thing understanding that this production and any of its alternatives are defined elsewhere"?

rbuckton:
Yes. And "check this thing" is what enables highlighting misspelled nonterminals, go to definition, and find all references in ecmarkup-vscode. There may also be a useful third state for the emu-grammar elements in static semantics sections where they are partially checked (e.g. find misspellings, go to def, but allow you to elide production parameters). So `definition` might make sense if we then also add a `reference` or `semantics` or some other name.

For `relaxed` I don't even parse the emu-grammar elements in the ecmarkup language server.

bterlson:
Yeah I think GMD's modes don't map cleanly here. I'm leaning toward "definition" attribute (with a default value of false if the attribute is not present). Definition=false elements are today ignored but I'm very interested in getting errors for referring to a non-existent RHS. Of course emu-prodref solves this in another way, but in a way that is totally not readable in the source document.

rbuckton:
I propose:
- `definition` - Defines the grammar of the spec. Performs full static checking of the grammar
  - reports errors for misspellings
  - reports errors for invalid grammarkdown syntax
  - reports errors for missing/unsupplied production parameters
  - enables go-to-definition in an editor
- `semantics` (or not set) - Refers to the defined grammar of the spec. Performs partial static checking of the grammar
  - reports errors for misspellings
  - reports errors for invalid grammarkdown syntax
  - does not report errors for missing/unsupplied production parameters
  - enables go-to-definition in an editor
- `example` (or some other term) 
  - not checked at all

jmdyck:
To me, it seems odd to define multiple boolean attributes if they're mutually exclusive. Why not a single attribute with multiple alternative values?

(Theoretically, 'example' isn't mutually exclusive, it's orthogonal to defining/using. I.e., you can have could have both defining productions and using productions in an example. But I'm not sure that's a useful viewpoint.)

-----

`semantics` could check correctness of prodn parameters *if* they appear.

Shouldn't the treatment for `example` include at least a check for grammarkdown syntax? Else why even bother handing it to grammarkdown?

-----

In the commit, some defining emu-grammar elements have not been marked as such:

* One in each of the 5 "Supplemental Syntax" sections.
* The one in "JSON.parse" (for DoubleStringCharacter in the weird parallel grammar for parsing JSON texts).
* The one in "FunctionDeclarations in IfStatement Statement Clauses" (augmenting IfStatement).
* The one in "Initializers in ForIn Statement Heads" (augmenting IterationStatement).

rbuckton:
`example` could just check syntax, yes. Regardless of the overlap, I'm primarily trying to address the expected level of correctness in each element, as well as to use the actual spec.html file as a source of truth for proposal specifications that augment the specification.

I'd prefer it if the default for emu-grammar was strict checking, so that a proposal author could easily start a new emu-grammar element and get the benefits of the checker (including completions and signature help for parametric productions, eventually). With the strict vs relaxed nomenclature, it made sense to be able to define a default for the document. That makes less sense with terms like 'definition', and there a significantly more emu-grammar nodes for static semantics in the spec than those for definitions, so strict checking by default makes less sense.

The states *are* mutually exclusive. It would make more sense to use an attribute like 'mode=strict'. It is the way it is now because it used to be that all you would do is specify an override for the document strictness setting. Writing `mode=` over and over again seemed inefficient.

rbuckton:
So here is what I came up with: 
- `<emu-grammar>` now has an optional `use` attribute.
- Valid values for `use` are: 
  - `"definition"` - Defines grammar for a specification, with stricter checking:
    - Report errors in grammarkdown syntax
    - Report errors in misspellings
    - Report errors in missing/unsupplied production parameters
    - Report errors in missing/unsupplied nonterminal parameters
    - Supports "Go To Definition" and "Find All References", etc. in supported editors
    - Productions show up as definitions in "Go To Definition"
  - `"reference"` - References defined grammar for refinements and static semantics, with relaxed checking:
    - Report errors in grammarkdown syntax
    - Report errors in misspellings
    - Report errors in missing/unsupplied nonterminal parameters
    - Supports "Go To Definition" and "Find All References", etc. in supported editors.
    - Productions **do not** show up as definitions in "Go To Definition"
  - `"example"` - Examples that are not intended to introduce or reference grammar, with limited checking:
    - Report errors in grammarkdown syntax
    - No support for "Go To Definition" or "Find All References"
- Default value for `use` is `"definition"`
- Invalid values are treated as `"example"`

bterlson:
I think this is getting to a great place and I am happy to see "definition" as the default (while more verbose for 262, it's better for proposals).

I am not sure about the checking requirements of "reference" however - the spec allows omitting grammar parameters in SDOs. Viz: `When an algorithm is associated with a production alternative, the alternative is typically shown without any “[ ]” grammar annotations. Such annotations should only affect the syntactic recognition of the alternative and have no effect on the associated semantics for the alternative.`. Let's make sure the checking semantics for reference allow for this.

rbuckton:
That sounds fine to me. My understanding is that for `reference`, we only check that _the parameters you supply_ are valid, and not error on "missing" parameters.

rbuckton:
I updated my [comment above](https://github.com/tc39/ecma262/pull/960#issuecomment-318468136) to reflect this.

bterlson:
I'm alright with this model, but I'll wait for @jmdyck's thoughts before I pull it. In the meantime, @rbuckton how about removing `use="definition"` since it's the default?

rbuckton:
I can do that. I thought it was worthwhile to be explicit.

jmdyck:
To me, `use=reference` reads like "these productions can be used as a reference", which sounds too much like they're definitional (like a reference kilogram or a reference implementation), so some proposal-writers might be puzzled. In my code, I say (in effect) `status = (definition|use|example)`.

jmdyck:
> My understanding is that for reference, we only check that the parameters you supply are valid, and not error on "missing" parameters.

For 262 at least, we can be even more constraining: if *any* parameters appear in the production, then *all* must appear (and be correct, of course). I don't know if proposals satisfy this as well, but that'd be my guess.

jmdyck:
@bterlson:
>  I am happy to see "definition" as the default (while more verbose for 262, it's better for proposals).

For 262's `emu-grammar` elements, about 92% are 'reference', 6% are 'definition' and 2% are 'example' (so yes, definitely more verbose for 262).

Are you saying that in proposals, the definitions outnumber the references?

bterlson:
@jmdyck that is true for the couple I examined (and it made sense since static semantics are not always required or implemented in early stages). Willing to go the other way, though.

rbuckton:
No, I'm requesting that the strictest checking should be the default as it adds the most value from a tooling perspective. Prior to the request to use something other than 'strict', it was possible to define a default strictness for the document. That makes less sense when the strictness setting is a semantic meaning (like 'definition') instead of an literal meaning (like 'strict' or 'relaxed').

jmdyck:
@bterlson: One thing to consider is that (in 262 at least) all definitional emu-grammars are multi-line, i.e. the `<emu-grammar>` tag is on its own line, so adding an attribute to it doesn't clutter the source much. On the other hand, many 'reference' emu-grammars are inline (e.g., `The MV of <emu-grammar>DecimalDigit :: '1'</emu-grammar> or of <emu-grammar>NonZeroDigit :: '1'</emu-grammar> or ...`), where adding an attribute makes the source line somewhat harder to read.

@rbuckton: So you believe that the checking for use=definition is stricter that use=reference? I'm not convinced it is. In fact, I'd probably argue the reverse, since any production in use=reference must match a production in some use=definition, whereas productions in use=definition don't have that constraint.

rbuckton:
@jmdyck That's a fair assessment, though neither is a proper subset of the other. Another option we could consider is setting the semantic meaning (like `definition` or `reference`) on the containing `emu-clause` (or `emu-annex`, etc.) similar to how the `namespace` attribute is used in the spec. Then the meaning of a specific `<emu-grammar>` element can override the ambient meaning.

Currently, grammarkdown does not yet even understand these semantic differences and only cares about `definition`. The other kind of checking for `reference` doesn't exist yet.

jmdyck:
> Another option we could consider is setting the semantic meaning (like `definition` or `reference`) on the containing `emu-clause` (or `emu-annex`, etc.) similar to how the `namespace` attribute is used in the spec. Then the meaning of a specific `<emu-grammar>` element can override the ambient meaning.

We could, but note that in such a scheme, we might end up putting `reference` on all the top-level clauses from "Abstract Operations" onward and then overriding that for definitions (and the odd example). Which would be roughly equivalent to just saying that `reference` is the default (but requiring more code to maintain/ascertain the 'ambient meaning').

rbuckton:
Fair enough. The tooling I have in ecmarkup-vscode doesn't yet support `reference`, so you won't get errors in those nodes yet.

rbuckton:
For now I will simplify this: `<emu-grammar>` nodes are checked, `<emu-grammar use="example">` nodes are not. I will defer `definition` vs `reference` to a later time.

jmdyck:
@bterlson: In issue #956, you say "Seems like #960 will solve this nicely", but it won't in its current form, as it only adds `use="example"` attributes, whereas issue #956 will require a (CSS-addressable) distinction between 'definition' and 'reference' productions, which @rbuckton has deferred from this PR. So:

1. Would you like the remaining attributes to be added in this PR or a separate PR?
2. Would you like the default to be 'definition' or 'reference'? (I argue the latter.)
3. Any opinion on nomenclature: `use=(definition|reference|example)` vs `status=(definition|use|example)`? (I prefer the latter.)

PR #961, opened 2017-07-27
Edit misleading note on return statements.

GeorgNeis:


bterlson:
I don't really like 'overruled' here but without a better suggestion I'll pull this in :)

anba:
Strictly speaking there are also other ways to intercept a `return` statement, e.g. 
```js
function f() {
    var it = {
        [Symbol.iterator]() { return this; },
        next() { return {}; },
        return() { throw "HI"; },
    };
    try {
        for (var x of it) return;
    } catch (e) {
        print("caught:", e);
    }
}
f();
```

So we may be better of not specifically mentioning `finally` as the only way to intercept a `return`...?

bterlson:
@anba you're right I think. I'll tweak this wording.

bterlson:
https://github.com/tc39/ecma262/blob/549cf469b50f620e08dac10a89226913530bf8a6/spec.html#L16923-L16924

(Sad that inline source context doesn't seem to work with large files :()

Here's the new note text:
>  A `return` statement causes a function to cease execution and, in most cases, returns a value to the caller. If |Expression| is omitted, the return value is *undefined*. Otherwise, the return value is the value of |Expression|. A `return` statement may not actually return a value to the caller depending on surrounding context. For example, in a `try` block, a `return` statement's completion record may be replaced with another completion record during evaluation of the `finally` block.

GeorgNeis:
Good catch!

PR #963, opened 2017-07-29
refactor EvaluateCall/EvaluateDirectCall

claudepache:
EvaluateDirectCall is used instead of EvaluateCall in only one place,
and this is because additional steps must be added after the first
GetValue(ref) step of EvaluateCall. So:
* factor GetValue(ref) out of EvaluateCall;
* now the call site of EvaluateDirectCall may use directly EvaluateCall;
* merge EvaluateDirectCall into EvaluateCall.
Incidently, correct two convention infringements in the invocation of
ArgumentListEvaluation.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12363,41 +12363,34 @@
           1. Let _func_ be ? GetValue(_ref_).
           1. If Type(_ref_) is Reference and IsPropertyReference(_ref_) is *false* and GetReferencedName(_ref_) is `"eval"`, then
             1. If SameValue(_func_, %eval%) is *true*, then
-              1. Let _argList_ be ? ArgumentListEvaluation(_arguments_).
+              1. Let _argList_ be ArgumentListEvaluation of _arguments_.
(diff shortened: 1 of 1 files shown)


claudepache:
For context, I came into the same issue (couldn’t use EvaluateCall because of steps needed after GetValue(ref)) when writing the spec algo for “optional function evaluation” `foo?.(bar)` in the context of the optional chaining proposal.

jmdyck:
The refactoring seems correct to me.

A few minor notes:
- There may be external links to EvaluateDirectCall. To handle these somewhat gracefully, we might want to preserve `id="sec-evaluatedirectcall"` (on the `h1`, say).
- It's unnecessary to replace the '?' shorthand with a call to `ReturnIfAbrupt`, as the shorthand is explicitly allowed for the invocation of a syntax-directed op.
- In the h1-header for an abstract op, we usually have a space directly inside the parens.

bterlson:
I'll fix this up now (using the new oldids for EvaluateDirectCall!)

PR #974, opened 2017-08-15
Meta: Add DFNs for function object and constructor

tobie:
A link to *function object* is needed by WebIDL (in particular by https://github.com/heycam/webidl/pull/405).

Additionally, it would be useful to link to *constructor* from various parts of the spec, notably from the introductory paragraphs to [interface objects](https://heycam.github.io/webidl/#interface-object): 

> If the interface is declared with a [Constructor] extended attribute, then the interface object can be called as a _constructor_ […]

This PR adds both.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1488,7 +1488,7 @@
             </tbody>
           </table>
         </emu-table>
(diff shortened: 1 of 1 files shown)


bterlson:
Just note that these are editorial changes. Meta used for metadata about the spec and other non-spec things (like package.json changes, build infra, etc.)

tobie:
Oops, sorry. WHATWG considers those as meta, so I got mixed up.

PR #977, opened 2017-08-16
Meta: Add DFN for array index

tobie:
A link to array index would be useful in WebIDL (for example to clarify
heycam/webidl#409).


bterlson:
Good idea, and congrats on your first contribution!

PR #983, opened 2017-08-28
Misc editorial fixes

anba:
Except for c70a81a42a8c7f6af8d2dfe1aacd976130b0208a, the other commits should be acceptable as non-normative changes.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -730,7 +730,7 @@
     <p>A step may specify the iterative application of its substeps.</p>
     <p>A step that begins with &ldquo;<dfn id="assert">Assert</dfn>:&rdquo; asserts an invariant condition of its algorithm. Such assertions are used to make explicit algorithmic invariants that would otherwise be implicit. Such assertions add no additional semantic requirements and hence need not be checked by an implementation. They are used simply to clarify algorithms.</p>
     <p>Algorithm steps may declare named aliases for any value using the form &ldquo;Let _x_ be _someValue_&rdquo;. These aliases are reference-like in that both _x_ and _someValue_ refer to the same underlying data and modifications to either are visible to both. Algorithm steps that want to avoid this reference-like behaviour should explicitly make a copy of the right-hand side: &ldquo;Let _x_ be a copy of _someValue_&rdquo; creates a shallow copy of _someValue_.</p>
-    <p>Once declared, aliases may be referenced in any subsequent step or substep thereof, but they may not be referenced from steps at a higher level. Aliases may be modified using the form &ldquo;Set _x_ to _someOtherValue_&rdquo;.</p>
+    <p>Once declared, an alias may be referenced in any subsequent steps and must not be referenced from steps prior to the alias's declaration. Aliases may be modified using the form &ldquo;Set _x_ to _someOtherValue_&rdquo;.</p>
 
     <emu-clause id=sec-algorithm-conventions-abstract-operations>
       <h1>Abstract Operations</h1>
--- a/spec.html
+++ b/spec.html
@@ -28893,7 +28893,7 @@ THH:mm:ss.sss
       <emu-clause id="sec-string.prototype.touppercase">
         <h1>String.prototype.toUpperCase ( )</h1>
         <p>This function interprets a String value as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>.</p>
-        <p>This function behaves in exactly the same way as `String.prototype.toLowerCase`, except that the String is mapped using to toUppercase algorithm of Unicode Default Case Conversion.</p>
+        <p>This function behaves in exactly the same way as `String.prototype.toLowerCase`, except that the String is mapped using the toUppercase algorithm of the Unicode Default Case Conversion.</p>
         <emu-note>
           <p>The `toUpperCase` function is intentionally generic; it does not require that its *this* value be a String object. Therefore, it can be transferred to other kinds of objects for use as a method.</p>
         </emu-note>
--- a/spec.html
+++ b/spec.html
@@ -10332,16 +10332,16 @@
             The SV of <emu-grammar>StringLiteral :: `'` SingleStringCharacters `'`</emu-grammar> is the SV of |SingleStringCharacters|.
           </li>
           <li>
-            The SV of <emu-grammar>DoubleStringCharacters :: DoubleStringCharacter</emu-grammar> is a sequence of one or two code units that is the SV of |DoubleStringCharacter|.
+            The SV of <emu-grammar>DoubleStringCharacters :: DoubleStringCharacter</emu-grammar> is a sequence of up to two code units that is the SV of |DoubleStringCharacter|.
           </li>
           <li>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -39905,8 +39905,8 @@
         <h1>Changes to Abstract Equality Comparison</h1>
         <p>The following steps are inserted after step 3 of the <emu-xref href="#sec-abstract-equality-comparison">Abstract Equality Comparison</emu-xref> algorithm:</p>
         <emu-alg>
-          1. If _x_ has an [[IsHTMLDDA]] internal slot and _y_ is either *null* or *undefined*, return *true*.
-          1. If _x_ is either *null* or *undefined* and _y_ has an [[IsHTMLDDA]] internal slot, return *true*.
+          1. If Type(_x_) is Object and _x_ has an [[IsHTMLDDA]] internal slot and _y_ is either *null* or *undefined*, return *true*.
```
bterlson: My feeling is "X has [[SomeSlot]] internal slot" implicitly carries an "if Type(_x_) is Object" but explicitness is fine here too (unless I'm missing the issue this is addressing).

spec.html
```diff
@@ -730,7 +730,6 @@
     <p>A step may specify the iterative application of its substeps.</p>
     <p>A step that begins with &ldquo;<dfn id="assert">Assert</dfn>:&rdquo; asserts an invariant condition of its algorithm. Such assertions are used to make explicit algorithmic invariants that would otherwise be implicit. Such assertions add no additional semantic requirements and hence need not be checked by an implementation. They are used simply to clarify algorithms.</p>
     <p>Algorithm steps may declare named aliases for any value using the form &ldquo;Let _x_ be _someValue_&rdquo;. These aliases are reference-like in that both _x_ and _someValue_ refer to the same underlying data and modifications to either are visible to both. Algorithm steps that want to avoid this reference-like behaviour should explicitly make a copy of the right-hand side: &ldquo;Let _x_ be a copy of _someValue_&rdquo; creates a shallow copy of _someValue_.</p>
-    <p>Once declared, aliases may be referenced in any subsequent step or substep thereof, but they may not be referenced from steps at a higher level. Aliases may be modified using the form &ldquo;Set _x_ to _someOtherValue_&rdquo;.</p>
```
bterlson: I don't think a total revert here is necessary - I still like most of this change. Can we just remove the step scoping requirement (i.e. should never refer to a subsequent step of any depth but may refer to a previous step at any depth)?

spec.html
```diff
@@ -10244,16 +10244,16 @@
             The SV of <emu-grammar>StringLiteral :: `'` SingleStringCharacters `'`</emu-grammar> is the SV of |SingleStringCharacters|.
           </li>
           <li>
-            The SV of <emu-grammar>DoubleStringCharacters :: DoubleStringCharacter</emu-grammar> is a sequence of one or two code units that is the SV of |DoubleStringCharacter|.
+            The SV of <emu-grammar>DoubleStringCharacters :: DoubleStringCharacter</emu-grammar> is a sequence of up to two code units that is the SV of |DoubleStringCharacter|.
```
bterlson: Wowza, good catch here.

spec.html
```diff
@@ -21345,6 +21345,7 @@
               </td>
               <td>
                 <p>Return the binding of a name exported by this module. Bindings are represented by a <dfn id="resolvedbinding-record">ResolvedBinding Record</dfn>, of the form {[[Module]]: Module Record, [[BindingName]]: String}. If the name cannot be resolved, return either the module record that is to blame for the resolution failure or, in case there isn't any, return *null* or `"ambiguous`" depending on the nature of the failure.</p>
+                <p>This operation must be idempotent if it completes normally. Each time it is called with a specific _exportName_, _resolveSet_ pair as arguments it must return the same result.</p>
```
bterlson: @domenic @GeorgNeis FYI.

spec.html
```diff
@@ -22053,11 +22053,13 @@
               1. Perform ? ModuleDeclarationEnvironmentSetup(_module_).
               1. Assert: _module_ occurs exactly once in _stack_.
               1. Assert: _module_.[[DFSAncestorIndex]] is less than or equal to _module_.[[DFSIndex]].
-              1. If _module_.[[DFSAncestorIndex]] equals _module_.[[DFSIndex]], repeat:
-                1. Let _requiredModule_ be the last element in _stack_.
-                1. Remove this element from _stack_.
-                1. Set _requiredModule_.[[Status]] to `"instantiated"`.
-                1. If _requiredModule_ is _module_, end the repetition.
+              1. If _module_.[[DFSAncestorIndex]] equals _module_.[[DFSIndex]], then
```
bterlson: I personally found the repeat notation fine. Do you prefer the explicit while loop, or is this about consistency with existing forms?

spec.html
domenic: Is that true? It seems likely, but I've recently been made aware of how horrifically complicated export resolution is (due to export from and circular dependencies), so I wouldn't want to bet on this at my current level of understanding.

spec.html
domenic: I don't think it does, unfortunately, given how many places in the spec precede internal slot checks with type checks.

domenic:
"Return the module evaluation result from ModuleExecution"

this doesn't seem like an editorial fix, just a straight-up bug fix.

spec.html
anba: It's not necessarily true, but it's the easiest way to enforce that the assertion in step 7 of https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver holds:
> ...
> 6. Let binding be ! m.ResolveExport(P, « »).
> 7. Assert: binding is a ResolvedBinding Record.
> ...

There are other ways to correct this assertion:
- We could make Source Text Module Records the only allowed Module Record type.
- We could disallow non-Source Text Module Records in Source Text Module Records module dependency graphs, and then restrict Module Namespace objects to Source Text Module Records.
- We could store the ResolveExport results in https://tc39.github.io/ecma262/#sec-getmodulenamespace into the Module Namespace object, so that https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver no longer needs to call ResolveExport.
- Or we could just make step 6 in https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver fallible again.

spec.html
anba: Just for consistency with the existing forms.

spec.html
anba: So we keep:
> Aliases may be modified using the form &ldquo;Set _x_ to _someOtherValue_&rdquo;.

And if necessary, maybe something along the lines of:
> Aliases cannot be used before their declaration.

spec.html
anba: What @domenic said: The spec always tests the type before accessing internal slots.

spec.html
bterlson: Yeah, that's my thinking. Want to update this? Or I can exclude this commit too.

spec.html
bterlson: @domenic I'm going to pull this, we can select another alternative above later if you like.

GeorgNeis: The added requirement lgtm.

GeorgNeis: There's a similar comment in https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver

bterlson:
Pushed the updated commits with editorial/normative tags and my own commit to remove the step scoping in lieu of reverting the previous related editorial change.

PR #984, opened 2017-08-28
Early errors for RegExp literals

### LINKED ISSUE #918: Invalid regexp property escape should throw an __early__ SyntaxError.
As verified in [Test262](https://github.com/tc39/test262/pull/1034#issuecomment-301541395), I'd like to confirm a mal formed property escape in a regexp should throw an Early Error.

I can't write a patch for this right now, but I want to keep track of this issue.

From @anba comments:

> ES 5.1 had:
> > An implementation must treat any instance of the following kinds of errors as an early error:
> > - [...]
> > - Errors in regular expression literals that are not implementation-defined syntax extensions.
>
> Somehow this was dropped in ES6 which leads to at least one contradiction, in https://tc39.github.io/ecma262/#sec-pattern, we've got:
> > [...] The algorithms in 21.2.2 are designed so that compiling a pattern may throw a SyntaxError exception; on the other hand, once the pattern is successfully compiled, applying the resulting internal procedure to find a match in a String cannot throw an exception (except for any host-defined exceptions that can occur anywhere such as out-of-memory).
>
> But a regular expression like `/a|(?![\d-\d])/u` may or may not evaluate the right-hand side disjunction (which contains an invalid range expression), it depends on the input string. 
> 
> So it should be listed as an early error in the spec, but no one spend time on correcting this issue...

anba:
Fixes #918

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -29130,6 +29130,266 @@ THH:mm:ss.sss
             It is a Syntax Error if _NcapturingParens_ &ge; 2<sup>32</sup>-1.
           </li>
         </ul>
+        <emu-grammar>QuantifierPrefix :: `{` DecimalDigits `,` DecimalDigits `}`</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if the MV of the first |DecimalDigits| is larger than the MV of the second |DecimalDigits|.
+          </li>
+        </ul>
+        <emu-grammar>AtomEscape :: DecimalEscape</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if the CapturingGroupNumber of |DecimalEscape| is larger than _NcapturingParens_ (<emu-xref href="#sec-notation"></emu-xref>).
+          </li>
+        </ul>
+        <emu-grammar>NonemptyClassRanges :: ClassAtom `-` ClassAtom ClassRanges</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if IsCharacterClass of the first |ClassAtom| is *true* or IsCharacterClass of the second |ClassAtom| is *true*.
+          </li>
+          <li>
+            It is a Syntax Error if IsCharacterClass of the first |ClassAtom| is *false* and IsCharacterClass of the second |ClassAtom| is *false* and the CharacterValue of the first |ClassAtom| is larger than the CharacterValue of the second |ClassAtom|.
+          </li>
+        </ul>
+        <emu-grammar>NonemptyClassRangesNoDash :: ClassAtomNoDash `-` ClassAtom ClassRanges</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if IsCharacterClass of |ClassAtomNoDash| is *true* or IsCharacterClass of |ClassAtom| is *true*.
+          </li>
+          <li>
+            It is a Syntax Error if IsCharacterClass of |ClassAtomNoDash| is *false* and IsCharacterClass of |ClassAtom| is *false* and the CharacterValue of |ClassAtomNoDash| is larger than the CharacterValue of |ClassAtom|.
+          </li>
+        </ul>
+      </emu-clause>
+
+      <emu-clause id="sec-patterns-static-semantics-capturing-group-number">
+        <h1>Static Semantics: CapturingGroupNumber</h1>
+        <emu-grammar>DecimalEscape :: NonZeroDigit</emu-grammar>
+        <emu-alg>
+          1. Return the MV of |NonZeroDigit|.
+        </emu-alg>
+        <emu-grammar>DecimalEscape :: NonZeroDigit DecimalDigits</emu-grammar>
+        <emu-alg>
+          1. Let _n_ be the number of code points in |DecimalDigits|.
+          1. Return (the MV of |NonZeroDigit| &times; 10<sup>_n_</sup>) plus the MV of |DecimalDigits|.
+        </emu-alg>
+        <p>The definitions of &ldquo;the MV of |NonZeroDigit|&rdquo; and &ldquo;the MV of |DecimalDigits|&rdquo; are in <emu-xref href="#sec-literals-numeric-literals"></emu-xref>.</p>
+      </emu-clause>
+
+      <emu-clause id="sec-patterns-static-semantics-is-character-class">
+        <h1>Static Semantics: IsCharacterClass</h1>
+        <emu-grammar>
+          ClassAtom :: `-`
+          ClassAtomNoDash :: SourceCharacter but not one of `\` or `]` or `-`
+          ClassEscape :: `b`
+          ClassEscape :: `-`
+          ClassEscape :: CharacterEscape
+        </emu-grammar>
+        <emu-alg>
+          1. Return *false*.
+        </emu-alg>
+        <emu-grammar>ClassEscape :: CharacterClassEscape</emu-grammar>
+        <emu-alg>
+          1. Return *true*.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-patterns-static-semantics-character-value">
+        <h1>Static Semantics: CharacterValue</h1>
+        <emu-grammar>
+          ClassAtom :: `-`
+        </emu-grammar>
+        <emu-alg>
+          1. Return the code point value of U+002D (HYPHEN-MINUS).
+        </emu-alg>
+        <emu-grammar>
+          ClassAtomNoDash :: SourceCharacter but not one of `\` or `]` or `-`
+        </emu-grammar>
+        <emu-alg>
+          1. Let _ch_ be the code point matched by |SourceCharacter|.
+          1. Return the code point value of _ch_.
+        </emu-alg>
+        <emu-grammar>
+          ClassEscape :: `b`
+        </emu-grammar>
+        <emu-alg>
+          1. Return the code point value of U+0008 (BACKSPACE).
+        </emu-alg>
+        <emu-grammar>
+          ClassEscape :: `-`
+        </emu-grammar>
+        <emu-alg>
+          1. Return the code point value of U+002D (HYPHEN-MINUS).
+        </emu-alg>
+        <emu-grammar>CharacterEscape :: ControlEscape</emu-grammar>
+        <emu-alg>
+          1. Return the code point value according to <emu-xref href="#table-47"></emu-xref>.
+        </emu-alg>
+        <emu-table id="table-47" caption="ControlEscape Code Point Values">
+          <table>
+            <tbody>
+            <tr>
+              <th>
+                ControlEscape
+              </th>
+              <th>
+                Code Point Value
+              </th>
+              <th>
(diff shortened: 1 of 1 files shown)


littledan:
This patch is great--it seems to address this issue as well: https://github.com/tc39/proposal-regexp-unicode-property-escapes/issues/29

spec.html
```diff
@@ -29130,6 +29130,266 @@
             It is a Syntax Error if _NcapturingParens_ &ge; 2<sup>32</sup>-1.
           </li>
         </ul>
+        <emu-grammar>QuantifierPrefix :: `{` DecimalDigits `,` DecimalDigits `}`</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if the MV of the first |DecimalDigits| is larger than the MV of the second |DecimalDigits|.
+          </li>
+        </ul>
```
littledan: From eshost, seems like four engines do have this early error.

mathiasbynens:
I’d upvote this a thousand times if I could. Abstracting away `IsCharacterClass` etc. makes so much sense.

spec.html
```diff
@@ -29130,6 +29130,266 @@
...
+        <emu-grammar>NonemptyClassRangesNoDash :: ClassAtomNoDash `-` ClassAtom ClassRanges</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if IsCharacterClass of |ClassAtomNoDash| is *true* or IsCharacterClass of |ClassAtom| is *true*.
+          </li>
+          <li>
+            It is a Syntax Error if IsCharacterClass of |ClassAtomNoDash| is *false* and IsCharacterClass of |ClassAtom| is *false* and the CharacterValue of |ClassAtomNoDash| is larger than the CharacterValue of |ClassAtom|.
+          </li>
```
littledan: Syntax error for everyone.

spec.html
```diff
@@ -38970,16 +39070,67 @@
...
 
+      <emu-annex id="sec-patterns-static-semantics-early-errors-annexb">
+        <h1>Static Semantics: Early Errors</h1>
+        <p>The semantics of <emu-xref href="#sec-patterns-static-semantics-early-errors"></emu-xref> is extended as follows:</p>
+        <emu-grammar>ExtendedAtom :: InvalidBracedQuantifier</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if any source text matches this rule.
+          </li>
+        </ul>
```
littledan: Error in all tested engines.

spec.html
```diff
@@ -38970,16 +39070,67 @@
...
+        <emu-grammar>NonemptyClassRanges :: ClassAtom `-` ClassAtom ClassRanges</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if IsCharacterClass of the first |ClassAtom| is *true* or IsCharacterClass of the second |ClassAtom| is *true* <ins>and this production has a [U] parameter</ins>.
+          </li>
```
littledan: This is what seems to be implemented in V8 and SpiderMonkey; the versions of JSC and ChakraCore I'm testing don't seem to throw the errors in non-Unicode mode.

spec.html
```diff
@@ -38970,16 +39070,67 @@
...
+        </ul>
+        <emu-grammar>NonemptyClassRangesNoDash :: ClassAtomNoDash `-` ClassAtom ClassRanges</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if IsCharacterClass of |ClassAtomNoDash| is *true* or IsCharacterClass of |ClassAtom| is *true* <ins>and this production has a [U] parameter</ins>.
+          </li>
```
littledan: This is what seems to be implemented in V8 and SpiderMonkey; the versions of JSC and ChakraCore I'm testing don't seem to throw the errors in non-Unicode mode.

spec.html
bterlson: Probably don't want to commit these \<ins\> tags.

bterlson:
Seems good. I just want to do my own littledan exercise of going through the various implementations and the early errors prescribed here.

Should this be a needs-consensus PR?

bakkot:
@littledan, I don't think this actually matches web reality in Chrome, does it? The relevant [v8 bug](https://bugs.chromium.org/p/v8/issues/detail?id=896) is still open, and I'm seeing that e.g. `if (false) /a{2,3}/` does not throw an error.

In particular, it looks like in v8 the error is thrown when the literal is *evaluated*, not when it's parsed (but still before it's exec'd).

bakkot:
Needs consensus PR seems right to me.

spec.html
littledan:
```
$ eshost -e "/a{5,3}/"
#### jsc
SyntaxError: Invalid regular expression: numbers out of order in {} quantifier

#### chakracore
SyntaxError: Syntax error in regular expression

#### spidermonkey
SyntaxError: numbers out of order in {} quantifier.:

#### d8
SyntaxError: Invalid regular expression: /a{5,3}/: numbers out of order in {} quantifier
```

spec.html
littledan:
```
$ eshost -e "/[z-a]/"
#### chakracore
SyntaxError: Invalid range in character set

#### jsc
SyntaxError: Invalid regular expression: range out of order in character class

#### spidermonkey
SyntaxError: invalid range in character class:

#### d8
SyntaxError: Invalid regular expression: /[z-a]/: Range out of order in character class
```

spec.html
littledan:
```
$ eshost -e "/{1}/"
#### jsc
SyntaxError: Invalid regular expression: nothing to repeat

#### chakracore
SyntaxError: Unexpected quantifier

#### spidermonkey
SyntaxError: nothing to repeat:

#### d8
SyntaxError: Invalid regular expression: /{1}/: Nothing to repeat
```

spec.html
littledan: ins is used elsewhere in Annex B, e.g., [here](https://tc39.github.io/ecma262/#sec-switch-duplicates-allowed-static-semantics).

```
$ eshost -e "/[\w-\b]/u"
#### jsc
SyntaxError: Invalid regular expression: invalid escaped character for unicode pattern

#### chakracore
/[\w-\b]/u

#### spidermonkey
SyntaxError: character class escape cannot be used in class range in regular expression:

#### d8
SyntaxError: Invalid regular expression: /[\w-\b]/: Invalid character class
```

littledan:
I want to clarify the tests that I included above. It's true that V8 has a known issue where, if a RegExp isn't reached in executing code, it's not parsed and early errors don't come.

However, there are still separate stages of parsing/compiling and executing on a string in V8. You can see this, for example, if you do `/\5/u.toString()`. In the current specification, that would not thrown an error, but in V8, it would. This is the same time as errors for RegExps which don't match the grammar are thrown, for example `/?/`.

I don't think we need to hold back this PR just because V8 has a known spec violation. Whether parsing errors are thrown when parsing the enclosing code, or when evaluating the RegExp for the first time, is not an open spec question given 3/4 of these engines giving early errors:

```
$ eshost -e "false ? /{1}/ : undefined"
#### jsc
SyntaxError: Invalid regular expression: nothing to repeat

#### chakracore
SyntaxError: Unexpected quantifier

#### spidermonkey
SyntaxError: nothing to repeat:

#### d8
undefined
```

It's currently web reality already that, when RegExps are first parsed/compiled, the errors are thrown, rather than later when they are going through a string. That's the delta that this PR makes; it doesn't change whether RegExp syntax errors come at the broader early error time or at a special, later time as V8 does.

littledan:
@bakkot For this to be marked "needs consensus": I think we ended up deciding that not all normative changes need to be discussed in committee, we should just discuss the ones that seem major or don't agree on here. Do you have doubts about this proposal, beyond the fact that V8 doesn't comply with the spec here (in a way that, I argued above, is unrelated)? If not, maybe we can land this patch without spending committee time on it. If you do have doubts, I'll add it to the agenda.

allenwb:
I believe the original assertion that this isn't already covered made in https://github.com/tc39/ecma262/issues/918 is incorrect.  See the comment I left there https://github.com/tc39/ecma262/issues/918#issuecomment-329024576

allenwb:
Below, I'm copying over most of https://github.com/tc39/ecma262/issues/918#issuecomment-329217142

The TL;DR summary is: This patch and all of the other existing syntax error checks in the RegExp evaluation algorithms should be refactored into Static Semantics Early Error rules.

----
The algorithmic syntax error detection in the RegExp evaluation algorithms really shouldn't be there. They are a carry over from Edition 3 where the spec.  which did not use separate static semantic clauses listing syntax-derived early errors (and hence the need for the Ch 16 rule).

What I should have done in ES6 is to factor all of those syntax error checks out of the RegExp evaluation algorithms and restated them as early error rules in [21.2.1.1](https://tc39.github.io/ecma262/#sec-patterns-static-semantics-early-errors). Not doing so was an oversight on my part, or perhaps I ran out of time.

The purpose of placing these errors into 21.2.1.1  (and with the early error static semantic errors, in general) is to clearly state the errors that identify malformed ES code that can be detected and reported independently of an run-time context or values.  The goal is to eliminate implementation variations in when such errors are reported.

allenwb:
Yes, I think it is probably fine.  For some reason, I thought I was seeing something different (new syntax error checks being added to RegExp evaluation rules) when I looked at the patch before sending my previous message

Gawd, I hate reading raw EMU and out of (global) context patch lines.

mathiasbynens:
[This PR gained consensus at the September 2017 TC39 meeting.](https://github.com/rwaldron/tc39-notes/blob/master/es8/2017-09/sep-26.md#12ii-early-errors-for-regexp-literals) Since the Unicode properties proposal is [blocked](https://github.com/tc39/proposal-regexp-unicode-property-escapes/issues/29) on this change, please merge this as soon as possible.

bterlson:
Sorry, I didn't know it was blocking. Happy to merge now. Result of investigation alluded to above: seems great. There are a few differences with what ChakraCore implements but are not concerning and I'll just track them as bugs :)

mathiasbynens:
Thanks!

spec.html
```diff
@@ -30152,12 +30247,12 @@ <h1>NonemptyClassRanges</h1>
           <h1>Runtime Semantics: CharacterRange ( _A_, _B_ )</h1>
           <p>The abstract operation CharacterRange takes two CharSet parameters _A_ and _B_ and performs the following steps:</p>
           <emu-alg>
-            1. If _A_ does not contain exactly one character or _B_ does not contain exactly one character, throw a *SyntaxError* exception.
+            1. Assert: _A_ and _B_ each contain exactly one character.
```
mathiasbynens: Doesn't e.g. `/[\p{Hex}-\uFFFF]/u` hit this case as well? The assert seems invalid in that case. What am I missing?

mathiasbynens: It doesn't because of https://github.com/tc39/test262/pull/1014/files/4843f049fafb6cdd61e3be011e3701cc94a9fe46#r200694312. Nothing to see here folks, move along! :)

PR #989, opened 2017-09-03
Editorial: fix trivial grammar error

xiaoyinl:
Fix a run-on sentence in Section 20.1.3.6.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -25634,7 +25634,7 @@
           1. Else, let _radixNumber_ be ? ToInteger(_radix_).
           1. If _radixNumber_ &lt; 2 or _radixNumber_ &gt; 36, throw a *RangeError* exception.
           1. If _radixNumber_ = 10, return ! ToString(_x_).
-          1. Return the String representation of this Number value using the radix specified by _radixNumber_. Letters `a`-`z` are used for digits with values 10 through 35. The precise algorithm is implementation-dependent; however the algorithm should be a generalization of that specified in <emu-xref href="#sec-tostring-applied-to-the-number-type"></emu-xref>.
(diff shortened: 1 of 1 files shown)


ConnerAiken:
Why not, "... dependent. However, xyz."

xiaoyinl:
@ConnerAiken I made the change you suggested. Thanks!
@ljharb Thanks for reviewing! If I remember correctly, "however" can't be used as a conjunction in this way in U.S. English, so I think it's better to change the punctuation.

ljharb:
@xiaoyinl http://www.quickanddirtytips.com/education/grammar/starting-a-sentence-with-however-right-or-wrong

ConnerAiken:
Exactly. "everyone knows it means 'nevertheless.'" .

littledan:
@ljharb The link you provided seems to say that it's fine to start a sentence with 'however'. I usually prefer the style in this PR. Long sentences, generally, can be hard to follow.

ljharb:
Totally, the link tells me that both styles are acceptable. I have no aesthetic preference for either.

bterlson:
"However" in this case is a conjunctive adverb and thus the sentence is not a run-on technically speaking, however (😀) whether it's too long is up to stylistic choice. I find the existing wording more clear personally. It's obvious "however" is qualifying the previous part of the sentence (with the PR, the however leaves me wondering, what are you howevering exactly?). Also, using However at the beginning of the sentence introduces an ambiguity about the definition of however until you reach the comma. Lastly, while I am more of a descriptivist, I recall style guides suggesting that starting a sentence with However when you mean "nevertheless".

@xiaoyinl thank you for raising this issue, and sorry for closing it. Please feel free to continue suggesting wording updates though!

PR #990, opened 2017-09-04
Editorial: Let "the Number value" be defined for 0

littledan:
This editorial change defines the Number value of 0 to be the IEEE 754
float +0. The previous definition was already accurate for this purpose;
the only change made here is to remove the requirement that "the Number
value" be applied to nonzero values. This change will make it easier
to specify conversions in the BigInt and WebAssembly specifications.

Closes https://github.com/tc39/proposal-bigint/issues/62


littledan:
This change is editorial because nowhere in the specification is "the Number value" of 0 used. It was previously undefined, and this PR makes it defined.

PR #994, opened 2017-09-07
Editorial: Clean up Abstract Relational Comparison

shvaikalesh:


spec.html
```diff
@@ -4339,10 +4340,13 @@
           1. If _ny_ is *+&infin;*, return *true*.
           1. If _ny_ is *-&infin;*, return *false*.
           1. If _nx_ is *-&infin;*, return *true*.
-          1. If the mathematical value of _nx_ is less than the mathematical value of _ny_ &mdash;note that these mathematical values are both finite and not both zero&mdash;return *true*. Otherwise, return *false*.
+          1. If the mathematical value of _nx_ is less than the mathematical value of _ny_ &mdash; note that these mathematical values are both finite and not both zero &mdash; return *true*. Otherwise, return *false*.
       </emu-alg>
       <emu-note>
-        <p>Step 3 differs from step 7 in the algorithm for the addition operator `+` (<emu-xref href="#sec-addition-operator-plus"></emu-xref>) in using &ldquo;and&rdquo; instead of &ldquo;or&rdquo;.</p>
+        <p>A String value _p_ is a prefix of String value _q_ if _q_ can be the string-concatenation of _p_ and some other String _r_. Note that any String is a prefix of itself, because _r_ may be the empty String.</p>
```
bterlson: I think we should normatively define prefix like the previous text. It's important I think!

spec.html
```diff
@@ -4339,10 +4340,13 @@
           1. If _ny_ is *+&infin;*, return *true*.
           1. If _ny_ is *-&infin;*, return *false*.
           1. If _nx_ is *-&infin;*, return *true*.
-          1. If the mathematical value of _nx_ is less than the mathematical value of _ny_ &mdash;note that these mathematical values are both finite and not both zero&mdash;return *true*. Otherwise, return *false*.
+          1. If the mathematical value of _nx_ is less than the mathematical value of _ny_ &mdash; note that these mathematical values are both finite and not both zero &mdash; return *true*. Otherwise, return *false*.
```
bterlson: The emdash should not have spaces around it (delete the extra space after `_ny_` instead).

spec.html
```diff
@@ -4253,6 +4253,17 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-isstringprefix" aoid="IsStringPrefix">
+      <h1>IsStringPrefix ( _p_, _q_ )</h1>
```
ljharb: I wonder if this abstract operation could be used in other places too, like `.includes` or `.startsWith`?

PR #997, opened 2017-09-08
Editorial: Use IsConstructor instead of checking [[Constructor]]

shvaikalesh:
Benefits: brevity, expressiveness, and consistency with other *constructor* checks.

With this change I was able to see all `IsConstructor` calls in references and make a [JS implementation](https://github.com/shvaikalesh/is-constructor/blob/master/index.js).


domenic:
(Minor nit: commit message is confusing; I'd say "instead of checking [[Constructor]]")

PR #1000, opened 2017-09-12
Adding explicit number conversion to %TypedArray%.prototype.sort

natashenka:
Clarifying that conversion to a number from comparefn should occur before the detachment check. This ambiguity led to this security problem: https://bugs.chromium.org/p/project-zero/issues/detail?id=983

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33221,7 +33221,7 @@ THH:mm:ss.sss
         <emu-alg>
           1. Assert: Both Type(_x_) and Type(_y_) is Number.
           1. If _comparefn_ is not *undefined*, then
-            1. Let _v_ be ToNumber(? Call(_comparefn_, *undefined*, &laquo; _x_, _y_ &raquo;)).
+            1. Let _v_ be ? ToNumber(? Call(_comparefn_, *undefined*, &laquo; _x_, _y_ &raquo;)).
             1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
             1. If _v_ is *NaN*, return *+0*.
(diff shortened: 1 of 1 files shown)


anba:
Can you merge this into a single step:
> Let _v_ be ? ToNumber(? Call(_comparefn_, *undefined*, &laquo; _x_, _y_ &raquo;)).

for consistency with the SortCompare algorithm in [22.1.3.25.1](https://tc39.github.io/ecma262/#sec-sortcompare)? Thanks!

spec.html
```diff
@@ -33221,7 +33221,7 @@
         <emu-alg>
           1. Assert: Both Type(_x_) and Type(_y_) is Number.
           1. If _comparefn_ is not *undefined*, then
-            1. Let _v_ be ? Call(_comparefn_, *undefined*, &laquo; _x_, _y_ &raquo;).
+            1. Let _v_ be ToNumber(? Call(_comparefn_, *undefined*, &laquo; _x_, _y_ &raquo;)).
```
ljharb: the `ToNumber` needs a `?` prefix as well, since it can throw

natashenka: Done

ljharb:
Does this need consensus?

allenwb:
shouldn't need consensus.  It's pretty clearly a bug that Array sort has a ToNumber but TypedArray sort does not.

bterlson:
Yep, seems like a bug to me (and a bad one, at that). Thanks!

bterlson:
@natashenka Congratulations on your first commit to ECMA262! 👌👍

PR #1001, opened 2017-09-16
Alternative of SLICE function.

Tech-Engineer:
Please review and let me know your thoughts on my proposal for a new way of finding sub array in javascript.


michaelficarra:
This is not the appropriate place for a proposal. See https://github.com/tc39/ecma262/blob/master/CONTRIBUTING.md for how to contribute.

PR #1004, opened 2017-09-18
Editorial: Add '(API)' after 'application programming interface'.

dilijev:
ECMA 262 Section 2: writes out "application programming interface" without also
writing "(API)" which is relevant because "API" is used throughout the rest of
the document and "(API)" is never made explicit before it's use.

API is a common acronym, so we could rewrite the long-form here to simply "API"
or add "(API)" since this is the first instance (which would be my preference).

Note that this section explicitly references ECMA 402 in the references below
that section, the title of which document is written in ECMA 262 as "ECMA-402,
ECMAScript 2015 Internationalization API Specification".


dilijev:
@ljharb lol thanks =P

bterlson:
Looks good to me (LGTM).

michaelficarra:
I wish I could've responded to this before it was merged. How about we use an [`<abbr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr)  to introduce the term instead?

bterlson:
@michaelficarra we don't use abbr anywhere so this aligns with current practices. I'd entertain a PR adding abbr to all the abbreviations we use, though...

dilijev:
Probably a good idea to also have the long-form introduction of `<abbr>`'d terms so that when it's rendered as PDF it's still readable in print form, with no loss of information.

michaelficarra:
@dilijev I'm sure that's something that can be done with some fancy CSS or very minimal JS. Just expand the first `<abbr>` for each term (if there's none marked as `<dfn>`) into its title followed by its contents in parens.

PR #1006, opened 2017-09-22
Do not remember module instantiation errors.

GeorgNeis:
A while ago, we changed the specification of modules such that instantiation and evaluation errors are recorded and then re-thrown by later instantiation/evaluation attempts. We have since realized that, in the case of instantiation errors, this was not a good idea, as it does not give us the kind of determinacy that we want.

The problem has to do with instantiating multiple "root" modules in a non-deterministic order, which happens e.g. due to multiple `<script type=module ...>` elements or multiple `import()` expressions. What we want to guarantee is at least "blame determinism", i.e., no matter
at which point an instantiation happens, if it fails then the error always originates from the same importing statement.

In cyclic graphs, having a failed instantiation record the error is in conflict with this goal. Consider this example:

```
<!-- index.html -->
<script type="module" src="./A.js"></script>
<script type="module" src="./B.js"></script>
```
```
// A.js
import "./B.js"
import {something} from "./empty.js"
```
```
// B.js
import "./A.js";
import {something_else} from "./empty.js"
```

If A gets instantiated first, both script loads result in an error blaming B's import of "something_else". If B gets instantiated first, both loads result in an error blaming A's import of "something".

Our solution is simply to not record instantiation errors but instead reset a module to "uninstantiated" if instantiation fails. In the example above, the load of A will then result in an error blaming B's import of "something_else", and the load of B will result in an error blaming A's import of "something" --- no matter which one gets instantiated first. (The error objects are not identical. While it's possible to enforce even this, it would require significant extra complexity, which
is why we decided against it.) 

The treatment of evaluation errors remains unchanged: unlike instantiation, evaluation must not happen more than once, and is not idempotent, so we definitely need to record the error there. (See `launchTheMissiles()` in https://github.com/tc39/ecma262/issues/862.)

Summary of the changes proposed in this PR:

Main changes:
- Instantiation errors are no longer remembered.
- Instantiation failure resets module to status "uninstantiated".
- Since the only recorded errors are now evaluation errors, instantiation can treat errored modules like instantiated modules.

Minor changes include:
- Reverted ResolveExport to what it was originally.
- Renamed [[ErrorCompletion]] to [[EvaluationError]].
- Replace "errored" status by "evaluated" with non-undefined
[[EvaluationError]].
- Moved [[Status]] and [[EvaluationError]] from Abstract MR to Source Text MR.
- Renamed InnerModuleDeclarationInstantiation to InnerModuleInstantiation.

Related pull request for HTML: https://github.com/whatwg/html/pull/2991


allenwb:
😀

bterlson:
Give me a little bit to read through the diff but the approach seems fine.

bterlson:
This looks good. Especially appreciate the updated copy in 15.2.1.16.6 Example Source Text Module Record Graphs. I introduced a merge conflict - will attempt to resolve soon unless someone beats me to it.

bterlson:
Alright I think I rebased correctly. @GeorgNeis @domenic could one of you take a quick look and make sure I didn't hose anything?

domenic:
LGTM

PR #1009, opened 2017-09-22
Normative: Remove detached array buffer check prior to calling CloneArrayBuffer

### LINKED ISSUE #1002: Let's remove the IsDetachedBuffer check in 22.2.4.3, step 18.a
I didn't remove the step in #844 to avoid normative changes in that PR, but since this check is only implemented correctly in SpiderMonkey and CloneArrayBuffer will again check for detached buffers, it doesn't seem terrible useful to me to test for detached buffers in this step. 

Test case for the current spec.
```js
function detach(ab) {
  if (ArrayBuffer.transfer) {
    ArrayBuffer.transfer(ab);
    detachArrayBuffer(ab);
    transferArrayBuffer(ab)
    try { eval("%ArrayBufferNeuter(ab)") } catch (e) {
      var w = new Worker("");
      w.postMessage(ab, [ab]);
      w.terminate();
    }
  } else {
  }
}

function test(from, to) {
    var ta = new from(0);
    ta.buffer.constructor = {
        get [Symbol.species]() {
            print("get @@species");
            detach(ta.buffer);
            return Object.defineProperty(function(){}.bind(), "prototype", {
                get() {
                    print("get prototype");
                    return null;
                }
            });
        }
    };
    try {
        new to(ta);
    } catch (e) {
        print("caught:", e);
        return;
    }
    print("No error");
}

// Expected: Prints "get @@species" and then throws a TypeError
test(Int32Array, Int32Array);

// Expected: Prints "get @@species", "get prototype", and then throws a TypeError
test(Int32Array, Uint32Array);
```

anba:
This aligns the detached array buffer checks in step 18 and 19 of 22.2.4.3.

Fixes #1002


littledan:
It looks like this only affects the ordering of detached buffer errors and errors from running out of memory. Is that right? If so, this seems like a good idea. If I'm understanding things right, this would be pretty difficult to write a reliable test for (which size buffer will run out of memory when you clone it?) so I think we should waive the testing requirement here.

anba:
I can create a test for test262 based on the example code in #1002.

anba:
Consensus was reached per https://github.com/tc39/tc39-notes/blob/master/es8/2017-11/nov-30.md#8ia-arraybuffer-error-reporting-tweak-by-andre-bargull

littledan:
Is there anything holding back this PR landing?

anba:
The change is also implemented in Firefox Nightly and will ship in Firefox 60 (https://bugzilla.mozilla.org/show_bug.cgi?id=1423878).

PR #1010, opened 2017-09-22
Resolve #1007 - say what the draft reprsents

### LINKED ISSUE #1007: Say what the draft represents
I understand from Domenic Denicola that the live copy at https://tc39.github.io/ecma262/ contains only previously-specified features and Stage 4 proposals (e.g., finished ones). Let's say that clearly up front.

tjcrowder:
Fixes #1007.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -74,6 +74,8 @@
     </li>
   </ul>
   <p>Refer to the <emu-xref href="#sec-colophon">colophon</emu-xref> for more information on how this document is created.</p>
+  <h1>About this Specification</h1>
+  <p>This document at <a href=".">https://tc39.es/ecma262/</a> is the most accurate up-to-date ECMAScript specification. It contains the content of the most recent yearly snapshot plus any <a href="https://github.com/tc39/proposals/blob/master/finished-proposals.md">finished proposals</a> (those that have reached Stage&nbsp;4 in the <a href="https://tc39.es/process-document/">proposal process</a> and thus are implemented in several implementations and will be in the next practical revision) since that snapshot was taken.</p>
 </div>
 <emu-intro id="sec-intro">
   <h1>Introduction</h1>


rwaldron:
This means that @bterlson has to remember to delete this section when for the version that gets voted on by the GA every June

allenwb:
Sure does, there is a whole set of steps required for Ecma publication.  That's part of what makes in an actual _standard_ rather than a spec. some random people decided to write and put out there.

(BTW, sometimes we've screwed up some of those steps.  we probably need a check list)

tjcrowder:
@rwaldron - Yes -- to allow for that, in the PR I made sure to put it in the `id=metadata-block` that's already removed for snapshots, so it wouldn't be an extra step.

littledan:
A couple nits for the text:
- It includes not just Stage 4 proposals but also things written as patches without going through the stage process.
- We're generally careful to avoid the word 'browser' in most places like this. How about 'implementation'?
- You could mention that we don't publish errata against older spec snapshots and bugs are fixed here (which gives meaning to 'most accurate').

rwaldron:
> I made sure to put it in the id=metadata-block that's already removed for snapshots, so it wouldn't be an extra step.

Ah, I actually didn't see that because: 



Thanks!

tjcrowder:
@littledan - Good points all, I can't believe I wrote "browser" there, I'm normally quite picky about *not* saying "browser." "Implementation" sounds good.

Also useful comments over on [the issue itself](https://github.com/tc39/ecma262/issues/1007) if anyone here wants to weigh in on them. Later this week I'll update the PR with changes from both here and there.

ljharb:
Are you no longer interested in the underlying issue, or just in this PR?

tjcrowder:
@ljharb - *sigh* Embarrassingly, I'd misremembered that this had been rejected declined when in fact it's still open. Note to self: Don't overdo cleanup. What's my best course of action here?

ljharb:
Undelete the branch and reopen the PR?

tjcrowder:
@ljharb - Thanks again. I thought I'd deleted my repo, but you're right, I just deleted the branch. *double sigh* I'm not an incompetent, I just play one on github sometimes... :-)

PR #1013, opened 2017-09-28
Add %JSONParse% for Fetch et al

### LINKED ISSUE #1012: Expose initial JSON.parse as %JSONParse%
That way Fetch and XMLHttpRequest can use it more directly (and other web platform entry points that need JSON parsing). Alternative is an abstract operation, which might be a little clearer usage wise.

Any preference?

annevk:
Fixes #1012.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2220,6 +2220,17 @@
             </tr>
             <tr>
               <td>
+                %JSONParse%
+              </td>
+              <td>
+                `JSON.parse`
+              </td>
+              <td>
+                The initial value of the `parse` data property of %JSON%
+              </td>
+            </tr>
+            <tr>
+              <td>
                 %Map%
               </td>
               <td>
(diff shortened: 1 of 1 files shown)


annevk:
Not mentioning it in prose was modeled after %Int32ArrayPrototype%, though I notice that for %ObjProto_valueOf% we do mention it in prose:

> This function is the %ObjProto_valueOf% intrinsic object.

The latter probably also has an ID therefore. Was not mentioning the former an oversight?

(I also noticed some other inconsistencies for which I'll submit a separate PR.)

ljharb:
I was wondering about that; I think it's better to have a dual link, and to mention it in both places (for all intrinsics).

annevk:
Added here for JSON.parse. Per https://tc39.github.io/ecma262/#sec-well-known-intrinsic-objects there's quite a few not linked. I'm happy to file a follow-up issue for those, but I'm not going to work on a PR for now.

domenic:
See https://github.com/tc39/ecma262/issues/1012#issuecomment-332867155 for a potential slightly cleaner alternative. But either will work. I'll try to ask @bterlson in person today.

bterlson:
I selfishly prefer this approach as it's cleaner in 262.

PR #1015, opened 2017-10-03
Decode doesn't need to check the reserved-set for non-ASCII characters

anba:
dc370115a779e34ec2998f46de6c87581152c215:
`reservedSet` contains only ASCII characters, so we don't need to check for `reservedSet` when decoding an escaped character. Passing a decoded ASCII character throws a URIError because of the checks in 4.d.vii.2 and 4.d.vii.8. Also see the note below Decode:
> RFC 3629 prohibits the decoding of invalid UTF-8 octet sequences. For example, the invalid sequence C0 80 must not decode into the code unit 0x0000. Implementations of the Decode algorithm are required to throw a URIError when encountering such invalid sequences.

c2b368efe45bbc51cc7ba4a81210299e8443beb3:
Replaces the non-defined term `array` with `List` and simplifies a loop to use list iteration.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -23575,13 +23575,10 @@
                   1. Let _kChar_ be the numeric value of the code unit at index _k_ within _string_.
                   1. If _kChar_ is less than 0xDC00 or greater than 0xDFFF, throw a *URIError* exception.
                   1. Let _V_ be UTF16Decode(_C_, _kChar_).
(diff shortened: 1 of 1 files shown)


bterlson:
Well now if we ever add anything into reservedSet (or another terminal to _uriReserved_) that is non-ascii we'll have to remember to restore this logic!! /s

Love this change, thank you.

PR #1016, opened 2017-10-04
Editorial: Use HTTPS for unicode.org URLs

mathiasbynens:
unicode.org is now available over HTTPS. Let’s make use of it.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -28433,7 +28433,7 @@ THH:mm:ss.sss
           <p>The `localeCompare` method itself is not directly suitable as an argument to `Array.prototype.sort` because the latter requires a function of two arguments.</p>
         </emu-note>
         <emu-note>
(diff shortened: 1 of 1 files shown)


annevk:
Is it appropriate to remove `www.` here? The couple non-relative links they have include it.

mathiasbynens:
What does that mean, though? Their non-relative links don’t include HTTPS either. If we want to stick to whatever they do, we should reject this PR.

annevk:
Well, generally when you go from HTTP to HTTPS you don't change the domain (unless it happens to redirect).

ljharb:
If they 301 from www to no-www, they should be updated in the text to match the destination; if they 302, they should retain the www. (I can't check rn; am on mobile).

annevk:
There's no redirects at play here. They just made HTTPS available.

ljharb:
In that case it makes sense to only change the protocol

bterlson:
I'll take this as-is having validated the more efficient www-less links work properly :)

PR #1025, opened 2017-10-31
The DataView constructor can return DVs backed by detached ArrayBuffers

jswalden:
Two ways this can happen.  One, "Let viewByteLength be ? ToIndex(byteLength)." step 9a could detach the already-checked buffer.  And two, "Let O be ? OrdinaryCreateFromConstructor(NewTarget, "%DataViewPrototype%", « [[DataView]], [[ViewedArrayBuffer]], [[ByteLength]], [[ByteOffset]] »)." step 10 could do so.

The super-easy fix, that alters the existing order of checks and potential side effects the least, is to doubly check for detachment by duplicating step 5 as a new step 10.5.

A more complicated fix, that would be consistent with how typed arrays handle the similar problem, but that would reorder some of the checking and side effects, would be to move the single `IsDetachedArrayBuffer` check to the end of the algorithm and adjust the other steps' ordering as needed.  Finicky, but doable.

I don't like creating newborn-and-incompletely-initialized objects, then putting them aside for other operations for a bit, as the typed array constructor algorithm does right now.  But avoiding doing that in `DataView` probably would warrant also doing so for typed arrays, which would mean *both* algorithms would have to change.  That might not be desirable.

This PR does the super-easy fix.  Is that the fix we want?

This is a followup to both [Mozilla bug 1285960](https://bugzilla.mozilla.org/show_bug.cgi?id=1285960#c5) and, it so happens, https://github.com/tc39/ecma262/pull/852#issuecomment-291791835 as well.  (I started from the former and was pointed at the latter.)


littledan:
The change makes sense to me, and fits with the rest of the design where we're trying to prevent this sort of thing from happening.

anba:
Consensus was reached per https://github.com/tc39/tc39-notes/blob/master/es8/2017-11/nov-30.md#8ib-avoid-detached-dataview-in-edge-case-by-jeff-walden

ljharb:
@jswalden would you be able to file a PR to test262 with tests for this change?

jswalden:
Filed https://github.com/tc39/test262/pull/1504 for the test262 change.

rwaldron:
> Filed tc39/test262#1504 for the test262 change.

Confirm: this was merged

PR #1029, opened 2017-11-12
[Normative] Add RegExp lookbehind assertions

mathiasbynens:
@littledan asked me to prepare the PR for the proposal adding lookbehind assertions to ECMAScript regular expressions.

Proposal repo: https://github.com/tc39/proposal-regexp-lookbehind by Gorkem Yakin, Nozomu Katō, and @littledan.

@littledan, I’ve added you as the commit author — I hope I’m not misrepresenting you :)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -29113,6 +29113,8 @@ THH:mm:ss.sss
           `\` `B`
           `(` `?` `=` Disjunction[?U] `)`
           `(` `?` `!` Disjunction[?U] `)`
+          `(` `?` `&lt;=` Disjunction[?U] `)`
+          `(` `?` `&lt;!` Disjunction[?U] `)`
 
         Quantifier ::
           QuantifierPrefix
@@ -29682,7 +29684,7 @@ THH:mm:ss.sss
         <h1>Pattern</h1>
         <p>The production <emu-grammar>Pattern :: Disjunction</emu-grammar> evaluates as follows:</p>
         <emu-alg>
-          1. Evaluate |Disjunction| to obtain a Matcher _m_.
+          1. Evaluate |Disjunction| with +1 as its _direction_ argument to obtain a Matcher _m_.
           1. Return an internal closure that takes two arguments, a String _str_ and an integer _index_, and performs the following steps:
             1. Assert: _index_ &le; the length of _str_.
             1. If _Unicode_ is *true*, let _Input_ be a List consisting of the sequence of code points of _str_ interpreted as a UTF-16 encoded (<emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>) Unicode string. Otherwise, let _Input_ be a List consisting of the sequence of code units that are the elements of _str_. _Input_ will be used throughout the algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref>. Each element of _Input_ is considered to be a character.
@@ -29701,6 +29703,7 @@ THH:mm:ss.sss
       <!-- es6num="21.2.2.3" -->
       <emu-clause id="sec-disjunction">
         <h1>Disjunction</h1>
+        <p>With argument _direction_.</p>
         <p>The production <emu-grammar>Disjunction :: Alternative</emu-grammar> evaluates as follows:</p>
         <emu-alg>
           1. Evaluate |Alternative| to obtain a Matcher _m_.
@@ -29708,8 +29711,8 @@ THH:mm:ss.sss
         </emu-alg>
         <p>The production <emu-grammar>Disjunction :: Alternative `|` Disjunction</emu-grammar> evaluates as follows:</p>
         <emu-alg>
-          1. Evaluate |Alternative| to obtain a Matcher _m1_.
-          1. Evaluate |Disjunction| to obtain a Matcher _m2_.
+          1. Evaluate |Alternative| with argument _direction_ to obtain a Matcher _m1_.
+          1. Evaluate |Disjunction| with argument _direction_ to obtain a Matcher _m2_.
           1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps when evaluated:
             1. Call _m1_(_x_, _c_) and let _r_ be its result.
             1. If _r_ is not ~failure~, return _r_.
@@ -29724,32 +29727,41 @@ THH:mm:ss.sss
           <pre><code class="javascript">["abc", "a", "a", undefined, "bc", undefined, "bc"]</code></pre>
           <p>and not</p>
           <pre><code class="javascript">["abc", "ab", undefined, "ab", "c", "c", undefined]</code></pre>
+          <p>The order in which the two alternatives are tried is independent of the value of _direction_.</p>
         </emu-note>
       </emu-clause>
 
       <!-- es6num="21.2.2.4" -->
       <emu-clause id="sec-alternative">
         <h1>Alternative</h1>
+        <p>With argument _direction_.</p>
         <p>The production <emu-grammar>Alternative :: [empty]</emu-grammar> evaluates as follows:</p>
         <emu-alg>
           1. Return a Matcher that takes two arguments, a State _x_ and a Continuation _c_, and returns the result of calling _c_(_x_).
         </emu-alg>
         <p>The production <emu-grammar>Alternative :: Alternative Term</emu-grammar> evaluates as follows:</p>
         <emu-alg>
-          1. Evaluate |Alternative| to obtain a Matcher _m1_.
-          1. Evaluate |Term| to obtain a Matcher _m2_.
-          1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps when evaluated:
-            1. Let _d_ be a Continuation that takes a State argument _y_ and returns the result of calling _m2_(_y_, _c_).
-            1. Call _m1_(_x_, _d_) and return its result.
+          1. Evaluate |Alternative| with argument _direction_ to obtain a Matcher _m1_.
+          1. Evaluate |Term| with argument _direction_ to obtain a Matcher _m2_.
+          1. If _direction_ is equal to +1, then
+            1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps when evaluated:
+              1. Let _d_ be a Continuation that takes a State argument _y_ and returns the result of calling _m2_(_y_, _c_).
+              1. Call _m1_(_x_, _d_) and return its result.
+          1. Else,
+            1. Assert: _direction_ is equal to -1.
+            1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps when evaluated:
+              1. Let _d_ be a Continuation that takes a State argument _y_ and returns the result of calling _m1_(_y_, _c_)
+              1. Call _m2_(_x_, _d_) and return its result.
         </emu-alg>
         <emu-note>
-          <p>Consecutive |Term|s try to simultaneously match consecutive portions of _Input_. If the left |Alternative|, the right |Term|, and the sequel of the regular expression all have choice points, all choices in the sequel are tried before moving on to the next choice in the right |Term|, and all choices in the right |Term| are tried before moving on to the next choice in the left |Alternative|.</p>
+          <p>Consecutive |Term|s try to simultaneously match consecutive portions of _Input_. When _direction_ is equal to +1, if the left |Alternative|, the right |Term|, and the sequel of the regular expression all have choice points, all choices in the sequel are tried before moving on to the next choice in the right |Term|, and all choices in the right |Term| are tried before moving on to the next choice in the left |Alternative|. When _direction_ is equal to -1, the evaluation order of |Alternative| and |Term| are reversed.</p>
         </emu-note>
       </emu-clause>
 
       <!-- es6num="21.2.2.5" -->
       <emu-clause id="sec-term">
         <h1>Term</h1>
+        <p>With argument _direction_.</p>
         <p>The production <emu-grammar>Term :: Assertion</emu-grammar> evaluates as follows:</p>
         <emu-alg>
           1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps when evaluated:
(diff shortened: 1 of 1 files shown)


littledan:
Thank you so much for writing this PR. To give credit where credit is due, the spec text here was written by @claudepache and @IgnoredAmbience . Review coming soon (maybe those folks will want to take a look as well).

spec.html
```diff
@@ -30057,62 +30091,68 @@
...
         </emu-alg>
         <p>The production <emu-grammar>Atom :: `(` `?` `:` Disjunction `)`</emu-grammar> evaluates as follows:</p>
         <emu-alg>
-          1. Return the Matcher that is the result of evaluating |Disjunction|.
+          1. Return the Matcher that is the result of evaluating |Disjunction| with argument _direction_.
         </emu-alg>
 
         <!-- es6num="21.2.2.8.1" -->
         <emu-clause id="sec-runtime-semantics-charactersetmatcher-abstract-operation" aoid="CharacterSetMatcher">
           <h1>Runtime Semantics: CharacterSetMatcher ( _A_, _invert_ )</h1>
```
littledan: Seems like the third argument should be added here.

spec.html
```diff
@@ -29571,32 +29572,40 @@
...
-          1. Evaluate |Alternative| to obtain a Matcher _m1_.
-          1. Evaluate |Term| to obtain a Matcher _m2_.
+          1. Evaluate |Alternative| with argument _direction_ to obtain a Matcher _m1_.
+          1. Evaluate |Term| with argument _direction_ to obtain a Matcher _m2_.
+          1. If _direction_ is equal to +1, then,
+            1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps when evaluated:
+              1. Let _d_ be a Continuation that takes a State argument _y_ and returns the result of calling _m2_(_y_, _c_).
+              1. Call _m1_(_x_, _d_) and return its result.
+          1. Else, _direction_ is equal to -1.
           1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps when evaluated:
```
littledan: Nit: Indent this line and the following two lines.

spec.html
mathiasbynens: Ack

spec.html
mathiasbynens: Ack

mathiasbynens:
Thanks for the thorough review, @littledan! Feedback addressed. 

> To give credit where credit is due, the spec text here was actually written by @claudepache and @IgnoredAmbience

Hmm… Should I change the commit author? Or does it make more sense to mention the authors in the commit message?

littledan:
I don't have a good idea of sharing credit here; these things are always the work of a lot of people. If I were you, I'd just leave Mathias as the commit author, and give callouts to @hashseed, @claudepache, @bterlson and everyone else who contributed when you get a chance, like in a blog post or talk. (This is a lot of work though!)

spec.html
```diff
@@ -30057,62 +30091,68 @@
...
           1. Let _parenIndex_ be the number of left-capturing parentheses in the entire regular expression that occur to the left of this |Atom|. This is the total number of <emu-grammar>Atom :: `(` Disjunction `)`</emu-grammar> Parse Nodes prior to or enclosing this |Atom|.
           1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps:
             1. Let _d_ be an internal Continuation closure that takes one State argument _y_ and performs the following steps:
               1. Let _cap_ be a copy of _y_'s _captures_ List.
               1. Let _xe_ be _x_'s _endIndex_.
               1. Let _ye_ be _y_'s _endIndex_.
-              1. Let _s_ be a new List whose characters are the characters of _Input_ at indices _xe_ (inclusive) through _ye_ (exclusive).
+              1. If _direction_ is equal +1, then
+                1. Let _s_ be a fresh List whose characters are the characters of _Input_ at indices _xe_ (inclusive) through _ye_ (exclusive).
+              1. Else, let _direction_ is equal to -1.
```
littledan: This wording is somewhat unusual. Maybe better wording would be, "

1. If _direction_ is equal **to** +1, then
     1. Let _s_ be a fresh List whose characters are the characters of _Input_ at indices _xe_ (inclusive) through _ye_ (exclusive).
1. Else,
    1. **Assert:** _direction_ is equal to -1
    1. Let _s_ be a fresh List whose characters are the characters of _Input_ at indices _ye_ (inclusive) through _xe_ (exclusive).

claudepache: > This wording is somewhat unusual.

I would even say, it’s plain typos: “If _direction_ is equal <ins>to</ins> +1” and “Else, <del>let</del> _direction_ is equal to -1”.

spec.html
```diff
@@ -29571,32 +29572,40 @@
...
         <p>The production <emu-grammar>Alternative :: Alternative Term</emu-grammar> evaluates as follows:</p>
         <emu-alg>
-          1. Evaluate |Alternative| to obtain a Matcher _m1_.
-          1. Evaluate |Term| to obtain a Matcher _m2_.
-          1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps when evaluated:
-            1. Let _d_ be a Continuation that takes a State argument _y_ and returns the result of calling _m2_(_y_, _c_).
-            1. Call _m1_(_x_, _d_) and return its result.
+          1. Evaluate |Alternative| with argument _direction_ to obtain a Matcher _m1_.
+          1. Evaluate |Term| with argument _direction_ to obtain a Matcher _m2_.
+          1. If _direction_ is equal to +1, then,
```
claudepache: no comma after “then”

spec.html
```diff
@@ -29571,32 +29572,40 @@
...
+            1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps when evaluated:
+              1. Let _d_ be a Continuation that takes a State argument _y_ and returns the result of calling _m2_(_y_, _c_).
+              1. Call _m1_(_x_, _d_) and return its result.
+          1. Else, _direction_ is equal to -1.
```
claudepache: “Else _direction_ is equal to -1,” (no comma after “else”, a comma at the end of the line).
Alternatively:

1. Else,
    1. Assert: _direction_ is equal to -1.

spec.html
```diff
@@ -30057,62 +30091,68 @@
...
-          1. Evaluate |Disjunction| to obtain a Matcher _m_.
+          1. Evaluate |Disjunction| with argument _direction_ to obtain a Matcher _m_.
           1. Let _parenIndex_ be the number of left-capturing parentheses in the entire regular expression that occur to the left of this |Atom|. This is the total number of <emu-grammar>Atom :: `(` Disjunction `)`</emu-grammar> Parse Nodes prior to or enclosing this |Atom|.
           1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps:
             1. Let _d_ be an internal Continuation closure that takes one State argument _y_ and performs the following steps:
               1. Let _cap_ be a copy of _y_'s _captures_ List.
               1. Let _xe_ be _x_'s _endIndex_.
               1. Let _ye_ be _y_'s _endIndex_.
-              1. Let _s_ be a new List whose characters are the characters of _Input_ at indices _xe_ (inclusive) through _ye_ (exclusive).
+              1. If _direction_ is equal +1, then
```
claudepache: “is equal <ins>to</ins> +1,”

spec.html
claudepache: “Else _direction_ is equal to -1,” (no comma after “else”, no extraneous “let”, a comma at the end of the line).
Alternatively:

1. Else,
    1. Assert: _direction_ is equal to -1.

mathiasbynens:
@claudepache Fixed!

spec.html
mathiasbynens: @claudepache Fixed!

wangyi7099:
I also expect this feature very much!!!

mathiasbynens:
@claudepache I’ve addressed your feedback. Please take another look and update your review status if everything seems okay now.

littledan:
LGTM

spec.html
```diff
@@ -30061,62 +30096,69 @@ <h1>Quantifier</h1>
...
           1. Let _parenIndex_ be the number of left-capturing parentheses in the entire regular expression that occur to the left of this |Atom|. This is the total number of <emu-grammar>Atom :: `(` Disjunction `)`</emu-grammar> Parse Nodes prior to or enclosing this |Atom|.
           1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps:
             1. Let _d_ be an internal Continuation closure that takes one State argument _y_ and performs the following steps:
               1. Let _cap_ be a copy of _y_'s _captures_ List.
               1. Let _xe_ be _x_'s _endIndex_.
               1. Let _ye_ be _y_'s _endIndex_.
-              1. Let _s_ be a new List whose characters are the characters of _Input_ at indices _xe_ (inclusive) through _ye_ (exclusive).
+              1. If _direction_ is equal to +1, then
+                1. Let _s_ be a fresh List whose characters are the characters of _Input_ at indices _xe_ (inclusive) through _ye_ (exclusive).
+              1. Else, let _direction_ is equal to -1.
```
claudepache: Has to be:

    1. Else,
        1. Assert: _direction_ is equal to -1.

Optionally: It is implicit in the formulation of these steps that the numerical order of _xe_ and _ye_ is bound to the value of _direction_ (i.e., _xe_ ≤ _ye_ if _direction_ is +1, etc.) I leave to your judgment to decide whether it adds value to make that relation explicit, i.e.

        1. Assert: _xe_ ≤ _ye_.
        1. Let _s_ be a fresh List whose characters are the characters of _Input_ at indices _xe_ (inclusive) through _ye_ (exclusive).
    1. Else,
        1. Assert: _direction_ is equal to -1.
        1. Assert: _ye_ ≤ _xe_.
        1. Let _s_ be a fresh List whose characters are the characters of _Input_ at indices _ye_ (inclusive) through _xe_ (exclusive).

spec.html
```diff
@@ -29575,32 +29576,41 @@ <h1>Disjunction</h1>
...
+              1. Let _d_ be a Continuation that takes a State argument _y_ and returns the result of calling _m1_(_y_, _c_)
+              1. Call _m2_(_x_, _d_) and return its result.
         </emu-alg>
         <emu-note>
-          <p>Consecutive |Term|s try to simultaneously match consecutive portions of _Input_. If the left |Alternative|, the right |Term|, and the sequel of the regular expression all have choice points, all choices in the sequel are tried before moving on to the next choice in the right |Term|, and all choices in the right |Term| are tried before moving on to the next choice in the left |Alternative|.</p>
+          <p>Consecutive |Term|s try to simultaneously match consecutive portions of _Input_. When _direction_ is equal to +1, if the left |Alternative|, the right |Term|, and the sequel of the regular expression all have choice points, all choices in the sequel are tried before moving on to the next choice in the right |Term|, and all choices in the right |Term| are tried before moving on to the next choice in the left |Alternative|. When _direction_ is equal to -1, the evaluation order of |Alternative| and |Term| are reversed.</p>
```
ljharb: is it clear here to use "+1" and "-1", and then also to use "left" and "right", which might not be correct in an RTL context?

spec.html
mathiasbynens: Done, and done. I agree being explicit makes it easier to follow.

spec.html
mathiasbynens: IMHO it’s clear that this refers to the grammar productions, which are always in a left-to-right order.

ljharb: Instead of -1/+1, what would you think about using the strings "left" and "right"? (obv they're merely spec artifacts either way)

claudepache: > Instead of -1/+1, what would you think about using the strings "left" and "right"?

No, it is not “left” and “right”, it is “in the direction of the beginning of the string” (-1) and “in the direction of the end of the string” (+1), regardless of the directionality of its characters.

mathiasbynens: Do you think that would be more clear? I’m in favor of making things more readable but personally don’t see how this change does that.

mathiasbynens: @claudepache This goes back to @ljharb’s first comment. IMHO it’s clear that directionality of the actual characters doesn’t play a role as we’re just talking about grammar productions here. Do you disagree?

ljharb: - I think that -1/+1 is confusing when the value isn't used directly in a math operation
 - i think "left" and "right" can be problematic and confusing, even when it makes sense (as we've seen with pad/trim/reduce)
 - I don't have better words for the direction argument in mind :-/ maybe "advance" (to go with AdvanceStringIndex) and "regress" (or some other antonym for "advance")?

anba: @ljharb Maybe `backwards` and `forwards`? Similar to the usage in the `Intl.Segmenter` proposal.

ljharb: Wow, that's way simpler - yeah that would make it much clearer for me.

(note that i'm not trying to block anything; just sharing my thoughts prior to merging. I wouldn't want to delay this landing :-) )

claudepache: Yes, `~forwards~` and `~backwards~` (in other parts of the spec, we have `~failure~`, `~empty~`) is probably better than `+1` and `-1`.

IgnoredAmbience:
Note: my effort in this PR was to port the changes from an older version of the spec to the most recent version for the proposal. I've no in-depth views on the content, unfortunately - hence removing myself from the review list.

spec.html
ljharb: @littledan if a regex is in unicode mode, it might advance +2 when it finds a surrogate pair, no?

littledan: (It'll only be advancing +-1 code point...) I don't feel strongly about this. I'm fine with switching to `~forwards~` and `~backwards~`.

claudepache: > if a regex is in unicode mode, it might advance +2 when it finds a surrogate pair, no?

In the context of RegExp matching, “one step forth/back” is always literally +1/-1 in terms of “indexing” the string. The difference is that, in the presence of the u-flag, the RegExp algorithm use an alternative way to “index” the string.

If I had to rewrite the algorithms now, I would probably use “forwards/backwards” instead of “+1/-1”, because it is more descriptive. But I don’t think it adds much clarity, because +1/-1 couldn’t mean anything else... unless you ask yourself inappropriate questions regarding RTL and/or astral characters, in which case you’ll have trouble to understand the algorithms anyway.

mathiasbynens:
Rebased against `master`; this involved rewriting the patch to change the (new) `BackreferenceMatcher`.

mysticatea:
I have a question here.

This PR didn't look to update [B.1.4 Regular Expressions Patterns](https://www.ecma-international.org/ecma-262/8.0/#sec-regular-expressions-patterns) section. Is there a plan to update the section?

PR #1034, opened 2017-11-21
Normative: Web Reality (4/4 engines): _day_ before _month_ in Date.prototype.toUTCString

dilijev:
Note: the other instance of _month_ before _day_ (in #sec-datestring) is correct w.r.t. web reality (4/4).

See the test case below for the web reality (4/4 engines) output for all 4 `Date.prototype.to*String()` method outputs:

    ## Source
    print(new Date('2008-10-13T05:16:33Z').toUTCString());
    print(new Date('2008-10-13T05:16:33Z').toString());
    print(new Date('2008-10-13T05:16:33Z').toDateString());
    print(new Date('2008-10-13T05:16:33Z').toTimeString());

    #### ch, d8, jsc, sm
    Mon, 13 Oct 2008 05:16:33 GMT
    Sun Oct 12 2008 22:16:33 GMT-0700 (Pacific Daylight Time)
    Sun Oct 12 2008
    22:16:33 GMT-0700 (Pacific Daylight Time)

See discussion at https://github.com/Microsoft/ChakraCore/pull/4067#issuecomment-345651062

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -28012,7 +28012,7 @@ THH:mm:ss.sss
           1. Let _month_ be the Name of the entry in <emu-xref href="#sec-todatestring-month-names"></emu-xref> with the Number MonthFromTime(_tv_).
           1. Let _day_ be the String representation of DateFromTime(_tv_), formatted as a two-digit number, padded to the left with a zero if necessary.
           1. Let _year_ be the String representation of YearFromTime(_tv_), formatted as a number of at least four digits, padded to the left with zeroes if necessary.
-          1. Return the string-concatenation of _weekday_, `","`, the code unit 0x0020 (SPACE), _month_, the code unit 0x0020 (SPACE), _day_, the code unit 0x0020 (SPACE), _year_, the code unit 0x0020 (SPACE), and TimeString(_tv_).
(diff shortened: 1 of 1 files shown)


dilijev:
Before merging, note that I'm looking into more inconsistencies with web reality in this area.

littledan:
Thank you for this patch! I wrote the spec text relatively recently, hoping to meet web reality. I believe this was a typo--I remember working to factor the text to allow for differences between month and day, but I guess I got it wrong in the end.

If you could also update test262 tests for this change, that would be great. The test might actually exist for this, but it would be good to double check it.

dilijev:
@littledan: I'm wondering why some of the date functions in the spec are runtime semantics and some are not -- they all seem to have algorithm steps

For example https://tc39.github.io/ecma262/#sec-todatestring versus https://tc39.github.io/ecma262/#sec-datestring

I understand the naming is just that one is an abstract operation and the other is not but why is ToDateString a runtime semantics thing (called from https://tc39.github.io/ecma262/#sec-date.prototype.tostring)

bterlson:
Nice find, and thanks for the patch!

dilijev:
@littledan I'm done puzzling out the other issues that had occurred to me. You can merge this whenever you're ready.

dilijev:
@bterlson: credit to @xiaoyinl for the find! :D

PR #1037, opened 2017-11-21
Editorial: Add missing 'The following steps are performed:' before several Date algorithms.

dilijev:
/cc @littledan


littledan:
I'm fine with this change, though I left that text out since I don't see what it adds.

dilijev:
@littledan fair enough, it just looked inconsistent since there are some sections that only say that and then have an algorithm. I wasn't sure if it was a spec convention or not, but it was an editorial consistency issue.

PR #1038, opened 2017-11-24
Es2017

alaaadnan:
<script type="module">
  import * as module from './utils.mjs';
  module.default();
  // → logs 'Hi from the default export!'
  module.doStuff();
  // → logs 'Doing stuff…'
</script>


ljharb:
@alaaadnan this seems like it might be a mistaken PR from an already-merged branch. I'm going to close it for now.

PR #1041, opened 2017-12-05
[Normative] Add RegExp Unicode property escapes

mathiasbynens:
Proposal repo: https://github.com/tc39/proposal-regexp-unicode-property-escapes


littledan:
LGTM

PR #1045, opened 2017-12-08
Normative: built-in functions are not allowed to have caller/arguments properties

bterlson:
Thus I think this statement is contradictory with the previous bullet. @allenwb, @jmdyck noticed [old bug 1075](https://bugs.ecmascript.org/show_bug.cgi?id=1075) seems relevant, but presumably this is not needed anymore.


allenwb:
I think you could further simply the second paragraph to start:

If an implementation extends ~~non-strict~~ any  function objects with an own property named `"caller"`...

The first paragraph identifies which kinds of functions may _not_ be extended. The second paragraph restricts the nature of the `caller` extension in the remaining places it might occur.

bterlson:
@allenwb ooh yeah I like that better. Alright, thanks!

Issue #623, opened 2016-06-28
Consider dropping the "optimization" for functions without parameter default value initializers


jorendorff:
Currently, FunctionDeclarationInstantiation branches at step 26:

    26. If hasParameterExpressions is false, then
        a. NOTE Only a single lexical environment is needed for the parameters and top-level vars.
        ...
    27. Else,
        a. NOTE A separate Environment Record is needed to ensure that closures created by expressions in the formal parameter list do not have visibility of declarations in the function body.
        ...

There's a special case (step 26) for functions without defaults; and a general case (step 27) that looks like it would work whether you have defaults or not.

Our implementation will in fact branch like this, somewhere. But why have the special case *in the spec*? It reads like a pure optimization, which seems undesirable for a spec. I think @allenwb wrote this - Allen, do you know what this is for?

allenwb:
I considered that possibility when I updated the algorithm to include the arguments environment. The reason I didn't go that route is that I wanted to emphasize that the separate environment was there solely to support parameter lists with defaults. Plus the overall algorithm is complicated enough that it seemed wise to normatively specify the intended semantics for the "optimized" case rather than leaving it to each implementation to figure that out.

The optimized branch could be eliminated and replaced with an informative note. However, at the time, carefully crafting the language of such a note seemed harder than just including both paths in the algorithm.

jorendorff:
> Plus the overall algorithm is complicated enough that it seemed wise to normatively specify the intended semantics for the "optimized" case rather than leaving it to each implementation to figure that out.

Thanks for thinking of us implementors, and thanks for the fast response.

The NOTE in step 26.a. should *say* that this is meant to illustrate a pure optimization that implementations may do. As it stands, I was concerned that the branch might have been done, rather, because the two branches were *not* the same, that adding even a trivial default like `= 0` changed behavior in some very subtle way that we were missing.

Honestly, though, I think I'd rather this be removed. Implementations will likely choose to optimize this some other way. For example, nested environments can often be fused; this particular optimization might fall out for free, as a special case of a more general analysis that looks for such opportunities.

littledan:
I like @jorendorff 's suggestion. Our implementation has a conditional too, but we only add the scope if we see sloppy direct eval being called anywhere in the function body or defaults.

ljharb:
@jorendorff any interest in making a PR?

PR #1046, opened 2017-12-14
Normative: Eliminate extra environment for eval in parameter initializers

ajklein:
This environment was originally added to the spec as the fix for a [bug](https://bugs.ecmascript.org/show_bug.cgi?id=3383) reported by @anba 
during the development of ES6. It first appeared in rev33, after a [brief discussion](https://mail.mozilla.org/pipermail/es-discuss/2014-December/040541.html)
on es-discuss.

Yet as explained in my [presentation](https://docs.google.com/presentation/d/11xRhQlcNGBdmKC43lEx7fUP0wl-QtFLg55FwgAvLE2k/edit?usp=sharing) at the November 2017 meeting,
it has not been uniformly implemented in engines, and its correct implementation in engines
that have attempted it has been a major source of bugs.

This PR implements "Option 4" of the linked slides, which coincides with the implementation
in JavaScriptCore, and thus is believed to be web-compatible.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -8620,8 +8620,6 @@
       -->
       <emu-alg>
         1. Let _calleeContext_ be the running execution context.
-        1. Let _env_ be the LexicalEnvironment of _calleeContext_.
-        1. Let _envRec_ be _env_'s EnvironmentRecord.
         1. Let _code_ be _func_.[[ECMAScriptCode]].
         1. Let _strict_ be _func_.[[Strict]].
         1. Let _formals_ be _func_.[[FormalParameters]].
@@ -8651,6 +8649,15 @@
         1. Else if _hasParameterExpressions_ is *false*, then
           1. If *"arguments"* is an element of _functionNames_ or if *"arguments"* is an element of _lexicalNames_, then
             1. Set _argumentsObjectNeeded_ to *false*.
+        1. If _strict_ is *true* or if _hasParameterExpressions_ is *false*, then
+          1. NOTE: Only a single lexical environment is needed for the parameters and top-level vars.
+          1. Let _env_ be the LexicalEnvironment of _calleeContext_.
+          1. Let _envRec_ be _env_'s EnvironmentRecord.
+        1. Else,
+          1. NOTE: A separate Environment Record is needed to ensure that bindings created by direct eval calls in the formal parameter list are outside the environment where parameters are declared.
+          1. Let _calleeEnv_ be the LexicalEnvironment of _calleeContext_.
+          1. Let _env_ be NewDeclarativeEnvironment(_calleeEnv_).
+          1. Let _envRec_ be _env_'s EnvironmentRecord.
         1. For each String _paramName_ in _parameterNames_, do
           1. Let _alreadyDeclared_ be _envRec_.HasBinding(_paramName_).
           1. NOTE: Early errors ensure that duplicate parameter names can only occur in non-strict functions that do not have parameter default values or rest parameters.
@@ -19560,40 +19567,12 @@
       </emu-alg>
       <emu-grammar>FormalParameter : BindingElement</emu-grammar>
       <emu-alg>
-        1. If ContainsExpression of |BindingElement| is *false*, return the result of performing IteratorBindingInitialization for |BindingElement| using _iteratorRecord_ and _environment_ as the arguments.
-        1. Let _currentContext_ be the running execution context.
-        1. Let _originalEnv_ be the VariableEnvironment of _currentContext_.
-        1. Assert: The VariableEnvironment and LexicalEnvironment of _currentContext_ are the same.
-        1. Assert: If _environment_ is not *undefined*, then _environment_ and _originalEnv_ are the same.
-        1. Let _paramVarEnv_ be NewDeclarativeEnvironment(_originalEnv_).
-        1. Set the VariableEnvironment of _currentContext_ to _paramVarEnv_.
-        1. Set the LexicalEnvironment of _currentContext_ to _paramVarEnv_.
-        1. Let _result_ be IteratorBindingInitialization of |BindingElement| with arguments _iteratorRecord_ and _environment_.
-        1. Set the VariableEnvironment of _currentContext_ to _originalEnv_.
-        1. Set the LexicalEnvironment of _currentContext_ to _originalEnv_.
-        1. Return _result_.
+        1. Return the result of performing IteratorBindingInitialization for |BindingElement| with arguments _iteratorRecord_ and _environment_.
       </emu-alg>
-      <emu-note>
-        <p>The new Environment Record created in step 6 is only used if the |BindingElement| contains a direct eval.</p>
-      </emu-note>
       <emu-grammar>FunctionRestParameter : BindingRestElement</emu-grammar>
       <emu-alg>
-        1. If ContainsExpression of |BindingRestElement| is *false*, return the result of performing IteratorBindingInitialization for |BindingRestElement| using _iteratorRecord_ and _environment_ as the arguments.
-        1. Let _currentContext_ be the running execution context.
-        1. Let _originalEnv_ be the VariableEnvironment of _currentContext_.
-        1. Assert: The VariableEnvironment and LexicalEnvironment of _currentContext_ are the same.
-        1. Assert: If _environment_ is not *undefined*, then _environment_ and _originalEnv_ are the same.
-        1. Let _paramVarEnv_ be NewDeclarativeEnvironment(_originalEnv_).
-        1. Set the VariableEnvironment of _currentContext_ to _paramVarEnv_.
-        1. Set the LexicalEnvironment of _currentContext_ to _paramVarEnv_.
-        1. Let _result_ be IteratorBindingInitialization of |BindingRestElement| with arguments _iteratorRecord_ and _environment_.
-        1. Set the VariableEnvironment of _currentContext_ to _originalEnv_.
-        1. Set the LexicalEnvironment of _currentContext_ to _originalEnv_.
-        1. Return _result_.
+        1. Return the result of performing IteratorBindingInitialization for |BindingRestElement| with arguments _iteratorRecord_ and _environment_.
       </emu-alg>
-      <emu-note>
-        <p>The new Environment Record created in step 6 is only used if the |BindingRestElement| contains a direct eval.</p>
-      </emu-note>
     </emu-clause>
 
     <emu-clause id="sec-function-definitions-runtime-semantics-instantiatefunctionobject">


anba:
> This PR implements "Option 4" of the linked slides, which coincides with the implementation
> in JavaScriptCore, and thus is believed to be web-compatible.

A bit hand-wavy because I haven't checked the changes in detail, but I don't think it completely matches JSC (which shouldn't be a requirement).

```js
function f0(a = eval("var b"), b) { }
// f0(); // No error

function f1(a = eval("var b = 0"), b) { }
// f1(); // ReferenceError, but no error in JSC

function f2(a = eval("function b(){}"), b) { }
// f2(); // ReferenceError, but SyntaxError in JSC

function f3(a = eval("{ function b(){} }"), b) { }
// f3(); // ReferenceError, but no error in JSC

function f4(b, a = eval("var b = 0")) { print(b) }
// f4(); // Prints "0"

function f5(b, a = eval("function b(){}")) { print(b) }
// f5(); // Prints "function b(){}", but SyntaxError in JSC

function f6(b, a = eval("{ function b(){} }")) { print(b) }
// f6(); // Prints "function b(){}", but "undefined" in JSC
```

ajklein:
@anba Thanks for the examples!

```js
function f0(a = eval("var b"), b) { }
// f0(); // No error
```

JSC seems to get this wrong in general: it doesn't detect conflicts between var declarations and lexical variables.

```js
function f1(a = eval("var b = 0"), b) { }
// f1(); // ReferenceError, but no error in JSC
```

I think the PR actually makes this a SyntaxError during EvalDeclarationInstantiation. That's its intention, anyway; if your reading is different, that's definitely something I want to change.

```js
function f2(a = eval("function b(){}"), b) { }
// f2(); // ReferenceError, but SyntaxError in JSC
```

Same as above, I think SyntaxError is correct.

```js
function f3(a = eval("{ function b(){} }"), b) { }
// f3(); // ReferenceError, but no error in JSC
```

Why do you think this should be a ReferenceError? I would expect this to result in no error.

```js
function f4(b, a = eval("var b = 0")) { print(b) }
// f4(); // Prints "0"
```

Yeah, this is the same JSC bug as f0; I think this should be a SyntaxError.

```js
function f5(b, a = eval("function b(){}")) { print(b) }
// f5(); // Prints "function b(){}", but SyntaxError in JSC
```

SyntaxError is what I expect here too.

```js
function f6(b, a = eval("{ function b(){} }")) { print(b) }
// f6(); // Prints "function b(){}", but "undefined" in JSC
```

"undefined" seems right to me, too. Are you expecting Annex B to cause the function to get assigned to `b`? I was thinking it wouldn't, since there's a conflicting binding in the outer scope.

sawaisom:
Yhdji

anba:
> ```js
> function f1(a = eval("var b = 0"), b) { }
> // f1(); // ReferenceError, but no error in JSC
> ```
> 
> I think the PR actually makes this a SyntaxError during EvalDeclarationInstantiation. That's its intention, anyway; if your reading is different, that's definitely something I want to change.

The parameter bindings are installed in the current execution context's VariableEnvironment, so the redeclaration check in EvalDeclarationInstantiation, step 5.d won't report any SyntaxErrors for the variable declaration `b`. Later in EvalDeclarationInstantiation (step 16.b) a new binding for `b` is not created, because the VariableEnvironment's environment record already contains a binding for `b`. So when we then even later evaluate `var b = 0`, we'll call PutValue (13.3.2.4, step 6), PutValue then calls SetMutableBinding, and 8.1.1.1.5 SetMutableBinding, step 4 then throws a ReferenceError because the binding for `b` isn't yet initialized.

And because EvalDeclarationInstantiation won't throw any SyntaxErrors, Annex B semantics apply for the other cases with `function b() {}`.

ajklein:
@anba Thanks, got it. The quickest fix I can think of would be, in sloppy mode, to introduce a LexicalEnvironment that sits just below the VariableEnvironment of the function, and declare parameters there. Do you think this would cause any additional problems?

anba:
Do you mean the following environment set-up? (I haven't yet checked if this approach is feasible!)

<p>Non-strict function with parameter expressions:<details><pre>
           Description                    Environment kind               Execution context state
+----------------------------------+     +----------------+     +-----------------------------------------+
|                                  |     |                |     |                                         |
| Var-bindings from eval in params +-----+  FunctionEnv   +-----+ VariableEnv during parameter evaluation |
|                                  |     |                |     |                                         |
+----------------------------------+     +----------------+     +-----------------------------------------+
                                                |
                                                |
        +--------------------------+     +------v---------+     +----------------------------------------+
        |                          |     |                |     |                                        |
        |  Parameter bindings      +-----+ DeclarativeEnv +-----+ LexicalEnv during parameter evaluation |
        |  "arguments" binding (?) |     |                |     |                                        |
        |                          |     +----------------+     +----------------------------------------+
        +--------------------------+            |
                                                |
   +--------------------------------+    +------v---------+     +----------------------------------------+
   |                                |    |                |     |                                        |
   | Var-bindings from body         +----+ DeclarativeEnv +-----+ VariableEnv after parameter evaluation |
   | Var-bindings from eval in body |    |                |     |                                        |
   |                                |    +----------------+     +----------------------------------------+
   +--------------------------------+           |
                                                |
   +--------------------------------+    +------v---------+     +---------------------------------------+
   |                                |    |                |     |                                       |
   | Lexical bindings from body     +----+ DeclarativeEnv +-----+ LexicalEnv after parameter evaluation |
   |                                |    |                |     |                                       |
   +--------------------------------+    +----------------+     +---------------------------------------+
</pre></details></p>

<p>Strict function with parameter expressions (no changes):<details><pre>
           Description              Environment kind               Execution context state
       +---------------------+     +----------------+     +-----------------------------------------+
       |                     |     |                |     |                                         |
       | Parameter bindings  +-----+  FunctionEnv   +-----+ VariableEnv during parameter evaluation |
       | "arguments" binding |     |                |     | LexicalEnv during parameter evaluation  |
       |                     |     +------+---------+     |                                         |
       +---------------------+            |               +-----------------------------------------+
                                          |
+----------------------------+     +------v---------+     +----------------------------------------+
|                            |     |                |     |                                        |
| Var-bindings from body     +-----+ DeclarativeEnv +-----+ VariableEnv after parameter evaluation |
| Lexical bindings from body |     |                |     | LexicalEnv after parameter evaluation  |
|                            |     +----------------+     |                                        |
+----------------------------+                            +----------------------------------------+
</pre></details></p>

<p>Non-strict function without parameter expressions (no changes):<details><pre>
           Description                    Environment kind               Execution context state
+----------------------------------+     +----------------+     +----------------------------------------+
|                                  |     |                |     |                                        |
|  Parameter bindings              +-----+  FunctionEnv   +-----+ VariableEnv                            |
|  "arguments" binding             |     |                |     | LexicalEnv during parameter evaluation |
|  Var-bindings from body          |     +------+---------+     |                                        |
|  Var-bindings from eval in body  |            |               +----------------------------------------+
|                                  |            |
+----------------------------------+            |
                                                |
  +--------------------------------+     +------v---------+     +---------------------------------------+
  |                                |     |                |     |                                       |
  | Lexical bindings from body     +-----+ DeclarativeEnv +-----+ LexicalEnv after parameter evaluation |
  |                                |     |                |     |                                       |
  +--------------------------------+     +----------------+     +---------------------------------------+
</pre></details></p>

<p>Strict function without parameter expressions (no changes):<details><pre>
           Description                    Environment kind       Execution context state
+----------------------------------+     +----------------+     +-----------------------+
|                                  |     |                |     |                       |
|  Parameter bindings              +-----+  FunctionEnv   +-----+ VariableEnv           |
|  "arguments" binding             |     |                |     | LexicalEnv            |
|  Var-bindings from body          |     +----------------+     |                       |
|  Lexical bindings from body      |                            +-----------------------+
|                                  |
+----------------------------------+
</pre></details></p>

With:
- FunctionEnv = Lexical Environment with a function Environment Record
- DeclarativeEnv = Lexical Environment with a declarative Environment Record

ajklein:
Yes, that's what I had in mind.

ajklein:
@bterlson have you had a chance to look at this? (By "this" I guess I mean both the PR and @anba's lovely ASCII art above).

syg:
@ajklein So based on @anba's diagrams, that means `eval` in parameter expression positions won't actually have access to any of the parameters, since the eval has the environment that *encloses* the parameter environment?

ajklein:
@syg That's not my understanding. The eval should have _access_ to the parameters; it's simply that any var bindings created in the eval will be "hoisted" above the parameters (or cause an error in case they conflict with the name of a parameter).

syg:
Ah, right, the direct eval will have its own LexicalEnv which has the parameter environment as the enclosing environment, while its VariablesEnv will be the function environment. That seems okay.

ajklein:
I've update the PR to match @anba's diagram. Another round of review would be greatly appreciated. I'm also not at all happy with my NOTEs, so any suggestions for how to add more clarity would be good.

rwaldron:
@leobalter @spectranaut @nomadtechie we should review these changes for potential Test262 updating.

ajklein:
Thanks for the review, @bakkot. I'd rather leave the strict branch as-is in this PR to leave the focus on the functional change, and leave it for #623 to replace this text with an informative note. If a fix for #623 gets merged before this, then of course I'd reflect that change here too.

woess:
IMHO, Option 4 (this PR) is a step back in that it reintroduces the undesirable and error-prone behavior that eval var declarations modify the caller's VariableEnvironment.

If implementation complexity was the rationale for this proposal, why was Option 3 "Make eval in parameters strict" dismissed? In comparison, that option seems the best: it's semantically sane and relatively easy to implement.

Admittedly, implicitly switching to strict mode might come as a bit of a surprise, but there are other precedents in the spec where new language features change legacy behavior.
And if that's a (compatibility) concern, there'd also be the option to leave the eval code in non-strict mode but with the strict mode env setup, but I don't think there's a necessity.

ajklein:
@woess It was indeed the concerns you lay out in your second paragraph I was worried about. We'd either have to decide that it was OK to change the strict-ness of the existing eval code, or create a new thing (strict eval with sloppy code inside).

I agree that the eval var declaration leakage is not great, but the amount of spec and implementation complexity (and security issues in the wild, in Chrome/V8's case) to "defend" against what I think is a very rare case doesn't seem worthwhile to me.

ajklein:
@bterlson @ljharb @bmeck Would like to pick this up again, pinging editors for some formal review (I believe @syg looked at this too).

bmeck:
@rwaldron did we ever figure out if there were any Test262 concerns?

ajklein:
Re-pinging this thread for a new year. I continue to believe this is a nice simplification, which if landed and implemented in engines would increase interop while reducing both cognitive and implementation complexity.

ljharb:
@ajklein would you be able to resummarize both a) what current browsers do, and b) what would change as a result of this patch? Thanks :-)

bakkot:
@ljharb / @ajklein: to jostle my own memory, I wrote up [my understanding / recollection](https://gist.github.com/bakkot/fb91c5182dc3de0e8833c5c37e533777) of the current and proposed behavior with some tests and their results on current engines.

Basically: the spec currently says that each parameter is evaluated in its own unique var scope which is just inside of the parameter scope. After this change, there would be a var scope just outside of the parameter scope in which all parameters would be evaluated. V8 and SpiderMonkey correctly implement the current spec, Chakra and JSC do their own unique things.

ljharb:
@ajklein would you mind rebasing this? after that, if @bakkot and @syg approve, I'll be happy to merge it.

erights:
What would be the observable effects for strict-only code? Examples?

bakkot:
@erights There would be no observable effects for strict-only code, because strict `eval`s already get their own VariableEnvironment which prevents hoisting declarations out of the `eval`. This change only affects code which have non-strict direct `eval`s in parameter expressions.

ajklein:
This isn't something I have bandwidth to push on right now, but I still think it's a good idea so would be happy for someone else to take it over.

syg:
I will be taking over this.

leobalter:
Tests have been merged

PR #1063, opened 2018-01-11
Normative: Add AsyncIteration

zbraniecki:
This is a PR to include the Stage 4 proposal for async iterators - https://github.com/tc39/proposal-async-iteration

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -968,6 +968,17 @@
           <table>
             <tbody>
             <tr>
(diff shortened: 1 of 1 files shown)


zbraniecki:
@domenic I included all the patches except of `Miscellaneous Patches`. I'm not familiar with the subject enough to change the algorithms around for that.

PR #1065, opened 2018-01-12
Editorial: fix grammar of synchronizes-with conditions

conrad-watt:
Read painfully literally, the conditions of synchronizes-with are mutually incompatible, or at least ungrammatical.

https://tc39.github.io/ecma262/#sec-synchronizes-with

The phrasing of these conditions seems to come, at least partially, from an earlier draft, where synchronizes-with was computed in a more algorithmic way.
https://tc39.github.io/ecmascript_sharedmem/shmem.html#MemoryModel.SynchronizesWith

The switch to the non-algorithmic definition with the phrasing "if all the following are true"  in the preamble means that the conditions must be rephrased.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38273,7 +38273,7 @@ THH:mm:ss.sss
           <ul>
             <li>_R_.[[Order]] is `"SeqCst"`.</li>
             <li>_W_.[[Order]] is `"SeqCst"` or `"Init"`.</li>
-            <li>If _W_.[[Order]] is `"SeqCst"` then _R_ and _W_ have equal ranges.</li>
+            <li>If _W_.[[Order]] is `"SeqCst"`, then _R_ and _W_ have equal ranges.</li>
             <li>If _W_.[[Order]] is `"Init"`, then for each event _V_ such that (_R_, _V_) is in _execution_.[[ReadsFrom]], _V_.[[Order]] is `"Init"`.</li>
           </ul>
         </li>
--- a/spec.html
+++ b/spec.html
@@ -38272,6 +38272,7 @@ THH:mm:ss.sss
           <p>For each pair (_R_, _W_) in _execution_.[[ReadsFrom]], (_W_, _R_) is in _execution_.[[SynchronizesWith]] if all the following are true.</p>
           <ul>
             <li>_R_.[[Order]] is `"SeqCst"`.</li>
+            <li>_W_.[[Order]] is `"SeqCst"` or `"Init"`.</li>
             <li>If _W_.[[Order]] is `"SeqCst"` then _R_ and _W_ have equal ranges.</li>
             <li>If _W_.[[Order]] is `"Init"`, then for each event _V_ such that (_R_, _V_) is in _execution_.[[ReadsFrom]], _V_.[[Order]] is `"Init"`.</li>
           </ul>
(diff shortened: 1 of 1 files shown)


conrad-watt:
I can maybe make this more clear with an example.

Given two events R and W, both SeqCst, with equal ranges and R reads-from W, it is the intention of the specification that W synchronizes-with R.

Consider each of the statements in turn (" if all the following are true"):

R.[[Order]] is "SeqCst"
This is true - trivially from our setup.

If W.[[Order]] is "SeqCst" and R and W have equal ranges.
This statement is true (although I'd argue that the presence of the word "if" is suspicious).

If W.[[Order]] is "Init", and for each event V such that (R, V) is in execution.[[ReadsFrom]], V.[[Order]] is "Init".
This statement is false, since W.[[Order]] is not "Init". Since one of the three statements is false, according to the specification as it currently stands it is not the case that W synchronizes-with R.

To emphasise, consider the different scenarios under which these two sentences are true (the question "is the following true")
A and B
If A then B

And consider the scenarios under which this sentence is true.
If A and B

ljharb:
Gotcha, that makes sense. In that case, I'd say a better way to clear this up would be to replace this list and the preceding statement with an abstract operation that can `return` true or false as needed, without the confusing prose.

syg:
"then" is fine, but I do agree with @ljharb that it reads awkwardly.

@conrad-watt How do you feel about the following?

- R.[[Order]] is "SeqCst".
- W.[[Order]] is "SeqCst" or "Init".
- When W.[[Order]] is "SeqCst", R and W have equal ranges.
- When W.[[Order]] is "Init", ...

Also @ljharb, I prefer to not change the relational stuff to abstract operations. The relational stuff was consciously made to not use step-by-step language to avoid having the reader mistakenly believe they could be reasoned about operationally.

ljharb:
@syg totally fine to defer to your preference here, and i like your rewording.

That said, abstract operations are just reusable bags of algorithm instructions, that imo always greatly increase readability.

syg:
> That said, abstract operations are just reusable bags of algorithm instructions, that imo always greatly increase readability.

Exactly right for algorithms, but this isn't an algorithm. I remember we went back and forth on the language and decided to use regular prose, so I'd like to not make an exception for this.

conrad-watt:
@syg you've made me realise that without your second bullet "W.[[Order]] is "SeqCst" or "Init" my proposal is incorrect, since it forces R to synchronize with "Unordered" writes that it reads from.

I chose "if... then" because I felt it most closely matches the existing specification language for the memory model (most notably in 27.7.4 bullet 3 https://tc39.github.io/ecma262/#sec-memory-order and 27.8 https://tc39.github.io/ecma262/#sec-races). Perhaps with the second bullet included, this reads more naturally, please take a look. I'm happy to change to "when" if it's preferred though.

For what it's worth, personally I prefer the axiomatic/prose style, since I'm approaching this from the point of view of formal specification/proof. I know this approach to the spec will be in the minority though :)

spec.html
```diff
@@ -38272,8 +38272,9 @@ <h1>synchronizes-with</h1>
           <p>For each pair (_R_, _W_) in _execution_.[[ReadsFrom]], (_W_, _R_) is in _execution_.[[SynchronizesWith]] if all the following are true.</p>
           <ul>
             <li>_R_.[[Order]] is `"SeqCst"`.</li>
-            <li>If _W_.[[Order]] is `"SeqCst"` and _R_ and _W_ have equal ranges.</li>
-            <li>If _W_.[[Order]] is `"Init"`, and for each event _V_ such that (_R_, _V_) is in _execution_.[[ReadsFrom]], _V_.[[Order]] is `"Init"`.</li>
+            <li>_W_.[[Order]] is `"SeqCst"` or `"Init"`.</li>
+            <li>If _W_.[[Order]] is `"SeqCst"` then _R_ and _W_ have equal ranges.</li>
```
syg: Existing nit: please add a , before the 'then' for consistency.

conrad-watt: done

PR #1074, opened 2018-01-23
Meta: Link to the new contributor form

littledan:
See https://github.com/tc39/Reflector/issues/104 for context.


getify:
Just FYI: that "context" link is apparently not publicly accessible. The rest of us can't tell what this is about.

ljharb:
That’s by design; the context is available to the intended reviewers, which are TC39 delegates.

littledan:
@getify To summarize that thread, the changes here are about:
- Moving to Google Forms, which should enable us do automated checking that contributors have signed the form (among other technical improvements)
- Making the text a little more explicit about the IP rights that are being licensed to Ecma, and cross-referencing more with the various relevant Ecma agreements

PR #1083, opened 2018-01-26
Normative: add explicit "length" property info for ThenFinally/CatchFinally

### LINKED ISSUE #1078: ThenFinally and CatchFinally functions should define explicit length
Anonymous built-in functions without a parameter list in the subclause headings which are reachable by user code should define an explicit value for their `length` property.

ljharb:
Fixes #1078

cc @anba

Technically this might be a normative bugfix?

(I also fixed the whitespace on the CatchFinally emu-clause)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37657,17 +37657,17 @@ THH:mm:ss.sss
               1. Let _thenFinally_ be _onFinally_.
               1. Let _catchFinally_ be _onFinally_.
           1. Else,
-              1. Let _thenFinally_ be a new built-in function object as defined in <emu-xref href="#sec-thenfinallyfunction">ThenFinally Function</emu-xref>.
-              1. Let _catchFinally_ be a new built-in function object as defined in <emu-xref href="#sec-catchfinallyfunction">CatchFinally Function</emu-xref>.
+              1. Let _thenFinally_ be a new built-in function object as defined in <emu-xref href="#sec-thenfinallyfunctions" title></emu-xref>.
+              1. Let _catchFinally_ be a new built-in function object as defined in <emu-xref href="#sec-catchfinallyfunctions" title></emu-xref>.
               1. Set _thenFinally_ and _catchFinally_'s [[Constructor]] internal slots to _C_.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -37678,24 +37678,26 @@ <h1>ThenFinally Function</h1>
...
-            1. Assert: IsCallable(_onFinally_) is *true*.
-            1. Let _result_ be ? Call(_onFinally_, *undefined*).
-            1. Let _C_ be _F_.[[Constructor]].
-            1. Assert: IsConstructor(_C_) is *true*.
-            1. Let _promise_ be ? PromiseResolve(_C_, _result_).
-            1. Let _thrower_ be equivalent to a function that throws _reason_.
-            1. Return ? Invoke(_promise_, `"then"`, &laquo; _thrower_ &raquo;).
-        </emu-alg>
+        <emu-clause id="sec-catchfinallyfunction" aoid="CatchFinallyFunction">
+          <h1>CatchFinally Function</h1>
```
littledan: Unrelated nit: Seems like similar closures in the spec have headings in plural and with words separated, e.g., [Promise Resolve Functions](http://www.ecma-international.org/ecma-262/#sec-promise-resolve-functions).

ljharb:
@littledan thanks, updated

gsathya:
don't normative changes require a test262 test?

ljharb:
@gsathya technically i'd consider this a normative bugfix, and I'd hope every implementation inferred this change on their own; however, I'll put up a test262 PR right now.

ljharb:
@gsathya https://github.com/tc39/test262/pull/1401

littledan:
@ljharb Should normative changes not require tests if they are bug fixes? I don't understand the distinction.

ljharb:
They all should indeed have tests; I’m not sure how rigorous we needed to be here about the ordering since we’re talking about such a brief period between `finally` landing, this bug fix landing, and the tests for this bugfix landing. In this case i think the previous spec implied the length correctly, the change in this PR makes it explicit.

ljharb:
For sure, that’s why i rushed to get a tests PR out as soon as it was pointed out :-)

PR #1086, opened 2018-01-29
Editorial: use "is 0" instead of "is zero" for consistency

ljharb:
Discovered [here](https://github.com/keithamus/proposal-array-last/pull/9#discussion_r164578959).

The spec seems to use `is 0` most often for comparing to zero, specifically with the results of a `ToLength` call, which can not return anything `<= 0` besides positive zero.

An equally valid and consistent alternative would be replacing all the `is 0` instances with `is zero`, if we'd find that clearer.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4197,7 +4197,7 @@
               String
             </td>
             <td>
-              If _argument_ is the empty String (its length is zero), return *false*; otherwise return *true*.
+              If _argument_ is the empty String (its length is 0), return *false*; otherwise return *true*.
             </td>
           </tr>
           <tr>
@@ -11235,7 +11235,7 @@
     <emu-clause id="sec-literals-string-literals">
       <h1>String Literals</h1>
       <emu-note>
-        <p>A string literal is zero or more Unicode code points enclosed in single or double quotes. Unicode code points may also be represented by an escape sequence. All code points may appear literally in a string literal except for the closing quote code points, U+005C (REVERSE SOLIDUS), U+000D (CARRIAGE RETURN), and U+000A (LINE FEED). Any code points may appear in the form of an escape sequence. String literals evaluate to ECMAScript String values. When generating these String values Unicode code points are UTF-16 encoded as defined in <emu-xref href="#sec-utf16encodecodepoint"></emu-xref>. Code points belonging to the Basic Multilingual Plane are encoded as a single code unit element of the string. All other code points are encoded as two code unit elements of the string.</p>
+        <p>A string literal is 0 or more Unicode code points enclosed in single or double quotes. Unicode code points may also be represented by an escape sequence. All code points may appear literally in a string literal except for the closing quote code points, U+005C (REVERSE SOLIDUS), U+000D (CARRIAGE RETURN), and U+000A (LINE FEED). Any code points may appear in the form of an escape sequence. String literals evaluate to ECMAScript String values. When generating these String values Unicode code points are UTF-16 encoded as defined in <emu-xref href="#sec-utf16encodecodepoint"></emu-xref>. Code points belonging to the Basic Multilingual Plane are encoded as a single code unit element of the string. All other code points are encoded as two code unit elements of the string.</p>
       </emu-note>
       <h2>Syntax</h2>
       <emu-grammar type="definition">


annevk:
I suggest 0, but please record the resolution in https://github.com/tc39/proposal-bigint/issues/10.

jmdyck:
Note that if `_x_` is a Number value, writing `_x_ is 0` or `_x_ is zero` is somewhat ill-defined, because Number doesn't have a (single) zero value, it has `*+0*` and `*-0*`. I think it generally means `_x_ is *+0* or *-0*`, but it might be better to say so explicitly.

Of course, if `_x_` were a "mathematical value", then comparison with `0` would be well-defined. However:
(a) distinguishing the two possibilities is currently non-trivial, and
(b) it's still an open question whether the second possibility exists (tc39/proposal-bigint#10).

leobalter:
@jmdyck I think `0` should be fine:

```
5.2.5 Mathematical Operations

...
If a mathematical operation or function is applied to a floating-point number, it should be understood as being applied to the exact mathematical value represented by that floating-point number; such a floating-point number must be finite, and if it is +0 or -0 then the corresponding mathematical value is simply 0.

...

6.1.6 The Number Type

...
Note that all the positive and negative integers whose magnitude is no greater than 253 are representable in the Number type (indeed, the integer 0 has two representations, +0 and -0).

```

If the specs requires strict comparisons to positive or negative zero, this is and should be always explicit. The `0` refers to any, and if needs to produce or cast any value, +0 should be used to avoid ambiguity.

jmdyck:
@leobalter: Relying on the Mathematical Operations clause to resolve this question requires you to say that "is" is a mathematical operation, which is a stretch just on its own, but also inconsistent with the spec's many non-mathematical uses of "is".

ljharb:
I’m content with anything that’s unambiguous and consistent. Would “is zero” remove the potential interpretation that it’s only comparing to +0?

bterlson:
Am I alone in thinking we should avoid depending on "mathematical values" if at all possible (in part because the long-standing question about whether/how they exist) and just go with explicit +/- 0? Or, define a term like "the value zero" which is just +0 and -0?

ljharb:
That sounds fine too - I could also replace all these cases with `SameValue(blah, 0)`, or `is positive or negative zero`.

jridgewell:
> just go with explicit +/- 0

"is ±0", "is not ±0"? 👍

bterlson:
I like PLUS-MINUS SIGN, yeah, good call.

ljharb:
Great, I'll update all of these to match that.

ljharb:
hmm, that doesn't really work for the conceptual places like when they're talking about string length. Would using `SameValueZero` maybe be more explicit?

domenic:
@bterlson you are not alone, but the general problem is pretty endemic in the spec right now (see https://github.com/tc39/proposal-bigint/issues/10 as mentioned by @jmdyck upthread). But yeah, if you think it's reasonable to make a piecemeal fix for these cases instead of waiting on the big-ol'-refactor that I'm sure @littledan will be contributing any day now, that seems like a good move.

bterlson:
@ljharb for string length, I think +0 is the only possibility.

@domenic yep, agreed it's a problem; being explicit should make any subsequent fixes easier too :)

ljharb:
@bterlson cool, i'll use `is +0` or `is ±0` based on where the variance is possible. Update coming shortly.

allenwb:
Isn't this all a matter of context?  If you compare an ECMAScript value to "zero" you must be dealing with ECMAScript number values not mathematical values.  If you are comparing  a value that was computed using a mathematical formula then you are comparing to a mathematical 0.  The context is usually pretty clear.  

If you are testing a mathematical value to see if it is mathematically equal to 0  you should use the mathematical = operator and the numeral 0.

If you are testing an ECMAScript Number value _x_ to see if it is a number value you should say: "if x is 0" meaning _x_ is either the number value +0 or the number value -0; "if _x_ is +0" meaning _x_ is the number value +0 but not the number value -0; or,  "if _x_ is -0" meaning _x_ is the number value -0 but not the number value +0.  I think these meanings should be pretty obvious but if you want to explicitly explain them somewhere the 5th paragraph of [Number Type](https://tc39.github.io/ecma262/#sec-ecmascript-language-types-number-type) seem like a fine place.

Never use "is zero" in an algorithm or formula.  It may be appropriate in some prose paragraphs that aren't explicitly talking about either number values or mathematical values.  I checked the current draft.  There appears to be 7 occurrences of "is zero" in an algorithm (or algorithm-like table).  They all appear to be very old text that was never updated.  They should all be fixed.

 I think that using ±0 in place of 0 would just be adding noise to the specification.

allenwb:
From ECMA/TC39/97/24, Notes from the 2/14/1997 ECMAScript  1st Edition Drafting Working Group Meeting:
> In several places the document uses zero with saying whether positive zero, negative zero or both are intended. It would be useful to do a scan through the doc looking specifically for these cases.

[97-024.PDF](https://github.com/tc39/ecma262/files/1683747/97-024.PDF)

ljharb:
That suggests that a `0`, without a `+`, `-`, or something akin to `±`, is something that should be changed to be unambiguous.

allenwb:
@ljharb Actually, don't think a comment about a 1997 draft suggests anything about the current specification.  The specification and its conventions  have changed drastically in the last 21 years...

However, I did think it highly assuming, in the context of this thread, that I ran into it this afternoon.  Had to share the fun.

jmdyck:
> If you compare an ECMAScript value to "zero" you must be dealing with ECMAScript number values not mathematical values. If you are comparing a value that was computed using a mathematical formula then you are comparing to a mathematical 0. The context is usually pretty clear.

There are a lot of contexts where it isn't clear, i.e. where a numeric quantity is neither the result of a mathematical formula nor obviously an ES language value. E.g., consider `Let _x_ be` followed by any of:
```
- 0
- the smallest nonnegative integer such that ...
- the code unit ...
- the 8-bit value represented by the two hexadecimal digits ...
- the integer represented by the four hexadecimal digits ...
- the Element Size value in Table 56 for ...
- the number of arguments passed to this function call
- the code point value of ...
- the length of _y_
- the number of bytes in _y_
- the number of characters contained in _y_
- the number of leading zero bits in the 32-bit binary representation of _y_
- the number of left-capturing parentheses in _y_
- (in general:) the number of [things] in [collection]
```
With any of these, it's unclear whether `_x_` is a Number value or a mathematical value.

Moreover, even when a quantity *is* defined with a mathematical formula, and so should be a mathematical value, it's not so clear. Consider the example given by @littledan in https://github.com/tc39/proposal-bigint/issues/10: [Array.prototype.indexOf](https://tc39.github.io/ecma262/#sec-array.prototype.indexof) has the step `Let _k_ be _len_ + _n_.` which implies that `_k_` is a mathematical value, and yet it's later passed to `ToString`, which only operates on ES language values. So maybe `_k_` is a mathematical value that's implicitly coerced to a Number value as necessary, or maybe it's actually a Number all along. It's unclear.

allenwb:
@jmdyck 
But is there any possibility of confusion about the meaning, in context>


>consider Let _x_ be followed by any of:
>- 0

doesn't explicitly say be "Let _x_ be -0" so clearly this is either an implicit +0 number value or a mathematical unsigned 0 value.  If it isn't clear from context needs to be clarified.  I haven't look at all such uses but in my sampling I haven't seen any needing to clarification.


>- the smallest nonnegative integer such that ...

https://tc39.github.io/ecma262/#sec-abstract-relational-comparison
https://tc39.github.io/ecma262/#sec-decode
Clear in context that these both deal with the bit encoding of Unicode code units and that -0 is not in the domain of discourse

Even without additional context, it is clear that in the following fragments the values could never be -0
>- the code unit ...
>- the 8-bit value represented by the two hexadecimal digits ...
>- the integer represented by the four hexadecimal digits ...
>- the Element Size value in Table 56 for ...
>- the number of arguments passed to this function call
>- the code point value of ...
>- the length of _y_
>- the number of bytes in _y_
>- the number of characters contained in _y_
>- the number of leading zero bits in the 32-bit binary representation of _y_
>- the number of left-capturing parentheses in _y_
>- (in general:) the number of [things] in [collection]

Turning all occurrences of "0" into either "±0", "+0", or "-0" going to force consideration of the possibility of signed 0 into many places within the specification where the concept of signed 0 is not at all relevant.  Seems to me that it would only make things less clear then they already are.

domenic:
It's interesting that you find all of those clear from context, Allen. I don't find it clear at all.

allenwb:
Which ones are unclear to you?  Don't see how counting integer values or bit values could possibly involve a IEEE 768 -0 value (or an Infinity or a NaN)

anba:
> Consider the example given by @littledan in tc39/proposal-bigint#10: Array.prototype.indexOf has the step Let _k_ be _len_ + _n_. which implies that _k_ is a mathematical value, and yet it's later passed to ToString, which only operates on ES language values. So maybe _k_ is a mathematical value that's implicitly coerced to a Number value as necessary, or maybe it's actually a Number all along. It's unclear.

I've always treated mathematical values and Number values as interchangeable, as long as the specific value is well-defined for both types. And this seems to be the case for the`Array.prototype.indexOf` example, because ToString is never called with a value which cannot be exactly represented by a Number value (because of the loop condition `k < len`).

Well, if we want to be super correct, more problematic is `len + n`, because `n` can be -Infinity, and mathematical values don't include infinities.

jmdyck:
> I haven't look at all such uses but in my sampling I haven't seen any needing to clarification.

I count 80 occurrences of `Let _foo_ be [numeral].` There are lots where I can't tell from context whether it's a Number value or a mathematical value, mainly because the spec doesn't say which contexts require/allow which, or if/where implicit coercions happen. (E.g., if the algorithm uses `_foo_` to index a List, does that tell you what `_foo_` is? It's unclear.)

> Even without additional context, it is clear that in the following fragments the values could never be -0

Knowing that a numeric value can't be -0 doesn't tell you whether it's a Number value or a mathematical value.

bterlson:
I just want to chime in to say that the confusion between values of the number type and mathematical values is far from clear regardless of context, and this is for experts. It's a problem that needs to be fixed even if the current regime can be justified by thinking hard about each usage. More importantly, I've yet to hear anyone say to me the distinction is useful to them in implementing the specification, and it has caused significant confusion, so being more clear everywhere just seems like a net win.

> I've always treated mathematical values and Number values as interchangeable, as long as the specific value is well-defined for both types. 

Yeah same, but isn't that just because that's the only way the current document makes sense? We could codify this identity, I suppose...

brabalan:
One of the side effect of the work we're doing on JSExplain is that we're producing a typed version of the spec, in the sense that every conversion is explicit. Once we have more of it implemented, we should be able to point to all the places where integer to Number conversions are needed.

littledan:
I share @bterlson 's worry about the ambiguity here, though I thought, as @allenwb says, that most uses of `0` @allenwb stood in for +0. I also thought `is` typically referred to `SameValue` if it's comparing JS values--would it be worth clarifying this in the notational conventions section? It's a little hard for me to understand how we could consistently avoid `is` when conditionals are phrased in terms of `is **true**`, though maybe we could limit `is` to just comparing to **true** and **false** if that would be more clear.

If we limit the number of mathematical values in the specification and switch to typically using JavaScript values, as in https://github.com/tc39/proposal-bigint/issues/10, I was imagining matching JavaScript literal syntax in the specification, making `0` mean `+0.0`. To be especially explicit, we could use `.0` on the end of all integral-valued Numbers in the specification, and put a `+` or `-` on the beginning of `0` literals, so we might say `is +0.0` for these cases. What do you think of 

Apologies for my delay in producing the number refactoring of the spec in https://github.com/tc39/proposal-bigint/issues/10 . I plan to put something out for review before the next meeting. As @jmdyck has pointed out, we can't get rid of every single mathematical value, but I think we can make them few enough that each one and each calculation involving one is called out explicitly. Since it's such a big change, we could review it in a short agenda item at the next meeting.

jmdyck:
> I thought, as @allenwb says, that most uses of 0 stood in for +0.

I don't think he said that. Rather, he said it *either* means the +0 Number value or a mathematical 0 value, and you should be able to (and usually can) tell which it is from context.

allenwb:
> I don't think he said that. Rather, he said it either means the +0 Number value or a mathematical 0 value, and you should be able to (and usually can) tell which it is from context.

Exactly. 

If we eliminate mathematical values and say all arithmetic in algorithms is done as if using ECMAScript Number values and ECMAScript operators then we are making a normative requirement that all such calculations are done using the rules of IEEE 768 double precision binary floating point.    Why would we want to do this for things like counting the elements of a list, doing offset calculations into an Array buffer, manipulating the encode of a Unicode code point, or any other place were we know we are dealing with simple integer values.

Is there any evidence at all that any implementor or casual reader has been at all confused by the existing language? Does anybody know of any implementation bugs that are traceable to such confusion?  Is forcing all computations to be "as if" double precision IEEE FP going to cause more or less confusion for such reader? 

As spec. writer we need to be hyper-conscience about being precise in our specifications.  But a trap we can  fall into is being more precise then is needed for our readers to easily understand our intent.  This can case more confusion or uncertainty than rather than less.

domenic:
We had several bugs in Web IDL and Streams implementations due to this; see the cited thread.

allenwb:
> We had several bugs in Web IDL and Streams implementations due to this; see the cited thread.

Which cited thread?  Can you repost the specific link you have in mind?

littledan:
@allenwb See also https://github.com/tc39/proposal-intl-relative-time/issues/58

ljharb:
Now that BigInt has merged, I've rebased this. Rereview would be appreciated.

spec.html
```diff
@@ -1918,8 +1918,8 @@ <h1>BigInt::sameValueZero ( _x_, _y_ )</h1>
         <emu-clause id="sec-binaryand" aoid="BinaryAnd">
           <h1>BinaryAnd ( _x_, _y_ )</h1>
           <emu-alg>
-            1. Assert: _x_ is 0 or 1.
-            1. Assert: _y_ is 0 or 1.
+            1. Assert: _x_ is +0 or 1.
+            1. Assert: _y_ is +0 or 1.
```
jmdyck: The arguments to `BinaryAnd`, `BinaryOr`, and `BinaryXor` are probably BigInt values, not Numbers (see issue #1723), so the tests should probably be `is *0n* or *1n*`.

spec.html
```diff
@@ -33488,7 +33488,7 @@ <h1>Array.prototype.pop ( )</h1>
         <emu-alg>
           1. Let _O_ be ? ToObject(*this* value).
           1. Let _len_ be ? LengthOfArrayLike(_O_).
-          1. If _len_ is zero, then
+          1. If _len_ is ±0, then
```
jmdyck: `LengthOfArrayLike` cannot return `-0`, so the test can be `is +0` (or `is 0` if you like).
(And similarly for 2 subsequent edits.)

spec.html
```diff
@@ -31113,10 +31113,10 @@ <h1>Term</h1>
           <h1>Runtime Semantics: RepeatMatcher ( _m_, _min_, _max_, _greedy_, _x_, _c_, _parenIndex_, _parenCount_ )</h1>
           <p>The abstract operation RepeatMatcher takes eight parameters, a Matcher _m_, an integer _min_, an integer (or &infin;) _max_, a Boolean _greedy_, a State _x_, a Continuation _c_, an integer _parenIndex_, and an integer _parenCount_, and performs the following steps:</p>
           <emu-alg>
-            1. If _max_ is zero, return _c_(_x_).
+            1. If _max_ is ±0, return _c_(_x_).
             1. Let _d_ be an internal Continuation closure that takes one State argument _y_ and performs the following steps:
-              1. If _min_ is zero and _y_'s _endIndex_ is equal to _x_'s _endIndex_, return ~failure~.
-              1. If _min_ is zero, let _min2_ be zero; otherwise let _min2_ be _min_ - 1.
+              1. If _min_ is ±0 and _y_'s _endIndex_ is equal to _x_'s _endIndex_, return ~failure~.
+              1. If _min_ is ±0, let _min2_ be zero; otherwise let _min2_ be _min_ - 1.
```
jmdyck: `_max_` and `_min_` come from a Quantifier, so can't be -0, so each of these 3 tests can be `is +0` (or `is 0` if you like).

spec.html
```diff
@@ -31173,7 +31173,7 @@ <h1>Assertion</h1>
         <emu-alg>
           1. Return an internal AssertionTester closure that takes a State argument _x_ and performs the following steps:
             1. Let _e_ be _x_'s _endIndex_.
-            1. If _e_ is zero, return *true*.
+            1. If _e_ is 0, return *true*.
```
annevk: Why is this a plain 0?

jmdyck: Since the merge of PR #1135, [5.2.5 Mathematical Operations](https://tc39.es/ecma262/#sec-mathematical-operations) says
> The Number value 0, alternatively written 0𝔽, is defined as the double-precision floating point positive zero value. In certain contexts, it may also be written as +0 for clarity.

So in the context of the above step, `0` and `*+0*` would mean the same thing.

annevk: Okay, and in the other contexts there is ambiguity? E.g., there's one set of steps that has n > 0 and then a following step with n is +0. It seems a little arbitrary.

jmdyck: In cases where this PR has used `*+0*`, I don't think `0` would have been ambiguous.

@ljharb, do you have a criterion/guideline for choosing between `*+0*` and `0`?

jmdyck:
(Should have included this in my review:) All occurrences of `+0` and `±0` introduced by this PR should probably be surrounded in asterisks.

spec.html
ljharb: This is likely an oversight; I've tried to avoid ever having an unsigned `0`. I'll change this to `+0`.

spec.html
```diff
@@ -1924,31 +1924,31 @@ <h1>BigInt::sameValueZero ( _x_, _y_ )</h1>
...
           <h1>BinaryXor ( _x_, _y_ )</h1>
           <emu-alg>
-            1. Assert: _x_ is 0 or 1.
-            1. Assert: _y_ is 0 or 1.
-            1. If _x_ is 1 and _y_ is 0, return 1.
-            1. Else if _x_ is 0 and _y_ is 1, return 1.
-            1. Else, return 0.
+            1. Assert: _x_ is *0n* or *1n*.
+            1. Assert: _y_ is *0n* or *1n*.
+            1. If _x_ is 1 and _y_ is *0n*, return *1n*.
```
```suggestion
            1. If _x_ is *1n* and _y_ is *0n*, return *1n*.
```

spec.html
```diff
@@ -27994,9 +27994,9 @@ <h1>Year Number</h1>
         <emu-eqn id="eqn-TimeFromYear" aoid="TimeFromYear">TimeFromYear(_y_) = msPerDay &times; DayFromYear(_y_)</emu-eqn>
         <p>A time value determines a year by:</p>
         <emu-eqn id="eqn-YearFromTime" aoid="YearFromTime">YearFromTime(_t_) = the largest integer _y_ (closest to positive infinity) such that TimeFromYear(_y_) &le; _t_</emu-eqn>
-        <p>The leap-year function is 1 for a time within a leap year and otherwise is zero:</p>
+        <p>The leap-year function is 1 for a time within a leap year and otherwise is *+0*:</p>
         <emu-eqn id="eqn-InLeapYear" aoid="InLeapYear">InLeapYear(_t_)
-          = 0 if DaysInYear(YearFromTime(_t_)) = 365
+          = *+0* if DaysInYear(YearFromTime(_t_)) = 365
           = 1 if DaysInYear(YearFromTime(_t_)) = 366
```
jmdyck: I'd suggest putting the two occurrences of "1" in asterisks as well, for balance.

spec.html
```diff
@@ -29766,7 +29766,7 @@ <h1>String.prototype.repeat ( _count_ )</h1>
           1. Let _n_ be ? ToInteger(_count_).
           1. If _n_ &lt; 0, throw a *RangeError* exception.
           1. If _n_ is *+&infin;*, throw a *RangeError* exception.
-          1. If _n_ is 0, return the empty String.
+          1. If _n_ is *+0*, return the empty String.
```
```suggestion
          1. If _n_ is *±0*, return the empty String.
```
Because `_count_` might be `-0`.

spec.html
```diff
@@ -31131,7 +31131,7 @@ <h1>Assertion</h1>
         <emu-alg>
           1. Return an internal AssertionTester closure that takes a State argument _x_ and performs the following steps:
             1. Let _e_ be _x_'s _endIndex_.
-            1. If _e_ is zero, return *true*.
+            1. If _e_ is 0, return *true*.
```
```suggestion
            1. If _e_ is *+0*, return *true*.
```

michaelficarra:
@ljharb Can you rebase and resolve conflicts?

michaelficarra:
@ljharb It seems you've missed two uses of "zero" in [21.2.2.5.1 Runtime Semantics: RepeatMatcher](https://deploy-preview-1086--ecma262-snapshots.netlify.com/#sec-runtime-semantics-repeatmatcher-abstract-operation) that would be appropriate to change.

spec.html
```diff
@@ -1939,31 +1939,31 @@ <h1>BigInt::sameValueZero ( _x_, _y_ )</h1>
         <emu-clause id="sec-binaryand" aoid="BinaryAnd">
           <h1>BinaryAnd ( _x_, _y_ )</h1>
           <emu-alg>
-            1. Assert: _x_ is 0 or 1.
-            1. Assert: _y_ is 0 or 1.
-            1. If _x_ is 1 and _y_ is 1, return 1.
-            1. Else, return 0.
+            1. Assert: _x_ is *0n* or *1n*.
```
bakkot: Hm, I'm not sure about this change. [BigIntBitwiseOp](https://tc39.es/ecma262/#sec-bigintbitwiseop), the caller of these operations, deals with unsuffixed numbers.

jmdyck: Yeah, those should probably be changed as well (in this PR or not, I don't mind). (Presumably this PR didn't touch BigIntBitwiseOp because it doesn't have any occurrences of "is zero" or "is 0".)

jmdyck (commented):
This PR has been completely overtaken by the merge of PR #2007, and should probably just be closed.

ljharb:
There's still one occurrence left of "is zero" that should be an explicit number; I've rebased this PR with that one change intact.

spec.html
```diff
@@ -4197,7 +4197,7 @@ <h1>ToBoolean ( _argument_ )</h1>
               String
             </td>
             <td>
-              If _argument_ is the empty String (its length is zero), return *false*; otherwise return *true*.
```
jmdyck, line 4200: No, the length of a String is a mathematical value.

ljharb: ah right, good call.

bakkot (approved):
WFM

jmdyck (approved):
No objection.

michaelficarra (approved):

moneyglitxh (commented):
Thanks

PR #1088, opened 2018-02-02
add a LICENSE file

michaelficarra:
The license text was taken from Appendix H, [Copyright & Software License](https://tc39.github.io/ecma262/#sec-copyright-and-software-license).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,34 @@
+All Software contained in this document ("Software") is protected by copyright
+and is being made available under the "BSD License", included below. This
+Software may be subject to third party rights (rights from parties other than
+Ecma International), including patent rights, and no licenses under such third
+party rights are granted under this license even if the third party concerned
+is a member of Ecma International. SEE THE ECMA CODE OF CONDUCT IN PATENT
+MATTERS AVAILABLE AT http://www.ecma-international.org/memento/codeofconduct.htm
+FOR INFORMATION REGARDING THE LICENSING OF PATENT CLAIMS THAT ARE REQUIRED TO
+IMPLEMENT ECMA INTERNATIONAL STANDARDS.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of the authors nor Ecma International may be used to
+  endorse or promote products derived from this software without specific
+  prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,34 @@
+All Software contained in this document ("Software") is protected by copyright
+and is being made available under the "BSD License", included below. This
+Software may be subject to third party rights (rights from parties other than
+Ecma International), including patent rights, and no licenses under such third
+party rights are granted under this license even if the third party concerned
+is a member of Ecma International. SEE THE ECMA CODE OF CONDUCT IN PATENT
+MATTERS AVAILABLE AT http://www.ecma-international.org/memento/codeofconduct.htm
```
mathiasbynens: Let’s use `https://` here?

michaelficarra: If you want that, I'd say you should PR the LIcense section of the spec first.

ljharb: Should we move the license section of the spec to this file, and have the build process dynamically include it, instead of having two sources of truth?

mathiasbynens: https://github.com/bterlson/ecmarkup/pull/130

littledan:
While I'd be very happy to have this license apply to the specification text, my understanding was that the license applies to source code and pseudo-code, while the main specification is licensed under the [Ecma text copyright policy](http://www.ecma-international.org/memento/Ecma%20copyright%20FAQ.htm). @ecmageneva @allenwb Any guidance here?

Note that this license would be, by itself, insufficient for our purposes. It's very important to also have the [Royalty Free Patent Policy](https://www.ecma-international.org/memento/Policies/Ecma_Royalty-Free_Patent_Policy_Extension_Option.htm) also apply to contributions. For this reason, TC39 has agreed that all non-member contributors of normative changes to the specification are required to sign [this form](https://tc39.github.io/agreements/contributor/). Ecma members in the TC39 "Royalty Free Task Group" have already signed analogous agreements with Ecma. If we want to have something populate our LICENSE file, maybe we should write something more detailed that reflects the IP status of the specification and explains how to contribute.

allenwb:
@littledan That's right the software license only applies to "code" and "pseudo-code".  Exactly which parts of the spec. are such code is something lawyers/judges would decided if it ever went to a court.

Also not that the licenses (for prose and code) as well as the patent pledges for final GA approved standards/reports is different from what applies to drafts, working documents, and member contributions. Most things on the github site would fall into that latter category.

mathiasbynens:
To elaborate on what @allenwb just said:

> Also note that the licenses (for prose and code) as well as the patent pledges for final GA approved standards/reports is different from what applies to drafts, working documents, and member contributions.

See <https://github.com/bterlson/ecmarkup/tree/master/boilerplate> for the different licenses.

littledan:
OK, given all of this, I'd rather not check in a LICENSE file with these contents, as it could be misread as saying that the BSD license applies to everything in the repository (which it does not).

bterlson:
I agree that the LICENSE text is misleading as is in this PR.

bterlson:
Sorry commented too soon. We'd either need multiple license files describing the different kinds of assets, or multiple licenses in a single file? Either way a standard LICENSE seems fundamentally misleading for this repo to me. I'd close this unless @michaelficarra disagrees!

LICENSE
mathiasbynens: @michaelficarra The upstream PR was merged. Can we update this PR too?

michaelficarra:
I'd appreciate a LICENSE file, even if it is more complicated than what I've included here. Someone shouldn't have to render the spec to figure out if they're allowed to render the spec (after having acquired the code). It's especially problematic since the legal text isn't even included in this repo; it's added in by ecmarkup.

littledan:
@michaelficarra I'd appreciate an accurate license file too; it'd be great to have a file that summarizes the IP situation referenced from this thread.

ljharb:
Closed by #1386.

Issue #632, opened 2016-07-06
Duplicate parameters of functions whose bodies are strict


bakkot:
It looks to me that function declarations whose bodies are strict but which are not within strict code are permitted to have duplicated parameter names, and I suspect this is a mistake.
## 

We have the following partial definition for strict mode code:

> Code is interpreted as [strict mode code](https://tc39.github.io/ecma262/#sec-strict-mode-code) in the following situations: [...]
> 
> Function code is strict mode code if the associated FunctionDeclaration [...] is contained in strict mode code or if the code that produces the value of the function's [[ECMAScriptCode]] internal slot begins with a Directive Prologue that contains a Use Strict Directive.
> 
> [Function code](https://tc39.github.io/ecma262/#sec-types-of-source-code) is source text that is parsed to supply the value of the [[ECMAScriptCode]] and [[FormalParameters]] internal slots (see 9.2) of an ECMAScript function object.

We also have [these Early Errors](https://tc39.github.io/ecma262/#sec-function-definitions-static-semantics-early-errors) for FunctionDeclarations:

> `FunctionDeclaration : function BindingIdentifier ( FormalParameters ) { FunctionBody }`
> If the source code matching this production is strict mode code, the Early Error rules for `StrictFormalParameters : FormalParameters` are applied.
> 
> `StrictFormalParameters : FormalParameters`
> It is a Syntax Error if BoundNames of FormalParameters contains any duplicate elements.

As I read this, "the source code matching this production [i.e. `FunctionDeclaration : function BindingIdentifier ( FormalParameters ) { FunctionBody }`]" is _not_ strict mode code just because its body contains a Use Strict Directive; only the `FormalParameters` and `FunctionBody` parts are. As such, the the Early Error rules for `StrictFormalParameters : FormalParameters` are _not_ applied.

Hence, a Script like `function foo(a, a){'use strict'}` would not have any Early Errors.

This seems wrong, and implementations do indeed enforce this Early Error. Either way, I think it would be good to have a NOTE here clarifying whether the strict-mode Early Errors for parameters and function names are intended to apply to function declarations (and expressions) which contain a Use Strict Directive but are not themselves inside of strict mode code.

See related issue [previously](https://bugs.ecmascript.org/show_bug.cgi?id=4243). (Also, when are we going to get the Bugzilla cert fixed?)

getify:
IANASA (I am not a spec author), but...

[...]

PR #1091, opened 2018-02-04
Normative: Expand text included in "function code"

jugglinmike:
Previously reported at https://bugs.ecmascript.org/show_bug.cgi?id=4243 This patch is at attempt to implement the solution agreed upon there. (I think this issue was simply overshadowed by a related discussion on non-simple parameter lists, which ultimately lead to the restriction introduced in ES2016.)

The current specification text allows both of the following forms:

    function static() { 'use strict'; }
    (function static() { 'use strict'; });

Existing implementations violate that text by producing parsing errors for both.

With this patch applied, only the FunctionExpression will be disallowed, so implementations will continue to be in violation of the spec.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9371,7 +9371,7 @@
         <em>Eval code</em> is the source text supplied to the built-in `eval` function. More precisely, if the parameter to the built-in `eval` function is a String, it is treated as an ECMAScript |Script|. The eval code for a particular invocation of `eval` is the global code portion of that |Script|.
       </li>
       <li>
-        <em>Function code</em> is source text that is parsed to supply the value of the [[ECMAScriptCode]] and [[FormalParameters]] internal slots (see <emu-xref href="#sec-ecmascript-function-objects"></emu-xref>) of an ECMAScript function object. The function code of a particular ECMAScript function does not include any source text that is parsed as the function code of a nested |FunctionDeclaration|, |FunctionExpression|, |GeneratorDeclaration|, |GeneratorExpression|, |AsyncFunctionDeclaration|, |AsyncFunctionExpression|, |MethodDefinition|, |ArrowFunction|, |AsyncArrowFunction|, |ClassDeclaration|, or |ClassExpression|.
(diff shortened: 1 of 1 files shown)


bterlson:
@ljharb @bmeck either of you want to take a stab at reviewing this? I don't want to pull this for 2018, so it will sit here at least until I create the 2018 branch later this week.

bmeck:
I'll get it done in next few days.

bakkot:
This looks good to me, but there's a closely related issue (#632) which this doesn't resolve. Actually, it maybe exacerbates that issue: if this is intended to resolve the bugzilla bug by making `function static() { 'use strict'; }` unambiguously not an error, that would imply `function f(a, a) { 'use strict'; }` would also unambiguously not be an error. That seems wrong.

Possibly we should fix #632 by some other means, but I'd be OK rewording the first two early errors [here](https://tc39.github.io/ecma262/#sec-function-definitions-static-semantics-early-errors) as part of this PR.

There's also some discussion of this problem with @RReverser at https://github.com/tc39/test262-parser-tests/issues/8#issuecomment-302775433.

bakkot:
By the way, this is probably needs-consensus: if I recall correctly, some people (maybe @ajklein?) were in favor of `function static() { 'use strict'; }` being an error, since it already is in all engines.

bmeck:
@jugglinmike is there a reason this isn't being applied to `FunctionDeclaration`s as well. The main note I see is https://bugs.ecmascript.org/show_bug.cgi?id=4243#c8 which I'm not quite sure I agree on some things being related to different scopes and differing in behavior based upon if they are used as an expression vs a declaration. In particular, the early error would not be related to the hoisted binding that gets applied to a sloppy outer scope, so I'm not sure the two concepts should be related.

jugglinmike:
> if this is intended to resolve the bugzilla bug by making `function eval() {
> 'use strict'; }` unambiguously not an error, that would imply `function f(a,
> a) { 'use strict'; }` would also unambiguously not be an error. That seems
> wrong.

@bakkot That was my intent. It doesn't seem wrong to me because as per this
patch, I consider the "function code" contour to be different for expressions
and declarations.

But to @bmek's point, it sounds like that is a decision from the BugZilla
discussion that we're revisiting here. If that's right, then we should decide
if it's better to have a simple definition of "function code" or one that more
closely aligns with the runtime semantics.

bakkot:
> It doesn't seem wrong to me because as per this patch, I consider the "function code" contour to be different for expressions and declarations.

@jugglinmike, the thing I'm pointing at here is that the [same wording](https://tc39.github.io/ecma262/#sec-function-definitions-static-semantics-early-errors) prohibits `function eval() { 'use strict'; }` as prohibits `function f(a, a) { 'use strict'; }` (note the duplicate parameter), and I think we definitely do want to keep the second one of these as an error regardless of how we come down on the function name issue.

jugglinmike:
@bakkot I believe the proposed modifications for each of these issues makes them orthogonal:

- this patch concerns whether function body strictness extends to the "name"
- gh-632 concerns whether function body strictness extends to the early error for duplicate parameters

It seems as though gh-632 can be resolved with a change dedicated to that issue (which I would be happy to write).

jugglinmike:
I raised this at the March 2019 TC39 meeting ([slides available here](http://jugglinmike.github.io/presentations/2018/tc39-fn-name/)), and the committee elected to change the specification to match current implementation status--that is, to unconditionally include the binding identifier in the definition of "function code". This more restrictive interpretation matches the majority of implementations today, but requires a different change to the specification. I'll follow up with a modified proposal within the next few days.

Issue #253, opened 2015-12-15
Notate abstract ops that can and cannot throw


domenic:
In streams, which is heavily promise-based, there are a lot of abstract operations that cannot throw. The ES spec has these too, in smaller number.

Of course, this kind of nothrow vs. throw characteristic can be contagious. If you only use nothrow abstract ops in your abstract op, then your abstract op is nothrow. Whereas if you use any throwing ops, you are now throwing. Unless you are certain for external reasons that the given throw operation will not throw in this context, in which case you might want to assert that it is a non-abrupt-completion.

There are a couple action items here, some of which can be done independently:
- Come up with a convention for asserting that the return value is never an abrupt completion, e.g. `Let _result_ be ! AbstractOp()`. Use it throughout. A lot of places in the spec currently use "perform" for this purpose, which is not defined anywhere.
- Introduce EMU support for notating with throws/nothrow. (Or perhaps, abrupt/noabrupt.) You can see this in action in Streams, e.g. https://streams.spec.whatwg.org/#rs-abstract-ops. For the source format, I'd suggest either `<emu-clause nothrow>` or `<emu-clause flair="nothrow">`.
- Go through the ES spec and update all abstract ops with whether or not they can throw. This could be done incrementally.
- Build automated checker tools to enforce the contagion rules, based on throw/nothrow and ! or ?.

ljharb:
What about the case where the result of `AbstractOp` is not needed? That's usually the cases where "Perform" appears. Having a useless `_result_` doesn't seem ideal. Do you have an alternative suggestion for an abstract op that solely causes side effects (some of which can throw, and some of which can't)?

In general I love this idea - being explicit about what can throw and what can't seems like an overall improvement.

domenic:
> Do you have an alternative suggestion for an abstract op that solely causes side effects (some of which can throw, and some of which can't)?

`1. ? AbstractOp()` to bubble errors, or `1. ! AbstractOp()` to assert it never errors. You could introduce the "perform" verb for that but I'd like it to be defined this time.

ljharb:
Sounds good - I like having an explicit "this is for side effects" mechanism like "Perform".

jmdyck:
Here's what I said when this was raised in the [ecmarkup repo](https://github.com/bterlson/ecmarkup/issues/19):

[...]

PR #1096, opened 2018-02-09
Editorial: add a SetFunctionLength abstract operation

tobie:


spec.html
```diff
@@ -7536,6 +7535,18 @@ <h1>SetFunctionName ( _F_, _name_ [ , _prefix_ ] )</h1>
     </emu-clause>
 
     <!-- es6num="9.2.12" -->
+    <emu-clause id="sec-setfunctionlength" aoid="SetFunctionLength">
+      <h1>SetFunctionLength ( _F_, _length_)</h1>
+      <p>The abstract operation SetFunctionLength requires a Function argument _F_, and a Number argument _length_. This operation adds a `length` property to _F_ by performing the following steps:</p>
+      <emu-alg>
+        1. Assert: _F_ is an extensible object that does not have a `length` own property.
+        1. Assert: Type(_length_) is Number.
+        1. Assert: ToLength(_length_) is equal to _length_.
```
tobie: Unsure if that's the best way to assert that `length` is an integer within the right bounds.

spec.html
```diff
@@ -7413,9 +7413,8 @@ <h1>FunctionAllocate ( _functionPrototype_, _strict_, _functionKind_ )</h1>
       <h1>FunctionInitialize ( _F_, _kind_, _ParameterList_, _Body_, _Scope_ )</h1>
       <p>The abstract operation FunctionInitialize requires the arguments: a function object _F_, _kind_ which is one of (Normal, Method, Arrow), a parameter list Parse Node specified by _ParameterList_, a body Parse Node specified by _Body_, a Lexical Environment specified by _Scope_. FunctionInitialize performs the following steps:</p>
       <emu-alg>
-        1. Assert: _F_ is an extensible object that does not have a `length` own property.
         1. Let _len_ be the ExpectedArgumentCount of _ParameterList_.
-        1. Perform ! DefinePropertyOrThrow(_F_, `"length"`, PropertyDescriptor{[[Value]]: _len_, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true*}).
+        1. Perform ! SetFunctionLength(_F_, _len_).
```
tobie: Not sure whether this should be prefixed with "!". I assume it does, but then SetFunctionName often isn't elsewhere in the spec and I'm not sure why.

tobie:
Thanks. As mentioned over irc, I'll make especially good use of it in WebIDL.

spec.html
jmdyck: If an operation cannot *ever* return an abrupt completion, then I'd say calls to it shouldn't be preceded by '!'. It isn't *wrong*, but it's unnecessary clutter. (This would be more persuasive if you could easily tell from an operation's 'declaration' whether it ever returns an abrupt completion. See issue #253.)

tobie: So just to clarify my understanding here:

While `DefinePropertyOrThrow` can throw in some circumstances, it can't in this case, because we know from the first assertion in `SetFunctionLength` that `DefinePropertyOrThrow`'s first argument is extensible and doesn't have a `length` own prop which could interfere with setting.

Thus we know that `SetFunctionLength` won't ever throw.

Preceding an abstract operation call by "!" means that we are in a situation where an abstract operation won't throw, _although it might do so in different circumstances_.

Operations which never throw should never be prefixed with "!"  even though we can't know from looking at the abstract operation's signature or its name that this is the case.

So for now, `SetFunctionLength` should not be prefixed with anything?

Is my understanding correct?

spec.html
```diff
@@ -7536,6 +7535,18 @@ <h1>SetFunctionName ( _F_, _name_ [ , _prefix_ ] )</h1>
...
+      <p>The abstract operation SetFunctionLength requires a Function argument _F_, and a Number argument _length_. This operation adds a `length` property to _F_ by performing the following steps:</p>
+      <emu-alg>
+        1. Assert: _F_ is an extensible object that does not have a `length` own property.
+        1. Assert: Type(_length_) is Number.
+        1. Assert: ToLength(_length_) is equal to _length_.
+        1. Return ! DefinePropertyOrThrow(_F_, `"length"`, PropertyDescriptor{[[Value]]: _length_, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true*}).
+      </emu-alg>
+    </emu-clause>
+
+    <!-- es6num="9.2.13" -->
```
bmeck: To my understanding we should not be moving/adding es6num since it is referring to where a clause was in the ES2015 specification: http://www.ecma-international.org/ecma-262/6.0/

spec.html
```diff
@@ -7413,9 +7413,8 @@ <h1>FunctionAllocate ( _functionPrototype_, _strict_, _functionKind_ )</h1>
       <h1>FunctionInitialize ( _F_, _kind_, _ParameterList_, _Body_, _Scope_ )</h1>
       <p>The abstract operation FunctionInitialize requires the arguments: a function object _F_, _kind_ which is one of (Normal, Method, Arrow), a parameter list Parse Node specified by _ParameterList_, a body Parse Node specified by _Body_, a Lexical Environment specified by _Scope_. FunctionInitialize performs the following steps:</p>
       <emu-alg>
-        1. Assert: _F_ is an extensible object that does not have a `length` own property.
         1. Let _len_ be the ExpectedArgumentCount of _ParameterList_.
-        1. Perform ! DefinePropertyOrThrow(_F_, `"length"`, PropertyDescriptor{[[Value]]: _len_, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true*}).
+        1. Perform SetFunctionLength(_F_, _len_).
```
ljharb: I disagree with the above; this should be prefixed with !.

Noise is less of a concern to me than not knowing whether an abstract operation might throw or not.

spec.html
```diff
@@ -24700,7 +24710,7 @@ <h1>Function.prototype.bind ( _thisArg_, ..._args_ )</h1>
               1. Let _targetLen_ be ToInteger(_targetLen_).
               1. Let _L_ be the larger of 0 and the result of _targetLen_ minus the number of elements of _args_.
           1. Else, let _L_ be 0.
-          1. Perform ! DefinePropertyOrThrow(_F_, `"length"`, PropertyDescriptor {[[Value]]: _L_, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true*}).
+          1. Perform SetFunctionLength(_F_, _L_).
```
ljharb: This also needs a !

spec.html
jmdyck: Your understanding matches mine. Whether it's *correct* is up to the editors.

spec.html
```diff
@@ -7535,6 +7534,17 @@ <h1>SetFunctionName ( _F_, _name_ [ , _prefix_ ] )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-setfunctionlength" aoid="SetFunctionLength">
+      <h1>SetFunctionLength ( _F_, _length_ )</h1>
+      <p>The abstract operation SetFunctionLength requires a Function argument _F_, and a Number argument _length_. This operation adds a `length` property to _F_ by performing the following steps:</p>
+      <emu-alg>
+        1. Assert: _F_ is an extensible object that does not have a `length` own property.
```
ljharb: Should this also assert on IsCallable? Should it call the internal IsExtensible methodminstead of using “is extensible” prose?

spec.html
jmdyck: Would you prefix *all* currently-unprefixed operation-invocations with "!"? There's about 2300 of them.

Alternatively, we could come up with a naming convention that distinguished those operations that can return abrupt completions from those that can't. That would convey the information you want to see at the invocation-point, in a maybe less obtrusive way.

Personally, I think I'd prefer if information about what the operation can return were only a click or mouse-hover away.

jmdyck: (I wonder if we should move this discussion to a separate issue.)

ljharb: Yes, i would, and yes, I’d say that’d be a separate issue.

ljharb: That there’s 2300 mean it’d be a tedious conversion; when I’m reading a specific part of the spec there wouldn’t be 2300 - there’d be just enough so I’d have to jump elsewhere as rarely as possible.

spec.html
domenic: I don't agree with this; I think we should use ! everywhere. (I am also not an editor.)

spec.html
tobie: I copied the proses used elsewhere in the spec word for word. I suggest changing all of it in a separate pull request.

ljharb: Sounds good :-)

spec.html
jmdyck: Separate issue now created: #1097

tobie:
What's the outcome of the discussion as to whether calls to `SetFunctionLength` should be prefixed by "!" _in this particular pull request_? I don't have an opinion either way, but would like to be able to move forward with this PR.

spec.html
```diff
@@ -7535,6 +7534,17 @@ <h1>SetFunctionName ( _F_, _name_ [ , _prefix_ ] )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-setfunctionlength" aoid="SetFunctionLength">
+      <h1>SetFunctionLength ( _F_, _length_ )</h1>
+      <p>The abstract operation SetFunctionLength requires a Function argument _F_, and a Number argument _length_. This operation adds a `length` property to _F_ by performing the following steps:</p>
```
anba: Nit: No comma here because there only two arguments.

spec.html
```diff
@@ -7535,6 +7534,17 @@ <h1>SetFunctionName ( _F_, _name_ [ , _prefix_ ] )</h1>
...
+      <emu-alg>
+        1. Assert: _F_ is an extensible object that does not have a `length` own property.
+        1. Assert: Type(_length_) is Number.
+        1. Assert: ToLength(_length_) is equal to _length_.
```
anba: ToLength is not correct here, because if called from `Function.prototype.bind` `length` can exceed 2^53-1. If a bounds check is wanted here, something like
  1. Assert: _length_ &ge; 0 and ToInteger(_length_) is equal to _length_.

is necessary.

ljharb:
@tobie yes, please include the ! in this PR.

tobie:
Hey, any idea what the timeline looks like to get this merged in?

bterlson:
Sorry for the delay. It looks great.

FWIW I think we should be consistently using !/? operators until it is made more clear which abstract ops can never throw.

tobie:
Thanks a lot. Getting put to good use here: https://github.com/heycam/webidl/pull/549. :)

PR #1102, opened 2018-02-16
Normative: add RegExp lookbehind to annex-B

mysticatea:
This PR adds the syntax of RegExp Lookbehind Assertions into B.1.4 Regular Expressions Patterns section.

Since #1029 didn't look to update the Annex-B section but I think it's needed (https://github.com/tc39/ecma262/pull/1029#issuecomment-364003143).

I apologize if it's intentional.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -40776,6 +40776,8 @@ THH:mm:ss.sss
           [+U] `(` `?` `=` Disjunction[+U, ?N] `)`
           [+U] `(` `?` `!` Disjunction[+U, ?N] `)`
           [~U] QuantifiableAssertion[?N]
+          `(` `?` `&lt;=` Disjunction[?U, ?N] `)`
+          `(` `?` `&lt;!` Disjunction[?U, ?N] `)`
 
         QuantifiableAssertion[N] ::
           `(` `?` `=` Disjunction[~U, ?N] `)`


mathiasbynens:
cc @anba @jmdyck @claudepache @IgnoredAmbience @littledan

mathiasbynens:
This indeed looks like an oversight to me.

spec.html
```diff
@@ -40918,7 +40922,7 @@ <h1>Pattern Semantics</h1>
           1. Return _m_.
         </emu-alg>
 
-        <p>Assertion (<emu-xref href="#sec-assertion"></emu-xref>) evaluation rules for the <emu-grammar>Assertion :: `(` `?` `=` Disjunction `)`</emu-grammar> and <emu-grammar>Assertion :: `(` `?` `!` Disjunction `)`</emu-grammar> productions are also used for the |QuantifiableAssertion| productions, but with |QuantifiableAssertion| substituted for |Assertion|.</p>
+        <p>Assertion (<emu-xref href="#sec-assertion"></emu-xref>) evaluation rules for the <emu-grammar>Assertion :: `(` `?` `=` Disjunction `)`</emu-grammar>, <emu-grammar>Assertion :: `(` `?` `!` Disjunction `)`</emu-grammar>, <emu-grammar>Assertion :: `(` `?` `&lt;=` Disjunction `)`</emu-grammar>, and <emu-grammar>Assertion :: `(` `?` `&lt;!` Disjunction `)`</emu-grammar> productions are also used for the |QuantifiableAssertion| productions, but with |QuantifiableAssertion| substituted for |Assertion|.</p>
```
littledan: Do we actually want to enable lookbehind in QuantifiableAssertion?

littledan: We should make sure to have a test262 test one way or the other.

mysticatea: Just a note: I confirmed it on Chrome 65 since Annex B is the section for browser compatibility. Lookbehind assertions are quantifiable on the Chrome.

claudepache:
We have to make a conscious decision between:

(A) be consistent with other assertions of the form `(?...)` wrt QuantifiableAssertion; or,
(B) limit QuantifiableAssertion to what is strictly needed for legacy reason.

This PR apparently chose (A). I would prefer (B).

(Also, because a choice is needed, this PR is not ”editorial”.)

spec.html
littledan: Good to know, though I think it's still within our ability to make this choice one way or another.

cc @hashseed @schuay

hashseed:
So if I understand correctly, the question is whether to allow something like `(?<=a){2,3}b`.

I honestly don't see a reason why this should be allowed since these assertions do not consume any characters, and therefore seem rather pointless.

spec.html
```diff
@@ -40775,11 +40775,15 @@ <h2>Syntax</h2>
...
           [+U] `(` `?` `!` Disjunction[+U, ?N] `)`
+          [+U] `(` `?` `&lt;=` Disjunction[?U, ?N] `)`
+          [+U] `(` `?` `&lt;!` Disjunction[?U, ?N] `)`
           [~U] QuantifiableAssertion[?N]
 
         QuantifiableAssertion[N] ::
           `(` `?` `=` Disjunction[~U, ?N] `)`
           `(` `?` `!` Disjunction[~U, ?N] `)`
+          `(` `?` `&lt;=` Disjunction[?U, ?N] `)`
+          `(` `?` `&lt;!` Disjunction[?U, ?N] `)`
```
jmdyck: It's an error to use '?U' here, because QuantifiableAssertion doesn't have a 'U' parameter.

littledan:
OK, seems like we're pretty unanimous in this thread so far that QuantifiableAssertion should not support positive or negative lookbehinds. @mysticatea would you be up for writing a new patch with that change, and then we can discuss it at the March 2018 TC39 meeting to confirm broader consensus?

mysticatea:
I updated this PR.

hashseed:
FWIW we have [this comment](https://cs.chromium.org/chromium/src/v8/src/regexp/regexp-parser.cc?l=1989&rcl=22ef1d88f2c7dc1f8c296e4ed4f32275915371fc) in V8 implemented back in [January 2016](https://codereview.chromium.org/1645573002):
```comment
// With /u, lookarounds are not quantifiable.
```

I guess the change would be to
 - disallow lookbehinds for quantifiable assertions
 - disallow lookaheads for quantifiable assertions in unicode mode
 - allow lookahead for quantifiable assertions in non-unicode mode

jmdyck:
> I think that's what the current patch does.

I think all three of @hashseed's bullets are accomplished in the current *spec*. This patch does something else.

hashseed:
V8 fix in flight: https://chromium-review.googlesource.com/c/v8/v8/+/926102

littledan:
@jmdyck How does the current spec support lookbehind at all when taking the Annex B grammar into account?

jmdyck:
The current spec doesn't, this patch does. (Allowing lookbehinds wasn't one of @hashseed's bullets.)

littledan:
@jmdyck I think we actually do want to enable lookbehinds, just not as a quantifiable assertion. Given that goal, do you think this patch seems good?

jmdyck:
Yes, the patch is necessary. 21.2.1 has lookbehinds, so annex B must have them too.

littledan:
See tests at https://github.com/tc39/test262/pull/1456

littledan:
At this point, we have
- Spec text, written by someone who I verified signed the IP form
- Test262 tests
- A full implementation in V8

Do we have consensus on this normative change? I've added it to the March 2018 meeting agenda so we can confirm consensus there if necessary.

bterlson:
Thanks for this PR, and sorry for the delay in getting it pulled in!

Issue #1110, opened 2018-02-20
The usage of the term "key" is ambiguous in relation to object properties


ogwh:
It seems there are (at least) two definitions of the term "key" within the specification that are not differentiated/explicitly defined.

The most common definition is summarised as: an object property identifier of type string or symbol.
For discussion I'll call this *pid* (for property ID).

The other definition is more traditional: a string that serves to identify an object in a hash table/map.
For discussion I'll call this *hid* (for hash table ID).

Where this generates confusion relates to the common usage of an object as a key/value store. In that use case a key is a *hid*. In all other use cases a key is a *pid*. This ambiguity is propagated throughout the language itself, with some functions on Object using one definition and others using the other without remark.

Some examples of Object functions that use *pid* (those that include symbol keys):

- assign: Assigns all properties including symbols
- create: Second param defines properties, including symbol properties
- defineProperty and defineProperties: self-explanatory
- seal: also seals symbol properties

Some examples of Object functions that use *hid*:

- keys: Only returns keys that are of type string
- entries: Only returns key-value pairs where the keys are strings
- values: Only returns values whose keys are strings

The above lists cover a good number of functions but they're not exhaustive.

Note that the documentation for those three functions on MDN states that all enumerable properties are included, but this is not the case. It's necessary to read the spec to see that these functions only include string keys currently (which is unrealistic to expect from everyone, however desirable).

For convenience this MDN page links to all functions on Object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object

Aside from the MDN documentation this has also caused confusion for the Typescript contributors (which is how I became aware of the problem):

Difficulty deciding on the correct type of the return value of Object.keys: https://github.com/Microsoft/TypeScript/pull/12253#issuecomment-263132208

Ambiguity causing the TS *keyof* keyword to be implemented using the *hid* definition of a key rather than the *pid* definition. Any change to which would now cause breaking changes to all Typescript programs: https://github.com/Microsoft/TypeScript/issues/20721 and https://github.com/Microsoft/TypeScript/issues/21983

In short it seems that the dual use of Object as both a hash table and as the base "object" type has generated (and will likely continue to generate) confusion because it's so easy to overlook.

For specifics with references to a few relevant parts of the ECMA specification itself see my comment on this issue here: https://github.com/Microsoft/TypeScript/issues/21983#issuecomment-366237751

[...]

PR #1113, opened 2018-02-21
Editorial: rename EnumerableOwnProperties to EnumerableOwnPropertyNames

bterlson:
See issue #1110.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "license": "SEE LICENSE IN https://tc39.github.io/ecma262/#sec-copyright-and-software-license",
   "homepage": "https://tc39.github.io/ecma262/",
   "dependencies": {
-    "ecmarkup": "^3.11.5"
+    "ecmarkup": "^3.12.0"
   },
   "devDependencies": {
     "@alrra/travis-scripts": "^2.0.0"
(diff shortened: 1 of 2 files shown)


anba:
Let's recap:

This operation was named "EnumerableOwnNames" in the [sixth](https://www.ecma-international.org/ecma-262/6.0/index.html#sec-enumerableownnames) and [seventh](https://www.ecma-international.org/ecma-262/7.0/index.html#sec-enumerableownnames) edition. And for the [eighth](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-abstract-operations) edition the "Names" suffix was dropped, probably because the operation no longer only returns property names, but also property values and property-key-value pairs. And now it's named "EnumerableOwnPropertyNames". So I guess the next step should be to remove the "Property" part, because it's kind of already implied by "Names". And then we're back to "EnumerableOwnNames" and have come full circle. :stuck_out_tongue_closed_eyes:

ljharb:
lol yeah that’s true. Maybe this should be reverted; I’m not sure i buy the conclusions in #1110 anyways. @bterlson, thoughts?

annevk:
This name change also affects HTML (albeit only a non-normative note).

PR #1116, opened 2018-02-23
Editorial: Fix typo

mathiasbynens:


bterlson:
I don't think there are enough approvals on this. Could we get a dozen more or so before I merge?

PR #1118, opened 2018-02-25
Fix Promise.prototype.finally to not incur extra ticks

raulsebastianmihaila:
https://esdiscuss.org/topic/promise-finally

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38579,8 +38579,8 @@ THH:mm:ss.sss
             1. If SameValue(_resolution_, _promise_) is *true*, then
               1. Let _selfResolutionError_ be a newly created *TypeError* object.
               1. Return RejectPromise(_promise_, _selfResolutionError_).
+            1. Let _finallySourcePromise_ be _F_.[[FinallySourcePromise]].
             1. If Type(_resolution_) is not Object, then
-              1. Let _finallySourcePromise_ be _F_.[[FinallySourcePromise]].
               1. If _finallySourcePromise_ is *undefined*
                 1. Return FulfillPromise(_promise_, _resolution_).
               1. Else
@@ -38594,7 +38594,7 @@ THH:mm:ss.sss
             1. Let _thenAction_ be _then_.[[Value]].
             1. If IsCallable(_thenAction_) is *false*, then
               1. Return FulfillPromise(_promise_, _resolution_).
-            1. Perform EnqueueJob(`"PromiseJobs"`, PromiseResolveThenableJob, &laquo; _promise_, _resolution_, _thenAction_, _F_.[[FinallySourcePromise]] &raquo;).
+            1. Perform EnqueueJob(`"PromiseJobs"`, PromiseResolveThenableJob, &laquo; _promise_, _resolution_, _thenAction_, _finallySourcePromise_ &raquo;).
             1. Return *undefined*.
           </emu-alg>
           <p>The `length` property of a promise resolve function is 1.</p>
--- a/spec.html
+++ b/spec.html
@@ -38532,14 +38532,15 @@ THH:mm:ss.sss
 
       <!-- es6num="25.4.1.3" -->
       <emu-clause id="sec-createresolvingfunctions" aoid="CreateResolvingFunctions">
-        <h1>CreateResolvingFunctions ( _promise_ )</h1>
+        <h1>CreateResolvingFunctions ( _promise_, _finallySourcePromise_ )</h1>
         <p>When CreateResolvingFunctions is performed with argument _promise_, the following steps are taken:</p>
         <emu-alg>
           1. Let _alreadyResolved_ be a new Record { [[Value]]: *false* }.
           1. Let _stepsResolve_ be the algorithm steps defined in Promise Resolve Functions (<emu-xref href="#sec-promise-resolve-functions"></emu-xref>).
-          1. Let _resolve_ be CreateBuiltinFunction(_stepsResolve_, &laquo; [[Promise]], [[AlreadyResolved]] &raquo;).
+          1. Let _resolve_ be CreateBuiltinFunction(_stepsResolve_, &laquo; [[Promise]], [[AlreadyResolved]], [[FinallySourcePromise]] &raquo;).
           1. Set _resolve_.[[Promise]] to _promise_.
           1. Set _resolve_.[[AlreadyResolved]] to _alreadyResolved_.
+          1. Set _resolve_.[[FinallySourcePromise]] to _finallySourcePromise_.
           1. Let _stepsReject_ be the algorithm steps defined in Promise Reject Functions (<emu-xref href="#sec-promise-reject-functions"></emu-xref>).
           1. Let _reject_ be CreateBuiltinFunction(_stepsReject_, &laquo; [[Promise]], [[AlreadyResolved]] &raquo;).
           1. Set _reject_.[[Promise]] to _promise_.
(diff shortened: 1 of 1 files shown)


raulsebastianmihaila:
@ljharb Please review

spec.html
```diff
@@ -39015,57 +39030,11 @@ <h1>Promise.prototype.finally ( _onFinally_ )</h1>
         <p>When the `finally` method is called with argument _onFinally_, the following steps are taken:</p>
         <emu-alg>
           1. Let _promise_ be the *this* value.
-          1. If Type(_promise_) is not Object, throw a *TypeError* exception.
+          1. If IsPromise(_promise_) is *false*, throw a *TypeError* exception.
```
ljharb: This isn’t going to work; finally must work with thenables.

spec.html
```diff
@@ -39015,57 +39030,11 @@ <h1>Promise.prototype.finally ( _onFinally_ )</h1>
...
-            1. Let _C_ be _F_.[[Constructor]].
-            1. Assert: IsConstructor(_C_) is *true*.
-            1. Let _promise_ be ? PromiseResolve(_C_, _result_).
-            1. Let _thrower_ be equivalent to a function that throws _reason_.
-            1. Return ? Invoke(_promise_, `"then"`, &laquo; _thrower_ &raquo;).
-          </emu-alg>
-          <p>The `length` property of a Catch Finally function is *1*.</p>
-        </emu-clause>
+          1. Let _resultCapability_ be ? NewPromiseCapability(_C_, _promise_).
+          1. Return PerformPromiseThen(_promise_, _onFinally_, _onFinally_, _resultCapability_).
```
ljharb: The fact that this does not observably call `.then` immediately means that a subclass has to override `.finally` - with the current design, it only has to override `.then` (which is a requirement)

spec.html
raulsebastianmihaila: Why is that? `Promise.prototype.then` does this check as well. Shouldn't `Promise.prototype.then` work with thenables as well?

spec.html
raulsebastianmihaila: I think that makes sense because in essence `then` and `finally` are different. Could you please clarify why you think otherwise? The current design is based on what I consider to be a workaround instead of being an exact expression of what `finally` means conceptually, hence the extra ticks that `finally` doesn't conceptually require.
I think `finally` should call `then` only if the behavior expressed by `then` is part of `finally`'s behavior. In my opinion it's not fully; they are similar only to a certain degree. Perhaps listing the pros and cons and the possible trade-offs and analyzing which one is better would help making the right choice.

raulsebastianmihaila:
I created the PR here because doing it in the other repo would have been much more difficult for me, as it differs a lot from this repo.

benjamn:
> It’d be more productive imo to make this PR to the proposal repo tho, where i can integrate it with my actual tests - this is going to take a lot of iteration before it’s ready to approach entering the spec.

@ljharb When I submitted [a similar PR](https://github.com/tc39/proposal-promise-finally/pull/48) to the proposal repository (exactly what you're asking @raulsebastianmihaila to do here), you [told me](https://github.com/tc39/proposal-promise-finally/pull/48#issuecomment-364986055) I should submit it to *this* repository and/or [your polyfill repo](https://github.com/es-shims/Promise.prototype.finally) instead. What gives?

I'm willing to believe there's a meaningful difference between these PRs that calls for submitting/reviewing them differently, but I feel the need to flag this apparent inconsistency in your preferences, and call out the possibility that you may be (consciously or unconsciously) derailing efforts to improve `Promise.prototype.finally` by asking for the work to be relocated, and playing up the amount of work involved (as if anyone contributing to the ecma262 spec doesn't realize it takes a lot of work), rather than just engaging with it as submitted. Please don't ask others to do needless work.

spec.html
raulsebastianmihaila: To clarify, you mean that if the subclass overrides `then`, it must then override `finally` as well, is that correct? Because a subclass doesn't have to override `then` unless it chooses to.

benjamingr: To be fair, no one likes promise subclassing anyway :D

benjamn: Though I am sympathetic to the goal of reducing the number of promises created by `Promise.prototype.finally` in general, I believe it is desirable (and widely expected) that a `Promise` subclass only has to override `then`, and all the other methods will automatically work in terms of the overridden `then`.

Regardless of any other observable semantic differences introduced by that PR, abandoning that property of `Promise` subclassing may be a deal-breaker for any approach that implements `finally` without calling `then`.

raulsebastianmihaila: @benjamn That would imply that whatever method is added to `Promise.prototype` in the future must be expressed in terms of `Promise.prototype.then` which is a limitation whose ground I'm not familiar with. In general a class doesn't have a 'heliocentric' method, and a subclass can override whatever behavior it wants to. I don't expect many people to subclass `Promise` and I certainly expect that people who subclass `Promise` know very well what they're doing, so that having to extend more than one method (if they need to) wouldn't be a deal-breaker.

spec.html
ljharb: Because of subclassing, that check must be restricted to .then so that it can be overridden.

spec.html
ljharb: It’s definitely a dealbreaker; i used to have it this way and that was a blocker for proceeding to stage 3.

ljharb:
@raulsebastianmihaila the “emu” stuff is identical to this spec.html file :-) they’re both ecmarkup (not html) no worries tho.

@benjamn sorry i was unclear; i asked to see what the spec diff would look like, but I’d intended you to update your PR in that repo (since that repo also has spec text and tests).

I’m trying to avoid needless work; specifically, the work involved of convincing the committee that this change should happen, before the necessary legwork has been done (ie, showing that the changes are even feasible).

spec.html
raulsebastianmihaila: @ljharb I assume there are meeting notes about this. Do you remember during which meeting this was discussed?

ljharb: https://github.com/rwaldron/tc39-notes/blob/b8da60318b564f136cbe8385f17f42abc0666cdd/es7/2017-01/jan-25.md#13iif-promiseprototypefinally-to-stage-3

raulsebastianmihaila: @ljharb Thank you for the link! Reading the notes I understood something else and not that it was a blocker for proceeding to the next stage. DE expressed some doubt about modifying `catch`. DE also said that if `then` is the builtin `then` then some optimization can be done. But if I understand correctly he meant that even though `then` must be looked up, optimizations are still possible. I don't see how not having to look up `then` would result in less optimized runtime. Then yourself said that it's good for `finally` to be consistent with `catch`. JLZ said that Angular might depend on the existing behavior of `catch`. AK didn't want the proposal to proceed because it would have been possible that the spec would be changed after that (he didn't say that he disagreed with `finally` not calling `then`). AWB expressed indifference. MM was being poetical.

People were pretty clear that changing `catch` could be problematic. But what I consider to be the essence of the discussion were having here, namely whether or not `finally` is conceptually expressed in terms of `then`, was not touched in that meeting. Perhaps if this matter was brought into the discussion more clearly, people would have looked at it differently. These notes don't show that this aspect was discussed seriously.

Nobody seemed firm about `finally` having to be consistent with `catch`. Somebody (we don't know who) even asked what was the motivation about not calling `then`. I think yourself were the most 'firm', by saying that it seems that consistency is the way to go, which doesn't seem very firm either. I think consistency with `catch` received a lot more attention than conceptual consistency. And in general the debate gives the impression of uncertainty. This contrasts a lot with saying that 'it was a blocker'.

ljharb: The general consensus in the room - or at least, the only way to achieve consensus, which is effectively the same - was that either a) `.catch` and `.finally` should both directly delegate to `.then`, or, b) `.catch` and `.finally` should both call abstract operations directly. "b" is not web-compatible.

I'm not going to enumerate which committee members were blockers there, both to not throw folks under the bus and because it's irrelevant (the notes are not transcripts, intentionally) - I hope you can simply trust that what I've said above is the case.

benjamingr: Can we phrase it so that engines have to implement `finally` through `then` only for non-subclasses? That would be sufficient - subclassing is almost never used anyway since `async` functions return the base class always anyway.

ljharb: Engines can already optimize out the `.then` calls when it's not a subclass and when `Promise.prototype.then` is unmodified; but no, if I replace `Promise.prototype.then` it needs to be able to observe the calls too.

raulsebastianmihaila: @ljharb I don't doubt what you're saying but I also can't help notice that the motivation for why those 2 options were the only ones considered by TC39 is missing. If TC39 isn't willing to evaluate (or re-evaluate, in case they already have) the aspects I've mentioned, this PR is DOA.

ljharb: The third option is what this PR does, which is what the finally proposal did originally, and what the committee rejected - see https://github.com/tc39/proposal-promise-finally/pull/34

ljharb:
Now that `finally` is stage 4, this is the right place to make the change.

However, I think #1250 is likely to be a better and more generic approach, and I suspect that if that achieves consensus, this may be able to be closed.

ljharb:
Closing, since #1250 has landed.

PR #1028, opened 2017-11-11
[Normative] Add RegExp `dotAll` mode (`s` flag)

mathiasbynens:
Proposal repo: https://github.com/tc39/proposal-regexp-dotall-flag

PR #1123, opened 2018-02-28
Normative: FlagText allows s

mysticatea:
This PR fixes an oversight.

[12.2.8.1 Static Semantics: Early Errors](https://www.ecma-international.org/ecma-262/8.0/#sec-primary-expression-regular-expression-literals-static-semantics-early-errors) disallows `s` flag currently. But it should allow `s` flag since #1028.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12143,7 +12143,7 @@
             It is a Syntax Error if BodyText of |RegularExpressionLiteral| cannot be recognized using the goal symbol |Pattern| of the ECMAScript RegExp grammar specified in <emu-xref href="#sec-patterns"></emu-xref>.
           </li>
           <li>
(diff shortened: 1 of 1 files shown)


bmeck:
+1 matches [RegExpInitialize](https://tc39.github.io/ecma262/#sec-regexpinitialize) if anyone was looking for the constructor / compile corollary.

littledan:
Please mark this PR as Normative: rather than Editorial:, as it fixes a bug and changes semantics.

mysticatea:
Thank you. I updated the commit message.

littledan:
This patch looks ready to land to me. Great work, @mysticatea ! What do you think, @bterlson ?

bterlson:
Looks good. I will also move this over to 2018 as it's just a simple bug fix :)

Issue #1095, opened 2018-02-09
Index conversion in ArrayAccumulation


GeorgNeis:
[ArrayAccumulation](https://tc39.github.io/ecma262/#sec-runtime-semantics-arrayaccumulation) computes the array index via `ToString(ToUint32(nextIndex+padding))`. This means the index may become 2^32-1. But that's not a valid array index anymore since it is the max value of the length property.

littledan:
What should the behavior be here? I imagine it would be pretty confusing if an Array actually had more than 2^32 elements and the property definitions wrapped around (rather than just overflowing as ordinary property definitions that happen to have property keys which look like numbers, and don't affect the length).

At the same time, I imagine many ECMAScript implementations run into resource limitations when processing such long literals or spreads (this seems to be the case with a version of V8 and SpiderMonkey that I tested long spreads with; maybe that's why you're filing this issue).

I wonder if we should instead throw an SyntaxError when Array literals are longer than 2^32-1, and a RangeError when spreads make an Array literal longer than that length at runtime. The change is likely to be web-compatible due to the resource limitations of these engines.

GeorgNeis:
I'm very much in favor of throwing once the index reaches 2^32-1. But for simplicity I'd prefer to always throw a RangeError. The distinction you made is not as clear as it sounds.

littledan:
OK, we already have parse-time ReferenceErrors, so I don't see the problem with a parse-time RangeError. Does anyone else? Any hesitation about restricting the length of array literals this way?

ljharb:
One thing i like about this approach is that it would necessitate a test262 test that ensures implementations *do* parse literals that are of 2^32 - 1 items (along with one that ensures it throws at 2^32).

littledan:
I don't see why it would necessitate that (any more than the current specification does). Resource limitations are outside of the scope of the specification, before and after this change. This change would just add an upper bound.

allenwb:
The only way this would happen in the real world would be by spreading a run-away iterator.  But if it does occur, what happens is fully-specified by [exotic array's 
\[\[DefineOwnProperty\]\]](https://tc39.github.io/ecma262/#sec-array-exotic-objects-defineownproperty-p-desc). 

Note that the behavior would be the same as if Array.from was used with the same iterator. (Except that Array.from imposes a MAX_SAFE_INTERGER limit, to accommodate array subclasses that don't have the 2<sup>32</sup>-1 limit).

[...]

PR #1124, opened 2018-03-02
Normative: Remove ToUint32 from array literal evaluation

littledan:
Previously, array literals (including spreads) had their indices and
lengths calculated with a ToUint32 operation. As Georg Neis noted
in #1095, such a calculation isn't quite right, as 2**32-1 is not
an array index. These semantics are unlikely to be observable in many
current JavaScript implementations due to resource limitations, but
the strange behavior may appear in future, resource-rich
implementations.

This patch specifies Georg Neis's suggestion for semantics: to write to
the length just in the case of elisions (regardless of where they are in
the Array literal), and don't bother to write to the length for writes for
ordinary elements (as Array indices will already lead to length updates).

As a bonus, the semantics for Array literals are singificantly simplified, involving
just one syntax-directed operation and pass rather than two. Note that, as before,
too-long Array literals throw only as a runtime exception, not as an early error.
This patch includes appropriate use of ? and ! in modified operations.

cc @allenwb @GeorgNeis @anba

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12637,51 +12637,57 @@
           `...` AssignmentExpression[+In, ?Yield, ?Await]
       </emu-grammar>
 
-      <emu-clause id="sec-static-semantics-elisionwidth">
-        <h1>Static Semantics: ElisionWidth</h1>
-        <emu-grammar>Elision : `,`</emu-grammar>
-        <emu-alg>
-          1. Return the numeric value 1.
-        </emu-alg>
-        <emu-grammar>Elision : Elision `,`</emu-grammar>
-        <emu-alg>
-          1. Let _preceding_ be the ElisionWidth of |Elision|.
-          1. Return _preceding_ + 1.
-        </emu-alg>
-      </emu-clause>
-
       <emu-clause id="sec-runtime-semantics-arrayaccumulation">
         <h1>Runtime Semantics: ArrayAccumulation</h1>
         <p>With parameters _array_ and _nextIndex_.</p>
+        <emu-grammar>Elision : `,`</emu-grammar>
+        <emu-alg>
+          1. Let _len_ be _nextIndex_ + 1.
+          1. Perform ? Set(_array_, `"length"`, _len_, *true*).
+          1. NOTE: The above Set throws if _len_ exceeds 2<sup>32</sup>-1.
+          1. Return _len_.
+        </emu-alg>
+        <emu-grammar>Elision : Elision `,`</emu-grammar>
+        <emu-alg>
+          1. Return the result of performing ArrayAccumulation for |Elision| with arguments _array_ and _nextIndex_ + 1.
+        </emu-alg>
         <emu-grammar>ElementList : Elision? AssignmentExpression</emu-grammar>
         <emu-alg>
-          1. Let _padding_ be the ElisionWidth of |Elision|; if |Elision| is not present, use the numeric value zero.
+          1. If |Elision| is present, then
+            1. Let _nextIndex_ be the result of performing ArrayAccumulation for |Elision| with arguments _array_ and _nextIndex_.
+            1. ReturnIfAbrupt(_nextIndex_).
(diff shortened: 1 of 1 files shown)


anba:
If ToUint32 is simply removed, the problem (unspecified behaviour for too large values when ToString is called) noted in https://github.com/tc39/ecma262/issues/1095#issuecomment-368177881 will arise:
>  If you could reach it and the bogus ToUint32 that was inserted as a "fix" is removed then the property key that is set is currently unspecified because nextIndex + 1 will compute a mathematical value that is not a Number value and ToString doesn't specify how to convert such a value to a string value.

littledan:
@anba I see; I was imagining that these were Numbers, but of course they are mathematical values. I'd be fine waiting to land this patch until after the fix to https://github.com/tc39/proposal-bigint/issues/10 lands, which will make them Numbers.

littledan:
@GeorgNeis and I discussed this change a bit further. Maybe the semantics here are a little off the mark--would it make more sense if we went with one version of @anba 's proposed semantics and used `min(length, 2**32-1)` when setting the length? This would give the final length that would result from the same series of property sets. The only exception we might want to make is if the array ends with several training commas which overflow--that could be considered equivalent to explicitly setting the length to a too-high value, and still throw. Any thoughts on these alternate semantics? If people like it, I can write this up into a new patch.

spec.html
```diff
@@ -11473,27 +11473,29 @@ <h1>Runtime Semantics: Evaluation</h1>
         <emu-alg>
           1. Let _array_ be ! ArrayCreate(0).
           1. Let _pad_ be the ElisionWidth of |Elision|; if |Elision| is not present, use the numeric value zero.
-          1. Perform Set(_array_, `"length"`, ToUint32(_pad_), *false*).
-          1. NOTE: The above Set cannot fail because of the nature of the object returned by ArrayCreate.
+          1. Perform ? Set(_array_, `"length"`, _pad_, *true*).
+          1. NOTE: The above Set may throw in the case of an Array literal's padding exceeds 2<sup>32</sup>-1 elements.
```
GeorgNeis: s/exceeds/exceeding/
or better s/may throw in the case of an Array literal's/throws if the/

littledan:
OK, @GeorgNeis had some more suggestions, which seem really on point to me. The result is a simpler and more obviously correct version, which is sightly different in semantics in even more obscure edge cases, but which I prefer. I've uploaded a commit from him to this PR, and edited the PR description to match the current proposal. If/when this PR lands, please squash the commits and use the description above as the commit message (minus the `@`mentions).

Given resource limits on all JS implementations Georg and I tested, we could not find a way to differentiate the current semantics from the semantics proposed in this patch (or the alternatives previously considered). I'd say, unfortunately, test262 tests will not be useful for this PR, and we should consider landing it without tests or implementations.

littledan:
Any thoughts on this PR? What more is needed to decide whether it's ready to merge?

ljharb:
Given that it's normative, it seems like it needs consensus (along with a rebase)

spec.html
```diff
@@ -11458,7 +11465,7 @@ <h1>Runtime Semantics: ArrayAccumulation</h1>
             1. Let _next_ be ? IteratorStep(_iteratorRecord_).
             1. If _next_ is *false*, return _nextIndex_.
             1. Let _nextValue_ be ? IteratorValue(_next_).
-            1. Let _status_ be CreateDataProperty(_array_, ToString(ToUint32(_nextIndex_)), _nextValue_).
+            1. Let _status_ be ! CreateDataProperty(_array_, ToString(_nextIndex_), _nextValue_).
```
```suggestion
            1. Let _status_ be ! CreateDataProperty(_array_, ! ToString(_nextIndex_), _nextValue_).
```

littledan:
@ljharb Feel free to land all of those suggestions. Somehow, the GitHub UI won't let me commit them.

Issue #1121, opened 2018-02-26
Allow set default value to CatchParameter


chicoxyzzy:
There is some inconsistency between [`CatchParameter`](https://tc39.github.io/ecma262/#prod-CatchParameter) and [`BindingElement `](https://tc39.github.io/ecma262/#prod-BindingElement) (i.e. function parameters).

Example from https://github.com/babel/babel/issues/7434:

```js
try {
  throw undefined;
} catch ({foo = 'bar'} = {}) {
  console.log(foo);
}
```

This is Syntax Error according to current spec grammar, but are there any reasons why assigning default value is not allowed here while destructuring (and default parameters inside of it) work fine?

chicoxyzzy:
My proposal is to change
```
CatchParameter[Yield, Await]:
  BindingIdentifier[?Yield, ?Await]
  BindingPattern[?Yield, ?Await]
```
to
```
CatchParameter[Yield, Await]:
  SingleNameBinding[?Yield, ?Await]
  BindingPattern[?Yield, ?Await]Initializer[+In, ?Yield, ?Await]opt
```

I'll be happy to file PR if this proposal sounds ok

chicoxyzzy:
To be clear, this problem was reported in Babel and it was caused by 3rd party library (regenerator in this case). Since it already possible to catch object, it looks weird that it's possible to use destructuring and default parameters inside it, but impossible to use set default value for CatchParameter.

littledan:
I can see the consistency argument for this change, but I'm wondering--when would you actually want to throw `undefined`?

chicoxyzzy:
@littledan personally I don't want to throw `undefined` or any other primitive or object :) This may happen in 3rd party libraries.

allenwb:
This was an intentional design decision. Because the expectation is there should always be some object coercible value (ie, not undefined or null) that was thrown.

If you want to explore changing something, how about exploring  to refusing to throw undefined. 

WRT changing  _CatchParameter_, I don't think we should do it.  Is somebody really wants to restructure a catch parameter and are concerned about null or undefined they can move the destructuring into the body of the catch clause.

ljharb:
can someone clarify: is it “a default value in the catch binding” that’s currently disallowed, or only in the presence of destructuring, or is it destructuring in the catch binding that’s currently disallowed?

ljharb:
[...]

PR #1126, opened 2018-03-03
Normative: Redefine CatchParameter as FormalParameter

chicoxyzzy:
This change removes default value restriction for catch parameter.

Closes https://github.com/tc39/ecma262/issues/1121

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -18287,8 +18287,7 @@
         `finally` Block[?Yield, ?Await, ?Return]
 
       CatchParameter[Yield, Await] :
-        BindingIdentifier[?Yield, ?Await]
-        BindingPattern[?Yield, ?Await]
+        FormalParameter[?Yield, ?Await]
     </emu-grammar>
     <emu-note>
       <p>The `try` statement encloses a block of code in which an exceptional condition can occur, such as a runtime error or a `throw` statement. The `catch` clause provides the exception-handling code. When a catch clause catches an exception, its |CatchParameter| is bound to that exception.</p>


allenwb:
Did you review all of the static and runtime semantics rules associated with _FormatParameter_ to make sure the will work for _CatchParameter_.  They are probably fine, but it needs to be proactively verified.

littledan:
@chicoxyzzy Thanks for preparing this patch. At a glance, it seems correct, though I'm not sure why we should bother having a _CatchParameter_ production at this point rather than just using _FormalParameter_ directly. This would require updating many more little parts of the spec, unfortunately.

Would you be interested in writing test262 tests for this change? (I think the test wouldn't change regardless of whether you take that editorial suggestion I made above.)

chicoxyzzy:
@littledan sure, no problem. I will create PR in test262 repo within the next few days.

anba:
I know Allen mentioned in #1121 to use `FormalParameter`, but I still don't see why we can't use `BindingElement` directly.

littledan:
I don't have a strong opinion about this editorial question, but in general, I don't think patches like this need to use a minimal number of lines of spec changed as a goal. It would be a little bit of work, but it'd be fine to update all the usages of CatchParameter, even though it's a bigger patch.

leobalter:
what happens when the initializer is the binding initializer?

```js
try {
  throw undefined;
} catch (x = x) {}
```

ljharb:
@leobalter i think that'd just be the same as `(x)`, and altho we could forbid that, we couldn't likely forbid `catch(x = (x))` or `catch(x = (0, x))` so it might end up being a linter thing (like `throw throw`)

leobalter:
Let me expand this a bit more:

```js
var x;
try {
  throw undefined;
} catch (x = x) {}
```

```js
var y = 42;
try {
  throw undefined;
} catch (x = y) {
  let y = 42;
}
``` 

```js
try {
  throw undefined;
} catch ([x = y, y] = x) {}
```


In functions we have the [IteratorBindingInitialization](https://tc39.github.io/ecma262/#sec-function-definitions-runtime-semantics-iteratorbindinginitialization) to solve these issues.

I believe the [catch clause evaluation](https://tc39.github.io/ecma262/#sec-runtime-semantics-catchclauseevaluation) should get some 👀 as well.

chicoxyzzy:
I'll take a look at it

anba:
> @anba I think that using FormalParameter could be guarantee of consistency

Can you clarify what you mean by "consistency"? FormalParameter is more or less only an alias for `BindingElement`, when ignoring its additional static semantics. And for the runtime semantics it seems unlikely we want to use https://tc39.github.io/ecma262/#sec-function-definitions-runtime-semantics-iteratorbindinginitialization here.

chicoxyzzy:
Sorry, I was too busy at work. I'll try to get back to this in next few weeks.

ljharb:
@chicoxyzzy have you had a chance to get to this?

chicoxyzzy:
I'm trying to make python test generation tool work, but no luck yet

chicoxyzzy:
Do I have any chance to find a champion who can present this normative change to the committee? I'm going to make another attempt at tests soon.

ljharb:
@chicoxyzzy Once tests are written I'm happy to present it to the committee on your behalf; I'll probably be able to do that in July.

ljharb:
@chicoxyzzy ping; we've still got some time, but if you can get those tests written I can add this to the July agenda :-)

chicoxyzzy:
Unfortunately I won't have time for that at least until the end of this week 😞

ljharb:
@chicoxyzzy that would work just fine! looking forward to it.

ljharb:
The committee discussed this today and could not come to consensus on merging the PR - there was skepticism about the value of this change, and concern about it being confusing. If @chicoxyzzy or anyone is still interested in this change, a full proposal would be appropriate.

I'll close this PR, but we can reopen it if this becomes a proposal that advances.

chicoxyzzy:
Yes, I'd like to create a proposal repo later.

Issue #1119, opened 2018-02-25
Atomics.wait/wake are weaker than underlying OS calls


conrad-watt:
Actions associated with Atomics.wake/wait are not tracked in the axiomatic memory model, other than the initial read by wake. This means that other reads and writes are permitted to be re-ordered around these instructions in surprising ways.

In particular, with the current semantics it is **_not_**  guaranteed that thread 1 will read 1 in the second line (assuming tA is initially zeroed).

| Line |Thread 1                      | Thread 2 |
|--|------------------------------|-------------- |
| 1 |   Atomics.wait(tA, 0, 0) |  tA[1] = 1   |
| 2 |    var x = tA[1];     |  Atomics.store(tA, 0, 1) |
| 3 |      -          | Atomics.wake(tA, 0, 1) |

This is because if thread 1 waits before thread 2 reaches line 2, no ordering constraint is induced between them by either thread 2's atomic store (which is never read from) or thread 2's wake, which does not induce any abstract action in the axiomatic model. Therefore, thread 1's non-atomic read is free to take the initial value of tA[1].

This is far weaker than the underlying OS calls that I assume wait/wake must be implemented using. On linux at least, when thread 1 sleeps, it will issue a full barrier. If thread 2 successfully wakes 1 or more threads, it will issue a write barrier (https://www.kernel.org/doc/Documentation/memory-barriers.txt). In my example, this is sufficient to guarantee that thread 1 will read 1 in the second line.

It also seems intuitive that wake should be used to indicate work completed on the part of the waking thread, but with the current semantics it is not guaranteed that this work will be visible.

I've previously discussed this over email with @lars-t-hansen, @syg, and @rossberg, and there seems to be a broad consensus that the semantics should be stronger. I personally believe that the act of waking and being woken should introduce abstract actions into the [[EventList]] of each thread such that the wake of thread 2 synchronizes-with a "being woken" action in thread 1. Indeed, based on our discussions, this may have been the informal intention of the spec during an earlier draft. However this might require additional barriers in some weaker architectures, and discussions in other possible specification solutions are still ongoing.

syg:
@conrad-watt Would you like to draft a normative PR for adding the s-w additions you outlined? I will review and champion if needed.

conrad-watt:
Ok! I'll put something together.

conrad-watt:
fixed by https://github.com/tc39/ecma262/commit/ef0cf19e78b227c768751b13d877fa27bfb38266

PR #1127, opened 2018-03-05
Normative: Strengthen Atomics.wait/notify synchronization to the level of other Atomics operations

conrad-watt:
See issue #1119 for background on the current behaviour.

This proposal strengthens the synchronization between a waking thread and its wakees as though they were each a coinciding atomic write/read (respectively) to the same memory location, in line with some previous comments by @lars-t-hansen on the informal language of an earlier draft of the memory model proposal.

Since each waker/wakee pair is determined purely operationally, there is no need to struggle to reassociate the events from scratch in the axiomatic semantics (a la the ReadsBytesFrom relation for Shared Data Block events). Instead, each agent tracks which waking events it is responsible for in a new [[AgentSynchronizesWith]] field of its Events Record, which is then unified with the global synchronizes-with relation.

I chose to introduce a rather general new "Synchronize" action rather than abuse any of the existing ones, since this spec machinery would be useful to specify a future fork or similar thread spawning operation, or explicit barrier operations, either in ECMAScript or in WebAssembly. This also limits the potential for unfortunate interactions with other parts of the spec which deal with Shared Data Block events.

I also rename the [[EventLists]] property of candidate executions to [[EventsRecords]] since Events Records now contain more than just EventLists, with the added bonus that it makes sentences which formerly read as "Let eventList be the [[EventList]] field of the element in execution.[[EventLists]] whose [[AgentSignifier]] is AgentSignifier()." easier to parse, considering the element of [[EventLists]] in question is actually an Events Record.

**Some initial discussion points**
1 -
Is "Synchronize" the right name for the new action? I mulled over "Scheduling" or something similar, but as I mentioned earlier these actions could potentially also be used to specify barriers in the future.

2 -
This change should not invalidate any user code - it is a strict subsetting of the previously allowed observable behaviour.

3 -
A scattering of informative notes throughout the PR may still be required.

4 -
**On second thoughts I don't believe this is true**
~~This new specification is still slightly weak. It doesn't incorporate all of the guarantees you might get from an OS barrier in Atomics.wait. For example, an execution for the following program is still allowed where x = 1, y = 1, z = 0.~~

| Line |Thread 1                      | Thread 2 |
|--|------------------------------|-------------- |
| 1 |   tA[2] = 1 | Atomics.store(tA, 0, 1)   |
| 2 |    tA[1] = 1 |  var x = Atomics.wake(tA, 0, 1)    |
| 3 |    Atomics.wait(tA, 0, 0)  | var y = tA[1] |
| 4 |      -                               | var z = tA[2] |

~~In a real implementation using OS calls for wake/wait, if x = 1 I'd expect to only observe y = z = 1. However this certainly isn't a property I'd expect as many people to care about or try to rely on. In particular, the execution where x = 0, y = 1, z = 0 must be allowed by the model, so it seems that the only programs that could benefit from a further strengthened model would be rather pathological.~~

5 -
On x86, this specification is satisfied by implementing Atomics.wake as a regular ECMAScript atomic load followed by a bare Linux wait, and Atomics.wake as a bare Linux wake, but (I believe) it is stronger than the same on ARM/Power, which will necessitate an extra read barrier at least (see below). I'm also not familiar enough with the barrier behaviour of other OS's to know what extra barriers are required, although I would be surprised if they do not at least issue a write barrier upon waking, as Linux does. It should be noted that this is not new - ECMAScript atomic writes are stronger than the (relaxed) atomic writes of ARM/Power, and this is reasonably well-known.

6 -
**Implementation schemes** - a "defensive" implementation scheme to ensure conformance after this change, assuming nothing about the underlying architecture/OS, is simply to do a full barrier immediately before any call to Atomics.wake and immediately after any call to Atomics.wait - this will always produce the correct behaviour, and is analogous to the "naive code generation scheme" suggested in the ECMAScript standard for atomic loads/stores.

An implementation wishing to do better than this (although I hope there are no benchmarks that depend on a very fast Atomics.wake!) must make use of knowledge about its particular architecture/OS combination. Below is my to-the-best-of-my-knowledge opinion on possible improvements.

**Atomics.wait**
On x86, a read barrier after Atomics.wait is all that is required for correctness. On ARM/Power, a full barrier is required.

On Linux a full barrier is carried out by the OS call to sleep, so no additional barrier is required by the ECMAScript implementation no matter the architecture. I don't know the behaviour of other OS's off the top of my head, but I imagine similar guarantees could be determined about their barrier behaviour.

**Atomics.wake**
No barrier at all is required at Atomics.wake if the call to Atomics.wake does not wake any waiting threads, so the cases afterwards will assume it wakes at least one.

On x86, a write barrier is sufficient for correctness, and therefore on an OS where the underlying syscall already performs this (Linux, at least, when at least one thread is woken) no additional barrier is required in the ECMAScript implementation.

On ARM/Power, a full barrier is required overall, but if the OS performs a write barrier already, the ECMAScript implementation need only perform an additional read barrier.

@lars-t-hansen, @syg, and @rossberg PTAL.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6714,7 +6714,7 @@
       <p>Examples of that type of termination are: operating systems or users terminating agents that are running in separate processes; the embedding itself terminating an agent that is running in-process with the other agents when per-agent resource accounting indicates that the agent is runaway.</p>
     </emu-note>
 
-    <p>Prior to any evaluation of any ECMAScript code by any agent in a cluster, the [[CandidateExecution]] field of the Agent Record for all agents in the cluster is set to the initial candidate execution. The initial candidate execution is an empty candidate execution whose [[EventsRecords]] field is a List containing, for each agent, an Agent Events Record whose [[AgentSignifier]] field is that agent's signifier and whose [[EventList]] field is an empty List.</p>
+    <p>Prior to any evaluation of any ECMAScript code by any agent in a cluster, the [[CandidateExecution]] field of the Agent Record for all agents in the cluster is set to the initial candidate execution. The initial candidate execution is an empty candidate execution whose [[EventsRecords]] field is a List containing, for each agent, an Agent Events Record whose [[AgentSignifier]] field is that agent's signifier, and whose [[EventList]] and [[AgentSynchronizesWith]] fields are empty Lists.</p>
 
     <emu-note>
       <p>All agents in an agent cluster share the same candidate execution in its Agent Record's [[CandidateExecution]] field. The candidate execution is a specification mechanism used by the memory model.</p>
@@ -35537,6 +35537,15 @@ THH:mm:ss.sss
         <emu-alg>
           1. Assert: The calling agent is in the critical section for _WL_.
           1. Assert: _W_ is on the list of waiters in _WL_.
+          1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
+          1. Let _eventsRecord_ be the Agent Events Record in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
+          1. Let _agentSynchronizesWith_ be _eventsRecord_.[[AgentSynchronizesWith]].
+          1. Let _notifierEventList_ be _eventsRecord_.[[EventList]].
+          1. Let _waiterEventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is _W_.
+          1. Let _notifyEvent_ and _waitEvent_ be new Synchronize events.
+          1. Append _notifyEvent_ to _notifierEventList_.
+          1. Append _waitEvent_ to _waiterEventList_.
+          1. Append (_notifierEvent_, _waitEvent_) to _agentSynchronizesWith_.
           1. Notify the agent _W_.
         </emu-alg>
         <emu-note>
@@ -38617,12 +38626,13 @@ The abstract operation PerformPromiseThen performs the &ldquo;then&rdquo; operat
     </emu-table>
 
     <p>These events are introduced by abstract operations or by methods on the Atomics object.</p>
-    <p>In addition to Shared Data Block events, there are host-specific events.</p>
+    <p>Some operations may also introduce <dfn>Synchronize</dfn> events. A <dfn>Synchronize event</dfn> has no fields, and exists purely to directly constrain the permitted orderings of other events.</p>
+    <p>In addition to Shared Data Block and Synchronize events, there are host-specific events.</p>
     <p>Let the range of a ReadSharedMemory, WriteSharedMemory, or ReadModifyWriteSharedMemory event be the Set of contiguous integers from its [[ByteIndex]] to [[ByteIndex]] + [[ElementSize]] - 1. Two events' ranges are equal when the events have the same [[Block]], and the ranges are element-wise equal. Two events' ranges are overlapping when the events have the same [[Block]], the ranges are not equal and their intersection is non-empty. Two events' ranges are disjoint when the events do not have the same [[Block]] or their ranges are neither equal nor overlapping.</p>
     <emu-note>
       <p>Examples of host-specific synchronizing events that should be accounted for are: sending a SharedArrayBuffer from one agent to another (e.g., by `postMessage` in a browser), starting and stopping agents, and communicating within the agent cluster via channels other than shared memory. It is assumed those events are appended to agent-order during evaluation like the other SharedArrayBuffer events.</p>
     </emu-note>
-    <p>Shared Data Block events are ordered within candidate executions by the relations defined below.</p>
+    <p>Events are ordered within candidate executions by the relations defined below.</p>
   </emu-clause>
 
   <emu-clause id="sec-agent-event-records">
@@ -38645,6 +38655,11 @@ The abstract operation PerformPromiseThen performs the &ldquo;then&rdquo; operat
             <td>[[EventList]]</td>
             <td>A List of events</td>
             <td>Events are appended to the list during evaluation.</td>
+          </tr>
+          <tr>
+            <td>[[AgentSynchronizesWith]]</td>
+            <td>A List of pairs of Synchronize events</td>
+            <td>Synchronize relationships introduced by the operational semantics.</td>
           </tr>
         </tbody>
       </table>
@@ -38760,6 +38775,17 @@ The abstract operation PerformPromiseThen performs the &ldquo;then&rdquo; operat
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-synchronizeeventset" aoid="SynchronizeEventSet">
+      <h1>SynchronizeEventSet( _execution_ )</h1>
+      <p>The abstract operation SynchronizeEventSet takes one argument, a candidate execution _execution_. It performs the following steps:</p>
+      <emu-alg>
+        1. Let _events_ be an empty Set.
+        1. For each event _E_ in EventSet(_execution_), do
+          1. If _E_ is a Synchronize event, add _E_ to _events_.
+        1. Return _events_.
+      </emu-alg>
+    </emu-clause>
+
     <emu-clause id="sec-hosteventset" aoid="HostEventSet">
       <h1>HostEventSet ( _execution_ )</h1>
       <p>The abstract operation HostEventSet takes one argument, a candidate execution _execution_. It performs the following steps:</p>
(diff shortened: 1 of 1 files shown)


littledan:
Why is this marked needs consensus? My understanding is that, although this is a normative change, it really reflects how any reasonable implementation would do things; it's more like a bug fix.

conrad-watt:
I'm not in a position to judge whether this needs consensus, but I briefly discuss in the **implementation schemes** section why this proposal might not match existing implementations, depending on how defensively they're currently synchronizing.

In short, if implementations have been treating wake/wait as like atomic loads/stores (in the absence of formal spec indication), they should already be conformant to this change. But implementations delegating to the OS directly without any barriers of their own might become non-conformant on weaker architectures.

ljharb:
My understanding is that any normative change needs consensus, unless in the editor's judgement it doesn't need consensus. I lack the context on Atomics to know one way or the other; I'm hoping @bterlson will have an opinion.

syg:
I suppose it's arguable whether it is a bug fix.

The strengthened behavior in the PR disallows some compiler transforms that the spec technically allows, e.g., moving non-atomic stores across `Atomic.wake` calls. (I say compiler because I don't think there are any OSes that would make such a thing observable.) Those transforms cause astonishment, and the broad consensus @conrad-watt referred to is that in e-mail correspondence, we agreed among us that those transforms were probably intended to be disallowed from the beginning, but slipped through the cracks.

I am comfortable treating this as a bug fix. But to fully cover our bases, we could discuss very quickly in March with other interested parties (Waldemar). I'll defer to editor group.

I'll make time to review this before next week.

conrad-watt:
If it's appropriate, I could also attend any discussion in person. I'm in London giving a presentation at the Formal Methods Meets JavaScript meeting the day before TC39 anyway.

I don't know what the rules are for external contributors attending TC39 meetings.

conrad-watt:
I should also point out that the current semantics of the spec are far more troubling than I've perhaps motivated in my earlier examples. In #1119 I gave an example which could look like an over-eager compiler transformation of non-atomics, but for the following program, an execution where thread 1 reads 0 at line 2 (assuming tA initially zeroed) is allowed, which seems completely incoherent.

| Line |Thread 1                      | Thread 2 |
|--|------------------------------|-------------- |
| 1 |   Atomics.wait(tA, 0, 0) |  Atomics.store(tA, 0, 1)   |
| 2 |    var x = Atomics.load(tA, 0)      |  Atomics.wake(tA, 0, 1)  |

This certainly won't be observable in practice. This all stems from the lack of happens-before between wake and wake.

lars-t-hansen:
Going back to a fairly random earlier draft, eg, https://github.com/tc39/ecmascript_sharedmem/blob/7c10268005c7ad11552aa333dc06229f29fd6c54/tc39/spec.html (mid 2016), we find that there are definitions of synchronizes-with and happens-before that include the wake -> wait edge in the synchronization order as a special case.

That draft also has a provision for "... an embedding-specific synchronizing event that has a sender and a receiver (such as sending a SharedArrayBuffer from one agent to another), the sending action synchronizes-with the receiving action".  That seems reflected in current prose in the `host-synchronizes-with` relation.

So it would appear that the wake -> wait edge was truly left out by accident, not being a memory operation and not being a host operation.  That is, this is simply a bug.  Which of course should not dissuade anyone from discussing it.

littledan:
@conrad-watt To attend TC39 as a subject matter expert and present on a proposal, one prerequisite is to sign [this form](https://tc39.github.io/agreements/contributor/) to license any IPR related to your presentation for use in the ECMAScript specification. We've historically had TC39 members invite experts to participate; not sure if any more formal signoff would be needed here, cc @ecmageneva @RexJaeschke.

It looks like you're associated with Cambridge University--Ecma [offers](http://www.ecma-international.org/memento/members.htm) free memberships to academic institutions which could make it easier to participate in the future.

spec.html
```diff
@@ -39692,13 +39701,32 @@ <h1>Memory Model Fundamentals</h1>
...
+      <table>
+        <tbody>
+          <tr>
+            <th>Field Name</th>
+            <th>Value</th>
+            <th>Meaning</th>
+          </tr>
+          <tr>
+            <td>[[Order]]</td>
+            <td>`"SeqCst"`</td>
```
syg: Is this intended future-proofing for release-acquire?

The reason RMW events have a fixed [[Order]] field is because they're quantified over along with other Shared Data Blocks that refer to [[Order]] generically. Synchronize events don't participate in that stuff, so we could leave it out for simplicity.

spec.html
```diff
@@ -39885,6 +39929,18 @@ <h1>ValueOfReadEvent( _execution_, _R_ )</h1>
         1. Return ComposeWriteEventBytes(_execution_, _R_.[[ByteIndex]], _Ws_).
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-agentssynchronizewith" aoid="AgentsSynchronizeWith">
+      <h1>AgentsSynchronizeWith( _execution_ )</h1>
```
syg: Having both Agent*s*SynchronizeWith and AgentSynchronize*s*With is confusing to me.

In conjunction with my above comment about making [[AgentSynchronizesWith]] a List, I recommend inline the logic in this AO directly into the synchronizes-with section, should be a 1-liner.

spec.html
```diff
@@ -39722,6 +39750,11 @@ <h1>Agent Events Records</h1>
             <td>A List of events</td>
             <td>Events are appended to the list during evaluation.</td>
           </tr>
+          <tr>
+            <td>[[AgentSynchronizesWith]]</td>
+            <td>A Relation on Synchronize events</td>
```
syg: Please change this to a List.

The framework I've tried to adhere to is that the operational semantics appends events directly to Lists, and Relations are built by the axiomatic semantics from those Lists. (Also the Relations are not constructed by AOs to avoid implying that the axiomatic semantics is a step-by-step thing.)

spec.html
conrad-watt: Partially this, and also to typographically give Synchronize events their own table structure on par with SDB events. I've changed to a textual description, which doubles as an explanatory note.

spec.html
conrad-watt: done

spec.html
conrad-watt: Done - I may submit a separate editorial PR to make the spec's use of the word "pair" slightly more consistent.

conrad-watt:
rebased + comments addressed, PTAL

conrad-watt:
I've been told that if I want to attend any discussion on this at next week's meeting (pending sign-off from the ECMA SG), I should note time constraints/preferences here. I would prefer the 20th, although on that day I would have to leave after 2pm. Failing that I'm available any other day for the duration.

I'm assuming this still needs to pass through the committee as a formality, in the absence of indications to the contrary.

Longer term I will look at convincing my department/university to affiliate with TC39 officially.

@IgnoredAmbience

spec.html
```diff
@@ -40009,7 +39979,7 @@ <h1>synchronizes-with</h1>
           </ul>
         </li>
         <li>For each pair (_E_, _D_) in _execution_.[[HostSynchronizesWith]], (_E_, _D_) is in _execution_.[[SynchronizesWith]].</li>
-        <li>For each pair of Synchronize events (_S_, _Sw_) in AgentsSynchronizeWith(_execution_), (_S_, _Sw_) is in _execution_.[[SynchronizesWith]].</li>
+        <li>For each element _eventsRecord_ of _execution_.[[EventsRecords]], for each pair (_S_, _Sw_) in _eventsRecord_.[[AgentSynchronizesWith]], (_S_, _Sw_) is in _execution_.[[SynchronizesWith]].</li>
```
syg: Editorial nit: I think it'd be a clearer to have the [[AgentSynchronizesWith]] addition be its own `<li>`

conrad-watt: done + rebased

nomadtechie:
Hi @conrad-watt @syg  - I've been working on adding test coverage to Atomics.wait and Atomics.wake. What implications (if any) does this change have on test coverage?

ajklein:
@binji @dtig @flagxor heads-up that this change reached consensus at TC39, want to make sure V8 folks have seen it.

binji:
> To what extent is test262 the right venue for these kind of concurrent "litmus" tests?

We discussed this some [here](https://github.com/tc39/test262/issues/1147), but the issue seems to have stalled. One issue is that the tool generated many litmus tests ("The litmus tests cover all possible memory interactions of sizes 1-4. They take up 202MB compressed and about 1.1GB uncompressed."). Another issue, as you mention, is that they are inherently probabilistic.

It seemed as though there was value to adding them to test262, though, perahps as an addendum.

littledan:
Thanks for the analysis of tests, @binji and @conrad-watt. Seems to me like this fix is ready to land without blocking on tests.

erights:
I think I do not need to review this. Can I painlessly withdraw as a reviewer of this?

bmeck:
@erights yup, just wanted to make sure since I was curious about this after your spectre talk and saw @syg 's [comment about ordering](https://github.com/tc39/ecma262/pull/1127#issuecomment-370537578). Was more curious if you agreed on that issue in particular. However, if nothing seems controversial we should land this at this point.

ljharb:
@erights i believe @bmeck had some thoughts that perhaps this change might not just be a "nice to have" but also a "must have", to achieve some security properties, and wanted your thoughts on that. You can certainly choose to decline the invitation :-D

ljharb:
@conrad-watt would you mind rebasing this on the latest master?

conrad-watt:
done

erights:
It does look like this change goes strictly in the right direction --- towards a smaller space of possible outcomes, where the remaining outcomes are more intuitive as the ones ruled out. So on these grounds I am in favor. But I have not dug into this enough to consider my positive reaction a review; nor do I expect to. Thanks.

spec.html
```diff
@@ -35537,6 +35537,15 @@ <h1>NotifyWaiter( _WL_, _W_ )</h1>
         <emu-alg>
           1. Assert: The calling agent is in the critical section for _WL_.
           1. Assert: _W_ is on the list of waiters in _WL_.
+          1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
+          1. Let _eventsRecord_ be the Agent Events Record in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
+          1. Let _agentSynchronizesWith_ be _eventsRecord_.[[AgentSynchronizesWith]].
+          1. Let _wakerEventList_ be _eventsRecord_.[[EventList]].
```
ljharb: `notifierEventList`?

ljharb:
(linking to #1220)

spec.html
conrad-watt: done

PR #1131, opened 2018-03-06
Restore original definition of DataView.length

### LINKED ISSUE #787: The length of DataView constructor doesn't match between the spec and implementations
https://tc39.github.io/ecma262/#sec-dataview-buffer-byteoffset-bytelength
> 24.2.2.1 DataView ( _buffer_ [ , _byteOffset_ [ , _byteLength_ ] ] )

https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects
> Unless otherwise specified, this value is equal to the largest number of
> named arguments shown in the subclause headings for the function description.
> Optional parameters (which are indicated with brackets: [ ]) or rest
> parameters (which are shown using the form «...name») are not included in the
> default argument count.

There is no explicit note for DataView constructor's length, so it should follow the above, and `DataView.length` should be 1.

Many implementations, however, return 3 for `DataView.length`.
  * Firefox Nightly 54.0a1 (2017-01-27) (64-bit)
  * Google Chrome Canary 58.0.2995.0 (Official Build) canary (64 bit)
  * WebKit 10.0.2 (12602.3.12.0.1, r211319)
(haven't checked on Edge since I don't have environment to test trunk)

anba:
Chakra, JSC, V8, and SpiderMonkey all agree that `DataView.length` is 3. test262 also tests that `DataView.length` is 3.

Fixes #787

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -36211,7 +36211,7 @@ THH:mm:ss.sss
 
       <!-- es6num="24.2.2.1" -->
       <emu-clause id="sec-dataview-buffer-byteoffset-bytelength">
-        <h1>DataView ( _buffer_ [ , _byteOffset_ [ , _byteLength_  ] ] )</h1>
+        <h1>DataView ( _buffer_, _byteOffset_, _byteLength_ )</h1>
         <p>When the `DataView` is called with at least one argument _buffer_, the following steps are taken:</p>
         <emu-alg>
           1. If NewTarget is *undefined*, throw a *TypeError* exception.
@@ -36221,7 +36221,7 @@ THH:mm:ss.sss
           1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
           1. Let _bufferByteLength_ be _buffer_.[[ArrayBufferByteLength]].
           1. If _offset_ &gt; _bufferByteLength_, throw a *RangeError* exception.
-          1. If _byteLength_ is either not present or *undefined*, then
+          1. If _byteLength_ is *undefined*, then
             1. Let _viewByteLength_ be _bufferByteLength_ - _offset_.
           1. Else,
             1. Let _viewByteLength_ be ? ToIndex(_byteLength_).


leobalter:
This is legacy pre dating ES6, from Khronos I guess.

`length` was [explicitly set to `3`](https://www.ecma-international.org/ecma-262/6.0/#sec-properties-of-the-dataview-constructor), I believe to match the pre existing constructor. It was removed after ES6. 

Even though, I would recover this text explicitly setting the `length` property rather than removing the optional parameters as they are intended to work as optional. The property can be set for legacy.

At this point I'm way more motivated for a breaking change for resetting the length properties of DataView, ArrayBuffer and TypedArray to match their simple - non optional - parameters.

leobalter:
We would not even make the arguments required but flag them as such, they would still behave as optional. Fixing the respective length is good for consistency through the spec.

leobalter:
I'm bringing this to the agenda as I want to make sure we gather some feedback here, what I'm proposing is a breaking change, and it's slightly different than this PR contents, so it's a normative change anyway.

I should follow up with a table of values for length including TypedArrays, DataView and ArrayBuffer constructors.

@anba are you interested to update - or create a new - this PR to reflect this change on the length?

The modifications would be:

[Here](https://tc39.github.io/ecma262/#sec-typedarray-constructors), remove the line saying:

```
The length property of the TypedArray constructor function is 3.
```

[Here](https://tc39.github.io/ecma262/#sec-arraybuffer-length), flag the optional parameter for the ArrayBuffer constructor.

```
24.1.2.1 ArrayBuffer ( [ length ] )
```

The DataView constructor wouldn't need any modifications.

leobalter:
Current:

| | ch | v8 | jsshell | jsc | es6 | latest spec draft |
|-----------|---|---|---|---|---|---|
| Int8Array.length | 3 | 3 | 3 | 3 | 3 | 3 |
| ArrayBuffer.length | 1 | 1 | 1 | 1 | 1 | 1 |
| DataView.length | 1 | 3 | 3 | 3 | 3 | 1`*` |

`*`- using default `length` value

Proposal 1, reflecting consistency with the specs:

| | ch | v8 | jsshell | jsc |
|-----------|---|---|---|---|
| Int8Array.length | 0 | 0 | 0 | 0 |
| ArrayBuffer.length | 0 | 0 | 0 | 0 |
| DataView.length | 1 | 1 | 1 | 1 |

Proposal 2, reflecting legacy/web reality (fallback)

| | ch | v8 | jsshell | jsc |
|-----------|---|---|---|---|
| Int8Array.length | 3 | 3 | 3 | 3 |
| ArrayBuffer.length | 1 | 1 | 1 | 1 |
| DataView.length | 3 | 3 | 3 | 3 |

anba:
DataView.length is also 3 in ChakraCore: https://github.com/Microsoft/ChakraCore/blame/9189a4bbbb4345be3ef53b70581ee5b3c131eec3/lib/Runtime/Library/JavascriptLibrary.cpp#L1892
It was changed six month ago from 1 to 3: https://github.com/Microsoft/ChakraCore/commit/f1daa9b8cc343f560bbf5bae56a2340019ba58fb

anba:
Changing `ArrayBuffer.length` from 1 to 0 makes no sense to me. While it is possible to call `new ArrayBuffer()` to create a zero length ArrayBuffer, this is definitely not how the ArrayBuffer constructor is normally used. If we change the "length" property of built-ins to reflect the minimum necessary number of arguments, methods like `Map.prototype.set` should have a length of 0, too. But just because `map.set()` is equivalent to `map.set(undefined, undefined)`, it seems more than questionable to me to change the definition of `Map.prototype.set` to `Map.prototype.set ( [ key [ , value ] ] )`

ljharb:
The length isn't for what's common, it's for what's required - setting aside legacy concerns, anything that you can safely omit is generally something that shouldn't be counted in the length.

That omitting an argument becomes "undefined" is being overly pedantic about something that's just part of the nature of JS; Map set requires two arguments, so its length should be 2 (which it is).

anba:
Oh, great. We had this exact conversation some time ago, with the exact some differences over what `length` means or does not mean. I'm almost tempted to close this PR, so maybe we can try in another year or so to change the spec to reflect what's universally implemented and tested for DataView.length.... :expressionless:

anba:
And the ArrayBuffer case is exactly the same as Map.prototype.set. The ArrayBuffer constructor has no special behaviour when its parameter is absent or undefined.

leobalter:
@anba I'm willing to take your opinion in consideration and take it as a valid rejection for my first proposal. Although, I'd like to know what you consider as the best option here for these values?

ljharb:
@anba the spec should match web reality when needed, and certainly the web reality here is that this function's length is 3.

*Conceptually*, this function's length should be 1 - so either way, I agree with leo's comment that the signature should indicate a length of 1, with a note overriding it to 3 if needed.

The real question is, would changing it to 1 everywhere break things? If it wouldn't break any code, then there's no reason that all 4 browsers couldn't just flip it to 1.

anba:
> Although, I'd like to know what you consider as the best option here for these values?

IMO, `ArrayBuffer.length` should definitely be 1, because, except for certain special cases, ArrayBuffers are always created with a non-zero length.
As mentioned [last year](https://github.com/tc39/ecma262/issues/787#issuecomment-277688972), aligning `DataView.length` with `<TypeName>TypedArray.length` seems useful to me (so either 1 or 3 for these constructors). But that's just my personal opinion. More important to me is finally aligning the spec, the implementations, and test262. And because there was no change in the implementations and test262 to follow the definitions from spec, instead we even saw ChakraCore changing `DataView.length` from 1 to 3, the spec should simply match the web reality and what's tested.

ljharb:
Except that we should match web reality when it's required, not just for the sake of matching it. If web reality is subpar and *can* be changed, we should change it.

anba:
But should the spec be allowed to change definitions between different editions, ignoring what's already is implemented?

ljharb:
Of course? The spec represents the way things "should" be, modulo places where web compatibility forces it to deviate.

anba:
For example I'd personally like to change the `If regexp is neither undefined nor null, then` test in various String.prototype methods to `If Type(regexp) is Object, then`, because I feel the spec/web reality is subpar and means we need to perform questionable checks like [this one](https://searchfox.org/mozilla-central/rev/efce4ceddfbe5fe4e2d74f1aed93894bada9b273/js/src/builtin/String.js#7-13,31) in SpiderMonkey. But is that sufficient enough to warrant a spec change?

ljharb:
If it's web compatible (that one would not be), then absolutely it is.

anba:
Why shouldn't it be web-compatible to change the test to `If Type(regexp) is Object, then`? I'd be really surprised if any application defines a `String.prototype[@@match]` method.

ljharb:
I think we're off-topic. In general, if you want to make *any change* that is a) web compatible, b) engines are willing to implement, and c) improves the quality of the spec, regardless of web reality, then it's definitely something that can be changed.

In this case, I think that DataView.length *should* be 1. That all the browsers currently have it as 3 is irrelevant; what's relevant is if it's web compatible to make it be 1. If so, browsers should change it. If not, the spec should use the established convention for noting that a function's length deviates from convention, which is a separate note that contradicts what the function signature would otherwise imply.

anba:
There has been no changes in implementations **and** in test262 during the course of a complete year to reflect what's now (= starting with ES2017) spec'ed for DataView.length. So I have to assume that there is absolutely no interest to implement this spec change. We even saw ChakraCore to change their implementation to reflect web reality. Doesn't that mean the "engines are willing to implement" requirement isn't fulfilled?

ljharb:
Did they change it to reflect web reality, or to reflect test262?

What we clearly forgot to do is update test262 to match the spec. We should try that for a year before claiming that it shows an unwillingness to implement.

leobalter:
> More important to me is finally aligning the spec, the implementations, and test262.

That's my goal.

Reading #787 we could follow up this discussion to propose DataView.length, TypedArrays.length, and ArrayBuffer.length to 1. I'm assuming this would be aligned with @anba thoughts.

This would be a less intrusive change, treating empty instances of ArrayBuffer and TypedArrays as not really useful, which I somehow agree.

WRT ChakraCore, I have the latest release (1.8.1) returning 1 for DataView.length:

```bash
> ch -v
ch version 1.8.1.0

> eshost -e 'DataView.length'
#### ch
1

#### d8
3

#### jsshell
3

#### node
3

#### jsc
3
```

anba:
> Did they change it to reflect web reality, or to reflect test262?

Reading https://github.com/Microsoft/ChakraCore/pull/3606, probably because of test262.

> What we clearly forgot to do is update test262 to match the spec. We should try that for a year before claiming that it shows an unwillingness to implement.

I was just about to write about that problem, but I was still pondering how to phrase it, because it would mean I'm going to contradict, at least partially, my previous comment. :laughing:

leobalter:
On Test262: I wish I could have the available time to reflect all of these changes in there. In any case, I'll open some PR to match what we have today, I can change it to reflect the outcome of this thread.

anba:
> WRT ChakraCore, I have the latest release (1.8.1) returning 1 for DataView.length:

1.9 and 1.10 apparently will have `DataView.length === 3`. IIRC the release schedule is tied to some Windows updates, so it can take a while until changes are present in released versions.

spec.html
```diff
@@ -36211,7 +36211,7 @@ <h1>The DataView Constructor</h1>
 
       <!-- es6num="24.2.2.1" -->
       <emu-clause id="sec-dataview-buffer-byteoffset-bytelength">
-        <h1>DataView ( _buffer_ [ , _byteOffset_ [ , _byteLength_  ] ] )</h1>
+        <h1>DataView ( _buffer_, _byteOffset_, _byteLength_ )</h1>
```
ljharb: Whether we decide on 1 or 3, this signature shouldn't be changed; if we go with 3, then like other legacy/deviating lengths, let's use a note.

leobalter:
TC39 discussed this in the meeting today - I was not present - and the outcome was to leave the specs as they are today. So DataView.length is 1. 

Test262 is already fixed to match the specs here: https://github.com/tc39/test262/commit/ccab9a18b0c5f3afa94237d9b89a65c4574384d4#diff-36b234ab13c21bd5452b4c9718810dfdR28

ljharb:
The `byteLength` change in this PR seems like it should still go in, however, and had no controversy I'm aware of?

leobalter:
the consensus was "leave the specs as it is", if we want to pursue anything further we need to bring this back to the meeting. Once again, this discussion happened while I was not present, despite my name was on the agenda and I shared with some other people I would be eventually online. Feel like time restraints can't be fully observed.

ljharb:
Then given the committee consensus, it seems that this PR and the linked issue should be closed?

littledan:
@ljharb Agreed.

littledan:
I thought I mentioned that the original length was 3 and it was deleted by mistake. Sorry, I didn't get around to editing those notes for completeness. Some things were lost.

anba:
@littledan So now that TC39 decided that `DataView.length` should be 1, the next step would be to consider aligning the length of the TypedArray constructors (from 3 to 1) for consistency with DataView. Was that point discussed at the meeting?

A possible justification: Most TypedArray constructor variants only take a single argument and the one which takes three arguments has the same signature as the DataView constructor, so it doesn't seem unreasonable to me to apply the same reasoning and change the length from 3 to 1. (There's also a TypedArray constructor variant which takes zero arguments, but that one is more or less only a special case for 22.2.4.2.)

littledan:
Yes, we discussed the proposed alignment that @leobalter mentioned above. The feedback from @bterlson and @ajklein was a bit skeptical of the value of spending time and effort changing the length of functions; since we weren't in agreement about the three possibilities, we ended up deciding to leave things how they are and not pursue this alignment.

Issue #917, opened 2017-05-14
await in parameters


pemrouz:
Following on from [this thread](https://twitter.com/pemrouz/status/831157518302867456), just creating this issue to help nudge discussion forward. Based on various comments, it seems doable but requires spec changes first. 

Clearly the following lines need to be removed, but how exactly the parsing logic needs to change to accommodate this is less clear to me:

> 26. If kind is "generator", then
>  If parameters Contains YieldExpression is true, throw a SyntaxError exception.
>
> 27. If kind is "async", then
>  If parameters Contains AwaitExpression is true, throw a SyntaxError exception.

Potential syntax variations for discussion:

* await just default value

  ```js
  async function foo(bar = await baz){  }
  ```

* await param, not just default value:

  ```js
  async function foo(await bar){  }
  ```

* await param, with default value

  ```js
  async function foo(await bar = baz){  }
  ```

ljharb:
While I like this idea, see https://github.com/tc39/ecma262#contributing-new-proposals - I think es-discuss might be a better place to suggest this.

Jamesernator:
The funny thing is in V8 if you add the flag --harmony_do_expressions `await` already works in that position if wrapped in a do expression:

```js
// Do expressions just for the same of demonstrating that the code
// needed for `await` in default parameter is already implemented
// in V8 at least
async function foo(bar=do { await 3 }) {
    return bar
}

foo() // Promise { 3 }
```

So `yield/await` should probably work in the default argument position, the fact it doesn't seems like a spec bug/oversight to me, but that's up for implementors to decide if it needs a proposal or not.

The additional syntax for `async function foo(await bar)` as much as I've always wanted that exact syntax, is probably better suited to a whole new proposal.

caitp:
> Just noticed it's explicitly excluded so it's not a spec bug. Is the current restriction better for optimizations or something?

We need parameter bindings to be heap-allocated rather than on-stack, so that they can be accessed when the generator is resumed.

Imagine:

```js
async function f(id, account = await accounts.lookup(id), transaction = await
                 account.mostRecentTransaction()) {

}
f().then(...);

// WALKTHROUGH:
// MOVQ $SP[-2], $HeapBinding($2) << undefined
// JNE account, UndefinedConstant(), &next_binding << Does not jump
// SuspendGenerator($generator);
// MOVQ $GeneratorInput, $HeapBinding($2) << result of accounts.lookup(id)
[...]

PR #1136, opened 2018-03-18
normative: allow await in parameter defaults (#917)

pemrouz:
This addresses a spec bug to allow using the await keyword in parameter
defaults for non-arrow async functions:

```js
async function foo(bar = await baz) { ... }
```

This enables developers to have a consistent way to default parameters
using both async and non-async values, which otherwise quickly becomes
a common pain point. A complete motivating example is described in [1].

The solution converged on here after several iterations has been refined
to avoid creating unnecessary duplication between the different cases. In
EvaluateBody (14.7.11) the promise is already created before
FunctionDeclarationInstantiation (which is also already rejected if
initialisation throws). Instead of creating an
AsyncFunctionDeclarationInstantiation, creating additional promises, or
duplicating the boilerplate in AsyncFunctionStart, the call to FDI has
simply been moved into AsyncFunctionStart which then allows usage of the
await keyword during initialisation. It also removes the corresponding
early errors in 14.7.1 (Static Semantics: Early Errors) and 19.2.1.1.1
(CreateDynamicFunction).

This PR is to use a basis for discussion in the committee. We are already
aware of and plan to make the analogous change to async generators before
landing.

Reviewers: @littledan, @bterlson, @syg
Tests:
https://github.com/tc39/test262/compare/master...pemrouz:normative/await-in-parameter-defaults?expand=1

[1] https://github.com/tc39/ecma262/issues/917#issuecomment-336945991

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -20556,7 +20556,6 @@
       <ul>
         <li>It is a Syntax Error if ContainsUseStrict of |AsyncFunctionBody| is *true* and IsSimpleParameterList of |UniqueFormalParameters| is *false*.</li>
         <li>It is a Syntax Error if HasDirectSuper of |AsyncMethod| is *true*.</li>
-        <li>It is a Syntax Error if |UniqueFormalParameters| Contains |AwaitExpression| is *true*.</li>
         <li>It is a Syntax Error if any element of the BoundNames of |UniqueFormalParameters| also occurs in the LexicallyDeclaredNames of |AsyncFunctionBody|.</li>
       </ul>
       <emu-grammar>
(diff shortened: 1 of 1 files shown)


littledan:
The plan is to extend this change to async generators, right?

bakkot:
I'm going to close this in light of https://github.com/tc39/ecma262/issues/917#issuecomment-374730357 and [the discussion in plenary](https://github.com/tc39/notes/blob/master/meetings/2018-03/mar-20.md#10ij-await-in-parameter-defaults-for-non-arrow-async-functions).

PR #1137, opened 2018-03-19
Editorial: clarify when early errors for functions apply

### LINKED ISSUE #632: Duplicate parameters of functions whose bodies are strict

bakkot:
Fixes #632.

This is _arguably_ normative, in that it changes the interpretation of a strict reading of the spec (which, I argue in the above issue, implies that `function f(a, a) { 'use strict'; }` would be legal, contrary to the intent of the author of the relevant text, implementations, and common sense).

In practice I think we're regarding this as either fixing a spec bug or merely a clarification, depending on whether you buy my above reading. So I've marked it as editorial.

Note that I've also changed the rule for names to be based on the strictness of `BindingIdentifier` rather than the strictness of the entire `FunctionDeclaration` (or similar) production. I don't think that actually changes anything, since the full production can't be strict unless all of its parts are. I'm mainly making this change to minimize ambiguity and to better match #1091.

cc @jugglinmike.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -18057,10 +18057,10 @@
       </emu-grammar>
       <ul>
         <li>
-          If the source code matching this production is strict mode code, the Early Error rules for <emu-grammar>UniqueFormalParameters : FormalParameters</emu-grammar> are applied.
+          If the source code matching |FormalParameters| is strict mode code, the Early Error rules for <emu-grammar>UniqueFormalParameters : FormalParameters</emu-grammar> are applied.
         </li>
         <li>
-          If the source code matching this production is strict mode code, it is a Syntax Error if |BindingIdentifier| is present and the StringValue of |BindingIdentifier| is `"eval"` or `"arguments"`.
+          If |BindingIdentifier| is present and the source code matching |BindingIdentifier| is strict mode code, it is a Syntax Error if the StringValue of |BindingIdentifier| is `"eval"` or `"arguments"`.
         </li>
         <li>
           It is a Syntax Error if ContainsUseStrict of |FunctionBody| is *true* and IsSimpleParameterList of |FormalParameters| is *false*.
@@ -19002,10 +19002,10 @@
       </emu-grammar>
       <ul>
         <li>
-          If the source code matching this production is strict mode code, the Early Error rules for <emu-grammar>UniqueFormalParameters : FormalParameters</emu-grammar> are applied.
+          If the source code matching |FormalParameters| is strict mode code, the Early Error rules for <emu-grammar>UniqueFormalParameters : FormalParameters</emu-grammar> are applied.
         </li>
         <li>
-          If the source code matching this production is strict mode code, it is a Syntax Error if |BindingIdentifier| is present and the StringValue of |BindingIdentifier| is `"eval"` or `"arguments"`.
+          If |BindingIdentifier| is present and the source code matching |BindingIdentifier| is strict mode code, it is a Syntax Error if the StringValue of |BindingIdentifier| is `"eval"` or `"arguments"`.
         </li>
         <li>
           It is a Syntax Error if ContainsUseStrict of |GeneratorBody| is *true* and IsSimpleParameterList of |FormalParameters| is *false*.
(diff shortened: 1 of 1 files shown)


bakkot:
This should hold off until whatever replaces #1091, I think.

jmdyck:
semi-recap: The edits re parameters would be sufficient to resolve #632, and could be merged as is. The edits re BindingIdentifiers, as currently worded, depend on a change that extends a function's strictness to its BindingIdentifier, which is now under discussion at PR #1158.

jugglinmike:
Thanks for the recap, @jmdyck!

@bterlson @bradleymeck @ljharb do you think we could move forward with this? I'll be happy to update gh-1158 once this is is available.

jugglinmike:
@bterlson @bradleymeck @ljharb I'm still interested in helping out here, especially as it relates to gh-1158. That's a normative change that received consensus earlier this year, so I'd like to see it landed in time for the next edition.

Could you let me know if there's anything blocking this change?

ljharb:
@jugglinmike there's nothing blocking this change yet; it's in @bterlson's queue to review. I don't believe #1158 has gone before the committee, though.

bakkot:
It was also my impression that we had consensus for the normative change made by the combination of this PR and #1158 (specifically the version in https://github.com/tc39/ecma262/pull/1158#issuecomment-413916320 or something like it).

Whether the normative change is in this PR, that one, or their sum depends on your interpretation of the current spec, but regardless I think we want both, and the normative change has consensus. I believe the clearest interpretation is that this PR is editorial and #1158 normative.

bakkot:
@zenparsing If you have alternative suggestions, I'm all ears! I think the current wording is more confusing because it is _not_ the case that the function production as a whole is either strict or non-strict: the definition of "strict mode code" is given in terms of "function code", which specifically does not include the whole production despite that being how you might tend to think of it; hence #632 and this PR.

This PR doesn't currently seek to change the definition "function code", just fix early errors which are worded as if it included the whole production. The thread at #1158 talks through some possible more comprehensive changes to how "function code" and/or "strict mode code" are defined. I think it would be reasonable to pursue some such change, but would like to land the normative change implied by the combination of this PR and #1158 before we go down that rabbit hole any further.

littledan:
As long as we're talking about readability: Spec text in this area is very repetitive across different, but similar, productions! If someone has an idea about how to unify these sorts of things a bit more, that could make it all a lot easier to read and maintain.

jmdyck:
(@littledan: If early error rules were more 'algorithmic', you could abstract over the commonalities by introducing an abstract operation that would be called for each of the different-but-similar productions.

But we probably shouldn't distract this PR by discussing that here.)

PR #1142, opened 2018-03-20
Editorial: create a ThrowCompletion abstract operation to reduce boilerplate

ljharb:
Using the record notation for creating a thrown TypeError completion is super awkward; this new abstract operation makes it read much more clearly for me.

Thoughts?

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -812,7 +812,7 @@
         </emu-alg>
         <p>mean the same things as:</p>
         <emu-alg>
-          1. Return Completion{[[Type]]: ~throw~, [[Value]]: a newly created *TypeError* object, [[Target]]: ~empty~}.
+          1. Return ThrowCompletion(a newly created *TypeError* object).
         </emu-alg>
       </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -3138,6 +3138,18 @@ <h1>NormalCompletion</h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-throwcompletion" aoid="ThrowCompletion">
+        <h1>ThrowCompletion</h1>
+        <p>The abstract operation ThrowCompletion with a single _argument_, such as:</p>
+        <emu-alg>
+          1. Return ThrowCompletion(_argument_).
+        </emu-alg>
+        <p>Is a shorthand that is defined as follows:</p>
```
jmdyck: I don't think there's any need to make it a shorthand. I suppose this is by analogy with NormalCompletion, but I don't think that needs to be a shorthand either.

ljharb: I find NormalCompletion to be awesome at simplifying the spec; what would you suggest instead of both?

jmdyck: Sorry, I'm not saying it shouldn't be defined, just that it doesn't need to be defined *as a "shorthand"* (macro). That is, it can be defined as a conventional abstract operation.

ljharb: Ah, sure. Yes, this was just consistency with NormalCompletion; I think changing both is fine but should be done in a followup (or, NormalCompletion should be changed prior to this PR going in).

jmdyck: E.g., `ReturnIfAbrupt` *has* to be defined as shorthand/macro, because 'invoking' it can cause a return from the invoking algorithm, which a conventional abstract operation can't do.

jmdyck: I'm fine to submit a follow-up to change both.

ljharb: Yay! Then your stamp on this PR would be appreciated ;-)

jmdyck: My stamp?

ljharb: A PR approval via Github's PR Reviews feature :-) only if you approve, of course!

bterlson:
I like this, thanks!

PR #1144, opened 2018-03-21
Normative: Cleanup Time Values and Time Range

apaprocki:
Implementations are not currently consistently limiting *time values* to the spec defined ranges of both *time value* and *extended years*.

This change is editorial cleanup that more clearly states the allowed *time value* range and the supported range of *extended years* that fit within the allowed *time value* range.

Implementations will need to be updated to conform with the spec (SM currently adheres) and pass the accompanying test in tc39/test262#1501.

**Concrete examples of what will be fixed are detailed in #1145.**

Time Values and Time Range:

- Language about `Number` range in hundreds-of-thousands of years plus/minus UTC epoch is not immediately relevant; move it to an `emu-note`
- New `emu-note` text fixes inconsistency in spec where this section was not properly accounting for leap years when computing the approximate number of years in the maximum range but 20.3.1.15.1 *extended years* was stating it correctly
- Add `emu-xref` for Number integer range
- Add comment referring to POSIX day for the source of 86,400 seconds
- Simplify remaining spec in section into clear, consistent language

Extended Years:

- Add `dfn` to *extended years* so it is linked from format section
- Add `emu-xref` for *time value* range
- Explicitly state *extended years* relation to *time value* range
- Explicitly state ISO 8601 strings outside *time value* range are not accepted
- Modify *extended years* example table to reflect the proper minimum/maximum values allowed by the *time value* range

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -26482,10 +26482,15 @@
 
       <emu-clause id="sec-time-values-and-time-range">
         <h1>Time Values and Time Range</h1>
-        <p>A Date object contains a Number indicating a particular instant in time to within a millisecond. Such a Number is called a <dfn>time value</dfn>. A time value may also be *NaN*, indicating that the Date object does not represent a specific instant of time.</p>
-        <p>Time is measured in ECMAScript in milliseconds since 01 January, 1970 UTC. In time values leap seconds are ignored. It is assumed that there are exactly 86,400,000 milliseconds per day. ECMAScript Number values can represent all integers from -9,007,199,254,740,992 to 9,007,199,254,740,992; this range suffices to measure times to millisecond precision for any instant that is within approximately 285,616 years, either forward or backward, from 01 January, 1970 UTC.</p>
-        <p>The actual range of times supported by ECMAScript Date objects is slightly smaller: exactly -100,000,000 days to 100,000,000 days measured relative to midnight at the beginning of 01 January, 1970 UTC. This gives a range of 8,640,000,000,000,000 milliseconds to either side of 01 January, 1970 UTC.</p>
-        <p>The exact moment of midnight at the beginning of 01 January, 1970 UTC is represented by the value *+0*.</p>
+        <p>A Date object contains a Number representing an instant in time with millisecond precision. Such a Number is called a <dfn>time value</dfn>. A time value may also be *NaN*, indicating that the Date object does not represent a specific instant in time.</p>
+        <p>Time is measured in ECMAScript as milliseconds since midnight at the beginning of 01 January, 1970 UTC. Time in ECMAScript does not observe leap seconds; they are ignored. Time calculations assume each and every day contains exactly <emu-eqn>60 &times; 60 &times; 24 &times; 1000 = 86,400,000</emu-eqn> milliseconds, to align with the POSIX specification of each and every day containing exactly 86,400 seconds.</p>
+        <p>A Number can exactly represent all integers from -9,007,199,254,740,992 to 9,007,199,254,740,992 (<emu-xref href="#sec-number.min_safe_integer"></emu-xref> and <emu-xref href="#sec-number.max_safe_integer"></emu-xref>). A time value supports a slightly smaller range of exactly -100,000,000 days to 100,000,000 days measured relative to midnight at the beginning of 01 January, 1970 UTC. This yields an exact supported time value range of -8,640,000,000,000,000 to 8,640,000,000,000,000 milliseconds relative to midnight at the beginning of 01 January, 1970 UTC.</p>
+        <p>The exact moment of midnight at the beginning of 01 January, 1970 UTC is represented by the time value *+0*.</p>
+        <emu-note>
+          <p>The 400 year cycle of the Gregorian calendar contains 97 leap years. This yields an average of 365.2425 days per year, or an average of 31,556,952,000 milliseconds per year under the Gregorian calendar. ECMAScript applies a proleptic Gregorian calendar for all time computations.</p>
+          <p>As specified by this section, the maximum year range a Number can represent exactly with millisecond precision is approximately -285,426 to 285,426 years relative to midnight at the beginning of 01 January, 1970 UTC.</p>
+          <p>As specified by this section, the maximum year range a time value can represent is approximately -273,790 to 273,790 years relative to midnight at the beginning of 01 January, 1970 UTC.</p>
+        </emu-note>
       </emu-clause>
 
       <emu-clause id="sec-day-number-and-time-within-day">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -27388,10 +27388,15 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
-        <p>The actual range of times supported by ECMAScript Date objects is slightly smaller: exactly -100,000,000 days to 100,000,000 days measured relative to midnight at the beginning of 01 January, 1970 UTC. This gives a range of 8,640,000,000,000,000 milliseconds to either side of 01 January, 1970 UTC.</p>
-        <p>The exact moment of midnight at the beginning of 01 January, 1970 UTC is represented by the value *+0*.</p>
+        <p>A Date object contains a Number representing an instant in time with millisecond precision. Such a Number is called a <dfn>time value</dfn>. A time value may also be *NaN*, indicating that the Date object does not represent a specific instant in time.</p>
+        <p>Time is measured in ECMAScript as milliseconds since midnight at the beginning of 01 January, 1970 UTC. Time in ECMAScript does not observe leap seconds; they are ignored. Time calculations assume a day contains exactly <emu-eqn>60 &times; 60 &times; 24 &times; 1000 = 86,400,000</emu-eqn> milliseconds.</p>
```
maggiepint: If we are going this deep, do we want to discuss *why* we don't support leap seconds? I feel like that gets asked a lot, and of course it goes back to the fact that we're fundamentally using Unix time which doesn't support them here.

Small nit - I believe we should say `time calculations assume a UTC day contains exactly...`

jmdyck: >  I believe we should say `time calculations assume a UTC day contains exactly...`

(I.e., inserting "UTC".) I think we should go in the opposite direction, removing most mentions of UTC, for the very reason that ES doesn't support leap seconds.

apaprocki: @maggiepint I can ask around tomorrow, but there might be some reason why Unix / POSIX can't be referenced specifically in the spec (neither are currently).

@jmdyck I'd tend to agree re: leaving UTC out when referring to the length of a day. One of the most common references for people (Wikipedia) clearly states a UTC day sometimes contains an extra second, so it would be like the spec saying "A UTC day is not a UTC day".

As for removing most mentions of UTC, the spec currently has 69 lines that mention UTC at least once. Even if removal is technically sound, careful wording would have to replace it and it's a larger proposal that would have to be discussed separately.

maggiepint:
I suppose UTC isn't right.

What I'm going for is *not local*, because of course the spec does support
local time which, due to DST or zone splitting, has days not of that length.

spec.html
jmdyck: > As for removing most mentions of UTC, [...] it's a larger proposal that would have to be discussed separately.

Agreed.

maggiepint: I suppose UTC isn't right.

What I'm going for is *not local*, because of course the spec does support
local time which, due to DST or zone splitting, has days not of that length.

apaprocki: I asked @bterlson and he didn't see an issue with mentioning the alignment with the POSIX specification of a day, so see my last push for proposed wording.

@jmdyck When looking back over the POSIX spec, it appears they did exactly as you are suggesting and changed all references from "**seconds since 00:00:00 UTC (Coordinated
Universal Time), January 1 1970**" to "**seconds since the Epoch**". I'd be ok with attempting to tackle that as a subsequent PR. I'm guessing that might need to be brought up to the committee because it would likely involve defining the ES Epoch explicitly as 1970-01-01T00:00:00 TAI and adding similar implementation-defined language that POSIX resorted to:

> The relationship between the actual time of day and the current value for seconds since the Epoch is unspecified.

> How any changes to the value of seconds since the Epoch are made to align to a desired relationship with the current actual time is implementation-defined. As represented in seconds since the Epoch, each and every day shall be accounted for by exactly 86400 seconds.

spec.html
```diff
@@ -27753,6 +27762,10 @@ <h1>Extended Years</h1>
                     <td>+002009-12-15T00:00:00Z</td>
                     <td>2009 A.D.</td>
                   </tr>
+                  <tr>
+                    <td>+275760-09-13T00:00:00Z</td>
+                    <td>275760 A.D.</td>
+                  </tr>
```
littledan: What is the purpose of the addition of these extended years? It doesn't look like they are boundary cases, due to the more extremely values preceding and following them.

littledan: Oh I see, these are what you are separately proposing to be the new extreme permitted values, right?

apaprocki:
@littledan I reformatted this as a single normative commit. I'll add the PR onto the agenda for May.

littledan:
For good measure, let's see if we can get in touch with implementers ahead of time to see how they would feel about making this change in their JS engines:

cc @hashseed @jungshik @bterlson @kmiller68

spec.html
```diff
@@ -27723,15 +27728,15 @@ <h1>Date Time String Format</h1>
...
+          <p>ECMAScript requires the ability to specify 6 digit years, or <dfn>extended years</dfn>, to render the full range of a time value; -100,000,000 days to 100,000,000 days, or approximately -273,790 to 273,790 years, relative to midnight at the beginning of 01 January, 1970 UTC (<emu-xref href="#sec-time-values-and-time-range"></emu-xref>). To represent years before 0 or after 9999, ISO 8601 permits the expansion of the year representation, but only by prior agreement between the sender and the receiver. In the simplified ECMAScript format such an expanded year representation shall have 2 extra year digits and is always prefixed with a + or - sign. The year 0 is considered positive and hence prefixed with a + sign. ISO 8601 formatted strings with extended years representing points in time outside the range of a time value are not accepted.</p>
```
littledan: I'm not sure if "ISO 8601 formatted strings with extended years representing points in time outside the range of a time value are not accepted." makes it clear that these date/time strings will result in an Invalid Date rather than being allowed to fall back to the implementation-specific parser. I think it'd be best if this behavior were explicitly described in `Date.parse`. (Apologies for not mentioning this the first time around.)

apaprocki:
@littledan I believe I addressed your last concern by explicitly stating

> are represented as the Number *NaN* when parsed by `Date.parse()` and do not fall back to implementation-specific behavior or heuristics.

instead of

> are not accepted.

This area of `Date.parse` could be cleaned up later under work you were suggesting, but this should be very clear in the meantime.

spec.html
littledan: The new text

> ISO 8601 formatted strings with extended years representing points in time outside the range of a time value are represented as the Number *NaN* when parsed by `Date.parse()` and do not fall back to implementation-specific behavior or heuristics.

Looks great to me. Thanks for the clarification.

apaprocki:
@ljharb can this be merged now? @rwaldron already merged the related test262

apaprocki:
@littledan it seems like somehow this lost your review, could you re-ack?

apaprocki:
@ljharb @bterlson can this be merged now? It’s been lingering, thanks

PR #1146, opened 2018-03-21
Editorial: refactor PerformPromiseThen to allow no capability

### LINKED ISSUE #694: Spec factoring: allow PerformPromiseThen with no capability
PerformPromiseThen is used often to run some spec code after a promise settles. However, often what's being done is not a promise transform, but instead just some reaction to the promise. So you don't need the promise-returning behavior that `Promise.prototype.then` does.

Examples of this are in https://github.com/tc39/ecma262/pull/692 and in the cancelable promises proposal at two locations: https://tc39.github.io/proposal-cancelable-promises/#sec-canceltoken.race and https://tc39.github.io/proposal-cancelable-promises/#sec-promise.withcanceltoken. In general we've been signaling this by creating a new "throwawayCapability" to pass in. As @littledan pointed out in #692, this is error prone since you need to make sure to mark the throwawayCapability.[[Promise]] as handled.

The ideal solution, in my opinion, would be to make the capability argument to PerformPromiseThen optional. In more detail:
- Change the definition of the [[Capability]] field of PromiseReaction records to allow undefined
- In PerformPromiseThen, mark _resultCapability_ as optional, and add a line saying that if it's not provided, it gets set to undefined. (I think that's how we handle optional args to abstract ops...)
- In PromiseReactionJob, before "If _handlerResult_ is an abrupt completion", add an "Assert: if _promiseCapability_ is undefined, _handlerResult_ is not an abrupt completion." (I.e., specs using PerformPromiseThen must never throw an error in their fulfill/reject handlers.) Then, if _promiseCapability_ is undefined, return NormalCompletion(undefined) instead of doing the rest of the steps.

ljharb:
Fixes #694.

I followed @domenic's steps in the linked issue here; I'm not sure if other changes are needed.

Specifically, which `PerformPromiseThen` calls should now be able to omit a capability?

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3005,9 +3005,7 @@
           1. Let _stepsRejected_ be the algorithm steps defined in <emu-xref href="#await-rejected" title></emu-xref>.
           1. Let _onRejected_ be CreateBuiltinFunction(_stepsRejected_, &laquo; [[AsyncContext]] &raquo;).
           1. Set _onRejected_.[[AsyncContext]] to _asyncContext_.
-          1. Let _throwawayCapability_ be ! NewPromiseCapability(%Promise%).
-          1. Set _throwawayCapability_.[[Promise]].[[PromiseIsHandled]] to *true*.
-          1. Perform ! PerformPromiseThen(_promiseCapability_.[[Promise]], _onFulfilled_, _onRejected_, _throwawayCapability_).
+          1. Perform ! PerformPromiseThen(_promiseCapability_.[[Promise]], _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
           1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
         </emu-alg>
@@ -37232,9 +37230,7 @@ THH:mm:ss.sss
                 1. Let _stepsRejected_ be the algorithm steps defined in <emu-xref href="#async-generator-resume-next-return-processor-rejected" title></emu-xref>.
                 1. Let _onRejected_ be CreateBuiltinFunction(_stepsRejected_, &laquo; [[Generator]] &raquo;).
                 1. Set _onRejected_.[[Generator]] to _generator_.
-                1. Let _throwawayCapability_ be ! NewPromiseCapability(%Promise%).
-                1. Set _throwawayCapability_.[[Promise]].[[PromiseIsHandled]] to *true*.
-                1. Perform ! PerformPromiseThen(_promiseCapability_.[[Promise]], _onFulfilled_, _onRejected_, _throwawayCapability_).
+                1. Perform ! PerformPromiseThen(_promiseCapability_.[[Promise]], _onFulfilled_, _onRejected_).
                 1. Return *undefined*.
               1. Else,
                 1. Assert: _completion_.[[Type]] is ~throw~.
@@ -37443,7 +37439,7 @@ THH:mm:ss.sss
                 [[Capability]]
               </td>
               <td>
-                A PromiseCapability Record
+                A PromiseCapability Record, or *undefined*
               </td>
               <td>
                 The capabilities of the promise for which this record provides a reaction handler.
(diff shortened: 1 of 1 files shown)


domenic:
Search for "throwaway capability"

ljharb:
@domenic thanks, updated!

spec.html
```diff
@@ -39095,11 +39094,14 @@ <h1>Promise.prototype.then ( _onFulfilled_, _onRejected_ )</h1>
...
-          <h1>PerformPromiseThen ( _promise_, _onFulfilled_, _onRejected_, _resultCapability_ )</h1>
+          <h1>PerformPromiseThen ( _promise_, _onFulfilled_, _onRejected_ [ , _resultCapability_ ] )</h1>
           <p>The abstract operation PerformPromiseThen performs the &ldquo;then&rdquo; operation on _promise_ using _onFulfilled_ and _onRejected_ as its settlement actions. The result is _resultCapability_'s promise.</p>
           <emu-alg>
             1. Assert: IsPromise(_promise_) is *true*.
-            1. Assert: _resultCapability_ is a PromiseCapability Record.
+            1. If _resultCapability_ is present, then
+              1. Assert: _resultCapability_ is a PromiseCapability Record.
+            1. Else,
+              1. Set _resultCapability to *undefined*.
```
domenic: Missing `_`

I think maybe you should use "Let" here instead of "Set" since the variable doesn't technically exist yet? (Which might argue for moving the else clause into a single line, so the "block scoping" is less confusing.)

spec.html
```diff
@@ -38204,7 +38202,7 @@ <h1>AsyncGeneratorRequest Records</h1>
             </tr>
             <tr>
               <td>[[Capability]]</td>
-              <td>A PromiseCapability record</td>
+              <td>A PromiseCapability record, or *undefined*</td>
```
domenic: I think you meant to edit PromiseReaction, not AsyncGeneratorRequest.

spec.html
```diff
@@ -39095,11 +39094,14 @@ <h1>Promise.prototype.then ( _onFulfilled_, _onRejected_ )</h1>
 
         <!-- es6num="25.4.5.3.1" -->
         <emu-clause id="sec-performpromisethen" aoid="PerformPromiseThen">
-          <h1>PerformPromiseThen ( _promise_, _onFulfilled_, _onRejected_, _resultCapability_ )</h1>
+          <h1>PerformPromiseThen ( _promise_, _onFulfilled_, _onRejected_ [ , _resultCapability_ ] )</h1>
           <p>The abstract operation PerformPromiseThen performs the &ldquo;then&rdquo; operation on _promise_ using _onFulfilled_ and _onRejected_ as its settlement actions. The result is _resultCapability_'s promise.</p>
```
domenic: Should modify the intro text. Suggestion:

> If _resultCapability_ is passed, the result is stored by updating _resultCapability_'s promise. (If it is not passed, then PerformPromiseThen is being called by a specification-internal operation where the result does not matter.)

spec.html
ljharb: The variable exists because it's in the abstract operation signature, so I think "Set" is correct.

Thanks, I've added the underscore

domenic: Yeah, I guess that depends on how you interpret "is present", but your way makes sense too.

ljharb:
Thanks, updated!

ljharb:
@anba @domenic @gsathya @msaboff any objections to merging this editorial change into ES2018?

domenic:
I don't like spending resources on snapshots, but whatevs...

littledan:
@mayalekova We chatted about this change and it sounded like you might have had some reservations about it. Are any embedding environments depending on the throwaway Promise, based on embedder-specific hooks into the Promise lifecycle, even if it's unobservable without those hooks?

MayaLekova:
For now it seems we won't fire the promise hooks on the throwawayPromise, don't know if this is the only change. @ak239 - please comment as well!

ak239:
On DevTools side to implement async stacks we rely on kBefore and kAfter hook calls for throwaway promise:
- each time when async function is suspended - we capture stack trace at this moment,
- when async function is resumed next time - we use kBefore and kAfter hooks - between these hooks we use captured async stack as current one.

From implementation point of view I believe we can instrument microtask by itself with inspector-only hooks or we can stop using promise hooks at all by proper instrumentation async function suspended, async function resumed and async function finished. It will work for us but I am not sure about some external developers who relies on promise hooks.

littledan:
@ak239 Could you explain more how DevTools currently takes advantage of the kBefore and kAfter events related to the throwawayPromise?

ak239:
When function is suspended we get instrumentation call with async function promise as argument and record stack trace for this promise.
On await we record mapping between throwaway promise and async function promise and expect that when async function is running next time we will get kBefore hook with throwaway promise before that run and kAfter hook after. We use recorded mapping between throwaway and async function promise here to map these hooks to async function.

As external developer I think that I can use:
- kInit when throwaway promise is created to capture stack,
- kBefore and kAfter when async function is resumed to set captured stack as current one.

What we actually would like to have: instrumentation every time when async function suspended, every time when async function is resumed and once when async function finished.

ak239:
On inspector side we can implement required for us instrumentation without using `throwawayPromise`: [PoC](https://chromium-review.googlesource.com/c/v8/v8/+/1114311). So from inspector point of view this PR looks good to me.

I am not sure how much important is providing these hooks for external developers.

Issue #1130, opened 2018-03-06
Spec bug with object rest destructuring assignment


woess:
Consider this test case from [test262](https://github.com/tc39/test262/blob/master/test/language/expressions/assignment/dstr-obj-rest-val-null.js) which expects a TypeError:
```js
var rest;
assert.throws(TypeError, function() {
  0, {...rest} = null;
});
```
The problem: I cannot find the corresponding passage in the spec.

I believe that **12.15.5.2 Runtime Semantics: DestructuringAssignmentEvaluation** is missing a `Perform ? RequireObjectCoercible(value).` as first step.

I'll walk you through the path I took, so you can correct me if I went wrong somewhere.

> 12.15.4 Runtime Semantics: Evaluation
> `AssignmentExpression : LeftHandSideExpression = AssignmentExpression`
>   1. If LeftHandSideExpression is neither an ObjectLiteral nor an ArrayLiteral [...]
>   2. Let assignmentPattern be the AssignmentPattern that is covered by LeftHandSideExpression.
>   3. Let rref be the result of evaluating AssignmentExpression.
>   4. Let rval be ? GetValue(rref).
>   5. Perform ? **DestructuringAssignmentEvaluation** of assignmentPattern using rval as the argument.
>   6. Return rval.
> 
> 12.15.5.2 Runtime Semantics: **DestructuringAssignmentEvaluation** (value)
> `ObjectAssignmentPattern : { AssignmentRestProperty }`
>   1. Let excludedNames be a new empty List.
>   2. Return the result of performing **RestDestructuringAssignmentEvaluation** of AssignmentRestProperty with value and excludedNames as the arguments.
> 
> 12.15.5.4 Runtime Semantics: **RestDestructuringAssignmentEvaluation** (value, excludedNames)
> `AssignmentRestProperty : ... DestructuringAssignmentTarget`
>   1. Let lref be the result of evaluating DestructuringAssignmentTarget.
>   2. ReturnIfAbrupt(lref).
>   3. Let restObj be ObjectCreate(%ObjectPrototype%).
>   4. Perform ? **CopyDataProperties**(restObj, value, excludedNames).
>   5. Return PutValue(lref, restObj).
> 
> 7.3.23 **CopyDataProperties** (target, source, excludedItems)
>   1. Assert: Type(target) is Object.
>   2. Assert: excludedItems is a List of property keys.
>   3. If source is **undefined** or **null**, return target.

anba:
[...]

PR #1163, opened 2018-04-05
Normative: specify RequireObjectCoercible as first step in destructuring assignment semantics

musgravejw:
Issue tracked here:  https://github.com/tc39/ecma262/issues/1130

This PR attempts to specify that the runtime semantics for **DestructuringAssignmentEvaluation (12.14.5.2)** should perform `RequireObjectCoercible(value)` as the first step, since assignment to `null` throws a TypeError as specified in [this test assertion](https://github.com/tc39/test262/blob/master/test/language/expressions/assignment/dstr-obj-rest-val-null.js).

Thank you.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -14192,12 +14192,14 @@
         </emu-alg>
         <emu-grammar>ObjectAssignmentPattern : `{` AssignmentRestProperty `}`</emu-grammar>
         <emu-alg>
+          1. Perform ? RequireObjectCoercible(_value_).
           1. Let _excludedNames_ be a new empty List.
           1. Return the result of performing RestDestructuringAssignmentEvaluation of |AssignmentRestProperty| with _value_ and _excludedNames_ as the arguments.
         </emu-alg>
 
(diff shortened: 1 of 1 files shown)


littledan:
To make a normative patch like this, you need to sign https://tc39.github.io/agreements/contributor/ . I don't see your name on the list of signatories. Would you be OK with filling that out?

The patch itself LGTM.

Are there test262 tests which verify the behavior in this case?

Cc @gsathya @caiolima

musgravejw:
@littledan: Absolutely.  Thanks!

littledan:
Thanks for signing. I can confirm that the form seems to be sent in properly.

caiolima:
@littledan If I understood correctly, this [test](https://github.com/tc39/test262/blob/master/test/language/expressions/assignment/dstr-obj-rest-val-null.js) is covering this update.

domenic:
For the web we've started making all instances of Assert link to https://infra.spec.whatwg.org/#assert

littledan:
I don't see any test updates to make. The patch looks ready to merge to me.

PR #1164, opened 2018-04-05
Editorial: Default littleEndian to false in DataView.p.getInt32

jridgewell:
All the other getter/setter methods default `littleEndian` to `false`
when it is not present.

This doesn't affect the overall algorithm.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -36699,7 +36699,7 @@ THH:mm:ss.sss
         <p>When the `getInt32` method is called with argument _byteOffset_ and optional argument _littleEndian_, the following steps are taken:</p>
         <emu-alg>
           1. Let _v_ be the *this* value.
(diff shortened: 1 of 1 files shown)


jridgewell:
`GetViewValue` is called with user-input `littleEndian`, so we can't remove the `ToBoolean`.

bmeck:
LGTM. Curious on history of why are we not pulling this off of the [Agent Record](https://tc39.github.io/ecma262/#agent-record), but not strictly related.

annevk:
Pretty sure this existed before agents. Should probably be refactored.

jmdyck:
(DataView, including DataView.prototype.getInt32, was added in rev 10 of the ES6 draft in Sept 2012. Agents were added to the ES2017 draft in commit 2e6a2546 in December 2016.)

PR #1177, opened 2018-04-19
Make Math & Reflect prose consistent with other no-construct no-call objects

rwaldron:


ljharb:
AsyncFunction, GeneratorFunction, and AsyncGeneratorFunction all can be both called and constructed; and i think Math and Reflect have been updated in the spec. There may not be anything left to do here.

rwaldron:
This has been corrected via another changeset.

Issue #1169, opened 2018-04-12
Date.parse cannot return an abrupt completion when called with a String value


anba:
https://tc39.github.io/ecma262/#sec-date-value, step 4.b.ii
> 1. Let tv be the result of parsing v as a date, in exactly the same manner as for the parse method (20.3.3.2). If the parse resulted in an abrupt completion, tv is the Completion Record.
> 2. ReturnIfAbrupt(tv).

should instead assert that calling `Date.parse` will never result in an abrupt completion:
> 1. Assert: The next step never returns an abrupt completion because v is a String value.
> 1. Let tv be the result of parsing v as a date, in exactly the same manner as for the parse method (20.3.3.2).

PR #1179, opened 2018-04-21
Editorial: specify Date.parse cannot return an abrupt completion

musgravejw:
Closes https://github.com/tc39/ecma262/issues/1169

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27849,8 +27849,8 @@ THH:mm:ss.sss
             1. Else,
               1. Let _v_ be ? ToPrimitive(_value_).
               1. If Type(_v_) is String, then
-                1. Let _tv_ be the result of parsing _v_ as a date, in exactly the same manner as for the `parse` method (<emu-xref href="#sec-date.parse"></emu-xref>). If the parse resulted in an abrupt completion, _tv_ is the Completion Record.
-                1. ReturnIfAbrupt(_tv_).
+                1. Assert: The next step never returns an abrupt completion because v is a String value.
+                1. Let _tv_ be the result of parsing _v_ as a date, in exactly the same manner as for the `parse` method (<emu-xref href="#sec-date.parse"></emu-xref>).
               1. Else,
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -27849,8 +27849,8 @@ <h1>Date ( _value_ )</h1>
             1. Else,
               1. Let _v_ be ? ToPrimitive(_value_).
               1. If Type(_v_) is String, then
+                1. Assert: parsing of String value _v_ does not return an abrupt completion.
                 1. Let _tv_ be the result of parsing _v_ as a date, in exactly the same manner as for the `parse` method (<emu-xref href="#sec-date.parse"></emu-xref>). If the parse resulted in an abrupt completion, _tv_ is the Completion Record.
```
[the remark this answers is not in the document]
musgravejw: @ljharb You're right.  Updated.

spec.html
```diff
@@ -27849,8 +27849,8 @@ <h1>Date ( _value_ )</h1>
             1. Else,
               1. Let _v_ be ? ToPrimitive(_value_).
               1. If Type(_v_) is String, then
-                1. Let _tv_ be the result of parsing _v_ as a date, in exactly the same manner as for the `parse` method (<emu-xref href="#sec-date.parse"></emu-xref>). If the parse resulted in an abrupt completion, _tv_ is the Completion Record.
-                1. ReturnIfAbrupt(_tv_).
+                1. Assert: parsing of String value _v_ does not return an abrupt completion.
```
bterlson: I actually liked @anba's wording better here. I think it's  useful to underscore that it's _because_ _v_ is a string that Date.parse won't throw (current text could imply that Date.parse doesn't throw ever or the string is constructed in a particular way so as not to throw).

musgravejw: Makes sense.

bterlson:
Looks good now! Thanks a lot, and congrats on your first spec text contribution to ECMA262!!

PR #1183, opened 2018-04-23
Editorial: replace obj.[[Get]](j) with Get(obj, j)

### LINKED ISSUE #1180: obj.[[Get]](j) in Array.prototype.sort section lacks the 2nd argument
In https://tc39.github.io/ecma262/#sec-array.prototype.sort section, the following sentence exists:

> Here the notation `old[j]` is used to refer to the hypothetical result of calling `obj.[[Get]](j)` before this function is executed, and the notation `new[j]` to refer to the hypothetical result of calling `obj.[[Get]](j)` after this function has been executed.

Now, `obj.[[Get]](j)` in the sentence doesn't have the 2nd argument `receiver`, so I guess that `this` in accessor properties are `undefined` in sorting. For example:

```js
const impl = Symbol()
class MyArrayLike {
    constructor() {
        this[impl] = [1, 2]
    }

    get 0() { return this[impl][0] } // this `this` is undefined in sorting.
    set 0(value) { this[impl][0] = value }
    get 1() { return this[impl][1] }
    set 1(value) { this[impl][1] = value }
    get length() { return 2 }
}

const xs = new MyArrayLike()
Array.prototype.sort.call(xs)
```

However, those `this`s are not `undefined` in real world.
I'm suspecting that we should fix those `obj.[[Get]](j)` to `obj.[[Get]](j, obj)`.

mysticatea:
Fixes #1180.

This PR replaces those `obj.[[Get]](j)`s by the abstract operation `Get(obj, j)` since those lacked the 2nd argument `Receiver`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33724,11 +33724,11 @@ THH:mm:ss.sss
         </ul>
         <p>The following steps are taken:</p>
         <emu-alg>
-          1. Perform an implementation-dependent sequence of calls to the [[Get]] and [[Set]] internal methods of _obj_, to the DeletePropertyOrThrow and HasOwnProperty abstract operation with _obj_ as the first argument, and to SortCompare (described below), such that:
-            * The property key argument for each call to [[Get]], [[Set]], HasOwnProperty, or DeletePropertyOrThrow is the string representation of a nonnegative integer less than _len_.
-            * The arguments for calls to SortCompare are values returned by a previous call to the [[Get]] internal method, unless the properties accessed by those previous calls did not exist according to HasOwnProperty. If both prospective arguments to SortCompare correspond to non-existent properties, use *+0* instead of calling SortCompare. If only the first prospective argument is non-existent use +1. If only the second prospective argument is non-existent use -1.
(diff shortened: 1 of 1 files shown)


ljharb:
@mysticatea can you update this PR?

mysticatea:
Hi. I apologize that I have lost on this in my busy life.

Thank you for your review. I have updated this PR.

- https://github.com/tc39/ecma262/pull/1183/commits/d5f0a7d6250860c02a0ac102e54969fecf665bff I updated "`[[Get]]` internal method" and "`[[Set]]` internal method" to "`Get` abstract operation" and "`Set` abstract operation". 
- https://github.com/tc39/ecma262/pull/1183/commits/92afd1686ddce84ff0da856183dd442e276e8888 Because the `Set` abstract operation has the functionality that throws `TypeError` if the `[[Set]]` internal method returned `false`, I updated to use it.
- https://github.com/tc39/ecma262/pull/1183/commits/31df05abacb1a8e709b4cc7926c9b0895b809ac3 Because the auto linker doesn't link `Set` words which are not followed by `(`, I enclosed the `Set` words by `<emu-xref>` tag.

PR #1185, opened 2018-04-25
allow catch bindings to be omitted

michaelficarra:
This implements the stage 3 proposal available at https://tc39.github.io/proposal-optional-catch-binding/.


bakkot:
This achieved [consensus](https://github.com/rwaldron/tc39-notes/blob/master/es9/2018-05/may-22.md#11id-optional-catch-binding-for-stage-4) for stage 4 at the May meeting. Anything holding up merging it?

ljharb:
Thanks for the ping; we wanted to wait until 2018 was cut, but I think we should be able to merge this by the next editor's meeting.

PR #1188, opened 2018-05-07
Extend ECMA-262 syntax into a superset of JSON

gibson042:
This implements https://github.com/tc39/proposal-json-superset , currently at stage 3.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9569,7 +9569,7 @@
 
   <emu-clause id="sec-line-terminators">
     <h1>Line Terminators</h1>
-    <p>Like white space code points, line terminator code points are used to improve source text readability and to separate tokens (indivisible lexical units) from each other. However, unlike white space code points, line terminators have some influence over the behaviour of the syntactic grammar. In general, line terminators may occur between any two tokens, but there are a few places where they are forbidden by the syntactic grammar. Line terminators also affect the process of automatic semicolon insertion (<emu-xref href="#sec-automatic-semicolon-insertion"></emu-xref>). A line terminator cannot occur within any token except a |StringLiteral|, |Template|, or |TemplateSubstitutionTail|. Line terminators may only occur within a |StringLiteral| token as part of a |LineContinuation|.</p>
+    <p>Like white space code points, line terminator code points are used to improve source text readability and to separate tokens (indivisible lexical units) from each other. However, unlike white space code points, line terminators have some influence over the behaviour of the syntactic grammar. In general, line terminators may occur between any two tokens, but there are a few places where they are forbidden by the syntactic grammar. Line terminators also affect the process of automatic semicolon insertion (<emu-xref href="#sec-automatic-semicolon-insertion"></emu-xref>). A line terminator cannot occur within any token except a |StringLiteral|, |Template|, or |TemplateSubstitutionTail|. &lt;LF&gt; and &lt;CR&gt; line terminators cannot occur within a |StringLiteral| token except as part of a |LineContinuation|.</p>
     <p>A line terminator can occur within a |MultiLineComment| but cannot occur within a |SingleLineComment|.</p>
     <p>Line terminators are included in the set of white space code points that are matched by the `\\s` class in regular expressions.</p>
     <p>The ECMAScript line terminator code points are listed in <emu-xref href="#table-33"></emu-xref>.</p>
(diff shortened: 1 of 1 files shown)


ljharb:
Tests: https://github.com/tc39/test262/pull/1543

littledan:
Thanks for your contribution, @gibson042 .

spec.html
```diff
@@ -9908,7 +9908,7 @@ <h2>Syntax</h2>
...
+    <p>Like white space code points, line terminator code points are used to improve source text readability and to separate tokens (indivisible lexical units) from each other. However, unlike white space code points, line terminators have some influence over the behaviour of the syntactic grammar. In general, line terminators may occur between any two tokens, but there are a few places where they are forbidden by the syntactic grammar. Line terminators also affect the process of automatic semicolon insertion (<emu-xref href="#sec-automatic-semicolon-insertion"></emu-xref>). A line terminator cannot occur within any token except a |StringLiteral|, |Template|, or |TemplateSubstitutionTail|. &lt;LF&gt; and &lt;CR&gt; line terminators cannot occur within a |StringLiteral| token except as part of a |LineContinuation|.</p>
```
ljharb: For clarity and future explorers, can you elaborate on why this restriction is narrowed?

gibson042: It's the fundamental change of [proposal-json-superset](https://github.com/tc39/proposal-json-superset), just in a part of the spec that I didn't realize included affected text until working on https://github.com/tc39/test262/pull/1565 . As of this PR landing, _StringLiteral_ will be able to include any _LineTerminator_ as part of a _LineContinuation_ (backslash followed by _LineTerminatorSequence_, used to statically define long strings over multiple lines without affecting their content), and—invalidating the old text—will **also** be able to include U+2028 LINE SEPARATOR and U+2029 PARAGRAPH SEPARATOR line terminators in any position (as is the case for JSON strings).

mathiasbynens: Previously, no line terminators could occur in string literals (except as part of a `LineContinuation`).

With the change, U+2028 and U+2029 can occur in string literals, even outside of `LineContinuation`.

In other words, this change clarifies that `LineContinuation`s are unaffected by the change, i.e. `\` followed by U+2028 or U+2029 in a string literal still evaluates to the empty string. This was already clear from the proposal, but updating this note as part of this patch improves accuracy.

littledan:
This proposal [reached Stage 4](https://github.com/tc39/tc39-notes/blob/master/es9/2018-05/may-22.md#11ie-ecmascript-as-a-superset-of-json) in the May 2018 TC39 meeting. Is the patch ready to land?

mathiasbynens:
Side note: this commit has just been merged, yet it doesn’t show up in the first few entries on https://github.com/tc39/ecma262/commits/master. Instead, it’s further down the list here:



Note that May 7th is the date the original PR was created.

Is this confusing to anyone else or just me? 

AFAICT until now, the commit log always followed merge order. Are we changing that now?

ljharb:
This is actually a bug in github; it sorts by (and has sorted by) commit date, not actual merge order. cc @keithamus; is this something that we could request?

(The reason this happened here is that when I rebased the PR, i used `--committer-date-is-author-date`, which avoids obscuring the original commit date)

mathiasbynens:
> (The reason this happened here is that when I rebased the PR, i used `--committer-date-is-author-date`, which avoids obscuring the original commit date)

This way, the commit order doesn’t make sense even in `git log`. Can we not do that please?

ljharb:
My `git log` locally works in proper merge order (ie, starting at the branch HEAD, and moving backwards one commit parent at a time). I'm not sure why it's working differently for you; perhaps aliases, or an outdated git version?

mathiasbynens:
You’re right, `git log` does show in proper merge order. My bad — I was confused again by the fact that a) the commit dates don’t match the merge dates for recent commits, as explained before, and b) [the commit overview](https://github.com/tc39/ecma262/commits/master) doesn’t link to the PRs for any of the recently-merged commits (another difference from how things have been done until now).

Please don’t use `--committer-date-is-author-date`, and include the PR links in the commit message when merging, just like we’ve always done for this repository.

ljharb:
I'll definitely include the PR links; that's a convention that wasn't stressed in our editors' meeting (however, if you click on any commit, the header on github shows the PR numbers, branches, and tags that it's contained in - so that link isn't strictly required).

PR #1193, opened 2018-05-12
Added 'yield' example for line terminator grammar restrictions.

### LINKED ISSUE #1192: Should the note about restricted productions include expressionless 'yield'?
In https://tc39.github.io/ecma262/#sec-rules-of-automatic-semicolon-insertion, the noted restricted productions are not complete, but notice that

> *ReturnExpression*
> &nbsp;&nbsp;&nbsp;&nbsp;`return;`

is documented, but not

> *YieldExpression*
> &nbsp;&nbsp;&nbsp;&nbsp;`yield`

I believe the idea is just to illustrate the two possible conditions where an expression may or may not be present, so I *think* `yield` should be documented the same.

image

DanielRosenwasser:
Fixes #1192.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -10975,6 +10975,7 @@
             ArrowParameters[?Yield, ?Await] [no LineTerminator here] `=&gt;` ConciseBody[?In]
 
           YieldExpression[In, Await] :
+            `yield`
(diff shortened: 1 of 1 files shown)


ljharb:
ping @DanielRosenwasser

littledan:
I think this PR is probably worth landing even though it's not exhaustive.

PR #1204, opened 2018-05-23
Editorial: separate `let [` into two tokens

RReverser:
As all the other consecutive tokens are explicitly separated into own literals, it seems beneficial to separate `let [` as well for clarity (that it doesn't special-case exact literal sequence, but like all others, matches consecutive tokens).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -16289,7 +16289,7 @@
     <h2>Syntax</h2>
     <emu-grammar type="definition">
       ExpressionStatement[Yield, Await] :
-        [lookahead &lt;! {`{`, `function`, `async` [no |LineTerminator| here] `function`, `class`, `let [`}] Expression[+In, ?Yield, ?Await] `;`
+        [lookahead &lt;! {`{`, `function`, `async` [no |LineTerminator| here] `function`, `class`, `let` `[`}] Expression[+In, ?Yield, ?Await] `;`
     </emu-grammar>
     <emu-note>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -16297,7 +16297,7 @@ <h1>Expression Statement</h1>
     <h2>Syntax</h2>
     <emu-grammar type="definition">
       ExpressionStatement[Yield, Await] :
-        [lookahead &lt;! {`{`, `function`, `async` [no |LineTerminator| here] `function`, `class`, `let [`}] Expression[+In, ?Yield, ?Await] `;`
+        [lookahead &lt;! {`{`, `function`, `async` [no |LineTerminator| here] `function`, `class`, `let` `[`}] Expression[+In, ?Yield, ?Await] `;`
```
ljharb: note the NLT requirement between `async` and `function` - might this change allow:
```js
let
[
```
where currently it must be `let [`? (grammar isn't my strong suit, so please correct me if i'm reading this wrong)

bakkot: It currently matches
```js
let
[
```
, and should continue to. (This is so that 
```js
let
[x] = [x]
```
is a variable declaration.)

RReverser: @ljharb This doesn't affect NLT requirement, it would be stated explicitly either way.

RReverser: But this kind of potential questions is exactly the reason I think it's worth fixing this construct to align with others :)

ljharb: Thanks for clarifying; this LGTM but I'm going to defer to other editors.

bmeck:
I believe this is on purpose due to lookahead by @waldemarhorwat , I could be wrong though.

michaelficarra:
This is merely an editorial error. They are two separate tokens.

Issue #1175, opened 2018-04-18
Wonky evaluation order for SuperProperty


jridgewell:
Per the spec:

```
SuperProperty: super [ Expression ]
  1. Let propertyNameReference be the result of evaluating Expression.
  2. Let propertyNameValue be ? GetValue(propertyNameReference).
  3. Let propertyKey be ? ToPropertyKey(propertyNameValue).
  ...
  5. Return ? MakeSuperPropertyReference(propertyKey, strict).

Runtime Semantics: MakeSuperPropertyReference(propertyKey, strict)
  1. Let env be GetThisEnvironment( ).
  ...
  3. Let actualThis be ? env.GetThisBinding().
  ...
```

This differs from regular `MemberExpression`:

```
MemberExpression: MemberExpression [ Expression ]
  1. Let baseReference be the result of evaluating MemberExpression.
  ...
  3. Let propertyNameReference be the result of evaluating Expression.

```

The fact that we evaluate the the inner `[ Expression ]` before evaluating the `super` feels wrong. I realize `super.property` isn't really a `MemberExpression`, but it looks just like it. This is observable with `super[super(), "prop"]`, allowing me to initialize the this binding (`GetThisEnvironment ().BindThisValue(...)`) even after I've referenced it.

Can we move the `env.GetThisBinding()` into `SuperProperty`'s `RS: Evaluation`?

As for real-world implementations of `super[super(), "prop"]`:

| Implemntation | Throws? |
| --- | --- |
| Chrome | Yes |
| Firefox | No |
| Safari | Yes |
| Edge | ~🤷‍♂️~ No |
| Babel | No, but I'm about to change it... |

ljharb:
It’d be good to know what edge does here.

jridgewell:
I'll try downloading an IE VM. In the meantime, try https://output.jsbin.com/miderik/1/quiet if you already have one.

jridgewell:
Whoo, SauceLabs to the rescue. Edge 16 doesn't throw.

allenwb:
I agree, resolving `this` first would be more consistent with normal left-to-right evaluation order.  Since this is only observable in a constructor using unlikely code pattern I don't think there will be any problem with changing it.

bakkot:
This is kind of an interesting question. On the one hand, it's inconsistent with member access. On the other, the current behavior is strictly more useful, albeit only in extremely weird cases.

[...]

PR #1207, opened 2018-05-24
Evaluate this binding first in SuperProperty

jridgewell:
Fixes https://github.com/tc39/ecma262/issues/1175 by moving the `getThisBinding()` before evaluating the computed property expression.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12501,17 +12501,21 @@
         <h1>Runtime Semantics: Evaluation</h1>
         <emu-grammar>SuperProperty : `super` `[` Expression `]`</emu-grammar>
         <emu-alg>
+          1. Let _env_ be GetThisEnvironment( ).
+          1. Let _actualThis_ be ? _env_.GetThisBinding().
           1. Let _propertyNameReference_ be the result of evaluating |Expression|.
           1. Let _propertyNameValue_ be ? GetValue(_propertyNameReference_).
           1. Let _propertyKey_ be ? ToPropertyKey(_propertyNameValue_).
           1. If the code matched by this |SuperProperty| is strict mode code, let _strict_ be *true*, else let _strict_ be *false*.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -12959,12 +12963,11 @@ <h1>Runtime Semantics: GetSuperConstructor ( )</h1>
...
       <!-- es6num="12.3.5.3" -->
       <emu-clause id="sec-makesuperpropertyreference" aoid="MakeSuperPropertyReference">
-        <h1>Runtime Semantics: MakeSuperPropertyReference ( _propertyKey_, _strict_ )</h1>
-        <p>The abstract operation MakeSuperPropertyReference with arguments _propertyKey_ and _strict_ performs the following steps:</p>
+        <h1>Runtime Semantics: MakeSuperPropertyReference ( _actualThis_, _propertyKey_, _strict_ )</h1>
+        <p>The abstract operation MakeSuperPropertyReference with arguments _actualThis_, _propertyKey_, and _strict_ performs the following steps:</p>
         <emu-alg>
           1. Let _env_ be GetThisEnvironment( ).
           1. Assert: _env_.HasSuperBinding() is *true*.
-          1. Let _actualThis_ be ? _env_.GetThisBinding().
```
ljharb: Should there be an assertion here about the value, or is that unnecessary?

jridgewell: Unnecessary, these are the only callers. Additionally, `getThisBinding` itself has an assertion.

ljharb: I was mainly thinking to ensure future callers use it properly, but yeah you’re probably right.

ljharb:
Tests PR: https://github.com/tc39/test262/pull/1572

jridgewell:
@ljharb: Those tests are actually for the pre-change behavior.

jridgewell:
Tests with the post-change behavior have been merged.

jridgewell:
Whoo, I changed JS.

PR #1210, opened 2018-05-30
Editorial: reduce indirection in NumberToString

ljharb:
`ToString` on a guaranteed Number just calls into `NumberToString`, so let's remove a conceptual stack frame from this recursion.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4088,7 +4088,7 @@
         <emu-alg>
           1. If _m_ is *NaN*, return the String `"NaN"`.
           1. If _m_ is *+0* or *-0*, return the String `"0"`.
-          1. If _m_ is less than zero, return the string-concatenation of `"-"` and ! ToString(-_m_).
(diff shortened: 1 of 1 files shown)


ljharb:
@allenwb lol yes, I know that, that's why i said "conceptual" - because in my brain, the extra step made it slightly harder to understand the algorithm.

In other words, this improvement has nothing to do with implementations, it solely has to do with making the spec easier to understand by removing wholly unnecessary indirection.

allenwb:
Note that this is actually a naming issue.  In ES1-ES2017 this algorithm did not have a distinct name, but instead was considered to be a subpart of the ToString algorithm.  See: https://ecma-international.org/ecma-262/8.0/index.html#sec-tostring-applied-to-the-number-type  So it recursed back to ToString.

Subsequent to ES 2017, the subpart was given the name  NumbrToString, but the target of the recursion was left unchanged.  I think this patch is a fine way to finish the incomplete work of the renaming.

PR #1212, opened 2018-06-04
Editorial: Don't explicitly set [[Extensible]] to true

### LINKED ISSUE #1201: Should remove unnecessary [[Extensible]] assignments
Default behavior is for it to be `true` per https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects , but lots of builtins specify it to be true. This can lead to some confusion and thinking that `true` is not the default.

rkirsling:
Resolves #1201.

Follows from:

[§17](https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects):
> Unless specified otherwise, the [[Extensible]] internal slot of a built-in object initially has the value **true**.

[§9.1](https://tc39.github.io/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots):
> Once the value of an object's [[Extensible]] internal slot has been set to **false** it may not be subsequently changed to **true**.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7324,7 +7324,6 @@
         1. Let _obj_ be a newly created object with an internal slot for each name in _internalSlotsList_.
         1. Set _obj_'s essential internal methods to the default ordinary object definitions specified in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>.
         1. Set _obj_.[[Prototype]] to _proto_.
-        1. Set _obj_.[[Extensible]] to *true*.
         1. Return _obj_.
       </emu-alg>
     </emu-clause>
@@ -7618,7 +7617,6 @@
         1. Set _F_.[[Strict]] to _strict_.
         1. Set _F_.[[FunctionKind]] to _functionKind_.
         1. Set _F_.[[Prototype]] to _functionPrototype_.
-        1. Set _F_.[[Extensible]] to *true*.
         1. Set _F_.[[Realm]] to the current Realm Record.
         1. Return _F_.
       </emu-alg>
@@ -7962,7 +7960,6 @@
         1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_. The initial value of each of those internal slots is *undefined*.
         1. Set _func_.[[Realm]] to _realm_.
         1. Set _func_.[[Prototype]] to _prototype_.
-        1. Set _func_.[[Extensible]] to *true*.
         1. Set _func_.[[ScriptOrModule]] to *null*.
         1. Return _func_.
       </emu-alg>
@@ -8072,7 +8069,6 @@
           1. If IsConstructor(_targetFunction_) is *true*, then
             1. Set _obj_.[[Construct]] as described in <emu-xref href="#sec-bound-function-exotic-objects-construct-argumentslist-newtarget"></emu-xref>.
           1. Set _obj_.[[Prototype]] to _proto_.
-          1. Set _obj_.[[Extensible]] to *true*.
           1. Set _obj_.[[BoundTargetFunction]] to _targetFunction_.
           1. Set _obj_.[[BoundThis]] to _boundThis_.
           1. Set _obj_.[[BoundArguments]] to _boundArgs_.
@@ -8129,7 +8125,6 @@
           1. Set _A_'s essential internal methods except for [[DefineOwnProperty]] to the default ordinary object definitions specified in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>.
           1. Set _A_.[[DefineOwnProperty]] as specified in <emu-xref href="#sec-array-exotic-objects-defineownproperty-p-desc"></emu-xref>.
           1. Set _A_.[[Prototype]] to _proto_.
-          1. Set _A_.[[Extensible]] to *true*.
           1. Perform ! OrdinaryDefineOwnProperty(_A_, `"length"`, PropertyDescriptor{[[Value]]: _length_, [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *false*}).
           1. Return _A_.
         </emu-alg>
@@ -8270,7 +8265,6 @@
           1. Set _S_.[[DefineOwnProperty]] as specified in <emu-xref href="#sec-string-exotic-objects-defineownproperty-p-desc"></emu-xref>.
           1. Set _S_.[[OwnPropertyKeys]] as specified in <emu-xref href="#sec-string-exotic-objects-ownpropertykeys"></emu-xref>.
           1. Set _S_.[[Prototype]] to _prototype_.
-          1. Set _S_.[[Extensible]] to *true*.
           1. Let _length_ be the number of code unit elements in _value_.
           1. Perform ! DefinePropertyOrThrow(_S_, `"length"`, PropertyDescriptor{[[Value]]: _length_, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }).
           1. Return _S_.
@@ -8443,7 +8437,6 @@
           1. Set _obj_.[[Delete]] as specified in <emu-xref href="#sec-arguments-exotic-objects-delete-p"></emu-xref>.
           1. Set the remainder of _obj_'s essential internal methods to the default ordinary object definitions specified in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>.
           1. Set _obj_.[[Prototype]] to %ObjectPrototype%.
-          1. Set _obj_.[[Extensible]] to *true*.
           1. Let _map_ be ObjectCreate(*null*).
           1. Set _obj_.[[ParameterMap]] to _map_.
           1. Let _parameterNames_ be the BoundNames of _formals_.
@@ -8649,7 +8642,6 @@
           1. Set _A_.[[Set]] as specified in <emu-xref href="#sec-integer-indexed-exotic-objects-set-p-v-receiver"></emu-xref>.
           1. Set _A_.[[OwnPropertyKeys]] as specified in <emu-xref href="#sec-integer-indexed-exotic-objects-ownpropertykeys"></emu-xref>.
           1. Set _A_.[[Prototype]] to _prototype_.
-          1. Set _A_.[[Extensible]] to *true*.
           1. Return _A_.
         </emu-alg>
       </emu-clause>


jmdyck:
Hm. [§17](https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects) is only talking about built-in objects, but this PR affects lots of objects that aren't built-in.

In fact, I think the only deleted line that *does* pertain to a built-in is the one in `CreateBuiltinFunction`. (And that one, it seems to me, was 'implementing' the default stated in §17.)

ljharb:
@jmdyck do you think this makes the patch incorrect?

rkirsling:
@jmdyck Hmm, it seems I misinterpreted the coverage of §17... I do notice that subsections of §19 set `[[Extensible]]` to `true` on the prototypes for [`Object`](https://tc39.github.io/ecma262/#sec-properties-of-the-object-prototype-object), [`Function`](https://tc39.github.io/ecma262/#sec-properties-of-the-function-prototype-object), etc. but I'm not sure whether this means we can leave `[[Extensible]]` implicit on instances. :confused: Either way, I think you're right about `CreateBuiltinFunction`.

jmdyck:
> do you think this makes the patch incorrect?

Yup.

ljharb:
@jmdyck Alright, should we revert it, or is there a way we could make a change that would preserve the benefit of not having to specify extensible true?

rkirsling:
Alright, let me step through this:

- The eight abstract functions I've modified in this patch are [`ObjectCreate`](https://tc39.github.io/ecma262/#sec-objectcreate), [`FunctionAllocate`](https://tc39.github.io/ecma262/#sec-functionallocate), [`CreateBuiltinFunction`](https://tc39.github.io/ecma262/#sec-createbuiltinfunction), [`BoundFunctionCreate`](https://tc39.github.io/ecma262/#sec-boundfunctioncreate), [`ArrayCreate`](https://tc39.github.io/ecma262/#sec-arraycreate), [`StringCreate`](https://tc39.github.io/ecma262/#sec-stringcreate), [`CreateMappedArgumentsObject`](https://tc39.github.io/ecma262/#sec-createmappedargumentsobject), and [`IntegerIndexedObjectCreate`](https://tc39.github.io/ecma262/#sec-integerindexedobjectcreate).

- Each function creates an ordinary object, a function object, or an exotic object. Since we're only concerned with `[[Extensible]]` here though, we can overlook the function / exotic bits and consider them all as ordinary objects.

- Ordinary objects are not built-ins. (_`Object` is a built-in but its instances aren't. D'oh._)

- [§9.1](https://tc39.github.io/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots) says all ordinary objects have an `[[Extensible]]` slot, but specifies no default value.

- [§6.1.7.2](https://tc39.github.io/ecma262/#sec-object-internal-methods-and-internal-slots) suggests that internal slots default to `undefined` unless otherwise specified.

If I didn't make a misstep, I think this means we need to do one of the following:
- Revert this patch.
- Add the "unless otherwise specified" text from §17 to §9.1, so it applies to ordinary objects in general.

...of which I think the latter could possibly be more convenient? :smile:

ljharb:
The latter seems good, since it's a very rare case when things are not extensible. Let's see what @bmeck and @bterlson think.

rkirsling:
One other thing I noticed while sorting through all of this:

It appears that `String` instances in the wild really aren't extensible.
```js
// V8 / JSC / SM
[true, 1, 'a', [1], { a: 1 }, /\s+/, new Date(), new Map(), new Error()]
   .map(x => { x.foo = 'foo'; return x.foo; });

> [undefined, undefined, undefined, "foo", "foo", "foo", "foo", "foo", "foo"]
```

Now, [§21.1.3](https://tc39.github.io/ecma262/#sec-properties-of-the-string-prototype-object) doesn't explicitly initialize `[[Extensible]]` on the `String` prototype, though I believe this(!) case should be optional by the clause in [§17](https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects). (_One way or another, it shouldn't be coming from the `Object` prototype, as [§6.1.7.2](https://tc39.github.io/ecma262/#sec-object-internal-methods-and-internal-slots) says internal slots aren't inherited._)

That's potentially fine, but does this mean it was a bug to have [`StringCreate`](https://tc39.github.io/ecma262/#sec-stringcreate) set `[[Extensible]]` to `true` in the first place? (_Sorry if I'm making another rookie mistake!_ 😅)

ljharb:
@rkirsling that's a string primitive; try `Object('a')`.

rkirsling:
@ljharb Rookie mistake confirmed. Thanks for that. 😅

bmeck:
+1 to unless otherwise specified.

jmdyck:
@rkirsling:

> The eight abstract functions I've modified ...

(We call them abstract *operations* to better distinguish them from functions.)

> Each function creates an ordinary object, a function object, or an exotic object. Since we're only concerned with [[Extensible]] here though, we can overlook the function / exotic bits and consider them all as ordinary objects.

I don't understand why you think we can do this. "Ordinary" and "exotic" are mutually exclusive, and "ordinary" and "function" are orthogonal.

> Ordinary objects are not built-ins.

Well, *some* are, but presumably your point is that some *aren't*, and thus aren't covered by §17. But what's more relevant is whether the ordinary objects *touched by this PR* are built-ins or not. (And again, I believe the answer is: some are, some aren't.)

IgnoredAmbience:
I don't think the original patch solves the intent of the bug it closed. I think the intent was that language of the form: "has an [[Extensible]] internal slot whose initial value is true" in sections 18 through 25 (eg, in [19.1.3](https://tc39.github.io/ecma262/#sec-properties-of-the-object-prototype-object)) should be removed in favour of the stated default of section 17.

ljharb:
There’s no rush to revert here imo; can someone come up with an alternative PR that will correct the problem, and achieve the intent of the issue?

rkirsling:
@jmdyck:

> (We call them abstract _operations_ to better distinguish them from functions.)

That makes sense, thanks. :+1:

> I don't understand why you think we can do this. "Ordinary" and "exotic" are mutually exclusive, and "ordinary" and "function" are orthogonal.

I guess the sentence I wrote wasn't self-evident, but I believe my meaning was correct.

- For `FunctionAllocate`, [§9.2](https://tc39.github.io/ecma262/#sec-ecmascript-function-objects) says: 
    > An ECMAScript function object is an ordinary object and has the same internal slots and the same internal methods as other ordinary objects.

    And the operation itself reiterates this:

    > Set F's essential internal methods to the default ordinary object definitions specified in 9.1.

- For `CreateBuiltinFunction`, [§9.3](https://tc39.github.io/ecma262/#sec-built-in-function-objects) says:
    > If a built-in function object is implemented as an exotic object it must have the ordinary object behaviour specified in 9.1. All such function exotic objects also have [[Prototype]], [[Extensible]], [[Realm]], and [[ScriptOrModule]] internal slots.

- For `BoundFunctionCreate`, the operation itself says:
    > Set obj's essential internal methods to the default ordinary object definitions specified in 9.1.

- For `StringCreate`, `CreateMappedArgumentsObject`, `IntegerIndexedObjectCreate`, [§9.4.3](https://tc39.github.io/ecma262/#sec-string-exotic-objects)-5 say that String, arguments, and Integer-Indexed exotic objects "have the same internal slots as ordinary objects".

- For `ArrayCreate`, [§9.4.2](https://tc39.github.io/ecma262/#sec-array-exotic-objects) strangely lacks any mention of internal slots, but this seems to be an oversight, because then it wouldn't have a `[[Prototype]]` either, right?

> Well, _some_ are, but presumably your point is that some _aren't_ ...

I meant "are not" to be read as an negated is-a statement, yes. This line was really just me making fun of myself for my newbie misunderstanding... 😞

@IgnoredAmbience:

> I don't think the original patch solves the intent of the bug it closed.

It is possible. I'd like to think that this §9 refactoring was worthwhile in its own right, but I can certainly drop the explicit initializations in §18-25 at the same time.

@ljharb:

> ... can someone come up with an alternative PR ...

Assuming everyone's alright with it, I can add the line to §9.1 and remove the explicit initializations from §18-25 later today.

zenparsing:
I think that the initializations of `[[Extensible]]` found in the algorithm steps of section 9 should probably remain as they are.

The change described in https://github.com/tc39/ecma262/pull/1212#issuecomment-394742429 would resolve the original issue, though.

jmdyck:
> I guess the sentence I wrote wasn't self-evident, but I believe my meaning was correct.

What you wrote was "we can overlook the function / exotic bits and consider them all as ordinary objects". I don't understand what meaning that could have that would be correct.

>  For `FunctionAllocate`, [...]

I agree that `FunctionAllocate` creates ordinary objects.

> For CreateBuiltinFunction, §9.3 says:
>
>  `If a built-in function object is implemented as an exotic object it must have the ordinary object behaviour specified in 9.1. All such function exotic objects also have [[Prototype]], [[Extensible]], [[Realm]], and [[ScriptOrModule]] internal slots.`

But in such a case, it's still not an ordinary object.

> For BoundFunctionCreate, the operation itself says:
>
> `Set obj's essential internal methods to the default ordinary object definitions specified in 9.1.`

Yes, but then it sets obj's [[Call]] to an exotic setting, so obj is not ordinary.

> For StringCreate, CreateMappedArgumentsObject, IntegerIndexedObjectCreate, §9.4.3-5 say that String, arguments, and Integer-Indexed exotic objects "have the same internal slots as ordinary objects".

Yes, they have [[Prototype]] and [[Extensible]] slots. That doesn't make them ordinary.

> For ArrayCreate, §9.4.2 strangely lacks any mention of internal slots, but this seems to be an oversight, because then it wouldn't have a [[Prototype]] either, right?

I disagree. §9.4.2 doesn't need to say that Array objects have [[Prototype]] and [[Extensible]] if ArrayCreate() explicitly sets them. Anyhow, we agree that an Array object has these slots. It's still not an ordinary object.

----

The overall point being that if we "add the "unless otherwise specified" text from §17 to §9.1, so it applies to ordinary objects in general", that text would *not* apply to the objects created by BoundFunctionCreate, ArrayCreate, StringCreate, CreateMappedArgumentsObject, IntegerIndexedObjectCreate, or (in some implementations) CreateBuiltinFunction. So that would be insufficient to restore the semantics that existed before this edit.

jmdyck:
@zenparsing: 
> I think that the initializations of [[Extensible]] found in the algorithm steps of section 9 should probably remain as they are.

I'm guessing you mean "as they *were*" (before this edit was merged), because no such initializations exist now. (Clause 9 does still have one 'Set' of an object's [[Extensible]] slot, but it's in OrdinaryPreventExtensions, so I don't think you'd call that an initialization.)

zenparsing:
@jmdyck 

> I'm guessing you mean "as they were"

That's right.

rkirsling:
@jmdyck:

AFAICT, the second paragraph of §9.1 seems to be _the_ definition of `[[Extensible]]` (just as the first would be for `[[Prototype]]`) since this has no alternate definition for objects that are not strictly ordinary. If so, then these function / exotic objects were already abiding by that definition, no? After all, they're explicitly referring back to §9.1 / "ordinary objects", as I've demonstrated.

(They needn't _be_ ordinary objects, and I'm sorry that my prior phrasings suggested inheritance; I thought it could be a helpful simplifying assumption to _view_ them as ordinary, if we're exclusively concerned with `[[Extensible]]` in the present discussion, but I evidently carried this assumption too far.)

Regardless of the Array exotic object case, this should certainly mean that the other seven would respect a "_Unless specified otherwise, the initial value of this slot is **true**._" sentence added to §9.1 (like [this](https://github.com/rkirsling/ecma262/commit/8530ba7fe6eff551d38a3109a461ef19aa47a714)).

@zenparsing:

If that's the consensus view, that's okay too.

I don't mean to express attachment to the merged change, I really just want to understand what is and is not wrong with it. 🙂 In that regard, this _did_ turn out to be a "good" first patch—I've learned a lot here!

jmdyck:
@rkirsling:

> AFAICT, the second paragraph of §9.1 seems to be _the_ definition of [[Extensible]]

Alternatively, one could argue that "the" definition of [[Extensible]] is whatever the abstract operations do with it. (Not that I think it's a fruitful argument. It seems that, at least for ordinary objects, there are two 'definitions', and we hope they agree.)

> If so, then these function / exotic objects were already abiding by that definition, no? After all, they're explicitly referring back to §9.1 / "ordinary objects", as I've demonstrated.

I think what you're saying is that, although the objects in question are exotic, their 'Table 5' essential internal methods (EIMs) are defined to be the ordinary internal methods of §9.1, and so, to that extent, they must treat the [[Extensible]] internal slot as would an ordinary object. If that is what you're saying, then:

(a) Not *all* of their EIMs are ordinary. ArrayCreate, StringCreate, CreateMappedArgumentsObject, and IntegerIndexedObjectCreate all set some of the EIMs to something exotic. In particular, they all set [[DefineOwnProperty]], which is one of the EIMs that (ordinarily) accesses [[Extensible]].

(b) Anyway, I don't think it's relevant, because the initial value of [[Extensible]] is not defined by those internal methods. That is, a requirement on some of an exotic object's internal methods to handle the [[Extensible]] slot in an ordinary way does not constitute or imply a requirement on it to *initialize* the slot in an ordinary way.

> Regardless of the Array exotic object case, this should certainly mean that the other seven would respect a "Unless specified otherwise, the initial value of this slot is true." sentence added to §9.1 (like this).

I'm fairly certain it would not mean that. That sentence would only place a requirement on ordinary objects.

rkirsling:
@jmdyck:

> That is, a requirement on some of an exotic object's internal methods to handle the [[Extensible]] slot in an ordinary way does not constitute or imply a requirement on it to _initialize_ the slot in an ordinary way.

I think this probably is the source of my confusion, thanks!
I'll make a new PR to revert this and do what @IgnoredAmbience / @zenparsing suggested then.

rkirsling:
Please allow me one more comment here before we create any follow-up issues for §9:

After a bunch more discussion and reflection, I now feel able to explain what was unclear to me.
- When an exotic object, such as 9.4.3's String object, has "the same internal slots as ordinary objects", it's clear that the interpretation of this must be found in the first two paragraphs of 9.1. So far so good.

- The surprising part as a newcomer is that those two paragraphs aren't _enforcing_ anything, but are more of a "statement of intention" that the `[[Prototype]]` and `[[Extensible]]` internal slots will uphold 6.1.7.3's EIM invariants within the algorithms that follow.

    - In particular, the following is evidently not a prohibition so much as a declaration:
        > Once the value of an object's [[Extensible]] internal slot has been set to false it may not be subsequently changed to true.

        That is, it _won't_ be changed in this way (so as not to break the `[[IsExtensible]]()` invariant).

- And so, the reason why adding an "`[[Extensible]]` defaults to `true`" stipulation to 9.1 doesn't help isn't a matter of exotic objects; rather, it's because those paragraphs don't do any work of their own. Even ordinary objects still have to realize that behavior in `ObjectCreate`, just as the default value for built-ins in §17 is realized in `CreateBuiltinFunction`.

Given that, I'm no longer sure whether there's a point to refactoring these eight "set to `true`" lines.

On the other hand, I would very much like to improve §9, to help others not fall into the same traps I did! 
- I think 9.1 should be clear that the slots are _meant to do a thing_ but that the text there isn't the one making it so. How about "will not" instead of "may not"? Should we turn this into a table?

- I think 9.4.2 should absolutely mention internal slots for Array exotic objects. No other ordinary or exotic object omits this information.

jmdyck:
> those two paragraphs aren't *enforcing* anything, but are more of a "statement of intention" that the `[[Prototype]]` and `[[Extensible]]` internal slots will uphold 6.1.7.3's EIM invariants within the algorithms that follow.

That's one way to read them. The problem then is determining, for any piece of spec prose, whether it's an actual requirement or a "statement of intention", something that merely helps you understand the pseudo-code that more precisely/explicitly states the requirement. (Conceivably, we could mark+style the two types of prose differently, but I'm not sure it would be worth the effort.)

Instead, I think some people might say that those sentences do indeed enforce something (i.e., they are normative requirements), they just happen to be redundant, given the algorithms. The problems with that are (a) possible divergence/disagreement, and (b) the temptation to delete the pseudo-code because the requirement is already stated in prose.

> And so, the reason why adding an "`[[Extensible]]` defaults to `true`" stipulation to 9.1 doesn't help isn't a matter of exotic objects; rather, it's because those paragraphs don't do any work of their own. Even ordinary objects still have to realize that behavior in `ObjectCreate`, just as the default value for built-ins in §17 is realized in `CreateBuiltinFunction`.

Something along those lines would have been my deeper objection, but it was easier to object on the basis of exotics.

> Given that, I'm no longer sure whether there's a point to refactoring these eight "set to `true`" lines.

I think it depends on the refactoring.

> I think 9.1 should be clear that the slots are *meant to do a thing* but that the text there isn't the one making it so. How about "will not" instead of "may not"?

I don't think I'd object to such a change, but I'll point out that the spec almost never uses "will not" outside of notes, so it might sound odd to some readers.

(Tangentially, I'll note that "may not" should be avoided anyway, because it can be interpreted as "must not" *or* "might not".)

> Should we turn this into a table?

Not sure what you have in mind there.

> I think 9.4.2 should absolutely mention internal slots for Array exotic objects. No other ordinary or exotic object omits this information.

No objection to adding it, but I'm curious if that would have helped you not "fall into a trap".

rkirsling:
> Not sure what you have in mind there.

I was just noticing that a number of sections (9.2, 9.4.1, 9.4.6, etc.) have "Internal Slots of ... Objects" tables. I thought this difference of presentation might aid in clarifying the text's purpose, though it's also possible it'd be overkill for just two slots.

> No objection to adding it, but I'm curious if that would have helped you not "fall into a trap".

Admittedly I don't think I can claim that. It just came up in the course of this thread and I thought it'd be a nice little cleanup item, so as to increase explicitness and consistency.

PR #1215, opened 2018-06-06
 Editorial: Built-ins needn't initialize [[Extensible]] to true

rkirsling:
Reverts #1212 and addresses #1201 properly.

[§17](https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects) gives `[[Extensible]]` a default value of `true` for built-ins. This has no impact on §9 but rather implies that we should remove the superfluous initializations in §18-25.


zenparsing:
Thanks @rkirsling !

PR #1217, opened 2018-06-07
Editorial: Introduce and use Unicode "surrogate" definitions

gibson042:
As requested by @ljharb: https://github.com/tc39/proposal-well-formed-stringify/issues/6#issuecomment-394888970


spec.html
```diff
@@ -948,7 +948,7 @@ <h1>The Boolean Type</h1>
...
...gate-code-unit">high-surrogate code unit</dfn> (or <dfn id="leading-surrogate">leading surrogate</dfn>) is a code unit with a numeric value in the inclusive range 0xD800 to 0xDBFF and a <dfn id="low-surrogate-code-unit">low-surrogate code unit</dfn> (or <dfn id="trailing-surrogate">trailing surrogate</dfn>) is a code unit with a numeric value in the inclusive range 0xDC00 to 0xDFFF. However, ECMAScript does not place any restrictions or requirements on the sequence of code units in a String value, so they may be ill-formed when interpreted as UTF-16 code unit sequences. Operations that do not interpret String contents treat them as sequences of undifferentiated 16-bit unsigned integers. The function `String.prototype.normalize` (see <emu-xref href="#sec-string.prototype.normalize"></emu-xref>) can be used to explicitly normalize a String value. `String.prototype.localeCompare` (see <emu-xref href="#sec-string.prototype.localecompare"></emu-xref>) internally normalizes String values, but no other operations implicitly normalize the strings upon which they operate. Only operations that are explicitly specified to be language or locale sensitive produce language-sensitive results.</p>
```
mathiasbynens: While we're changing this anyway, I'd suggest starting a new paragraph before:

> The function `String.prototype.normalize` [...]

Currently, it reads as if normalization has any kind of effect on surrogates (but it doesn't).

gibson042: Good point. I reorganized a bit more.

spec.html
```diff
@@ -948,22 +948,24 @@ <h1>The Boolean Type</h1>
...
-      </emu-note>
-      <p>Some operations interpret String contents as UTF-16 encoded Unicode code points. In that case the interpretation is:</p>
+      <p>ECMAScript operations that do not interpret String contents apply no further semantics.</p>
+      <p>ECMAScript operations that do interpret String values treat each element as a single UTF-16 code unit. As defined in section 3.8 of the Unicode Standard (Surrogates), a <dfn id="leading-surrogate">leading surrogate</dfn> (or more formally, a <dfn id="high-surrogate-code-unit">high-surrogate code unit</dfn>) is a code unit with a numeric value in the inclusive range 0xD800 to 0xDBFF and a <dfn id="trailing-surrogate">trailing surrogate</dfn> (or more formally, a <dfn id="low-surrogate-code-unit">low-surrogate code unit</dfn>) is a code unit with a numeric value in the inclusive range 0xDC00 to 0xDFFF. However, ECMAScript does not restrict the value of String elements or relationships between them, so String values may be ill-formed when interpreted as UTF-16 code unit sequences.</p>
```
michaelficarra: I would either add a comma after 0xDBFF or split that sentence into two.

michaelficarra: Drop "code unit sequences" at the end.

spec.html
```diff
@@ -948,22 +948,24 @@ <h1>The Boolean Type</h1>
...
+      <p>Some ECMAScript operations further interpret String contents as sequences of Unicode code points in the UTF-16 encoding form. In that case the following rules apply:</p>
```
michaelficarra: I would say

> Some ECMAScript operations further interpret String contents as sequences of Unicode code points encoded using UTF-16. In that case the following rules apply:

gibson042:
Thanks for the feedback, @michaelficarra. I've incorporated it in the latest commit.

ljharb:
Rebase done.

bterlson:
@ljharb lol I think we rebased at the same time. It's only fair that you should do this though since you probably saw whose fault all this conflictiness was ;)

ljharb:
lol no worries, it wasn't a hard one

PR #620, opened 2016-06-23
Normative: Require the latest available Unicode version instead of a fixed version number

mathiasbynens:
As of June 21st, Unicode 9.0.0 is the latest version.

**Update July 27:** This PR has been updated to refer to the latest available Unicode version rather than v9.0.0 specifically, as per the July 27 meeting.

PR #1218, opened 2018-06-07
Update Unicode property lists per Unicode v11

mathiasbynens:
https://unicode.org/versions/Unicode11.0.0/

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -30145,7 +30145,7 @@ THH:mm:ss.sss
           </emu-alg>
           <p>Implementations must support the Unicode property value names and aliases listed in <emu-xref href="#table-unicode-general-category-values"></emu-xref> and <emu-xref href="#table-unicode-script-values"></emu-xref>. To ensure interoperability, implementations must not support any other property value names or aliases.</p>
           <emu-note>
-            <p>For example, `Xpeo` and `Old_Persian` are valid `Script_Extension` values, but `xpeo` and `Old Persian` aren't.</p>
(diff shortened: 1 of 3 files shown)


mathiasbynens:
I believe this is _not_ a needs-consensus PR and as such am hoping this can be merged sooner rather than later. Rationale: the committee already reached consensus to follow the latest Unicode version back in 2016 per https://github.com/tc39/ecma262/pull/620 and https://tc39.github.io/tc39-notes/2016-07_jul-27.html#10ia-require-unicode-900.

Once this PR goes in, I’ll update Test262 and the surrounding ecosystem w.r.t. the Unicode property escapes changes:

- https://github.com/mathiasbynens/unicode-property-escapes-tests
- https://github.com/tc39/test262/tree/master/test/built-ins/RegExp/property-escapes
- https://github.com/mathiasbynens/unicode-property-aliases-ecmascript
- https://github.com/mathiasbynens/unicode-property-value-aliases-ecmascript
- https://github.com/mathiasbynens/regexpu-core
- https://github.com/mathiasbynens/regexpu
- [@babel/plugin-proposal-unicode-property-regex](https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-unicode-property-regex)

littledan:
@mathiasbynens Many normative changes follow a convention of using commit messages starting with `Normative:`. Could be useful here. I agree that this upgrade process should not generally be considered "needs-consensus", but I think it'd be good to mention in a quick status update to the committee that this upgrade is happening, and you expect it to happen in the future.

If you like upgrading Unicode things, see [this bug in ECMA-402](https://github.com/tc39/ecma402/issues/183).

PR #1226, opened 2018-06-15
Editorial: [[Extensible]] description should reference EIM invariants

rkirsling:
Addresses the first bullet at the bottom of https://github.com/tc39/ecma262/pull/1212#issuecomment-395663772:

As a newcomer, I found it confusing to determine whether or not the second paragraph of [9.1](https://tc39.github.io/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots) could be considered normative (and thus overestimated the effect of adding a sentence to this paragraph).

As it turns out, the `may not`s here all follow from the EIM invariants of [6.1.7.3](https://tc39.github.io/ecma262/#sec-invariants-of-the-essential-internal-methods), so I think the simplest way to prevent confusion for others to come might just be to point that out explicitly.


claudepache:
I don’t think it is right to invoke the [invariants of the essential methods](https://tc39.github.io/ecma262/#sec-invariants-of-the-essential-internal-methods) here, because they don’t concern about internal slots (i.e., state), but only about internal methods (i.e., behavior).

As I understand, the incriminated paragraph is descriptive, and follows from the algorithms of the internal methods given just below, not of the invariant of essential methods.

At the very least, it should not be implied that there is some *directed* logical relation (I mean, that one is the consequence of the other) between the [[Extensible]] slot value and the invariants of essential methods.

allenwb:
> At the very least, it should not be implied that there is some directed logical relation (I mean, that one is the consequence of the other) between the [[Extensible]] slot value and the invariants of essential methods.

@claudepache is absolutely correct here.  [[Extensible]] is simply a device used to specify the MOP level behavior of ordinary objects.  The essential invariants don't require or imply the existence of [[Extensible]].

rkirsling:
I understand that it's not a one-step implication, but since [[Extensible]] must behave in a way that does not cause an invariant violation and since the description of its behavior echoes the invariant text almost exactly, it seems misleading to me to restate these facts as if they were independent truths about the [[Extensible]] slot.

Is there another way we can explicitly indicate that either (1) the non-independence of this behavior or (2) the fact that this text is just describing the algorithms that follow?

I suppose one way would be to just delete the last two sentences of the paragraph, though I imagine we're viewing this recapitulation as worthwhile.

Alternatively, if it can truly be said that (2) should go without saying to the spec reader, then maybe I should just go make a PR to the "how to read the spec" guide instead.😅 I just don't want other newcomers to spend days in confusion about this like I did.

jmdyck:
Re (1): I think it's possible for this paragraph to make a valid reference to the EIM invariants. First, I think you'd need to say why this slot has any connection to the invariants, something like:

    An ordinary object's [[Extensible]] internal slot is *false* if and only if
    the object is *non-extensible* in the sense of [the invariants clause].

Then you could say something like:

    In order to maintain the invariants of that clause,
    once the object's [[Extensible]] internal slot has been set to *false*, ...

Alternatively, re (2), we could just say:

    As shown in the following algorithms,
    once an ordinary object's [[Extensible]] internal slot has been set to *false*, ...

(You could even combine them -- here's how the following algorithms behave wrt `[[Extensible]]` in order to maintain the invariants -- but that might be overkill.)

rkirsling:
Thanks @jmdyck, that's super helpful!

I gave (1) a try, but I fear it won't work due to the temporal phrasing of the 'non-extensible' definition:
> A target is _non-extensible_ if it has been observed to return false from its [[IsExtensible]] internal method, or true from its [[PreventExtensions]] internal method.

That "has been" (versus, say, "can be") is tricky—since the observation of these return values would come after setting [[Extensible]] to `false`, it seems ill-founded to define the latter in terms of the former. 😞 But if [[Prototype]] "is used for implementing inheritance", perhaps we could say something like this?
```
Every ordinary object has a Boolean-valued [[Extensible]] internal slot which is used to
implement (non-)extensibility in the sense of 6.1.7.3. Once an object's [[Extensible]]
internal slot has been set to *false*, ...
```

If not, your suggestion for (2) works for me too.

jmdyck:
> That "has been" (versus, say, "can be") is tricky—since the observation of these return values would come after setting [[Extensible]] to `false`

Urgh, yup. We could word around that, but it wouldn't be worth it.

> `Every ordinary object has a Boolean-valued [[Extensible]] internal slot which is used to
implement (non-)extensibility in the sense of 6.1.7.3.`

Sounds good to me. Also removes some redundant phrasing.

bterlson:
I like making it clear in this section that the text "falls out" of the implementation of the essential invariants, and I really like cleaning up this para into a comma separated clause, but I don't much like the verbage at the end of the first sentence. `(non-)` is somewhat confusing and `in the sense of` is too imprecise for my liking. How about a first sentence like:

> Every ordinary object has a Boolean-valued [[Extensible]] internal slot which is used to implement the extensibility-related <emu-xref href="#sec-invariants-of-the-essential-internal-methods"></emu-xref>.

allenwb:
@bterison says:
> Every ordinary object has a Boolean-valued [[Extensible]] internal slot which is used to implement Invariants of the Essential Internal Methods related to extensibility.

"used to implement"?? That sounds like you are mandating a specific implementation strategy for the required behaviors.

I'd say: "...[[Extensible]] internal slot that is used to specify extensibility related internal method behaviors of ordinary object. Those behaviors conform to the Invariants of the Essential Internal Methods."

Of course, that second sentence is redundant, because it would be a specification bug if any specified behaviors did not conform to the invariants.

jmdyck:
> "used to implement"?? That sounds like you are mandating a specific implementation strategy for the required behaviors.

(Note that, as @rkirsling  pointed out, we already say that `[[Prototype]]` "is used for implementing inheritance". In fact, that wording goes back to the first edition.)

allenwb:
> (Note that, as @rkirsling pointed out, we already say that [[Prototype]] "is used for implementing inheritance". In fact, that wording goes back to the first edition.)

doesn't mean it is correct.  Things were _much_ less precise in the 1st edition then they are now.

rkirsling:
Updated with (a slight variation of) my suggestion from the previous comment.

rkirsling:
@bterlson, @allenwb: Does the current wording meet your concerns adequately?

allenwb:
> Every ordinary object has a Boolean-valued [[Extensible]] internal slot which is used to realize the extensibility-related internal method behaviours described in \<emu-xref href="#sec-invariants-of-the-essential-internal-methods">\</emu-xref>.

I still find the new language awkward.  Invariants are exactly behaviors.They are requirements imposed upon the behavior. Also, don't really like realize.  How about:

> Every ordinary object has a Boolean-valued [[Extensible]] internal slot which is used to fulfill the object extensibility internal method invariants specified in \<emu-xref href="#sec-invariants-of-the-essential-internal-methods">\</emu-xref>.

ljharb:
@allenwb are you ok with the updated changes?

PR #1237, opened 2018-06-18
Editorial: remove useless es6num comments

ljharb:
Since these comments don't actually get converted into anchors, they serve no purpose. Let's get rid of them.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -88,13 +88,11 @@
   </p>
 </emu-intro>
 
-<!-- es6num="1" -->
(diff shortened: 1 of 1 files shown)


bterlson:
Alright alright, it's time to dispense with this legacy.

(For those wondering, these IDs were preserved in the source text only, and did not impact the rendering in any way. I used them as bookmarks in the source document since section numbers are computed rather than hard-coded making searching difficult. But they're so badly out of date and half-missing now that it's not super useful).

PR #1239, opened 2018-06-19
Editorial: "length" should be used consistently

### LINKED ISSUE #1236: Treatment of `length` is inconsistent in 9.4.2 and 9.4.3
- 9.4.2 has a long first paragraph describing the `length` property:
    > Every Array object has a `length` property ...

    Yet it also has the following sentence dangling as its own paragraph afterwards:
    > Array exotic objects always have a non-configurable property named `"length"`.

    Seems like the latter should be deleted and the word "non-configurable" added to the former.

- 9.4.3 exclusively uses `"length"` in its exposition:
     > String exotic objects always have a data property named `"length"` ... Both the code unit data properties and the `"length"` property ...

    This is locally consistent, but I think it could be changed to just `length` for consistency with 9.4.2.

-----
Note: The string `"length"` is used for good reason in the algorithmic steps; this issue is purely about the explanatory text.

rkirsling:
Resolves #1236.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3046,7 +3046,7 @@
             1. Return *undefined*.
           </emu-alg>
 
-          <p>The `length` property of an Await fulfilled function is 1.</p>
+          <p>The `"length"` property of an Await fulfilled function is 1.</p>
         </emu-clause>
 
         <emu-clause id="await-rejected">
@@ -3066,7 +3066,7 @@
             1. Return *undefined*.
           </emu-alg>
 
-          <p>The `length` property of an Await rejected function is 1.</p>
+          <p>The `"length"` property of an Await rejected function is 1.</p>
         </emu-clause>
       </emu-clause>
 
@@ -7458,7 +7458,7 @@
           1. Throw a *TypeError* exception.
         </emu-alg>
         <p>The value of the [[Extensible]] internal slot of a %ThrowTypeError% function is *false*.</p>
-        <p>The `length` property of a %ThrowTypeError% function has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>
+        <p>The `"length"` property of a %ThrowTypeError% function has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>
       </emu-clause>
     </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -7921,7 +7920,7 @@ <h1>ArraySetLength ( _A_, _Desc_ )</h1>
 
     <emu-clause id="sec-string-exotic-objects">
       <h1>String Exotic Objects</h1>
-      <p>A <em>String object</em> is an exotic object that encapsulates a String value and exposes virtual integer-indexed data properties corresponding to the individual code unit elements of the String value. String exotic objects always have a data property named `"length"` whose value is the number of code unit elements in the encapsulated String value. Both the code unit data properties and the `"length"` property are non-writable and non-configurable.</p>
```
ljharb: hm, i kind of feel like when it's a name, it should have quotes, and when it's referring to the property, it shouldn't?

rkirsling: So "property named `"length"`" versus "the `length` property", you mean?

Any consistent approach is fine by me, though I would counterargue that in `let foo = 3;` we have an "identifier named `foo`" and no strings to be found. :smile:

ljharb: Because in the property case it’s actually a string :-)

rkirsling: Right, I just mean that the verb 'named' shouldn't make a difference (_assuming I understood your concern correctly_)—I figure the two consistent approaches would be:
1. `"length"` in the algorithms and `length` in the exposition
2. `"length"` across the board

(...noting that the same ought to apply to `"prototype"`, `"constructor"`, etc.)

ljharb: Any of these consistent approaches is an improvement; let's see what the other editors think.

bmeck: I have a preference for quotes; right now we use `@@` to prefix well known symbols and the rest of properties are String values including integer indices; however, I can see how this could be confusing if you perform some lookups in algorithm by the value of a variable with name `length` if we did not distinguish `_length_` from `` `length` ``.  Since this is about consistency and not about distinguishing the values I'd lean towards @ljharb here and have it use quotes. It better represents what is going on. My main concern is actually about future key types being added and wanting to be sure we are clear about what the key types are when we are performing algorithm steps.

rkirsling: That works—"quotes everywhere" is the most trivial to maintain anyway. :smile:

rkirsling: ...on the other hand, we're pretty close to (1) already, while (2) would be a more substantial shift—there are ~70 instances of `length`, ~70 of `prototype`, etc., not to mention that this also applies to methods like `toString`. Would we really want to change our policy on all of these?

rkirsling: So I went through every instance of "length" in the spec and there are four situations:
1. length - an English word
2. _length_ - a formal parameter
3. `length` - a property, either referenced in prose (including assert text) or dot-accessed
4. `"length"` - a string literal passed as an argument (even in text like "a [[Get]] of `"length"`")

There were three outstanding cases outside of section 9, so I've added them to this PR.

If this is acceptable / helpful, I'd be happy to do similar consistency checks for other properties / methods as a follow-up...but if I'm just being a nuisance, please let me know. 😅

rkirsling:
@bterlson Are you okay with the approach here? Should I continue forward with this "manual linting" effort? :smile:

bterlson:
@rkirsling I very much appreciate the manual linting!

But I think the convention should be '\*"foo"\* property' as properties are strings or symbols and *'s [are used for ES language values](http://bterlson.github.io/ecmarkup/#emd-overview). Since there are many instances \`"string"` I'd be ok with flipping this PR to change \`length\` to \`"length"\`.

rkirsling:
@bterlson All set! Regarding the asterisks, the link you provided makes sense, but evidently there are just 8 instances of \*"something"\* (i.e. `\*"[^"\s]+"\*`) in the codebase, compared to over a thousand for \`"something"\`. I guess as a separate item from this property-quoting task, we'll need to decide whether to do away with those 8 or to make a global style change (:scream:).

jmdyck:
(Asterisks vs backticks has come up before. The problem is that pretty much anything that the spec uses to denote a "language value" would also be valid code that has the same value. So to decide between asterisk and backtick, you have to decide whether the spec is talking about a value, or talking about code with that value, which is not always clear. Probably it should go to the former more than it currently does.)

PR #1241, opened 2018-06-20
Editorial: Consistency for the term "fully populated Property Descriptor"

IgnoredAmbience:
Previously a mixture of "complete property descriptor" and "fully populated property descriptor" was used. Change to use the defined "fully populated property descriptor" term.

Also capitalise "Property Descriptor" so that definition detection (hopefully) kicks in elsewhere, to be consistent with the rest of the spec.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2570,7 +2570,7 @@
             The normal return type is either Property Descriptor or Undefined.
           </li>
           <li>
-            If the Type of the return value is Property Descriptor, the return value must be a <emu-xref href="#sec-completepropertydescriptor">complete property descriptor</emu-xref>.
+            If the Type of the return value is Property Descriptor, the return value must be a fully populated Property Descriptor.
           </li>
           <li>
             If _P_ is described as a non-configurable, non-writable own data property, all future calls to [[GetOwnProperty]] ( _P_ ) must return Property Descriptor whose [[Value]] is SameValue as _P_'s [[Value]] attribute.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -1554,7 +1554,7 @@ <h2>[[GetOwnProperty]] (_P_)</h2>
             The Type of the return value must be either Property Descriptor or Undefined.
           </li>
           <li>
-            If the Type of the return value is Property Descriptor, the return value must be a complete property descriptor (see <emu-xref href="#sec-completepropertydescriptor"></emu-xref>).
+            If the Type of the return value is Property Descriptor, the return value must be a fully populated Property Descriptor (see <emu-xref href="#sec-completepropertydescriptor"></emu-xref>).
```
ljharb: Since the section is named “complete”, should we not go with that for the wording?

IgnoredAmbience: I opted to use the term as already defined in the section defining Property Descriptors ([`sec-property-descriptor-specification-type`](https://tc39.github.io/ecma262/#sec-property-descriptor-specification-type)). Swapping all to be "complete Property Descriptors" is also ok with me, a similar number of changed lines are required.

ljharb: Fair - https://tc39.github.io/ecma262/#sec-completepropertydescriptor seems to me the stronger precedent. Let’s see what others think.

IgnoredAmbience: On the other hand, we may wish to avoid possible confusion between complete (verb) and complete (adjective).

ljharb: “completed” for the adjective case would work fine.

zenparsing: I think the link to "CompletePropertyDescriptor" in this context is a bit odd; I would expect a definition rather than an algorithm.

I like the change to "fully populated...". Can we just remove the link to the algorithm?

bakkot:
Rebased. I applied @zenparsing's suggestion to just drop the link to CompletePropertyDescriptor. And since `Property Descriptor` is a defined term, both of the changed lines now link to ["The Property Descriptor Specification Type"](https://ci.tc39.es/preview/tc39/ecma262/sha/31f5c305069d12fe46a3c979aa27f9efff9aeeba/#sec-property-descriptor-specification-type), which is the section that defines "fully populated property descriptor".

I considered adding a `dfn` for "fully populated property descriptor", but it didn't seem worth it; the link would go the same section either way, and it wouldn't get linked for "fully populated data Property Descriptor".

PR #1245, opened 2018-06-23
Editorial: Remove "Outside" wording for eval errors

littledan:
It took me a few times over the specification to understand that the
headings and introductory text with descriptions like
"when a direct eval call occurs outside of any function"
were non-normative. Here's a confusing case:

new class { constructor() { function f() { eval("super()") } f(); } }

This example will throw a SyntaxError because the enclosing
ThisEnvironment when eval runs is not a derived constructor. However,
in the ordinary English sense of "outside", I'd say that the eval
call is not outside of a constructor.

This patch removes the use of the term "outside" when describing the
distribution of this error case. It puts the error text inline in
PerformEval, where it's easier to see the relationship with where
the error condition comes from.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -23415,7 +23415,7 @@
         1. Return ? PerformEval(_x_, _callerRealm_, *false*, *false*).
       </emu-alg>
 
-      <emu-clause id="sec-performeval" aoid="PerformEval">
+      <emu-clause id="sec-performeval" aoid="PerformEval" oldids="sec-performeval-rules-outside-functions,sec-performeval-rules-outside-methods,sec-performeval-rules-outside-constructors">
         <h1>Runtime Semantics: PerformEval ( _x_, _callerRealm_, _strictCaller_, _direct_ )</h1>
         <p>The abstract operation PerformEval with arguments _x_, _callerRealm_, _strictCaller_, and _direct_ performs the following steps:</p>
         <emu-alg>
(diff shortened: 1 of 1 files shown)


littledan:
Thanks @caitp for raising the issue.

anba:
Nit: There are some unnecessary commas before "and" in the added steps. 

This will introduce the same error report ordering issue we already have in [CreateDynamicFunction](https://tc39.github.io/ecma262/#sec-createdynamicfunction) (and ought to fix at some point) to PerformEval. When following the spec *really* closely, CreateDynamicFunction requires that for example `Function("super(); 0 = 0")` always throws a ReferenceError (because of the invalid assignment `0 = 0`), because the `SuperCall` check in CreateDynamicFunction, step 23 is outside of the "Parsing and early error detection may be interweaved in an implementation-dependent manner." clause.

littledan:
@anba Do you have an idea for a different wording which doesn't use "outside", or some other way to avoid the confusion?

anba:
I'd suggest to add a normative note before (or after) the newly added steps to clarify that implementations are allowed to treat these steps as early errors and hence are allowed to interweave them with parsing the source text. 

Obligatory comment: I still hope we can remove early ReferenceErrors at some point, so we no longer have to care about issues like this one. :smile:

littledan:
What's a normative note?

anba:
What I meant is to use some sort of descriptive step in the algorithm to clarify that the additional error reporting can be interwoven with the parse of the source code. A `NOTE` wouldn't work, because they are normally only informative, therefore we somehow need to explicitly state that these notes/remarks are meant to be read as normative. 

Let's say something like:

1. The next three steps may be interweaved with parsing and early error detection in an implementation-dependent manner.
1. If _inFunction_ is *false*, and _body_ Contains |NewTarget|, throw a *SyntaxError* exception.
1. etc.

jmdyck:
Maybe a more 'structured' form, like:

    1. Perform the following substeps in an implementation-dependent order,
       possibly interleaving parsing and error detection:
        1. ... parse ...
        1. ... check for early errors ...
        1. ... check for 'ad hoc' errors ...

It makes the 'scope' of the interweaving clearer, and doesn't doesn't have the drawback of needing to update the number in "the next three steps" when you add/subtract an error-check.

(Would also work for CreateDynamicFunction.)

jmdyck:
Re CreateDynamicFunction: the proposed change does allow interleaving the detection of early errors and 'ad hoc' errors, which is enough to resolve the problem raised by @anba [above](#issuecomment-401438699). But step 20 indicates that *parsing* can also be interleaved, so it would make more sense if the "perform the following substeps" block started 3 steps earlier, to include the parsing of _P_ and _bodyText_.

And then in both spots, you can drop the sentence
`Parsing and early error detection may be interweaved in an implementation-dependent manner.`
from substeps, since it's covered by the overall step.

littledan:
@jmdyck Good catch, done.

jmdyck:
You should probably preserve the `sec-performeval-rules-outside-foo` ids, since they've been around for a couple years. I.e., add them as oldids on the PerformEval clause. (For someone following an old link, it might be more helpful to go to the corresponding step that now enforces the rule, but I don't know how the editors feel about adding anchors to steps.)

Apart from that, looks good to me.

littledan:
@jmdyck Done, thanks.

PR #1248, opened 2018-06-26
Editorial: Simplify %AsyncFromSyncIteratorPrototype% methods

MayaLekova:
Extract common steps from the following clauses:
25.1.4.2.1 %AsyncFromSyncIteratorPrototype%.next
25.1.4.2.2 %AsyncFromSyncIteratorPrototype%.return
25.1.4.2.3 %AsyncFromSyncIteratorPrototype%.throw
into new primitive AsyncFromSyncIteratorContinuation.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -36433,19 +36433,9 @@ THH:mm:ss.sss
               1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _invalidIteratorError_ &raquo;).
               1. Return _promiseCapability_.[[Promise]].
             1. Let _syncIteratorRecord_ be _O_.[[SyncIteratorRecord]].
-            1. Let _nextResult_ be IteratorNext(_syncIteratorRecord_, _value_).
-            1. IfAbruptRejectPromise(_nextResult_, _promiseCapability_).
-            1. Let _nextDone_ be IteratorComplete(_nextResult_).
-            1. IfAbruptRejectPromise(_nextDone_, _promiseCapability_).
-            1. Let _nextValue_ be IteratorValue(_nextResult_).
-            1. IfAbruptRejectPromise(_nextValue_, _promiseCapability_).
-            1. Let _valueWrapperCapability_ be ! NewPromiseCapability(%Promise%).
-            1. Perform ! Call(_valueWrapperCapability_.[[Resolve]], *undefined*, &laquo; _nextValue_ &raquo;).
-            1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-async-from-sync-iterator-value-unwrap-functions" title></emu-xref>.
-            1. Let _onFulfilled_ be CreateBuiltinFunction(_steps_, &laquo; [[Done]] &raquo;).
-            1. Set _onFulfilled_.[[Done]] to _nextDone_.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -36501,22 +36481,12 @@ <h1>%AsyncFromSyncIteratorPrototype%.throw ( _value_ )</h1>
...
-            1. Let _throwValue_ be IteratorValue(_throwResult_).
-            1. IfAbruptRejectPromise(_throwValue_, _promiseCapability_).
-            1. Let _valueWrapperCapability_ be ! NewPromiseCapability(%Promise%).
-            1. Perform ! Call(_valueWrapperCapability_.[[Resolve]], *undefined*, &laquo; _throwValue_ &raquo;).
-            1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-async-from-sync-iterator-value-unwrap-functions" title></emu-xref>.
-            1. Let _onFulfilled_ be CreateBuiltinFunction(_steps_, &laquo; [[Done]] &raquo;).
-            1. Set _onFulfilled_.[[Done]] to _throwDone_.
-            1. Perform ! PerformPromiseThen(_valueWrapperCapability_.[[Promise]], _onFulfilled_, *undefined*, _promiseCapability_).
-            1. Return _promiseCapability_.[[Promise]].
+            1. Return AsyncFromSyncIteratorContinuation(_result_, _promiseCapability_).
```
ljharb: Should all these calls be prefixed with a `?` or a `!`?

spec.html
```diff
@@ -36466,22 +36456,12 @@ <h1>%AsyncFromSyncIteratorPrototype%.return ( _value_ )</h1>
...
+            1. If Type(_result_) is not Object, then
               1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; a newly created *TypeError* object &raquo;).
               1. Return _promiseCapability_.[[Promise]].
-            1. Let _returnDone_ be IteratorComplete(_returnResult_).
-            1. IfAbruptRejectPromise(_returnDone_, _promiseCapability_).
-            1. Let _returnValue_ be IteratorValue(_returnResult_).
-            1. IfAbruptRejectPromise(_returnValue_, _promiseCapability_).
-            1. Let _valueWrapperCapability_ be ! NewPromiseCapability(%Promise%).
-            1. Perform ! Call(_valueWrapperCapability_.[[Resolve]], *undefined*, &laquo; _returnValue_ &raquo;).
-            1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-async-from-sync-iterator-value-unwrap-functions" title></emu-xref>.
```
ljharb: @domenic can you confirm that this step is fine to move into an abstract operation?

spec.html
MayaLekova: Thanks for the comment!

After some help from @mathiasbynens sounded like prefixing with "!" is the right thing to do (since the called function contains "!" calls as well).

ljharb: Seems right to me!

bterlson:
Beautiful.

jmdyck:
The invocations of `AsyncFromSyncIteratorContinuation` are prefixed by `!`, but can't it return an abrupt completion via `IfAbruptRejectPromise`?

(Also, for the clause id, the convention is ` sec-asyncfromsynciteratorcontinuation` rather than `sec-async-from-sync-iterator-continuation`.)

PR #1250, opened 2018-06-28
Normative: Reduce the number of ticks in async/await

MayaLekova:
JavaScript programmers may expect that the following
two programs are largely similar in terms of how they perform with
respect to the ECMAScript job queue (if inside of an async function):

promise.then(f);                f(await promise);

However, if `promise` is a built-in promise, then these two code fragments
will differ in the number of iterations through the job queue are taken: because
`await` always wraps a Promise with another Promise, there are three job
queue items enqueued and dequeued before calling `f` in the `await` example,
whereas there is just a single item for the `then` usage.

In discussions with JavaScript programmers, the number of job queue items
in the current semantics turns out to be surprising. For example, the difference
has become more visible in conjunction with new V8 features for Promise
visibility and debugging, which are sometimes used in Node.js.

This patch changes the semantics of await to reduce the number of
job queue turns that are taken in the common `await` Promise case by replacing
the unconditional wrapping with a call to PromiseResolve. Analogous changes
are made in async iterators.

The patch preserves key design goals of async/await:
- Job queue processing remains deterministic, including both ordering and the number of jobs enqueued (which is observable by interspersing other jobs)
- Userland Promise libraries with "then" methods ("foreign thenables") are usable within `await`, and trigger a turn of the job queue
- Non-Promises can be awaited, and this takes a turn of the native job queue (as do all usages of `await`)

Reducing the number of job queue turns also improves performance
on multiple highly optimized async/await implementations. In a draft
implementation of this proposal in V8 behind a flag [1]:
- The doxbee async/await performance benchmark [2] improved with 48%
- The fibonacci async/await performance benchmark [3] improved with 23%
- The Hapi throughput benchmark [4] improved with 50% (when async hooks are enabled) and with 20% (when async hooks are disabled)

[1] https://chromium-review.googlesource.com/c/v8/v8/+/1106977
[2] https://github.com/bmeurer/promise-performance-tests/blob/master/lib/doxbee-async.js
[3] https://github.com/bmeurer/promise-performance-tests/blob/master/lib/fibonacci-async.js
[4] https://github.com/fastify/benchmarks

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2990,22 +2990,21 @@
         <p>Algorithm steps that say</p>
 
         <emu-alg>
-          1. Let _completion_ be Await(_promise_).
+          1. Let _completion_ be Await(_value_).
         </emu-alg>
 
         <p>mean the same thing as:</p>
 
         <emu-alg>
           1. Let _asyncContext_ be the running execution context.
-          1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
-          1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, &laquo; _promise_ &raquo;).
+          1. Let _promise_ be ? PromiseResolve(&laquo; _value_ &raquo;).
           1. Let _stepsFulfilled_ be the algorithm steps defined in <emu-xref href="#await-fulfilled" title></emu-xref>.
           1. Let _onFulfilled_ be CreateBuiltinFunction(_stepsFulfilled_, &laquo; [[AsyncContext]] &raquo;).
           1. Set _onFulfilled_.[[AsyncContext]] to _asyncContext_.
           1. Let _stepsRejected_ be the algorithm steps defined in <emu-xref href="#await-rejected" title></emu-xref>.
           1. Let _onRejected_ be CreateBuiltinFunction(_stepsRejected_, &laquo; [[AsyncContext]] &raquo;).
           1. Set _onRejected_.[[AsyncContext]] to _asyncContext_.
-          1. Perform ! PerformPromiseThen(_promiseCapability_.[[Promise]], _onFulfilled_, _onRejected_).
+          1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
           1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
           1. Return.
@@ -3018,14 +3017,14 @@
           <p>Await can be combined with the `?` and `!` prefixes, so that for example</p>
 
           <emu-alg>
-            1. Let _value_ be ? Await(_promise_).
+            1. Let _result_ be ? Await(_value_).
           </emu-alg>
 
           <p>means the same thing as:</p>
 
           <emu-alg>
-            1. Let _value_ be Await(_promise_).
-            1. ReturnIfAbrupt(_value_).
+            1. Let _result_ be Await(_value_).
+            1. ReturnIfAbrupt(_result_).
           </emu-alg>
         </emu-note>
 
@@ -36909,12 +36908,11 @@ THH:mm:ss.sss
           1. IfAbruptRejectPromise(_done_, _promiseCapability_).
           1. Let _value_ be IteratorValue(_result_).
           1. IfAbruptRejectPromise(_value_, _promiseCapability_).
-          1. Let _valueWrapperCapability_ be ! NewPromiseCapability(%Promise%).
-          1. Perform ! Call(_valueWrapperCapability_.[[Resolve]], *undefined*, &laquo; _value_ &raquo;).
+          1. Let _valueWrapper_ be ? PromiseResolve(&laquo; _value_ &raquo;).
           1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-async-from-sync-iterator-value-unwrap-functions" title></emu-xref>.
           1. Let _onFulfilled_ be CreateBuiltinFunction(_steps_, &laquo; [[Done]] &raquo;).
           1. Set _onFulfilled_.[[Done]] to _done_.
-          1. Perform ! PerformPromiseThen(_valueWrapperCapability_.[[Promise]], _onFulfilled_, *undefined*, _promiseCapability_).
+          1. Perform ! PerformPromiseThen(_valueWrapper_, _onFulfilled_, *undefined*, _promiseCapability_).
           1. Return _promiseCapability_.[[Promise]].
         </emu-alg>
       </emu-clause>
@@ -37535,15 +37533,14 @@ THH:mm:ss.sss
             1. If _state_ is `"completed"`, then
               1. If _completion_.[[Type]] is ~return~, then
                 1. Set _generator_.[[AsyncGeneratorState]] to `"awaiting-return"`.
-                1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
-                1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, &laquo; _completion_.[[Value]] &raquo;).
+                1. Let _promise_ be ? PromiseResolve(&laquo; _completion_.[[Value]] &raquo;).
                 1. Let _stepsFulfilled_ be the algorithm steps defined in <emu-xref href="#async-generator-resume-next-return-processor-fulfilled" title></emu-xref>.
                 1. Let _onFulfilled_ be CreateBuiltinFunction(_stepsFulfilled_, &laquo; [[Generator]] &raquo;).
                 1. Set _onFulfilled_.[[Generator]] to _generator_.
                 1. Let _stepsRejected_ be the algorithm steps defined in <emu-xref href="#async-generator-resume-next-return-processor-rejected" title></emu-xref>.
                 1. Let _onRejected_ be CreateBuiltinFunction(_stepsRejected_, &laquo; [[Generator]] &raquo;).
                 1. Set _onRejected_.[[Generator]] to _generator_.
-                1. Perform ! PerformPromiseThen(_promiseCapability_.[[Promise]], _onFulfilled_, _onRejected_).
+                1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
                 1. Return *undefined*.
               1. Else,
                 1. Assert: _completion_.[[Type]] is ~throw~.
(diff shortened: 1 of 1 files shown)


MayaLekova:
Please note a possible caveat of the change noted by @zenparsing  in [the original commit](https://github.com/MayaLekova/ecma262/commit/f3637934640dc43ebf505bbda83786b0ba1a471b#r29488080). Any feedback on what's the importance of this issue and suggestions how it can be worked around are more than welcome!

ljharb:
I like that this change leaves `await` robust, and uses the existing PromiseResolve mechanism. I think this also relates to #1118.

spec.html
```diff
@@ -3045,15 +3045,14 @@ <h1>Await</h1>
...
+          1. Let _completion_ be Await(_value_).
         </emu-alg>
 
         <p>mean the same thing as:</p>
 
         <emu-alg>
           1. Let _asyncContext_ be the running execution context.
-          1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
-          1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, &laquo; _promise_ &raquo;).
+          1. Let _promise_ be ? PromiseResolve(&laquo; _value_ &raquo;).
```
zenparsing: `PromiseResolve` currently takes a constructor in the first argument position. I suppose that we'll need to provide it with `%Promise%`.

zenparsing:
I'm curious about what the group thinks about the issue that @MayaLekova referenced upthread: with these new semantics "then" will not be called on a native, non-subclassed promise when `await`ed (because we are doing `PerformPromiseThen` directly):

```js
async function f() {
  let promise = Promise.resolve(1);
  promise.then = function(...args) {
    print('then called');
    return Promise.prototype.then.apply(this, args);
  };
  let result = await promise;
  return result;
}

f().then(v => print(v));
```

Currently, we log "then called", but I think after the change we will not. Is this edge case a problem that we should worry about?

ljharb:
Since it’s only observable if you monkeypatch the then method, and since Promise.resolve behaves this way already, and since `await` is often described as conceptually wrapping its operand in Promise.resolve, it seems like the proper choice to me - and anyone relying on this behavior already can’t necessarily rely on it happening.

benjamn:
Will this behave differently for thenables that are not `Promise`s? Presumably `.then` would have to be called by `await` for thenable objects?

spec.html
benjamn: What are the chances we could tolerate any `Promise` subclass here? In other words, pass _value_`.constructor` as the first argument to `PromiseResolve` if _value_ `instanceof %Promise%` (pardon my hand waving; there may be a better spec notation for this), thereby giving _value_ a chance to be returned as-is by `PromiseResolve`.

I realize the `async` function can only return instances of the original, native `Promise` constructor, but it seems like a good thing for `await` to treat `Promise` subclasses the same way it treats ordinary `Promise`s. That is, unless part of the deal with subclassing `Promise` is that you're guaranteed your `.prototype.then` method will be called by `await`?

I don't have strong feelings either way, but it seems like a discussion worth having.

zenparsing: A Promise subclass would (presumably) fail the `SameValue` test on 2.b of `PromiseResolve`, and we would then fallback to creating a new promise (exactly as we currently do).

ljharb: I believe it was an intentional design decision, discussed in committee, not to support subclassed Promises in `await`.

ljharb: (Specifically, using .constructor here would mean that any object coercible value in the language would pass the test in PromiseResolve, which would break the semantics)

benjamn: So it's exactly as if `await` calls `Promise.resolve`, which [passes `this` as the constructor argument to `PromiseResolve`](https://tc39.github.io/ecma262/#sec-promise.resolve). Although `this` could be a subclass of `Promise` in arbitrary user code, it would always be `%Promise%` in an `async` function.

benjamn: > (Specifically, using .constructor here would mean that any object coercible value in the language would pass the test in PromiseResolve, which would break the semantics)

That's why we would have to perform an external check that `.constructor` is a `Promise` subclass before calling `PromiseResolve(_value_.constructor, _value_)`, or perhaps modify `PromiseResolve` to enforce the subclass relationship.

How do folks feel about introducing a single-argument `PromiseResolve` variant that does the right thing (whatever we decide that is), so that this code (i.e. `Let _promise_ be ? PromiseResolve(&laquo; _value_ &raquo;)`) works as written?

zenparsing: I would probably prefer the explicit constructor argument to `PromiseResolve` for now. And I believe it should be `%Promise%` in any case: subclasses should go through the slower but semantically more flexible path of calling their overridden "then" method from a new tick.

ljharb: I also think it's useful to keep the explicit `%Promise%`.

MayaLekova: Thanks for the discussion!
Added `%Promise%` as a parameter in a follow-up commit.

MayaLekova:
Rebased the commits on top of master.

Here are the links to the original changes, for the purpose of comment tracking:
[Version 1](https://github.com/MayaLekova/ecma262/commit/f3637934640dc43ebf505bbda83786b0ba1a471b#r29488080)
[Version 2](https://github.com/MayaLekova/ecma262/commit/320b6c289e1c82c825d31f945821cfc102641874)

tolmasky:
Am I understanding correctly that a *subclassed* Promise with a custom `then`, await PromiseSubclassInstance(), you *can* expect the custom then to be called?

For some context, I've considered this as a way to allow chaining methods to serve "double duty" as awaitable or chainable:

```javascript
const frame = await object.frame(0);
await object.frame(0).click("selector");
```

Here, frame is a custom "LazyPromiseSubclass" that does not actually commit to looking for the frame until `then` is called (either through `await`, explict `then`, or eventually though something like `Promise.all`). That way, on the second line, the work is never actually done (and immediately lost) to grab the frame.

zenparsing:
@MayaLekova This change achieved consensus at the September meeting. Thank you!

MayaLekova:
@zenparsing Thank you a lot for presenting it! How should we proceed with merging the change?

ljharb:
@MayaLekova the PR still needs test262 tests; once those are written and merged, this will be eligible for merging.

gsathya:
> How should we proceed with merging the change?

@MayaLekova Apart from what @ljharb said, we'd also need to wait until there's implementation feedback confirming that this change is web compatible.

littledan:
About web compatibility, I am not so worried about this change, given the experience with Edge and the lack of visible complaints in the upgrade from Babel to native async/await. What additional sort of evidence are you looking for, @gsathya?

hax:
I believe this change will affect order of async operations. I used to see questions about such issue, for example: https://www.zhihu.com/question/268007969 (in Chinese). But I don't think it have web compatibility issue, because old versions of Node/Chrome already use such semantic, and the older versions of Node, browsers, promise polyfills never have consistent async order, see old issue: https://github.com/promises-aplus/promises-spec/issues/183 . So I guess real world codes never rely on that.

gsathya:
> What additional sort of evidence are you looking for, @gsathya?

I don't think there's enough evidence to show that there isn't going to be a breakage. I don't consider upgrading from Babel (fast path) to native (slow path) as enough evidence that the reverse is fine too. 

When Maya implemented this in V8, we had several tests breaking inadvertently (because they relied on the ordering) which gives me some concern.

I don't see any reason to prematurely merge this into the spec and then running into web compat problems later, especially when we're so close to having two shipping implementations.

littledan:
OK, as long as the delay in merging the patch isn't blocking implementations from collecting further evidence, @gsathya's plan sounds good to me.

jugglinmike:
Should the new semantics also be applied to the return value of async functions?

For example, the program

```js
const p = Promise.resolve();
(async () => { await p; print('after:await'); })();

p
  .then(() => print('tick'))
  .then(() => print('tick'));
```

Prints "tick tick after:await" today, but soon, it will print "after:await tick tick" instead. Compare with the program

```js
const p = Promise.resolve();
(async () => p)().then(() => print('after:return'));

p
  .then(() => print('tick'))
  .then(() => print('tick'));
```

Today, that prints "tick tick after:return", and this patch doesn't change that. For consistency, ought it to print "tick after:return tick" instead? If so, would a similar change be appropriate for YieldExpressions in async generators?

hax:
@jugglinmike 

> For consistency, ought it to print "tick after:return tick" instead?

Old v8 (node 8, 9) output "after:return tick tick". 
Current v8 (node 10) output "tick tick after:return"

😂 Oh, async order always make me confused!

Kovensky:
The first program does read to me like the correct output should be `after:return tick tick`.

I don't understand, though why the second one _doesn't_ print `tick after:return tick`.

MayaLekova:
@jugglinmike Thank you for raising this question!

This PR already takes care for async functions which include the `await` statement, but doesn't change the behaviour of the [implicitly created promise](https://tc39.github.io/ecma262/#sec-async-function-definitions-EvaluateBody) (step 1) once the async function starts executing. Furthermore any promise resolving with another promise gets an [implicit tick](https://tc39.github.io/ecma262/#sec-promise-resolve-functions) (step 13) on the microtask queue, meaning promise chaining is deferred. That's where the 2 ticks come from. 

For better understanding, the line
```javascript
(async () => p)()
  .then(() => console.log('after:return'));  // 1
```
from your second example is currently equivalent to
```javascript
(new Promise(resolve => resolve(p)))
  .then(() => console.log('after:return'));  // 2
```
If we want to get rid of these 2 ticks, the line in question would become equivalent to
```javascript
(function() { return Promise.resolve(p); })()
  .then(() => console.log('after:return'));  // 3
```

Let's discuss removing each of the 2 ticks separately.

Deferring the implicit creation of the wrapper promise inside async functions in case we actually need to await on an asynchronous task (which excludes async functions without `await` or with `await`s only on resolved promises) will indeed remove the performance overhead of turning a synchronous function to asynchronous. It will though introduce the possibility to create starvation, for instance if the `await` statement is in a loop. This possibility outweighs the performance gain IMO, so it's better not to do it.

Second, if we want to remove the extra tick for chaining native, non-patched promises (introduce "eager" chaining), this will effectively change observable behaviour for applications already shipping with native promises. Think of it as changing the behaviour of line 2 from above to be equivalent to line 3, which will actually introduce further inconsistency.

As @hax already noted, Node 8 shipped with this behaviour after optimization in V8, but that resulted in a bug in the [async hooks in Node](https://github.com/nodejs/node/issues/20274).

I hope this answers the question why we have the current behaviour from the second example and also shows the downsides of the possible changes.

joshuaaguilar20:
@MayaLekova  thanks for this beautiful write up. This was a great addition to node

eddyaics (commented):
lgtm

PR #1252, opened 2018-07-03
Editorial: Update date-time format language

gibson042:
Contains non-normative improvements from [proposal-uniform-interchange-date-parsing](https://github.com/gibson042/ecma262-proposal-uniform-interchange-date-parsing).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -26697,7 +26697,7 @@
 
       <emu-clause id="sec-date-time-string-format">
         <h1>Date Time String Format</h1>
-        <p>ECMAScript defines a string interchange format for date-times based upon a simplification of the ISO 8601 Extended Format. The format is as follows: `YYYY-MM-DDTHH:mm:ss.sssZ`</p>
+        <p>ECMAScript defines a string interchange format for date-times based upon a simplification of the ISO 8601 calendar date extended format. The format is as follows: `YYYY-MM-DDTHH:mm:ss.sssZ`</p>
         <p>Where the fields are as follows:</p>
         <figure>
           <table class="lightweight-table">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -26748,19 +26748,20 @@ <h1>Date Time String Format</h1>
...
+          <p>There exists no international standard that specifies abbreviations for civil time zones like CET, EST, etc. and sometimes the same abbreviation is even used for two very different time zones. For this reason, both ISO 8601 and this format specify numeric representations of time zone offsets.</p>
         </emu-note>
 
-        <emu-clause id="sec-extended-years">
-          <h1>Extended Years</h1>
-          <p>ECMAScript requires the ability to specify 6 digit years (extended years); approximately 285,426 years, either forward or backward, from 01 January, 1970 UTC. To represent years before 0 or after 9999, ISO 8601 permits the expansion of the year representation, but only by prior agreement between the sender and the receiver. In the simplified ECMAScript format such an expanded year representation shall have 2 extra year digits and is always prefixed with a + or - sign. The year 0 is considered positive and hence prefixed with a + sign.</p>
+        <a id="sec-extended-years"></a>
```
gibson042: Is it valuable to preserve old anchors like this? And if so, am I doing it right?

jmdyck: It's valuable to preserve them, but the preferred way would be to add `oldids="sec-extended-years"` to the following `<emu-clause>` tag. (The rendering process then takes care of generating an anchor for the old id.)

spec.html
```diff
@@ -26748,19 +26748,19 @@ <h1>Date Time String Format</h1>
...
-        <p>Illegal values (out-of-bounds as well as syntax errors) in a format string means that the format string is not a valid instance of this format.</p>
+        <p>Illegal values (out-of-bounds as well as syntax errors) in a format string mean that the format string is not a valid instance of this format.</p>
         <emu-note>
-          <p>As every day both starts and ends with midnight, the two notations `00:00` and `24:00` are available to distinguish the two midnights that can be associated with one date. This means that the following two notations refer to exactly the same point in time: `1995-02-04T24:00` and `1995-02-05T00:00`</p>
+          <p>As every day both starts and ends with midnight, the two notations `00:00` and `24:00` are available to distinguish the two midnights that can be associated with one date. This means that the following two notations refer to exactly the same point in time: `1995-02-04T24:00` and `1995-02-05T00:00`. ISO 8601 allows an hour of 24 only to indicate the end of a calendar day within a time interval, but ECMAScript permits it in representations of single time points.</p>
```
bterlson: I find this addition here confusing personally. "Hour of 24" is hard to parse, it's not clear that "time interval" is not an ES thing but refers to an ISO8601 concept, and "single time points" could also be confused for an ES term. Suggestions:

* Use "an hour string of `24`" or similar
* Use "point in time" over "time points"

I'm struggling to understand the relevance of this note, though. @gibson042 can you say more about what you're trying to clarify?

gibson042: It's intended to highlight how ECMAScript is coloring outside the lines by allowing strings that ISO 8601 does not. I'll reword.

ljharb:
@apaprocki i think this is ready to land, but i'd love your review first :-)

ljharb:
@apaprocki i've just merged #1144; i'll rebase this one on top of that. Thanks!

ljharb:
@gibson042 just kidding, i combined everything into one commit, but it's not clear to me how to rebase your edit on top of #1144. Would you mind doing that rebase? :-D

gibson042:
@ljharb Rebased and added one more update (9f4da04d54a10268620a90454aea5af1a7ccebf0 ) to remove questionable use of the term "illegal".

I resisted the temptation to clarify priority of "implementation-specific heuristics" vs. "unrecognizable/out-of-bounds" `Date.parse` rules in the process, which I believe would be a normative change better handled by [proposal-uniform-interchange-date-parsing](https://github.com/gibson042/ecma262-proposal-uniform-interchange-date-parsing).

PR #1274, opened 2018-07-25
Normative: add `Object.fromEntries`

ljharb:
This PR adds `Object.fromEntries` to the spec.

It is currently stage 3; and still needs merged test262 tests, and consensus, to reach stage 4 and be merged here.

Closes https://github.com/tc39/proposal-object-from-entries/issues/21.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -24101,6 +24101,34 @@
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-object.fromentries">
+        <h1>Object.fromEntries ( iterable )</h1>
+        <p>When the `fromEntries` method is called with argument _iterable_, the following steps are taken:</p>
+        <emu-alg>
+          1. Perform ? RequireObjectCoercible(_iterable_).
+          1. Let _obj_ be ObjectCreate(%ObjectPrototype%).
+          1. Assert: _obj_ is an extensible ordinary object with no own properties.
+          1. Let _stepsDefine_ be the algorithm steps defined in <emu-xref href="#sec-create-data-property-on-object-functions">CreateDataPropertyOnObject Functions</emu-xref>.
+          1. Let _adder_ be CreateBuiltinFunction(_stepsDefine_, &laquo; &raquo;).
+          1. Return ? AddEntriesFromIterable(_obj_, _iterable_, _adder_).
+        </emu-alg>
+        <emu-note>
+          The function created for _adder_ is never directly accessible to ECMAScript code.
+        </emu-note>
+
+        <emu-clause id="sec-create-data-property-on-object-functions">
+          <h1>CreateDataPropertyOnObject Functions</h1>
+          <p>A CreateDataPropertyOnObject function is an anonymous built-in function. When a CreateDataPropertyOnObject function is called with arguments _key_ and _value_, the following steps are taken:</p>
+          <emu-alg>
+            1. Let _O_ be the *this* value.
+            1. Assert: Type(_O_) is Object.
+            1. Assert: _O_ is an extensible ordinary object.
+            1. Let _propertyKey_ be ? ToPropertyKey(_key_).
+            1. Perform ! CreateDataPropertyOrThrow(_O_, _propertyKey_, _value_).
+          <emu-alg>
+        </emu-clause>
+      </emu-clause>
+
       <emu-clause id="sec-object.entries">
         <h1>Object.entries ( _O_ )</h1>
         <p>When the `entries` function is called with argument _O_, the following steps are taken:</p>


spec.html
```diff
@@ -24069,6 +24069,34 @@ <h1>Object.defineProperty ( _O_, _P_, _Attributes_ )</h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-Object.fromEntries">
+        <h1>Object.fromEntries ( iterable )</h1>
+        <p>When the `fromEntries` method is called with argument _iterable_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _obj_ be ObjectCreate(%ObjectPrototype%).
+          1. Assert: _obj_ is an extensible ordinary object with no own properties.
+          1. If _iterable_ is *undefined* or *null*, return _obj_.
```
bakkot: This differs from the behavior we originally had, which was to throw. This behavior contradicts [this point](https://github.com/tc39/proposal-object-from-entries/blob/master/DETAILS.md#non-optionality-of-argument) in the DETAILS document.

This was a normative change included as part of [this commit](https://github.com/tc39/proposal-object-from-entries/commit/3b1c261276289a66f78cfb148b729479e1fdf540#diff-3540caefa502006d8a33cb1385720803), I'm guessing by accident. (That commit does what the Map constructor does, but we explicitly want to do something different.) I think the fix is to change this to `If _iterable_ is *undefined* or *null*, throw a TypeError exception` and move it to the first step.

spec.html
```diff
@@ -24069,6 +24069,34 @@ <h1>Object.defineProperty ( _O_, _P_, _Attributes_ )</h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-Object.fromEntries">
```
bakkot: To match the rest of the spec, I think this should be spelled `sec-object.fromentries`. Existing sections are


```
      <emu-clause id="sec-object.assign">
      <emu-clause id="sec-object.create">
      <emu-clause id="sec-object.defineproperties">
      <emu-clause id="sec-object.defineproperty">
      <emu-clause id="sec-object.entries">
      <emu-clause id="sec-object.freeze">
      <emu-clause id="sec-object.getownpropertydescriptor">
      <emu-clause id="sec-object.getownpropertydescriptors">
      <emu-clause id="sec-object.getownpropertynames">
      <emu-clause id="sec-object.getownpropertysymbols">
      <emu-clause id="sec-object.getprototypeof">
      <emu-clause id="sec-object.is">
      <emu-clause id="sec-object.isextensible">
      <emu-clause id="sec-object.isfrozen">
      <emu-clause id="sec-object.issealed">
      <emu-clause id="sec-object.keys">
      <emu-clause id="sec-object.preventextensions">
      <emu-clause id="sec-object.prototype">
      <emu-clause id="sec-object.seal">
      <emu-clause id="sec-object.setprototypeof">
      <emu-clause id="sec-object.values">
      <emu-clause id="sec-object.prototype.constructor">
      <emu-clause id="sec-object.prototype.hasownproperty">
      <emu-clause id="sec-object.prototype.isprototypeof">
      <emu-clause id="sec-object.prototype.propertyisenumerable">
      <emu-clause id="sec-object.prototype.tolocalestring">
      <emu-clause id="sec-object.prototype.tostring">
      <emu-clause id="sec-object.prototype.valueof">
```

spec.html
ljharb: Note that this same confusion was brought up and addressed [here](
https://github.com/tc39/proposal-object-from-entries/pull/20#discussion_r204927185)

bakkot: It does not have the same semantics. `GetIterator` throws if passed undefined or null (via [GetIterator](https://tc39.github.io/ecma262/#sec-getiterator) -> [GetMethod](https://tc39.github.io/ecma262/#sec-getmethod) -> [GetV](https://tc39.github.io/ecma262/#sec-getv) -> [ToObject](https://tc39.github.io/ecma262/#sec-toobject)).

This is about the case where `GetIterator` is passed `undefined`, not an object whose `Symbol.iterator` property is `undefined`.

ljharb: aha, thank you. Can you file this on the proposal repo itself, and I'll address it there before updating here?

bakkot: [Filed](https://github.com/tc39/proposal-object-from-entries/issues/22)

ljharb: https://github.com/tc39/proposal-object-from-entries/pull/23

ljharb:
@bakkot merged into the proposal; updated here

bakkot:
Tests [landed](https://github.com/tc39/test262/pull/1676).

Issue #219, opened 2015-11-25
Is Note 2 from Array.prototype.slice meaningful?


littledan:
http://tc39.github.io/ecma262/#sec-array.prototype.slice

> NOTE 2 The explicit setting of the length property of the result Array in step 11 is necessary to ensure that its value is correct in situations where the trailing elements of the result Array are not present.

The explicit setting of length is a new change vs ES5. V8 implements it, and I don't see a downside from an implementation standpoint. But I don't understand what this note means. If this happens to be creating an %Array%, then doesn't ArraySpeciesCreate already start it out at the right final length, so the trailing holes will be present already?

anba:
The explicit Put() was erroneously removed in ES5, it is present in ES3 (ES3, 15.4.4.10 Array.prototype.slice, step 19): https://bugs.ecmascript.org/show_bug.cgi?id=417

littledan:
I can see how that's needed in ES5, since it constructs an empty array. However, in ES2015, we start with ArraySpeciesCreate of the appropriate length.

anba:
I just wanted to provide the history behind the Set() call, I don't think I care whether or not it's present. But if it gets removed from Array.prototype.slice, it should also get removed from Array.prototype.splice (`Set(A, "length", actualDeleteCount, true)`).

bterlson:
Leaving aside the issue of whether the set to length is still required (which we can consider if someone wants to seriously propose that), the note is confusing in light of ES2015 semantics. I will fix it presently.

PR #1277, opened 2018-07-31
Editorial: Step 12 of Array#splice sets the length of the result Array

cscott:
Step 19 sets the length of the target array, O, but the note is
referring to the array which is returned, A, which has its length
set in step 12 (and is initialized to the correct length in step 9).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -36701,7 +36701,7 @@ THH:mm:ss.sss
         </emu-alg>
         <p>The *"length"* property of the `concat` method is *1*<sub>𝔽</sub>.</p>
         <emu-note>
-          <p>The explicit setting of the *"length"* property in step <emu-xref href="#step-array-proto-concat-set-length"></emu-xref> is necessary to ensure that its value is correct in situations where the trailing elements of the result Array are not present.</p>
+          <p>The explicit setting of the *"length"* property in step <emu-xref href="#step-array-proto-concat-set-length"></emu-xref> is necessary to ensure that its value is correct in situations where the trailing elements of Array are not present.</p>
         </emu-note>
         <emu-note>
           <p>The `concat` function is intentionally generic; it does not require that its *this* value be an Array. Therefore it can be transferred to other kinds of objects for use as a method.</p>
@@ -37666,7 +37666,7 @@ THH:mm:ss.sss
           1. Return _A_.
         </emu-alg>
         <emu-note>
-          <p>The explicit setting of the *"length"* property of the result Array in step <emu-xref href="#step-array-proto-splice-set-length"></emu-xref> was necessary in previous editions of ECMAScript to ensure that its length was correct in situations where the trailing elements of the result Array were not present. Setting *"length"* became unnecessary starting in ES2015 when the result Array was initialized to its proper length rather than an empty Array but is carried forward to preserve backward compatibility.</p>
+          <p>The explicit setting of the *"length"* property of _O_ in step <emu-xref href="#step-array-proto-splice-set-length"></emu-xref> was necessary in previous editions of ECMAScript to ensure that its length was correct in situations where its trailing elements are not present.</p>
         </emu-note>
         <emu-note>
           <p>The `splice` function is intentionally generic; it does not require that its *this* value be an Array. Therefore it can be transferred to other kinds of objects for use as a method.</p>


allenwb:
The reference to step 19 is correct and should not be changed.  There is, however, a bug in the note that originated in ES2015:

>the explicit setting of the `length` property of the **>>>>result<<<<** Array in step 19

It should be:

>the explicit setting of the `length` property of the **this** Value Array in step 19

jmdyck:
I think it would help if, rather than referring to `the result Array` or `the *this* Value Array`, the note just referred to `O` and `A`. (Note that `O` isn't even necessarily an Array, as pointed out in note 3.)

In the second sentence, where it says `the result Array was initialized to its proper length rather than an empty Array`, it must be that `the result Array` means `A`. Which seems to mean that the second sentence is talking about setting the length of `A`, whereas the first is talking about setting the length of `O`. Which is certainly not how it reads to me currently.

jmdyck:
This note and the corresponding one in `Array.prototype.slice`, were updated (and the second sentence added) in 61f3fc33 to resolve issue #219.

bterlson:
The explicit setting of the length property of <del>the result Array</del><ins>_O_</ins> in step 19 was necessary in previous editions of ECMAScript to ensure that its length was correct in situations where the trailing elements of <del>the result Array</del><ins>_O_</ins> were not present.

<ins>(parabreak)</ins>

Setting the length <ins>of _A_ in step 12</ins> became unnecessary starting in ES2015 when the result Array was initialized to its proper length rather than an empty Array but is carried forward to preserve backward compatibility.

Eh?

jmdyck:
So isn't setting the `length` of `O` still necessary?

allenwb:
The current text of the note in question is:
> The explicit setting of the length property of the result Array in step 19 was necessary in previous editions of ECMAScript to ensure that its length was correct in situations where the trailing elements of the result Array were not present. Setting length became unnecessary starting in ES2015 when the result Array was initialized to its proper length rather than an empty Array but is carried forward to preserve backward compatibility.

The only thing wrong with that paragraph, AFAIK, is the use of the term "result Array". It should be _O_ instead. Every thing in that paragraph is intended to be about step 19 and _O_ so don't get confused by "result Array" into thinking about step 12. 

FWIW, step 19 is still as necessary as it ever was. The compatibility issue is that setting the length is potentially observable via a proxy or other means such as _O_ being an array-like where 'length' is an accessor property.

jmdyck:
> Every thing in that paragraph is intended to be about step 19 and O

That was true of the original note (in ES6), but not of the current version.

Here's the second sentence:
> Setting length became unnecessary starting in ES2015 when the result Array was initialized to its proper length rather than an empty Array

This says that the array it's talking about was initialized to its proper length in ES2015 but  was initialized to an empty Array before then.

ES5.1 says:
> Let O be the result of calling ToObject passing the this value as the argument.
> Let A be a new array created as if by the expression new Array()

and ES6/ES2015 says:
> Let O be ToObject(this value).
> Let A be ArraySpeciesCreate(O, actualDeleteCount).

Thus, the second sentence is talking about `A`, not `O`.

----

> FWIW, step 19 is still as necessary as it ever was.

That's what I thought. So why does the note say it "was necessary in previous editions", suggesting that it's no longer necessary? Again, talking about a change in necessity only applies to `A`, not `O`.

Moreover, the discussion in #219 is talking about `A`: @littledan refers to  `ArraySpeciesCreate` and @anba quotes `Set(A, "length", actualDeleteCount, true)`. So it seems they're mis-interpreting the original version of the note as talking about `A`.

It looks to me like 61f3fc3 changed the note from a correct statement about `O` to a mishmash of things about `O` and `A`.

allenwb:
@jmdyck I think you're right.

What the note said in ES2015 was:
> The explicit setting of the length property of the result Array in step 24 is necessary to ensure that its value is correct in situations where its trailing elements are not present.

My original sin was saying "result Array" when I really meant _O_.  Explicitly setting it's length is necessary because _O_ is not necessarily an Array exotic object. 

I'm not sure how that got morphed into a statement about backwards compatibility, but it probably is because of the  "result Array" mistake.

The ES2015 note should be restored using _O_ and step 19.

A statement about setting the length of _A_ for backward compatibility is also incorrect.  ES5.1 does not explicitly set the length of _A_ because it knows that _A_ is an Array exotic object.  ES2015 added the step of explicitly setting _A_'s length because there is no guarantee that ArraySpeciesCreate will return an Array exotic object or that it will have preset the length of the object it does return.  

So, I think all of the backwards compatibility verbiage in the current note should be eliminated.

It's debatable whether or not a note should be added explaining why it is necessary to explicitly set the length of _A_. 

I could just as well argue to eliminate the note about _O_.  Basically, all of the generic array methods must take care to explicitly set the length of any object, being used in an array-like manner, that is not guaranteed to be an Array exotic object.  So, these two occurrences in splice aren't really special cases.  Historically, such notes where added when an explicit set of length was found to be missing or is somebody had explicitly questioned why setting the length was needed.

Every generic array method has note that identifies it as such.  Perhaps that note could be elaborated with a reminder about the need to explicitly set lengths.

ljharb:
@jmdyck are you able to update this PR?

jmdyck:
@ljharb: I wouldn't think so -- I didn't open it.

I could open a new PR (assuming I can re-learn what I figured out 6 months ago).

ljharb:
oh whoops :-) @cscott are you able to update this PR?

If not, then @jmdyck, instead of opening a new PR, if you could post a link to your branch here, I can update the PR with your commit(s)?

cscott:
Wow.  Just stumbled across this, 9 months later.  Has the text been changed/fixed in the interim?  I do have access to update this PR, but given that it was only a 2-character change to begin with then opening a new PR to address @allenwb's more substantial rewrite would probably be reasonable.

ljharb:
@cscott it'd be great to update this one and preserve the context in a single PR :-)

jmdyck:
@cscott:
>  Has the text been changed/fixed in the interim?

It's been changed (hence the branch conflicts), but only to add quotes to the `length` property (in 89e3938c). So not fixed, no.

@ljharb:
> it'd be great to update this one and preserve the context in a single PR :-)

I disagree. Given the convoluted history and misunderstandings around the text in question, it'll be hard enough to clearly resolve the situation with a fresh PR; trying to do it instead as an outgrowth of this PR will just make things muddier. (Which is partly why I didn't respond to your [earlier invitation](https://github.com/tc39/ecma262/pull/1277#issuecomment-457847237).)

ljharb:
I think the evolution of understanding is what has value.

littledan:
@ljharb It'd be great if you could document the practices you encourage on reusing/making new GitHub PRs somewhere, so we can discuss whether we want to adopt those practices as a group.

cscott:
Regardless, you can preserve content by linking the two PRs in a comment.

But I'm willing to do a force-push to put a new patch on this branch.  What I don't have time to do (at the moment) is page in my lost state and write new text that matches what Alan wanted.  If someone wants to do that and push a commit to a private repo or a new pr on this repo or send a patch to me by email or whatever -- I can then force-push the patch here if that makes folks happy.

spec.html
```diff
@@ -37666,7 +37666,7 @@ <h1>Array.prototype.splice ( _start_, _deleteCount_, ..._items_ )</h1>
           1. Return _A_.
         </emu-alg>
         <emu-note>
-          <p>The explicit setting of the *"length"* property of the result Array in step <emu-xref href="#step-array-proto-splice-set-length"></emu-xref> was necessary in previous editions of ECMAScript to ensure that its length was correct in situations where the trailing elements of the result Array were not present. Setting *"length"* became unnecessary starting in ES2015 when the result Array was initialized to its proper length rather than an empty Array but is carried forward to preserve backward compatibility.</p>
+          <p>The explicit setting of the *"length"* property of _O_ in step <emu-xref href="#step-array-proto-splice-set-length"></emu-xref> was necessary in previous editions of ECMAScript to ensure that its length was correct in situations where its trailing elements are not present.</p>
```
ljharb, line 37669:
```suggestion
```

PR #1284, opened 2018-08-10
Editorial: Correct and make readable for-in/of/await-of static semantics 

TimothyGu:
First commit:

Currently, VarScopedDeclarations is defined for

    IterationStatement :
      `for` `await` `(` `var` ForBinding `of` AssignmentExpression `)` Statement

in two different places with conflicting definitions. Remove the incorrect definition.

Second commit:

Coalesce children of `for`-`in`, `for`-`of`, and `for`-`await`-`of` statements with identical definitions for VarDeclaredNames and VarScopedDeclarations for better readability: instead of having six different algorithm blocks for each of VarDeclaredNames and VarScopedDeclarations now there are only two for each.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -17527,30 +17527,21 @@
       <emu-clause id="sec-for-in-and-for-of-statements-static-semantics-vardeclarednames">
         <h1>Static Semantics: VarDeclaredNames</h1>
         <emu-see-also-para op="VarDeclaredNames"></emu-see-also-para>
-        <emu-grammar>IterationStatement : `for` `(` LeftHandSideExpression `in` Expression `)` Statement</emu-grammar>
-        <emu-alg>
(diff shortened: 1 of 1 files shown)


TimothyGu:
> Not "several", but one RHS (`for await ( var ForBinding of ...`) has two conflicting definitions.

Indeed. Commit message fixed.

> but I'm not certain if the fix qualifies as editorial, since an implementation that happened to follow the wrong definition would now have to change their code.

I'll let the editors decide on this, but my belief is that it is editorial.

> The first commit eliminates the first of the conflicting definitions, which looks like the correct fix to me. However, the commit somewhat obscures that fix by also moving a different RHS from one group to another. It would be clearer to accomplish that move in the second commit.

Done.

> No, I don't think any RHSs are missing.

Commit message has been changed.

PR #1286, opened 2018-08-13
Normative: Use GetValue to evaluate class heritage.

### LINKED ISSUE #934: missing GetValue() of 'superclass' in 14.5.13 ClassDefinitionEvaluation
In 14.5.13 ClassDefinitionEvaluation

GetValue() is missing after the line:
   b. Let superclass be the result of evaluating ClassHeritage.

loganfsmyth:
It seems like all other cases of `"Let X be the result of evaluation SomeGrammar."` that actually use the value are followed by a call to `GetValue`, but this one is missing. I was talking through the spec with someone on StackOverflow and they noted that it was missing, and thing I realized #934 had been filed, so a PR seemed like the best next step to get the conversation going.

Fixes #934.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19731,9 +19731,9 @@
           1. Let _constructorParent_ be the intrinsic object %FunctionPrototype%.
         1. Else,
           1. Set the running execution context's LexicalEnvironment to _classScope_.
-          1. Let _superclass_ be the result of evaluating |ClassHeritage|.
+          1. Let _superclassRef_ be the result of evaluating |ClassHeritage|.
           1. Set the running execution context's LexicalEnvironment to _lex_.
-          1. ReturnIfAbrupt(_superclass_).
+          1. Let _superclass_ be ? GetValue(_superclassRef_).
           1. If _superclass_ is *null*, then
             1. Let _protoParent_ be *null*.
             1. Let _constructorParent_ be the intrinsic object %FunctionPrototype%.


zenparsing:
The result of evaluating *ClassHeritage* can be a Reference, so it does need a `GetValue`.

magnusriga:
Since `GetValue(V)` does indeed just return `V` if the value is not *Reference* type, it seems safe to include a `GetValue` there to cover all scenarios.

If *ClassHeritage* is a *CallExpression* (and returns a *Completion Record*), we are covered by this clause (https://www.ecma-international.org/ecma-262/8.0/#sec-implicit-completion-values):

spec.html
```diff
@@ -19731,7 +19731,8 @@ <h1>Runtime Semantics: ClassDefinitionEvaluation</h1>
           1. Let _constructorParent_ be the intrinsic object %FunctionPrototype%.
         1. Else,
           1. Set the running execution context's LexicalEnvironment to _classScope_.
-          1. Let _superclass_ be the result of evaluating |ClassHeritage|.
+          1. Let _superclassRef_ be the result of evaluating |ClassHeritage|.
+          1. Let _superclass_ be ? GetValue(_superclassRef_).
```
claudepache: Beware: In case of abrupt completion, the running execution context's LexicalEnvironment won’t be restored.

Fix: move this step after the next one.

spec.html
```diff
@@ -19731,7 +19731,8 @@ <h1>Runtime Semantics: ClassDefinitionEvaluation</h1>
...
           1. Set the running execution context's LexicalEnvironment to _lex_.
           1. ReturnIfAbrupt(_superclass_).
```
claudepache: With your change, this step is now void and can be removed.

spec.html
loganfsmyth: Ah great point, thanks for catching that

spec.html
```diff
@@ -19731,9 +19731,9 @@ <h1>Runtime Semantics: ClassDefinitionEvaluation</h1>
           1. Let _constructorParent_ be the intrinsic object %FunctionPrototype%.
         1. Else,
           1. Set the running execution context's LexicalEnvironment to _classScope_.
-          1. Let _superclass_ be the result of evaluating |ClassHeritage|.
+          1. Let _superclassRef_ be the result of evaluating |ClassHeritage|.
           1. Set the running execution context's LexicalEnvironment to _lex_.
-          1. ReturnIfAbrupt(_superclass_).
+          1. Let _superclass_ be ? GetValue(_superclassRef_).
```
TimothyGu: Couldn't _superclassRef_ also be an abrupt completion here?

TimothyGu: Yes, but [GetValue](https://tc39.github.io/ecma262/#sec-getvalue) handles that gracefully by having `ReturnIfAbrupt` as its first step.

PR #1288, opened 2018-08-16
Normative: Use GetMethod instead of GetV to get iterator `next`

isiahmeadows:
This is purely for consistency. Currently, only the first throws (tested in Node 10), but with this patch, the second would also throw:

```js
[] = {[Symbol.iterator]: () => ({get next() { throw "foo" }})};
[] = {[Symbol.iterator]: () => ({next: 0})};
```

In every other case, this has no effect since they all call IteratorNext before any other ECMAScript-visible behavior, and I feel this better aligns intuitively with what the spec implies.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4855,7 +4855,8 @@
           1. Otherwise, set _method_ to ? GetMethod(_obj_, @@iterator).
         1. Let _iterator_ be ? Call(_method_, _obj_).
         1. If Type(_iterator_) is not Object, throw a *TypeError* exception.
-        1. Let _nextMethod_ be ? GetV(_iterator_, `"next"`).
+        1. Let _nextMethod_ be ? GetMethod(_iterator_, `"next"`).
+        1. If _nextMethod_ is *undefined*, throw a *TypeError* exception.
         1. Let _iteratorRecord_ be Record { [[Iterator]]: _iterator_, [[NextMethod]]: _nextMethod_, [[Done]]: *false* }.
         1. Return _iteratorRecord_.
       </emu-alg>


domenic:
This is not "editorial" if it changes behavior of implementations. As such it would require testing against engines, test262 tests, and (assuming engines follow the current spec) sign-off on changing it.

isiahmeadows:
Okay. I'll fix the commit message, then. I misunderstood the distinction.

isiahmeadows:
@domenic Done.

isiahmeadows:
So Node/Chrome, Firefox, and Safari all have the same behavior. I haven't yet tested on Edge (don't have a VM ready), but it appears this would in fact be a theoretically breaking change.

ljharb:
With a file of:
```js
[] = {[Symbol.iterator]: () => ({get next() { throw new Error("foo") }})};

[] = {[Symbol.iterator]: () => ({next: 0})};
```

I get these results:
```sh
#### ch


#### v8
Error: foo

#### sm
Error: foo

#### jsc
Error: foo
```

When I comment out the first line, no engines throw.

ljharb:
@isiahmeadows the committee discussed this and had the following feedback:
 1. It'd be great to see more of a motivating use case here for the normative change
 1. before merging a breaking change like this, we'd like to see at least one implementation of it (ie, one browser) both to signal web compatibility, and to indicate implementor buy-in.

ljharb:
@isiahmeadows are you still interested in pursuing this change? if so, the two points in https://github.com/tc39/ecma262/pull/1288#issuecomment-424457388 need to be addressed - if not, this should be closed.

isiahmeadows:
@ljharb I do, but it's low priority. I'll close this for now, and I'll open a new issue and/or PR if/once I get the time to do it and address the feedback question. (I do also want to run it by V8 to get their thoughts on it.)

bakkot:
@isiahmeadows I'm going to close this per your previous comment.

Issue #1178, opened 2018-04-20
Is there a reason `ArraySpeciesCreate` (section 9.4.2.3) doesn't just delegate to `SpeciesConstructor`?


isiahmeadows:
There's very little variation in practice between the algorithm for ArraySpeciesCreate(*originalArray*, *length*) and the following steps:

1. Let *thisRealm* be the current Realm Record.
1. If *length* is **-0**, let *length* be **+0**
1. Let *defaultArray* be *thisRealm*.[[Intrinsics]].[[%Array%]].
1. Let *C* be ? SpeciesConstructor(*originalArray*, *defaultArray*).
1. Return ? Construct(*C*, *length*).

Is there a reason why the variation exists?

Specifically, this in practice means `ArraySpeciesCreate` does two things that `SpeciesConstructor` callees generally don't:

1. It ignores `Symbol.species` on functions, but not objects (bug?)
2. It defaults to the current realm's `%Array%` instead of the `this` value's `%Array%` (documented)

I accounted for 2 in the steps above, but 1 seems like a bug. Would changing it to be aligned with `SpeciesConstructor` change the behavior to become not web-compatible?

ljharb:
cc @allenwb

allenwb:
ArraySpeciesCreate was crafted to specifically support the five functions that call it.  Those are all legacy methods that have both backwards compatibility and web reality requirements and which needed to be extended (in ES6) to support Array subclassing.  The choice of realm is the web reality issue and the fall back to ArrayCreate if _originalArray_ does not have a `constructor` property is one of the backwards compatibility requirements.

> It ignores Symbol.species on functions, but not objects (bug?)

No it doesn't. Note that the test is `Type(C) is Object`.  `Type()` is not `typeof` and the Type of a function is object,  so that predicate will be true when C is a function.

Why; doesn't ArraySpeciesCreate use SpeciesConstructor? One reason is that it was probably written before SpeciesConstructor.  But, it is also probably better for operations like this that are trying to exactly duplicate legacy behavior to have all the essential pseudo code in one place.

You could probably refactor it but you will need to be very careful to exactly duplicate the sequencing of the observable behaviors and the backwards compat (with ES3/5) constraints.  I don't think it would be worth the effort or necessarily result in a clearer specification.

If it's not broken, don't fix it!

domenic:
I would suggest that the editors add a non-normative note explaining the things above, so they aren't forever locked up inside @allenwb's brain.

PR #1289, opened 2018-08-17
Normative: allow ArraySpeciesCreate to create non-arrays

bakkot:
Summary: this PR modifies the methods `Array.prototype.{concat, filter, map, slice, splice}` such that they return an instance of the same class<sup>1</sup> as the object on which they are invoked even if that object is not an array, rather than, as currently, only in the case that the object is an array (that is, was created by `Array` or a subclass thereof). 

<sup>1</sup> to be precise, of the class given by `original.constructor[Symbol.species]`, if it exists

---

`Array.prototype.map` and similar methods which create new arrays defer to `Symbol.species` to determine the constructor for the resulting object. However, unlike other usages of `Symbol.species` such as `RegExp.prototype[Symbol.split]`, the array prototype methods will only use `Symbol.species` when the object on which the method is invoked is an actual array. (I believe this may have been an oversight introduced in ES6 when specifying the magic cross-realm behavior of `Array.prototype.map` and friends.)

This means that in the following code
```js
class KindaArray { // NB: not extending `Array`, possibly because it needs to extend some other thing
  static get [Symbol.species]() {
    console.log('reached'); // not reached
    return KindaArray;
  }

  get map() {
    return [].map;
  }
}

let x = new KindaArray;
x[0] = 'foo';
x.length = 1;

let y = x.map(t => t + 'bar');
y instanceof KindaArray
```
the last line evaluates to `false`. Since Array prototype methods are explicitly intended to be usable on things which are not arrays, and `Symbol.species` is explicitly intended to allow methods which create new instances to return instances of the same class as the original instance rather the class on which the method was originally defined, this is surprising to me. The current behavior also seems not particularly useful. This is especially so because a proxy for an array is treated as an array for the purposes of these algorithms, and such a proxy might not behave like an array in any way.

This PR changes the behavior of the above code so that `reached` is printed and the last line evaluates to `true`.

Note that this change is observable even in code which does not explicitly reference `Symbol.species`: currently `Array.prototype.map.call(new Uint8Array([0,1,2]), x => x + 1)` returns an Array; with this change it would return a `Uint8Array`, just like `Uint8Array.prototype.map.call(new Uint8Array([0,1,2]), x => x + 1)`.

In practice, I don't expect this to affect much if any existing code. That people might be relying on the behavior in the previous paragraph is my only real worry, and I'm hopeful that's not the case. There might also be a slight performance regression when applying these methods to array-likes, as in the fairly common pattern of `[].slice.call(arguments, 0)`, since these will need to do two additional lookups (to resolve `arguments.constructor[Symbol.species]` to `undefined`).

See [twitter discussion](https://twitter.com/awbjs/status/1029927157668044801). Also see #1178.

cc @tabatkins @allenwb

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7868,14 +7868,13 @@
 
       <emu-clause id="sec-arrayspeciescreate" aoid="ArraySpeciesCreate">
         <h1>ArraySpeciesCreate ( _originalArray_, _length_ )</h1>
-        <p>The abstract operation ArraySpeciesCreate with arguments _originalArray_ and _length_ is used to specify the creation of a new Array object using a constructor function that is derived from _originalArray_. It performs the following steps:</p>
+        <p>The abstract operation ArraySpeciesCreate with arguments _originalArray_ and _length_ is used to specify the creation of a new object using a constructor function that is derived from _originalArray_, with special behavior when _originalArray_ is an ordinary array. It performs the following steps:</p>
         <emu-alg>
           1. Assert: _length_ is an integer Number &ge; 0.
           1. If _length_ is *-0*, set _length_ to *+0*.
           1. Let _isArray_ be ? IsArray(_originalArray_).
-          1. If _isArray_ is *false*, return ? ArrayCreate(_length_).
           1. Let _C_ be ? Get(_originalArray_, `"constructor"`).
-          1. If IsConstructor(_C_) is *true*, then
+          1. If _isArray_ is *true* and IsConstructor(_C_) is *true*, then
             1. Let _thisRealm_ be the current Realm Record.
             1. Let _realmC_ be ? GetFunctionRealm(_C_).
             1. If _thisRealm_ and _realmC_ are not the same Realm Record, then


allenwb:
Note that this revised behavior is exactly what I originally intended for ArraySpeciesCreate and the built-in methods that use it.  The original version in ES6 had a bug that wasn't discovered and and subsequent refactoring were careful to preserve that buggy behavior.

allenwb:
@bakkot 
various places
> Return a new array whose prototype is %Array%.

that should be

Return a new exotic array object whose prototype is %ArrayPrototype%

allenwb:
Rather than inferring intent from the (buggy) algorithm, let's just capture the original intent

*  In the absence of legacy compatibility behaviors and malformed object structures, ArraySpeciesCreate should do the following:
  1. Get the original array object's "constructor"
  2. Get the value the constructors @@species
  3. If the species property is missing or its value is null or undefined, return result of ArrayCreate
  4. If the value of species is not a constructor throw a TypeError
  5. Return the result of invoking the species value as a constructor.

Note that the built-in @@species property of %Array% is a get accessor that returns `this` so normally constructed subclasses of %Array% that don't over-ride @@species will produce the subclass constructor as the @@species values.

*  In the absence of legacy compatibility concerns, there would have been additional error conditions such as throwing if the original array object did not have a "constructor" property.   But instead, in all of the following legacy special cases the result of ArrayCreate is returned:

  1. The original array does not have a constructor property.
  2. The value of the original array's constructor property is not a constructor function.
  3. The original array is an array exotic object whose constructor is the %Array% object from a realm other than the current realm.

Note that prior to ES6, all places that currently use ArraySpeciesCreate did the equivalent of ArrayCreate so there were no observable property accesses. The original (buggy) structuring of the ES6 ArraySpeciesCreate was primarily about implementing the above normal and special cases in a way that avoid any redundant observable property accesses.

bakkot:
@allenwb:

> Return a new exotic array object whose prototype is %ArrayPrototype%

Whoops, thanks. Edited my original comment to avoid confusion.

> let's just capture the original intent

Thanks, this matches my understanding except for the following caveat:

> The value of the original array's constructor property is not a constructor function.

To be precise I think this should say "is an object which is not a constructor function". The algorithm still throws for, for example, `x = []; x.constructor = 0; x.map(()=>{});`.

allenwb:
@bakkot 

>To be precise I think this should say "is an object which is not a constructor function". The algorithm still throws for, for example, `x = []; x.constructor = 0; x.map(()=>{});`.

Then that's another bug in the algorithm, because the about sequence would have been valid in ES3/5 and hence needs to be preserved. 

Special case 2 should probably be: The value of the original array's constructor property is null or undefined.

The `x.constructor = 0` case should work by looking up 'constructor' on %NumberPrototype", to get %Number% and then looking up @@species on %Number% which should return undefined which should cause ArrayCreate in step 3 of the normal behavior.  In other words, for the purposes of this algorithm (actually most algorithms) constructor values that are primitive types are coerced to wrapper objects for property access purposes.

I'll update https://github.com/tc39/ecma262/pull/1289#issuecomment-413903958

bakkot:
> Then that's another bug in the algorithm, because the about sequence would have been valid in ES3/5 and hence needs to be preserved.

Since engines have been throwing for this case for a while now, I think it's probably fine to continue doing so, despite that breaking with ES5.

> The `x.constructor = 0` case should work by looking up 'constructor' on %NumberPrototype%, to get %Number% and then looking up `@@species` on %Number% which should return undefined

Uh, that looks like one too many lookups for a `constructor` property. Shouldn't this be

looking up 'constructor' on `x` to get `0` and then looking up `@@species` on %NumberPrototype% which should return undefined

?

allenwb:
@bakkot 
> Since engines have been throwing for this case for a while now...

Yes that occurred to me to, but:
1.  This is an obscure edge case.  Who knows how many things the current implementations have broken they they've never heard about because the breakage wasn't a major website.
2.. Throwing on primitive values introduces a hard to explain inconsistency into the language. Property accesses on primitive values normally do property lookups using the corresponding wrapper prototype.  Why should this be treated inconsistently as a special case.  WTF JS!

WRT "took many look ups" . Just forget about that paragraph that is trying to (poorly and unnecessarily) explain property lookup on primitive values.

Instead, I think special case 2 should be:

2. The Type of the value of the original array's constructor property is not Object.

anba:
Requiring `constructor` and `@@species` lookups for all objects will likely lead to performance regressions, for example in Speedometer about 1/3 of all ArraySpeciesCreate calls for `slice` and `splice` is performed with non-arrays (https://bugzilla.mozilla.org/show_bug.cgi?id=1376572#c5). The same bug also has some numbers showing the general overhead when performing `constructor` and `@@species` lookups.

Apart from the performance issue, it will also be necessary to validate that this change won't break any existing applications. Specifically when `Array.prototype.{concat,filter,map,slice,splice}` is called with non-array objects which already have a `@@species` property. For example `[].slice.call(new Int32Array(4))` will throw with the proposed change.

bakkot:
The committee decided not to do this, in light of potential performance and web compatibility concerns. See further discussion in #1313.

PR #1302, opened 2018-09-10
Editorial: Quote properties across the board.

rkirsling:
Follow-up to #1239.

On IRC, two ideas were raised as to a consistent, enforceable approach to quoting properties:

- @ljharb suggested always using quotes unless that property can also be considered an identifier (hence "`"prototype"` property" but "`Function` property of the global object" ).

- @jmdyck suggested using quotes just when referring to it _as_ a property (hence "`"Function"` property" but "`Function` constructor", "`"toString"` property" but "`toString` method").

Either approach is acceptable to me, but I am starting with the latter, as the diff is more conservative.

I've also included a few related commits which fix neighboring typos, add missing backticks (e.g. to Array.prototype in plaintext), and turn "value `foo`" into "value **foo**" where trivial. The last of these could be a separate PR if desired; I've included it here since it too was brought up in the last PR.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -88,10 +88,10 @@
   <p>The fifth edition of ECMAScript (published as ECMA-262 5<sup>th</sup> edition) codified de facto interpretations of the language specification that have become common among browser implementations and added support for new features that had emerged since the publication of the third edition. Such features include accessor properties, reflective creation and inspection of objects, program control of property attributes, additional array manipulation functions, support for the JSON object encoding format, and a strict mode that provides enhanced error checking and program security. The fifth edition was adopted by the Ecma General Assembly of December 2009.</p>
   <p>The fifth edition was submitted to ISO/IEC JTC 1 for adoption under the fast-track procedure, and approved as international standard ISO/IEC 16262:2011. Edition 5.1 of the ECMAScript Standard incorporated minor corrections and is the same text as ISO/IEC 16262:2011. The 5.1 Edition was adopted by the Ecma General Assembly of June 2011.</p>
   <p>Focused development of the sixth edition started in 2009, as the fifth edition was being prepared for publication. However, this was preceded by significant experimentation and language enhancement design efforts dating to the publication of the third edition in 1999. In a very real sense, the completion of the sixth edition is the culmination of a fifteen year effort. The goals for this edition included providing better support for large applications, library creation, and for use of ECMAScript as a compilation target for other languages. Some of its major enhancements included modules, class declarations, lexical block scoping, iterators and generators, promises for asynchronous programming, destructuring patterns, and proper tail calls. The ECMAScript library of built-ins was expanded to support additional data abstractions including maps, sets, and arrays of binary numeric values as well as additional support for Unicode supplemental characters in strings and regular expressions. The built-ins were also made extensible via subclassing. The sixth edition provides the foundation for regular, incremental language and library enhancements. The sixth edition was adopted by the General Assembly of June 2015.</p>
-  <p>ECMAScript 2016 was the first ECMAScript edition released under Ecma TC39's new yearly release cadence and open development process. A plain-text source document was built from the ECMAScript 2015 source document to serve as the base for further development entirely on GitHub. Over the year of this standard's development, hundreds of pull requests and issues were filed representing thousands of bug fixes, editorial fixes and other improvements. Additionally, numerous software tools were developed to aid in this effort including Ecmarkup, Ecmarkdown, and Grammarkdown. ES2016 also included support for a new exponentiation operator and adds a new method to Array.prototype called `includes`.</p>
+  <p>ECMAScript 2016 was the first ECMAScript edition released under Ecma TC39's new yearly release cadence and open development process. A plain-text source document was built from the ECMAScript 2015 source document to serve as the base for further development entirely on GitHub. Over the year of this standard's development, hundreds of pull requests and issues were filed representing thousands of bug fixes, editorial fixes and other improvements. Additionally, numerous software tools were developed to aid in this effort including Ecmarkup, Ecmarkdown, and Grammarkdown. ES2016 also included support for a new exponentiation operator and adds a new method to `Array.prototype` called `includes`.</p>
   <p>ECMAScript 2017 introduced Async Functions, Shared Memory, and Atomics along with smaller language and library enhancements, bug fixes, and editorial updates. Async functions improve the asynchronous programming experience by providing syntax for promise-returning functions. Shared Memory and Atomics introduce a new memory model that allows multi-agent programs to communicate using atomic operations that ensure a well-defined execution order even on parallel CPUs. It also included new static methods on Object: `Object.values`, `Object.entries`, and `Object.getOwnPropertyDescriptors`.</p>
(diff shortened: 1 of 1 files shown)


mathiasbynens:
> using quotes for any property that cannot also be considered an identifier (hence "`"prototype"` property" but "`Function` property of the global object" )

I don't understand this part. `Function` can be considered an identifier.

(slightly off-topic because this is not the approach you went with)

ljharb:
`Function` is an identifier, and a property of the global object, which is why it's a tricky one - so i think it can be argued either way, unfortunately.

mathiasbynens:
> unlike `prototype` which is not itself an identifier

`prototype` is an identifier.

domenic:
Everything changed in this PR is an [identifier](https://tc39.github.io/ecma262/#prod-Identifier), in fact.

rkirsling:
@mathiasbynens @domenic Hmm, do you think "bound identifier" would be adequately clear then?

It would seem that given these...
- [12.1](https://tc39.github.io/ecma262/#sec-identifiers): 
    ```
    Identifier :
        IdentifierName but not ReservedWord
    ```
- [12.3](https://tc39.github.io/ecma262/#sec-left-hand-side-expressions):
    ```
    MemberExpression[Yield, Await] :
        ...
        MemberExpression[?Yield, ?Await] . IdentifierName
        ...
    ```
...and the fact that `foo.function = 3;` is indeed legal, that to say "a property is not an identifier" shouldn't be incorrect so much as confusable (particularly if the specification of `IdentifierName` delegates to a [Unicode annex](http://unicode.org/reports/tr31/) that simply talks of "identifiers")? 🤔

But either way, it would be nicer to avoid confusing people, so please let me know. :smile:

domenic:
I'm not sure what would be clearer, as I'm not sure what distinction you're trying to capture. I gather you think some things should be quoted, and some shouldn't. But so far the justification given is that identifiers should be unquoted, and that doesn't help, because that means everything in this PR should be unquoted. So I'm not sure what you're trying to accomplish.

Separately it seems you are discussing the distinction between identifiers and identifier names, but I'm not sure I see the relevance.

spec.html
```diff
@@ -23369,7 +23369,7 @@ <h1>isNaN ( _number_ )</h1>
         1. Otherwise, return *false*.
       </emu-alg>
       <emu-note>
-        <p>A reliable way for ECMAScript code to test if a value `X` is a *NaN* is an expression of the form `X !== X`. The result will be *true* if and only if `X` is a *NaN*.</p>
+        <p>A reliable way for ECMAScript code to test if a value *X* is a *NaN* is an expression of the form `X !== X`. The result will be *true* if and only if *X* is a *NaN*.</p>
```
jmdyck: There isn't a spec value named `X`. The sentence refers to it as a value, but then uses it to construct an expression, so it seems more appropriate to treat it as code (i.e., keep the backticks).

We could also change the wording slightly:
```
    ... test if the value of `X` ...
```

rkirsling:
I mean, as Mathias noted, this discussion is rather tangential to the current state of the PR. Insofar as it's just me seeking knowledge or opinions, it could be taken to IRC instead.

I imagine the context of this PR is a bit unclear. Admittedly my [original idea](https://github.com/tc39/ecma262/pull/1239#discussion_r198345731) was rather the opposite (_to never quote property names except when actually passing a string literal_), but my goal here is nothing other to find a consistent solution that pleases our editors and some hypothetical future linting tool alike.

Thus if we're going to quote property names in general, the key questions are what to do about properties of the global object like `Function` and what to do about methods like `toString`.

bmeck:
I'd agree with @domenic that all the changes here seem to be around already valid identifier tokens. I'm not entirely sure if this increases consistency in any way. If it could be explained why these particular tokens are being distinguished and for what reason. Primordials like the `Function constructor` generally are referred to via primordial syntax like [%Function%](https://tc39.github.io/ecma262/#sec-function-constructor). Is that the distinguishing feature between these 2 categories is?

jmdyck:
> I'm not entirely sure if this increases consistency in any way.

It does. E.g., in the current spec, there are 74 occurrences of
```
`"foo"` property
```
and 133 occurrences of
```
`foo` property
```

In rkirsling's quote-all-properties branch, the respective counts are 185 and 23. (The unquoted remnant is mostly due to properties of the global object, as rkirsling raised above.)

It would certainly be possible to resolve the inconsistency in the other direction, it's just that this direction is roughly what came out of the discussion for #1239.

rkirsling:
Turns out a chunk of the "unquoted remnant" that @jmdyck mentioned was missed spots (_must've been juggling too many regexes, as I have every recollection of addressing them..._ 🤔), but indeed, the other ones are of two varieties:
- '`super` property', which of course does _not_ mean 'a property named `"super"`'
- stuff like 'the `Number.parseInt` data property', which would need to be rewritten without the dot
    (_easily done but probably not necessary, unless the editors would prefer it_)

@bterlson also suggested going ahead with converting these `"foo"` cases into **"foo"** right away, due to viewing these as ES language values—I'm happy to do so if others are not opposed to it.

rkirsling:
Rebased, rechecked, and resquashed.

rkirsling:
Any chance our current editors could review this?

(Just to reiterate, my singular goal here has been to address existing inconsistencies; if we'd like to split this up or reconsider the approach or what have you, that's all fine by me.)

rkirsling:
Resolved conflicts again.

spec.html
```diff
@@ -24630,7 +24630,7 @@ <h1>Object.prototype.toLocaleString ( [ _reserved1_ [ , _reserved2_ ] ] )</h1>
         </emu-alg>
         <p>The optional parameters to this function are not used but are intended to correspond to the parameter pattern used by ECMA-402 `toLocaleString` functions. Implementations that do not include ECMA-402 support must not use those parameter positions for other purposes.</p>
         <emu-note>
-          <p>This function provides a generic `toLocaleString` implementation for objects that have no locale-specific `toString` behaviour. `Array`, `Number`, `Date`, and `Typed Arrays` provide their own locale-sensitive `toLocaleString` methods.</p>
+          <p>This function provides a generic `toLocaleString` implementation for objects that have no locale-specific `toString` behaviour. `Array`, `Number`, `Date`, and %TypedArray% provide their own locale-sensitive `toLocaleString` methods.</p>
```
ljharb: i think technically this would be `%TypedArray_prototype%`?

rkirsling: I think 'provide' here is meant to mean "on the prototype" though? Because it certainly is `%TypedArray%.prototype.toLocaleString`, but that's also true for `Array`, `Number`, and `Date`.

ljharb: ah true, fair enough :-)

jmdyck:
> it would be really nice if we had some kind of programmatic enforcement of the formatting, to prevent inevitable future drift.

A start would be to check that you don't get any output from:
```
< spec.html grep -o '`[^`".]*` property' | grep -v '`super` property'
```
and
```
< spec.html grep -o 'the String `[^`"]*`'
```

@rkirsling: Running the first of those commands on the current branch finds
```
`name` property
```
in Function.prototype.toString, which maybe snuck in on a rebase.

rkirsling:
(Resolved conflicts.)

rkirsling:
Rebased again.

rkirsling:
Rebased following #1592.

On that note, if jettisoning the last commit here would make the PR as a whole easier to swallow, that would be fine—it is, after all, barely even the tip of a massive iceberg in itself.

rkirsling:
Rebased.

PR #1303, opened 2018-09-10
Layering: Lower module graph algorithms onto Abstract Module Record

guybedford:
This PR moves the graph algorithms of module instantiation and evaluation and their associated fields from Source Text Module record down to Abstract Module Record.

This allows new module record types like Web Assembly or Dynamic Modules to share the same graph algorithms. This enables circular references for WASM<->JS, while avoiding the need to rewrite the entire graph algorithm and module status handling for each module type.

The following fields are moved to Abstract Module Record: [[RequestedModules]], [[Status]], [[EvaluationError]], [[DFSIndex]] and [[DFSAncestorIndex]].

The Instantiate and Evaluate concrete methods of Source Text Module Record are renamed InstantiateAll and EvaluateAll and moved to Abstract Module Record as well.

Individual module records then define the concrete methods `Instantiate()` and `Evaluate()` as per-module instantiation and evaluation routines which are called once and provide the individual module type implementations, providing a convenient specification boundary for these new module records.

Initial review by @littledan.

//cc @bmeck @allenwb

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -21379,6 +21379,62 @@
             </tr>
             <tr>
               <td>
+                [[RequestedModules]]
+              </td>
+              <td>
+                List of String
+              </td>
+              <td>
+                A List of all the |ModuleSpecifier| strings used by the module represented by this record to request the importation of a module. The List is source code occurrence ordered.
+              </td>
+            </tr>
+            <tr>
+              <td>
+                [[Status]]
+              </td>
+              <td>
+                String
+              </td>
+              <td>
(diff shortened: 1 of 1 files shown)


linclark:
I did something pretty similar in #1199 (which I regrettably did not update in a timely fashion, so apologies). 

I presented about that change at the May meeting. Instead of pulling this logic into the Abstract Module Record, I had created a separate Cyclic Module Record. This was based on discussions with a few people, including @domenic and Georg, who had worked on the original factoring. The reasoning was that we still want module types that don't allow cycles to be able to participate in these graphs.

Does aligning it with that design—moving the DFS fields and methods to a Cyclic Module Record class that's a subclass of Abstract Module Record—work for you? (Note: There was also some logic in my PR to add a "subphase". That is no longer necessary.)

PR #1306, opened 2018-09-12
Normative: GetExportedNames adjustments for late-defined export names

guybedford:
This PR provides the absolute minimum changes necessary for the [Dynamic Modules proposal](https://github.com/guybedford/proposal-dynamic-modules) to be possible as a non-ECMA262 specification.

There are primarily two features:
1. Providing the namespace module as a second argument to GetExportedNames allowing the Dynamic Module Record implementation to track namespaces for late defined export population (https://github.com/guybedford/proposal-dynamic-modules#handling-namespace-and-star-exports)
2. Supporting a null return from GetExportedNames to indicate that the namespace is not yet ready to be finalized for dynamic modules and that it should be left empty and delegated to the module record subclass to handle namespace finalization. This is necessary for well-defined edge case handling (https://github.com/guybedford/proposal-dynamic-modules#uninstantiated-circular-edge-case).

This PR will hopefully be covered in more detail at the coming meeting.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -8452,17 +8452,6 @@
           </tr>
           <tr>
             <td>
-              [[HostDefined]]
-            </td>
-            <td>
-              Any, default value is *undefined*.
-            </td>
-            <td>
-              Field reserved for use by host environments that need to associate additional information with a Module Namespace Exotic Object.
-            </td>
-          </tr>
-          <tr>
-            <td>
               [[Prototype]]
             </td>
             <td>
@@ -8617,7 +8606,6 @@
           1. Otherwise,
             1. Let _sortedExports_ be a new List containing the same values as the list _exports_ where the values are ordered as if an Array of the same values had been sorted using `Array.prototype.sort` using *undefined* as _comparefn_.
             1. Set _M_.[[Exports]] to _sortedExports_.
-          1. Set _M_.[[HostDefined]] to *undefined*.
           1. Create own properties of _M_ corresponding to the definitions in <emu-xref href="#sec-module-namespace-objects"></emu-xref>.
           1. Set _module_.[[Namespace]] to _M_.
           1. Return _M_.
@@ -21436,7 +21424,7 @@
             </tr>
             <tr>
               <td>
-                GetExportedNames(_exportStarSet_, _nsModule_)
+                GetExportedNames(_exportStarSet_)
               </td>
               <td>
                 Return a list of all names that are either directly or indirectly exported from this module.
@@ -22028,12 +22016,11 @@
         </emu-clause>
 
         <emu-clause id="sec-getexportednames">
-          <h1>GetExportedNames ( _exportStarSet_, _nsModule_ ) Concrete Method</h1>
+          <h1>GetExportedNames ( _exportStarSet_ ) Concrete Method</h1>
           <p>The GetExportedNames concrete method of a Source Text Module Record implements the corresponding Module Record abstract method.</p>
           <p>It performs the following steps:</p>
           <emu-alg>
             1. Assert: _exportStarSet_ is a List of Module Records.
-            1. Assert: _nsModule_ is a Module Record.
             1. Let _module_ be this Source Text Module Record.
             1. If _exportStarSet_ contains _module_, then
               1. Assert: We've reached the starting point of an `import *` circularity.
@@ -22048,9 +22035,9 @@
               1. Append _e_.[[ExportName]] to _exportedNames_.
             1. For each ExportEntry Record _e_ in _module_.[[StarExportEntries]], do
               1. Let _requestedModule_ be ? HostResolveImportedModule(_module_, _e_.[[ModuleRequest]]).
-              1. Let _starNames_ be ? _requestedModule_.GetExportedNames(_exportStarSet_, _nsModule_).
+              1. Let _starNames_ be ? _requestedModule_.GetExportedNames(_exportStarSet_).
               1. If _starNames_ is *null*, then
-                 1. Return *null*.
+                1. Return *null*.
               1. For each element _n_ of _starNames_, do
                 1. If SameValue(_n_, `"default"`) is *false*, then
                   1. If _n_ is not an element of _exportedNames_, then
@@ -22058,7 +22045,7 @@
             1. Return _exportedNames_.
           </emu-alg>
           <emu-note>
-            <p>GetExportedNames does not filter out or throw an exception for names that have ambiguous star export bindings.</p>
+            <p>GetExportedNames does not filter out or throw an exception for names that have ambiguous star export bindings. It also doesn't throw for modules that star export from non-source text Module Records with deferred bindings.</p>
           </emu-note>
         </emu-clause>
 
@@ -22204,6 +22191,11 @@
                   1. Let _resolution_ be ? _importedModule_.ResolveExport(_in_.[[ImportName]], &laquo; &raquo;).
                   1. If _resolution_ is *null* or `"ambiguous"`, throw a *SyntaxError* exception.
                   1. Call _envRec_.CreateImportBinding(_in_.[[LocalName]], _resolution_.[[Module]], _resolution_.[[BindingName]]).
+              1. For each ExportEntry Record _e_ in _module_.[[StarExportEntries]], do
+                1. Let _requestedModule_ be ? HostResolveImportedModule(_module_, _e_.[[ModuleRequest]]).
+                1. Let _exportNames_ be ? _requestedModule_.GetExportedNames(_exportStarSet_).
+                1. If _exportNames_ is *null*, then
+                  1. Throw a *SyntaxError* exception.
               1. Let _code_ be _module_.[[ECMAScriptCode]].
               1. Let _varDeclarations_ be the VarScopedDeclarations of _code_.
               1. Let _declaredVarNames_ be a new empty List.
@@ -22225,6 +22217,9 @@
                     1. Call _envRec_.InitializeBinding(_dn_, _fo_).
               1. Return NormalCompletion(~empty~).
             </emu-alg>
+            <emu-note>
+              <p>A *null* return from GetExportedNames indicates a non-source text Module Record with deferred exports. Star exports from these modules are not supported and throw a Syntax Error.</p>
+            </emu-note>
           </emu-clause>
         </emu-clause>
 
@@ -22398,7 +22393,7 @@
           1. Assert: _module_.[[Status]] is not `"uninstantiated"`.
           1. Let _namespace_ be _module_.[[Namespace]].
           1. If _namespace_ is *undefined*, then
-            1. Let _exportedNames_ be ? _module_.GetExportedNames(&laquo; &raquo;, _module_).
+            1. Let _exportedNames_ be ? _module_.GetExportedNames(&laquo; &raquo;).
             1. If _exportedNames_ is *null*, then
               1. Set _namespace_ to ModuleNamespaceCreate(_module_, *null*).
             1. Otherwise,
--- a/spec.html
+++ b/spec.html
@@ -8444,10 +8444,10 @@
               [[Exports]]
             </td>
             <td>
-              List of String | *null*
+              List of String | Null
             </td>
             <td>
-              A List containing the String values of the exported names exposed as own properties of this object. The list is ordered as if an Array of those String values had been sorted using `Array.prototype.sort` using *undefined* as _comparefn_.
+              A List containing the String values of the exported names exposed as own properties of this object. A *null* value indicates a non-Source-Text Module Record with deferred export bindings. The list is ordered as if an Array of those String values had been sorted using `Array.prototype.sort` using *undefined* as _comparefn_.
             </td>
           </tr>
           <tr>
@@ -22218,7 +22218,7 @@
               1. Return NormalCompletion(~empty~).
             </emu-alg>
             <emu-note>
-              <p>A *null* return from GetExportedNames indicates a non-source text Module Record with deferred exports. Star exports from these modules are not supported and throw a Syntax Error.</p>
+              <p>A *null* return from GetExportedNames indicates a non-Source Text Module Record with deferred exports. Star exports from these modules are not currently supported and throw a Syntax Error. This constraint could be removed in future.</p>
             </emu-note>
           </emu-clause>
         </emu-clause>
--- a/spec.html
+++ b/spec.html
@@ -8444,10 +8444,21 @@
               [[Exports]]
             </td>
             <td>
-              List of String
+              List of String | *null*
             </td>
             <td>
               A List containing the String values of the exported names exposed as own properties of this object. The list is ordered as if an Array of those String values had been sorted using `Array.prototype.sort` using *undefined* as _comparefn_.
+            </td>
+          </tr>
+          <tr>
+            <td>
+              [[HostDefined]]
+            </td>
+            <td>
+              Any, default value is *undefined*.
+            </td>
+            <td>
+              Field reserved for use by host environments that need to associate additional information with a Module Namespace Exotic Object.
             </td>
           </tr>
           <tr>
@@ -8478,6 +8489,8 @@
         <h1>[[IsExtensible]] ( )</h1>
         <p>When the [[IsExtensible]] internal method of a module namespace exotic object _O_ is called, the following steps are taken:</p>
         <emu-alg>
+          1. If _O_.[[Exports]] is *null*, then
+            1. Return *true*.
           1. Return *false*.
         </emu-alg>
       </emu-clause>
@@ -8486,6 +8499,8 @@
         <h1>[[PreventExtensions]] ( )</h1>
         <p>When the [[PreventExtensions]] internal method of a module namespace exotic object _O_ is called, the following steps are taken:</p>
         <emu-alg>
+          1. If _O_.[[Exports]] is *null*, then
+            1. Return *false*.
           1. Return *true*.
         </emu-alg>
       </emu-clause>
@@ -8506,7 +8521,9 @@
         <h1>[[DefineOwnProperty]] ( _P_, _Desc_ )</h1>
         <p>When the [[DefineOwnProperty]] internal method of a module namespace exotic object _O_ is called with property key _P_ and Property Descriptor _Desc_, the following steps are taken:</p>
         <emu-alg>
-          1. If Type(_P_) is Symbol, return OrdinaryDefineOwnProperty(_O_, _P_, _Desc_).
+          1. If Type(_P_) is Symbol, then
+            1. Let _current_ be ? _O_.[[GetOwnProperty]](_P_).
+            1. Return ValidateAndApplyPropertyDescriptor(_O_, _P_, *false*, _Desc_, _current_).
           1. Let _current_ be ? _O_.[[GetOwnProperty]](_P_).
           1. If _current_ is *undefined*, return *false*.
           1. If IsAccessorDescriptor(_Desc_) is *true*, return *false*.
@@ -8591,12 +8608,16 @@
         <emu-alg>
           1. Assert: _module_ is a Module Record.
           1. Assert: _module_.[[Namespace]] is *undefined*.
-          1. Assert: _exports_ is a List of String values.
+          1. Assert: _exports_ is either *null*, or a List of String values.
           1. Let _M_ be a newly created object.
           1. Set _M_'s essential internal methods to the definitions specified in <emu-xref href="#sec-module-namespace-exotic-objects"></emu-xref>.
           1. Set _M_.[[Module]] to _module_.
-          1. Let _sortedExports_ be a new List containing the same values as the list _exports_ where the values are ordered as if an Array of the same values had been sorted using `Array.prototype.sort` using *undefined* as _comparefn_.
-          1. Set _M_.[[Exports]] to _sortedExports_.
+          1. If _exports_ is *null*, then
+            1. Set _M_.[[Exports]] to *null*.
+          1. Otherwise,
+            1. Let _sortedExports_ be a new List containing the same values as the list _exports_ where the values are ordered as if an Array of the same values had been sorted using `Array.prototype.sort` using *undefined* as _comparefn_.
+            1. Set _M_.[[Exports]] to _sortedExports_.
+          1. Set _M_.[[HostDefined]] to *undefined*.
           1. Create own properties of _M_ corresponding to the definitions in <emu-xref href="#sec-module-namespace-objects"></emu-xref>.
           1. Set _module_.[[Namespace]] to _M_.
           1. Return _M_.
@@ -21415,7 +21436,7 @@
             </tr>
             <tr>
               <td>
-                GetExportedNames(_exportStarSet_)
+                GetExportedNames(_exportStarSet_, _nsModule_)
               </td>
               <td>
                 Return a list of all names that are either directly or indirectly exported from this module.
@@ -22007,10 +22028,12 @@
         </emu-clause>
 
         <emu-clause id="sec-getexportednames">
-          <h1>GetExportedNames ( _exportStarSet_ ) Concrete Method</h1>
+          <h1>GetExportedNames ( _exportStarSet_, _nsModule_ ) Concrete Method</h1>
           <p>The GetExportedNames concrete method of a Source Text Module Record implements the corresponding Module Record abstract method.</p>
           <p>It performs the following steps:</p>
           <emu-alg>
+            1. Assert: _exportStarSet_ is a List of Module Records.
+            1. Assert: _nsModule_ is a Module Record.
             1. Let _module_ be this Source Text Module Record.
             1. If _exportStarSet_ contains _module_, then
               1. Assert: We've reached the starting point of an `import *` circularity.
@@ -22025,7 +22048,9 @@
               1. Append _e_.[[ExportName]] to _exportedNames_.
             1. For each ExportEntry Record _e_ in _module_.[[StarExportEntries]], do
               1. Let _requestedModule_ be ? HostResolveImportedModule(_module_, _e_.[[ModuleRequest]]).
-              1. Let _starNames_ be ? _requestedModule_.GetExportedNames(_exportStarSet_).
+              1. Let _starNames_ be ? _requestedModule_.GetExportedNames(_exportStarSet_, _nsModule_).
+              1. If _starNames_ is *null*, then
+                 1. Return *null*.
               1. For each element _n_ of _starNames_, do
                 1. If SameValue(_n_, `"default"`) is *false*, then
                   1. If _n_ is not an element of _exportedNames_, then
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -22007,7 +22007,7 @@ <h1>ParseModule ( _sourceText_, _realm_, _hostDefined_ )</h1>
         </emu-clause>
 
         <emu-clause id="sec-getexportednames">
-          <h1>GetExportedNames ( _exportStarSet_ ) Concrete Method</h1>
+          <h1>GetExportedNames ( _exportStarSet_, _nsModule_ ) Concrete Method</h1>
```
ljharb: It’d be nice if there was some sort of assertion on the type of the arguments here - for both exportStarSet and nsModule.

bmeck:
Since this is purely for downstream specifications and doesn't affects any mechanisms of ES Modules, shouldn't this be marked editorial? No behavior changes for anything in the JS semantics, just some extra data is being exposed for downstream.

ljharb:
Requiring extra data to be exposed imo is a normative requirement.

spec.html
guybedford: Sure, I can add that.

guybedford: Added.

allenwb:
> Requiring extra data to be exposed imo is a normative requirement.

How can that possibly be the case?  A conforming ECMAScript implementation is not required to support anything other than ECMAScript Source Text Modules and isn't required to provide a mechanism to for extending itself with other kinds of modules or to expose any data to support such a mechanism.

The whole abstract module hierarchy is there only to informatively say other implementation defined module types could exist and if they exist here is the data they must expose to interoperate with Source Text Modules.

ljharb:
Ok, then that’s an interface contract - and updating a contract is a normative change.

spec.html
zenparsing: Should `nsModule` be passed into this function for the base (non-recursive) case?

guybedford: The way Dynamic Modules are written, both cases are treated the same - the module tracks its own namespace equally to any other namespace which reflects its own export names, there are no conditionals in this handling at all.

allenwb:
Ok, just looked at the Dynamic Modules Proposal readme for the first time...

Here is my fundamental concern about the change to GetExportedNames.  Some significant though went into the original versions of these algorithms to make sure that (at least for source text modules) module linking could be done ahead of time by a compiler or static linker.  On implication of that is that there can't be any link-time dependencies upon ECMAScript runtime entities/state.

Namespace exotic objects are runtime entities.  They don't exist statically outside of a running ES engine. If _nsModule_ is a reference to a namespace object then by making it a parameter to GetExportedNames (as defined by AbstractModuleRecords) you are introducing a runtime state dependency into the linking algorithms.  However, just from the read me, I'm not sure whether or not that is what _nsModule_ is actually a reference to, so I could be wrong.

guybedford:
@allenwb thanks for taking a look. I previously called this `exportStarModule`. It is not the namespace exotic runtime object, but rather a module record whose namespace exotic object exposes the named exports of a dynamic module, which is needed for tracking purposes. Great care is taken to ensure that from a user-observability perspective, exported names behaviours remain fully well-defined, and this is what a lot of the spec is about.

spec.html
zenparsing: In https://guybedford.github.io/proposal-dynamic-modules/#sec-getmodulenamespace, `module` is passed to GetExportedNames. Did you intend for the same change here?

allenwb:
@guybedford Some of this may come down to naming conventions. Names ending in Object are runtime entities that are reified to ECMAScript user land.  Names ending  in Record are state holders that aren't directly reified but whether they represent runtime state or static (ie, available outside runtime) state depends upon careful consideration of the nature of the state. For most of the language, the distinction between "Static Semantics" and "Runtime Semantics" is sufficient because the static semantics state is generally derivable by source code inspection of a _single_ script/module. But module loading is in the middle ground because  it occurs (at least in browsers) during runtime but is intended to  primarily deal with static analyzable relationships between _multiple_ perhaps dynamically acquired _modules_ . It's still fundamentally about relationships that are fixed during module linking, even if the module was dynamically acquired or even dynamically constructed immediately before module linking.

Ideally naming choices should help spec. readers/implementors better understand this distinction.

spec.html
guybedford: Yes, thanks for spotting that!

littledan:
Thanks for writing this up and for the reviews here. I'd classify this as a "Layering:" change, as it doesn't change the runtime semantics of anything now, but does add hooks for certain embedders. I'm looking forward to discussing this patch at TC39--getting it down to the minimal change that would need to be layered into ECMAScript really makes it easier for the committee to review whether we are comfortable with the change.

guybedford:
Further to the last meeting where it was determined that implementation feedback was needed here, I went ahead and implemented the Dynamic Modules Spec (https://github.com/nodejs/dynamic-modules), which depends on this and it all worked out exactly as expected here, with the spec changes as in this PR.

Complete implementation at https://chromium-review.googlesource.com/c/v8/v8/+/1303725.

guybedford:
I've updated this PR to reflect the latest directions of the dynamic modules spec where instead of throwing an error when there are cycles with unexecuted dynamic module namespaces (which may be very hard to debug) instead the namespace is just treated as uninitialized through a null return here.

spec.html
```diff
@@ -8450,6 +8450,17 @@ <h1>Module Namespace Exotic Objects</h1>
               A List containing the String values of the exported names exposed as own properties of this object. The list is ordered as if an Array of those String values had been sorted using `Array.prototype.sort` using *undefined* as _comparefn_.
             </td>
           </tr>
+          <tr>
+            <td>
+              [[HostDefined]]
+            </td>
+            <td>
+              Any, default value is *undefined*.
```
ljharb: The typical term used here is `Any ECMAScript language type`, i believe

spec.html
```diff
@@ -8450,6 +8450,17 @@ <h1>Module Namespace Exotic Objects</h1>
...
+            </td>
+            <td>
+              Field reserved for use by host environments that need to associate additional information with a Module Namespace Exotic Object.
```
ljharb: I'm a bit confused why this is needed - implementations can always add any number of internal slots (or side-channel metadata) to any value.

spec.html
```diff
@@ -22373,11 +22389,12 @@ <h1>Runtime Semantics: GetModuleNamespace ( _module_ )</h1>
...
           1. Let _namespace_ be _module_.[[Namespace]].
           1. If _namespace_ is *undefined*, then
-            1. Let _exportedNames_ be ? _module_.GetExportedNames(&laquo; &raquo;).
+            1. Let _exportedNames_ be ? _module_.GetExportedNames(&laquo; &raquo;, _module_).
             1. Let _unambiguousNames_ be a new empty List.
-            1. For each _name_ that is an element of _exportedNames_, do
-              1. Let _resolution_ be ? _module_.ResolveExport(_name_, &laquo; &raquo;).
-              1. If _resolution_ is a ResolvedBinding Record, append _name_ to _unambiguousNames_.
+            1. If _exportedNames_ is not *null*, then
+              1. For each _name_ that is an element of _exportedNames_, do
```
ljharb: this already exists in master, but it's a somewhat confusing way of talking about iterating through a List to me. Typically I'd expect `For each _name_ of _exportedNames_, do`

spec.html
guybedford: I've applied this in https://github.com/nodejs/dynamic-modules/blob/master/spec.html#L203.

The nice thing about this is that it avoids having to declare the field in the dynamic modules spec, retaining the spec separation of simply implementing a subclass of module record and not changing any other objects or fields.

guybedford: Other suggestions welcome too.

spec.html
guybedford: The wording here was already used in eg Realm.

spec.html
guybedford: Sure I've updated this.

spec.html
ljharb: It seems like if the only spec to use it is the dynamic modules spec, then that's the place it should be declared :-/

guybedford: Are you saying that a spec can add fields to any object in the ECMA-262 spec while remaining valid against ECMA-262? I thought the point of [[HostDefined]] was to clarify these cases?

domenic: Although technically hosts can do crazy things, it's much more friendly to spec readers and to implementers if integration points are called out explicitly and made to interoperate in the same way they would in implementations. I strongly support adding [[HostDefined]] fields where appropriate in this way to increase the fidelity and readability of the spec's layering.

domenic:
Although technically hosts can do crazy things, it's much more friendly to spec readers and to implementers if integration points are called out explicitly and made to interoperate in the same way they would in implementations. I strongly support adding [[HostDefined]] fields where appropriate in this way to increase the fidelity and readability of the spec's layering.

ljharb:
@domenic i agree in principle, but how do different hosts and specs use the HostDefined slot currently that requires interop? Why not implicitly define that slot on every value, if it’s needed?

In this PR, the slot is initially set to `undefined` and then never referenced again. Why does the slot need to exist for a hot, or other spec, to install it as needed?

domenic:
I think @guybedford linked you to a pull request that explicitly uses the [[HostDefined]] slot to give better layering and implementability.

> Why not implicitly define that slot on every value, if it’s needed?

The point of my above post was that explicit is better than implicit, and that it's not needed on every value: it's needed on specific ones, which benefit from having clear specs emplacing them on those values.

> Why does the slot need to exist for a hot, or other spec, to install it as needed?

The point of my above post is that hosts should not have to do wacky things like installing slots (which, remember, in ECMAScript are immutable at object creation time) as a [monkey-patch](https://annevankesteren.nl/2014/02/monkey-patch) onto other specs. They should instead explicitly coordinate with the base spec to create explicit, clear hook points (in this case, a [[HostDefined]] slot).

zenparsing:
The "monkey patch" point certainly makes sense, but I'm wondering how this doesn't devolve into having `[[HostDefined]]` on gradually more and more things.

Currently we have `[[HostDefined]]` fields on the following Record types:

- Module Record
- Realm
- Pending Job
- Script Record

We don't currently have it on any object internal slot.

Separately, I'm having a difficult time understanding how this works with star exports. If I understand correctly, with this patch the namespace object for any module that exports `*` can in principle have an empty `[[Exports]]` list. Presumably that list is going to be dynamically updated during evaluation somehow.

Given that `[[Exports]]` is going to be updated during the evaluation phase, how do we know that the object model invariants of namespace objects will hold?

spec.html
ljharb: Fair enough; we can update them all separately.

jmdyck:
@domenic:
> slots (which, remember, in ECMAScript are immutable at object creation time)

The *value* of an internal slot can certainly change after object creation, so they're not immutable in that sense. I'm guessing what you meant was that the *set* of internal slots is fixed at object creation time, based on:

> Unless explicitly specified otherwise, internal slots are allocated as part of the process of creating an object and may not be dynamically added to an object.

but note that even that has the "unless otherwise specified" loophole. (I don't know if another spec would be allowed to use that loophole, or whether it implicitly means "... specified *within this specification*".)

(This is just side-clarification, not an argument in any direction. Explicit, clear hook points are good.)

spec.html
guybedford: @jmdyck thanks for looking into this further, it's good to ensure these conventions. Do you think sticking with one of the more explicit options would be better then? Although as @ljharb mentions perhaps this could be a separate editorial PR too.

jmdyck: I think what you've got is fine for this PR, and consistification should be left to a separate PR. (= agree with @ljharb, sounds like)

guybedford:
@zenparsing 

> Separately, I'm having a difficult time understanding how this works with star exports. If I understand correctly, with this patch the namespace object for any module that exports * can in principle have an empty [[Exports]] list. Presumably that list is going to be dynamically updated during evaluation somehow.

Yes, it's almost like there is a "namespace object TDZ" here while the dynamic module hasn't yet executed just in that cycle phase. In most cases this isn't observable, but where it is it is similar to how CommonJS circular references work except that it is not progressive - the namespace will go from being empty to having all exports defined. If you're interested in the edge cases on this one, the tests for this go through the various object behaviours during this unfinished phase at https://chromium-review.googlesource.com/c/v8/v8/+/1303725/6/test/cctest/test-modules.cc#621.

zenparsing:
Thanks for the link, that helps.

Specifically, I'm thinking about the invariant for `[[GetOwnProperty]]` which [states](https://tc39.github.io/ecma262/#sec-invariants-of-the-essential-internal-methods):

> If the target is non-extensible and P is non-existent, then all future calls to [[GetOwnProperty]] (P) on the target must describe P as non-existent (i.e. [[GetOwnProperty]] (P) must return undefined).

In the context of the current spec, calling `Reflect.getOwnPropertyDescriptor("test")` on one of these "dynamic/not-ready" namespace objects will return `undefined` because "test" isn't in the namespace object's `[[Exports]]` list yet.

Later, when the namespace object has been fully populated, calling `Reflect.getOwnPropertyDescriptor("test")` would return a property descriptor, which would seem to contradict the invariant above.

guybedford:
To summarize the suggested change, what we currently have is:

```js
// During Namespace "TDZ":
Object.isExtensible(ns); // Returns *false*
Object.preventExtensions(ns); // Success (no-op)
Object.defineProperty(ns, 'x', { value: 'asdf' }); // Throws

// (Same behaviour after "TDZ")
```

The suggested change to ensure the "IsExtensible invariant" would be:

```js
// During Namespace "TDZ":
Object.isExtensible(ns); // Returns *true*
Object.preventExtensions(ns); // Throws
Object.defineProperty(ns, 'x', { value: 'asdf' }); // Throws

// After Namespace "TDZ":
Object.isExtensible(ns); // Returns *false*
Object.preventExtensions(ns); // Success (no-op)
Object.defineProperty(ns, 'x', { value: 'asdf' }); // Throws
```

This one will likely need some implementation feedback as well, so will update how this goes looking into the v8 side of things.

Further feedback / suggestions here very welcome.

spec.html
```diff
@@ -8602,12 +8608,15 @@ <h1>ModuleNamespaceCreate ( _module_, _exports_ )</h1>
         <emu-alg>
           1. Assert: _module_ is a Module Record.
           1. Assert: _module_.[[Namespace]] is *undefined*.
-          1. Assert: _exports_ is a List of String values.
+          1. Assert: _exports_ is a List of String values or *null*.
```
ljharb: This confused me; do you mean, “exports is either null, or a List of a string values”, or do you mean that the list contains strings or nulls? If the former, maybe using that wording would help.

zenparsing:
Allowing the namespace object to be extensible would resolve the invariant issue, but perhaps goes against the current spirit of namespace objects as object model manefestations of cross-module bindings.

Just curious: have we considered simply disallowing export star from dynamic modules? How important is the ability to export star from a CJS module?

bmeck:
@zenparsing to my knowledge the invariant isn't broken since the act of freezing/preventingExtensions/etc. is allowed to add descriptors.

guybedford:
> Allowing the namespace object to be extensible would resolve the invariant issue, but perhaps goes against the current spirit of namespace objects as object model manefestations of cross-module bindings.

The latest commit doesn't allow the namespace object to be extensible - rather it simply marks the [[IsExtensible]] field as true if it is not yet finalized, and false afterwards, while still not permitting user-extensions during its [[IsExtensible]] phase (if any).

> Just curious: have we considered simply disallowing export star from dynamic modules? How important is the ability to export star from a CJS module?

It is very important, as `import * as m from 'cjs'` is a Common pattern already.

zenparsing:
OK, so it makes namespaces "extensible but refuses extension."

Another option might be disallow any property reflection on dynamic namespace objects until evaluation, so that `[[OwnPropertyKeys]]`, `[[Get]]`, `[[Set]]`, `[[HasProperty]]`, `[[DefineOwnProperty]]` , etc. all fail.

zenparsing:
The more I think about it, the more I dislike the idea of namespace objects observably mutating. Is it feasible to have these "TDZ" namespaces refuse all reflection?

> It is very important, as import * as m from 'cjs' is a Common pattern already.

It certainly needs to support `import *` (and therefore we need to deal with the "TDZ namespace" issue), but does it need to support `export * from 'dynamic'`? Dynamic export star results in non-local proliferation of dynamic namespace objects.

Specwise, `dynamicModule.Instantiate` could eagerly create its own "TDZ" namespace object, and always throw on "GetExportedNames".

ljharb:
@zenparsing wouldn't that mean you could determine if a module was dynamic or not by dynamically importing a file that attempted to re-export it? It seems very undesirable to make some kinds of modules' formats observable from outside, as it would hinder refactoring from, say, CJS to ESM.

domenic:
I don't think we should make any changes to how Source Text Module Records work.

jdalton:
@domenic 
> I don't think we should make any changes to how Source Text Module Records work.

Are you suggesting the work be done in the Abstract Module Record or that late population of the namespace objects for non-dynamic modules is something you don't want to be possible?

domenic:
I think whatever goals are desired for this Node.js-specific module type should be done by not affecting the behavior of Source Text Module Records.

jdalton:
@allenwb 
> I don't see any need to change the semantics of ECMAScript source modules to accommodate a reasonably specified foreign module mechanism.

The issue we're facing in Node-land is CJS modules (dynamic-modules) don't have their `module.exports` fully realized until the module executes. For Source Text modules importing dynamic-modules like `import {a} from "cjs"` or `export {a} from "cjs"` the WIP approach taken is for the dynamic-module to [resolve an export](https://nodejs.github.io/dynamic-modules/#sec-dynamicresolveexport) binding for `a`, before we know if it really exists, so that it can work with Source Text modules. In cyclical load situations, before the dynamic-module is executed, the `a` binding would even be observable on the namespace object of Source Text module that `export {a} from "cjs"` _(though if you tried to access `ns.a` it would throw since it's uninitialized)_. Then in the [evaluation phase](https://nodejs.github.io/dynamic-modules/#sec-dynamicmoduleevaluation) of the dynamic-module `a` is validated. 

This trick works OK but falls down for `export * from "cjs"` because we don't know the export names to fake-out the Source Text module with, which is why this PR has dynamic-modules throwing for that case.

In my own experiments one way around it, [that I've gotten to work](https://twitter.com/jdalton/status/1089100922083454976), is in part due to deferring the Source Text Module namespace object population/finalization for cases of `export {a} from "cjs"` and `export * from "cjs"`. Though the full solution is more involved, the minimum bit of foundation spec to enable a possible future where `export * from "cjs"` doesn't throw would be to enable deferring Source Text Module namespace object population/finalization in a way similar to how dynamic-modules [defer their namespace object population/finalization](https://github.com/tc39/ecma262/pull/1306/files).

While deferring Source Text Module namespace object population/finalization is one possible way I'd :heart: other ideas for, and for folks to noodle on, how to accommodate `export * from "cjs"` so we can avoid throwing.

domenic:
It seems to me that, if Node can't uphold the semantics of a Source Text Module Record for whatever reason, they should instead use their own type of module record, instead of bringing everyone else down to their level by changing the core language semantics.

Another way of looking at it would be that, as far as I know at least, the only environment so far that's managed to implement the Source Text Module Record is the browser. Maybe it'd be better to move the STMR spec into the web spec space, instead of trying to pretend it's part of a core, host-agnostic language semantic. In this way it would join the jobs infrastructure (#735) in the pile of things where ECMA-262 overreached and tried to specify something as "language" level, but it ended up not being suitable for realistic implementations. In this world, "Web JavaScript Module Records" and "Node.js Module Records" would be on equal footing, alongside WebAssembly Module Records, HTML Module Records, JSON Module Records, etc. They would all be extensions of Abstract Module Record. (Or, even better if #1311 ever lands, extensions of Cyclic Module Record.)

littledan:
There's a bunch going on in this discussion, so I want to try to break it out a bit, summarizing the above discussion to make next steps more clear. Let's consider individual Q/A questions here:

**Q**: *Given* the goal of allowing static and dynamic modules to interact in Node.js, what does TC39 think of this proposal?

**A**: Many TC39 delegates (including @zenparsing above, and more at the November 2018 breakout sessions) expressed support for working on this problem, but skepticism about the semantics of the particular re-export case. This current PR seems to address those concerns.

**Q**: What does TC39 think of integrating dynamic and static modules with this level of dynamism?

**A**: @allenwb [expressed some concerns](https://github.com/tc39/ecma262/pull/1306#issuecomment-457913749) about this integration above. We don't seem to have consensus as a committee as to whether this is a good thing to pursue, given that others expressed support. Offline, I have heard another angle of concern, that it would be unfortunate to implement this interoperability in Node.js when we don't expect it to ever make it to the Web.

**Q**: For host environments that will only support source text module records, and have no plans to add dynamic module records, is TC39 OK with adding additional conditionals in the code path for source text module records?

**A**: I think @domenic is [arguing](https://github.com/tc39/ecma262/pull/1306#issuecomment-457860511) "no" above.

**Q**: Is TC39 interested in layering patches for non-web environments? Are we interested in being consulted on host environment integration issues like this?

**A**: (I think we should be open to both of these!) This PR is based on the assumption that the answer is yes, and no one is arguing explicitly against that, but I am not sure if we all agree on this point. Probably Node could move forward with a parallel type to Source Text Module Records to manage all of this, with the minor change, and leave all the rest in tact. The big change would be that some embedding environments would be modifying the `[[Exports]]` list at a later time, but Source Text Module Records would never see such a list, and I see no assertion in the spec that it doesn't change.

The result of doing everything in a separate document would be either inconsistency/vagueness, a lot of duplication or a little bit of monkey patching, all of which @domenic and @annevk have done a lot of good work to reduce over time with the Web embedding. Many JS engines are used in both the Web and Node.js, so implementers of those engines need to be aware of the set of relevant spec documents regardless of whether it goes into ECMA-262 or not. If this is to be maintained as a separate document, I think the clearest way to present it would be ins tags around part of the TC39 text, like the current proposal. If we are, in general, not interested in such layering patches for other environments, and just for technical review, let's make this clear to help everyone save time. 

So, what are our recommendations for @guybedford from here? The options I see are:
1. Recommend that Node.js doesn't integrate CJS modules this way.
2. Encourage @guybedford to make a separate standard-like document with the two module types included, if he wants to formally specify this, without touching ECMA-262. This could come with or without TC39's "blessing" for this path.
3. Land something like this patch, but consider adding an assertion somewhere that the null check path is never hit in environments with only Source Text Module Records (or Circular Module Records).

allenwb:
@domenic
 There are ways for node modules and ECMAScript modules to co-exist that doesn't have these problems.  For example:
1.  Use `require` to access cjs modules from a ESM. Don't pretend that a CJS module provides the export semantics needed to import bindings into an ESM.
2. Treat a CJS module's module.exports as a ES default export. The CJS exported properties are accessed as properties of the imported default objet but are not individually linkable bindings.
3. Associate static metadata with any CJS module that wants to provide exported bindings that can be imported by a ESM. The metadata would declare the exports and perhaps provide a mapping from module.export names to export bindings.
4. ...

It isn't clear to me why CJS modules need to start acting as if they where ESMs.  There is no backwards compatibility reason for a CJS module to export bindings to an ESM or to participate in circularities involving ESMs. There are speculative scenarios that can only exist in an environment that has both ESMs (conforming to the specified semantics) coexisting with CJS.  But in such an environment if you want to export bindings or participate in circuaritwes  why wouldn't you simply define a ESM rather than defining a new CJS module?

allenwb:
@littledan 
see the aside in my previous response about the "dynamic module" terminology.  I really think that "dynamic module" is too broad of a term to be applied to the CJS specific issues that are being discussed.  There are ways to support the dynamic creation of other kinds of  modules (including reflectively using JS code)  that could support the existing source code module linking semantics and not run into these issues.

ljharb:
@allenwb the reason is that both conceptually, and in practice for many years via babel, that’s how virtually the entire npm-using javascript community has expected to be able to import CJS modules. Whether you agree with it or not, this interoperability point is critical for some in the node modules working group to be able to ship ESM at all.

If ESM with STMR can’t be implemented by a non-web implementation like node, then perhaps @domenic’s suggestion above would make sense.

jdalton:
To follow up on @ljharb and +1 _(I was writing when his reply came in)_.

@allenwb 

The goal of dynamic-module is to improve CJS+ESM interoperability by taking the implementation from its current experimental status of
> Treat a CJS module's module.exports as a ES default export.

which won't likely ship unflagged in Node. To one with named exports, which has a higher likelihood of shipping. One of the wins with dynamic-modules is that it more closely meets developers expectations of interoperability (established by Babel, TypeScript, Webpack, etc.).

>  Associate static metadata with any CJS module that wants to provide exported bindings that can be imported by a ESM. The metadata would declare the exports and perhaps provide a mapping from module.export names to export bindings.

It would be ideal to avoiding associating yet more metadata, and taxing developers with that backfilling/maintenance cost, so that the existing ecosystem of hundreds of thousands of CJS modules can be loaded.

I think @domenic may be on to something with [his suggestion](https://github.com/tc39/ecma262/pull/1306#issuecomment-457944215):
> Maybe it'd be better to move the STMR spec into the web spec space

spec.html
```diff
@@ -22179,6 +22191,11 @@ <h1>ModuleDeclarationEnvironmentSetup ( _module_ )</h1>
                   1. Let _resolution_ be ? _importedModule_.ResolveExport(_in_.[[ImportName]], &laquo; &raquo;).
                   1. If _resolution_ is *null* or `"ambiguous"`, throw a *SyntaxError* exception.
                   1. Call _envRec_.CreateImportBinding(_in_.[[LocalName]], _resolution_.[[Module]], _resolution_.[[BindingName]]).
+              1. For each ExportEntry Record _e_ in _module_.[[StarExportEntries]], do
```
domenic: This is a change to Source Text Module Records, so I am unsure why you are saying there are no changes to STMR in this PR.

spec.html
```diff
@@ -22026,14 +22036,16 @@ <h1>GetExportedNames ( _exportStarSet_ ) Concrete Method</h1>
             1. For each ExportEntry Record _e_ in _module_.[[StarExportEntries]], do
               1. Let _requestedModule_ be ? HostResolveImportedModule(_module_, _e_.[[ModuleRequest]]).
               1. Let _starNames_ be ? _requestedModule_.GetExportedNames(_exportStarSet_).
+              1. If _starNames_ is *null*, then
```
domenic: Also a change to STMR

spec.html
guybedford: Ah, yes, apologies I mispoke on that one.

MylesBorins:
@linclark @guybedford do you think it would be possible to rebase against https://github.com/tc39/ecma262/pull/1311 and extent cyclic module records instead of source text module record?

@domenic would that appease your concerns related to modifying STMR?

domenic:
No, it's the modification to core ES infrastructure that's problematic, not the particular name "Source Text Module Record" vs. "Cyclic Module Record".

guybedford:
Closing as this is blocked on various constraints. Here is a brief summary for future reference:

1. No consensus on the ability for namespaces to have this undefined state.
2. No consensus on the ability for the ECMA-262 module system to define this state to support dynamic modules.

The only alternative to the above is to change the way execution order happens in cycles such that the dynamic modules namespaces in cycles are no longer observable at all. This was created at https://github.com/nodejs/dynamic-modules/pull/3 if anyone wants to pick it up.

Apart from the above, this means that defining export names after instantiation for any module record simply seems like it will never be possible in ECMA-262. As a result CommonJS must either execute during instantiation, or only define a default export.

PR #1311, opened 2018-09-24
Layering: Enable cyclic dependencies with non-STMR module types

linclark:
Currently, only Source Text Module Records (STMR) can participate in cycles. From my conversations with those who put that limitation in the spec, this restriction was only added because there wasn't an immediate need for cyclic dependencies with non-STMR modules. 

This PR makes it possible for other languages to add module record types that can participate in cycles with Source Text Module Records. To do this, it moves the logic for managing cycles to a new module type, Cyclic Module Record (CMR). STMR are a subclass of the CMR, and other specs such as WebAssembly can also subclass this module record type.

To help with review, I've broken this into smaller commits, but can squash them before merge if that is preferred.

Note: This is an update to #1199. In this version, I removed the concept of "subphases" that was introduced there because they are no longer needed.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -21691,6 +21691,48 @@
             </emu-alg>
           </emu-clause>
         </emu-clause>
+
+        <emu-clause id="sec-example-cyclic-module-record-graphs">
+          <h1>Example Cyclic Module Record Graphs</h1>
+
+          <p>This non-normative section gives a series of examples of the instantiation and evaluation of a few common module graphs, with a specific focus on how errors can occur.</p>
+
+          <p>First consider the following simple module graph:</p>
+
+          <emu-figure id="figure-module-graph-simple" caption="A simple module graph">
+            <img alt="A module graph in which module A depends on module B" width="121" height="211" src="img/module-graph-simple.svg">
+          </emu-figure>
+
+          <p>Let's first assume that there are no error conditions. When a host first calls _A_.Instantiate(), this will complete successfully by assumption, and recursively instantiate modules _B_ and _C_ as well, such that _A_.[[Status]] = _B_.[[Status]] = _C_.[[Status]] = `"instantiated"`. This preparatory step can be performed at any time. Later, when the host is ready to incur any possible side effects of the modules, it can call _A_.Evaluate(), which will complete successfully (again by assumption), recursively having evaluated first _C_ and then _B_. Each module's [[Status]] at this point will be `"evaluated`".</p>
+
+          <p>Consider then cases involving instantiation errors. If InnerModuleInstantiation of _C_ succeeds but, thereafter, fails for _B_, for example because it imports something that _C_ does not provide, then the original _A_.Instantiate() will fail, and both _A_ and _B_'s [[Status]] remain `"uninstantiated"`. _C_'s [[Status]] has become `"instantiated"`, though.</p>
+
+          <p>Finally, consider a case involving evaluation errors. If InnerModuleEvaluation of _C_ succeeds but, thereafter, fails for _B_, for example because _B_ contains code that throws an exception, then the original _A_.Evaluate() will fail. The resulting exception will be recorded in both _A_ and _B_'s [[EvaluationError]] fields, and their [[Status]] will become `"evaluated"`. _C_ will also become `"evaluated"` but, in contrast to _A_ and _B_, will remain without an [[EvaluationError]], as it successfully completed evaluation. Storing the exception ensures that any time a host tries to reuse _A_ or _B_ by calling their Evaluate() method, it will encounter the same exception. (Hosts are not required to reuse Cyclic Module Records; similarly, hosts are not required to expose the exception objects thrown by these  methods. However, the specification enables such uses.)</p>
+
+          <p>The difference here between instantiation and evaluation errors is due to how evaluation must be only performed once, as it can cause side effects; it is thus important to remember whether evaluation has already been performed, even if unsuccessfully. (In the error case, it makes sense to also remember the exception because otherwise subsequent Evaluate() calls would have to synthesize a new one.) Instantiation, on the other hand, is side-effect-free, and thus even if it fails, it can be retried at a later time with no issues.</p>
+
+          <p>Now consider a different type of error condition:</p>
+
+          <emu-figure id="figure-module-graph-missing" caption="A module graph with an unresolvable module">
+            <img alt="A module graph in which module A depends on a missing (unresolvable) module, represented by ???" width="121" height="121" src="img/module-graph-missing.svg">
+          </emu-figure>
+
+          <p>In this scenario, module _A_ declares a dependency on some other module, but no Module Record exists for that module, i.e. HostResolveImportedModule throws an exception when asked for it. This could occur for a variety of reasons, such as the corresponding resource not existing, or the resource existing but ParseModule throwing an exception when trying to parse the resulting source text. Hosts can choose to expose the cause of failure via the exception they throw from HostResolveImportedModule. In any case, this exception causes an instantiation failure, which as before results in _A_'s [[Status]] remaining `"uninstantiated"`.</p>
+
+          <p>Lastly, consider a module graph with a cycle:</p>
+
+          <emu-figure id="figure-module-graph-cycle" caption="A cyclic module graph">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -22082,264 +22317,83 @@ <h1>ResolveExport ( _exportName_, _resolveSet_ ) Concrete Method</h1>
...
-            <h1>ModuleDeclarationEnvironmentSetup ( _module_ )</h1>
-
-            <p>The ModuleDeclarationEnvironmentSetup abstract operation is used by InnerModuleInstantiation to initialize the Lexical Environment of the module, including resolving all imported bindings.</p>
-
-            <p>This abstract operation performs the following steps:</p>
-
-            <emu-alg>
-              1. For each ExportEntry Record _e_ in _module_.[[IndirectExportEntries]], do
-                1. Let _resolution_ be ? _module_.ResolveExport(_e_.[[ExportName]], &laquo; &raquo;).
+            1. 1. _module_ be this Source Text Module Record.
```
jmdyck: "1." -> "Let"

spec.html
```diff
@@ -21337,8 +21337,7 @@ <h1>Static Semantics: VarScopedDeclarations</h1>
...
+        <p>For specification purposes Module Record values are values of the Record specification type and can be thought of as existing in a simple object-oriented hierarchy where Module Record is an abstract class with both abstract and concrete subclasses. This specification defines the abstract subclass named Cyclic Module Record and its concrete subclass named Source Text Module Record. Other specifications and implementations may define additional Module Record subclasses corresponding to alternative module definition facilities that they defined.</p>        <p>Module Record defines the fields listed in <emu-xref href="#table-36"></emu-xref>. All Module Definition subclasses include at least those fields. Module Record also defines the abstract method list in <emu-xref href="#table-37"></emu-xref>. All Module definition subclasses must provide concrete implementations of these abstract methods.</p>
```
jmdyck: There should be a linebreak after the first `</p>`.

spec.html
```diff
@@ -21452,15 +21451,296 @@ <h1>Abstract Module Records</h1>
...
+                1. Return _index_.
+              1. If _module_.[[Status]] is `"instantiating"`, `"instantiated"`, or `"evaluated"`, then
+                1. Return _index_.
+              1. Assert: _module_.[[Status]] is `"uninstantiated"`.
+              1. Set _module_.[[Status]] to `"instantiating"`.
+              1. Set _module_.[[DFSIndex]] to _index_.
+              1. Set _module_.[[DFSAncestorIndex]] to _index_.
+              1. Increase _index_ by 1.
+              1. Append _module_ to _stack_.
+              1. For each String _required_ that is an element of _module_.[[RequestedModules]], do
```
jmdyck: `_module_.[[RequestedModules]]` is defined only if `_module_` is a Source Text Module Record.

spec.html
```diff
@@ -22082,264 +22317,83 @@ <h1>ResolveExport ( _exportName_, _resolveSet_ ) Concrete Method</h1>
           </emu-alg>
         </emu-clause>
 
-        <emu-clause id="sec-moduledeclarationinstantiation">
-          <h1>Instantiate ( ) Concrete Method</h1>
+        <emu-clause id="sec-moduledeclarationenvironmentsetup" aoid="ModuleDeclarationEnvironmentSetup">
+          <h1>ModuleDeclarationEnvironmentSetup ( )</h1>
```
jmdyck: Clause header should end with "Concrete Method".

spec.html
```diff
@@ -22082,264 +22317,83 @@ <h1>ResolveExport ( _exportName_, _resolveSet_ ) Concrete Method</h1>
...
+                  1. Perform ! _envRec_.CreateMutableBinding(_dn_, *false*).
+                1. If _d_ is a |FunctionDeclaration|, a |GeneratorDeclaration|, an |AsyncFunctionDeclaration|, or an |AsyncGeneratorDeclaration|, then
+                  1. Let _fo_ be the result of performing InstantiateFunctionObject for _d_ with argument _env_.
+                  1. Call _envRec_.InitializeBinding(_dn_, _fo_).
+            1. Return NormalCompletion(~empty~).
+          </emu-alg>
         </emu-clause>
+        
+        <emu-clause id="sec-moduleexecution" aoid="ModuleExecution">
+          <h1>ModuleExecution ( )</h1>
```
jmdyck: Clause header should end with "Concrete Method".

spec.html
```diff
@@ -22082,264 +22317,83 @@ <h1>ResolveExport ( _exportName_, _resolveSet_ ) Concrete Method</h1>
...
         </emu-clause>
 
-        <emu-clause id="sec-moduledeclarationinstantiation">
-          <h1>Instantiate ( ) Concrete Method</h1>
+        <emu-clause id="sec-moduledeclarationenvironmentsetup" aoid="ModuleDeclarationEnvironmentSetup">
+          <h1>ModuleDeclarationEnvironmentSetup ( )</h1>
 
-          <p>The Instantiate concrete method of a Source Text Module Record implements the corresponding Module Record abstract method.</p>
-          <p>On success, Instantiate transitions this module's [[Status]] from `"uninstantiated"` to `"instantiated"`. On failure, an exception is thrown and this module's [[Status]] remains `"uninstantiated"`.</p>
+          <p>The ModuleDeclarationEnvironmentSetup concrete method implements the corresponding Cyclic Module Record abstract method.</p>
```
jmdyck: After "concrete method", insert "of a Source Text Module Record".

spec.html
```diff
@@ -22082,264 +22317,83 @@ <h1>ResolveExport ( _exportName_, _resolveSet_ ) Concrete Method</h1>
...
+        
+        <emu-clause id="sec-moduleexecution" aoid="ModuleExecution">
+          <h1>ModuleExecution ( )</h1>
 
-        <emu-clause id="sec-moduleevaluation">
-          <h1>Evaluate ( ) Concrete Method</h1>
-
-          <p>The Evaluate concrete method of a Source Text Module Record implements the corresponding Module Record abstract method.</p>
-          <p>Evaluate transitions this module's [[Status]] from `"instantiated"` to `"evaluated"`.</p>
+          <p>The ModuleExecution concrete method implements the corresponding Cyclic Module Record abstract method.</p>
```
jmdyck: After "concrete method", insert "of a Source Text Module Record".

spec.html
zenparsing: It seems like `[[RequestedModules]]` needs to be defined on Cyclic Module Record in order for the algorithm to traverse the graph. Is that right?

guybedford: Yes exactly, the field just needs to be moved across.

spec.html
```diff
@@ -21452,15 +21451,296 @@ <h1>Abstract Module Records</h1>
...
+                <th>
+                  Method
+                </th>
+                <th>
+                  Purpose
+                </th>
+              </tr>
+              <tr>
+                <td>
+                  ModuleDeclarationEnvironmentSetup()
```
zenparsing: We might want to give these names that are consistent with the other abstract methods (i.e. a verb form). I'm not sure what would be the best name here though. `InitializeEnvironment`? Something else?

littledan:
PR to address the issues noted above: https://github.com/linclark/ecma262/pull/1

littledan:
BTW, as this is a layering change (and not normative), and all concerns raised have been addressed, do we need to discuss it in a TC39 meeting, or can it be merged without that? Asking just to know whether I should prepare a presentation for November.

domenic:
I've noticed a lot of approved-but-not-merged PRs on this repository. This particular one is somewhat blocking to HTML modules work, and we'd love to see it land as soon as possible.

ljharb:
Most of those aren't approved by all the editors yet (specifically, the editor in chief), thus they're not actually mergeable yet. I'll make sure we address this one in the next editor's meeting.

littledan:
Another month has passed. What's the process from here?

zenparsing:
@bterlson This PR is labelled "es2019" but I don't know if we talked about whether you wanted to bring this one in for the 2019 branch. Thoughts?

MylesBorins:
quick ping to see where this is going. It seems like cyclic module records are going to be a very useful abstraction

bterlson:
Sorry for the delay, folks. This is good to go! Thanks for sticking with it.

littledan:
Thanks for landing this!

PR #1312, opened 2018-09-24
 Layering: Rename Module.Instantiate to Module.Link 

linclark:
The terminology between WebAssembly and ES modules differs when it comes to instantiation, which causes some confusion for the integration of the two systems. Ideally, we wouldn't use the same term to mean two different things between the specs.

After talking about it with multiple people, it seems like calling it "the linking phase" might be a better name for this phase anyway. Developers are often confused about what actually happens during instantiation and why it's separated from evaluation. I believe "linking" more clearly communicates the justification for this being its own phase.

Note: This depends on #1311. The relevant commit for this PR is the last one. Once #1311  is merged, I can rebase this to remove those commits.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -21567,7 +21567,7 @@
                 Lexical Environment | *undefined*
               </td>
               <td>
-                The Lexical Environment containing the top level bindings for this module. This field is set when the module is instantiated.
+                The Lexical Environment containing the top level bindings for this module. This field is set when the module is linked.
               </td>
             </tr>
             <tr>
@@ -21625,7 +21625,7 @@
             </tr>
             <tr>
               <td>
-                Instantiate()
+                Link()
               </td>
               <td>
                 <p>Prepare the module for evaluation by transitively resolving all module dependencies and creating a module Environment Record.</p>
@@ -21637,7 +21637,7 @@
               </td>
               <td>
                 <p>If this module has already been evaluated successfully, return *undefined*; if it has already been evaluated unsuccessfully, throw the exception that was produced. Otherwise, transitively evaluate all module dependencies of this module and then evaluate this module.</p>
-                <p>Instantiate must have completed successfully prior to invoking this method.</p>
+                <p>Link must have completed successfully prior to invoking this method.</p>
               </td>
             </tr>
             </tbody>
@@ -21671,7 +21671,7 @@
                   String
                 </td>
                 <td>
-                  Initially `"uninstantiated"`. Transitions to `"instantiating"`, `"instantiated"`, `"evaluating"`, `"evaluated"` (in that order) as the module progresses throughout its lifecycle.
+                  Initially `"unlinked"`. Transitions to `"linking"`, `"linked"`, `"evaluating"`, `"evaluated"` (in that order) as the module progresses throughout its lifecycle.
                 </td>
               </tr>
               <tr>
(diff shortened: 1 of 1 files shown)


annevk:
If this change is made, it would also require downstream changes, e.g., to https://html.spec.whatwg.org/.

cc @domenic

allenwb:
I also like this name change.

dherman:
I've always been a fan of Legend of Zelda. 👍

spec.html
```diff
@@ -21337,8 +21337,7 @@ <h1>Static Semantics: VarScopedDeclarations</h1>
...
+        <p>For specification purposes Module Record values are values of the Record specification type and can be thought of as existing in a simple object-oriented hierarchy where Module Record is an abstract class with both abstract and concrete subclasses. This specification defines the abstract subclass named Cyclic Module Record and its concrete subclass named Source Text Module Record. Other specifications and implementations may define additional Module Record subclasses corresponding to alternative module definition facilities that they defined.</p>        <p>Module Record defines the fields listed in <emu-xref href="#table-36"></emu-xref>. All Module Definition subclasses include at least those fields. Module Record also defines the abstract method list in <emu-xref href="#table-37"></emu-xref>. All Module definition subclasses must provide concrete implementations of these abstract methods.</p>
```
TimothyGu: This should be divided into two lines at `<p>` boundaries. It seems like the two original lines got coalesced.

littledan:
The name change seems good to me; I'm happy to see the positive reaction here. Let's discuss this in the November 2018 TC39 meeting.

littledan:
Oh, sure, I'd be fine with landing this outside of plenary. Let's just not delay this longer than that.

Editors, what do you think?

bterlson:
We're fine with this change; will finish reviewing #1311 and will pull these in together.

littledan:
Ping on this change; I was hoping it would be pulled in together with #1311 per https://github.com/tc39/ecma262/pull/1312#issuecomment-432792042 .

ljharb:
@littledan this one won't be in 2019, and it'll need to be rebased as well (that will also help it be more easily reviewed, since it wasn't set up as a stacked PR)

littledan:
Rebase out for review at https://github.com/tc39/ecma262/pull/1465 . Please let me know if you need anything else.

ljharb:
To clarify, please do not open redundant PRs; either the PR author has to rebase it, or you can post a link to your branch, and this PR can updated by an editor with those commits.

littledan:
Oops, didn't realize I had write access to this branch. Rebased here.

spec.html
```diff
@@ -21708,58 +21708,58 @@ <h1>Cyclic Module Records</h1>
           </table>
         </emu-table>
 
-        <emu-clause id="sec-moduledeclarationinstantiation">
-          <h1>Instantiate ( ) Concrete Method</h1>
+        <emu-clause id="sec-moduledeclarationlinking">
```
zenparsing: @ljharb Will this result in links to the previous fragment not working? If so, do we care?

jmdyck: Could add `oldids="sec-moduledeclarationinstantiation"`

ajklein:
I like the name change conceptually, so happy to see this move forward. I don't think that we'd change V8's C++ naming just because of this spec change; rather we'd use the new name if we had other reasons that the API needed changing. @GeorgNeis in case he agrees or disagrees

spec.html
littledan: Good point; Added a couple oldids.

littledan:
Do you need anything else from me to review this patch? From https://github.com/tc39/ecma262/pull/1312#issuecomment-469400510 , it sounds like implementation burden isn't a big issue here.

Ms2ger:
Is someone following up with other specs that currently use "Instantiate"? I'm thinking at least HTML, WebIDL, top-level-await and wasm/esm-integration.

littledan:
@Ms2ger I haven't done this yet; if you are interested in fixing these other specs, it would be greatly appreciated.

Issue #64, opened 2015-09-23
Keyword list does not have line breaks.


mysticatea:
http://tc39.github.io/ecma262/#prod-Keyword

![image](https://cloud.githubusercontent.com/assets/1937871/10038879/d2dacff6-6206-11e5-8d9d-ce7cf3ce2aac.png)

Windows 7, Chrome 46.0.2490.33, Firefox 42.0a2.

bterlson:
Nice find... because the non-terminal elements don't have spaces between them they don't break. Maybe the right approach is for emu to put \\u200A HAIR SPACE between non-terminals or something. Will think about this.

michaelficarra:
Why hair space? Is there no zero-width way to force a break? If not, a regular space is probably fine since the margin on each of the `emu-t`s is larger than a space.

mysticatea:
Also, I have found a wrong line break in the `Punctuator` list.

![image](https://cloud.githubusercontent.com/assets/1937871/10063241/86345884-62a6-11e5-8f01-4a3ba6a10374.png)


So my idea is:

```diff
  emu-t {
      font-family: monospace;
      font-weight: bold;
+     white-space: nowrap;
  }
     
- emu-production emu-t {
-     margin-right: 1ex;
+ emu-production emu-t::after {
+     content: " ";
+     white-space: normal;
  }
```

![image](https://cloud.githubusercontent.com/assets/1937871/10063452/d81942f2-62a8-11e5-8be2-cf5c66a6b60b.png)
![image](https://cloud.githubusercontent.com/assets/1937871/10063460/e84ac4ac-62a8-11e5-9c32-af691df832f9.png)

bterlson:
Looks nice to me. Does it look ok with other productions that have non-terminals next to terminals? If so, want to send a PR to Ecmarkup? :)

mysticatea:
> Does it look ok with other productions that have non-terminals next to terminals?

Looks OK to me.

![image](https://cloud.githubusercontent.com/assets/1937871/10064156/70901626-62b0-11e5-8f29-fa3d0e55374d.png)

> If so, want to send a PR to Ecmarkup?

Yes, I will do :)

bterlson:
Awesome, looks great! I can commit this tomorrow too if you don't get around to it.

PR #1314, opened 2018-09-27
Editorial: Convert @@ notation to intrinsics notation for well-known Symbols

ljharb:
- this avoids a conceptual conflict with decorators, when they land

Open to bikesheds; other alternatives I considered:
 - explicitly using intrinsics, ie, `%SymbolSpecies%`, and hardcoding that table, while removing the "well-known symbols table". This felt like it would lose some clarity around intrinsics vs well-known-symbols.
 - Using `%Symbol%species%` - this would require upstream support in ecmarkup, so i opted not to do it for now.
 - something else?

### DIFF
--- a/img/figure-2.svg
+++ b/img/figure-2.svg
@@ -83,14 +83,14 @@
       </g>
     </g>
 
-    <g transform="translate(370,160)">
-      <rect class="class" x="0" y="0" width="160" height="30" />
-      <rect class="class" x="0" y="30" width="160" height="70" />
+    <g transform="translate(360,160)">
+      <rect class="class" x="0" y="0" width="180" height="30" />
+      <rect class="class" x="0" y="30" width="180" height="70" />
       <g transform="translate(0,4)">
-        <text x="80" y="16" text-anchor="middle">Function.prototype</text>
+        <text x="90" y="16" text-anchor="middle">Function.prototype</text>
       </g>
       <g transform="translate(4,32)">
-        <text x="0" y="16">@@hasInstance()</text>
+        <text x="0" y="16">%Symbol.hasInstance%()</text>
         <text x="0" y="32">apply()</text>
         <text x="0" y="48">bind()</text>
         <text x="0" y="64">call()</text>
@@ -104,7 +104,7 @@
         <text x="110" y="16" text-anchor="middle">%GeneratorFunction.prototype%</text>
       </g>
       <g transform="translate(4,32)">
-        <text x="0" y="16">@@toStringTag =</text>
+        <text x="0" y="16">%Symbol.toStringTag% =</text>
         <text x="16" y="32">"GeneratorFunction"</text>
       </g>
     </g>
@@ -127,26 +127,26 @@
       </g>
     </g>
 
-    <g transform="translate(720,120)">
-      <rect class="class" x="0" y="0" width="160" height="30" />
-      <rect class="class" x="0" y="30" width="160" height="30" />
+    <g transform="translate(680,120)">
+      <rect class="class" x="0" y="0" width="200" height="30" />
+      <rect class="class" x="0" y="30" width="200" height="30" />
       <g transform="translate(0,4)">
-        <text x="80" y="16" text-anchor="middle">%IteratorPrototype%</text>
+        <text x="100" y="16" text-anchor="middle">%IteratorPrototype%</text>
       </g>
       <g transform="translate(4,32)">
-        <text x="0" y="16">@@iterator() : object</text>
+        <text x="0" y="16">%Symbol.iterator%() : object</text>
       </g>
     </g>
 
-    <g transform="translate(670,300)">
-      <rect class="class" x="0" y="0" width="210" height="30" />
-      <rect class="class" x="0" y="30" width="210" height="24" />
-      <rect class="class" x="0" y="54" width="210" height="56" />
+    <g transform="translate(630,300)">
+      <rect class="class" x="0" y="0" width="250" height="30" />
+      <rect class="class" x="0" y="30" width="250" height="24" />
+      <rect class="class" x="0" y="54" width="250" height="56" />
       <g transform="translate(0,4)">
-        <text x="110" y="16" text-anchor="middle">%GeneratorPrototype%</text>
+        <text x="120" y="16" text-anchor="middle">%GeneratorPrototype%</text>
       </g>
       <g transform="translate(4,32)">
-        <text x="0" y="16">@@toStringTag = "Generator"</text>
+        <text x="0" y="16">%Symbol.toStringTag% = "Generator"</text>
       </g>
       <g transform="translate(4,56)">
         <text x="0" y="16">next() : object</text>
@@ -189,7 +189,7 @@
       </g>
     </g>
 
-    <g transform="translate(570,200)">
+    <g transform="translate(570,195)">
       <path class="note" d="M 0,0 V 80 H 220 V 10 h -10 10 l -10,-10 v 10 -10 z" />
       <g transform="translate(4,4)">
         <text x="0" y="16">%GeneratorFunction.prototype%</text>
@@ -242,12 +242,12 @@
       <path class="proto-slot-arrowend" d="M 0,0 L 9,5 L 0,10 z" />
     </marker>
 
-    <path class="proto-slot-arrow" d="M 450,160 V 120" />
-    <path class="proto-slot-arrow" d="M 450,300 V 260" />
-    <path class="proto-slot-arrow" d="M 450,440 V 370" />
+    <path class="proto-slot-arrow" d="M 450,160 V 120" /> <!-- between Object.prototype and Function.prototype -->
+    <path class="proto-slot-arrow" d="M 450,300 V 260" /> <!-- between Function.prototype and %GeneratorFunction.prototype% -->
+    <path class="proto-slot-arrow" d="M 450,440 V 370" /> <!-- below %GeneratorFunction.prototype% -->
 
-    <path class="proto-slot-arrow" d="M 100,200 V 175 H 370" />
-    <path class="proto-slot-arrow" d="M 100,300 V 260" />
+    <path class="proto-slot-arrow" d="M 100,200 V 175 H 360" /> <!-- between Function and Function.prototype -->
+    <path class="proto-slot-arrow" d="M 100,300 V 260" /> <!-- between Function and %GeneratorFunction% -->
 
     <path class="proto-slot-arrow" d="M 800,120 V 105 H 530" />
     <path class="proto-slot-arrow" d="M 800,300 V 180" />
@@ -272,8 +272,8 @@
     <text x="335" y="306" text-anchor="end">constructor</text>
     <text x="185" y="334">prototype</text>
 
-    <path class="prototype-constructor-arrow" d="M 560,315 H 670" />
-    <text x="665" y="306" text-anchor="end">constructor</text>
+    <path class="prototype-constructor-arrow" d="M 560,315 H 630" />
+    <text x="665" y="295" text-anchor="end">constructor</text>
     <text x="565" y="334">prototype</text>
 
     <path class="prototype-arrow" d="M 530,455 H 720" />
--- a/spec.html
+++ b/spec.html
@@ -1207,7 +1207,8 @@
       <emu-clause id="sec-well-known-symbols">
         <h1>Well-Known Symbols</h1>
         <p>Well-known symbols are built-in Symbol values that are explicitly referenced by algorithms of this specification. They are typically used as the keys of properties whose values serve as extension points of a specification algorithm. Unless otherwise specified, well-known symbols values are shared by all realms (<emu-xref href="#sec-code-realms"></emu-xref>).</p>
-        <p>Within this specification a well-known symbol is referred to by using a notation of the form @@name, where “name” is one of the values listed in <emu-xref href="#table-well-known-symbols"></emu-xref>.</p>
+        <p>Within this specification a well-known symbol is referred to using the standard <emu-xref href="#sec-well-known-intrinsic-objects">intrinsic notation</emu-xref> where the intrinsic is one of the values listed in <emu-xref href="#table-well-known-symbols"></emu-xref>.</p>
+        <emu-note>Previous editions of this specification used a notation of the form @@name, where the current edition would use `%Symbol.name%`. In particular, the following names were used: @@asyncIterator, @@hasInstance, @@isConcatSpreadable, @@iterator, @@match, @@matchAll, @@replace, @@search, @@species, @@split, @@toPrimitive, @@toStringTag, and @@unscopables.</emu-note>
         <emu-table id="table-well-known-symbols" caption="Well-known Symbols" oldids="table-1">
           <table>
             <thead>
@@ -1225,7 +1226,7 @@
             </thead>
             <tr>
               <td>
-                <dfn>@@asyncIterator</dfn>
+                <dfn>%Symbol.asyncIterator%</dfn>
               </td>
               <td>
                 *"Symbol.asyncIterator"*
@@ -1236,7 +1237,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@hasInstance</dfn>
+                <dfn>%Symbol.hasInstance%</dfn>
               </td>
               <td>
                 *"Symbol.hasInstance"*
@@ -1247,7 +1248,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@isConcatSpreadable</dfn>
+                <dfn>%Symbol.isConcatSpreadable%</dfn>
               </td>
               <td>
                 *"Symbol.isConcatSpreadable"*
@@ -1258,7 +1259,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@iterator</dfn>
+                <dfn>%Symbol.iterator%</dfn>
               </td>
               <td>
                 *"Symbol.iterator"*
@@ -1269,7 +1270,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@match</dfn>
+                <dfn>%Symbol.match%</dfn>
               </td>
               <td>
                 *"Symbol.match"*
@@ -1280,7 +1281,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@matchAll</dfn>
+                <dfn>%Symbol.matchAll%</dfn>
               </td>
               <td>
                 *"Symbol.matchAll"*
@@ -1291,7 +1292,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@replace</dfn>
+                <dfn>%Symbol.replace%</dfn>
               </td>
               <td>
                 *"Symbol.replace"*
@@ -1302,7 +1303,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@search</dfn>
+                <dfn>%Symbol.search%</dfn>
               </td>
               <td>
                 *"Symbol.search"*
@@ -1313,7 +1314,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@species</dfn>
+                <dfn>%Symbol.species%</dfn>
               </td>
               <td>
                 *"Symbol.species"*
@@ -1324,7 +1325,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@split</dfn>
+                <dfn>%Symbol.split%</dfn>
               </td>
               <td>
                 *"Symbol.split"*
@@ -1335,7 +1336,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@toPrimitive</dfn>
+                <dfn>%Symbol.toPrimitive%</dfn>
               </td>
               <td>
                 *"Symbol.toPrimitive"*
@@ -1346,7 +1347,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@toStringTag</dfn>
+                <dfn>%Symbol.toStringTag%</dfn>
               </td>
               <td>
                 *"Symbol.toStringTag"*
@@ -1357,7 +1358,7 @@
             </tr>
             <tr>
               <td>
-                <dfn>@@unscopables</dfn>
+                <dfn>%Symbol.unscopables%</dfn>
               </td>
               <td>
                 *"Symbol.unscopables"*
@@ -4836,7 +4837,7 @@
       </dl>
       <emu-alg>
         1. If _input_ is an Object, then
-          1. Let _exoticToPrim_ be ? GetMethod(_input_, @@toPrimitive).
+          1. Let _exoticToPrim_ be ? GetMethod(_input_, %Symbol.toPrimitive%).
           1. If _exoticToPrim_ is not *undefined*, then
             1. If _preferredType_ is not present, then
               1. Let _hint_ be *"default"*.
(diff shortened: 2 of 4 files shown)


gibson042:
Will this interact with special treatment of `%%` by ecmarkup (e.g., for "%%Internal%%" in [spec](https://github.com/bterlson/ecmarkup/blob/a5ce61c6a8bdbd28e9c087760d450e58927b880c/test/algorithms.html) → [HTML](https://github.com/bterlson/ecmarkup/blob/a5ce61c6a8bdbd28e9c087760d450e58927b880c/test/baselines/reference/algorithms.html))? And as a meta question, do we want to _introduce_ auto-linking of well-known symbols if we're changing their representation?

ljharb:
@gibson042 i would love to see auto-linking of well-known symbols, with or without this change.

I've tested the HTML output locally and atm they don't link anywhere, before or after.

annevk:
This would also affect the HTML Standard.

cc @domenic

littledan:
Would it be possible to use a syntax more similar to `Symbol.foo`, so that it looks more similar to how JavaScript developers refer to the same thing?

ljharb:
What’d you have in mind? It would have to be apparent that it’s not a property lookup on %Symbol%, i suspect.

littledan:
Sorry, I don't have a concrete suggestion, and agree it shouldn't imply actual property access. But I would prefer that editorial churn like this would somehow inch us closer to being readable by normal programmers. `@@` is now understood by a small set of people outside TC39; using a new syntax will make it harder to understand for that see while keeping it unintuitive for the big majority.

bakkot:
My preference would be to use the %SymbolSpecies% notation and move the well-known symbols table to immediately following the intrinsics table. That way we reduce the total number of novel syntaxes in the spec. I don't actually think there is much of an important distinction between intrinsics and well-known symbols except for the cross-realm interaction. They're similar enough that it makes sense to put them near each other and use the same syntax, IMO.

ljharb:
@bakkot i initially looked at doing that, but it seemed like a loss to remove the useful info in this [section](https://tc39.github.io/ecma262/#sec-well-known-symbols).

bakkot:
[figure-2.png](https://github.com/tc39/ecma262/blob/7acacc524213058a2368b5fa751fac7ea08ba230/img/figure-2.png) also needs to be updated, since it contains the text `@@hasInstance`.

zenparsing:
I'm thinking that we should leave things the way they are for now, since `@@` is well-understood, fairly readable, and used by other specs.

michaelficarra:
Decision from the editor call today was to change `@@symbolName` notation to `%Symbol.symbolName%` for consistency with other instrinsics.

michaelficarra:
@ljharb Don't forget @bakkot's comment about the usage in figures: https://github.com/tc39/ecma262/pull/1314#issuecomment-441111136

bakkot:
Also would be good to have a NOTE near the symbol table mentioning the old notation, I think.

jmdyck:
In this PR, do well-known symbols qualify as intrinsics?

- If they do, that would require some changes to [6.1.7.4 Well-Known Intrinsic Objects](https://tc39.es/ecma262/#sec-well-known-intrinsic-objects), which indicates that intrinsics (or at least, the well-known ones) are all objects.

- If they don't, then the preamble to the "Well-Known Symbols" table shouldn't say that it lists intrinsics.

Either way, it would be nice if the definition of the %X.Y% notation didn't say that it means an object.

(For reference, here are some other discussions about the definition of _intrinsics_:
- [how-we-work PR #64](https://github.com/tc39/how-we-work/pull/64)
- [comment](https://github.com/tc39/ecma262/pull/1376#issuecomment-492257182) in PR #1376 
- issue #1540 
- issue #1703 

)

bakkot:
I think it's easier if we don't try to make them into intrinsics. Apart from being primitives, they also do not have different identities across realms. They're really a different kin of thing.

> If they don't, then the preamble to the "Well-Known Symbols" table shouldn't say that it lists intrinsics.

Agreed. I think it needn't mention that at all: just "Within this specification a well-known symbol is referred to using the standard intrinsic notation", possibly followed by "defined in section 6.1.7.4", is probably sufficient.

---

> Either way, it would be nice if the definition of the %X.Y% notation didn't say that it means an object.

Where does it say that? The only relevant text I see says

> Within this specification a reference such as %name% means the intrinsic object, associated with the current realm, corresponding to the name. A reference such as %name.a.b% means, as if the "b" property of the "a" property of the intrinsic object %name% was accessed prior to any ECMAScript code being evaluated.

Which indicates that %X% is an object (which will still be the case after this PR), but does not say that %X.Y% must be.

allenwb:
This is exactly why two distinct notations were used in the first place. Well known symbols are not intrinsics and the distinction is emphasized by using two notations.

This change seems like pointless churm that in the end blurs the important semantic distinctions.

jmdyck:
> > Either way, it would be nice if the definition of the %X.Y% notation didn't say that it means an object.
> 
> Where does it say that?

Whoops, right. I saw "means the intrinsic object" and failed to see that it only applied to the dotless form.

zenparsing:
I agree with Allen's comment and think that, in general, the use of `%X.Y%` does not result in an overall increase in spec clarity.

ljharb:
Well-known symbols are nonconfigurable nonwritable properties on an intrinsic; what's the important semantic distinction?

@bakkot @michaelficarra note added and image updated.

michaelficarra:
@ljharb Any reason this PR is a draft currently? It LGTM.

ljharb:
@michaelficarra the previous feeling on it had been that it would cause too much churn for not enough gain. Let's chat about it next week!

michaelficarra:
@ljharb I'm happy to talk about it more if you want, but we already made the call to make this change: https://github.com/tc39/ecma262/pull/1314#issuecomment-556553406. Has something changed since then?

ljharb:
oh oops :-) i guess i forgot/missed it. In that case, I’ll convert it back, and rebase!

bakkot:
Eh, revisiting this since the last time it was discussed in light of the significant amount of other nontrivial churn we've made recently or propose to make soon (#1914, #1964, #1796, #1950, #1889, #1595, etc), my preference is now to hold off on this until we've dealt with some of the higher-value stuff and let things rest for a while.

ljharb:
That’s fine too; but the work is already done and won’t likely impact those other PRs much (and I’m happy to rebase them all myself after this lands, if that’s a concern)

bakkot:
I am less concerned with rebasing and more concerned with the amount of change that readers of the spec need to keep up with over time.

littledan:
Personally, I'm really positive about this change. The @@ syntax has caused some confusing spill-over into documentation and design discussions, and it really doesn't make sense given that ES6 didn't end up including that syntax in the language. This PR adds a lot of clarity, and can help guide others towards communicating with clarity. Good work, @ljharb .

spec.html
```diff
@@ -26540,7 +26541,7 @@ <h1>Properties of the Symbol Constructor</h1>
 
       <emu-clause id="sec-symbol.asynciterator">
         <h1>Symbol.asyncIterator</h1>
-        <p>The initial value of `Symbol.asyncIterator` is the well known symbol @@asyncIterator (<emu-xref href="#table-1"></emu-xref>).</p>
+        <p>The initial value of `Symbol.asyncIterator` is the well known symbol %Symbol.asyncIterator% (<emu-xref href="#table-1"></emu-xref>).</p>
```
ExE-Boss, line 26544: These should have `<dfn>` on them, otherwise auto‑linking won’t work:
```suggestion
        <p>The initial value of `Symbol.asyncIterator` is the well known symbol <dfn>%Symbol.asyncIterator%</dfn> (<emu-xref href="#table-1"></emu-xref>).</p>
```

spec.html
```diff
@@ -40856,9 +40857,9 @@ <h1>Module Namespace Objects</h1>
     <p>A Module Namespace Object is a module namespace exotic object that provides runtime property-based access to a module's exported bindings. There is no constructor function for Module Namespace Objects. Instead, such an object is created for each module that is imported by an |ImportDeclaration| that includes a |NameSpaceImport|.</p>
     <p>In addition to the properties specified in <emu-xref href="#sec-module-namespace-exotic-objects"></emu-xref> each Module Namespace Object has the following own property:</p>
 
-    <emu-clause id="sec-@@tostringtag">
-      <h1>@@toStringTag</h1>
-      <p>The initial value of the @@toStringTag property is the String value *"Module"*.</p>
+    <emu-clause oldids="sec-@@tostringtag" id="sec-%symbol.tostringtag%">
```
ExE-Boss, line 40860: This should have a more specific ID that is unlikely to conflict:

```suggestion
    <emu-clause oldids="sec-@@tostringtag" id="sec-module-namespace-objects-tostringtag">
```
or, depending on which section ID style gets chosen:
```suggestion
    <emu-clause oldids="sec-@@tostringtag" id="sec-module-namespace-objects-%symbol.tostringtag%">
```

spec.html
```diff
@@ -40856,9 +40857,9 @@ <h1>Module Namespace Objects</h1>
...
+      <h1>%Symbol.toStringTag%</h1>
```
ExE-Boss, line 40861: This should probably be wrapped in brackets:

```suggestion
      <h1>[ %Symbol.toStringTag% ]</h1>
```

spec.html
ljharb: It will work without dfns once ecmarkup is modified to accept the new syntax, which it will be before merging this PR.

spec.html
ljharb: it won’t conflict with anything else because there will never be another top level section for this value

spec.html
ExE-Boss: **ecmarkup** already supports this syntax, since `%TypedArray.prototype%` links to <https://tc39.es/ecma262/#sec-properties-of-the-%typedarrayprototype%-object>.

ljharb: In that case i think intrinsics should all not require the dfn, unless all the rest of them have it?

ExE-Boss: All the rest of the intrinsics have `<dfn>`s, as it’s necessary for auto‑linking.

spec.html
ExE-Boss: Well, I still feel like we should fix the section ID.

ljharb: I don't see the value of that; `sec-%symbol.tostringtag%` isn't going to ever conflict with anything else. this change just makes the URL longer for no value.

spec.html
ljharb: fair enough, then these should too, thanks :-)

ExE-Boss: <https://github.com/tc39/ecma262/pull/1996> has now landed, so this needs to be rebased on top of it.

spec.html
ljharb: That's not the current convention, or the previous text; the property key is a Symbol, which is what `%Symbol.toStringTag%` is.

ljharb:
@ExE-Boss The async generator style was a mistake that hasn't been worthwhile yet to correct; the former style is the reigning convention.

spec.html
```diff
@@ -40770,9 +40771,9 @@ <h1>AsyncFunction.prototype.constructor</h1>
       </emu-clause>
 
       <emu-clause id="sec-async-function-prototype-properties-toStringTag">
```
ExE-Boss, line 40773:
```suggestion
      <emu-clause id="sec-asyncfunction.prototype-%symbol.tostringtag%" oldids="sec-async-function-prototype-properties-toStringTag">

File: spec.html [hunk 5]
```diff
@@ -39223,8 +39224,8 @@ <h1>AsyncGeneratorFunction.prototype.prototype</h1>
       </emu-clause>
 
       <emu-clause id="sec-asyncgeneratorfunction-prototype-tostringtag">
```
ExE-Boss, line 39226:
```suggestion
      <emu-clause id="sec-asyncgeneratorfunction.prototype-%symbol.tostringtag%" oldids="sec-asyncgeneratorfunction-prototype-tostringtag">
```

ljharb:
Why does the order matter?

ExE-Boss:
For consistency with the rest of the specification.

ljharb:
Right but why does this consistency matter? the only thing that ends up in the rendered output is `id`.

bakkot:
Per discussion today, let's do this, but add a sentence for each symbol somewhere (probably under each [Symbol.whatever subclause](https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-symbol.hasinstance)) to the effect of "In previous versions of this specification, this was referred to as <dfn>@@hasInstance</dfn>".

ljharb:
PR's rebased.

> but add a sentence for each symbol somewhere

it seems a bit awkward to add it there, but I can - there's already a sentence in the well-known symbols table ```Previous editions of this specification used a notation of the form @@name, where the current edition would use `%Symbol.name%` ```, but if it's important that each former form be directly greppable, then it could also be another column in the well-known symbols table. Thoughts?

bakkot:
I would like the old names to be greppable, yeah. I'm fine going with a column in that table ("previous name" or whatever) if the other editors are.

ljharb:
added it as a separate commit, so i can drop it easily if needed.

michaelficarra:
@bakkot I don't think we need to mention every single old name. If someone wants to know what happened to `@@species` and find-in-page for `@@species` doesn't turn anything up, find-in-page for `@@` will bring them to the explanation of what happened. Similarly, find-in-page for `species` will also make them aware of the notation change. That seems sufficient.

Josh-Cena:
MDN (and probably other developer references) use the `@@` notation, so hard to say this is a spec-only editorial change. I'm not against the idea in general but I'd admit it makes text much longer.

michaelficarra:
@Josh-Cena Brevity isn't that high on our list of editorial concerns.

Josh-Cena:
I understand it's not a concern for the spec, just wanted to share my opinions as the MDN editor supposing I received a change like this—because this will certainly have an impact on us. Not sure if the community churn of old browser release notes, blog posts, and developer references have been taken into account in the decision making.

zloirock:
@michaelficarra It's not about brevity. It affects a huge amount of documentation - many projects I participate in use `@@` notation. A huge amount of archive discussions where it's used will become more confusing for beginners.

ljharb:
@Josh-Cena the feedback i'd have given to MDN authors back in 2015 (or to anyone writing documentation) would have been to staunchly avoid using spec-internal notation, and to just use `Symbol.whatever` - which is what I'd suggest changing MDN to regardless, before or after this PR. It's certainly a lot of churn, which is unfortunate.

Josh-Cena:
I don't have strong opinions about changing it for MDN for the future, just wondering if the editors have thought about it for materials from the past.

So would you recommend `Array.prototype[Symbol.iterator]` or `Array.prototype[%Symbol.iterator%]`, for both pedagogical and pedantic concerns?

ljharb:
The former, exclusively. Spec-internal notation may change at any time based on the whims of the editors, and shouldn’t be relied upon externally. I’ve made that same mistake in a few of my packages.

Josh-Cena:
`%Symbol.iterator%` is very, very unlikely to change, and may even be part of the language with `getIntrinsic` though.

ljharb:
@Josh-Cena nope, the getIntrinsic proposal (which I'm championing) explicitly avoids exposing the percent signs to users, for the reasons indicated.

Josh-Cena:
> explicitly avoids exposing the percent signs to users

Really? That's not the impression I get from either the README or the spec:

> ```js
> const $String = getIntrinsic('%String%');
> ```
>
> Provide a new global or static function, `getIntrinsic`, that takes a string argument and effectively replicates the `get-intrinsic` package API, as well as the specification's internal `%a.b.c%` notation. It would return the "original" value being requested.

Was there a recent API change?

ljharb:
Looks like I just forgot to update the readme, I'll do so now :-)

ljharb:
Updated text per editor call; this is now only blocked by editor review, and ecmarkup supporting Symbol intrinsics in AO h1s.

Jack-Works:
What about things like `%RegExp.prototype[@@match]%`? Does it become `%RegExp.prototype[%Symbol.match%]%`?

ljharb:
Yep, exactly.

spec.html
```diff
@@ -1203,7 +1203,8 @@ <h1>The Symbol Type</h1>
...
-        <p>Within this specification a well-known symbol is referred to by using a notation of the form @@name, where “name” is one of the values listed in <emu-xref href="#table-well-known-symbols"></emu-xref>.</p>
+        <p>Within this specification a well-known symbol is referred to using the standard <emu-xref href="#sec-well-known-intrinsic-objects">intrinsic notation</emu-xref> where the intrinsic is one of the values listed in <emu-xref href="#table-well-known-symbols"></emu-xref>.</p>
+        <emu-note>Previous editions of this specification used a notation of the form @@name, where the current edition would use `%Symbol.name%` In particular, the following names were used: @@asyncIterator, @@hasInstance, @@isConcatSpreadable, @@iterator, @@match, @@matchAll, @@replace, @@search, @@species, @@split, @@toPrimitive, @@toStringTag, @@unscopables)</emu-note>
```
syg, line 1207:
```suggestion
        <emu-note>Previous editions of this specification used a notation of the form @@name, where the current edition would use `%Symbol.name%`. In particular, the following names were used: @@asyncIterator, @@hasInstance, @@isConcatSpreadable, @@iterator, @@match, @@matchAll, @@replace, @@search, @@species, @@split, @@toPrimitive, @@toStringTag, and @@unscopables.</emu-note>
```

syg (approved):
lgtm % wording

bakkot:
OK, if this is rebased and updated to incorporate ecmarkup 19.0.0 it should be ready to go.

ljharb:
k, should be all rebased and up to date, and the SVG updated (i had to move some of the coordinates around to make the longer text fit nicely)

spec.html
```diff
@@ -30616,7 +30617,7 @@ <h1>Properties of the Symbol Constructor</h1>
 
       <emu-clause id="sec-symbol.asynciterator">
         <h1>Symbol.asyncIterator</h1>
-        <p>The initial value of `Symbol.asyncIterator` is the well-known symbol @@asyncIterator (<emu-xref href="#table-well-known-symbols"></emu-xref>).</p>
+        <p>The initial value of `Symbol.asyncIterator` is the well known symbol %Symbol.asyncIterator% (<emu-xref href="#table-well-known-symbols"></emu-xref>).</p>
```
jmdyck, line 30620: Reinstate hyphen, just for consistency:
```suggestion
        <p>The initial value of `Symbol.asyncIterator` is the well-known symbol %Symbol.asyncIterator% (<emu-xref href="#table-well-known-symbols"></emu-xref>).</p>
```

zloirock:
I still hope that it will not be merged for [this reason](https://github.com/tc39/ecma262/pull/1314#issuecomment-1728529600).

img/figure-2.svg
Josh-Cena, line 1: FYI: the object names are no longer centered, and the "Function" -> "Function.prototype" arrow is not sticking to the edge of the box.

ljharb:
https://github.com/tc39/ecma262/pull/1314#issuecomment-1728535369

> the feedback i'd have given to MDN authors back in 2015 (or to anyone writing documentation) would have been to staunchly avoid using spec-internal notation, and to just use `Symbol.whatever` - which is what I'd suggest changing MDN to regardless, before or after this PR. It's certainly a lot of churn, which is unfortunate.

Josh-Cena:
Lots of historical materials will stick around and will confuse readers. Again I'm willing to do the work so MDN aligns with the spec, but the ship has sailed (as tc39 always says).

ljharb:
True enough, but that's often unavoidable, and in this case, it'll be far more confusing if we have a spec with both decorators and `@@` in it.

Josh-Cena:
Decorators don't use `@@` though, and code entities are typeset differently in the spec, if they are used at all. I see it can be confusing for MDN (which is another thing to think about, for us!), but not for the spec.

jmdyck (changes requested):

spec.html
```diff
@@ -30838,9 +30839,9 @@ <h1>
         </emu-clause>
       </emu-clause>
 
-      <emu-clause id="sec-symbol.prototype-@@toprimitive">
-        <h1>Symbol.prototype [ @@toPrimitive ] ( _hint_ )</h1>
-        <p>This method is called by ECMAScript language operators to convert a Symbol object to a primitive value.</p>
+      <emu-clause oldids="sec-symbol.prototype-@@toprimitive" id="sec-symbol.prototype-%symbol.toprimitive%">
+        <h1>Symbol.prototype [ %Symbol.toPrimitive% ] ( _hint_ )</h1>
+        <p>This function is called by ECMAScript language operators to convert a Symbol object to a primitive value.</p>
```
jmdyck, line 30844: Accidental revert of PR #2592?
```suggestion
        <p>This method is called by ECMAScript language operators to convert a Symbol object to a primitive value.</p>
```

jmdyck (commented):
You gave a new clause id to every intrinsic with the property key `%Symbol.foo%`, except for:
- `RegExp.prototype [ %Symbol.matchAll% ]`
- `%AsyncIteratorPrototype% [ %Symbol.asyncIterator% ]`
- `AsyncFunction.prototype [ %Symbol.toStringTag% ]`

Did you intend to skip those?

Apart from that, looks okay to me.

ljharb:
I didn't do it intentionally, it's just that I only changed the ones that had `@@` in them.

I could change them to be consistent, but I'm not sure if that's worth the churn + oldids.

jmdyck:
> I didn't do it intentionally, it's just that I only changed the ones that had `@@` in them.

That was my guess at first, but then I noticed that you had introduced new ids to replace
- `sec-asyncgeneratorfunction-prototype-tostringtag`
- `sec-asyncgenerator-prototype-tostringtag`

which don't have `@@`.

ljharb:
Guess I was just being inconsistent in my application then :-) I'll update them to match.

michaelficarra (approved):
Still LGTM.

spec.html
```diff
@@ -46398,8 +46399,8 @@ <h1>The %AsyncIteratorPrototype% Object</h1>
         <p>All objects defined in this specification that implement the AsyncIterator interface also inherit from %AsyncIteratorPrototype%. ECMAScript code may also define objects that inherit from %AsyncIteratorPrototype%. The %AsyncIteratorPrototype% object provides a place where additional methods that are applicable to all async iterator objects may be added.</p>
       </emu-note>
 
-      <emu-clause id="sec-asynciteratorprototype-asynciterator">
-        <h1>%AsyncIteratorPrototype% [ @@asyncIterator ] ( )</h1>
+      <emu-clause oldids="sec-asynciteratorprototype-asynciterator" id="sec-asynciteratorprototype-%symbol.asynciterator%">
```
jmdyck, line 46402:
```suggestion
      <emu-clause oldids="sec-asynciteratorprototype-asynciterator" id="sec-%asynciteratorprototype%-%symbol.asynciterator%">
```
because we usually include the base object's percents in the id. (The status quo id is anomalous on this point.)

michaelficarra: Either one WFM. I'm not too worried about consistency in clause IDs.

Josh-Cena:
@ljharb Out of curiosity: what's up with "luke_skywalker"?

ljharb:
@Josh-Cena this PR kills at-at's

Josh-Cena:
Oh! lmao 😄

domenic:
Just want to drop in with an extra thanks to @ljharb for doing the HTML and Web IDL PRs. That kind of cross-ecosystem citizenship is much appreciated.

linusg:
PR for ECMA-402: https://github.com/tc39/ecma402/pull/905

ljharb:
whoops, i should have made that one too. thanks!

PR #1317, opened 2018-09-28
Editorial: Update date-time format language

gibson042:
* Use ISO 8601 vocabulary
* Move interpretation of absent date time elements into Date.parse


littledan:
Given that this change is editorial and approved by both editors, can we let it land?

ljharb:
@littledan To ensure that any issues that need to be backported into ES2019 land cleanly, we're not merging anything new for the time being. We'll address this during the editors' update in plenary.

PR #1325, opened 2018-10-10
Editorial: Better relate ECMAScript time values to POSIX and UTC

gibson042:
This should be the last of my editorial Date changes in advance of proposing normative updates for https://github.com/gibson042/ecma262-proposal-uniform-interchange-date-parsing .


spec.html
```diff
@@ -26506,14 +26506,11 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
+        <p>Time measurement in ECMAScript is analogous to time measurement in POSIX, in particular sharing an epoch of midnight at the beginning of 01 January, 1970 UTC, use of the proleptic Gregorian calendar, and an accounting of every day as comprising exactly 86,400 seconds (each of which is itself 1000 milliseconds). An ECMAScript <dfn>time value</dfn> is a Number representing an instant in time with millisecond precision as a signed count of such milliseconds since the epoch, or *NaN* representing no specific instant. Alignment of time values to UTC values (deviating by no more than 1000 milliseconds) is maintained by equating every multiple of 86,400,000 with the start of a UTC day, ignoring leap seconds and the duration variance they entail.</p>
```
jmdyck: If you ignore UTC's leap seconds, then you're not aligning to UTC.

gibson042: Agreed, but I want to make clear that every UTC day is a fresh start. How would you feel about "approximate alignment to UTC"?

jmdyck: Also, I think "equating every multiple of 86,400,000 with the start of a UTC day" is a requirement that the current spec does not make, so I'm not sure this change qualifies as editorial.

E.g., I think one possible interpretation of the current spec is that each ES day is exactly 86,400 SI seconds long, so (due to UTC's leap seconds) each ES start-of-day currently occurs 27 seconds before the UTC start-of-day.

Mind you, I wouldn't be surprised if all real implementations do have ES and UTC start-of-day coincide, in which case this might as well be editorial.

gibson042: No, ECMAScript matches POSIX in its use of SI seconds with discontinuities for daily alignment with UTC. Both behaviors can be derived from the abstract operations of 20.3.1 such as [TimeWithinDay](https://tc39.github.io/ecma262/#sec-day-number-and-time-within-day) and [MakeDate](https://tc39.github.io/ecma262/#sec-makedate), but I'm trying to make them explicit in the introduction (though I'm open to ideas on how to do so).

jmdyck: > ECMAScript matches POSIX in its use of SI seconds with discontinuities for daily alignment with UTC.

If that were the case, I'd expect the spec to at least mention SI seconds and discontinuities, but it doesn't.

Do all ECMAScript implementations exhibit discontinuities in ES time?

> Both behaviors can be derived from the abstract operations of 20.3.1 such as TimeWithinDay and MakeDate

Those two operations merely confirm that there are 86,400 ES seconds per ES day. To make conclusions re SI seconds or discontinuities, I think you'd need the spec to make a stronger statement about the relationship between ES time and UTC.

gibson042: > I'd expect the spec to at least mention SI seconds and discontinuities, but it doesn't.

That's the kind of thing that I'm trying to address with this PR.

> Those two operations merely confirm that there are 86,400 ES seconds per ES day. To make conclusions re SI seconds or discontinuities, I think you'd need the spec to make a stronger statement about the relationship between ES time and UTC.

The relationship really is already there, but buried so deep throughout 20.3 that you can't see it. I'll demonstrate with respect to the very first post-epoch UTC leap second. `1972-06-30T23:59:60Z` has exactly the same time value as `1972-07-01T00:00:00Z`, and you can observe how the abstract operations implicitly treat time values as _full-UTC-days-since-epoch × millisecondsPerDay + min(milliseconds-since-start-of-UTC-day, 86400000)_, which forces an interpretation of that time value as the start of 01 July 1972 UTC rather than the leap second at the end of June, even though one millisecond back is still part of the 1972-06-30T23:59:59Z second.
```js
const leap = Date.UTC(1972, 6 - 1, 30,  23,59,60);
// → TimeClip(MakeDate(MakeDay(1972, 5, 30), MakeTime(23, 59, 60, 0)))
// → TimeClip(MakeDate(911, 86400000))
// → TimeClip(78796800000)
// → 78796800000

let O = new Date(leap);
O.toUTCString();
// → Table46NameFromNumber(WeekDay(78796800000)) + "," + "\u0020" +
//     ZeroPad2(DateFromTime(78796800000)) + "\u0020" +
//     Table47NameFromNumber(MonthFromTime(78796800000)) + "\u0020" +
//     ZeroPad4(YearFromTime(78796800000)) + "\u0020" +
//     TimeString(78796800000)
// → Table46NameFromNumber(6) + "," + "\u0020" +
//     ZeroPad2(1) + "\u0020" +
//     Table47NameFromNumber(6) + "\u0020" +
//     ZeroPad4(1972) + "\u0020" +
//     ZeroPad2(HourFromTime(78796800000)) + ":" +
//       ZeroPad2(MinFromTime(78796800000)) + ":" +
//       ZeroPad2(SecFromTime(78796800000)) +
//       "\u0020" + "GMT"
// → "Sat" + "," + "\u0020" +
//     "01" + "\u0020" +
//     "Jul" + "\u0020" +
//     "1972" + "\u0020" +
//     ZeroPad2(0) + ":" + ZeroPad2(0) + ":" + ZeroPad2(0) + "\u0020" + "GMT"
// → "Sat, 01 Jul 1972 00:00:00 GMT"

O = new Date(leap - 1);
O.toUTCString();
// → Table46NameFromNumber(WeekDay(78796799999)) + "," + "\u0020" +
//     ZeroPad2(DateFromTime(78796799999)) + "\u0020" +
//     Table47NameFromNumber(MonthFromTime(78796799999)) + "\u0020" +
//     ZeroPad4(YearFromTime(78796799999)) + "\u0020" +
//     TimeString(78796799999)
// → Table46NameFromNumber(5) + "," + "\u0020" +
//     ZeroPad2(30) + "\u0020" +
//     Table47NameFromNumber(5) + "\u0020" +
//     ZeroPad4(1972) + "\u0020" +
//     ZeroPad2(HourFromTime(78796799999)) + ":" +
//       ZeroPad2(MinFromTime(78796799999)) + ":" +
//       ZeroPad2(SecFromTime(78796799999)) +
//       "\u0020" + "GMT"
// → "Fri" + "," + "\u0020" +
//     "30" + "\u0020" +
//     "Jun" + "\u0020" +
//     "1972" + "\u0020" +
//     ZeroPad2(23) + ":" + ZeroPad2(59) + ":" + ZeroPad2(59) + "\u0020" + "GMT"
// → "Fri, 30 Jun 1972 23:59:59 GMT"
```

jmdyck: Yup, that's what the operations do, but that's not enough to establish the relationship between ES time and UTC.

I think you're assuming that ES time relates to UTC something like this:

    Given an ES time value _tv_,
    let "<foo> GMT" be the result of ToDateString(_tv_) when LocalTZA() is zero.
    Then "<foo>Z" denotes a UTC time
    (except when a negative leap-second occurs, which hasn't ever happened yet),
    and that UTC time is what we define 'the UTC time denoted by _tv_' to be.

(Or, very roughly, stringify `_tv_` and read that as if it's a UTC time.)

Which seems like a reasonable assumption, but I don't think it's required by the spec.

Perhaps it's even the consensus interpretation within TC39, but I haven't found evidence of that. If you could point me to a consensus, that'd be great.

So I ask again: Do all ECMAScript implementations exhibit discontinuities in ES time?
Is an ES implementation that uses "leap smears" deemed non-conformant?

gibson042: I just demonstrated how ECMAScript requires overlap discontinuities where UTC has a positive leap second, and here's an example of gap discontinuity requirements at negative leap seconds:
```js
Date.parse(`${dateOfNegativeLeap}T23:59:58Z`) + 2000 ===
  Date.parse(`${dateFollowingNegativeLeap}T00:00:00Z`);
// → true
```

ECMAScript time values are literally [defined in terms of UTC](https://tc39.github.io/ecma262/#sec-time-values-and-time-range), as is the correlation between time values and strings representing date and time of day. It would absolutely be non-conforming to insert or remove milliseconds to or from a plain interpretation of any date-time string, whether in parsing or in string generation.

Leap smearing deals with the relationship between _external_ time and reported time (e.g., misreporting an actual time of T23:59:60.500Z as T23:59:59.750Z), and as such is completely independent of pure functional abstract operations involving time values.

jmdyck: > ECMAScript time values are literally defined in terms of UTC

ES time value *zero* is equated with a particular UTC instant, yes, but that's not enough to define the relationship between ES time and UTC.

> as is the correlation between time values and strings representing date and time of day.

No, that correlation is not defined in terms of UTC, it's defined by the relevant algorithms in the ES spec.

> Neither kind of value is ever internally discontinuous, but the relationship between them is. I just demonstrated examples at positive leap seconds [...], and here's an example at negative leap seconds (proof of existence for ES time values that don't map to any UTC value):

I agree that if there were ever a negative leap second, then you could write a date-string that ES time would happily map to an ES time value, but that UTC would deem invalid. But that doesn't necessarily mean that the ES time value doesn't map to a UTC value, because we don't know that same-date-string is the basis of the relationship between ES time and UTC. You appear to be assuming that it is, and I'm saying that I don't think that assumption is warranted by the spec. (And again, maybe that assumption is the TC39 consensus, and if so it'd be great to see that documented somewhere.)

Consider an analogy from the world of calendars: Julian vs Gregorian. (Note that this has nothing to do with how dates are represented in ES.) If you write "Feb 29, 1900", that's meaningful in the Julian calendar but not the Gregorian. But that *doesn't* mean that "Feb 29, 1900" Julian doesn't map to any any Gregorian date; in fact it maps to "March 13, 1900" Gregorian (because same-date-string is not the basis of the relationship between the two calendars).

> It would absolutely be non-conforming to insert or remove milliseconds to or from a plain interpretation of any date-time string, whether in parsing or in string generation.

I'm not quite sure what you mean by "plain interpretation", but if you mean that it would be non-conforming to do date parsing or stringifying other than as required in the spec, I agree.

> Leap smearing deals with the relationship between external time and reported time [...], and as such is completely independent of pure functional abstract operations involving time values.

Again, I agree. But I'm not sure if that means you think an ES implementation that uses leap smears can be conformant.

gibson042: > maybe that assumption is the TC39 consensus, and if so it'd be great to see that documented somewhere

It is not clear to me what consensus you're seeking. The relevant algorithms already require implementations to equate a negative leap second <i>**YYYY**</i>-<i>**MM**</i>-<i>**DD**</i>T23:59:59Z (which is not part of the UTC timeline) as 86_400_000 × _**n**_ + 86_399_000 where _**n**_ is an integer, and more importantly require a one-to-one correspondence between every integer in the range [86_400_000 × _**n**_, 86_400_000 × (_**n**_ + 1) − 1] with a distinct <i>**YYYY**</i>-<i>**MM**</i>-<i>**DD**</i>T<i>HH</i>:<i>mm</i>:<i>ss</i>.<i>sss</i>Z (where <i>**YYYY**</i>-<i>**MM**</i>-<i>**DD**</i> is fixed by _**n**_) for _every_ integer _**n**_ in [−100_000_000, 99_999_999].

Note that the second requirement maps the full range of every ECMAScript time value into a distinct UTC value, but leaves no room for positive leap seconds T23:59:60Z—they simply don't exist in the ECMAScript timeline, just as they don't exist in the POSIX timeline.

> I'm not sure if that means you think an ES implementation that uses leap smears can be conformant.

What could "uses leap smears" possibly mean in the context of these abstract operations on discrete values? ECMAScript time value 86_400_000 × (_**n**_ + 1) − 1 is necessarily equated with UTC value <i>**YYYY**</i>-<i>**MM**</i>-<i>**DD**</i>T23:59:59.999Z, whether or not <i>**YYYY**</i>-<i>**MM**</i>-<i>**DD**</i>T23:59:60Z is part of the UTC timeline. An implementation could add functionality to convert between ECMAScript time values or Dates and some other time scale (such as Hebrew, Julian, TAI, or even actual leap-second-inclusive UTC), but that has no bearing on the required behavior of time values and Dates in and of theirselves that I'm addressing here.

What exactly is it that you would have changed in the text of this PR?

jmdyck: > It is not clear to me what consensus you're seeking.

Yes, I realize that, but I'm not sure I can make it any clearer. I'll try though.


> What could "uses leap smears" possibly mean in the context of these abstract operations on discrete values?

That isn't the context in which it means something.

Instead, I'll go way back to "the question of exactly what happens to ES time on a day when UTC has a leap second." Say you call Date.now() every 200 milliseconds for a few seconds around the time that UTC experiences a leap second. What behavior does the ES spec require of a conforming implementation?  
 - steadily increasing values, each about 200 more than the previous;
 - mostly as above, but with one additional jump of 1000 (up or down, depending on the sign of the leap second); or
 - implementation-defined/implementation-dependent.

(The first is what I'd expect from an implementation that uses a time source that implements leap smears. The second is supposedly what you'd get from an implementation that uses a time source that implements POSIX.1 time.)

> ECMAScript time value 86_400_000 × (n + 1) - 1 is necessarily equated with UTC value YYYY-MM-DDT23:59:59.999Z

It's necessarily mapped to such a *date-string*, but:
 - whether that date-string represents a UTC time is, in the general case, unknown; and
 - in the case that it does represent a UTC time, whether the ES time value "is equated to" that UTC time is unclear (I claim). 

> An implementation could add functionality to convert between ECMAScript time values and [...] actual leap-second-inclusive UTC ...

The latter phrase is telling, in that it acknowledges that the "UTC" that ES time values supposedly convert to is not "actual UTC". If one excludes leap-seconds from UTC, the result is *not* UTC, and shouldn't be called UTC.

> What exactly is it that you would have changed in the text of this PR?

I'd like 20.3.1.1 to say whether there is any required relation between ES time and "external" time, and if so, what. (And then, depending on that, maybe other things.)

gibson042: > Say you call Date.now() every 200 milliseconds for a few seconds around the time that UTC experiences a leap second. What behavior does the ES spec require of a conforming implementation?

That question is unrelated to the content of this PR, but i don't believe that the spec requires implementations to even have access to an _accurate_ clock, let alone one with any particular behavior around leap seconds. It only requires Date.now to always return a Number that is a valid time value.

> whether the ES time value "is equated to" that UTC time is unclear (I claim).

Date.now, Date.parse, Date.UTC, and the abstract operation UTC all include language like "the time value designating the UTC date and time" and use the abstract operations I've been referencing to convert between calendar datetimes and time values. The intention is clearly to map them directly, even though each timeline includes values that the other lacks (hence the discontinuities). It's exactly the same compromise made by POSIX, and for exactly the same reasons (static deterministic date-time arithmetic without access to a necessarily-dynamic table of leap seconds). _That_ is what I'm trying to convey here.

> I'd like 20.3.1.1 to say whether there is any required relation between ES time and "external" time, and if so, what.

Like POSIX time values, ES time values have a piecewise linear relationship with UTC. The specifics of that relationship are detailed in abstract operations, but summarized as an accounting of every UTC day having exactly 86400000 milliseconds. Do you dispute that? Do you believe it is not sufficiently clear in thr proposed text?

I am not covering implementation use of any clock in this PR, but believe that there is no requirement to even have one (e.g., a `Date.now` could always return `NaN` and still be conforming).

jmdyck: > > Say you call Date.now() every 200 milliseconds for a few seconds around the time that UTC experiences a leap second. What behavior does the ES spec require of a conforming implementation?
>
> That question is unrelated to the content of this PR,

The blurb for Date.now() says: `The *now* function returns a Number value that is the time value designating the UTC date and time of the occurrence of the call to *now*.` So the above question  appears to be about how ES time values relate to UTC, and so would seem to be covered by the title of this PR. Which is not to say that this PR is obliged to answer it, just that it's not unreasonable to expect that it would. (And I hope some PR does answer it.)


> but i don't believe that the spec requires implementations to even have access to an accurate clock, let alone one with any particular behavior around leap seconds. It only requires Date.now to always return a Number that is a valid time value.

It could be that that's the intent of TC39, but a phrase like `the UTC date and time of the occurrence of the call` seems, on the face of it, to require the implementation to have access to an accurate UTC clock (or to a clock that it can convert to an accurate UTC clock). So I think it's a point that could stand some clarification. 

> > whether the ES time value "is equated to" that UTC time is unclear (I claim).
>
> Date.now, Date.parse, Date.UTC, and the abstract operation UTC(t) all include language like "the time value designating the UTC date and time"

 - Date.parse says `it returns a Number, the UTC time value corresponding to the date and time [extracted from the string argument]`, which is clearly a misuse of the term UTC: ES time values are not UTC. (Similarly: algorithm references to `the time value (UTC) identifying the current time`.)

 - Date.UTC says (in a Note) that it `interprets the arguments in UTC rather than as local time`.

 - UTC(t) says it `converts t from local time to UTC`.

For all of these, I get the impression that "UTC" is only used to mean (roughly) the zero-offset timezone, not the real-world time-scale.


> and use the abstract operations I've been referencing to convert between time values and ISO 8601 (which is based upon UTC) calendar date and time of day. The intention is clearly to map them directly, even though each timeline includes values that the other lacks (hence the discontinuities).

In my opinion, the current spec is sufficiently muddled concerning UTC that I can't agree that its intention is clear.

But if this PR makes it clear, that'll be great.


> > I'd like 20.3.1.1 to say whether there is any required relation between ES time and "external" time, and if so, what.
>
> Like POSIX time values, ES time values have a piecewise linear relationship with UTC.

Sorry, I didn't mean a relation between two time systems, but rather a relation between ES's idea of *current* time and an external current time. I.e., wording that would answer my question above involving Date.now().


> The specifics of that relationship are detailed in abstract operations, but summarized as an accounting of every UTC day having exactly 86400000 milliseconds. Do you dispute that? Do you believe it is not sufficiently clear in thr proposed text?

I think saying "every UTC day having exactly 86400000 milliseconds" would confuse/annoy people who understand leap seconds, but the actual proposed text is better.

However, when the text says `An ECMAScript <dfn>time value</dfn> is a Number representing an instant in time with millisecond precision as a signed count of such milliseconds since the epoch`, that implies a time-scale that's monotonic in the real world, which contradicts what you want to convey. So that sentence needs to be toned down.

Note that POSIX says "Seconds Since the Epoch" is merely "a value that *approximates* the number of seconds that have elapsed since the Epoch" (emphasis mine). (Moreover, it doesn't explicitly tie the Epoch to a particular real-world time-point.)


> I am not covering implementation use of any clock in this PR, but believe that there is no requirement to even have one (e.g., a Date.now could always return NaN and still be conforming).

Okay, but in that case, the other text (that doesn't talk about a clock) shouldn't mislead the reader into thinking that an accurate clock *is* required. (See above re "the UTC date and time of the occurrence of the call".)

gibson042: Your points about clarifying the requirement (or lack thereof) for implementations to have access clocks with certain characteristics are valid, but out of scope for this PR (which merely attempts to clarify the relationship between ECMAScript time values and points on the UTC timeline).

I have updated the text to define time value as a linear sum of "day number" and "time within the day" rather than a pure count of milliseconds, corresponding with treatment in the immediately following operations. It's certainly more complex than before, but should resolve some misunderstandings. We can tackle requirements relating to "current time" next.

spec.html
```diff
@@ -26507,7 +26507,7 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
+		<p>An ECMAScript <dfn>time value</dfn> is a Number representing an instant in time to millisecond precision, or *NaN* representing no specific instant. A time value that is a multiple of <emu-eqn>24 &times; 60 &times; 60 &times; 1000 = 86,400,000</emu-eqn> (i.e., is equal to 86,400,000 &times; _d_ for some integer _d_) represents the instant at the start of the UTC day that follows the epoch by _d_ whole UTC days (preceding the epoch for negative _d_). Every other finite time value _t_ is defined relative to the greatest preceding time value _s_ that is such a multiple, and represents the instant that occurs _t_ &minus; _s_ milliseconds after the instant represented by _s_. Time values do not account for UTC leap seconds, but this definition yields piecewise alignment with UTC, discontinuities only at leap second boundaries, and zero difference outside of leap seconds.</p>
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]
jmdyck: Would it be useful to be more explicit about leap seconds? Something like:

 - For any instant within a positive leap second, there is no time value that represents that  instant.
- For any instant within the second immediately following a negative leap second, there are 2 time values that represent that instant.

(If so, it might be worthwhile to pull all the leap second stuff to a separate para.)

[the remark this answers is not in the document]

gibson042:
@ljharb et al.: This PR appears to have stalled, but I'd really like to see it land. My goal with it is just to clarify the piecewise-discontinuous relationship between ECMAScript time values and the UTC timeline, and I believe that goal has been met (with the changes marked Editorial because that relationship is already inherent in the algorithms of section 20.3, I'm just trying to better express it in prose). @jmdyck seems to agree that I have met that goal, but is also concerned with the relationship between ECMAScript time values and _external_ time on Earth, in particular pertaining to the requirement (or lack thereof) for implementations to have access to clocks with certain characteristics (accuracy being most notable among them). I think those concerns are valid, but out of scope for this PR, since it is limited to abstract rather than physical relationships.

So where do we go from here? Should this be tagged "needs consensus" and discussed in plenary?

ljharb:
@gibson042 sorry for the oversight. if you wouldn't mind rebasing this PR, we'll take a look at it this week.

gibson042:
Done.

spec.html
```diff
@@ -27999,14 +27999,13 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
+        <p>An ECMAScript <dfn>time value</dfn> is a Number, either a finite integer representing an instant in time to millisecond precision or *NaN* representing no specific instant. A time value that is a multiple of <emu-eqn>24 &times; 60 &times; 60 &times; 1000 = 86,400,000</emu-eqn> (i.e., is equal to 86,400,000 &times; _d_ for some integer _d_) represents the instant at the start of the UTC day that follows the epoch by _d_ whole UTC days (preceding the epoch for negative _d_). Every other finite time value _t_ is defined relative to the greatest preceding time value _s_ that is such a multiple, and represents the instant that occurs within the same UTC day as _s_ but follows it by _t_ &minus; _s_ milliseconds.</p>
+        <p>Time values do not account for UTC leap seconds&mdash;there are no time values representing instants within positive leap seconds, but there are time values representing instants removed from the UTC timeline by negative leap seconds. However, the definition of time values nonetheless yields piecewise alignment with UTC, discontinuities only at leap second boundaries, and zero difference outside of leap seconds.</p>
```
jmdyck: I think that "but" should be an "and", since it adds to the ways in which time values don't handle leap seconds.

I think there's a word missing before "discontinuities": maybe "with" or "having".

spec.html
```diff
@@ -27999,14 +27999,13 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
+        <p>Time measurement in ECMAScript is analogous to time measurement in POSIX, in particular sharing definition in terms of the proleptic Gregorian calendar, an epoch of midnight at the beginning of 01 January, 1970 UTC, and an accounting of every day as comprising exactly 86,400 seconds (each of which is 1000 milliseconds long).</p>
+        <p>An ECMAScript <dfn>time value</dfn> is a Number, either a finite integer representing an instant in time to millisecond precision or *NaN* representing no specific instant. A time value that is a multiple of <emu-eqn>24 &times; 60 &times; 60 &times; 1000 = 86,400,000</emu-eqn> (i.e., is equal to 86,400,000 &times; _d_ for some integer _d_) represents the instant at the start of the UTC day that follows the epoch by _d_ whole UTC days (preceding the epoch for negative _d_). Every other finite time value _t_ is defined relative to the greatest preceding time value _s_ that is such a multiple, and represents the instant that occurs within the same UTC day as _s_ but follows it by _t_ &minus; _s_ milliseconds.</p>
```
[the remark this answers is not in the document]
gibson042: Since the more restrictive range limitation only applies to the [[DateValue]] slot rather than to all time values, I'm going to leave this as-is and improve things further with #1564.

spec.html
gibson042: Accepted both suggestions.

spec.html
```diff
@@ -27999,14 +27999,13 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
+        <p>Time values do not account for UTC leap seconds&mdash;there are no time values representing instants within positive leap seconds, but there are time values representing instants removed from the UTC timeline by negative leap seconds. However, the definition of time values nonetheless yields piecewise alignment with UTC, discontinuities only at leap second boundaries, and zero difference outside of leap seconds.</p>
+        <p>A Number can exactly represent all integers from -9,007,199,254,740,992 to 9,007,199,254,740,992 (<emu-xref href="#sec-number.min_safe_integer"></emu-xref> and <emu-xref href="#sec-number.max_safe_integer"></emu-xref>). A time value supports a slightly smaller range of -8,640,000,000,000,000 to 8,640,000,000,000,000 milliseconds. This yields a supported time value range of exactly -100,000,000 days to 100,000,000 days relative to midnight at the beginning of 01 January, 1970 UTC.</p>
         <p>The exact moment of midnight at the beginning of 01 January, 1970 UTC is represented by the time value *+0*.</p>
```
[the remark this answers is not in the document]
gibson042: I'll leave this and let #1564 do the rearranging.

PR #1326, opened 2018-10-11
Editorial: Correct variable terminology

TimothyGu:
The spec "variables" are not variables, but aliases as defined in [5.2 Algorithm Conventions](https://tc39.github.io/ecma262/#sec-algorithm-conventions).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4055,7 +4055,7 @@
           1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
         </emu-alg>
 
-        <p>where all variables in the above steps, with the exception of _completion_, are ephemeral and visible only in the steps pertaining to Await.</p>
+        <p>where all aliases in the above steps, with the exception of _completion_, are ephemeral and visible only in the steps pertaining to Await.</p>
 
         <emu-note>
           <p>Await can be combined with the `?` and `!` prefixes, so that for example</p>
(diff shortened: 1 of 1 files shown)


zenparsing:
The only other such usage of "variable" occurs in reference to `InputLength` at https://tc39.github.io/ecma262/#sec-pattern.

Should that usage be fixed as well?

TimothyGu:
@zenparsing Should be fixed.

spec.html
```diff
@@ -29434,7 +29434,7 @@ <h1>Notation</h1>
...
+            A <em>Continuation</em> procedure is an internal closure (i.e. an internal procedure with some arguments already bound to values) that takes one State argument and returns a MatchResult result. If an internal closure references aliases which are bound in the function that creates the closure, the closure uses the values that these aliases had at the time the closure was created. The Continuation attempts to match the remaining portion (specified by the closure's already-bound arguments) of the pattern against _Input_, starting at the intermediate state given by its State argument. If the match succeeds, the Continuation returns the final State that it reached; if the match fails, the Continuation returns ~failure~.
```
ljharb: i'm confused; closures reference variables, how can they reference aliases?

zenparsing: My understanding is that the "continuations" in that section aren't JS closures, and the bindings that they capture aren't JS variables. They are the EcmaSpeak "aliases" instead (e.g. [Disjunction](https://tc39.github.io/ecma262/#sec-disjunction)). Does that sound right?

TimothyGu: Precisely as @zenparsing said. See https://tc39.github.io/ecma262/#sec-algorithm-conventions, where “alias” is defined.

@allenwb do you have any insights?

ljharb:
Still hoping for some clarification re https://github.com/tc39/ecma262/pull/1326#discussion_r224886849

spec.html
jmdyck: This paragraph dates back to the 3rd edition. At the time, the spec used the word "closure" only to denote a spec artifact (what we might now think of as a dynamic abstract operation), used to describe the semantics of RegExps, and not to describe the behavior of ES functions. (The spec started to use the term more generally in the 5th edition, but it still has that specific meaning within the RegExp semantics.)

The phrase "the function that creates the closure" compounds the confusion. It would probably help to change that occurrence of "function" to "algorithm".

bakkot: https://github.com/tc39/ecma262/pull/1889 defined abstract closures formally, resolving this issue.

PR #1335, opened 2018-10-25
Add a link to the collection of historical editions

gibson042:
It was surprisingly hard to find when researching https://twitter.com/gibson042/status/1055265958326099968 .


spec.html
```diff
@@ -72,8 +72,8 @@ <h1>Contributing to this Specification</h1>
 </div>
 <emu-intro id="sec-intro">
   <h1>Introduction</h1>
-  <p>This Ecma Standard defines the ECMAScript 2019 Language. It is the tenth edition of the ECMAScript Language Specification. Since publication of the first edition in 1997, ECMAScript has grown to be one of the world's most widely used general-purpose programming languages. It is best known as the language embedded in web browsers but has also been widely adopted for server and embedded applications.</p>
-  <p>ECMAScript is based on several originating technologies, the most well-known being JavaScript (Netscape) and JScript (Microsoft). The language was invented by Brendan Eich at Netscape and first appeared in that company's Navigator 2.0 browser. It has appeared in all subsequent browsers from Netscape and in all browsers from Microsoft starting with Internet Explorer 3.0.</p>
+  <p>This Ecma Standard defines the ECMAScript 2019 Language. It is the tenth edition of the ECMAScript Language Specification. Previous "historical" editions are available at <a href="http://ecma-international.org/publications/standards/Ecma-262-arch.htm">http://ecma-international.org/publications/standards/Ecma-262-arch.htm</a>.</p>
```
ljharb: this should be https

gibson042: Whoops, agreed.

gibson042:
I believe it is important to provide links directly from a document towards those it updates and/or replaces (cf. the "Obsoletes" and "Updates" front matter in RFCs such as [7230](https://tools.ietf.org/html/rfc7230)). This particular reference _could_ be placed elsewhere in the spec, but to me it seems very appropriate to follow up language like "This… is the tenth edition" with content that describes where to find previous editions.

littledan:
Let's look into surfacing historical information as part of @jorydotcom 's archiving project. If you want to understand the history of something, there's a lot more to look at than the past annual versions.

gibson042:
I agree, but in my opinion that doesn't invalidate the benefits (I would say "requirements") of providing a link that points from a document towards the content it updates and/or replaces.

allenwb:
I agree that this is perhaps giving too much attention to historical superseded editions. Also, remember that the current official standard is not the same as the current working draft on github. Perhaps a more general statement would be better:

> The current and past editions of ECMA-262 is available in several formats at https://www.ecma-international.org/publications/standards/Ecma-262.htm. The  current working draft for the next edition is available at https://tc39.github.io/ecma262/ .

This is a more general statement with more information. It says where to look for historic editions but doesn't over emphasize them.

ljharb:
I like that suggestion a lot - especially if we can flip the order, so the current working draft link is mentioned first.

gibson042:
I like that too. Done.

littledan:
I'm not sure, I'm still skeptical of directing people to the historical editions as the third sentence of the specification. Understanding the history also includes understanding the revision history on GitHub, and the revisions of the docs that Allen made; many intermediate versions have shipped to the web. Many people see the historical spec versions already and draw false conclusions for them. I'd prefer for this to be presented with more clear context, as part of the archival project.

gibson042:
@littledan are you arguing for the https://www.ecma-international.org/publications/standards/Ecma-262.htm link to appear elsewhere in the spec, or for it not to appear in the spec at all?

If the former, then where would you propose if not alongside the already-existing discussion of other editions? I want it to be early but nonintrusive, both to place each revision in its proper context and also to make it easy for someone who has pulled up the _wrong_ edition (e.g., from having bookmarked a now-outdated edition) to quickly correct their mistake.

But if you instead are arguing for the link to not appear in the spec at all, then we just fundamentally disagree and I don't know how to move forward.

zenparsing:
@littledan Would you prefer a link to the archival project (when it is ready), possibly at the end of this "history" section?

gibson042:
The introduction is currently structured as a self-definition of the document, followed by a history of the language and a final acknowledgement of contributors:
* _This standard defines ECMAScript, and is the Nth edition of the ECMAScript Language Specification. Since the first edition in 1997, ECMAScript has taken over the world. It is best known from web browsers, but has also been adopted for server and embedded applications._
* _ECMAScript is based on JavaScript and JScript, appearing in all Netscape browsers since Navigator 2.0 and all Microsoft browsers since IE 3.0._
* _Development of the Language Specification started in November 1996. The first edition of this Ecma Standard was adopted by the Ecma General Assembly of June 1997._
* _(approval of the second edition)_
* …
* _(changes in previous edition)_
* _(changes in this edition)_
* _(acknowledgement of community effort)_
* _(identification of editors)_

@littledan I currently have the link to all editions and formats right after the self-definition and before the history lesson, for easy outbound navigation when someone finds theirself in the wrong place. It obviously doesn't belong _in_ the history, so where would you like it moved? Before the community acknowledgements? After the list of editors? Under a new Introduction subheading?

littledan:
Maybe right after the history? I think the historical editions are logically part of the history and make sense to explain as such.

gibson042:
Right, but the history provides edition-by-edition detail in chronological order. So "after the history" in practice will mean right after a description of changes in the latest edition, very far removed from the document's self-description and immediately before community acknowledgments. Do you really think that's better than immediately _before_ the detailed history, where it is now?

domenic:
Although I don't want to make committee unanimity required for editorial changes, I would urge the editor not to accept this as-is, and if such an archival link is provided at all, for it to be much further down in the document, as discussed upthread.

ljharb:
(don’t mind the netlify rule failures; I’m trying to get it configured for #1525)

gibson042:
Updated to resolve merge conflicts, and I'm happy to see the CI success.

spec.html
```diff
@@ -195,8 +195,8 @@ <h1>Contributing to this Specification</h1>
...
-  <p>ECMAScript is based on several originating technologies, the most well-known being JavaScript (Netscape) and JScript (Microsoft). The language was invented by Brendan Eich at Netscape and first appeared in that company's Navigator 2.0 browser. It has appeared in all subsequent browsers from Netscape and in all browsers from Microsoft starting with Internet Explorer 3.0.</p>
+  <p>This Ecma Standard defines the ECMAScript 2026 Language. It is the seventeenth edition of the ECMAScript Language Specification. The current working draft for the next edition to be published is available at <a href="https://tc39.es/ecma262/">https://tc39.es/ecma262/</a>. The latest and past editions are available in multiple formats at <a href="https://ecma-international.org/publications-and-standards/standards/ecma-262/">https://ecma-international.org/publications-and-standards/standards/ecma-262/</a>.</p>
```
michaelficarra: I think I'd rather include this in `#metadata-block`. Also the draft link is redundant with what is already there.

spec.html
```diff
@@ -195,8 +195,8 @@ <h1>Contributing to this Specification</h1>
...
+  <p>ECMAScript is based on several originating technologies, the most well-known being JavaScript (Netscape) and JScript (Microsoft). The language was invented by Brendan Eich at Netscape and first appeared in that company's Navigator 2.0 browser. It has appeared in all subsequent browsers from Netscape and in all browsers from Microsoft starting with Internet Explorer 3.0. Though best known as the language embedded in web browsers, it has also been widely adopted for server and embedded applications. Since publication of the first edition in 1997, ECMAScript has grown to be one of the world's most widely used general-purpose programming languages.</p>
```
michaelficarra: I like where you're going with this, but I'd prefer a separate PR so we can discuss it separately.

gibson042: #3735

spec.html
gibson042: I wasn't sure that #metadata-block survived snapshot publication, but looking at https://tc39.es/ecma262/2024/multipage/ I think we're good. Updated.

spec.html
```diff
@@ -164,7 +164,7 @@
...
+  <p>The document at <a href="https://tc39.es/ecma262/">https://tc39.es/ecma262/</a> is the most accurate and up-to-date ECMAScript specification. It contains the content of the most recent yearly snapshot plus any <a href="https://github.com/tc39/proposals/blob/HEAD/finished-proposals.md">finished proposals</a> (those that have reached Stage&nbsp;4 in the <a href="https://tc39.es/process-document/">proposal process</a> and thus are implemented in several implementations and will be in the next practical revision) since that snapshot was taken. The latest and past editions are available in multiple formats at <a href="https://ecma-international.org/publications-and-standards/standards/ecma-262/">https://ecma-international.org/publications-and-standards/standards/ecma-262/</a>.</p>
```
```suggestion
  <p>The document at <a href="https://tc39.es/ecma262/">https://tc39.es/ecma262/</a> is the most accurate and up-to-date ECMAScript specification. It contains the content of the most recent yearly snapshot plus any <a href="https://github.com/tc39/proposals/blob/HEAD/finished-proposals.md">finished proposals</a> — those that have reached Stage&nbsp;4 in the <a href="https://tc39.es/process-document/">proposal process</a> and thus are implemented in several implementations and will be included in the next yearly snapshot. Historical snapshots are available at <a href="https://ecma-international.org/publications-and-standards/standards/ecma-262/">https://ecma-international.org/publications-and-standards/standards/ecma-262/</a>.</p>
```

I think this phrasing better emphasises that this is the preferred specification document and any other document is simply a possibly-outdated snapshot of what was once here.

michaelficarra (approved):

syg (approved):

PR #1342, opened 2018-11-05
Editorial: Trivial fix for header of Well-Known symbols table

rhysd:
I found a table header has unnecessary `[[` and `]]` at https://tc39.github.io/ecma262/#sec-ecmascript-language-types-string-type . Removed them.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -930,7 +930,7 @@
                 Specification Name
               </th>
               <th>
-                [[Description]]
+                Description
               </th>
               <th>
                 Value and Purpose


jmdyck:
The double-square-brackets are intentional. The content of that column is not just a general description of the symbol, but is rather the symbol's associated `[[Description]]` value.

ljharb:
Agreed; this is intentional.

rhysd:
Oh, I see. Thank you for correcting me.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1345, opened 2018-11-07
Update abstract line of JSON.stringify section

### LINKED ISSUE #1344: Incorrect JSON.stringify description
https://tc39.github.io/ecma262/#sec-json.stringify

    The stringify function returns a String in UTF-16 encoded JSON format representing an ECMAScript value. 

This needs to read as: "The stringify function *may* return a String in UTF-16 encoded JSON format representing an ECMAScript value. "

If you call it as such:

JSON.stringify(undefined)

The resultant value is *NOT* of type string, and therefore cannot be a string representation of a JSON object.

The simple test case is that: JSON.parse(JSON.stringify(undefined)) should throw an exception on unencodable data eagerly (at the stringify step), not at the later parse step - otherwise, every call to JSON.stringify has to type check the resultant value when the input is non deterministic at run time.

ahungry:
- Previously did not account for any obvious mention of non-string return value.

Fixes #1344.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -35867,7 +35867,7 @@ THH:mm:ss.sss
 
     <emu-clause id="sec-json.stringify">
       <h1>JSON.stringify ( _value_ [ , _replacer_ [ , _space_ ] ] )</h1>
-      <p>The `stringify` function returns a String in UTF-16 encoded JSON format representing an ECMAScript value. It can take three parameters. The _value_ parameter is an ECMAScript value, which is usually an object or array, although it can also be a String, Boolean, Number or *null*. The optional _replacer_ parameter is either a function that alters the way objects and arrays are stringified, or an array of Strings and Numbers that acts as an inclusion list for selecting the object properties that will be stringified. The optional _space_ parameter is a String or Number that allows the result to have white space injected into it to improve human readability.</p>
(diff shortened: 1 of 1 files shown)


ahungry:
Related to: https://github.com/tc39/ecma262/issues/1344

spec.html
```diff
@@ -35867,7 +35867,7 @@ <h1>Runtime Semantics: InternalizeJSONProperty ( _holder_, _name_ )</h1>
...
+      <p>The `stringify` function returns a String in UTF-16 encoded JSON format representing an ECMAScript value, or undefined. It can take three parameters. The _value_ parameter is an ECMAScript value, which is usually an object or array, although it can also be a String, Boolean, Number or *null*. The optional _replacer_ parameter is either a function that alters the way objects and arrays are stringified, or an array of Strings and Numbers that acts as an inclusion list for selecting the object properties that will be stringified. The optional _space_ parameter is a String or Number that allows the result to have white space injected into it to improve human readability.</p>
```
```suggestion
      <p>The `stringify` function returns a String in UTF-16 encoded JSON format representing an ECMAScript value, or *undefined*. It can take three parameters. The _value_ parameter is an ECMAScript value, which is usually an object or array, although it can also be a String, Boolean, Number or *null*. The optional _replacer_ parameter is either a function that alters the way objects and arrays are stringified, or an array of Strings and Numbers that acts as an inclusion list for selecting the object properties that will be stringified. The optional _space_ parameter is a String or Number that allows the result to have white space injected into it to improve human readability.</p>
```

ljharb: Separately, I’m a bit confused why object and array are lowercase, but String, Number, and Boolean are capitalized. @tc39/ecma262-editors?

bterlson: Lowercase o object is used to refer to anything that is an Object (https://tc39.github.io/ecma262/#sec-terms-and-definitions-object). Just an editorial convention, but one I like as Object is used frequently. Array should probably be capitalized however.

PR #1346, opened 2018-11-14
add Symbol.prototype.description getter

michaelficarra:
This integrates the stage 3 [symbol description proposal](https://github.com/tc39/proposal-Symbol-description) in preparation for stage 4.

As of this writing, there are [three compatible browser implementations](https://test262.report/browse/built-ins/Symbol/prototype/description).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -25027,6 +25027,16 @@
         </emu-clause>
       </emu-clause>
 
+      <emu-clause id="sec-symbol.prototype.description">
+        <h1>get Symbol.prototype.description</h1>
+        <p>`Symbol.prototype.description` is an accessor property whose set accessor function is *undefined*. Its get accessor function performs the following steps:</p>
+        <emu-alg>
+          1. Let _s_ be the *this* value.
+          1. Let _sym_ be ? thisSymbolValue(_s_).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -25027,6 +25027,16 @@ <h1>Runtime Semantics: SymbolDescriptiveString ( _sym_ )</h1>
         </emu-clause>
       </emu-clause>
 
+      <emu-clause id="sec-symbol.prototype.description">
+        <h1>get Symbol.prototype.description</h1>
+        <p>`Symbol.prototype.description` is an accessor property whose set accessor function is *undefined*. Its get accessor function performs the following steps:</p>
+        <emu-alg>
+          1. Let _s_ be the *this* value.
+          1. Let _sym_ be ? thisSymbolValue(_s_).
```
zenparsing: You could also do

```
Let _sym_ be ? thisSymbolValue(*this* value).
```

ljharb: The reigning convention seems to be to have it in a separate variable; i don’t have a strong opinion either way as long as we’re consistent.

michaelficarra:
@tc39/ecma262-editors I've removed the `pending stage 4` label.

PR #1363, opened 2018-11-29
fixes #1133: replace some uses of "idempotent" with alternative wording

### LINKED ISSUE #1133: change misuse of "idempotent" to something else
There are occasions throughout the spec where the word "idempotent" is used to mean something other than [its common usage in mathematics](https://en.wikipedia.org/wiki/Idempotence). In these cases, the spec is trying to claim that a host-defined procedure indeed behaves as a mathematical function. For example,

> This operation must be idempotent if it completes normally. Each time it is called with a specific exportName, resolveSet pair as arguments it must return the same result.

Confusingly, there are also instances that are consistent with common usage.

> The ToInt32 abstract operation is idempotent: if applied to a result that it produced, the second application leaves that value unchanged.

I think we can just remove the incorrect usage, since the intended meaning is always described immediately after. If someone has a better term or phrasing to use as an alternative, I'd love to hear it.

michaelficarra:
Fixes #1133. The other uses of "idempotent" (to mean "if applied to any of its own outputs, it behaves as the identity function") were left intact.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9634,7 +9634,7 @@
           1. Return ? _targetEnvRec_.GetBindingValue(_binding_.[[BindingName]], *true*).
         </emu-alg>
         <emu-note>
-          <p>ResolveExport is idempotent and side-effect free. An implementation might choose to pre-compute or cache the ResolveExport results for the [[Exports]] of each module namespace exotic object.</p>
+          <p>ResolveExport is side-effect free. Each time this operation is called with a specific _exportName_, _resolveSet_ pair as arguments it must return the same result. An implementation might choose to pre-compute or cache the ResolveExport results for the [[Exports]] of each module namespace exotic object.</p>
         </emu-note>
       </emu-clause>
 
@@ -22533,7 +22533,7 @@
               </td>
               <td>
                 <p>Return the binding of a name exported by this module. Bindings are represented by a <dfn id="resolvedbinding-record">ResolvedBinding Record</dfn>, of the form { [[Module]]: Module Record, [[BindingName]]: String }. If the export is a Module Namespace Object without a direct binding in any module, [[BindingName]] will be set to *"\*namespace\*"*. Return *null* if the name cannot be resolved, or *"ambiguous"* if multiple bindings were found.</p>
-                <p>This operation must be idempotent if it completes normally. Each time it is called with a specific _exportName_, _resolveSet_ pair as arguments it must return the same result.</p>
+                <p>Each time this operation is called with a specific _exportName_, _resolveSet_ pair as arguments it must return the same result if it completes normally.</p>
               </td>
             </tr>
             <tr>
@@ -23572,7 +23572,7 @@
             If a Module Record corresponding to the pair _referencingScriptOrModule_, _specifier_ does not exist or cannot be created, an exception must be thrown.
           </li>
           <li>
-            This operation must be idempotent if it completes normally. Each time it is called with a specific _referencingScriptOrModule_, _specifier_ pair as arguments it must return the same Module Record instance.
+            Each time this operation is called with a specific _referencingScriptOrModule_, _specifier_ pair as arguments it must return the same Module Record instance if it completes normally.
           </li>
         </ul>
         <p>Multiple different _referencingScriptOrModule_, _specifier_ pairs may map to the same Module Record instance. The actual mapping semantic is implementation-defined but typically a normalization process is applied to _specifier_ as part of the mapping process. A typical normalization process would include actions such as alphabetic case folding and expansion of relative and abbreviated path specifiers.</p>


littledan:
I'm not sure about this change; I prefer the "idempotent" wording, for the reasons described in https://github.com/tc39/ecma262/pull/1482#discussion_r269241101 . However, I don't feel very strongly, and I'm fine to adopt the "deterministic" wording if everyone else likes that better. I'll land that change in #1482 if this one lands.

allenwb:
I'm with @littledan but stronger.  "Determinist" is not an appropriate substitute for for "idempotent" in these cases.

claudepache:
“idempotent with respect to the program state”?

ljharb:
@michaelficarra are you comfortable with the above [suggestion](https://github.com/tc39/ecma262/pull/1363#issuecomment-476828792)?

michaelficarra:
@ljharb No, not at all. I still fail to see how "deterministic with respect to its parameters" is insufficient to describe these scenarios. If the parameters are the same across many calls, the function's behaviour will be the same. We shouldn't continue to use "idempotent" in two very different ways.

bathos:
Deterministic, even with the ‘with respect to’ qualification, I would usually expect to mean only that it introduces no entropy to the application state, which doesn’t necessarily imply the function returns the same value, only that the value is predictable. Maybe if the qualifier were ‘with respect _only_ to’?

TimothyGu:
How about just remove this sentence? The rest of the paragraphs explain what are meant in these contexts anyway, though we should still keep "if it completes normally" around.

michaelficarra:
@bathos IME, "with respect to" used in this way always means "with respect only to". But I wouldn't refuse if that makes it clearer for some.

@TimothyGu That is an okay compromise if the editors don't like the proposed wording.

bathos:
@michaelficarra You’re right — I suppose I figured the ‘only’ might shift emphasis in a way that made the intended meaning clearer, but maybe that’s not true.

ljharb:
@littledan, @allenwb - you two seem to still be against this change; do you have additional thoughts, or has your opinion changed?

littledan:
Has something changed in this PR? My feeling is still what I said in https://github.com/tc39/ecma262/pull/1363#issuecomment-476808132 .

bakkot:
FWIW I am strongly in favor of this change for the reasons given in the OP of #1133. However, if people are getting tripped up by the use of "deterministic with respect to its parameters" (for reasons which I do not entirely understand), we just not pick a word for this, and instead say something like

> ResolveExport is side-effect free. Each time it is called with a specific _exportName_, _resolveSet_ pair as arguments it must return the same result. An implementation might choose to pre-compute or cache the ResolveExport results for the [[Exports]] of each module namespace exotic object.</p>

> Each time this operation is called with a specific _exportName_, _resolveSet_ pair as arguments it must return the same result if it completes normally.

> Each time this operation is called with a specific _referencingModule_, _specifier_ pair as arguments it must return the same Module Record instance if it completes normally.

zenparsing:
@bakkot I like it.

ljharb:
That seems equally correct but less concise - but if it’s more widely acceptable than it seems like a good improvement.

ljharb:
@littledan what do you think about [this suggestion](https://github.com/tc39/ecma262/pull/1363#issuecomment-497051942)?

littledan:
@bakkot That wording seems fine to me.

michaelficarra:
@tc39/ecma262-editors Rebased and updated to use the above suggested wording.

spec.html
```diff
@@ -22644,7 +22644,7 @@ <h1>Runtime Semantics: HostResolveImportedModule ( _referencingScriptOrModule_,
             If a Module Record corresponding to the pair _referencingScriptOrModule_, _specifier_ does not exist or cannot be created, an exception must be thrown.
           </li>
           <li>
-            This operation must be idempotent if it completes normally. Each time it is called with a specific _referencingScriptOrModule_, _specifier_ pair as arguments it must return the same Module Record instance.
+            Each time this operation is called with a specific _referencingScriptOrModule_, _specifier_ pair as arguments it must return the same Module Record instance if it completes normally.
```
michaelficarra: @tc39/ecma262-editors Please confirm that this changed wording matches the intended normative requirement that the mathematical-function-like behaviour only applies for normal completions. It was unclear whether the qualifier in the previous sentence was meant to apply to this whole `<li>` or whether the two sentences were meant to be read completely independently.

ljharb: (relates somewhat to #1645)

the rewording seems correct to me; cc @domenic to confirm.

domenic: That was the intent.

michaelficarra:
Rebased.

PR #1368, opened 2018-12-11
Editorial: Add "async generator" to [[FunctionKind]] description

### LINKED ISSUE #1308: Fixing description of [[FunctionKind]] internal slot of Function Object.
In `FunctionAllocate` there is few logic.
> 2. Assert: functionKind is either "normal", "non-constructor", "generator", "async", or "async generator"
> 11. Set F.[[FunctionKind]] to functionKind.

So, we can say, F.[[FunctionKind]] can be "async generator".
but, description about [[FunctionKind]] in specification, It isn't
> Either "normal", "classConstructor", "generator" or "async"

I think, to change this to blow will be better. 
> Either "normal", "classConstructor", "generator", "async" or "async generator"

PS. why FunctionKind of async generator is "async generator" ? even class constructor is "classConstructor". I think it have to be "asyncGenerator", because specification usually use camel case.

viktmv:
Fixes #1308

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7170,7 +7170,7 @@
             String
           </td>
           <td>
-            Either `"normal"`, `"classConstructor"`, `"generator"`, or `"async"`.
+            Either `"normal"`, `"classConstructor"`, `"generator"`, `"async"`, or `"async generator"`.
           </td>
         </tr>
         <tr>


bterlson:
@viktmv thank you and congrats on your first contribution to ECMA262! 🎉🎉

Issue #486, opened 2016-03-20
!? usage


annevk:
#424 was closed though the remaining calls of OrdinaryGetOwnPropery were not updated to include !. So that probably needs to be fixed.

While closing @bterlson also mentioned that all abstract operations return a completion value and therefore must use ? or !.

That does not seem consistent with the draft, e.g., IsCallable is never prefixed.

There should probably be some more reliable set of rules.

allenwb:
@annevk @bterlson 

and also probably worth noting that abstraction operations used for static semantics specification can't produce abrupt completions because they happen outside of the ES runtime execution environment.

bterlson:
? or ! are not mandatory. If not used the assumption is that you're operating on a completion record (static semantics is the one exception I think?). There is a rule in 5.2 that allows us to treat a completion record as if it were a value but it's something I'd like to get away from. It's in that spirit that I closed #424.

However, I understand wanting to avoid ! when calling abstract operations that cannot throw. We should probably fix #253 and notate certain operations as non-throwing and specify that these return values not completion records. That would necessitate not using ! and ? on such operations. I'll add this to emu.

jmdyck:
> We should probably fix #253 and notate certain operations as non-throwing and specify that these return values not completion records.

(I have a related proposal, but it's waiting for #452 to be merged.)

annevk:
@jmdyck so given the proposal in #497 we would remove ! for OrdinaryGetOwnProperty calls since that never returns an abrupt completion?

domenic:
> However, I understand wanting to avoid ! when calling abstract operations that cannot throw. We should probably fix #253 and notate certain operations as non-throwing and specify that these return values not completion records. That would necessitate not using ! and ? on such operations. I'll add this to emu.

Hmm, I do not like this approach. The reason I wanted #253 in the first place was so that I know, when writing other "code" that uses that abstract operation, whether I should use ! or ? at the call site, and consequently whether the call site itself can throw or not throw. Eliminating the ! from the call site reduces clarity, making me jump to the definition of the abstract operation to find out if we're assuming it returns a non-completion record, or a completion record.

I'd much rather have AbstractOp(x) always return a completion record, that must always be unwrapped via ! or ?. The extra character is a very minimal burden and adds a lot of good explictness.

jmdyck:
@annevk:
> so given the proposal in #497 we would remove ! for OrdinaryGetOwnProperty calls since that never returns an abrupt completion?

[...]

PR #1372, opened 2018-12-17
Add "name" property for classes as part of ClassDefinitionEvaluation

anba:
SetFunctionName is currently called after ClassDefinitionEvaluation, which means the class name is not available in static class field initializers (https://github.com/tc39/proposal-class-fields/issues/85) and may cause issues for a `@frozen` decorator (https://github.com/tc39/proposal-decorators/issues/168). 

This PR moves the initialization of the "name" property into ClassDefinitionEvaluation, which should avoid both aforementioned issues. I've split this PR into three parts for easier reviewing:

- The first commit 7bb51dd87436dcf655e8cbf1ff2285d3688b28e7 adds calls to a new runtime semantic `NamedEvaluation` to evaluate anonymous function definitions (as identified by `IsAnonymousFunctionDefinition`). 

- 3d752d728202f4c8511df4d9720972e85ddd976f adds the actual definitions for `NamedEvaluation`.

- And the last commit 03fa5ba4efda9788c07e589b019702928d2fb562 moves the "name" property assignment into ClassDefinitionEvaluation.

This PR doesn't result in any observable changes for non-class anonymous function definitions. For class definitions (with explicit _and_ inferred names), the "name" property is now defined earlier, which is observable when calling `[[OwnPropertyKeys]]`. I don't expect to see any web-compatibility issues for this change given that all major engines already return the property keys of classes in a different order:
```js
function printNames(kind, c) {
    c.foo = 0;
    print(`${kind}: ${Object.getOwnPropertyNames(c)}`);
}

// SM: prototype,m,foo,length,name
// JSC: length,name,prototype,m,foo
// V8: length,prototype,m,name,foo
// CH: prototype,length,name,m,foo
//
// Spec: length,prototype,m,name,foo
// Proposed: length,prototype,name,m,foo
printNames("class", class C { static m(){} });

// For comparison a normal function (Strict mode enabled to avoid non-standard arguments and caller.)
// SM: foo,prototype,length,name
// JSC: length,name,foo,prototype
// V8: length,name,prototype,foo
// CH: prototype,length,name,foo
//
// Spec: length,prototype,name,foo
printNames("function", function f(){ "use strict" });
```

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -11774,11 +11774,11 @@
         <emu-alg>
           1. Let _propKey_ be the result of evaluating |PropertyName|.
           1. ReturnIfAbrupt(_propKey_).
-          1. Let _exprValueRef_ be the result of evaluating |AssignmentExpression|.
-          1. Let _propValue_ be ? GetValue(_exprValueRef_).
           1. If IsAnonymousFunctionDefinition(|AssignmentExpression|) is *true*, then
-            1. Let _hasNameProperty_ be ? HasOwnProperty(_propValue_, `"name"`).
-            1. If _hasNameProperty_ is *false*, perform SetFunctionName(_propValue_, _propKey_).
+            1. Let _propValue_ be the result of performing NamedEvaluation for |AssignmentExpression| with argument _propKey_.
+          1. Else,
+            1. Let _exprValueRef_ be the result of evaluating |AssignmentExpression|.
+            1. Let _propValue_ be ? GetValue(_exprValueRef_).
           1. Assert: _enumerable_ is *true*.
           1. Assert: _object_ is an ordinary, extensible object with no non-configurable properties.
           1. Return ! CreateDataPropertyOrThrow(_object_, _propKey_, _propValue_).
@@ -12126,6 +12126,21 @@
         <emu-grammar>ParenthesizedExpression : `(` Expression `)`</emu-grammar>
         <emu-alg>
           1. Return AssignmentTargetType of |Expression|.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-grouping-operator-runtime-semantics-namedevaluation">
+        <h1>Runtime Semantics: NamedEvaluation</h1>
+        <p>With parameter _name_.</p>
+        <emu-grammar>PrimaryExpression : CoverParenthesizedExpressionAndArrowParameterList</emu-grammar>
+        <emu-alg>
+          1. Let _expr_ be CoveredParenthesizedExpression of |CoverParenthesizedExpressionAndArrowParameterList|.
+          1. Return the result of performing NamedEvaluation for _expr_ with argument _name_.
+        </emu-alg>
+        <emu-grammar>ParenthesizedExpression : `(` Expression `)`</emu-grammar>
+        <emu-alg>
+          1. Assert: IsAnonymousFunctionDefinition(|Expression|) is *true*.
+          1. Return the result of performing NamedEvaluation for |Expression| with argument _name_.
         </emu-alg>
       </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -18677,6 +18703,17 @@ <h1>Runtime Semantics: EvaluateBody</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-arrow-function-definitions-runtime-semantics-namedevaluation">
+      <h1>Runtime Semantics: NamedEvaluation</h1>
+      <p>With parameter _name_.</p>
+      <emu-grammar>ArrowFunction : ArrowParameters `=&gt;` ConciseBody</emu-grammar>
+      <emu-alg>
+        1. Let _closure_ be the result of evaluating this |ArrowFunction|.
+        1. Perform SetFunctionName(_closure_, _name_).
```
```suggestion
        1. Perform ! SetFunctionName(_closure_, _name_).
```

spec.html
```diff
@@ -19118,6 +19155,17 @@ <h1>Runtime Semantics: PropertyDefinitionEvaluation</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-generator-function-definitions-runtime-semantics-namedevaluation">
+      <h1>Runtime Semantics: NamedEvaluation</h1>
+      <p>With parameter _name_.</p>
+      <emu-grammar>GeneratorExpression : `function` `*` `(` FormalParameters `)` `{` GeneratorBody `}`</emu-grammar>
+      <emu-alg>
+        1. Let _closure_ be the result of evaluating this |GeneratorExpression|.
+        1. Perform SetFunctionName(_closure_, _name_).
```
```suggestion
        1. Perform ! SetFunctionName(_closure_, _name_).
```

spec.html
```diff
@@ -19438,6 +19486,19 @@ <h1>Runtime Semantics: PropertyDefinitionEvaluation</h1>
...
 
+    <emu-clause id="sec-asyncgenerator-definitions-namedevaluation">
+      <h1>Runtime Semantics: NamedEvaluation</h1>
+      <p>With parameter _name_.</p>
+      <emu-grammar>
+        AsyncGeneratorExpression : `async` [no LineTerminator here] `function` `*` `(` FormalParameters `)` `{` AsyncGeneratorBody `}`
+      </emu-grammar>
+      <emu-alg>
+        1. Let _closure_ be the result of evaluating this |AsyncGeneratorExpression|.
+        1. Perform SetFunctionName(_closure_, _name_).
```
```suggestion
        1. Perform ! SetFunctionName(_closure_, _name_).
```

spec.html
```diff
@@ -19767,6 +19828,8 @@ <h1>Runtime Semantics: ClassDefinitionEvaluation</h1>
         1. If |ClassHeritage_opt| is present, set _F_.[[ConstructorKind]] to `"derived"`.
         1. Perform MakeConstructor(_F_, *false*, _proto_).
         1. Perform MakeClassConstructor(_F_).
+        1. If _className_ is not *undefined*, then
+          1. Perform SetFunctionName(_F_, _className_).
```
```suggestion
          1. Perform ! SetFunctionName(_F_, _className_).
```

spec.html
```diff
@@ -20072,6 +20133,20 @@ <h1>Runtime Semantics: PropertyDefinitionEvaluation</h1>
...
+
+    <emu-clause id="sec-async-function-definitions-runtime-semantics-namedevaluation">
+      <h1>Runtime Semantics: NamedEvaluation</h1>
+      <p>With parameter _name_.</p>
+      <emu-grammar>
+        AsyncFunctionExpression : `async` [no LineTerminator here] `function` `(` FormalParameters `)` `{` AsyncFunctionBody `}`
+      </emu-grammar>
+      <emu-alg>
+        1. Let _closure_ be the result of evaluating this |AsyncFunctionExpression|.
+        1. Perform SetFunctionName(_closure_, _name_).
```
```suggestion
        1. Perform ! SetFunctionName(_closure_, _name_).
```

spec.html
```diff
@@ -20348,6 +20423,20 @@ <h1>Runtime Semantics: EvaluateBody</h1>
...
+    <emu-clause id="sec-async-arrow-function-definitions-runtime-semantics-namedevaluation">
+      <h1>Runtime Semantics: NamedEvaluation</h1>
+      <p>With parameter _name_.</p>
+      <emu-grammar>
+        AsyncArrowFunction : `async` [no LineTerminator here] AsyncArrowBindingIdentifier [no LineTerminator here] `=>` AsyncConciseBody
+        AsyncArrowFunction : CoverCallExpressionAndAsyncArrowHead [no LineTerminator here] `=>` AsyncConciseBody
+      </emu-grammar>
+      <emu-alg>
+        1. Let _closure_ be the result of evaluating this |AsyncArrowFunction|.
+        1. Perform SetFunctionName(_closure_, _name_).
```
```suggestion
        1. Perform ! SetFunctionName(_closure_, _name_).
```

littledan:
We got consensus on this change in the January 2019 TC39 meeting! Thanks everybody.

@anba Would you be interested in writing tests for this change?

spec.html
anba: I'd like to keep this without !, because there's still no resolution for #486 and it matches the majority of existing SetFunctionName callers.

anba:
Created tc39/test262#2057

spec.html
ljharb: Totally fine for this PR; but I do see #486 as resolved, so I'll be happy to put up a followup PR that converts all SetFunctionName callsites to use `!` or `?` as needed.

Issue #354, opened 2016-02-03
Spec convention for branding


domenic:
As we write more Ecmaspeak specs (Streams, 402, Loader), we find the way that ES does branding a bit onerous and opaque. That is, it picks a privileged internal slot and says "if you have this internal slot, you can pass go."

I propose the following instead:

- In a class definition, when defining the internal slots for a class (e.g. ["Properties of Map instances"](https://tc39.github.io/ecma262/#sec-properties-of-map-instances) in ES), we can say a magic sentence: "[[MapData]] is Map's _branding internal slot_."
- We provide two abstract operations, IsBranded(o, Class) and ThrowOnBrandMismatch(o, Class) that do what they say. Example usage:
  - replace [steps 2 and 3 of Map.prototype.size](https://tc39.github.io/ecma262/#sec-get-map.prototype.size) with "Perform ! ThrowOnBrandMismatch(_M_, Map)".
  - replace [step 8 of %MapIteratorPrototype%.next](https://tc39.github.io/ecma262/#sec-%mapiteratorprototype%.next) with "Assert: IsBranded(_m_, Map)."

Potential variant: use a small badge in the heading of the method/abstract op, like [streams does for throw/nothrow](https://cloud.githubusercontent.com/assets/617481/8471094/a55b5fe6-20cf-11e5-956b-11bf6d9343fc.png), like `brandcheck(Map)`.

Does this sound interesting? If so I could work on a PR, because streams is getting really sick of defining IsStreamWhatever(x) abstract ops and repeating the boilerplate everywhere. Having ES lead the way would be great.

allenwb:
We intentionally avoiding any "branding" language in the ES6 spec. because the precious use of [[Class]] as a branding mechanism had been too broadly assumed to imply things other than what was actually required by the specification. And somebody thought it was something that actually had to exist as a string value or tried to misuse it in various ways. 

Generally, all that we want is to require with the replacement language is that that specified functions that depend upon specific object-level encodings of internal slot data throw an error when applied to objects that do not encode those internal slots.

Another way to describe the [[MapData]] requirement might be to say something like "is an object that was successfully allocated and initialized by invoking the  [[Construct]] internal method of a built-in Map constructor"  (but can't say "the Map constructor" because of multiple Realms).

I'm not particularly attached to the ES6 language, but I think we should avoid any language that implies  anything more than "an object that meets the requirements of this method".

domenic:
[...]

PR #1373, opened 2018-12-17
Editorial: add RequireInternalSlot abstract operation

ljharb:
To me, this seems like a nice way to DRY up a common operation - asserting that a value is an Object with a given internal slot.

What are your thoughts on this? Is this too much indirection, or does this reduction in spec lines make it easier to read?

It was a pretty quick change, mostly a single regex operation, so I'm happy to close it if we don't think this is an improvement.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7131,6 +7131,15 @@
         <p>If _constructor_ does not supply a [[Prototype]] value, the default value that is used is obtained from the realm of the _constructor_ function rather than from the running execution context.</p>
       </emu-note>
     </emu-clause>
+
+    <emu-clause id="sec-requireinternalslot" aoid="RequireInternalSlot">
+      <h1>RequireInternalSlot ( _O_, _internalSlot_ )</h1>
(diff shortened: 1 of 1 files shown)


domenic:
Cf. https://github.com/tc39/ecma262/issues/354 . This seems like a simpler version of what was proposed there: slightly less readable/intuitive IMO, but also a smaller delta from the existing conventions.

ljharb:
@domenic Thanks, good callout! if we can figure out a better convention as in #354, I don't think this change would make that transition harder (although it probably wouldn't make it all that much easier).

zenparsing:
I like this approach; it aids readability while not sacrificing explicitness.

bakkot:
Thinking about internal slots as first-class values which can be passed around (in the spec) makes my heard hurt a bit, but I don't see a fundamental problem it it.

bakkot:
> we already make liberal use of it with ObjectCreate

So we do! I'd forgotten.

PR #1376, opened 2018-12-18
Editorial: define and use improved intrinsics notation

ljharb:
Per editors' group discussion, separating this out from #1314.

The intention here is to prevent unbounded future growth of the intrinsics table, while also making intrinsic references slightly more readable/intuitive by making them more closely mirror the JS you'd use to access them (in an unmodified realm).

This ideally would be coupled with a change in ecmarkup to auto-link intrinsic syntax to the value in question, for even easier consumption.

Closes #1377. Closes #1105.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1704,7 +1704,7 @@
       <emu-clause id="sec-well-known-intrinsic-objects">
         <h1>Well-Known Intrinsic Objects</h1>
         <p>Well-known intrinsics are built-in objects that are explicitly referenced by the algorithms of this specification and which usually have realm-specific identities. Unless otherwise specified each intrinsic object actually corresponds to a set of similar objects, one per realm.</p>
-        <p>Within this specification a reference such as %name% means the intrinsic object, associated with the current realm, corresponding to the name. Determination of the current realm and its intrinsics is described in <emu-xref href="#sec-execution-contexts"></emu-xref>. The well-known intrinsics are listed in <emu-xref href="#table-7"></emu-xref>.</p>
+        <p>Within this specification a reference such as %name% means the intrinsic object, associated with the current realm, corresponding to the name. A reference such as %name.a.b% means, as if the "b" property of the "a" property of the intrinsic object %name% was accessed prior to any ECMAScript code being evaluated. Determination of the current realm and its intrinsics is described in <emu-xref href="#sec-execution-contexts"></emu-xref>. The well-known intrinsics are listed in <emu-xref href="#table-7"></emu-xref>.</p>
         <emu-table id="table-7" caption="Well-Known Intrinsic Objects">
           <table>
             <tbody>
@@ -1749,7 +1749,7 @@
                 `ArrayBuffer.prototype`
               </td>
               <td>
-                The initial value of the `prototype` data property of %ArrayBuffer%.
+                The initial value of the `prototype` data property of %ArrayBuffer%; i.e., %ArrayBuffer.prototype%
               </td>
             </tr>
             <tr>
@@ -1759,7 +1759,7 @@
               <td>
               </td>
               <td>
-                The prototype of Array iterator objects (<emu-xref href="#sec-array-iterator-objects"></emu-xref>)
+                The prototype of Array iterator objects (<emu-xref href="#sec-array-iterator-objects"></emu-xref>); i.e., %ArrayIterator.prototype%
               </td>
             </tr>
             <tr>
@@ -1770,7 +1770,7 @@
                 `Array.prototype`
               </td>
               <td>
-                The initial value of the `prototype` data property of %Array% (<emu-xref href="#sec-properties-of-the-array-prototype-object"></emu-xref>)
+                The initial value of the `prototype` data property of %Array% (<emu-xref href="#sec-properties-of-the-array-prototype-object"></emu-xref>); i.e., %Array.prototype%
               </td>
             </tr>
             <tr>
@@ -1781,7 +1781,7 @@
                 `Array.prototype.entries`
               </td>
               <td>
-                The initial value of the `entries` data property of %ArrayPrototype% (<emu-xref href="#sec-array.prototype.entries"></emu-xref>)
+                The initial value of the `entries` data property of %Array.prototype% (<emu-xref href="#sec-array.prototype.entries"></emu-xref>); i.e., %Array.prototype.entries%
               </td>
             </tr>
             <tr>
@@ -1792,7 +1792,7 @@
                 `Array.prototype.forEach`
               </td>
               <td>
-                The initial value of the `forEach` data property of %ArrayPrototype% (<emu-xref href="#sec-array.prototype.foreach"></emu-xref>)
+                The initial value of the `forEach` data property of %Array.prototype% (<emu-xref href="#sec-array.prototype.foreach"></emu-xref>); i.e., %Array.prototype.forEach%
               </td>
             </tr>
             <tr>
@@ -1803,7 +1803,7 @@
                 `Array.prototype.keys`
               </td>
               <td>
-                The initial value of the `keys` data property of %ArrayPrototype% (<emu-xref href="#sec-array.prototype.keys"></emu-xref>)
+                The initial value of the `keys` data property of %Array.prototype% (<emu-xref href="#sec-array.prototype.keys"></emu-xref>); i.e., %Array.prototype.keys%
               </td>
             </tr>
             <tr>
@@ -1814,7 +1814,7 @@
                 `Array.prototype.values`
               </td>
               <td>
-                The initial value of the `values` data property of %ArrayPrototype% (<emu-xref href="#sec-array.prototype.values"></emu-xref>)
+                The initial value of the `values` data property of %Array.prototype% (<emu-xref href="#sec-array.prototype.values"></emu-xref>); i.e., %Array.prototype.values%
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -1733,7 +1733,7 @@ <h1>Well-Known Intrinsic Objects</h1>
                 `Array.prototype`
               </td>
               <td>
-                The initial value of the `prototype` data property of %Array% (<emu-xref href="#sec-properties-of-the-array-prototype-object"></emu-xref>)
+                The initial value of the `prototype` data property of %Array% (<emu-xref href="#sec-properties-of-the-array-prototype-object"></emu-xref>); e.g., %Array.prototype%
```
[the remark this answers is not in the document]
ljharb: Did some googling and found https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/ which fully agrees with you :-) I'll update these.

spec.html
```diff
@@ -1667,7 +1667,7 @@ <h2>[[Construct]] ( )</h2>
...
         <p>Well-known intrinsics are built-in objects that are explicitly referenced by the algorithms of this specification and which usually have realm-specific identities. Unless otherwise specified each intrinsic object actually corresponds to a set of similar objects, one per realm.</p>
-        <p>Within this specification a reference such as %name% means the intrinsic object, associated with the current realm, corresponding to the name. Determination of the current realm and its intrinsics is described in <emu-xref href="#sec-execution-contexts"></emu-xref>. The well-known intrinsics are listed in <emu-xref href="#table-7"></emu-xref>.</p>
+        <p>Within this specification a reference such as %name% means the intrinsic object, associated with the current realm, corresponding to the name. A reference such as %name.a.b% means, as if the "b" property of the "a" property of thhe intrinsic object %name% was accessed prior to any ECMAScript code being evaluated. Determination of the current realm and its intrinsics is described in <emu-xref href="#sec-execution-contexts"></emu-xref>. The well-known intrinsics are listed in <emu-xref href="#table-7"></emu-xref>.</p>
```
gibson042: thhe → the

jmdyck:
Rather than adding (for instance) `i.e., %ArrayBuffer.prototype%` in the third column ("ECMAScript Language Association"), I think it would make more sense to add `%ArrayBuffer.prototype%` in the first column ("Intrinsic Name").

ljharb:
@jmdyck there are potentially other specs that reference the original intrinsic name.

domenic:
The only such specs should be HTML, Web IDL, and Streams, which are all living and can be updated. So indeed, if you want to remove all `%XPrototype%` entries entirely, we could probably get away with that, with a bit of extra work.

jmdyck:
> @jmdyck there are potentially other specs that reference the original intrinsic name.

They can continue to do so, as I didn't suggest removing the original names. I specifically said *add* the new name to column 1.

Ms2ger:
What's the status here?

spec.html
```diff
@@ -1689,7 +1689,7 @@ <h2>[[Construct]] ( )</h2>
...
         <p>Well-known intrinsics are built-in objects that are explicitly referenced by the algorithms of this specification and which usually have realm-specific identities. Unless otherwise specified each intrinsic object actually corresponds to a set of similar objects, one per realm.</p>
-        <p>Within this specification a reference such as %name% means the intrinsic object, associated with the current realm, corresponding to the name. Determination of the current realm and its intrinsics is described in <emu-xref href="#sec-execution-contexts"></emu-xref>. The well-known intrinsics are listed in <emu-xref href="#table-7"></emu-xref>.</p>
+        <p>Within this specification a reference such as %name% means the intrinsic object, associated with the current realm, corresponding to the name. A reference such as %name.a.b% means, as if the "b" property of the "a" property of the intrinsic object %name% was accessed prior to any ECMAScript code being evaluated. Determination of the current realm and its intrinsics is described in <emu-xref href="#sec-execution-contexts"></emu-xref>. The well-known intrinsics are listed in <emu-xref href="#table-7"></emu-xref>.</p>
```
```suggestion
        <p>Within this specification a reference such as %name% means the intrinsic object, associated with the current realm, corresponding to the name. A reference such as %name.a.b% means the object that is the "b" property of the object that is the "a" property of the intrinsic object %name%, before any ECMAScript code is evaluated. Determination of the current realm and its intrinsics is described in <emu-xref href="#sec-execution-contexts"></emu-xref>. The well-known intrinsics are listed in <emu-xref href="#table-7"></emu-xref>.</p>
```

jmdyck:
Tangential thought for the future (not a problem for this PR):

@ljharb
> my hope is to [...] come back here and file a final one to "clean up" all the no-longer-necessary explicit intrinsics.

Note that if you remove table rows for things that can be expressed as `%Foo.X%`, that will raise a question about the definition of "well-known intrinsic". There are currently two definitions: "built-in objects that are explicitly referenced by the algorithms of this specification", and "the objects listed in Table 7", and I think they currently agree. But if an algorithm references `%Foo.X%` and there's no row for `%Foo.X%` in Table 7, then the definitions disagree.

You could resolve it either way (i.e., pick either to become the "true" definition), but you'd need to alter the other so that it no longer sounds like a definition.

jmdyck:
CreateIntrinsics creates `_realmRec_.[[Intrinsics]]` with a field for each intrinsic listed in Table 7, but this PR will allow specs to refer to intrinsics that are not listed in Table 7, so it's not clear how such references would be 'resolved' (in the specification model). It seems like `[[Intrinsics]]` would need a field for every intrinsic, whether listed in Table 7 or not, which will require some change of wording.

allenwb:
> I'm mildly concerned that this change encourages an incorrect mental model with regards to the well-known intrinsics: namely, that there exists some kind of static relationship between objects named using this notation.

In fact, this change will be contrary to the primary reason I introduced the _Well-known_ Intrinsics table and naming convention.  The problem it was try to solve was maintaining a list of which intrinsics satisfied dependencies of other parts of the specification or other specification.  It serves as a an inventory and warning that these intrinsics provide more than callable user-land functionality.

Just like actual code, we should not be optimizing for writing to the detriment of reading.

ljharb:
That's turned out not to be sustainable, however - as the list continues to grow rapidly. To me, this change improves readability of the spec by removing a lot of otherwise unhelpful noise.

Since external specs depend on abstract operations, too, without an explicit list, the goal of "show what things external specs depend on" isn't met - someone making a change already has to review external specs to know that. For the current specification, a grep should indicate it with either approach.

allenwb:
> That's turned out not to be sustainable, however - as the list continues to grow rapidly.

So, I check.  The ES6 spec. had (+/- counting errors) listed 90 "well known intrinsics"  while the current working draft lists lists 110. So that's an addition of 4 a year averaged over 5 years.  Not sure I'd call a 5% yearly growth rate rapid.  Most of new table entries (12) are new built-in classes/prototypes.  More interesting is the number of "well known intrinsic methods" which went from 2 in 2015 to 10 in the current draft.  

The two well know methods from 2015 are both internally referenced within ECMA-262.  None of the 8 added well known methods are internally referenced.  From that I infer they were added to the list so they could be referenced by external specs.  I find the fact that such external dependencies must be inferred and are not explicitly marked in the table somewhat disturbing.  I would prefer a separate external dependencies table.

"rapid growth" of such external intrinsic method dependencies seems contrary to one of the goals for ES Harmony. That was that built-ins and "host methods" should be self-hostable using ECMAScript code.
When ever an external spec. takes a dependency upon an "intrinsic method" it is specifying behavior that cannot be implemented exclusively use standard ES code. Or flipping it around if widely used frameworks have to depend upon current value access through the global object why isn't that also good enough for the external specs.  If access to intrinsic methods is essential then perhaps time should be spent on formalizing ES code accessible APIs to access them, rather than giving privileged access to "external specs".

It is even worse if you think it is ok for external specs. to depend upon abstract operations  without permission and documentation. Besides precluding "self-hosting" it is the the path to a fragile unmaintainable specification.  This is a basic encapsulation issue.  If external actors have full and unrestricted access and permission to use internal specification devices such as abstract operations then TC39 will loose the ability to maintain the specification.

It's important to make sure we support related external specifications.  But we should not be doing it in a manner prevents self-hosting or that makes it easy (and implies permission) for external specification to take undocumented dependencies upon the non-normative internal design details of our specifications.

ljharb:
I would also prefer an external dependencies table (for both intrinsics and abstract operations), and am happy to pursue that separately. I very much do *not* think it's OK for external specs to have undocumented (in 262) dependencies on 262, but we don't have a solution for that (and the intrinsics table isn't that solution).

If such a table existed, what would be the point of the current intrinsics notation/table? My premise is that these two things are orthogonal, and that "undocumented external dependencies" is a distinct - and important - problem to solve.

ljharb:
I’m planning to rebase and merge this within the next few days.

Either before or after it’s merged, a PR that attempts to somehow document known external dependencies on any part of this spec (even if it’s not visible in the rendered spec, just in the authoring markup) would be *very* welcome. If we could further find a way to normatively prohibit layered specs from depending on parts of the spec without documenting that dependency, that would also be worth investigating (separately).

jmdyck:
> a PR that attempts to somehow document known external dependencies on any part of this spec (even if it’s not visible in the rendered spec, just in the authoring markup) would be _very_ welcome.

E.g., adding atributes like `referenced-by="HTML Intl"` to the appropriate `<emu-clause>` and `<dfn>` elements?

ljharb:
@jmdyck sure, something like that would work - anything's worth exploring. The main goal would be to ensure that when changes are made to something, the impact on layered specs is considered - and also to make such dependencies greppable.

mostafapes:
happy

PR #1378, opened 2018-12-18
Normative: LeftHandSideExpression -> UnaryExpression in Static Semantics: Early Errors

rwaldron:
Fixes this spec bug: 

image


bterlson:
Merged, relabeled as editorial as I think that's accurate (we typically don't mark simple bugs as normative when the intent was clear from the start).

PR #1381, opened 2018-12-27
Editorial: Fix formatting of grammar productions ending in html entities

NilSet:
A bug in grammarkdown (which is fixed in master and not yet released)
causes a newline after an html entity to get consumed, making
alternatives run together and look like a seqence of symbols.

This commit works around the bug by adding a trailing space after html
entities in grammar productions.

See https://github.com/rbuckton/grammarkdown/pull/29 for the proper fix
which has yet to be released.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,6 @@
     "ecmarkup": "^3.16.0"
   },
   "devDependencies": {
-    "@alrra/travis-scripts": "^2.0.0"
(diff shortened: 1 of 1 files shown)


rbuckton:
`grammarkdown@2.0.12` has been published on npm with the fix mentioned above.

ljharb:
(I'll repurpose this to force the dep update and the rebuild)

PR #1387, opened 2019-01-02
Editorial: Clarify IEEE 754-2008 terminology

TimothyGu:
The name of the standard is properly "IEEE Standard for Floating-Point Arithmetic", not "IEEE Standard for Binary Floating-Point Arithmetic".

Appropriately refer to the binary64 format as "double-precision 64-bit binary format", rather than just "double-precision 64-bit format".

Replace the legacy "round to nearest, ties to even" rounding mode to the modern "roundTiesToEven" rounding direction mode.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1086,7 +1086,7 @@
 
     <emu-clause id="sec-ecmascript-language-types-number-type">
       <h1>The Number Type</h1>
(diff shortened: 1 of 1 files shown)


zenparsing:
Thanks @TimothyGu. Do we have another IEEE expert that can sign off on these changes?

apaprocki:
I can at least say that I have the specification and what is represented in the PR is correct and matches the text of the specification, so it’s a good editorial change.

PR #1393, opened 2019-01-04
Normative: Treat IterationStatements uniformly in Annex B.3.5.

### LINKED ISSUE #1392: Annex B.3.5: Should we stop making for-of a special case?
[Annex B.3.5](https://tc39.github.io/ecma262/#sec-variablestatements-in-catch-blocks), **VariableStatements in Catch Blocks**, says:
> It is a Syntax Error if any element of the BoundNames of CatchParameter also occurs in the VarDeclaredNames of Block unless CatchParameter is CatchParameter : BindingIdentifier and that element is only bound by a VariableStatement, the VariableDeclarationList of a for statement, the ForBinding of a for-in statement, or the BindingIdentifier of a for-in statement.

That is, we allow var-redeclaration of a non-destructured catch parameter...
```js
try {} catch (e) { var e; }
```
...except in the particular case where the var declaration is a for-of initializer.
```js
try {} catch (e) { for (var e of whatever) {} }
```
(_Note that this behavior applies equally to sloppy and strict modes._)

Based on some earlier discussion I see in #150, it seems like the motivation here was to prohibit as much as possible without jeopardizing web compatibility. Still, if the for-of case is neither easy to remember nor trivial to implement, I wonder whether it provides concrete benefit to anyone. Should we simplify it away?

-----

Current status in the major engines:
- [V8](https://github.com/v8/v8/blob/64c85cf403fe19a38c452b07b080368756bb079c/src/parsing/parser.cc#L1914-L1920) and [SpiderMonkey](https://hg.mozilla.org/mozilla-central/file/aac6849c6ff0/js/src/frontend/ParseContext.cpp#l461) fully implement Annex B.3.5 and have special codepaths for the for-of case
- [JSC](https://github.com/WebKit/webkit/blob/8b7a022d2061c6b306b9ed3204df37b8f1ede31c/Source/JavaScriptCore/parser/Parser.h#L1232-L1233) implements everything but the for-of case, for which it has a `FIXME` (and a [bug](https://bugs.webkit.org/show_bug.cgi?id=192830))
- ChakraCore allows any var-redeclaration of a catch parameter, destructured or otherwise

rkirsling:
Resolves #1392.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -40800,17 +40800,16 @@ THH:mm:ss.sss
           It is a Syntax Error if any element of the BoundNames of |CatchParameter| also occurs in the LexicallyDeclaredNames of |Block|.
         </li>
         <li>
-          It is a Syntax Error if any element of the BoundNames of |CatchParameter| also occurs in the VarDeclaredNames of |Block| unless |CatchParameter| is <emu-grammar>CatchParameter : BindingIdentifier</emu-grammar> and that element is only bound by a |VariableStatement|, the |VariableDeclarationList| of a for statement, the |ForBinding| of a for-in statement, or the |BindingIdentifier| of a for-in statement.
+          It is a Syntax Error if any element of the BoundNames of |CatchParameter| also occurs in the VarDeclaredNames of |Block| unless |CatchParameter| is <emu-grammar>CatchParameter : BindingIdentifier</emu-grammar>.
         </li>
       </ul>
       <emu-note>
-        <p>The |Block| of a |Catch| clause may contain `var` declarations that bind a name that is also bound by the |CatchParameter|. At runtime, such bindings are instantiated in the VariableDeclarationEnvironment. They do not shadow the same-named bindings introduced by the |CatchParameter| and hence the |Initializer| for such `var` declarations will assign to the corresponding catch parameter rather than the `var` binding. The relaxation of the normal static semantic rule does not apply to names only bound by for-of statements.</p>
+        <p>The |Block| of a |Catch| clause may contain `var` declarations that bind a name that is also bound by the |CatchParameter|. At runtime, such bindings are instantiated in the VariableDeclarationEnvironment. They do not shadow the same-named bindings introduced by the |CatchParameter| and hence the |Initializer| for such `var` declarations will assign to the corresponding catch parameter rather than the `var` binding.</p>
       </emu-note>
       <p>This modified behaviour also applies to `var` and `function` declarations introduced by direct eval calls contained within the |Block| of a |Catch| clause. This change is accomplished by modifying the algorithm of <emu-xref href="#sec-evaldeclarationinstantiation"></emu-xref> as follows:</p>
       <p>Step 5.d.ii.2.a.i is replaced by:</p>
       <emu-alg type="i">
         1. If _thisEnvRec_ is not the Environment Record for a |Catch| clause, throw a *SyntaxError* exception.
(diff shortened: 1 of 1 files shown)


bakkot:
> It is a Syntax Error if any element of the BoundNames of |CatchParameter| also occurs in the VarDeclaredNames of |Block| unless |CatchParameter| is <emu-grammar>CatchParameter : BindingIdentifier</emu-grammar> and **that element is only bound by a |VariableStatement|, or the |VariableDeclarationList|, |ForBinding|, or |BindingIdentifier| of an |IterationStatement|**.

Are there any other cases? I'm not sure the bolded part of this rule is ever unsatisfied.

rkirsling:
I think you're right, since function declarations add to a block's [LexicallyDeclaredNames](https://tc39.github.io/ecma262/#sec-block-static-semantics-lexicallydeclarednames) and not to its [VarDeclaredNames](https://tc39.github.io/ecma262/#sec-block-static-semantics-vardeclarednames).

I'm getting a bit confused trying to figure out what this implies for the second case though. Technically, what's written there is not exhaustive—if it read HoistableDeclaration instead of FunctionDeclaration, then we could delete the entire sentence, but as it reads now, it would not allow, e.g., GeneratorDeclaration.

However, it appears that not a single engine actually distinguishes among HoistableDeclarations here!
```js
// SyntaxError in V8, SM, XS; permitted in JSC, CH
try {} catch (e) { function e() {} }
try {} catch (e) { function* e() {} }
try {} catch (e) { async function e() {} }
try {} catch (e) { async function* e() {} }

// permitted in V8, SM, CH; SyntaxError in JSC, XS
try { throw 1; } catch (e) { eval('function e() {}'); }
try { throw 1; } catch (e) { eval('function* e() {}'); }
try { throw 1; } catch (e) { eval('async function e() {}'); }
try { throw 1; } catch (e) { eval('async function* e() {}'); }

// (Side note: CH doesn't actually have async generators now.)
```
So perhaps web reality suggests that we could do away with this whole sentence after all? 🤔

syg:
Your example gives rise to another piece of esoterica: currently Annex B.3.3 interaction with B.3.5 means the following, which may or may not be surprising to you depending on how long you've stared at the abyss:

```js
function foo() { try { throw 1; } catch (e) { { function e() {} } } print(e); }
foo() // prints "function e() {}"
```
but
```js
function bar() { try { throw 1; } catch (e) { function e() {} } print(e); } // SyntaxError
```

rkirsling:
@syg Thankfully though, that too applies equally to the four HoistableDeclarations. :smile:

rkirsling:
Made @bakkot's suggested simplification as well as the web reality simplification from https://github.com/tc39/ecma262/pull/1393#issuecomment-451587882.

rkirsling:
Implementation status, just for the record:
- Landed JSC patch: https://trac.webkit.org/r240686
- Opened SM bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1523720
- Opened V8 bug: https://crbug.com/v8/8760

mathiasbynens:
@rkirsling Are you gonna patch Test262 as well?

rkirsling:
@mathiasbynens https://github.com/tc39/test262/pull/2023 is just awaiting the green button. :smile:

rkirsling:
Tests merged, two implementations landed. 🎊
(Thanks @mathiasbynens!)

mathiasbynens:
…and now it [landed in SpiderMonkey](https://bugzilla.mozilla.org/show_bug.cgi?id=1523720) as well!

PR #1401, opened 2019-01-12
Editorial: Clarify internal slots during object creation

TimothyGu:
- Do not redundantly set initial value, as "Unless specified otherwise, the initial value of an internal slot is the value **undefined**."
- Do explicitly state what the internal slots are when creating a new exotic object.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7396,7 +7396,7 @@
         1. If _functionKind_ is `"normal"`, let _needsConstruct_ be *true*.
         1. Else, let _needsConstruct_ be *false*.
         1. If _functionKind_ is `"non-constructor"`, set _functionKind_ to `"normal"`.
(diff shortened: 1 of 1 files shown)


annevk:
Note that IDL/HTML isn't this explicit currently and I think even has some dynamic slots (though you could redo those in terms of a map or some such).

I wonder if we want to formalize this better if there shouldn't be some kind of Create abstract operation that handles allocation, internal slots, and internal methods. (And perhaps also the Realm it's created in, which is somewhat relevant in HTML and might also become more relevant in ECMAScript with the realms proposal.)

TimothyGu:
As clarified on IRC, @annevk actually meant adding fields to a Record after its creation, which is not affected by this change.

PR #1402, opened 2019-01-13
Editorial: Fix SubstitutionTemplate ArgumentListEvaluation

TimothyGu:
Previously it was referencing an undeclared _TemplateLiteral_.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13420,15 +13420,20 @@
           1. Let _siteObj_ be GetTemplateObject(_templateLiteral_).
           1. Return a List containing the one element which is _siteObj_.
         </emu-alg>
-        <emu-grammar>SubstitutionTemplate : TemplateHead Expression TemplateSpans</emu-grammar>
+        <emu-grammar>TemplateLiteral : SubstitutionTemplate</emu-grammar>
         <emu-alg>
           1. Let _templateLiteral_ be this |TemplateLiteral|.
           1. Let _siteObj_ be GetTemplateObject(_templateLiteral_).
+          1. Let _remaining_ be ? ArgumentListEvaluation of |SubstitutionTemplate|.
+          1. Return a List whose first element is _siteObj_ and whose subsequent elements are the elements of _remaining_, in order.
+        </emu-alg>
+        <emu-grammar>SubstitutionTemplate : TemplateHead Expression TemplateSpans</emu-grammar>
+        <emu-alg>
           1. Let _firstSubRef_ be the result of evaluating |Expression|.
           1. Let _firstSub_ be ? GetValue(_firstSubRef_).
           1. Let _restSub_ be ? SubstitutionEvaluation of |TemplateSpans|.
           1. Assert: _restSub_ is a List.
-          1. Return a List whose first element is _siteObj_, whose second elements is _firstSub_, and whose subsequent elements are the elements of _restSub_, in order. _restSub_ may contain no elements.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -11952,13 +11952,15 @@ <h1>Runtime Semantics: ArgumentListEvaluation</h1>
           1. Let _siteObj_ be GetTemplateObject(_templateLiteral_).
           1. Return a List containing the one element which is _siteObj_.
         </emu-alg>
-        <emu-grammar>SubstitutionTemplate : TemplateHead Expression TemplateSpans</emu-grammar>
+        <emu-grammar>TemplateLiteral : SubstitutionTemplate</emu-grammar>
         <emu-alg>
           1. Let _templateLiteral_ be this |TemplateLiteral|.
           1. Let _siteObj_ be GetTemplateObject(_templateLiteral_).
-          1. Let _firstSubRef_ be the result of evaluating |Expression|.
+          1. Let _first_ be the |Expression| of |SubstitutionTemplate|.
```
zenparsing: The pattern `|symbol| of |otherSymbol|` is currently only used in Annex B, I believe, and I'm concerned about encouraging it.

Is there another way to fix this bug that doesn't involve directly digging into the expansion of `SubstitutionTemplate`?

What do you think of defining `ArgumentListEvaluation` for *SubstitutionTemplate* as before, but where it doesn't return *siteObj*, and then define `ArgumentListEvaluation` for *TemplateLiteral : SubstitutionTemplate* which prepends the *siteObj* to that list?

jmdyck: > The pattern `|symbol| of |otherSymbol|` is currently only used in Annex B, I believe, ...

It's used a few times outside of Annex B:

Names and Keywords:
    `the |CodePoint| of the |UnicodeEscapeSequence|`
Async Function Definitions:
    `the |FormalParameters| of an |AsyncFunctionDeclaration|, ...`
    `the |BindingIdentifier| of a |FunctionExpression|, ...`
Tail Position Calls:
    `the |FunctionBody| of a |GeneratorBody|`
    `the |FunctionBody| of an |AsyncFunctionBody|`
    `the |FunctionBody| of an |AsyncGeneratorBody|`
Source Text Module Records:
    `the |ModuleSpecifier| of the |ImportDeclaration|`
    `the |ModuleSpecifier| of the |ExportDeclaration|`
Static Semantics: Early Errors (Patterns):
    `the |RegExpIdentifierName| of this production's |GroupName|`

(Not arguing for or against, just supplying info.)

TimothyGu: > What do you think of defining `ArgumentListEvaluation` for *SubstitutionTemplate* as before, but where it doesn't return *siteObj*, and then define `ArgumentListEvaluation` for *TemplateLiteral : SubstitutionTemplate* which prepends the *siteObj* to that list?

That's a fair approach, though as @jmdyck confirms I got the pattern from other places.

bakkot:
I'd be OK with either approach, but I mildly lean towards the one @zenparsing suggested.

@TimothyGu Did you want to pick this up again? If not, do you mind if I push commits to your branch? Otherwise I'll open a PR of my own in a week or two.

ljharb:
@bakkot the "allow edits" box is checked; i'd prefer we push to this one rather than opening a new one.

bakkot:
I've rebased this to resolve a merge conflict and pushed a commit applying @zenparsing's suggestion.

TimothyGu:
Apologies for the lack of response here. Please feel free to take over.

PR #1403, opened 2019-01-13
Editorial: Fix Completion handling for != and !==

TimothyGu:
The Abstract Equality Comparison operation may return abrupt completions, but the Strict Equality Comparison operation should always return Boolean values. Indicate that.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13793,6 +13793,7 @@
         1. Let _rref_ be the result of evaluating |RelationalExpression|.
         1. Let _rval_ be ? GetValue(_rref_).
         1. Let _r_ be the result of performing Abstract Equality Comparison _rval_ == _lval_.
+        1. ReturnIfAbrupt(_r_).
         1. If _r_ is *true*, return *false*. Otherwise, return *true*.
       </emu-alg>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -13768,6 +13769,7 @@ <h1>Runtime Semantics: Evaluation</h1>
         1. Let _rref_ be the result of evaluating |RelationalExpression|.
         1. Let _rval_ be ? GetValue(_rref_).
         1. Let _r_ be the result of performing Strict Equality Comparison _rval_ === _lval_.
+        1. Assert: Type(_r_) is Boolean.
```
zenparsing: Is `r` technically a completion record here? Should we be explicit and unwrap it?

TimothyGu:
@zenparsing Fixed, thanks.

jmdyck:
Wait, how can Abstract Equality Comparison return an abrupt completion?

ljharb:
If toString or valueOf throws?

jmdyck:
Ah. In that case, shouldn't its calls to `ToPrimitive` be prefixed with '?' ?

ljharb:
Yes, that’d be a great follow up :-)

PR #1404, opened 2019-01-13
Editorial: Fix typo in AsyncFunctionStart

TimothyGu:
We want to use undefined for return, not normal.


TimothyGu:
Wait, this is not quite correct… `(async () => { 3; })()` should obviously resolve to **undefined** and not 3…

PR #1406, opened 2019-01-13
Editorial: Fix return value logic for async arrow functions

TimothyGu:
Introduce a new production _ExpressionBody_ for both plain and async arrow functions, and allow correctly handling return value logic for both. Currently, _AsyncFunctionStart_ does not correctly handle the return value of async arrow functions correctly.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19825,8 +19825,11 @@
         CoverParenthesizedExpressionAndArrowParameterList[?Yield, ?Await] #parencover
 
       ConciseBody[In] :
-        [lookahead != `{` ] AssignmentExpression[?In, ~Yield, ~Await]
+        [lookahead != `{` ] ExpressionBody[?In, ~Await]
         `{` FunctionBody[~Yield, ~Await] `}`
+
+      ExpressionBody[In, Await] :
+        AssignmentExpression[?In, ~Yield, ?Await]
     </emu-grammar>
     <h2>Supplemental Syntax</h2>
     <p>When the production
@@ -19909,7 +19912,7 @@
     <emu-clause id="sec-arrow-function-definitions-static-semantics-containsusestrict">
       <h1>Static Semantics: ContainsUseStrict</h1>
       <emu-see-also-para op="ContainsUseStrict"></emu-see-also-para>
-      <emu-grammar>ConciseBody : AssignmentExpression</emu-grammar>
+      <emu-grammar>ConciseBody : ExpressionBody</emu-grammar>
       <emu-alg>
         1. Return *false*.
       </emu-alg>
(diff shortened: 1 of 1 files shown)


TimothyGu:
@zenparsing Yes I believe so

anba:
I wonder if a better approach to fix this issue is to add a new runtime semantics to evaluate AssignmentExpressions for arrow-function like contexts. 

[14.2.15 Runtime Semantics: EvaluateBody](https://tc39.github.io/ecma262/#sec-arrow-function-definitions-runtime-semantics-evaluatebody) has:
> 1. ...
> 2. Let exprRef be the result of evaluating AssignmentExpression.
> 3. Let exprValue be ? GetValue(exprRef).
> 4. Return Completion { [[Type]]: return, [[Value]]: exprValue, [[Target]]: empty }. 

which is exactly what we need here for async arrow functions, i.e. evaluate an AssignmentExpression and then convert the result into a Return completion. (Plus it also handles GetValue which also seems to be missing for async arrow functions.).

The same steps are executed in the class fields proposal at <https://tc39.github.io/proposal-class-fields/#runtime-semantics-class-public-field-definition-evaluation>, so we *soon* have three places which could use a single shared definition.

----

I haven't verified if this works, but the quick'n dirty approach to get this working may be to add a *wrapper* around AssignmentExpression, so we get:
```
FunkyAssignmentExpression[In, Yield, Await]:
  AssignmentExpression[In?, Yield?, Await?]

ConciseBody[In]:
  [lookahead ≠ {] FunkyAssignmentExpression[?In, ~Yield, ~Await]
  { FunctionBody[~Yield, ~Await] }

AsyncConciseBody[In]:
  [lookahead ≠ {] FunkyAssignmentExpression[?In, ~Yield, +Await]
  { AsyncFunctionBody }

FieldDefinition[Yield, Await]:
  ClassElementName[?Yield, ?Await] FieldInitializer opt

FieldInitializer:
  = FunkyAssignmentExpression[In, ~Yield, ~Await]
```

with:
```
Runtime Semantics: Evaluation

FunkyAssignmentExpression: AssignmentExpression
1. Let exprRef be the result of evaluating AssignmentExpression.
2. Let exprValue be ? GetValue(exprRef).
3. Return Completion { [[Type]]: return, [[Value]]: exprValue, [[Target]]: empty }. 
```

TimothyGu:
@anba I went with your approach. @tc39/ecma262-editors please take a look.

ljharb:
@TimothyGu can you rebase this and resolve the conflicts?

TimothyGu:
@ljharb Rebased

ljharb:
ping @waldemarhorwat

spec.html
```diff
@@ -20463,23 +20470,17 @@ <h1>Runtime Semantics: IteratorBindingInitialization</h1>
...
         1. Else,
           1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _declResult_.[[Value]] &raquo;).
         1. Return Completion { [[Type]]: ~return~, [[Value]]: _promiseCapability_.[[Promise]], [[Target]]: ~empty~ }.
       </emu-alg>
-      <emu-grammar>
-        AsyncConciseBody : `{` AsyncFunctionBody `}`
-      </emu-grammar>
-      <emu-alg>
-        1. Return the result of EvaluateBody of |AsyncFunctionBody| passing _functionObject_ and _argumentsList_ as the arguments.
-      </emu-alg>
```
syg: Musing: I wonder if we should have ecmarkup auto-generate chain production expansions somehow so the reader can get a clickable link for the RHS.

jmdyck: Theoretically possible, but I think it would go substantially beyond ecmarkup's current capabilities just to deduce the chain rules that have been left implicit.

PR #1407, opened 2019-01-14
Normative: Require at least four digits in string representations of negative years

### LINKED ISSUE #1035: Normative: Clarification on Date.prototype.to(UTC)String formatting of negative years.
@ljharb, @bterlson, @littledan 
/cc @xiaoyinl

The engines disagree with each other over how to format negative years in `Date.prototype.toString()`

See test case:

```
## Source
let year1BC = new Date("0001-10-13T05:16:33Z");
let year11BC = new Date("0001-10-13T05:16:33Z");
year1BC.setFullYear(-1, 10, 13);
year11BC.setFullYear(-11, 10, 13);

print(year1BC.toString());
print(year11BC.toString());
print(year1BC.toUTCString());
print(year11BC.toUTCString());


┌─────┬────────────────────────────────────────────────────────────┐
│ d8  │ Sat Nov 13 -001 22:16:33 GMT-0800 (Pacific Standard Time)  │
│ jsc │ Mon Nov 13 -011 22:16:33 GMT-0800 (Pacific Standard Time)  │
│     │ Sun, 14 Nov -001 06:16:33 GMT                              │
│     │ Tue, 14 Nov -011 06:16:33 GMT                              │
├─────┼────────────────────────────────────────────────────────────┤
│ sm  │ Sat Nov 13 -0001 22:16:33 GMT-0800 (Pacific Standard Time) │
│     │ Mon Nov 13 -0011 22:16:33 GMT-0800 (Pacific Standard Time) │
│     │ Sun, 14 Nov -0001 06:16:33 GMT                             │
│     │ Tue, 14 Nov -0011 06:16:33 GMT                             │
├─────┼────────────────────────────────────────────────────────────┤
│ ch  │ Sat Nov 13 -1 22:16:33 GMT-0800 (Pacific Standard Time)    │
│     │ Mon Nov 13 -11 22:16:33 GMT-0800 (Pacific Standard Time)   │
│     │ Sun, 14 Nov 2 B.C. 06:16:33 GMT                            │
│     │ Tue, 14 Nov 12 B.C. 06:16:33 GMT                           │
└─────┴────────────────────────────────────────────────────────────┘
```

In Microsoft/ChakraCore#4067 we are addressing the fact that ChakraCore does not pad the year with leading 0's and uses B.C. instead of a negative year. However the spec versus current plurality implementation seems to be unclear.

In our opinion (ChakraCore), SpiderMonkey's output is more true to the spec text. 

@xiaoyinl at https://github.com/Microsoft/ChakraCore/pull/4067#issuecomment-345651062 writes:

> [Section 20.3.4.43](https://tc39.github.io/ecma262/#sec-date.prototype.toutcstring) specifies the exact format of `toUTCString()`. It says:
> > 7. Let year be the String representation of YearFromTime(tv), formatted as a number of at least four digits, padded to the left with zeroes if necessary.

Because the spec reads that the year should be (emphasis mine):

> formatted as a number of at least four **digits**, padded to the left with zeroes if necessary.

Thus the spec seems to imply the digits should be treated separately (by padding the absolute value of the year out to 4 digits) and then if the year was negative, inserting a `-` before it.

As I wrote in https://github.com/Microsoft/ChakraCore/pull/4067#issuecomment-345914246:

> jsc and d8 behave as `printf` formatting a negative number with `printf("%04d", negativeNumber);`, which for `int negativeNumber = -3;` would print `-003`. (Java's `String.format` also agrees with that interpretation of `%04d` with a negative argument.)

> This is what `d8` and `jsc` have done in terms of output, and I'd imagine that the implemented the formatting behavior with something akin to `sprintf` instead of rolling their own as we've done here.

See discussion here:

https://github.com/Microsoft/ChakraCore/pull/4067#issuecomment-342685858
https://github.com/Microsoft/ChakraCore/pull/4067#issuecomment-345914246

gibson042:
Fixes #1035

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27839,8 +27839,9 @@ THH:mm:ss.sss
             1. Let _day_ be the String representation of DateFromTime(_tv_), formatted as a two-digit decimal number, padded to the left with a zero if necessary.
             1. Let _yv_ be YearFromTime(_tv_).
             1. If _yv_ &ge; 0, let _yearSign_ be the empty string; otherwise, let _yearSign_ be `"-"`.
-            1. Let _year_ be the String representation of abs(_yv_), formatted as a decimal number of at least four digits, padded to the left with zeroes if necessary.
-            1. Return the string-concatenation of _weekday_, the code unit 0x0020 (SPACE), _month_, the code unit 0x0020 (SPACE), _day_, the code unit 0x0020 (SPACE), _yearSign_, and _year_.
+            1. Let _year_ be the String representation of abs(_yv_), formatted as a decimal number.
+            1. Let _paddedYear_ be ! StringPad(_year_, 4, *"0"*, *"start"*).
+            1. Return the string-concatenation of _weekday_, the code unit 0x0020 (SPACE), _month_, the code unit 0x0020 (SPACE), _day_, the code unit 0x0020 (SPACE), _yearSign_, and _paddedYear_.
           </emu-alg>
           <emu-table id="sec-todatestring-day-names" caption="Names of days of the week">
             <table>
@@ -28075,8 +28076,9 @@ THH:mm:ss.sss
           1. Let _day_ be the String representation of DateFromTime(_tv_), formatted as a two-digit decimal number, padded to the left with a zero if necessary.
           1. Let _yv_ be YearFromTime(_tv_).
           1. If _yv_ &ge; 0, let _yearSign_ be the empty string; otherwise, let _yearSign_ be `"-"`.
-          1. Let _year_ be the String representation of abs(_yv_), formatted as a decimal number of at least four digits, padded to the left with zeroes if necessary.
-          1. Return the string-concatenation of _weekday_, `","`, the code unit 0x0020 (SPACE), _day_, the code unit 0x0020 (SPACE), _month_, the code unit 0x0020 (SPACE), _yearSign_, _year_, the code unit 0x0020 (SPACE), and TimeString(_tv_).
+          1. Let _year_ be the String representation of abs(_yv_), formatted as a decimal number.
+          1. Let _paddedYear_ be ! StringPad(_year_, 4, *"0"*, *"start"*).
+          1. Return the string-concatenation of _weekday_, `","`, the code unit 0x0020 (SPACE), _day_, the code unit 0x0020 (SPACE), _month_, the code unit 0x0020 (SPACE), _yearSign_, _paddedYear_, the code unit 0x0020 (SPACE), and TimeString(_tv_).
         </emu-alg>
       </emu-clause>
 
--- a/spec.html
+++ b/spec.html
@@ -28065,7 +28065,7 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-date.prototype.toutcstring">
         <h1>Date.prototype.toUTCString ( )</h1>
-        <p>The following steps are performed:</p>
+        <p>The `toUTCString` method returns a String value representing the instance in time corresponding to this time value. The format of the String is based upon "HTTP-date" from RFC 7231, generalized to support the full range of times supported by ECMAScript Date objects. It performs the following steps:</p>
         <emu-alg>
           1. Let _O_ be this Date object.
           1. Let _tv_ be ? thisTimeValue(_O_).
@@ -41492,6 +41492,9 @@ THH:mm:ss.sss
     </li>
     <li>
       <i>RFC 3629 &ldquo;UTF-8, a transformation format of ISO 10646&rdquo;</i>, available at &lt;<a href="https://tools.ietf.org/html/rfc3629">https://tools.ietf.org/html/rfc3629</a>&gt;
+    </li>
+    <li>
+      <i>RFC 7231 &ldquo;Hypertext Transfer Protocol (HTTP/1.1): Semantics and Content&rdquo;</i>, available at &lt;<a href="https://tools.ietf.org/html/rfc7231">https://tools.ietf.org/html/rfc7231</a>&gt;
     </li>
   </ol>
 </emu-annex>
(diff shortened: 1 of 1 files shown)


littledan:
Spec text looks good to me. FWIW, it seems like ChakraCore and SpiderMonkey currently ship these semantics.

spec.html
```diff
@@ -27574,8 +27574,10 @@ <h1>Runtime Semantics: DateString ( _tv_ )</h1>
             1. Let _weekday_ be the Name of the entry in <emu-xref href="#sec-todatestring-day-names"></emu-xref> with the Number WeekDay(_tv_).
             1. Let _month_ be the Name of the entry in <emu-xref href="#sec-todatestring-month-names"></emu-xref> with the Number MonthFromTime(_tv_).
             1. Let _day_ be the String representation of DateFromTime(_tv_), formatted as a two-digit decimal number, padded to the left with a zero if necessary.
-            1. Let _year_ be the String representation of YearFromTime(_tv_), formatted as a decimal number of at least four digits, padded to the left with zeroes if necessary.
-            1. Return the string-concatenation of _weekday_, the code unit 0x0020 (SPACE), _month_, the code unit 0x0020 (SPACE), _day_, the code unit 0x0020 (SPACE), and _year_.
+            1. Let _yv_ be YearFromTime(_tv_).
+            1. If _yv_ &ge; 0, let _yearSign_ be the empty string; otherwise, let _yearSign_ be `"-"`.
+            1. Let _year_ be the String representation of abs(_yv_), formatted as a decimal number of at least four digits, padded to the left with zeroes if necessary.
```
ljharb: i wonder if it might be useful here to extract out and share the internal logic for https://tc39.github.io/ecma262/#sec-string.prototype.padstart, such that you could call `PadStart(formatted as a decimal number of at least four digits, 4, '0')` or something?

(it'd be fine not to do that in this PR, ofc)

spec.html
```diff
@@ -27800,16 +27802,18 @@ <h1>Date.prototype.toTimeString ( )</h1>
 
       <emu-clause id="sec-date.prototype.toutcstring">
         <h1>Date.prototype.toUTCString ( )</h1>
-        <p>The following steps are performed:</p>
+        <p>The `toUTCString` method returns a String value representing the instance in time corresponding to this time value. The format of the String is based upon "HTTP-date" from RFC 7231, generalized to support the full range of times supported by ECMAScript Date objects. It performs the following steps:</p>
```
ljharb: can we make this link to the section below that has a URL?

spec.html
```diff
@@ -27800,16 +27802,18 @@ <h1>Date.prototype.toTimeString ( )</h1>
...
           1. Let _weekday_ be the Name of the entry in <emu-xref href="#sec-todatestring-day-names"></emu-xref> with the Number WeekDay(_tv_).
           1. Let _month_ be the Name of the entry in <emu-xref href="#sec-todatestring-month-names"></emu-xref> with the Number MonthFromTime(_tv_).
           1. Let _day_ be the String representation of DateFromTime(_tv_), formatted as a two-digit decimal number, padded to the left with a zero if necessary.
-          1. Let _year_ be the String representation of YearFromTime(_tv_), formatted as a decimal number of at least four digits, padded to the left with zeroes if necessary.
-          1. Return the string-concatenation of _weekday_, `","`, the code unit 0x0020 (SPACE), _day_, the code unit 0x0020 (SPACE), _month_, the code unit 0x0020 (SPACE), _year_, the code unit 0x0020 (SPACE), and TimeString(_tv_).
+          1. Let _yv_ be YearFromTime(_tv_).
+          1. If _yv_ &ge; 0, let _yearSign_ be the empty string; otherwise, let _yearSign_ be `"-"`.
+          1. Let _year_ be the String representation of abs(_yv_), formatted as a decimal number of at least four digits, padded to the left with zeroes if necessary.
```
ljharb: ditto here re PadStart

spec.html
gibson042: I agree, and contemplated doing so in this PR but ultimately decided not to. I think a followup should replace all use of "formatted as a decimal number of at least $N digits, padded to the left with zeroes if necessary" with logic extracted from String.prototype.padStart.

spec.html
gibson042: I followed the existing pattern for IETF RFC references (cf. https://ecma-international.org/ecma-262/9.0/#sec-uri-syntax-and-semantics and https://ecma-international.org/ecma-262/9.0/#sec-escape-string ), so I think any change to that would also belong in a separate PR.

gibson042:
Tests: https://github.com/tc39/test262/pull/2114

gibson042:
XS (and only XS) seems to always format negative years with six digits: https://circleci.com/gh/tc39/test262/340?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link#action-107

This is interesting because that behavior is required of `Date.prototype.toISOString` for conformance with ISO 8601, but `toDateString`/`toUTCString`/etc. have no such requirement. It superficially makes sense here, but gets trickier when we remember ISO _also_ requires a plus sign for representing years after 9999. In other words, the same conformance with ISO 8601 that motivates use of six digits to represent every negative year from those methods would also motivate use of a plus sign and six digits to represent large positive years.

I personally foresee more problems than benefits stemming from e.g. "Mon, 01 Jul +065535 10:23:42 GMT", so I'm inclined to keep the tests as they are (requiring the minimum number of digits greater than or equal to 4, with no tainting from ISO 8601 year expansion).

gibson042:
Update: I have introduced a StringPad operation as suggested above by @ljharb, for precisely defining the representation of negative years in a way that prohibits e.g. "-001234" (as undesirably exhibited by XS).

ljharb:
@gibson042 5e0b005bf0493c91275f0a8b1d8479c120595c31 seems editorial to me, and something i'd love to land separately/first. Would you be willing to make a separate PR for that commit? (you can leave it here, ofc, it will rebase out after the other one lands)

PR #1408, opened 2019-01-14
Normative: Suppress GetMethod errors in IteratorClose

### LINKED ISSUE #1398: Suppress GetMethod(iterator.return) errors in IteratorClose
IteratorClose is currently specified as

`IteratorClose ( iteratorRecord, completion )`
  1. Assert: `Type(iteratorRecord.[[Iterator]])` is `Object`.
  2. Assert: `completion` is a Completion Record.
  3. Let `iterator` be `iteratorRecord.[[Iterator]]`.
  4. Let `return` be `? GetMethod(iterator, "return")`.
  5. If `return` is undefined, return `Completion(completion)`.
  6. Let `innerResult` be `Call(return, iterator, « »)`.
  7. If `completion.[[Type]]` is `throw`, return `Completion(completion)`.
  8. If `innerResult.[[Type]]` is `throw`, return `Completion(innerResult)`.
  9. If `Type(innerResult.[[Value]])` is not `Object`, throw a `TypeError` exception.
  10. Return `Completion(completion)`.

If the outer completion is `throw`, then it is commonly understood that exceptions throw by iterator.return() should be suppressed. However, there is one exception that is not suppressed, which is iterator.return not being callable -- on 4., `? GetMethod(iterator, "return")` will return the abrupt completion of GetMethod (caused by IsCallable failing), rather than the outer completion.

Therefore, errors in `GetMethod(iterator, "return")` and in `Call(return, iterator, « »)` are inconsistent, and the access and call have to be split across an exception handling boundary (try-catch or similar). I suggest that we amend this to:

  3. ...
  4. Let `return` be `GetMethod(iterator, "return")`.
  5. If `return.[[Value]]` is undefined, return `Completion(completion)`.
  6. If `return.[[Type]]` is `throw` and `completion.[[Type]]` is `throw`, return `Completion(completion)`.
  7. Let `innerResult` be `Call(return.[[Value]], iterator, « »)`.
  8. ...

This is now consistent, and can be desugared more simply to wrap iterator.return() in a try-catch (or the  equivalent of such an operation in an engine's IR).

This should not be a web-compat issue, as it is already currently inconsistent across implementations: current versions of V8 and SpiderMonkey are spec compatible, but JSC, Chakra and XS are not. This can be tested with:

    (function() {
      try {
        var it = {
          [Symbol.iterator]: function() {
            return {
              i: 0,
              next: function() {
                if (this.i > 0) return { done: true }
                return { value: this.i++, done: false };
              },
              return: 2
            }
          }
        }

        for (var x of it) {
          throw 3;
        }
      } catch (e) {
        if (e == 3) {
          print("Not spec compatible")
        } else {
          print("Spec compatible")
        }
      }
    })();

Additionally, babel desugaring currently does not take this into account, desugaring the loop to something that does not check IsCallable:

    var _iteratorNormalCompletion = true;
    var _didIteratorError = false;
    var _iteratorError = undefined;

    try {
      for (var _iterator = it[Symbol.iterator](), _step; !(_iteratorNormalCompletion = (_step = _iterator.next()).done); _iteratorNormalCompletion = true) {
        var x = _step.value;

        throw 3;
      }
    } catch (err) {
      _didIteratorError = true;
      _iteratorError = err;
    } finally {
      try {
        if (!_iteratorNormalCompletion && _iterator.return) {
          _iterator.return();
        }
      } finally {
        if (_didIteratorError) {
          throw _iteratorError;
        }
      }
    }

A similar spec incompatibility can be constructed for array destructuring.

zenparsing:
Fixes #1398 

### Current behavior:

If the "return" property of an iterator is not callable, a TypeError will be thrown from `IteratorClose`, regardless of the completion provided as an argument.

### Proposed behavior:

If a "throw" completion is provided to `IteratorClose`, then it will rethrow that error even if the return property is not callable.

### Test case:

```js
let iter = {
  [Symbol.iterator]() { return this; },
  next() { return { value: 1, done: false }; },
  return: 0, // Not a function
};
try {
  for (let i of iter)
    throw 1; // Throw and trigger IteratorClose
} catch (e) {
  // Does this print "1" or "TypeError: 0 is not a function"?
  print(e);
}
```

### Current engine behavior:

```
eshost -e '(() => { try { for (let i of { [Symbol.iterator]() { return this }, next() { return { value: 1, done: false }; }, return: 0 }) throw 1; } catch (e) { print(e) } })()'

#### ch
1

#### jsc
1

#### v8
TypeError: The iterator's 'return' method is not callable

#### sm
TypeError: property 'return' of iterator is not callable

#### xs
1
```

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6109,9 +6109,11 @@
         1. Assert: Type(_iteratorRecord_.[[Iterator]]) is Object.
         1. Assert: _completion_ is a Completion Record.
         1. Let _iterator_ be _iteratorRecord_.[[Iterator]].
-        1. Let _return_ be ? GetMethod(_iterator_, *"return"*).
-        1. If _return_ is *undefined*, return Completion(_completion_).
-        1. Let _innerResult_ be Call(_return_, _iterator_).
+        1. Let _innerResult_ be GetMethod(_iterator_, *"return"*).
+        1. If _innerResult_.[[Type]] is ~normal~, then
+          1. Let _return_ be _innerResult_.[[Value]].
+          1. If _return_ is *undefined*, return Completion(_completion_).
+          1. Set _innerResult_ to Call(_return_, _iterator_).
         1. If _completion_.[[Type]] is ~throw~, return Completion(_completion_).
         1. If _innerResult_.[[Type]] is ~throw~, return Completion(_innerResult_).
         1. If Type(_innerResult_.[[Value]]) is not Object, throw a *TypeError* exception.
@@ -6126,10 +6128,12 @@
         1. Assert: Type(_iteratorRecord_.[[Iterator]]) is Object.
         1. Assert: _completion_ is a Completion Record.
         1. Let _iterator_ be _iteratorRecord_.[[Iterator]].
-        1. Let _return_ be ? GetMethod(_iterator_, *"return"*).
-        1. If _return_ is *undefined*, return Completion(_completion_).
-        1. Let _innerResult_ be Call(_return_, _iterator_).
-        1. If _innerResult_.[[Type]] is ~normal~, set _innerResult_ to Await(_innerResult_.[[Value]]).
+        1. Let _innerResult_ be GetMethod(_iterator_, *"return"*).
+        1. If _innerResult_.[[Type]] is ~normal~, then
+          1. Let _return_ be _innerResult_.[[Value]].
+          1. If _return_ is *undefined*, return Completion(_completion_).
+          1. Set _innerResult_ to Call(_return_, _iterator_).
+          1. If _innerResult_.[[Type]] is ~normal~, set _innerResult_ to Await(_innerResult_.[[Value]]).
         1. If _completion_.[[Type]] is ~throw~, return Completion(_completion_).
         1. If _innerResult_.[[Type]] is ~throw~, return Completion(_innerResult_).
         1. If Type(_innerResult_.[[Value]]) is not Object, throw a *TypeError* exception.


shvaikalesh:
Tests PR: tc39/test262#2565

spec.html
```diff
@@ -4912,9 +4912,11 @@ <h1>IteratorClose ( _iteratorRecord_, _completion_ )</h1>
...
         1. Assert: _completion_ is a Completion Record.
         1. Let _iterator_ be _iteratorRecord_.[[Iterator]].
-        1. Let _return_ be ? GetMethod(_iterator_, `"return"`).
-        1. If _return_ is *undefined*, return Completion(_completion_).
-        1. Let _innerResult_ be Call(_return_, _iterator_, &laquo; &raquo;).
+        1. Let _innerResult_ be GetMethod(_iterator_, `"return"`).
+        1. If _innerResult_.[[Type]] is ~normal~,
+          1. Let _return_ be _innerResult_.[[Value]].
+          1. If _return_ is *undefined*, return Completion(_completion_).
+          1. Set _innerResult_ to Call(_return_, _iterator_, &laquo; &raquo;).
```
shvaikalesh, line 4919: Last argument of [`Call`](https://tc39.es/ecma262/#sec-call) defaults to empty list.
Please also see https://github.com/tc39/ecma262/pull/1414.

spec.html
```diff
@@ -6113,9 +6113,11 @@ <h1>IteratorClose ( _iteratorRecord_, _completion_ )</h1>
         1. Assert: Type(_iteratorRecord_.[[Iterator]]) is Object.
         1. Assert: _completion_ is a Completion Record.
         1. Let _iterator_ be _iteratorRecord_.[[Iterator]].
-        1. Let _return_ be ? GetMethod(_iterator_, *"return"*).
-        1. If _return_ is *undefined*, return Completion(_completion_).
-        1. Let _innerResult_ be Call(_return_, _iterator_).
+        1. Let _innerResult_ be GetMethod(_iterator_, `"return"`).
```
bakkot, line 6116:
```suggestion
        1. Let _innerResult_ be GetMethod(_iterator_, *"return"*).
```

spec.html
```diff
@@ -6130,10 +6132,12 @@ <h1>AsyncIteratorClose ( _iteratorRecord_, _completion_ )</h1>
         1. Assert: Type(_iteratorRecord_.[[Iterator]]) is Object.
         1. Assert: _completion_ is a Completion Record.
         1. Let _iterator_ be _iteratorRecord_.[[Iterator]].
-        1. Let _return_ be ? GetMethod(_iterator_, *"return"*).
-        1. If _return_ is *undefined*, return Completion(_completion_).
-        1. Let _innerResult_ be Call(_return_, _iterator_).
-        1. If _innerResult_.[[Type]] is ~normal~, set _innerResult_ to Await(_innerResult_.[[Value]]).
+        1. Let _innerResult_ be GetMethod(_iterator_, `"return"`).
```
bakkot, line 6135:
```suggestion
        1. Let _innerResult_ be GetMethod(_iterator_, *"return"*).
```

spec.html
```diff
@@ -6130,10 +6132,12 @@ <h1>AsyncIteratorClose ( _iteratorRecord_, _completion_ )</h1>
...
+        1. If _innerResult_.[[Type]] is ~normal~,
```
syg, line 6136:
```suggestion
        1. If _innerResult_.[[Type]] is ~normal~, then
```

PR #1410, opened 2019-01-15
Editorial: Use IsExtensible abstract op

shvaikalesh:


anba:
I'd probably keep `Reflect.isExtensible` as is to match the other `Reflect` functions and to reduce the extra indirections. Otherwise we may end up with changing `Reflect.set` to call `7.3.3 Set`, `Reflect.get` to call `7.3.2 GetV`, and `Reflect.has` to call `7.3.10 HasProperty`. And then I wouldn't be too surprised if someone else starts proposing to add wrappers around the remaining MOP methods to section 7.3, so we end up with more indirections of dubious worth.

PR #1411, opened 2019-01-15
Editorial: Treat not present parameters as `undefined`

shvaikalesh:
Using missing parameters directly eliminates extra steps, resulting in more descriptive variable names and increased visibility for functions where "X is (not) present" is absolutely required. ES6+ era algorithms already do treat missing parameters as `undefined` and use them directly.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -25239,7 +25239,7 @@
         <emu-alg>
           1. If NewTarget is neither *undefined* nor the active function, then
             1. Return ? OrdinaryCreateFromConstructor(NewTarget, *"%Object.prototype%"*).
-          1. If _value_ is *null*, *undefined* or not supplied, return ObjectCreate(%Object.prototype%).
+          1. If _value_ is *undefined* or *null*, return ObjectCreate(%Object.prototype%).
           1. Return ! ToObject(_value_).
         </emu-alg>
         <p>The *"length"* property of the `Object` constructor function is 1.</p>
@@ -26568,8 +26568,8 @@
         <h1>Number ( _value_ )</h1>
         <p>When `Number` is called with argument _value_, the following steps are taken:</p>
         <emu-alg>
-          1. If no arguments were passed to this function invocation, let _n_ be *+0*.
-          1. Else, let _n_ be ? ToNumber(_value_).
+          1. If _value_ is present, let _n_ be ? ToNumber(_value_).
+          1. Else, let _n_ be *+0*.
           1. If NewTarget is *undefined*, return _n_.
           1. Let _O_ be ? OrdinaryCreateFromConstructor(NewTarget, *"%Number.prototype%"*, &laquo; [[NumberData]] &raquo;).
           1. Set _O_.[[NumberData]] to _n_.
@@ -26865,13 +26865,12 @@
       <emu-clause id="sec-number.prototype.tostring">
         <h1>Number.prototype.toString ( [ _radix_ ] )</h1>
         <emu-note>
-          <p>The optional _radix_ should be an integer value in the inclusive range 2 to 36. If _radix_ is not present or is *undefined* the Number 10 is used as the value of _radix_.</p>
+          <p>The optional _radix_ should be an integer value in the inclusive range 2 to 36. If _radix_ is *undefined* the Number 10 is used as the value of _radix_.</p>
         </emu-note>
         <p>The following steps are performed:</p>
         <emu-alg>
           1. Let _x_ be ? thisNumberValue(*this* value).
-          1. If _radix_ is not present, let _radixNumber_ be 10.
-          1. Else if _radix_ is *undefined*, let _radixNumber_ be 10.
+          1. If _radix_ is *undefined*, let _radixNumber_ be 10.
           1. Else, let _radixNumber_ be ? ToInteger(_radix_).
           1. If _radixNumber_ &lt; 2 or _radixNumber_ &gt; 36, throw a *RangeError* exception.
           1. If _radixNumber_ = 10, return ! ToString(_x_).
@@ -29287,7 +29286,7 @@ THH:mm:ss.sss
         <h1>String ( _value_ )</h1>
         <p>When `String` is called with argument _value_, the following steps are taken:</p>
         <emu-alg>
-          1. If no arguments were passed to this function invocation, let _s_ be *""*.
+          1. If _value_ is not present, let _s_ be the empty String.
           1. Else,
             1. If NewTarget is *undefined* and Type(_value_) is Symbol, return SymbolDescriptiveString(_value_).
             1. Let _s_ be ? ToString(_value_).
@@ -29661,8 +29660,8 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _O_ be ? RequireObjectCoercible(*this* value).
           1. Let _S_ be ? ToString(_O_).
-          1. If _form_ is not present or _form_ is *undefined*, set _form_ to *"NFC"*.
-          1. Let _f_ be ? ToString(_form_).
+          1. If _form_ is *undefined*, let _f_ be *"NFC"*.
+          1. Else, let _f_ be ? ToString(_form_).
           1. If _f_ is not one of *"NFC"*, *"NFD"*, *"NFKC"*, or *"NFKD"*, throw a *RangeError* exception.
           1. Let _ns_ be the String value that is the result of normalizing _S_ into the normalization form named by _f_ as specified in <a href="https://unicode.org/reports/tr15/">https://unicode.org/reports/tr15/</a>.
           1. Return _ns_.
@@ -32747,7 +32746,6 @@ THH:mm:ss.sss
           1. If _mapfn_ is *undefined*, let _mapping_ be *false*.
           1. Else,
             1. If IsCallable(_mapfn_) is *false*, throw a *TypeError* exception.
-            1. If _thisArg_ is present, let _T_ be _thisArg_; else let _T_ be *undefined*.
             1. Let _mapping_ be *true*.
           1. Let _usingIterator_ be ? GetMethod(_items_, @@iterator).
           1. If _usingIterator_ is not *undefined*, then
@@ -32768,7 +32766,7 @@ THH:mm:ss.sss
                 1. Return _A_.
               1. Let _nextValue_ be ? IteratorValue(_next_).
               1. If _mapping_ is *true*, then
-                1. Let _mappedValue_ be Call(_mapfn_, _T_, &laquo; _nextValue_, _k_ &raquo;).
+                1. Let _mappedValue_ be Call(_mapfn_, _thisArg_, &laquo; _nextValue_, _k_ &raquo;).
                 1. If _mappedValue_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _mappedValue_).
                 1. Set _mappedValue_ to _mappedValue_.[[Value]].
               1. Else, let _mappedValue_ be _nextValue_.
@@ -32787,7 +32785,7 @@ THH:mm:ss.sss
             1. Let _Pk_ be ! ToString(_k_).
             1. Let _kValue_ be ? Get(_arrayLike_, _Pk_).
             1. If _mapping_ is *true*, then
-              1. Let _mappedValue_ be ? Call(_mapfn_, _T_, &laquo; _kValue_, _k_ &raquo;).
+              1. Let _mappedValue_ be ? Call(_mapfn_, _thisArg_, &laquo; _kValue_, _k_ &raquo;).
             1. Else, let _mappedValue_ be _kValue_.
             1. Perform ? CreateDataPropertyOrThrow(_A_, _Pk_, _mappedValue_).
             1. Set _k_ to _k_ + 1.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -25585,13 +25585,12 @@ <h1>Number.prototype.toPrecision ( _precision_ )</h1>
       <emu-clause id="sec-number.prototype.tostring">
         <h1>Number.prototype.toString ( [ _radix_ ] )</h1>
         <emu-note>
-          <p>The optional _radix_ should be an integer value in the inclusive range 2 to 36. If _radix_ is not present or is *undefined* the Number 10 is used as the value of _radix_.</p>
+          <p>The optional _radix_ should be an integer value in the inclusive range 2 to 36. If _radix_ is *undefined*, 0 is assumed.</p>
```
```suggestion
          <p>The optional _radix_ should be an integer value in the inclusive range 2 to 36. If _radix_ is *undefined*, 10 is assumed.</p>
```

spec.html
```diff
@@ -27868,10 +27867,9 @@ <h1>The String Constructor</h1>
         <h1>String ( _value_ )</h1>
         <p>When `String` is called with argument _value_, the following steps are taken:</p>
         <emu-alg>
-          1. If no arguments were passed to this function invocation, let _s_ be `""`.
-          1. Else,
-            1. If NewTarget is *undefined* and Type(_value_) is Symbol, return SymbolDescriptiveString(_value_).
-            1. Let _s_ be ? ToString(_value_).
+          1. If _value_ is not present, set _value_ to `""`.
+          1. If NewTarget is *undefined* and Type(_value_) is Symbol, return SymbolDescriptiveString(_value_).
+          1. Let _s_ be ? ToString(_value_).
```
ljharb: This call used to not happen when no arguments were passed; now it happens unconditionally. I'm not sure this is an improvement. Why de-nest these steps?

spec.html
shvaikalesh: Fixed in 9918c5b: "suggested change" feature seems broken.

spec.html
```diff
@@ -6348,7 +6348,7 @@ <h1>GetActiveScriptOrModule ( )</h1>
       <h1>ResolveBinding ( _name_ [ , _env_ ] )</h1>
       <p>The ResolveBinding abstract operation is used to determine the binding of _name_ passed as a String value. The optional argument _env_ can be used to explicitly provide the Lexical Environment that is to be searched for the binding. During execution of ECMAScript code, ResolveBinding is performed using the following algorithm:</p>
       <emu-alg>
-        1. If _env_ is not present or if _env_ is *undefined*, then
+        1. If _env_ is *undefined*, then
```
zenparsing: I don't think we can currently assume that "not present" is equivalent to "undefined" for abstract operations.

spec.html
```diff
@@ -33653,7 +33641,7 @@ <h1>AddEntriesFromIterable ( _target_, _iterable_, _adder_ )</h1>
         <p>The abstract operation AddEntriesFromIterable accepts a _target_ object, an _iterable_ of entries, and an _adder_ function to be invoked, with _target_ as the receiver.</p>
         <emu-alg>
           1. If IsCallable(_adder_) is *false*, throw a *TypeError* exception.
-          1. Assert: _iterable_ is present, and is neither *undefined* nor *null*.
+          1. Assert: _iterable_ is neither *undefined* nor *null*.
```
zenparsing: See above: this is another abstract operation.

spec.html
```diff
@@ -34662,7 +34646,6 @@ <h1>DetachArrayBuffer ( _arrayBuffer_ [ , _key_ ] )</h1>
         <emu-alg>
           1. Assert: Type(_arrayBuffer_) is Object and it has [[ArrayBufferData]], [[ArrayBufferByteLength]], and [[ArrayBufferDetachKey]] internal slots.
           1. Assert: IsSharedArrayBuffer(_arrayBuffer_) is *false*.
-          1. If _key_ is not present, set _key_ to *undefined*.
```
zenparsing: See above: this is an abstract operation.

spec.html
```diff
@@ -38060,8 +38043,6 @@ <h1>PerformPromiseThen ( _promise_, _onFulfilled_, _onRejected_ [ , _resultCapab
             1. Assert: IsPromise(_promise_) is *true*.
             1. If _resultCapability_ is present, then
               1. Assert: _resultCapability_ is a PromiseCapability Record.
-            1. Else,
```
zenparsing: See above: this is an abstract operation.

spec.html
```diff
@@ -27868,7 +27867,7 @@ <h1>The String Constructor</h1>
         <h1>String ( _value_ )</h1>
         <p>When `String` is called with argument _value_, the following steps are taken:</p>
         <emu-alg>
-          1. If no arguments were passed to this function invocation, let _s_ be `""`.
+          1. If _value_ is not present, set _value_ to `""`.
```
jmdyck: Presumably that should be `... set _s_ ...`

```suggestion
          1. If _value_ is not present, set _s_ to the empty String.
```

spec.html
bterlson: This is correct, however I would be open to some clause 5 changes to make this work for abstract ops. I was also confused by this pattern until Allen explained it to me some years ago.

spec.html
```diff
@@ -31383,7 +31381,7 @@ <h1>Array.prototype.entries ( )</h1>
         <h1>Array.prototype.every ( _callbackfn_ [ , _thisArg_ ] )</h1>
         <emu-note>
           <p>_callbackfn_ should be a function that accepts three arguments and returns a value that is coercible to the Boolean value *true* or *false*. `every` calls _callbackfn_ once for each element present in the array, in ascending order, until it finds one where _callbackfn_ returns *false*. If such an element is found, `every` immediately returns *false*. Otherwise, if _callbackfn_ returned *true* for all elements, `every` will return *true*. _callbackfn_ is called only for elements of the array which actually exist; it is not called for missing elements of the array.</p>
-          <p>If a _thisArg_ parameter is provided, it will be used as the *this* value for each invocation of _callbackfn_. If it is not provided, *undefined* is used instead.</p>
+          <p>_thisArg_ parameter is used as the *this* value for each invocation of _callbackfn_.</p>
```
zenparsing: I feel that the non-normative text here was more clear before the change. @ljharb what do you think?

ljharb: You might be right; it's a bit tricky given that callback might be in sloppy mode.

shvaikalesh: Thanks for pointing it out, I've reverted all non-normative changes.

jmdyck:
@shvaikalesh 
> resulting in more descriptive variable names

I don't understand this claim. Can you give an example?

shvaikalesh:
@jmdyck

That claim was regarding e3f2a97: `_thisArg_` is more descriptive than `T`.

jmdyck:
@shvaikalesh: Ah, thanks. I agree that `thisArg` is more descriptive than `T`.

Alternatively, one can avoid introducing `T` by replacing steps of the form:
```If _thisArg_ is present, let _T_ be _thisArg_; else let _T_ be *undefined*.```
with:
```If _thisArg_ is not present, set _thisArg_ to *undefined*.```

**Comment by michaelficarra** [comment]
@shvaikalesh Can you add wording to the Notational Conventions section to address [the concerns](https://github.com/tc39/ecma262/pull/1411#discussion_r248098984) of @zenparsing and @bterlson (and me) above?

File: spec.html [hunk 9]
```diff
@@ -25279,7 +25279,7 @@ <h1>Object ( [ _value_ ] )</h1>
         <emu-alg>
           1. If NewTarget is neither *undefined* nor the active function, then
             1. Return ? OrdinaryCreateFromConstructor(NewTarget, *"%Object.prototype%"*).
-          1. If _value_ is *null*, *undefined* or not supplied, return ObjectCreate(%Object.prototype%).
+          1. If _value_ is *undefined* or *null*, return ObjectCreate(%ObjectPrototype%).
```
michaelficarra: I think you accidentally changed `%Object.prototype%`.

spec.html
```diff
@@ -26906,13 +26906,12 @@ <h1>Number.prototype.toPrecision ( _precision_ )</h1>
       <emu-clause id="sec-number.prototype.tostring">
         <h1>Number.prototype.toString ( [ _radix_ ] )</h1>
         <emu-note>
-          <p>The optional _radix_ should be an integer value in the inclusive range 2 to 36. If _radix_ is not present or is *undefined* the Number 10 is used as the value of _radix_.</p>
+          <p>The optional _radix_ should be an integer value in the inclusive range 2 to 36. If _radix_ is *undefined*, 10 is assumed.</p>
```
michaelficarra: I don't like the "assumed" here. "the Number 10 is used as the value of _radix_" was fine IMO.

shvaikalesh:
> On this basis, the "missing -> *undefined*" equivalence is not specified to hold for optional parameters. Algorithms for built-in functions are obliged to explicitly handle missing arguments for optional parameters (typically via "is [not] present").

There are quite a few methods and constructors that assume "missing -> `undefined`" equivalence for optional parameters:

<details>
<summary>Complete list (22 functions, 25 parameters)</summary>

1. [`String.prototype.endsWith`](https://tc39.es/ecma262/#sec-string.prototype.endswith), `endPosition` at step 7
1. [`String.prototype.includes`](https://tc39.es/ecma262/#sec-string.prototype.includes), `position` at step 6
1. [`String.prototype.indexOf`](https://tc39.es/ecma262/#sec-string.prototype.indexof), `position` at step 4
1. [`String.prototype.lastIndexOf`](https://tc39.es/ecma262/#sec-string.prototype.lastindexof), `position` at step 4
1. [`String.prototype.padEnd`](https://tc39.es/ecma262/#sec-string.prototype.padend), `fillString` at step 2
1. [`String.prototype.padStart`](https://tc39.es/ecma262/#sec-string.prototype.padstart), `fillString` at step 2
1. [`String.prototype.startsWith`](https://tc39.es/ecma262/#sec-string.prototype.startswith), `position` at step 6
1. [`Array.from`](https://tc39.es/ecma262/#sec-array.from), `mapfn` at step 2
1. [`Array.prototype.copyWithin`](https://tc39.es/ecma262/#sec-array.prototype.copywithin), `end` at step 7
1. [`Array.prototype.fill`](https://tc39.es/ecma262/#sec-array.prototype.fill), `start` at step 3 and `end` at step 5
1. [`Array.prototype.flat`](https://tc39.es/ecma262/#sec-array.prototype.flat), `depth` at step 4
1. [`Array.prototype.includes`](https://tc39.es/ecma262/#sec-array.prototype.includes), `fromIndex` at step 4
1. [`Array.prototype.indexOf`](https://tc39.es/ecma262/#sec-array.prototype.indexof), `fromIndex` at step 4
1. [`Symbol`](https://tc39.es/ecma262/#sec-symbol-description), `description` at step 2
1. [`SharedArrayBuffer`](https://tc39.es/ecma262/#sec-sharedarraybuffer-length), `length` at step 2
1. [`DataView`](https://tc39.es/ecma262/#sec-dataview-buffer-byteoffset-bytelength), `byteOffset` at step 3
1. [`%TypedArray%`](https://tc39.es/ecma262/#sec-typedarray-buffer-byteoffset-length), `byteOffset` at step 6
1. [`%TypedArray%.prototype.copyWithin`](https://tc39.es/ecma262/#sec-%typedarray%.prototype.copywithin), `end` at step 8
1. [`%TypedArray%.prototype.fill`](https://tc39.es/ecma262/#sec-%typedarray%.prototype.fill), `start` at step 6 and `end` at step 8
1. [`%TypedArray%.prototype.set`](https://tc39.es/ecma262/#sec-%typedarray%.prototype.set-array-offset), `offset` at step 5
1. [`JSON.parse`](https://tc39.es/ecma262/#sec-json.parse), `reviver` at step 7
1. [`JSON.stringify`](https://tc39.es/ecma262/#sec-json.stringify), `replacer` at step 4 and `space` at step 5

</details>

Are [Clause 17](https://tc39.es/ecma262/#sec-ecmascript-standard-built-in-objects) changes welcome?

spec.html
shvaikalesh: Thank you, fixed in 0afecec.

spec.html
shvaikalesh: Resolved in 4178c0d.

spec.html
```diff
@@ -26609,8 +26609,8 @@ <h1>The Number Constructor</h1>
         <h1>Number ( _value_ )</h1>
         <p>When `Number` is called with argument _value_, the following steps are taken:</p>
         <emu-alg>
-          1. If no arguments were passed to this function invocation, let _n_ be *+0*.
-          1. Else, let _n_ be ? ToNumber(_value_).
+          1. If _value_ is present, let _n_ be ? ToNumber(_value_).
```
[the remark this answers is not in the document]
michaelficarra: I don't think it's odd. It's a mandatory argument (`Number.length` is `1`), and `Number()` is different than `Number(void 0)`.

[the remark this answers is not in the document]
jmdyck: > It's a mandatory parameter (Number.length is 1),

It could be made an 'optional' parameter and Number.length could still be defined to be 1.

> and `Number()` is different than `Number(void 0).`

Agreed. But that just means that the algorithm has to distinguish between "not present" and `*undefined*`, which I'm not questioning.

> Checking the presence of a mandatory parameter is the "otherwise specified".

Agreed. So it's allowed, but it's still odd. Or maybe I'll downgrade to "unusual". (I.e., I think `Number()` and `String()` are the only built-ins where the spec checks for the presence of a 'mandatory' parameter.)

When you say "it's necessary", are you saying that the presence of the "unless otherwise specified" phrase *necessitates* that there be some place in the spec where it *is* otherwise specified? Because I'm pretty sure there are several "unless otherwise specified" sentences where it never is otherwise specified.

---

Here's a thought experiment: if the spec explicitly gave the value of the `length` property for every built-in function, would there be any useful distinction between 'optional' and 'mandatory' arguments? (I had *thought* that one distinction was that optionals didn't automatically get undefined-for-missing, and so would need an explicit 'present' test, but apparently that's not the case in practice.)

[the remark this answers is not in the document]
ljharb: @jmdyck currently, the only explicit lengths are for where it differs from "the number of non-optional params in the signature", all of which are legacy.

imo every argument should be implicitly undefined when omitted, and may separately be present or absent.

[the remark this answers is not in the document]
jmdyck: > @jmdyck currently, the only explicit lengths are for where it differs from "the number of non-optional params in the signature",

(That's not quite true. E.g., the `length` of the `Array` constructor is given explicitly as 1, which is the same as its "default argument count". Similarly, the `length` of the `%TypedArray%` constructor is explicitly given as 0.)

> all of which are legacy.

Well, `Number()` and `String()` have both been around since ES1, so surely they qualify as legacy.

> imo every argument should be implicitly undefined when omitted, and may separately be present or absent.

Right, so to that extent, you're not drawing a distinction between 'optional' and 'required' parameters.

devsnek:
This left out Array.prototype.flatMap. I can open a PR soonish if no one else does.

shvaikalesh:
Thank you @devsnek. `BigInt.prototype.toString` was also left out, and possibly something else. I am making a full sweep now and will open a follow-up PR today.

devsnek:
@shvaikalesh #1781

PR #1412, opened 2019-01-15
Editorial: Make type checking explicit

shvaikalesh:


spec.html
```diff
@@ -9213,7 +9214,6 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
         1. Let _target_ be _O_.[[ProxyTarget]].
         1. Let _trap_ be ? GetMethod(_handler_, `"construct"`).
         1. If _trap_ is *undefined*, then
-          1. Assert: IsConstructor(_target_) is *true*.
```
shvaikalesh: `Construct` on the next step asserts this.

spec.html
```diff
@@ -8900,7 +8902,6 @@ <h1>[[GetOwnProperty]] ( _P_ )</h1>
           1. If _targetDesc_ is *undefined*, return *undefined*.
           1. If _targetDesc_.[[Configurable]] is *false*, throw a *TypeError* exception.
           1. Let _extensibleTarget_ be ? IsExtensible(_target_).
-          1. Assert: Type(_extensibleTarget_) is Boolean.
```
shvaikalesh: Description of `IsExtensible` states this.

spec.html
```diff
@@ -24457,7 +24457,6 @@ <h1>Runtime Semantics: CreateDynamicFunction ( _constructor_, _newTarget_, _kind
               1. Let _parameterGoal_ be the grammar symbol |FormalParameters[+Yield, ~Await]|.
               1. Let _fallbackProto_ be `"%Generator%"`.
             1. Else if _kind_ is `"async"`, then
-              1. Assert: _kind_ is `"async"`.
```
ljharb: perhaps instead of removing this assertion, the previous line should be changed to just "else"?

shvaikalesh: `Else if` on previous line is not the last one, please see expanded diff:

```diff
  1. If _kind_ is `"normal"`, then
    1. Let _goal_ be the grammar symbol |FunctionBody[~Yield, ~Await]|.
    1. Let _parameterGoal_ be the grammar symbol |FormalParameters[~Yield, ~Await]|.
    1. Let _fallbackProto_ be `"%FunctionPrototype%"`.
  1. Else if _kind_ is `"generator"`, then
    1. Let _goal_ be the grammar symbol |GeneratorBody|.
    1. Let _parameterGoal_ be the grammar symbol |FormalParameters[+Yield, ~Await]|.
    1. Let _fallbackProto_ be `"%Generator%"`.
  1. Else if _kind_ is `"async"`, then
-   1. Assert: _kind_ is `"async"`.
    1. Let _goal_ be the grammar symbol |AsyncFunctionBody|.
    1. Let _parameterGoal_ be the grammar symbol |FormalParameters[~Yield, +Await]|.
    1. Let _fallbackProto_ be `"%AsyncFunctionPrototype%"`.
  1. Else,
    1. Assert: _kind_ is `"async generator"`.
    1. Let _goal_ be the grammar symbol |AsyncGeneratorBody|.
    1. Let _parameterGoal_ be the grammar symbol |FormalParameters[+Yield, +Await]|.
    1. Let _fallbackProto_ be `"%AsyncGenerator%"`.
```

spec.html
zenparsing: I think the assertion here adds clarity. Can we leave it?

shvaikalesh: I've hoisted the assertion: this internal method is never created (and invoked) on a proxy if target is not a constructor.

PR #1413, opened 2019-01-15
Editorial: Remove direct inverse asserts

shvaikalesh:


spec.html
```diff
@@ -20046,10 +20046,10 @@ <h1>Runtime Semantics: EvaluateBody</h1>
       <emu-alg>
         1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
         1. Let _declResult_ be FunctionDeclarationInstantiation(_functionObject_, _argumentsList_).
-        1. If _declResult_ is not an abrupt completion, then
```
ljharb: personally this ordering reads better to me (success case, then failure case)

spec.html
```diff
@@ -20334,10 +20334,10 @@ <h1>Runtime Semantics: EvaluateBody</h1>
...
-          1. Perform ! AsyncFunctionStart(_promiseCapability_, |AssignmentExpression|).
```
ljharb: also here

spec.html
```diff
@@ -33396,12 +33396,12 @@ <h1>Runtime Semantics: AllocateTypedArray ( _constructorName_, _newTarget_, _def
             1. Let _obj_ be IntegerIndexedObjectCreate(_proto_, &laquo; [[ViewedArrayBuffer]], [[TypedArrayName]], [[ByteLength]], [[ByteOffset]], [[ArrayLength]] &raquo;).
             1. Assert: _obj_.[[ViewedArrayBuffer]] is *undefined*.
             1. Set _obj_.[[TypedArrayName]] to _constructorName_.
-            1. If _length_ is not present, then
```
ljharb: also here, defaults first reads better to me.

spec.html
```diff
@@ -8958,7 +8958,7 @@ <h1>[[DefineOwnProperty]] ( _P_, _Desc_ )</h1>
         1. If _targetDesc_ is *undefined*, then
           1. If _extensibleTarget_ is *false*, throw a *TypeError* exception.
           1. If _settingConfigFalse_ is *true*, throw a *TypeError* exception.
-        1. Else _targetDesc_ is not *undefined*,
```
jmdyck: The phrase `_targetDesc_ is not *undefined*` is, in effect, an inline assertion. Similar for the other two in this commit.

There are about 50 of these in the spec.

ljharb: that's true, but what's the point of the assertion here? `if (x) { } else { assert(!x); }` doesn't seem particularly useful :-/

shvaikalesh: If we agree that such assertions are extra, I am happy to remove them all in this PR.

jmdyck: > `if (x) { } else { assert(!x); }` doesn't seem particularly useful :-/

My guess is, that depends on how much is between the first pair of braces. And probably varies from person to person.

Also, I think there are a few cases where the else-assertion isn't simply the logical negation of the if-condition (where you might find the assertion more useful).

ljharb: I can’t conceive of that (in a single if/else; i can imagine the usefulness in an if/else if chain) do you have an example?

jmdyck: E.g., [GetValue](https://tc39.github.io/ecma262/#sec-getvalue) has:

    1. If IsPropertyReference(_V_) is *true*, then ...
    1. Else _base_ must be an Environment Record, ...

Other examples in:

    SetIntegrityLevel
    RunJobs
    OrdinaryGetOwnProperty
    ValidateAndApplyPropertyDescriptor
    OrdinarySetWithOwnDescriptor
    [[Get]] for Arguments Exotic Objects
    Evaluation of Delete operator
    Number.prototype.toPrecision
    String.prototype.split
    RegExpBuiltinExec
    Array.prototype.concat
    Array.prototype.copyWithin
    Array.prototype.pop
    Array.prototype.shift
    Array.prototype.splice
    Array.prototype.unshift

There might be more.

ljharb: In that one case I’d say a separate assertion is way clearer to me than the inline one, but fair point that it’s a common pattern.

jmdyck: I'm fine with replacing (some) else-inline-assertions with separate `Assert` steps. I just object to simply deleting all of them.

zenparsing: Agreed.

ljharb: In that case, perhaps this PR should be updated to delete only the inline `else` assertions in `if/else` chains that are a direct inverse of the `if`, and we can leave the rest for another PR?

jmdyck: I agree that In Number.prototype.toPrecision, steps 10 and 13 aren't examples of the situation I described. I wasn't talking about those, I was looking at step 10.c.iv.

All the others are of the form:

    1. If _X_ is *true*, then ...
    1. Else _X_ is *false*, ...

While it's debatable whether, in these cases, the else-assertion is "useful", it *is* an example of the situation I described, where "the else-assertion isn't simply the logical negation of the if-condition". In the logical negation of `_X_ is *true*`, you only know that `_X_` isn't `*true*` -- there's a vast universe of values that it could be, of which `*false*` is only one. (Of course, in these cases, there's an additional constraint that `_X_` is a Boolean value, which causes the else-assertion to be correct.)

shvaikalesh:
I've came up with a simple script that detects direct inverse asserts in `Else` clauses (simple cases only) and reduced this PR to removing only them.

spec.html
```diff
@@ -25479,7 +25479,7 @@ <h1>Number.prototype.toExponential ( _fractionDigits_ )</h1>
             1. Let _d_ be `"0"`.
           1. Else,
             1. If _e_ &gt; 0, let _c_ be `"+"`.
-            1. Else _e_ &le; 0,
+            1. Else _e_ &lt; 0,
```
shvaikalesh: `e` can't be equal `0` here: please see condition on line `25477`.

ljharb: in that case, can this be `else` with a separate `Assert`?

PR #1416, opened 2019-01-15
Editorial: Tweak String.prototype.repeat

shvaikalesh:


spec.html
```diff
@@ -28287,8 +28287,8 @@ <h1>String.prototype.repeat ( _count_ )</h1>
           1. Let _n_ be ? ToInteger(_count_).
           1. If _n_ &lt; 0, throw a *RangeError* exception.
           1. If _n_ is *+&infin;*, throw a *RangeError* exception.
-          1. Let _T_ be the String value that is made from _n_ copies of _S_ appended together. If _n_ is 0, _T_ is the empty String.
-          1. Return _T_.
+          1. If _n_ is 0, return the empty String `""`.
```
zenparsing: We can just use `the empty String` here.

PR #1418, opened 2019-01-18
Editorial: Explain why *_SAFE_INTEGER magnitude is not a power of two

gibson042:
The current text for MAX_SAFE_INTEGER and MIN_SAFE_INTEGER is somewhat indirect, resulting in repeated confusion and re-explanation of what those values signify and why they should not have magnitude 2<sup>53</sup> (i.e., failing to indicate the relevance of "n + 1" and "n - 1"). It also comes up in other contexts, cf. https://github.com/httpwg/http-extensions/issues/737#issuecomment-452189033 .

This PR attempts to remedy the issue with both prose and simple code examples of collision at 2<sup>53</sup>.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -30946,6 +30946,9 @@
 
       <emu-clause id="sec-number.issafeinteger">
         <h1>Number.isSafeInteger ( _number_ )</h1>
+        <emu-note>
+          <p>An integer _n_ is considered "<dfn id="safe-integer">safe</dfn>" if and only if the Number value for _n_ is not the Number value for any other integer.</p>
+        </emu-note>
         <p>This function performs the following steps when called:</p>
         <emu-alg>
           1. If IsIntegralNumber(_number_) is *true*, then
@@ -30957,7 +30960,7 @@
       <emu-clause id="sec-number.max_safe_integer">
         <h1>Number.MAX_SAFE_INTEGER</h1>
         <emu-note>
-          <p>The value of `Number.MAX_SAFE_INTEGER` is the largest integral Number n such that ℝ(n) and ℝ(n) + 1 are both exactly representable as a Number value.</p>
+          <p>Due to rounding behaviour necessitated by precision limitations of IEEE 754-2019, the Number value for every integer greater than `Number.MAX_SAFE_INTEGER` is shared with at least one other integer. Such large-magnitude integers are therefore not <emu-xref href="#safe-integer">safe</emu-xref>, and are not guaranteed to be exactly representable as Number values or even to be distinguishable from each other. For example, both `9007199254740992` and `9007199254740993` evaluate to the Number value *9007199254740992*<sub>𝔽</sub>.</p>
         </emu-note>
         <p>The value of `Number.MAX_SAFE_INTEGER` is *9007199254740991*<sub>𝔽</sub> (𝔽(2<sup>53</sup> - 1)).</p>
         <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>
@@ -30972,7 +30975,7 @@
       <emu-clause id="sec-number.min_safe_integer">
         <h1>Number.MIN_SAFE_INTEGER</h1>
         <emu-note>
-          <p>The value of `Number.MIN_SAFE_INTEGER` is the smallest integral Number n such that ℝ(n) and ℝ(n) - 1 are both exactly representable as a Number value.</p>
+          <p>Due to rounding behaviour necessitated by precision limitations of IEEE 754-2019, the Number value for every integer less than `Number.MIN_SAFE_INTEGER` is shared with at least one other integer. Such large-magnitude integers are therefore not <emu-xref href="#safe-integer">safe</emu-xref>, and are not guaranteed to be exactly representable as Number values or even to be distinguishable from each other. For example, both `-9007199254740992` and `-9007199254740993` evaluate to the Number value *-9007199254740992*<sub>𝔽</sub>.</p>
         </emu-note>
         <p>The value of `Number.MIN_SAFE_INTEGER` is *-9007199254740991*<sub>𝔽</sub> (𝔽(-(2<sup>53</sup> - 1))).</p>
         <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>


spec.html
```diff
@@ -25358,10 +25361,10 @@ <h1>Number.isSafeInteger ( _number_ )</h1>
 
       <emu-clause id="sec-number.max_safe_integer">
         <h1>Number.MAX_SAFE_INTEGER</h1>
+        <p>The value of Number.MAX_SAFE_INTEGER is 9007199254740991 (2<sup>53</sup> - 1).</p>
         <emu-note>
-          <p>The value of `Number.MAX_SAFE_INTEGER` is the largest integer n such that n and n + 1 are both exactly representable as a Number value.</p>
+          <p>Integers greater than `Number.MAX_SAFE_INTEGER` cannot by uniquely represented by Number values and are therefore not necessarily distinguishable. For example, both `9007199254740992` and `9007199254740993` identify the Number value 9007199254740992.</p>
```
zenparsing: I like the added clarity in this PR, but I'm not sure what "identify" means in this context. Would something like "are represented by" be more precise?

gibson042: I thought about that too, but it read weird to me when the same sentence referenced integers being represented by Number values and Number values being represented by ECMAScript literals. I'm open to other ideas, though, and can change it if you feel strongly.

annevk: Perhaps say they are equal/identical values?

gibson042: I changed "identify" to "evaluate to".

claudepache:
Small typo: ”cannot by uniquely represented”: s/by/be/.

gibson042:
Thanks @claudepache; fixed.

michaelficarra:
As an alternative proposal to this PR, I would suggest that a non-normative note give the intuition that `Number.MAX_SAFE_INTEGER` is the largest integral Number value that we can have confidence did not arise from rounding of its representation as an integer in source text. Of course, there are also other useful intuitions we can give about producing greater integral Number values through arithmetic means.

spec.html
```diff
@@ -25358,10 +25361,10 @@ <h1>Number.isSafeInteger ( _number_ )</h1>
 
       <emu-clause id="sec-number.max_safe_integer">
         <h1>Number.MAX_SAFE_INTEGER</h1>
+        <p>The value of Number.MAX_SAFE_INTEGER is 9007199254740991 (2<sup>53</sup> - 1).</p>
         <emu-note>
-          <p>The value of `Number.MAX_SAFE_INTEGER` is the largest integer n such that n and n + 1 are both exactly representable as a Number value.</p>
+          <p>Integers greater than `Number.MAX_SAFE_INTEGER` cannot be uniquely represented by Number values and are therefore not necessarily distinguishable. For example, both `9007199254740992` and `9007199254740993` evaluate to the Number value 9007199254740992.</p>
```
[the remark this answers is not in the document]
gibson042: I will concede that the previous explanation was more **clear**, and am open to feedback on how to recover that. But the two are equally **correct** (responding directly to your point, the integer 9007199254740992 is not uniquely represented by the Number value 9007199254740992, because the distinct integer 9007199254740993 is also represented by the same Number value—that you initially misunderstood the use of "uniquely represented" is an issue of clarity, not correctness), and the previous explanation is _less_ **useful** (because it fails to convey why anyone should care about n + 1 as opposed to e.g. n + 2, n + 10, or n − 1).

spec.html
```diff
@@ -25345,6 +25345,9 @@ <h1>Number.isNaN ( _number_ )</h1>
 
       <emu-clause id="sec-number.issafeinteger">
         <h1>Number.isSafeInteger ( _number_ )</h1>
+        <emu-note>
+          <p>An integer is considered "safe" if and only if it can be exactly represented by a Number value that uniquely distinguishes it from all other integers.</p>
```
tabatkins:
```suggestion
          <p>An integer is considered "safe" if and only if it can be exactly represented by a Number value, and no other integer is also best represented by that Number.</p>
```

spec.html
tabatkins:
```suggestion
          <p>Integers greater than `Number.MAX_SAFE_INTEGER` share Number representations with other integers, and are therefore not necessarily distinguishable. For example, both `9007199254740992` and `9007199254740993` evaluate to the Number value 9007199254740992.</p>
```

spec.html
```diff
@@ -25373,10 +25376,10 @@ <h1>Number.MAX_VALUE</h1>
 
       <emu-clause id="sec-number.min_safe_integer">
         <h1>Number.MIN_SAFE_INTEGER</h1>
+        <p>The value of Number.MIN_SAFE_INTEGER is -9007199254740991 (-(2<sup>53</sup> - 1)).</p>
         <emu-note>
-          <p>The value of `Number.MIN_SAFE_INTEGER` is the smallest integer n such that n and n - 1 are both exactly representable as a Number value.</p>
+          <p>Integers less than `Number.MIN_SAFE_INTEGER` cannot be uniquely represented by Number values and are therefore not necessarily distinguishable. For example, both `-9007199254740992` and `-9007199254740993` evaluate to the Number value -9007199254740992.</p>
```
tabatkins:
```suggestion
          <p>Integers less than `Number.MIN_SAFE_INTEGER` cannot be uniquely represented by Number values and are therefore not necessarily distinguishable. For example, both `-9007199254740992` and `-9007199254740993` evaluate to the Number value -9007199254740992.</p>
          <p>Integers less than `Number.MIN_SAFE_INTEGER` share their Number representations with other integers, and are therefore not necessarily distinguishable. For example, both `-9007199254740992` and `-9007199254740993` evaluate to the Number value -9007199254740992.</p>
```

spec.html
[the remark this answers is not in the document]
jmdyck: > the integer 9007199254740992 is not uniquely represented by the Number value 9007199254740992, because the distinct integer 9007199254740993 is also represented by the same Number value

I don't think the spec supports this phrasing. The Number value 9007199254740992 *represents* the integer 9007199254740992 and *only* that integer.

Instead, it sounds like the many-to-one relation you have in mind is the one defined for the phrase "the Number value of x". (I.e., every [non-NaN] Number value is the Number value of infinitely many mathematical values x.)

gibson042:
Thanks @tabatkins, I have incorporated your suggestions.

spec.html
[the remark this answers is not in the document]
gibson042: @jmdyck that text no longer appears in this PR.

spec.html
```diff
@@ -25345,6 +25345,9 @@ <h1>Number.isNaN ( _number_ )</h1>
 
       <emu-clause id="sec-number.issafeinteger">
         <h1>Number.isSafeInteger ( _number_ )</h1>
+        <emu-note>
+          <p>An integer is considered "<dfn id="safe-integer">safe</dfn>" if and only if it can be exactly represented by a Number value and no other integer is also best represented by that Number value.</p>
```
claudepache: Following @jmdyck’s remark:

”An integer is considered _safe_ when [its Number value](https://tc39.github.io/ecma262/#sec-ecmascript-language-types-number-type) is the Number value of no other integer.”

In order to avoid misinterpretation, we ought indeed to use language that is either self-evident or precisely defined. ”The Number value of (an exact real mathematical quantity)” is precisely defined in the spec. ”Exactly represented” is self-evident, but ”best represented” is not, because it is not clear that a Number value ”represents” more than one mathematical number (I don’t usually consider that `Number(2)` represents anything else than the integer 2.)

(Also, since it is a definition, I think that ”when” is more appropriate than ”if and only if”.)

gibson042: Excellent point; I have updated this language to use "the Number value for" as defined in [The Number Type](https://ecma-international.org/ecma-262/9.0/#sec-ecmascript-language-types-number-type) (but kept the "if and only if" for precision).

spec.html
```diff
@@ -27113,6 +27113,9 @@ <h1>Number.isNaN ( _number_ )</h1>
 
       <emu-clause id="sec-number.issafeinteger">
         <h1>Number.isSafeInteger ( _number_ )</h1>
+        <emu-note>
+          <p>A mathematical integer _n_ is considered "<dfn id="safe-integer">safe</dfn>" if and only if the Number value for _n_ is not the Number value for any other mathematical integer.</p>
```
syg, line 27117: We have a `<dfn>` for integer to mean mathematical values already, so we can drop "mathematical".

```suggestion
          <p>An integer _n_ is considered "<dfn id="safe-integer">safe</dfn>" if and only if the Number value for _n_ is not the Number value for any other integer.</p>
```

spec.html
```diff
@@ -27123,10 +27126,10 @@ <h1>Number.isSafeInteger ( _number_ )</h1>
 
       <emu-clause id="sec-number.max_safe_integer">
         <h1>Number.MAX_SAFE_INTEGER</h1>
+        <p>The value of Number.MAX_SAFE_INTEGER is 9007199254740991 (2<sup>53</sup> - 1).</p>
         <emu-note>
-          <p>The value of `Number.MAX_SAFE_INTEGER` is the largest integer n such that n and n + 1 are both exactly representable as a Number value.</p>
+          <p>Due to rounding behavior necessitated by precision limitations of IEEE 754-2008, the Number value for every mathematical integer greater than `Number.MAX_SAFE_INTEGER` is shared with at least one other mathematical integer. Such large-magnitude integers are therefore not <emu-xref href="#safe-integer">safe</emu-xref>, and are not guaranteed to be exactly representable as Number values or even to be distinguishable from each other. For example, both `9007199254740992` and `9007199254740993` evaluate to the Number value 9007199254740992.</p>
```
syg, line 27131:
```suggestion
          <p>Due to rounding behaviour necessitated by precision limitations of IEEE 754-2008, the Number value for every integer greater than `Number.MAX_SAFE_INTEGER` is shared with at least one other integer. Such large-magnitude integers are therefore not <emu-xref href="#safe-integer">safe</emu-xref>, and are not guaranteed to be exactly representable as Number values or even to be distinguishable from each other. For example, both `9007199254740992` and `9007199254740993` evaluate to the Number value 9007199254740992.</p>
```

spec.html
```diff
@@ -27138,10 +27141,10 @@ <h1>Number.MAX_VALUE</h1>
 
       <emu-clause id="sec-number.min_safe_integer">
         <h1>Number.MIN_SAFE_INTEGER</h1>
+        <p>The value of Number.MIN_SAFE_INTEGER is -9007199254740991 (-(2<sup>53</sup> - 1)).</p>
         <emu-note>
-          <p>The value of `Number.MIN_SAFE_INTEGER` is the smallest integer n such that n and n - 1 are both exactly representable as a Number value.</p>
+          <p>Due to rounding behavior necessitated by precision limitations of IEEE 754-2008, the Number value for every mathematical integer less than `Number.MIN_SAFE_INTEGER` is shared with at least one other mathematical integer. Such large-magnitude integers are therefore not <emu-xref href="#safe-integer">safe</emu-xref>, and are not guaranteed to be exactly representable as Number values or even to be distinguishable from each other. For example, both `-9007199254740992` and `-9007199254740993` evaluate to the Number value -9007199254740992.</p>
```
syg, line 27146:
```suggestion
          <p>Due to rounding behaviour necessitated by precision limitations of IEEE 754-2008, the Number value for every integer less than `Number.MIN_SAFE_INTEGER` is shared with at least one other integer. Such large-magnitude integers are therefore not <emu-xref href="#safe-integer">safe</emu-xref>, and are not guaranteed to be exactly representable as Number values or even to be distinguishable from each other. For example, both `-9007199254740992` and `-9007199254740993` evaluate to the Number value -9007199254740992.</p>
```

syg (approved):
lgtm with minor suggested edits below

gibson042:
@syg Updated!

syg (approved):
Thanks! Still lgtm.

PR #1429, opened 2019-01-24
Allow variable declaration within IfStatement conditions

jonlorusso:
In languages such as C and Perl, it is possible to declare a variable within the `Condition` of an If statement:

```
if (const paramKey = this.link.queryParamKey) {
    console.log(paramKey);
}
```

I recognize the fact that my modifications to `spec.html` are insufficient and likely incorrect.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -15793,8 +15793,8 @@
     <h2>Syntax</h2>
     <emu-grammar type="definition">
       IfStatement[Yield, Await, Return] :
-        `if` `(` Expression[+In, ?Yield, ?Await] `)` Statement[?Yield, ?Await, ?Return] `else` Statement[?Yield, ?Await, ?Return]
-        `if` `(` Expression[+In, ?Yield, ?Await] `)` Statement[?Yield, ?Await, ?Return]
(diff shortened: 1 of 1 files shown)


ljharb:
Thanks for the contribution! This seems like a new feature you'd like to add - please see https://github.com/tc39/ecma262/blob/master/CONTRIBUTING.md#creating-a-new-proposal for the best way to suggest additions to the language.

jonlorusso:
@ljharb Thanks for the reply!  [Earlier in that document](https://github.com/tc39/ecma262/blob/master/CONTRIBUTING.md#issues-and-pull-requests), it suggests creating a pull request.  I admittedly didn't make it as far as "creating a new proposal".  To be fair, my change isn't really a "small change".  Anyway, thanks for your time!

ljharb:
One possible wrinkle to keep in mind for `const` and `let` is that typically they're scoped to the curly braces they appear inside; it might be confusing if they were outside the block curlies and scoped to them, but also if they were scoped to the outer block instead of the relevant if block.

anurbol:
> it might be confusing if they were outside the block curlies and scoped to them
But there is already `for..of` and `for..in`?

PR #1431, opened 2019-01-26
Editorial: Evaluate Disjunction :: Alternative with argument direction

targos:


ljharb:
Can you elaborate a bit more on what this change accomplishes for you, and why it's needed?

targos:
Sure. I opened this a bit fast, sorry.

As part of https://github.com/devsnek/engine262, I am writing a RegExp evaluator that follows the algorithms of this spec.

In [21.2.2.4 Alternative](https://tc39.github.io/ecma262/#sec-alternative), we can see that evaluation of `Alternative` takes a parameter `direction`.
`Alternative` is evaluated from [21.2.2.3 Disjunction](https://tc39.github.io/ecma262/#sec-disjunction).
There are two cases:
- `Disjunction :: Alternative`: the one modified here, that doesn't tell to evaluate `Alternative` with the `direction` argument.
- `Disjunction :: Alternative | Disjunction`: that correctly evaluates `Alternative` with the `direction` argument.

PR #1437, opened 2019-01-31
Editorial: set default essential internal methods consistently

annevk:


annevk:
I suspect all of this is superseded by #1460. Probably best to tackle that first.

PR #1441, opened 2019-02-07
Normative: Remove implementation-defined typeof behavior

### LINKED ISSUE #1440: Implementation-defined typeof still necessary?
Now that we have the HTML document.all object defined, including its `typeof document.all = "undefined"` behavior, do we still need to permit non-standard, non-callable exotic objects to have implementation-defined typeof, as defined in https://tc39.github.io/ecma262/#sec-typeof-operator-runtime-semantics-evaluation ? I'm not aware of any use cases for this in the Web; do other embedding environments need it?

See related discussion at https://github.com/heycam/webidl/issues/512 @annevk @TimothyGu

littledan:
Previously, the ECMAScript specification permitted non-callable
non-standard exotic objects to have implementation-defined typeof,
within certain limits. Although some ECMAScript implementations
may have taken advantage of this possibility historically, it's not
clear that anyone needs it anymore. Instead, other mechanisms can
be used by embedders to indicate aspects of objects.

Closes #1440

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13051,15 +13051,7 @@
             </tr>
             <tr>
               <td>
-                Object (ordinary and does not implement [[Call]])
-              </td>
-              <td>
-                `"object"`
-              </td>
-            </tr>
-            <tr>
-              <td>
-                Object (standard exotic and does not implement [[Call]])
+                Object (does not implement [[Call]])
               </td>
               <td>
                 `"object"`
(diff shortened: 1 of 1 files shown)


ljharb:
Generally, every normative change requires committee approval; it certainly *seems* like something we could decide on GitHub, but I would be uncomfortable merging a non-bugfix normative change without that explicit consensus.

allenwb:
> Instead, @@toStringTag and other
mechanisms can be used by embedders to indicate aspects of objects

Just wanted to remind that @@toStringTag, unlike `typeof`, is forgeable  and was never intended to be used as a reliable type checking mechanism. We shouldn't encourage people to try to use it that way. 

Otherwise, I like the change this issue is proposing.

littledan:
Removed the reference to `@@toStringTag` as @allenwb suggested and the note that @annevk pointed out. Thanks for the quick reviews everyone!

To the editor group: Do you plan to put this on the agenda for the upcoming TC39 meeting, or should I do it?

ljharb:
It’s always preferred when committee members can drive their own needs-consensus changes.

littledan:
Alright, thanks for the clarification and quick review.

jmdyck:
> It’s always preferred when committee members can drive their own needs-consensus changes.

So what should happen when a needs-consensus change doesn't 'belong' to a committee member?

ljharb:
@jmdyck in that case (or when no committee member is driving it) it's up to the editors to proactively add it to the agenda, as time permits.

jmdyck:
Cool, thanks.

ljharb:
The HTML change for this seems to have landed https://github.com/heycam/webidl/pull/640

annevk:
To be clear, that text can be removed if this lands. It's mostly added there since this change is taking a while.

PR #1442, opened 2019-02-08
Editorial: Simplify tables removing end tags (tr, th, and td) and tbody

sirisian:
Per this discussion: https://github.com/bterlson/ecmarkup/issues/152#issuecomment-460369361

These "minor" modifications remove the ```</tr>```, ```</th>```, ```</td>```, ```<thead>```, ```</thead>```, ```<tbody>```, and ```</tbody>``` tags from the spec.html. The changes also move up lines to the ```<th>``` and ```<td>``` tag when the lines aren't nested tags. (When they were I kept them how they were in the original document).

Motivation for this change are that in my proposal I'm making table changes and adding rows and it was getting very large and easy to make mistakes. These changes make it much easier to follow.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -924,10 +924,12 @@
         <p>Within this specification a well-known symbol is referred to by using a notation of the form @@name, where &ldquo;name&rdquo; is one of the values listed in <emu-xref href="#table-1"></emu-xref>.</p>
         <emu-table id="table-1" caption="Well-known Symbols">
           <table>
-            <tr>
-              <th>Specification Name
-              <th>[[Description]]
-              <th>Value and Purpose
+            <thead>
+              <tr>
+                <th>Specification Name
+                <th>[[Description]]
+                <th>Value and Purpose
+            </thead>
             <tr>
               <td>@@asyncIterator
               <td>`"Symbol.asyncIterator"`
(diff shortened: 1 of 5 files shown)


mathiasbynens:
Did you forget the `table-*.html` files? e.g. https://github.com/tc39/ecma262/blob/master/table-binary-unicode-properties.html

domenic:
Probably you should not remove the opening `<thead>` and `<tbody>`, since those convey important semantic information.

sirisian:
@mathiasbynens Whoops, and I was thinking about those the other day (wondering if my proposal changed any). I'll go back and update those also.

@domenic Only a few of the tables had the thead? Were those tables special? I'll go back and check. Also none of the tables use more than one tbody? I didn't see the reason for it. The only styling referring to thead is:

```css
emu-table th, emu-table thead td, table.real-table th {
    background-color: #eeeeee;
}
```

I ran a document.querySelectorAll('emu-table thead td') and it returned no elements so I assumed it was just vestigial css. (Or tables looked differently a long time ago was my thought, but I didn't check). Those tbody lines add something like 140 extra lines to the document.

tabatkins:
`thead` is useful semantic info, yes. (And should be used more often, I see in the diff a number of tables with a header row that is not correctly marked as such.)  Screenreaders and such use heuristics to try and detect header rows that aren't marked up (like a first row containing only `th`s), but it's still worthwhile to transmit the semantic explicitly, for styling if nothing else.  (For example, the W3C stylesheet uses `thead` to put a thick border between the head and body of the table.)

And then, if you're keeping `thead`, you need to *either* keep the `</thead>` or the `<tbody>` tag; I recommend the former, so the main table can lose a level of indentation.

sirisian:
@tabatkins Done. The spec is now ever so slightly easier to read for the blind.

I ran it through ecmarkup also just to be sure and it all looks the same rendered. It's a lot of changes. (Sorry reviewer).

michaelficarra:
@sirisian Can you separate out the `<thead>` addition into a separate PR? Are there any other accessibility-related changes here?

sirisian:
I forgot this existed. I just tried to do a pull and there's too many conflicts for this. I don't have time at the moment to fix it. Weird this wasn't merged 10 months ago. Feel free to make the changes. They were just aesthetic.

PR #1463, opened 2019-03-01
Editorial: Fix description of [[ThisMode]] internal slot of Function Object

### LINKED ISSUE #1305: Fixing description of [[thisMode]] internal slot of Function Object.
in current, specification explain `[[thisMode]]` of Function Object as blow.

> Defines how this references are interpreted within the formal parameters
and code body of the function. lexical means that this refers to the this
value of a lexically enclosing function. strict means that the this value is used
exactly as provided by an invocation of the function. global means that a this
value of undefined is interpreted as a reference to the global object.

and I think, this line missed some action about how **global** this mode works.

>  global means that a this
value of undefined is interpreted as a reference to the global object.

because, In `OrdinaryCallBindThis` abstract operation that used to bind `this` value into function environment record. It interpret `null` as reference to global object like `undefined`.

So, In my opinion, changing that line to blow will make more clear specification.

>  global means that a this
value of undefined or null is interpreted as a reference to the global object.

viktmv:
Fixes #1305

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -8341,7 +8341,7 @@
             ~lexical~ | ~strict~ | ~global~
           </td>
           <td>
-            Defines how `this` references are interpreted within the formal parameters and code body of the function. ~lexical~ means that `this` refers to the *this* value of a lexically enclosing function. ~strict~ means that the *this* value is used exactly as provided by an invocation of the function. ~global~ means that a *this* value of *undefined* is interpreted as a reference to the global object.
+            Defines how `this` references are interpreted within the formal parameters and code body of the function. ~lexical~ means that `this` refers to the *this* value of a lexically enclosing function. ~strict~ means that the *this* value is used exactly as provided by an invocation of the function. ~global~ means that a *this* value of *undefined* or *null* is interpreted as a reference to the global object, and any other *this* value is first passed to ToObject.
           </td>
         </tr>
         <tr>


zenparsing:
I actually prefer the previous commit. This is just noting the possible values of `[[ThisMode]]`. For details about the actual behavior readers should consult the algorithm steps.

viktmv:
Adding the description was suggested in the issue, might have jumped the gun with that one 🙂

Should I revert the commit untill the decision is made, or what's the course of action here? Thanks!

spec.html
```diff
@@ -7236,7 +7236,7 @@ <h1>ECMAScript Function Objects</h1>
             (lexical, strict, global)
           </td>
           <td>
-            Defines how `this` references are interpreted within the formal parameters and code body of the function. ~lexical~ means that `this` refers to the *this* value of a lexically enclosing function. ~strict~ means that the *this* value is used exactly as provided by an invocation of the function. ~global~ means that a *this* value of *undefined* is interpreted as a reference to the global object.
+            Defines how `this` references are interpreted within the formal parameters and code body of the function. ~lexical~ means that `this` refers to the *this* value of a lexically enclosing function. ~strict~ means that the *this* value is used exactly as provided by an invocation of the function. ~global~ means that a *this* value of *undefined* or *null* is interpreted as a reference to the global object, and *this* value of primitive type is wrapped by its type's wrapper object.
```
ljharb: "the global object" is incorrect anyways - it's "the [[GlobalThisValue]] of the [[GlobalEnv]] value of the current realm", which probably wouldn't help make this much clearer :-)

If `~global~` also implies sloppy mode, then I'd maybe state this like:
```suggestion
            Defines how `this` references are interpreted within the formal parameters and code body of the function. ~lexical~ means that `this` refers to the *this* value of a lexically enclosing function. ~strict~ means that the *this* value is used exactly as provided by an invocation of the function. ~global~ means that a *this* value of *undefined* or *null* is interpreted as a reference to the global object, and the *this* value is first passed to ToObject.
```

ENvironmentSet: Humm...
I agree that the expression `the global object` is incorrect. but I think `the global object of current realm` or `the [[GlobalObject]] of the current realm` would be shorter and more readable then `the [[GlobalThisValue]] of the [[GlobalEnv]] value of the current realm`. do you have any reason to use that sentence?

PS. `currentRealm.[[GlobalEnv]].[[GlobalThisValue]]` is always eqauls `currentRealm.[[GlobalObject]]`. (8.2.3 SetRealmGlobalObject is only way to change "**the global object of current realm**" as much as I know.)

anyway, 
>  the *this* value is first passed to ToObject.
this looks better then my suggestion, I'm totally support this sentence.

jmdyck: > currentRealm.[[GlobalEnv]].[[GlobalThisValue]] is always eqauls currentRealm.[[GlobalObject]]

No, not always. It's either that *or* an implementation-defined object. See [InitializeHostDefinedRealm](https://tc39.github.io/ecma262/#sec-initializehostdefinedrealm), step 8.

ENvironmentSet: Did you checked next step of it? it calls SetRealmGlobalObject(and this abstract operation makes sure what I just said before.)

anyway, If you meant that `currentRealm.[[GlobalObject]]` could not be an "global object"(which spec explains at Clause 18), then I'd love to say that I missed that part.

and.. If I missed something, I'd love to say sorry to you :(

ljharb: @ENvironmentSet all browsers supply a different globalThis value from the globalObject value (a window proxy).

ENvironmentSet: @ljharb Oh, I see. thank you for comment :)
@jmdyck thank you too :))

bakkot:
@ljharb Github's UI is broken, so I manually pushed a commit with your change.

Re: "global object", I agree it's wrong, but I don't have a particular fix in mind for it (@ljharb, it sounded like you didn't think your suggestion was the right approach), so unless someone has a better idea for that right now let's land this as-is and follow up on fixing that description.

PR #1464, opened 2019-03-02
Editorial: correctly determine flags for static RegExp parsing

### LINKED ISSUE #1343: Editorial: statically enforcing regex grammar for literals
The grammar for regular expressions is pretty weird, in that there is a [very liberal grammar](https://tc39.github.io/ecma262/#sec-literals-regular-expression-literals) which is later [refined](https://tc39.github.io/ecma262/#sec-patterns). The intent of this, at least as I understand, test262 tests, and engines implement, is that the more conservative grammar is enforced statically.

As far as I can tell, that's done in [12.2.8.1](https://tc39.github.io/ecma262/#sec-primary-expression-regular-expression-literals-static-semantics-early-errors):

> It is a Syntax Error if BodyText of RegularExpressionLiteral cannot be recognized using the goal symbol Pattern of the ECMAScript RegExp grammar specified in [21.2.1](https://tc39.github.io/ecma262/#sec-patterns).

But this has a couple of problems: in particular, "the goal symbol Pattern" does not refer to a particular thing, because the Pattern goal symbol has parameters. And nothing in this rule refers to the regex's flags, which means that I don't think it would forbid `/\a/u` statically, though all engines ([except V8](https://bugs.chromium.org/p/v8/issues/detail?id=896)) reject it.

This is even more complicated because of the behavior of the named groups `\k`, which is responsible for the `N` flag on the Pattern grammar. The [runtime semantics](https://tc39.github.io/ecma262/#sec-regexpinitialize) which make for example `/\k<>(?<a>)/` an error live in RegExpInitialize, and are pretty complicated:

> Parse P using the grammars in [21.2.1](https://tc39.github.io/ecma262/#sec-patterns) and interpreting each of its 16-bit elements as a Unicode BMP code point. UTF-16 decoding is not applied to the elements. The goal symbol for the parse is Pattern[~U, ~N]. If the result of parsing contains a GroupName, reparse with the goal symbol Pattern[~U, +N] and use this result instead. Throw a SyntaxError exception if P did not conform to the grammar, if any elements of P were not matched by the parse, or if any Early Error conditions exist.

I think the rules in the runtime semantics (specifically, steps 6, 7/a, and 8/a) need to be essentially copied to the static semantics. That solves all three problems above:

- The ambiguous reference to "the goal symbol Pattern" is removed, because the flags are always explicitly specified.
- The unicode flag is properly taken into consideration.
- The reparse-if-named-group behavior is enforced.

On IRC, @jmdyck points out that the problematic static semantics text predates the addition of parameters to the Pattern grammar, so this is probably just an editorial oversight as the grammar became more complicated.

See also #1050.

bakkot:
Fixes #1343 according to the approach [discussed therein](https://github.com/tc39/ecma262/issues/1343#issuecomment-458646159); see that issue for full context, but briefly, the current spec text says to do a parse according to "the goal symbol Pattern", which is meaningless in the current spec because Pattern has flags, which must be provided. Determining these flags is a little involved.

As it says there, this is heavily duplicated with [the runtime semantics for RegExpInitialize](https://tc39.github.io/ecma262/#sec-regexpinitialize). If the editors think this is worth noting, or avoiding, I can try to come up with something.

I'd particularly appreciate reviews from @jmdyck and @waldemarhorwat.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12925,12 +12925,23 @@
         <emu-grammar>PrimaryExpression : RegularExpressionLiteral</emu-grammar>
         <ul>
           <li>
-            It is a Syntax Error if BodyText of |RegularExpressionLiteral| cannot be recognized using the goal symbol |Pattern| of the ECMAScript RegExp grammar specified in <emu-xref href="#sec-patterns"></emu-xref>.
-          </li>
-          <li>
-            It is a Syntax Error if FlagText of |RegularExpressionLiteral| contains any code points other than `g`, `i`, `m`, `s`, `u`, or `y`, or if it contains the same code point more than once.
+            It is a Syntax Error if IsValidRegularExpressionLiteral(|RegularExpressionLiteral|) is *false*.
           </li>
         </ul>
+      </emu-clause>
+
+      <emu-clause id="sec-isvalidregularexpressionliteral" aoid="IsValidRegularExpressionLiteral">
+        <h1>Static Semantics: IsValidRegularExpressionLiteral ( _literal_ )</h1>
+        <p>The abstract operation IsValidRegularExpressionLiteral determines if its argument is a valid regular expression literal. The following steps are taken:</p>
+        <emu-alg>
+          1. Assert: _literal_ is a |RegularExpressionLiteral|.
+          1. If FlagText of _literal_ contains any code points other than `g`, `i`, `m`, `s`, `u`, or `y`, or if it contains the same code point more than once, return *false*.
+          1. Let _P_ be BodyText of _literal_.
+          1. If FlagText of _literal_ contains `u`, then
+            1. Parse _P_ using the grammars in <emu-xref href="#sec-patterns"></emu-xref>. The goal symbol for the parse is |Pattern[+U, +N]|. If _P_ did not conform to the grammar, if any elements of _P_ were not matched by the parse, or if any Early Error conditions exist, return *false*.
+            1. Otherwise, return *true*.
+          1. Parse _P_ using the grammars in <emu-xref href="#sec-patterns"></emu-xref>. The goal symbol for the parse is |Pattern[~U, ~N]|. If the result of parsing contains a |GroupName|, reparse with the goal symbol |Pattern[~U, +N]|. If _P_ did not conform to the grammar, if any elements of _P_ were not matched by the parse, or if any Early Error conditions exist, return *false*. Otherwise, return *true*.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-regular-expression-literals-runtime-semantics-evaluation">


spec.html
```diff
@@ -11828,14 +11828,25 @@ <h1>Static Semantics: Early Errors</h1>
...
+        <emu-alg>
+          1. Assert: _literal_ is a Parse Node corresponding to a RegularExpressionLiteral.
+          1. If FlagText of _literal_ contains any code points other than `"g"`, `"i"`, `"m"`, `"s"`, `"u"`, or `"y"`, or if it contains the same code point more than once, return *false*.
+          1. Let _P_ be BodyText of _literal_.
+          1. If FlagText of _literal_ contains `"u"`,
+            1. If _P_ cannot be recognized using the goal symbol |Pattern[+U, +N]| of the ECMAScript RegExp grammar specified in <emu-xref href="#sec-patterns"></emu-xref>, return *false*.
+            1. Otherwise, return *true*.
+          1. Parse _P_ using the grammars in <emu-xref href="#sec-patterns"></emu-xref>. The goal symbol for the parse is |Pattern[~U, ~N]|. If the result of parsing contains a |GroupName|, reparse with the goal symbol |Pattern[~U, +N]| and use this result instead. If _P_ did not conform to the grammar, if any elements of _P_ were not matched by the parse, or if any Early Error conditions exist, return *false*. Otherwise, return *true*.
```
zenparsing: This is a mouthful, and I can see it's basically duplicated over [here](https://tc39.github.io/ecma262/#sec-regexpinitialize). An improvement in the future might be to attempt some common factoring (but it might also be awkard due to runtime/static differences).

jmdyck: I've found a way to factor out the common bits. I can submit that as a follow-up.

bakkot:
@jmdyck Addressed those two; thanks!

spec.html
```diff
@@ -11828,14 +11828,25 @@ <h1>Static Semantics: Early Errors</h1>
...
+            It is a Syntax Error if IsValidRegularExpressionLiteral(|RegularExpressionLiteral|) is *false*.
           </li>
         </ul>
       </emu-clause>
 
+      <emu-clause id="sec-primary-expression-regular-expression-literals-static-semantics-isvalidregularexpressionliteral" aoid="IsValidRegularExpressionLiteral">
+        <h1>Static Semantics: IsValidRegularExpressionLiteral ( _literal_ )</h1>
+        <p>The abstract operation IsValidRegularExpressionLiteral determines if its argument is a valid regular expression literal. The following steps are taken:</p>
+        <emu-alg>
+          1. Assert: _literal_ is |RegularExpressionLiteral|.
```
```suggestion
          1. Assert: _literal_ is a |RegularExpressionLiteral|.
```

bakkot: Gah, thanks.

spec.html
```diff
@@ -11828,14 +11828,25 @@ <h1>Static Semantics: Early Errors</h1>
...
 
+      <emu-clause id="sec-primary-expression-regular-expression-literals-static-semantics-isvalidregularexpressionliteral" aoid="IsValidRegularExpressionLiteral">
+        <h1>Static Semantics: IsValidRegularExpressionLiteral ( _literal_ )</h1>
+        <p>The abstract operation IsValidRegularExpressionLiteral determines if its argument is a valid regular expression literal. The following steps are taken:</p>
+        <emu-alg>
+          1. Assert: _literal_ is a |RegularExpressionLiteral|.
+          1. If FlagText of _literal_ contains any code points other than `"g"`, `"i"`, `"m"`, `"s"`, `"u"`, or `"y"`, or if it contains the same code point more than once, return *false*.
+          1. Let _P_ be BodyText of _literal_.
+          1. If FlagText of _literal_ contains `"u"`,
+            1. If _P_ cannot be recognized using the goal symbol |Pattern[+U, +N]| of the ECMAScript RegExp grammar specified in <emu-xref href="#sec-patterns"></emu-xref>, return *false*.
```
jmdyck: This should probably include the stuff about elements being left over or Early Errors.
```suggestion
            1. Parse _P_ using the grammars in <emu-xref href="#sec-patterns"></emu-xref>. The goal symbol for the parse is |Pattern[+U, +N]|. If _P_ did not conform to the grammar, if any elements of _P_ were not matched by the parse, or if any Early Error conditions exist, return *false*.
```

jmdyck:
Once this is merges, I should be able to submit the follow-up that I mentioned in March.

bakkot:
Addressed comments and merge conflicts.

spec.html
bakkot: @jmdyck If you still remember what your plan for factoring out the common bits is, this PR is now merged!

jmdyck: I do, but it'll be easier (fewer collisions) if I wait until after PR #1552 is merged.

bakkot: @jmdyck #1552 landed; do want to submit that follow-up?

jmdyck: Done: PR #1866.

PR #1471, opened 2019-03-07
Editorial: Grammar "so has the features..." -> "so have the features..."

Tomy8s:
Small grammar mistake:

If we consider we are discussing "the features and facities ECMAScript provides",

It would be grammatically incorrect to write:

```As the usage of ECMAScript has expanded, so has they.```

And also grammatically incorrect to write:

```As the usage of ECMAScript has expanded, so has the features and facilities it provides.```

---

It would be grammatically correct to write:

```As the usage of ECMAScript has expanded, so have they.```

And therefore would be grammatically correct for `spec.html` to read:

```As the usage of ECMAScript has expanded, so have the features and facilities it provides.```

### CODE REVIEW

**Comment by rkirsling** [comment]
Nice catch!

I suppose it's easy to overlook because one sort of wants it to be balanced as `has`/`has`. 🤔
Toward that end, we could alternatively make the subordinate subject singular:
> As the usage of ECMAScript has expanded, so **has the breadth of** the features and facilities it provides.

File: spec.html [hunk 1]
```diff
@@ -138,7 +138,7 @@ <h1>Overview</h1>
...
   <p>ECMAScript was originally designed to be a <em>Web scripting language</em>, providing a mechanism to enliven Web pages in browsers and to perform server computation as part of a Web-based client-server architecture. ECMAScript is now used to provide core scripting capabilities for a variety of host environments. Therefore the core language is specified in this document apart from any particular host environment.</p>
-  <p>ECMAScript usage has moved beyond simple scripting and it is now used for the full spectrum of programming tasks in many different environments and scales. As the usage of ECMAScript has expanded, so has the features and facilities it provides. ECMAScript is now a fully featured general-purpose programming language.</p>
+  <p>ECMAScript usage has moved beyond simple scripting and it is now used for the full spectrum of programming tasks in many different environments and scales. As the usage of ECMAScript has expanded, so have the features and facilities it provides. ECMAScript is now a fully featured general-purpose programming language.</p>
```
```suggestion
  <p>ECMAScript usage has moved beyond simple scripting and it is now used for the full spectrum of programming tasks in many different environments and at many scales. As the usage of ECMAScript has expanded, so have the features and facilities it provides. ECMAScript is now a fully featured general-purpose programming language.</p>
```

PR #1479, opened 2019-03-13
Normative: CreateDynamicFunction early concatenates bodyText

leobalter:
Ref tc39/test262#2102
Ref tc39/test262#2109

This normative change matches the current behavior of V8, SpiderMonkey, and JSC (e.g. `Function("-->")`. This is not only for the convenience of reflecting the observed behaviors, ~~but this will allow some guard for potential false positives like `Function("#!hashbang")`~~.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -25776,7 +25776,7 @@
                 1. Set _P_ to the string-concatenation of the previous value of _P_, *","* (a comma), and _nextArgString_.
                 1. Set _k_ to _k_ + 1.
               1. Let _bodyText_ be _args_[_k_].
-            1. Set _bodyText_ to ? ToString(_bodyText_).
+            1. Set _bodyText_ to be the string-concatenation of 0x000A (LINE FEED), ? ToString(_bodyText_), and 0x000A (LINE FEED).
             1. Perform the following substeps in an implementation-dependent order, possibly interleaving parsing and error detection:
               1. Let _parameters_ be the result of parsing _P_, interpreted as UTF-16 encoded Unicode text as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, using _parameterGoal_ as the goal symbol. Throw a *SyntaxError* exception if the parse fails.
               1. Let _body_ be the result of parsing _bodyText_, interpreted as UTF-16 encoded Unicode text as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, using _goal_ as the goal symbol. Throw a *SyntaxError* exception if the parse fails.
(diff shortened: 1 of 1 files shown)


anba:
> but this will allow some guard for potential false positives like `Function("#!hashbang")`.

I don't understand this part, hashbang is only allowed when parsing with the goal symbol `Script` or `Module`, so it never applies in `CreateDynamicFunction`.

leobalter:
@anba right you are. Maybe the only advantage is the convenience for reflecting engines behavior in a very specific usage.

leobalter:
@ljharb the current tests we have on Test262 matches this proposed behavior. If tc39/test262#2101 tc39/test262#2102 get merged those will invert. 

If we have those PRs on hold until the TC39 meeting the tag for need tests could be discarded.

----

Just to inform, I just want a closing answer for this problem. One way or another, I'm fine with any decision TC39 chooses for this PR.

leobalter:
tc39/test262#2101 is now merged, that means I need a new PR to now reflect this change.

leobalter:
Doing more investigation over this issue I found some similar cases:

```html
<script>-->
console.log('foo');</script>
```

Chrome, Firefox, Safari: will log `'foo'`.
Edge: SyntaxError
```

Anyway a file starting with `-->` will have the observed SyntaxError properly on v8, ChakraCore, SpiderMonkey, JSC, and Node.

The same issue is also observed with `eval`:

```
$ eshost -se 'eval("-->")'
#### d8, jsc, jsshell, node
undefined

#### ch
SyntaxError: Syntax error
```

----

The error is similar to what happens to `Function("-->")`, but I'm not convinced the fix should happen at the same place for whatever happens to `Function` (CreateDynamicFunction), `eval` (PerformEval), or the `script` tag (not ecma262).

**Comment by leobalter** [comment]
There is an interesting observation for the actual output of the Function ctor:

```
eshost -se 'Function(42).toString()'
#### ch
function anonymous(
) {42
}

#### d8, jsshell, node
function anonymous(
) {
42
}

#### jsc
function anonymous() {
42
}
```

That means that ChakraCore is also not compliant to CreateDynamicFunction when the SourceText is not correctly set, it misses the preceding line feed. I wonder if fixing this part would have the same issue observed for HTMLCloseComments.

A function like this will be a SyntaxError anywhere:

```js
function fn() {-->
}
```

but this function formatting is not:

```js
function fn() {
--> valid code
}
```

And this matches the revealed code from `Function("-->")`:

```
$ eshost -se 'Function("-->").toString()'
#### d8, jsshell, node
function anonymous(
) {
-->
}

#### ch
SyntaxError: Syntax error

#### jsc
function anonymous() {
-->
}
```

---

As seen in the examples, it makes sense to observe the function body in conjunction with the padding of line feeds added at CreateDynamicFunction.

I'm ok to observe the string: `\n-->\n` can be correctly parsed.

leobalter:
My conclusion is to stick with the proposed change and discuss the problems with eval elsewhere. These are another problem I don't have time and availability to have immediately fixed. 

Fixing the `Function` ctor in the specs will not preclude anything else to be fixed.

ljharb:
Discussed on the editor call; this is ready to go.

PR #1493, opened 2019-03-28
Normative: Extend specs to observe WebReality for HTMLCloseComment

leobalter:
Ref #1479

Observable productions from the current WebReality:

```js
eval('-->');
```

```html
<script>--></script>
```

What this does not do/allow:

```js
function fn() {-->
} // THIS REMAINS AS A SYNTAX ERROR WITH OR WITHOUT THE PR
```


spec.html
```diff
@@ -40074,6 +40074,7 @@ <h2>Syntax</h2>
 
         SingleLineHTMLCloseComment ::
           LineTerminatorSequence HTMLCloseComment
+          [no |SourceCharacter| here] HTMLCloseComment
```
mathiasbynens: As discussed offline, there’s no precedent for something like this in the spec. We’d need to find a better way to express this grammar. @waldemarhorwat can probably help.

bakkot: If I understand correctly the intent is to allow these comments in exactly the place that the [hashbang proposal](https://github.com/tc39/proposal-hashbang) allows `#!`. If that's so, I think the easiest thing to do is to extend the [`HashbangComment` production](https://tc39.github.io/proposal-hashbang/out.html#prod-HashbangComment) it adds so that it also matches `--> SingleLineCommentChars_opt` in the `Script` goal (and then possibly rename the production to `InitialComment` or something).

leobalter: @bakkot this has been considered but it's not really possible.

First of all, the hashbang proposal has a grammar and an exception (only allowed at the start of Script or Module). For the current proposal, we don't have a grammar without the preceding requirements. I could possibly have a simple `HTMLCloseComment` production and say it's only possible at the very start. The problem here is that I can't limit this to the first sequence of "the source text of an ECMAScript Script or Module" only, we'd need to go being Script and Module goals so we make it a fit for `eval` and cases like `Function`.

Another restriction goes to only specifying `-->` directly. We need to have it extended to all the productions from `HTMLCloseComment` including the `/* |line terminator| */ -->`

The intent is to have the specs for HTMLCloseComment matching web reality. So that should extend to the first sequence of source texts, `eval('-->')`, `Function('-->')`, etc, including the productions with the multi line comments.

I'd be happy with other suggestions for the grammar.

leobalter: > I don't understand why you'd need to make any change to the HTMLCloseComment production.

My goal is to reflect web reality

> Are you saying it would be insufficient to limit it to the start of Script and Module?

yes, this should include parts where SourceText is parsed without these goals.

> For "being", do you mean "beyond"?

yes, typo.

> For eval, the goal symbol is Script, so that's not a problem. But for Function, the body-text goal symbol is FunctionBody or GeneratorBody, etc, so that's the problem?

If that unblocks the suggestion, I'd like to know what's the suggested production. 

So far I'd like to unblock a grammar that reflects the current productions:

```js
-->
```

```js
/**/ -->
```

```js
/**/  /* *//*foo*/ -->
```

```js
/*
*/ /**/ -->
```

All of them being valid productions as the first sequence of any sourcetext:

- file (script src)
- script tag code (no src)
- eval
- Function and derivatives (AsyncFunction, GeneratorFunction, etc)
- possibly non-ecmascript parsing like `setTimeout('-->\nconsole.log(42)', 10);`

jmdyck: [Sorry, github wasn't showing me that you'd replied, so I modified my earlier answer. I've reverted it to what you replied to.]

jmdyck: > >  For eval, the goal symbol is Script, so that's not a problem. But for Function, the body-text goal symbol is FunctionBody or GeneratorBody, etc, so that's the problem?
> 
> If that unblocks the suggestion, I'd like to know what's the suggested production.

I don't understand that sentence. (But it sounds like, yes, parsing source-text with goals other than Script and Module is the problem.)

> So far I'd like to unblock a grammar that reflects the current productions:

Again, I don't understand what you mean by "unblock". I think you're saying that you want to change the grammar to allow each of the 4 source-texts in each of the 5 contexts. Is that correct?

gibson042: I suggest employing a strategy similar to the one used in [_NumericLiteral_](https://ecma-international.org/ecma-262/9.0/#sec-literals-numeric-literals) ("the _SourceCharacter_ immediately following a _NumericLiteral_ must not be an _IdentifierStart_ or _DecimalDigit_"), e.g.
<blockquote>

_Comment_::
&nbsp; _MultiLineComment_
&nbsp; _SingleLineComment_
&nbsp; _SingleLineHTMLOpenComment_
&nbsp; _SingleLineHTMLCloseComment_
&nbsp; **_InitialSingleLineHTMLCloseComment_**
&nbsp; _SingleLineDelimitedComment_

…

**_InitialSingleLineHTMLCloseComment_::**
&nbsp; **_HTMLCloseComment_**

**_InitialSingleLineHTMLCloseComment_ may not be preceded by any _SourceCharacter_.**
</blockquote>


Or if the intent is to apply this special treatment only if **no** other characters precede "-->" such that e.g. `/**/-->` is not recognized as _HTMLCloseComment_—which seems not to be the case per @leobalter's last message—then we would want e.g.
<blockquote>

**_InitialSingleLineHTMLCloseComment_::**
&nbsp; **`-->` SingleLineCommentChars<sub>opt</sub>**
</blockquote>

leobalter: @jmdyck:

> Again, I don't understand what you mean by "unblock". I think you're saying that you want to change the grammar to allow each of the 4 source-texts in each of the 5 contexts. Is that correct?

yes

leobalter: @gibson042 I like the solution from the example.

`/**/-->` with a preceding line terminator is a valid HTMLCloseComment, and also accepted as the first sequence of a source text. My proposed text - and yours as in the first example - matches that.

jmdyck: The 4th source-text:

    /*
    */ /**/ -->

is a `MultiLineComment` and thus a `Comment`, and so is already allowed in any context.

Each of the other source-texts is an initial `HTMLCloseComment` (i.e., not preceded by any other characters), which is not currently allowed, so those are the ones that require a grammar change.

@bakkot's suggestion would allow them at the start of `Script` and `Module`, which I think covers all contexts except the 4th: "Function and derivatives (AsyncFunction, GeneratorFunction, etc)".

To handle the latter, we could say that a bare `HTMLCloseComment` can also occur at the start of a `FunctionBody` (etc), but that would also allow `function fn() {-->\n}`, which we want to remain a syntax error.

Is that a fair summary of your objection to @bakkot's suggestion?

leobalter: > suggestion would allow them at the start of Script and Module (...)

In hashbang we have a grammar production with an extra text setting the restriction. How we would set this for HTMLCloseComment?

I'm not objecting, what I am saying we need to end up with a new production anyway. My recommendation is to use the current proposed in this PR and maybe add an additional text describing it, like @gibson042's suggestion:

> _HTMLCloseComment may not be preceded by any SourceCharacter._

or maybe saying

> [no |SourceCharacter| here] means that the following grammar is only valid if parsed as the first sequence of a SourceText.

For these suggestions, describing the goals is not relevant.

I'm welcoming to other alternatives and I'm only asking if anyone can prototype/draft them as spec text.

not-an-aardvark: As far as compatibility is concerned, how important is it to preserve the web reality here? It seems unlikely that many websites are relying on the `eval` and `Function` behavior of `-->`. (I presume the `-->`-as-a-comment behavior was originally introduced to support inline scripts.) It seems like this could be a case where web reality can be simplified at the same time as the spec is updated to match it.

waldemarhorwat: My recommendation would be similar to @gibson042's, but I'd do it one level down:

_SingleLineHTMLCloseComment_ ::
&nbsp;   _LineTerminatorSequence HTMLCloseComment_
&nbsp;   _HTMLCloseComment_  (only if no prior input elements have been parsed from the source text being parsed)

Here the intent is that the second expansion of _SingleLineHTMLCloseComment_ can only be done if the _SingleLineHTMLCloseComment_ would be the very first input element produced in the parse of some source text. An illustrative example would be useful to get the idea across.

You'll want to check whether this is actually universally true in all contexts, or, if not, whether we want this to be universally true in all contexts. I think it is true in all of the contexts @leobalter mentioned above. I'm not sure about the context in which the Function parameters are parsed (and that one seems to have a bug in at least implementation encountering an error if function parameters include `//`).

jmdyck: > > @bakkot's suggestion would allow them at the start of Script and Module (...)
> 
> In hashbang we have a grammar production with an extra text setting the restriction.

More precisely, the hashbang proposal adds another InputElement* nonterminal+production and (like all other such nonterminals) specifies when it should be used as the goal symbol for the lexical parser.

> How we would set this for HTMLCloseComment?

We could add `HTMLCloseComment` as another RHS of `InputElementHashbangOrRegExp`. (This is equivalent to @bakkot's suggestion, I think.)

> I'm not objecting, what I am saying we need to end up with a new production anyway.

I'm not sure what you mean by that, since this PR doesn't create a new production. (It just adds a RHS to the `SingleLineHTMLCloseComment` production.)

leobalter: > You'll want to check whether this is actually universally true in all contexts, or, if not, whether we want this to be universally true in all contexts. I think it is true in all of the contexts @leobalter mentioned above

Agreed, and I'll change this PR to match the suggestion from @waldemarhorwat, which I believe it's a good fit for the proposed change, considering where we place the production and how we restrict the production as the first part of the source text.

@jmdyck I'd rather not mix it with the hashbang, these will leave some parts open with unnecessary complexity. The suggestion I first proposed, along with @gibson042's and @waldemarhorwat are, IMO, effective and without much complexity. It's something I'd like to see fixed and never have to touch anymore unless we eventually plan to remove it.

jmdyck: > @jmdyck I'd rather not mix it with the hashbang, these will leave some parts open with unnecessary complexity.

Okay, but that's a very imprecise statement. (Which parts of what open to what?) I've been trying to nail down your objection/qualms/concerns with the mixing-with-hashbang suggestion, and [I thought I'd figured it out](https://github.com/tc39/ecma262/pull/1493#discussion_r270982122), but your responses since then have left me puzzled and uncertain. Please confirm if I've correctly characterized the problem with @bakkot's suggestion.

leobalter:
Closing this in favor of #3244

PR #1498, opened 2019-04-03
Provide HostEnsureCanCompileString implementors more context

### LINKED ISSUE #938: Include the string to be compiled in the call to `HostEnsureCanCompileStrings`
To improve the quality of CSP reports, it would be helpful for [`HostEnsureCanCompileStrings()`](https://tc39.github.io/ecma262/#sec-hostensurecancompilestrings) to include the string to be compiled as an argument. `HostEnsureCanCompileStrings(callerRealm, calleeRealm, source)` seems ideal. :)

The goal is to ensure that we can include a sample of the script which violates the policy when generating a CSP violation report. We're doing this for inline `<script>...</script>` blocks today, and layering `eval()` and the like on as well would be helpful.

mikesamuel:
Originally floated as a [strawman][1]

> ### `eval` is Evil
>
> The `eval` function is the most misused feature of JavaScript. Avoid it.
>
> -- <cite>Douglas Crockford, "JavaScript: The Good Parts"</cite>

`eval` and its friend `new Function` are problematic because, too
often, an attacker can turn it against the application.

Most code avoids `eval`, but JS programs are no longer small, and
self-contained as they were when Crock wrote that.

If one module uses `eval`, even if it's as simple as
[`Function('return this')()`][2] to get a handle to the
global object then `eval` has to work.

This prevents the use of security measures like:

*  [Content-Security-Policy][]
*  `node --disallow_code_generation_from_strings`

which turn off `eval` globally.

As JavaScript programs get larger, the chance that no part of it needs `eval` or `Function()`
to operate gets smaller.

----

It is difficult in JavaScript for a code reviewer to determine that
code never uses these operators.  For example, the below can when `x` is constructor.

```js
({})[x][x](y)()

// ({})[x]       === Object
// ({})[x][x]    === Function
// ({})[x][x](y)  ~  Function(y)
```

So code review and developer training are unlikely to prevent abuse of these
operators.

----

This aims to solve the problem by providing more context to host environments so that they
can make finer-grained trust decisions.

The [Trusted Types proposal][3] aims to allow JavaScript development to scale securely.

It makes it easy to separate:

*  the decisions to trust a chunk of code to load.
*  the check that an input to a sensitive operator like `eval` is trustworthy.

This allows trust decisions tto be made where the maximum context is
available and allows these decisions to be concentrated in small
amounts of thoroughly reviewed code

The checks can also be moved into host code so that they reliably happen before
irrevocable actions like code loading complete.

Specifically, Trusted Types would like to require that the code portion of the
inputs to %Function% and %eval% are [*TrustedScript*][4].

[1]: https://github.com/mikesamuel/proposal-hostensurecancompilestrings-passthru/
[2]: https://github.com/zloirock/core-js/blob/2a005abe68520248d4431cab70d86e40b55d6e98/packages/core-js/internals/global.js#L5
[3]: https://wicg.github.io/trusted-types/dist/spec/
[4]: https://wicg.github.io/trusted-types/dist/spec/#typedef-trustedscript
[Content-Security-Policy]: https://csp.withgoogle.com/docs/index.html

Fixes #938.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12510,7 +12510,7 @@
               1. Let _evalText_ be the first element of _argList_.
               1. If the source code matching this |CallExpression| is strict mode code, let _strictCaller_ be *true*. Otherwise let _strictCaller_ be *false*.
               1. Let _evalRealm_ be the current Realm Record.
-              1. Perform ? HostEnsureCanCompileStrings(_evalRealm_, _evalRealm_).
+              1. Perform ? HostEnsureCanCompileStrings(_evalRealm_, _evalRealm_, Script, _evalText_).
               1. Return ? PerformEval(_evalText_, _evalRealm_, _strictCaller_, *true*).
           1. Let _thisCall_ be this |CallExpression|.
           1. Let _tailCall_ be IsInTailPosition(_thisCall_).
@@ -23365,7 +23365,7 @@
         1. Let _callerContext_ be the second to top element of the execution context stack.
         1. Let _callerRealm_ be _callerContext_'s Realm.
         1. Let _calleeRealm_ be the current Realm Record.
-        1. Perform ? HostEnsureCanCompileStrings(_callerRealm_, _calleeRealm_).
+        1. Perform ? HostEnsureCanCompileStrings(_callerRealm_, _calleeRealm_, Script, _x_).
         1. Return ? PerformEval(_x_, _calleeRealm_, *false*, *false*).
       </emu-alg>
 
@@ -23449,11 +23449,17 @@
       </emu-clause>
 
       <emu-clause id="sec-hostensurecancompilestrings" aoid="HostEnsureCanCompileStrings">
-        <h1>HostEnsureCanCompileStrings ( _callerRealm_, _calleeRealm_ )</h1>
+        <h1>HostEnsureCanCompileStrings ( _callerRealm_, _calleeRealm_[, _goal_[, _bodyText_]] )</h1>
 
         <p>HostEnsureCanCompileStrings is an implementation-defined abstract operation that allows host environments to block certain ECMAScript functions which allow developers to compile strings into ECMAScript code.</p>
 
         <p>An implementation of HostEnsureCanCompileStrings may complete normally or abruptly. Any abrupt completions will be propagated to its callers. The default implementation of HostEnsureCanCompileStrings is to unconditionally return an empty normal completion.</p>
+
+        <emu-note>
+          <p>_goal_ is the production which will be used to parse _bodyText_.  For example, if called via %Function% it might be the grammar symbol |FunctionBody[~Yield, ~Await]| and from %eval% it might be the grammar symbol |Script|.</p>
+        </emu-note>
+
+        <p>Implementations of HostEnsureCanCompileStrings should not coerce _bodyText_ to a string since callers will later do that and HostEnsureCanCompileStrings must not base security decisions on both coercions producing the same string.  Clients may base security decisions on internal slots.</p>
       </emu-clause>
 
       <emu-clause id="sec-evaldeclarationinstantiation" aoid="EvalDeclarationInstantiation">
@@ -24673,7 +24679,6 @@
             1. Let _callerContext_ be the second to top element of the execution context stack.
             1. Let _callerRealm_ be _callerContext_'s Realm.
             1. Let _calleeRealm_ be the current Realm Record.
-            1. Perform ? HostEnsureCanCompileStrings(_callerRealm_, _calleeRealm_).
             1. If _newTarget_ is *undefined*, set _newTarget_ to _constructor_.
             1. If _kind_ is `"normal"`, then
               1. Let _goal_ be the grammar symbol |FunctionBody[~Yield, ~Await]|.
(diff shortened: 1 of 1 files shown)


mikesamuel:
I'd like to schedule this for a needs_consensus slot per https://github.com/tc39/proposals/pull/209#discussion_r271809863

This should be a semantics free change from the ES side.

Tracking issue filed on the HTML5 side: https://github.com/whatwg/html/issues/4501

spec.html
```diff
@@ -24693,9 +24698,12 @@ <h1>Runtime Semantics: CreateDynamicFunction ( _constructor_, _newTarget_, _kind
               1. Let _parameterGoal_ be the grammar symbol |FormalParameters[+Yield, +Await]|.
               1. Let _fallbackProto_ be `"%AsyncGenerator%"`.
             1. Let _argCount_ be the number of elements in _args_.
-            1. Let _P_ be the empty String.
             1. If _argCount_ = 0, let _bodyText_ be the empty String.
-            1. Else if _argCount_ = 1, let _bodyText_ be _args_[0].
+            1. Else _argCount_ &ge; 1,
```
mikesamuel: The goal of splitting the loop out of this if-tree is to preserve the order of side effects due to stringification while allowing the host callout access to bodyText.

The order of events should stay the same:

1. Host callout
2. Parameter name arguments stringified in order
3. Body text stringified

mikesamuel:
How might one specify tests?

AFAIK, there's no way for a test262 implementation to specify that a test runs in the context of a less-than-completely permissive host callout.   https://github.com/tc39/test262/issues/2120

Would it be sufficient to include a prose section on what to test like the below?

> Run in the context of a HostEnsureCanCompileStrings callout that
> does blah blah blah:
> 
> ```js
> // test code here
> ```

like https://gist.github.com/mikesamuel/5c71584fc10a603167111e933cdc650d

mikesamuel:
Fixes #938

ljharb:
It may indeed not be possible to test; that's something i think the test262 maintainers can help confirm.

littledan:
I think you'd need more changes than just this. Step 2 of [PerformEval](https://tc39.github.io/ecma262/#sec-performeval) will not evaluate non-String arguments, so it wouldn't really be very useful to apply with a TrustedScript.

mikesamuel:
@littledan, I put together https://github.com/mikesamuel/evalable#tc39-proposal-evalable-stage-0 based on https://github.com/tc39/ecma262/issues/938#issuecomment-457352474

Jamesernator:
Question. Do we even need to callout to `HostEnsureCanCompileStrings` for the case of `new Function()` (with no params)? It doesn't seem to me that `new Function()` on it's own can be used as a vector for XSS attacks.

mikesamuel:
@Jamesernator 
You're right that `(new Function)` is not an XSS risk but I suspect I'm missing something more in your question.

IIUC `new Function` reaches CreateDynamicFunction with zero arguments which reaches [19.2.1.1.1.1 step 13](https://tc39.github.io/ecma262/#sec-createdynamicfunction)

> 13. If argCount = 0, let bodyText be the empty String.

so `new Function('')` and `(new Function)` are equivalent and both are already guarded by the host callout.

Jamesernator:
It just seems surprising to me that `new Function()` needs to invoke the machinery for checking whether or not a string is safe given that it wasn't given a string in the first place. (Given that `eval(nonString)` doesn't bother invoking the HostEnsureCanCompileStrings machinery when it's not given a string).

spec.html
```diff
@@ -24693,9 +24698,12 @@ <h1>Runtime Semantics: CreateDynamicFunction ( _constructor_, _newTarget_, _kind
...
+              1. Let _bodyText_ be _args_[_argCount_ - 1].
+            1. Perform ? HostEnsureCanCompileStrings(_callerRealm_, _calleeRealm_, _goal_, _bodyText_).
+            1. Let _P_ be the empty String.
+            1. If _argCount_ = 1, let _bodyText_ be _args_[0].
```
littledan: I don't understand the purpose of this line. Wasn't this already handled four lines up?

littledan (commented):
This layering change generally LGTM. I'd like to wait to land it until we have a PR within a host environment that takes advantage of this PR.

spec.html
ExE-Boss, line 24702:
```suggestion
            1. Else,
              1. Assert: _argCount_ &ge; 1,
```

spec.html
```diff
@@ -24693,9 +24698,12 @@ <h1>Runtime Semantics: CreateDynamicFunction ( _constructor_, _newTarget_, _kind
...
             1. Else _argCount_ &gt; 1,
```
ExE-Boss, line 24707:
```suggestion
            1. If _argCount_ &gt; 1,
```

spec.html
ExE-Boss: It indeed was.

caridy:
We would try to revamp this PR (cc @ptomato)

ptomato:
I've opened a rebased version of this over at https://github.com/tc39/ecma262/pull/3222.

caridy:
@ptomato should we close this, or do you prefer to keep it around for when the time comes to work on trusted types?

ljharb:
If it should be closed, it’d be done on merging of the replacement, I’d expect.

nicolo-ribaudo:
Note that the other PR only replaces _part_ of the motivation of this one. If we'll want to ever support trusted types, we'll need to also pass the uncoerced objects to the host (which is what this PR does).

PR #1499, opened 2019-04-03
Editorial: Introduce StringPad operation backing padStart and padEnd

gibson042:
Per https://github.com/tc39/ecma262/pull/1407#issuecomment-479619396

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -28518,23 +28518,8 @@ THH:mm:ss.sss
         <p>When the `padEnd` method is called, the following steps are taken:</p>
         <emu-alg>
           1. Let _O_ be ? RequireObjectCoercible(*this* value).
-          1. Let _S_ be ? ToString(_O_).
-          1. Let _intMaxLength_ be ? ToLength(_maxLength_).
-          1. Let _stringLength_ be the length of _S_.
-          1. If _intMaxLength_ is not greater than _stringLength_, return _S_.
-          1. If _fillString_ is *undefined*, let _filler_ be the String value consisting solely of the code unit 0x0020 (SPACE).
-          1. Else, let _filler_ be ? ToString(_fillString_).
-          1. If _filler_ is the empty String, return _S_.
-          1. Let _fillLen_ be _intMaxLength_ - _stringLength_.
-          1. Let _truncatedStringFiller_ be the String value consisting of repeated concatenations of _filler_ truncated to length _fillLen_.
-          1. Return the string-concatenation of _S_ and _truncatedStringFiller_.
+          1. Return ? StringPad(_O_, _maxLength_, _fillString_, *"end"*).
         </emu-alg>
-        <emu-note>
-          <p>The first argument _maxLength_ will be clamped such that it can be no smaller than the length of the *this* value.</p>
-        </emu-note>
-        <emu-note>
-          <p>The optional second argument _fillString_ defaults to *" "* (the String value consisting of the code unit 0x0020 SPACE).</p>
-        </emu-note>
       </emu-clause>
 
       <emu-clause id="sec-string.prototype.padstart">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -28485,47 +28485,43 @@ <h1>String.prototype.padEnd ( _maxLength_ [ , _fillString_ ] )</h1>
...
-          1. If _fillString_ is *undefined*, let _filler_ be the String value consisting solely of the code unit 0x0020 (SPACE).
-          1. Else, let _filler_ be ? ToString(_fillString_).
-          1. If _filler_ is the empty String, return _S_.
-          1. Let _fillLen_ be _intMaxLength_ - _stringLength_.
-          1. Let _truncatedStringFiller_ be the String value consisting of repeated concatenations of _filler_ truncated to length _fillLen_.
-          1. Return the string-concatenation of _truncatedStringFiller_ and _S_.
+          1. Return ? StringPad(_O_, _maxLength_, _fillString_, *"start"*).
         </emu-alg>
-        <emu-note>
-          <p>The first argument _maxLength_ will be clamped such that it can be no smaller than the length of the *this* value.</p>
```
ljharb: we should probably keep this non-normative note here?

spec.html
```diff
@@ -28485,47 +28485,43 @@ <h1>String.prototype.padEnd ( _maxLength_ [ , _fillString_ ] )</h1>
...
-        </emu-note>
-        <emu-note>
-          <p>The optional second argument _fillString_ defaults to *" "* (the String value consisting of the code unit 0x0020 SPACE).</p>
```
ljharb: also here?

spec.html
```diff
@@ -28485,47 +28485,43 @@ <h1>String.prototype.padEnd ( _maxLength_ [ , _fillString_ ] )</h1>
...
-          1. Let _fillLen_ be _intMaxLength_ - _stringLength_.
-          1. Let _truncatedStringFiller_ be the String value consisting of repeated concatenations of _filler_ truncated to length _fillLen_.
-          1. Return the string-concatenation of _S_ and _truncatedStringFiller_.
+          1. Return ? StringPad(_O_, _maxLength_, _fillString_, *"end"*).
         </emu-alg>
-        <emu-note>
-          <p>The first argument _maxLength_ will be clamped such that it can be no smaller than the length of the *this* value.</p>
-        </emu-note>
-        <emu-note>
-          <p>The optional second argument _fillString_ defaults to *" "* (the String value consisting of the code unit 0x0020 SPACE).</p>
```
ljharb: also here?

spec.html
```diff
@@ -28485,47 +28485,43 @@ <h1>String.prototype.padEnd ( _maxLength_ [ , _fillString_ ] )</h1>
...
-          1. If _fillString_ is *undefined*, let _filler_ be the String value consisting solely of the code unit 0x0020 (SPACE).
-          1. Else, let _filler_ be ? ToString(_fillString_).
-          1. If _filler_ is the empty String, return _S_.
-          1. Let _fillLen_ be _intMaxLength_ - _stringLength_.
-          1. Let _truncatedStringFiller_ be the String value consisting of repeated concatenations of _filler_ truncated to length _fillLen_.
-          1. Return the string-concatenation of _S_ and _truncatedStringFiller_.
+          1. Return ? StringPad(_O_, _maxLength_, _fillString_, *"end"*).
         </emu-alg>
-        <emu-note>
-          <p>The first argument _maxLength_ will be clamped such that it can be no smaller than the length of the *this* value.</p>
```
ljharb: we should probably keep this non-normative note here?

spec.html
```diff
@@ -28485,47 +28485,43 @@ <h1>String.prototype.padEnd ( _maxLength_ [ , _fillString_ ] )</h1>
...
+            1. Let _stringLength_ be the length of _S_.
+            1. If _intMaxLength_ is not greater than _stringLength_, return _S_.
+            1. If _fillString_ is *undefined*, let _filler_ be the String value consisting solely of the code unit 0x0020 (SPACE).
+            1. Else, let _filler_ be ? ToString(_fillString_).
+            1. If _filler_ is the empty String, return _S_.
+            1. Let _fillLen_ be _intMaxLength_ - _stringLength_.
+            1. Let _truncatedStringFiller_ be the String value consisting of repeated concatenations of _filler_ truncated to length _fillLen_.
+            1. If _placement_ is *"end"*, return the string-concatenation of _S_ and _truncatedStringFiller_.
+            1. Else,
+              1. Assert: _placement_ is *"start"*.
```
ljharb: let's maybe assert that `_placement_` is start or end at the top of the operation, before the observable things happen? seems a bit more clear to me.

spec.html
gibson042: Are you saying you want the notes in the prototype methods _and_ the abstract op, or just the former?

ljharb: That's a good question. It seems useful to have them on the methods in particular, since that's what people are likely to read first, but then it seems strange not to include them on the op as well. Let's see what @zenparsing or others think?

spec.html
gibson042: Sure; done.

spec.html
zenparsing: I think we ought to either move the notes to the abstract op, as @gibson042 has done, or leave them in each method, but probably not both. Since the point of the PR is to reduce duplication, I would probably move them into the abstract op (closer to the algorithms which they reference), but either way is fine.

ljharb: In that case, seems like we can leave it as-is.

PR #1500, opened 2019-04-03
Meta: Improve the explanation of normative changes

gibson042:
Definition copied from the "normative change" label
(cf. https://github.com/tc39/ecma262/labels?q=normative )

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ To file an issue, go to the ecma262 [issues page](https://github.com/tc39/ecma26
 To make a pull request (PR), [fork](https://help.github.com/articles/fork-a-repo/) the [ecma262](https://github.com/tc39/ecma262), apply changes to the spec.html, and upload it to your fork on GitHub, using the web interface to file a pull request. Locally, to see how your change renders in HTML, use [ecmarkup](https://github.com/bterlson/ecmarkup) to build spec.html into an actual HTML file.
 
 Commits in pull requests should have a first line which starts with a tag, followed by a colon, indicating which type of patch they are:
-  * Normative: changes impacting JavaScript developers in some way
+  * Normative: any changes that affect behavior required to correctly evaluate some ECMAScript source text (such as a script or module)
   * Editorial: any non-normative changes to spec text including typo fixes, changes to the document style, etc.
   * Markup: non-visible changes to markup in the spec
   * Meta: changes to documents about this repository (e.g. readme.md or contributing.md) and other supporting documents or scripts (e.g. package.json, design documents, etc.)


ljharb:
hmm, I thought I'd provided all the label descriptions, but I don't recall that exact one - I'm not sure where it comes from.

ljharb:
(this commit message will need a "Meta: " prefix before landing)

CONTRIBUTING.md
```diff
@@ -11,7 +11,7 @@ To file an issue, go to the ecma262 [issues page](https://github.com/tc39/ecma26
 To make a pull request (PR), [fork](https://help.github.com/articles/fork-a-repo/) the [ecma262](https://github.com/tc39/ecma262), apply changes to the spec.html, and upload it to your fork on GitHub, using the web interface to file a pull request. Locally, to see how your change renders in HTML, use [ecmarkup](https://github.com/bterlson/ecmarkup) to build spec.html into an actual HTML file.
 
 Commits in pull requests should have a first line which starts with a tag, followed by a colon, indicating which type of patch they are:
-  * Normative: changes impacting JavaScript developers in some way
+  * Normative: any changes that affect the interpretation of some source texts
```
zenparsing: I've typically heard and used the terminology "observable semantics" in these kinds of contexts.

Would something like this be more clear?

```
any changes that affect the observable behavior of JavaScript implementations
```

gibson042: I like that very much. Updated in the PR, and it would be a good idea to update the GitHub label description as well.

littledan: This wording feels a bit obscure. You have to know what "observable" and "implementation" mean to understand it. Is there any way we could phrase this in more basic terms, while still getting the same point across? e.g., "Any changes that affect what some JavaScript program does when executed."

littledan:
It'd be great to have your help in the https://github.com/tc39/how-we-work repository, as well.

CONTRIBUTING.md
ljharb: Are you expecting a pull request from someone that doesn’t understand these (imo relatively self-explanatory) terms?

littledan: Yes. Both of those two terms have been mentioned as confusing in the past. Currently, observable is a not-yet-filled-in definition in the how-we-work glossary.

gibson042: So the goal here is to maximize comprehensibility for inexperienced TC39 contributors while maintaining precise correctness for everyone. How about this?
> any changes that affect behavior required to correctly evaluate some ECMAScript source text (such as a script or module)

Not all normative changes are necessarily observable to running code anyway (e.g., WeakMap memory reclamation), nor are all or even any implementations necessarily affected by every normative change (e.g., those that specify alignment on preexisting reality such as `Array.prototype.sort` stability). But I believe all normative changes **do** affect the set of implementation _requirements_ (usually adding to them but conceptually also including reductions), and furthermore that this is the distinguishing characteristic of a normative change.

gibson042: I have updated this PR to the above suggestion.

PR #242, opened 2015-12-10
Layering: track execution contexts' script/module

domenic:
Fixes #78. Introduces the concept of a Script Record as a counterpart to a Module Record, and adds [[HostDefined]] fields to both of them. Every execution environment now has ScriptOrModule component pointing back to its "creator" script or module:

- Top level script/module execution contexts point to the appropriate script/module.
- Function execution contexts for functions declared at top level point to that of the containing script/module.
- Function execution contexts for functions declared inside other functions point to that of the containing function.
- Eval execution contexts point to that of their containing function.
- Job execution contexts (which are largely a spec artifact) point to the script/module that originally enqueued the job.

In the course of doing this, factored out ParseScript and ScriptEvaluation abstract operations, which now better parallel modules.

----

I think this is pretty good but review would be appreciated, especially where I converted ScriptEvaluation from Runtime Semantics into an abstract operation. The grammar interactions aren't 100% clear to me there.

PR #1501, opened 2019-04-04
Layering: Add GetRequestedModules() to Module Record

dandclark:
Add the GetRequestedModules() abstract method to Abstract Module Record, with a corresponding implementation for Source Text Module Record.  Move Cyclic Module Record's [[RequestedModules]] down to Source Text Module Record, as it is now used directly only by STMR, and its presence on Cyclic MR causes problems for derived module types like HTML Module Record that don't necessarily have strings associated with all their child modules.

The purpose of GetRequestedModules() is to abstract out from InnerModuleInstantiation/InnerModuleEvaluation the details of how a Module Record type stores its requested modules, since these are different for HTML Module Records as implemented [here](https://github.com/whatwg/html/pull/4505).

It is not time yet to merge this in given that there are still open questions about the HTML Modules design (mostly tracked over in the [w3c/webcomponents repo](https://github.com/w3c/webcomponents/labels/modules)).  The intent here is to facilitate discussion and tease out issues that can be better found when making changes against the actual spec, with the intention of eventually merging in the updated PR once consensus has been reached on all open issues.  

I've placed the built result of this change here: https://dandclark.github.io/built-specs/ecma262/get-requested-modules.html

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -21709,7 +21709,7 @@
                   GetRequestedModules()
                 </td>
                 <td>
-                  Returns a List of Module Records constituting the modules imported from this module.  The List is source code occurrence ordered.
+                  Returns a List of Module Records constituting the modules imported from this module. The List is source code occurrence ordered.
                 </td>
               </tr>
             </tbody>
@@ -21762,8 +21762,8 @@
               1. Set _module_.[[DFSAncestorIndex]] to _index_.
               1. Increase _index_ by 1.
               1. Append _module_ to _stack_.
-              1. Let _requiredModules_ be _module_.GetRequestedModules()
-              1. For each module _requiredModule_ that is an element of _requiredModules_, do
+              1. Let _requiredModules_ be _module_.GetRequestedModules().
+              1. For each element _requiredModule_ of _requiredModules_ in List order, do
                 1. Set _index_ to ? InnerModuleInstantiation(_requiredModule_, _stack_, _index_).
                 1. Assert: _requiredModule_.[[Status]] is either `"instantiating"`, `"instantiated"`, or `"evaluated"`.
                 1. Assert: _requiredModule_.[[Status]] is `"instantiating"` if and only if _requiredModule_ is in _stack_.
@@ -21834,7 +21834,7 @@
               1. Increase _index_ by 1.
               1. Append _module_ to _stack_.
               1. Let _requiredModules_ be _module_.GetRequestedModules().
-              1. For each module _requiredModule_ that is an element of _requiredModules_, do
+              1. For each element _requiredModule_ of _requiredModules_ in List order, do
                 1. Set _index_ to ? InnerModuleEvaluation(_requiredModule_, _stack_, _index_).
                 1. Assert: _requiredModule_.[[Status]] is either `"evaluating"` or `"evaluated"`.
                 1. Assert: _requiredModule_.[[Status]] is `"evaluating"` if and only if _requiredModule_ is in _stack_.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -21754,8 +21762,8 @@ <h1>InnerModuleInstantiation ( _module_, _stack_, _index_ )</h1>
               1. Set _module_.[[DFSAncestorIndex]] to _index_.
               1. Increase _index_ by 1.
               1. Append _module_ to _stack_.
-              1. For each String _required_ that is an element of _module_.[[RequestedModules]], do
-                1. Let _requiredModule_ be ? HostResolveImportedModule(_module_, _required_).
+              1. Let _requiredModules_ be _module_.GetRequestedModules()
+              1. For each module _requiredModule_ that is an element of _requiredModules_, do
```
ljharb: This suggestion mirrors the language used here: https://tc39.github.io/ecma262/#sec-enumerableownpropertynames
```suggestion
              1. For each element _requiredModule_ of _requiredModules_ in List order, do
```

Not sure if we have a prevailing convention on this or not.

spec.html
```diff
@@ -21704,6 +21704,14 @@ <h1>Cyclic Module Records</h1>
                   Initialize the execution context of the module and evaluate the module's code within it.
                 </td>
               </tr>
+              <tr>
+                <td>
+                  GetRequestedModules()
+                </td>
+                <td>
+                  Returns a List of Module Records constituting the modules imported from this module.  The List is source code occurrence ordered.
```
```suggestion
                  Returns a List of Module Records constituting the modules imported from this module. The List is source code occurrence ordered.
```

dandclark: Fixed.  Thanks!

spec.html
dandclark: Fixed.  Thanks!

spec.html
```diff
@@ -22543,6 +22550,24 @@ <h1>ExecuteModule ( ) Concrete Method</h1>
...
+
+          <p>The GetRequestedModules concrete method of a Source Text Module Record implements the corresponding Cyclic Module Record abstract method.</p>
+
+          <p>This abstract method performs the following steps:</p>
+
+          <emu-alg>
+            1. Let _module_ be this Source Text Module Record.
+            1. Let _requestedModules_ be an empty List of Abstract Module Records.
+            1. For each String _requested_ that is an element of _module_.[[RequestedModules]], do
+              1. Let _requestedModule_ be ! HostResolveImportedModule(_module_, _requested_).
```
littledan: The specification previously had `?` rather than `!` from some paths, so I think it would be best to leave it that way here. Then, the two callsites can invoke it with either of those, based on whether they anticipate errors.

littledan (commented):
This change seems good to me. It'll cause a tiny bit of duplication for WebAssembly modules, but that's fine.

A nit: Maybe `[[RequiredModules]]` should move down to SourceTextModuleRecord if its usages are exclusively there.

Thanks for following up with these appropriate layering changes; I am excited to see HTML modules moving forward.

spec.html
dandclark: Thanks!  I like the idea of moving [[RequiredModules]] down to STMR...I'll look into this.

littledan:
Nit: I see this marked as "Normative" in the PR title. However, I don't know what normative changes are in this PR. To me, it looks more like a "Layering" (or "Editorial") PR, which should be possible to land without tests or prior review in plenary (unless it seems controversial or worth noting in committee).

ljharb:
Layering PRs are often normative; and we don’t have a different commit category for them. In particular, almost any change to Modules is effectively normative due to the partial nature in which 262 specifies them.

littledan:
Just for context, we have used the "Layering:" prefix on commits before, e.g., https://github.com/tc39/ecma262/pull/242 , even if I failed to list this when writing our current CONTRIBUTING.md. I still don't understand how this patch is normative, in a way that would make other layering patches not normative (since embedders can hook into any of the layering changes). I don't expect any JavaScript programs to change their behavior due to this change.

ljharb:
That's a fair point, and maybe that makes this one editorial; either way, I'd expect anything that impacts Modules, especially CMRs and STMRs, to require committee consensus (otherwise the "dynamic modules" agenda item wouldn't have required it either).

spec.html
dandclark: I pushed an update that:
- Switched '?' to '!' in GetRequestedModules and removed the note stating that the HostResolveImportedModules call must have succeeded.
- Pushed [[RequestedModules]] down to Source Text Module Record.

Thanks for the comments!

dandclark:
Yeah, when creating the PR I thought it didn't quite fit into either 'Normative' or 'Editorial' but didn't know there was precedent for using 'Layering' as a category. :)  I've updated the title to 'Layering'.

In any case I agree that a module chage like this likely requires committee consensus.

littledan (approved):
I don't see any issues with this PR. Good work! However, maybe we should wait to land it until there's a bit more clarity on what the semantics of HTML modules will be.

zenparsing:
I'm curious if the desired indirection cannot instead be applied in `HostResolveImportedModule`.

The `[[RequestedModules]]` list can be any arbitrary strings: their semantics depends upon the module record that contains them. A cyclic module record _m_ could have auto-generated keys in `[[RequestedModules]]`, say `#1` and `#2`. `HostResolveImportedModule(m, '#1')` would then resolve to the desired dependency.

PR #1504, opened 2019-04-09
Normative: `eval(nonString)` should not have observable side effects

### LINKED ISSUE #1495: inconsistent order of host and type checks for eval
JS engines are not consistent in the order in which they apply type checks and host callouts during `eval`.

`eval` applied to non string arguments is the identity function because of step 2 of the definition of [PerformEval](https://tc39.github.io/ecma262/#sec-performeval):

> ### 18.2.1.1. Runtime Semantics: PerformEval ( *x*, *evalRealm*, *strictCaller*, *direct* )
> <p>The abstract operation PerformEval with arguments *x*, *evalRealm*, *strictCaller*, and *direct* performs the following steps:</p>
>
> 1. Assert: If *direct* is **false**, then *strictCaller* is also **false**.
> 1. If Type(*x*) **is not String**, return *x*.
> 1. ...

This type check happens after the call to HostEnsureCanCompileStrings:

> ### 18.2.1 eval ( _x_ )
> <p>The `eval` function is the <dfn>%eval%</dfn> intrinsic object. When the `eval` function is called with one argument *x*, the following steps are taken:</p>
>
> 1. Assert: The execution context stack has at least two elements.
> 1. Let *callerContext* be the second to top element of the execution context stack.
> 1. Let *callerRealm* be *callerContext*'s Realm.
> 1. Let *calleeRealm* be the current Realm Record.
> 1. Perform ? **HostEnsureCanCompileStrings(*callerRealm*, *calleeRealm*)**.
> 1. Return ? PerformEval(*x*, *calleeRealm*, **false**, **false**).

but when the argument is not a string, some engines do not seem to perform HostEnsureCanCompileStrings.

```sh
$ npx node@10 --disallow_code_generation_from_strings -e 'console.log(eval(0))'; \
   echo Exit status $?
0
Exit status 0

$ npx node@10 --disallow_code_generation_from_strings -e 'console.log(eval("0"))'; \
  echo Exit status $?
[eval]:1
console.log(eval("0"))
        ^

EvalError: Code generation from strings disallowed for this context
    at [eval]:1:9
    at Script.runInThisContext (vm.js:119:20)
    at Object.runInThisContext (vm.js:326:38)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at evalScript (internal/bootstrap/node.js:589:27)
    at startup (internal/bootstrap/node.js:265:9)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
Exit status 1
```

----

I tested this in various browsers via

```html
<meta http-equiv="Content-Security-Policy" content="script-src 'nonce-test'">
<script nonce="test">
  console.group('eval number');
  try {
    console.log(eval(0));
  } finally {
    console.groupEnd();
  }
</script>
<script nonce="test">
  console.group('eval string');
  try {
    console.log(eval("0"));
  } finally {
    console.groupEnd();
  }
</script>
```

| Browser | `eval(0)` |
| --------- | ------- |
| Chrome 73.0 | allowed |
| Firefox 66.0 | rejected |
| Safari 12.0 | allowed |

I don't have IE / Edge handy.

mikesamuel:
Fixes #1495
which identifies the ordering of host callouts and input
type checks as a spec bug.

This uses option 1 from
https://github.com/tc39/ecma262/issues/1495#issuecomment-478626560
to ensure that type checks happen before host callouts.

This can't be tested in test262 but if this lands, I will follow up
with tests via web-platform-tests/wpt.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12569,13 +12569,12 @@
               1. Let _evalText_ be the first element of _argList_.
               1. If the source code matching this |CallExpression| is strict mode code, let _strictCaller_ be *true*. Otherwise let _strictCaller_ be *false*.
               1. Let _evalRealm_ be the current Realm Record.
-              1. Perform ? HostEnsureCanCompileStrings(_evalRealm_, _evalRealm_).
               1. Return ? PerformEval(_evalText_, _evalRealm_, _strictCaller_, *true*).
           1. Let _thisCall_ be this |CallExpression|.
           1. Let _tailCall_ be IsInTailPosition(_thisCall_).
           1. Return ? EvaluateCall(_func_, _ref_, _arguments_, _tailCall_).
         </emu-alg>
-        <p>A |CallExpression| evaluation that executes step 6.a.vii is a <dfn>direct eval</dfn>.</p>
+        <p>A |CallExpression| evaluation that executes step 6.a.vi is a <dfn>direct eval</dfn>.</p>
         <emu-grammar>CallExpression : CallExpression Arguments</emu-grammar>
         <emu-alg>
           1. Let _ref_ be the result of evaluating |CallExpression|.
@@ -23417,17 +23416,17 @@
         1. Assert: The execution context stack has at least two elements.
         1. Let _callerContext_ be the second to top element of the execution context stack.
         1. Let _callerRealm_ be _callerContext_'s Realm.
-        1. Let _calleeRealm_ be the current Realm Record.
-        1. Perform ? HostEnsureCanCompileStrings(_callerRealm_, _calleeRealm_).
-        1. Return ? PerformEval(_x_, _calleeRealm_, *false*, *false*).
+        1. Return ? PerformEval(_x_, _callerRealm_, *false*, *false*).
       </emu-alg>
 
       <emu-clause id="sec-performeval" aoid="PerformEval">
-        <h1>Runtime Semantics: PerformEval ( _x_, _evalRealm_, _strictCaller_, _direct_ )</h1>
-        <p>The abstract operation PerformEval with arguments _x_, _evalRealm_, _strictCaller_, and _direct_ performs the following steps:</p>
+        <h1>Runtime Semantics: PerformEval ( _x_, _callerRealm_, _strictCaller_, _direct_ )</h1>
+        <p>The abstract operation PerformEval with arguments _x_, _callerRealm_, _strictCaller_, and _direct_ performs the following steps:</p>
         <emu-alg>
           1. Assert: If _direct_ is *false*, then _strictCaller_ is also *false*.
           1. If Type(_x_) is not String, return _x_.
+          1. Let _evalRealm_ be the current Realm Record.
+          1. Perform ? HostEnsureCanCompileStrings(_callerRealm_, _evalRealm_).
           1. Let _thisEnvRec_ be ! GetThisEnvironment().
           1. If _thisEnvRec_ is a function Environment Record, then
             1. Let _F_ be _thisEnvRec_.[[FunctionObject]].


spec.html
```diff
@@ -23365,16 +23364,16 @@ <h1>eval ( _x_ )</h1>
...
         1. Let _calleeRealm_ be the current Realm Record.
-        1. Perform ? HostEnsureCanCompileStrings(_callerRealm_, _calleeRealm_).
-        1. Return ? PerformEval(_x_, _calleeRealm_, *false*, *false*).
+        1. Return ? PerformEval(_x_, _callerRealm_, _calleeRealm_, *false*, *false*).
       </emu-alg>
 
       <emu-clause id="sec-performeval" aoid="PerformEval">
-        <h1>Runtime Semantics: PerformEval ( _x_, _evalRealm_, _strictCaller_, _direct_ )</h1>
-        <p>The abstract operation PerformEval with arguments _x_, _evalRealm_, _strictCaller_, and _direct_ performs the following steps:</p>
+        <h1>Runtime Semantics: PerformEval ( _x_, _callerRealm_, _calleeRealm_, _strictCaller_, _direct_ )</h1>
```
zenparsing: I believe that in all cases, _evalRealm_ is "the current Realm Record".

I suppose it would be possible to only pass in the _callerRealm_ to PerformEval and do `Let _evalRealm_ be the current Realm Record` right before the call to `HostEnsureCanCompileStrings`. Thoughts?

mikesamuel: You convinced me.  Response I wrote while chewing over the idea below.

----

I believe you're correct about all existing calls, and there're unlikely to be any new calls to PerformEval since MakeDynamicFunction is a probably a better alternative for new dynamic operators.

I'd rather not do unrelated fixup that makes quick consensus harder but that would avoid expanding the parameter list and it eliminates the need to rename _evalRealm_ to _calleeRealm_.

erights: I find the spec language for PerformEval terribly confusing because it tries to cover both direct and indirect eval in one spec. @littledan and I tried to decipher it wrt a similar issue and were not able to do so quickly.

In both cases, only one realm should be relevant. But I don't know how what I'm about to say maps onto the spec's terms "callerRealm", "calleeRealm", and "current realm".

Indirect eval is just a function call. JS is a lexically scoped language, both in syntax and in philosophy. We have taken pains to keep caller sensitivity out of the semantics of functions being called. The function being called is passed explicit arguments, and, if you want to describe it that way, one implicit argument --- the continuation. Nothing not passed by the caller influences the semantics of the callee once invoked. There remains a de-facto sloppy mode exception, `.caller`, which is one of our very few successes at successful deprecation.

The "lexical capture" analog for builtins is that implicit access that one intrinsic has to another is to that intrinsic from its own state. For indirect eval, the only realm that should matter is the realm it was born into.

The direct eval *syntax* is a special form. The meaning of the special form is conditioned on a test whether the `eval` variable reference evaluates to a special `eval` intrinsic. If not, then the special form has the semantics of the function call it seems to be. Since in this case the thing called is not the `eval` intrinsic, we need not worry about this further.

The remaining direct eval is a full special form. The identity of the eval object that passed that test has no further influence on the meaning of the operation. As a special form, its meaning is as part of the execution of the code the special form appears in, including the realm for this execution.

If someone could please explain how these distinctions correspond to the spec language ---"callerRealm", "calleeRealm", and "current realm" --- separately for both direct and indirect eval, it would help us discuss both the current issue and the puzzle @littledan and I were trying to figure out. Thanks.

mikesamuel: @erights Are you asking for a non-normative note explaining the relationship between the inputs to PerformEval?

I may be wrong, but as I understand it, assuming you're running in a browser context where globals like `self`, `parent`, etc. have their [usual meanings](https://searchfox.org/mozilla-central/rev/5053031ba7621fa8f63f42de4c204ab3561e4e59/dom/webidl/Window.webidl#35):

```js
self.parent.eval('0')
```

*  *evalRealm* and *calleeRealm* and the current execution context correspond to `self.parent` since [PrepareCall](https://tc39.github.io/ecma262/#sec-prepareforordinarycall) created an execution context for the `eval` builtin function and pushed it to the top of the stack.
    > The value of the Realm component of the running execution context is also called the current Realm Record.
*  *callerRealm* corresponds to `self` since the execution context above was in the scope of `self`.

Does that accord with your understanding?  Is it useful at all?

erights: At this time, I am not asking for anything to be added to the spec, normative or not. I seek only to understand what the spec means.

Your explanation reiterates the explanatory structure I already find confusing. It is not yet useful to me. Perhaps it may be once I understand more. The main thing is to case split between direct and indirect eval. They are completely different, indeed opposite in some ways, regarding the issues I'm trying to understand. Also, let's avoid mentioning the call stack as much as we can. Given

```js
const outerRealm = Realm.makeRootRealm();
const outerEval = outerRealm.global.eval;
const innerRealm = Realm.makeRootRealm();
const innerEval = Realm.makeRootRealm().global.eval;
```

For each of the following two cases, for the evaluation of the code '1+2', what are the evalRealm, the callerRealm, the calleeRealm, the current realm:

direct eval:
```js
outerEval('eval => eval("1 + 2;");')(innerEval);
```

indirect eval:
```js
outerEval('innerEval => innerEval("1 + 2;");')(innerEval);
```

mikesamuel: > direct eval:

So `outerEval('eval => eval("1 + 2;");'` performs an indirect eval to produces a function whose realm is outerRealm, right?  Let's call that *producedByEval*, so that code snippet could be rewritten as

```js
const producedByEval = outerEval('eval => eval("1 + 2;");');
producedByEval(innerEval);
```

From https://tc39.github.io/ecma262/#sec-function-calls-runtime-semantics-evaluation directEval only happens when
> a. If SameValue(func, %eval%) is true, then
>    1. [Rest of direct eval code]

Is there one %eval% intrinsic object shared by all realms?

If so, then you're correct that there is a direct eval.

If not, the call to that *producedByEval* does not perform a direct eval since outerRealm's %eval% is not the same as innerEval, right?

mikesamuel: Mark, maybe we should have this discussion on a different forum.

One thing, should

```js
const innerEval = Realm.makeRootRealm().global.eval;
```

be

```js
const innerEval = innerRealm.global.eval;
```

erights: > One thing, should [...] be [...]

Yes, good catch. Corrected in place above.

> Is there one %eval% intrinsic object shared by all realms?
>
> If so, then you're correct that there is a direct eval.
>
> If not, the call to that producedByEval does not perform a direct eval since outerRealm's %eval% is not the same as innerEval, right?

There is definitely an intrinsic per root realm. We can see that this must be the case since it is a mutable function object that inherits from its realm's `Function.prototype`.

However, the deeper question had not occurred to me, and I was making an unexamined assumption: That the runtime test that determines whether it is a direct eval only care that it is an %eval% intrinsic of some realm. Now that you raise it, we should examine both the spec language and what multi-realm implementations, including all browser implementations, actually do.

These test should *not* use the current Realms shim, as it must replace the binding of `eval` and so cannot support direct eval. This is one of the necessary differences be the shim and the proposed behavior.

erights: > Mark, maybe we should have this discussion on a different forum.

I'm happy to move. Where?

mikesamuel: > I'm happy to move. Where?

https://ocapjs.org/t/what-does-performeval-mean/88

allenwb:
@erights
 > That the runtime test that determines whether it is a direct eval only care that it is an %eval% intrinsic of some realm. 

That's not correct.  See [12.3.4.1](https://tc39.github.io/ecma262/#sec-function-calls-runtime-semantics-evaluation) step 6.a.  %eval% is a reference to the `eval` intrinsic of the _current_ realm. In other words the `eval` intrinsic of the realm that is lexically associated with the code containing the _CallExpression_ that is being evaluated.

littledan:
Maybe @domenic can help clarify, as I believe he wrote much of this specification text.

spec.html
```diff
@@ -23364,17 +23363,17 @@ <h1>eval ( _x_ )</h1>
...
-        1. Let _calleeRealm_ be the current Realm Record.
-        1. Perform ? HostEnsureCanCompileStrings(_callerRealm_, _calleeRealm_).
-        1. Return ? PerformEval(_x_, _calleeRealm_, *false*, *false*).
+        1. Return ? PerformEval(_x_, _callerRealm_, *false*, *false*).
       </emu-alg>
 
       <emu-clause id="sec-performeval" aoid="PerformEval">
-        <h1>Runtime Semantics: PerformEval ( _x_, _evalRealm_, _strictCaller_, _direct_ )</h1>
-        <p>The abstract operation PerformEval with arguments _x_, _evalRealm_, _strictCaller_, and _direct_ performs the following steps:</p>
+        <h1>Runtime Semantics: PerformEval ( _x_, _callerRealm_ _strictCaller_, _direct_ )</h1>
```
domenic: Missing comma

domenic:
```suggestion
        <h1>Runtime Semantics: PerformEval ( _x_, _callerRealm_, _strictCaller_, _direct_ )</h1>
```

mikesamuel: [Commit suggestion] didn't work, so did it manually.

Issue #1505, opened 2019-04-10
Lookup `resolve` property only once in PerformPromiseAll and PerformPromiseRace


gsathya:
See https://github.com/tc39/proposal-promise-allSettled/pull/40 for background

This would be a change similar to the one proposed with Promise.allSettled

bakkot:
This change seems consistent with other parts of the language: for example, the [Map constructor](https://tc39.github.io/ecma262/#sec-map-iterable) (and related constructors elsewhere) look up `"set"` exactly once, outside of the loop.

mathiasbynens:
It’s [also](https://github.com/tc39/proposal-promise-allSettled/pull/40#issuecomment-481126774) in the spirit of changes we’ve made to the iterator protocol, and then some: #976, #988, #1021, #1398, #1408.

ljharb:
I’d love to see it happen; do we have any data about web compatibility?

bakkot:
For this to break anyone, they would need to be replacing `Promise.resolve` with a not-semantically-equivalent version as a side effect of an iterator passed to `Promise.{all, race}`, right? That seems like the sort of thing which is unlikely to be happening in practice.

I would love web compat data, but this seems like one of those cases where we could decide to make the change without data and hope engines don't find they need to back it out (assuming of course that engines were on board).

ljharb:
If we’re all comfortable with that, then i am too :-) i agree it seems highly unlikely.

erights:
agreed

leobalter:
```js
var promise = new Promise(function() {});

var resolveCount = 0
var proxy = new Proxy(Promise, {
    get(_, name) {
        if (name === 'resolve') {
            resolveCount++;
        }
        return Promise[name];
    }
});

Promise.all.call(proxy, [promise, promise, promise]);

console.log(resolveCount); // should be 1 after this change, it's currently 3
```

If you prefer avoiding Proxies:

```js
var resolveCount = 0;
function NotPromise(executor) {
    return new Promise(executor);
}

Object.defineProperty(NotPromise, 'resolve', {
    get() {
        resolveCount++;
        return function(v) {
            return Promise.resolve(v);
        };
    }
});

Promise.all.call(NotPromise, [promise, promise, promise]);

console.log(resolveCount);
```

Here are some tests you can use for `Promise.all` and `Promise.allSettled`. `Promise.race` would need a few tweaks.

I support this change.

mathiasbynens:
[...]

PR #1506, opened 2019-04-11
Lookup constructor.resolve only once in PerformPromise{All, Race}

gsathya:
Fixes https://github.com/tc39/ecma262/issues/1505

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38347,6 +38347,8 @@ THH:mm:ss.sss
             1. Assert: _resultCapability_ is a PromiseCapability Record.
             1. Let _values_ be a new empty List.
             1. Let _remainingElementsCount_ be a new Record { [[Value]]: 1 }.
+            1. Let _promiseResolve_ be ? Get(_constructor_, `"resolve"`).
+            1. If ! IsCallable(_promiseResolve_) is *false*, throw a *TypeError* exception.
             1. Let _index_ be 0.
             1. Repeat,
               1. Let _next_ be IteratorStep(_iteratorRecord_).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -38265,6 +38265,7 @@ <h1>Runtime Semantics: PerformPromiseAll ( _iteratorRecord_, _constructor_, _res
             1. Assert: _resultCapability_ is a PromiseCapability Record.
             1. Let _values_ be a new empty List.
             1. Let _remainingElementsCount_ be a new Record { [[Value]]: 1 }.
+            1. Let _promiseResolve_ be ? GetMethod(_constructor_, `"resolve"`).
```
zenparsing: The result of `GetMethod` could be a function or `undefined`. In that case where it's `undefined`, do you want an error thrown here or in the `Call(_promiseResolve_ ...)` below?

gsathya: I don't have a preference, but in general I guess it's better to throw earlier rather than later.

Would it make sense to just use a `Get` and do the `IsCallable` check inline instead of using `GetMethod` in this case? Otherwise, we'd do a double check for `undefined` (once in the `GetMethod` call and again in this method).

gibson042: I very much want the former, because otherwise an expression like `(class extends Promise { static resolve(){ throw "NOPE" }}).all(iterable)` will behave differently based solely on the emptiness vs. non-emptiness of `iterable` without regard for the state of iterated values.

bakkot:
Early throwing also matches [Map etc](https://tc39.github.io/ecma262/#sec-add-entries-from-iterable), which check before iteration. (They do `Get` + `IsCallable` rather than `GetMethod`.)

gsathya:
> Early throwing also matches Map etc, which check before iteration. (They do Get + IsCallable rather than GetMethod.)

SGTM, I hear folks like consistency. PR updated.

gsathya:
This has reached consensus at the TC39 meeting today.

Issue #1354, opened 2018-11-20
Memory model bug for data-race free programs


syg:
@conrad-watt, in his work on adopting the JS memory model for wasm, has found a bug in the memory model that causes data-race free programs to be *not* sequentially consistent.

## The Bug

Consider the following example,

Example 1
```
           Agent 1                          Agent 2
           -------                          -------
[W1] Atomics.store(x, 0, 1);  ||  [W2] Atomics.store(x, 0, 2);
[Wy] Atomics.store(y, 0, 1);  ||  [Ry] if (Atomics.load(y, 0) === 1) {
                              ||  [R1]   print(x[0]);
                              ||  [R2]   print(x[0]);
                              ||       }
```

The current memory model admits an execution where `2,1` is printed. Let *sw*, *hb*, and *mo* abbreviate synchronizes-with, happens-before, and memory-order respectively.

[Wy] <*sw* [Ry], therefore [Wy] <*hb* [Ry]
[W1] <*hb* [Wy]
[W2] <*hb* [Ry] <*hb* [R1] <*hb* [R2]
[W1] <*mo* [Wy] <*mo* [W2] <*mo* [Ry] <*mo* [R1] <*mo* [R2]

[R1] reads-from [W2]
[R2] reads-from [W1]

Note that there is no data race because [W1] and [W2], while racing, are not a *data* race since both are SC atomic.

The above is a valid but non-SC execution because "Coherent Reads" only says that a read must read the most recent write event on some range with respect to happens-before. So while there is a demonstrable total memory order as written above, it is still valid for [R2] to read from a memory-order-older [W1].

## Strong Fix

It is natural to try to fix the above by disallowing reads of stale memory-order write events, which is a stronger condition than only disallowing reads of stale happens-before write events.

This fix might be too strong, however. Consider a related example where W1 and W2 are changed to non-atomic writes.

Example 2
```
           Agent 1                          Agent 2
           -------                          -------
[W1'] x[0] = 1;                ||  [W2'] x[0] = 2;
[Wy ] Atomics.store(y, 0, 1);  ||  [Ry ] if (Atomics.load(y, 0) === 1) {
                               ||  [R1 ]   print(x[0]);
                               ||  [R2 ]   print(x[0]);
                               ||       }
```

Under the strong fix, there is no valid execution where `2,1` could be printed. That is, both Examples 1 and 2 cannot print `2,1`.

The **pros** are:
- Easier-to-reason-about meaning of SC atomics as full SC fences.
[...]

PR #1511, opened 2019-04-17
Normative: fix axiomatic model to be SC-DRF, and allow ARMv8 compilation

conrad-watt:
(PR submitted following co-ordination with @syg)

### Background

In the November TC39 meeting, [consensus was reached](http://tc39.github.io/tc39-notes/2018-11_nov-28.html#memory-model-bug-drf-sc-bug) to strengthen the axiomatic model to provide Sequential Consistency of Data-Race-Free Programs (SC-DRF). This property was [always intended to be provided by the model](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-data-race-freedom) (but see https://github.com/tc39/ecma262/issues/1354 for a counter-example).

A PR to amend the model was drafted (https://github.com/tc39/ecma262/pull/1362), but I asked for a [delay in landing it](https://github.com/tc39/ecma262/pull/1362#issuecomment-443306833) after @stedolan noticed that the JS model (independent of our fix) appeared to be too strong to allow compilation of SC atomics to ARMv8 lda/stl instructions. We've confirmed that this compilation scheme is used "in the wild" in Chrome ([introducing commit](https://codereview.chromium.org/2760963002)), and the violation can be observed experimentally in multi-threaded WebAssembly programs running end-to-end in Chrome on ARMv8 machines. Other engines are also discussing/in the process of implementing this compilation scheme.

In the January TC39 meeting, [consensus was reached](http://tc39.github.io/tc39-notes/2019-01_jan-30.html#amending-the-memory-model-to-support-armv8-ldastl-codegen) to amend the axiomatic model to support compilation from JS SC atomics to ARMv8 lda/stl instructions. To be clear, this makes the model _weaker_ than before, since it previously provided a guarantee on interleaving (racy) non-atomics with SC atomics that ARMv8 does not (see [slides](https://docs.google.com/presentation/d/1qif7z-Y8C-nvJM20UNJQzAKJgLN4wmXS_5NN2Wgipb4/) for a counter-example).

### This PR

This PR incorporates both fixes as normative changes, based on previously proposed wording by @syg for the SC-DRF fix (https://github.com/tc39/ecma262/pull/1362). The fixes are split imprecisely between two commits so the change that weakens the model for the ARMv8 fix can be seen in isolation. That being said, both commits should be taken as a package.

Note that the SC-DRF fix has changed slightly from the precise formulation of the original PR, but the "weak fix" philosophy remains the same. In fact, the new version is a slightly weaker strengthening (I shouldn't have called the previous version "weakest"), and makes one of the SC-DRF conditions more closely mirror an analogous condition in the C++ model.

In support of our changes, we have some proof work/verification that I'd be happy to talk about in more detail over email, including a Coq proof done by @anlun. These proofs don't handle every feature of the memory model (really every proof in this field has a caveat attached), but we've verified at least that the previous model was incorrect, and the revised model isn't incorrect in the same way.

We also observed that one of the SC-DRF additions makes a clause about synchronizing **init** events redundant, and therefore include a separate editorial change to remove it. **init** events were originally included in _synchronizes-with_ so that the following shape is not permitted, where R_sc _reads-from_ W_init and W_sc and R_sc have equal ranges

- W_init _memory-order_ W_sc
- W_sc _memory-order_ R_sc

However, it is guaranteed that W_init _happens-before_ W_sc and W_init _happens-before_ R_sc, and our SC-DRF fix disallows exactly this shape anyway. The third bullet disallows the following shape (for all consistencies of W), where R_sc _reads-from_ W and W_sc and R_sc have equal ranges

- W _happens-before_ W_sc
- W _happens-before_ R_sc
- W_sc _memory-order_ R_sc

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -39418,6 +39418,10 @@ THH:mm:ss.sss
       </emu-note>
 
       <emu-note>
+        <p>`"Init"` events do not participate in synchronizes-with, and are instead constrained directly by happens-before.</p>
+      </emu-note>
+
+      <emu-note>
         <p>Not all `"SeqCst"` events related by reads-from are related by synchronizes-with. Only events that also have equal ranges are related by synchronizes-with.</p>
       </emu-note>
 
--- a/spec.html
+++ b/spec.html
@@ -39511,7 +39511,12 @@ THH:mm:ss.sss
       <ul>
         <li>For each pair (_E_, _D_) in _execution_.[[HappensBefore]], (_E_, _D_) is in memory-order.</li>
         <li>
-          <p>For each pair (_E_, _D_) in _execution_.[[SynchronizesWith]], (_E_, _D_) is in memory-order if there is no WriteSharedMemory or ReadModifyWriteSharedMemory event _W_ in SharedDataBlockEventSet(_execution_) with equal range as _D_ such that _W_ is not _E_, and the pairs (_E_, _W_) and (_W_, _D_) are in memory-order.</p>
+          <p>For each pair (_R_, _W_) in _execution_.[[ReadsFrom]], there is no WriteSharedMemory or ReadModifyWriteSharedMemory event _V_ in SharedDataBlockEventSet(_execution_) such that the pairs (_W_, _V_) and (_V_, _R_) are in memory-order, and any of the following conditions are true.</p>
+          <ul>
+            <li>The pair (_W_, _R_) is in _execution_.[[SynchronizesWith]], and _V_ and _R_ have equal ranges.</li>
+            <li>The pairs (_W_, _R_) and (_V_, _R_) are in _execution_.[[HappensBefore]], _W_.[[Order]] is `"SeqCst"`, and _W_ and _V_ have equal ranges.</li>
+            <li>The pairs (_W_, _R_) and (_W_, _V_) are in _execution_.[[HappensBefore]], _R_.[[Order]] is `"SeqCst"`, and _V_ and _R_ have equal ranges.</li>
+          </ul>
           <emu-note>
             <p>This clause additionally constrains `"SeqCst"` events on equal ranges.</p>
           </emu-note>
--- a/spec.html
+++ b/spec.html
@@ -39511,7 +39511,7 @@ THH:mm:ss.sss
       <ul>
         <li>For each pair (_E_, _D_) in _execution_.[[HappensBefore]], (_E_, _D_) is in memory-order.</li>
         <li>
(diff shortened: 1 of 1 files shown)


ljharb:
@conrad-watt does/can this have test262 tests?

conrad-watt:
I can write tests which embody the difference between the memory models, but they would work very differently from (my understanding of) a traditional test262 test.

Under perfect conditions, using WebAssembly to "force" efficient code generation, I've been able to make a test which shows the ARMv8 behaviour roughly 1 in 100,000 times. There was some previous discussion [here](https://github.com/tc39/ecma262/pull/1127#issuecomment-376281035) about adding tests like this to a test262 addendum, but it's not gone anywhere so far. The model as a whole is not being validated currently.

Another factor might be that, the majority of the time, if an implementation is doing something the model doesn't allow, it's the model that should change, not the implementation. That's the case with all of these changes.

syg:
Yeah, I'm still pretty stumped on how to check in litmus tests. Litmus tests are qualitatively different from unit tests -- we can certainly put them **into** test262 for now, but running them along with the other ones doesn't really make sense.

spec.html
```diff
@@ -39501,7 +39495,12 @@ <h1>Sequentially Consistent Atomics</h1>
       <ul>
         <li>For each pair (_E_, _D_) in _execution_.[[HappensBefore]], (_E_, _D_) is in memory-order.</li>
         <li>
-          <p>For each pair (_E_, _D_) in _execution_.[[SynchronizesWith]], (_E_, _D_) is in memory-order if there is no WriteSharedMemory or ReadModifyWriteSharedMemory event _W_ in SharedDataBlockEventSet(_execution_) with equal range as _D_ such that _W_ is not _E_, and the pairs (_E_, _W_) and (_W_, _D_) are in memory-order.</p>
+          <p>For each pair (_R_, _W_) in _execution_.[[ReadsFrom]], there is no WriteSharedMemory or ReadModifyWriteSharedMemory event _V_ in SharedDataBlockEventSet(_execution_) such that _V_.[[Order]] is Seqcst, the pairs (_W_, _V_) and (_V_, _R_) are in memory-order, and any of the following conditions are true.</p>
```
syg: Formatting: Seqcst -> `` `"SeqCst"` ``

conrad-watt: done in rebase

conrad-watt:
> My only suggestion is a non-normative note that the happens-before subclauses are now what constrains `"Init"` events would be helpful.

done

PR #1513, opened 2019-04-23
Editorial: replace `lookahead ∉ { ele }` with `lookahead ≠ ele`

bakkot:
This is explicitly permitted, and I think clearer. Mostly because I just misread the right-hand side of `lookahead ∉ { let [ }` as a two-element set.


syg:
Hear hear.

PR #1515, opened 2019-04-23
Normative: Add new BigInt primitive

caiolima:
This PR is include changes to integrate [tc39/proposal-bigint](https://github.com/tc39/proposal-bigint) into ECMA262 text. It contains changes on:

1. Grammar to support BigInt literals.
2. Abstract Operations to consider this new primitive.
3. Semantics of arithmetic expressions to support BigInts.
4. The specification of BigInt objects, including the constructor and BigInt.prototype.
5. Including `BigInt64Array` and `BigUint64Array` as new types of `TypedArray`.
6. Changes on `DataView` operations to support new types.

Proposal author: Daniel Ehrenberg and Brendan Eich

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1110,30 +1110,898 @@
       </emu-clause>
     </emu-clause>
 
-    <emu-clause id="sec-ecmascript-language-types-number-type">
-      <h1>The Number Type</h1>
-      <p>The Number type has exactly 18437736874454810627<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>64<sub>ℝ</sub></sup> - 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> + 3<sub>ℝ</sub></emu-eqn>) values, representing the double-precision 64-bit format IEEE 754-2008 values as specified in the IEEE Standard for Binary Floating-Point Arithmetic, except that the 9007199254740990<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> - 2<sub>ℝ</sub></emu-eqn>) distinct &ldquo;Not-a-Number&rdquo; values of the IEEE Standard are represented in ECMAScript as a single special *NaN* value. (Note that the *NaN* value is produced by the program expression `NaN`.) In some implementations, external code might be able to detect a difference between various Not-a-Number values, but such behaviour is implementation-dependent; to ECMAScript code, all *NaN* values are indistinguishable from each other.</p>
+    <emu-clause id="sec-numeric-types">
+      <h1>Numeric Types</h1>
+      <p>ECMAScript has two built-in numeric types: Number and BigInt. In this specification, every numeric type _T_ contains a multiplicative identity value denoted _T_::unit. The specification types also have the following abstract operations, likewise denoted _T_::<i>op</i> for a given operation with specification name <i>op</i>. All argument types are _T_. The "Result" column shows the return type, along with an indication if it is possible for some invocations of the operation to return an abrupt completion.</p>
+      <emu-table id="table-numeric-type-ops" caption="Numeric Type Operations">
+        <table>
+          <tbody>
+          <tr>
+            <th>
+              Invocation Synopsis
+            </th>
+            <th>
+              Example source
+            </th>
+            <th>
+              Invoked by the Evaluation semantics of ...
+            </th>
+            <th>
+              Result
+            </th>
+          </tr>
+          <tr>
+            <td>
+              _T_::unaryMinus(x)
+            </td>
+            <td>
+              `-x`
+            </td>
+            <td>
+              <emu-xref href="#sec-unary-minus-operator" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::bitwiseNOT(x)
+            </td>
+            <td>
+              `~x`
+            </td>
+            <td>
+              <emu-xref href="#sec-bitwise-not-operator" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::exponentiate(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;**&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-exp-operator" title></emu-xref>
+              and <emu-xref href="#sec-math.pow" title></emu-xref>
+            </td>
+            <td>
+              _T_, may throw *RangeError*
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::multiply(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;*&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-multiplicative-operators" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::divide(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;/&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-multiplicative-operators" title></emu-xref>
+            </td>
+            <td>
+              _T_, may throw *RangeError*
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::remainder(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;%&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-multiplicative-operators" title></emu-xref>
+            </td>
+            <td>
+              _T_, may throw *RangeError*
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::add(x,&nbsp;y)
+            </td>
+            <td>
+              `x ++`<br>`++ x`<br>`x&nbsp;+&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-postfix-increment-operator" title></emu-xref>,
+              <emu-xref href="#sec-prefix-increment-operator" title></emu-xref>,
+              and <emu-xref href="#sec-addition-operator-plus" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::subtract(x,&nbsp;y)
+            </td>
+            <td>
+              `x --`<br>`-- x`<br>`x&nbsp;-&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-postfix-decrement-operator" title></emu-xref>,
+              <emu-xref href="#sec-prefix-decrement-operator" title></emu-xref>,
+              and <emu-xref href="#sec-subtraction-operator-minus" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::leftShift(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;&lt;&lt;&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-left-shift-operator" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::signedRightShift(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;>>&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-signed-right-shift-operator" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::unsignedRightShift(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;>>>&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-unsigned-right-shift-operator" title></emu-xref>
+            </td>
+            <td>
+              _T_, may throw *TypeError*
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::lessThan(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;&lt;&nbsp;y`<br>`x&nbsp;>&nbsp;y`<br>`x&nbsp;&lt;=&nbsp;y`<br>`x&nbsp;>=&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-relational-operators" title></emu-xref>,
+              via <emu-xref href="#sec-abstract-relational-comparison" title></emu-xref>
+            </td>
+            <td>
+              Boolean or *undefined* (for unordered inputs)
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::equal(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;==&nbsp;y`<br>`x&nbsp;!=&nbsp;y`<br>`x&nbsp;===&nbsp;y`<br>`x&nbsp;!==&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-equality-operators" title></emu-xref>,
+              via <emu-xref href="#sec-strict-equality-comparison" title></emu-xref>
+            </td>
+            <td>
+              Boolean
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::sameValue(x,&nbsp;y)
+            </td>
+            <td>
+            </td>
+            <td>
+              Object internal methods,
+              via <emu-xref href="#sec-samevalue" title></emu-xref>,
+              to test exact value equality
+            </td>
+            <td>
+              Boolean
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::sameValueZero(x,&nbsp;y)
+            </td>
+            <td>
+            </td>
+            <td>
+              Array, Map, and Set methods,
+              via <emu-xref href="#sec-samevaluezero" title></emu-xref>,
+              to test value equality ignoring differences among members of the zero cohort (e.g., *-0* and *+0*)
+            </td>
+            <td>
+              Boolean
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::bitwiseAND(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;&amp;&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-binary-bitwise-operators" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::bitwiseXOR(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;^&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-binary-bitwise-operators" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::bitwiseOR(x,&nbsp;y)
+            </td>
+            <td>
+              `x&nbsp;|&nbsp;y`
+            </td>
+            <td>
+              <emu-xref href="#sec-binary-bitwise-operators" title></emu-xref>
+            </td>
+            <td>
+              _T_
+            </td>
+          </tr>
+          <tr>
+            <td>
+              _T_::toString(x)
+            </td>
+            <td>
+              `String(x)`
+            </td>
+            <td>
+              Many expressions and built-in functions, via <emu-xref href="#sec-tostring" title></emu-xref>
+            </td>
+            <td>
+              String
+            </td>
+
+          </tr>
+          </tbody>
+        </table>
+      </emu-table>
+      <p>The _T_::unit value and _T_::_op_ operations are not a part of the ECMAScript language; they are defined here solely to aid the specification of the semantics of the ECMAScript language. Other abstract operations are defined throughout this specification.</p>
+      <p>Because the numeric types are in general not convertible without loss of precision or truncation, the ECMAScript language provides no implicit conversion among these types. Programmers must explicitly call `Number` and `BigInt` functions to convert among types when calling a function which requires another type.</p>
       <emu-note>
-        <p>The bit pattern that might be observed in an ArrayBuffer (see <emu-xref href="#sec-arraybuffer-objects"></emu-xref>) or a SharedArrayBuffer (see <emu-xref href="#sec-sharedarraybuffer-objects"></emu-xref>) after a Number value has been stored into it is not necessarily the same as the internal representation of that Number value used by the ECMAScript implementation.</p>
+        <p>The first and subsequent editions of ECMAScript have provided, for certain operators, implicit numeric conversions that could lose precision or truncate. These legacy implicit conversions are maintained for backward compatibility, but not provided for BigInt in order to minimize opportunity for programmer error, and to leave open the option of generalized <em>value types</em> in a future edition.</p>
       </emu-note>
-      <p>There are two other special values, called *positive Infinity* and *negative Infinity*. For brevity, these values are also referred to for expository purposes by the symbols *+&infin;* and *-&infin;*, respectively. (Note that these two infinite Number values are produced by the program expressions `+Infinity` (or simply `Infinity`) and `-Infinity`.)</p>
-      <p>The other 18437736874454810624<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>64<sub>ℝ</sub></sup> - 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup></emu-eqn>) values are called the finite numbers. Half of these are positive numbers and half are negative numbers; for every finite positive Number value there is a corresponding negative value having the same magnitude.</p>
-      <p>Note that there is both a *positive zero* and a *negative zero*. For brevity, these values are also referred to for expository purposes by the symbols *+0* and *-0*, respectively. (Note that these two different zero Number values are produced by the program expressions `+0` (or simply `0`) and `-0`.)</p>
-      <p>The 18437736874454810622<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>64<sub>ℝ</sub></sup> - 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> - 2<sub>ℝ</sub></emu-eqn>) finite nonzero values are of two kinds:</p>
-      <p>18428729675200069632<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>64<sub>ℝ</sub></sup> - 2<sub>ℝ</sub><sup>54<sub>ℝ</sub></sup></emu-eqn>) of them are normalized, having the form</p>
-      <div class="math-display">
-        _s_ &times; _m_ &times; 2<sup>_e_</sup>
-      </div>
-      <p>where _s_ is +1<sub>ℝ</sub> or -1<sub>ℝ</sub>, _m_ is a positive mathematical integer less than 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> but not less than 2<sub>ℝ</sub><sup>52<sub>ℝ</sub></sup>, and _e_ is a mathematical integer ranging from -1074<sub>ℝ</sub> to 971<sub>ℝ</sub>, inclusive.</p>
-      <p>The remaining 9007199254740990<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> - 2<sub>ℝ</sub></emu-eqn>) values are denormalized, having the form</p>
-      <div class="math-display">
-        _s_ &times; _m_ &times; 2<sup>_e_</sup>
-      </div>
-      <p>where _s_ is +1<sub>ℝ</sub> or -1<sub>ℝ</sub>, _m_ is a positive mathematical integer less than 2<sub>ℝ</sub><sup>52<sub>ℝ</sub></sup>, and _e_ is -1074<sub>ℝ</sub>.</p>
-      <p>Note that all the positive and negative mathematical integers whose magnitude is no greater than 2<sup>53</sup> are representable in the Number type (indeed, the mathematical integer 0 has two representations, *+0* and *-0*).</p>
-      <p>A finite number has an <em>odd significand</em> if it is nonzero and the mathematical integer _m_ used to express it (in one of the two forms shown above) is odd. Otherwise, it has an <em>even significand</em>.</p>
-      <p>In this specification, the phrase &ldquo;the <dfn id="number-value">Number value</dfn> for _x_&rdquo; where _x_ represents an exact real mathematical quantity (which might even be an irrational number such as &pi;) means a Number value chosen in the following manner. Consider the set of all finite values of the Number type, with *-0* removed and with two additional values added to it that are not representable in the Number type, namely 2<sub>ℝ</sub><sup>1024<sub>ℝ</sub></sup> (which is <emu-eqn>+1<sub>ℝ</sub> &times; 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> &times; 2<sub>ℝ</sub><sup>971<sub>ℝ</sub></sup></emu-eqn>) and <emu-eqn>-2<sub>ℝ</sub><sup>1024<sub>ℝ</sub></sup></emu-eqn> (which is <emu-eqn>-1<sub>ℝ</sub> &times; 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> &times; 2<sub>ℝ</sub><sup>971<sub>ℝ</sub></sup></emu-eqn>). Choose the member of this set that is closest in value to _x_. If two values of the set are equally close, then the one with an even significand is chosen; for this purpose, the two extra values 2<sub>ℝ</sub><sup>1024<sub>ℝ</sub></sup> and <emu-eqn>-2<sub>ℝ</sub><sup>1024<sub>ℝ</sub></sup></emu-eqn> are considered to have even significands. Finally, if 2<sub>ℝ</sub><sup>1024<sub>ℝ</sub></sup> was chosen, replace it with *+&infin;*; if <emu-eqn>-2<sub>ℝ</sub><sup>1024<sub>ℝ</sub></sup></emu-eqn> was chosen, replace it with *-&infin;*; if *+0* was chosen, replace it with *-0* if and only if _x_ is less than zero; any other chosen value is used unchanged. The result is the Number value for _x_. (This procedure corresponds exactly to the behaviour of the IEEE 754-2008 roundTiesToEven mode.)</p>
-      <p>Some ECMAScript operators deal only with integers in specific ranges such as <emu-eqn>-2<sup>31</sup></emu-eqn> through <emu-eqn>2<sup>31</sup> - 1</emu-eqn>, inclusive, or in the range 0 through <emu-eqn>2<sup>16</sup> - 1</emu-eqn>, inclusive. These operators accept any value of the Number type but first convert each such value to an integer value in the expected range. See the descriptions of the numeric conversion operations in <emu-xref href="#sec-type-conversion"></emu-xref>.</p>
+
+      <emu-clause id="sec-ecmascript-language-types-number-type">
+        <h1>The Number Type</h1>
+        <p>The Number type has exactly 18437736874454810627<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>64<sub>ℝ</sub></sup> - 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> + 3<sub>ℝ</sub></emu-eqn>) values, representing the double-precision 64-bit format IEEE 754-2008 values as specified in the IEEE Standard for Binary Floating-Point Arithmetic, except that the 9007199254740990<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> - 2<sub>ℝ</sub></emu-eqn>) distinct &ldquo;Not-a-Number&rdquo; values of the IEEE Standard are represented in ECMAScript as a single special *NaN* value. (Note that the *NaN* value is produced by the program expression `NaN`.) In some implementations, external code might be able to detect a difference between various Not-a-Number values, but such behaviour is implementation-dependent; to ECMAScript code, all *NaN* values are indistinguishable from each other.</p>
+        <emu-note>
+          <p>The bit pattern that might be observed in an ArrayBuffer (see <emu-xref href="#sec-arraybuffer-objects"></emu-xref>) or a SharedArrayBuffer (see <emu-xref href="#sec-sharedarraybuffer-objects"></emu-xref>) after a Number value has been stored into it is not necessarily the same as the internal representation of that Number value used by the ECMAScript implementation.</p>
+        </emu-note>
+        <p>There are two other special values, called *positive Infinity* and *negative Infinity*. For brevity, these values are also referred to for expository purposes by the symbols *+&infin;* and *-&infin;*, respectively. (Note that these two infinite Number values are produced by the program expressions `+Infinity` (or simply `Infinity`) and `-Infinity`.)</p>
+        <p>The other 18437736874454810624<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>64<sub>ℝ</sub></sup> - 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup></emu-eqn>) values are called the finite numbers. Half of these are positive numbers and half are negative numbers; for every finite positive Number value there is a corresponding negative value having the same magnitude.</p>
+        <p>Note that there is both a *positive zero* and a *negative zero*. For brevity, these values are also referred to for expository purposes by the symbols *+0* and *-0*, respectively. (Note that these two different zero Number values are produced by the program expressions `+0` (or simply `0`) and `-0`.)</p>
+        <p>The 18437736874454810622<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>64<sub>ℝ</sub></sup> - 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> - 2<sub>ℝ</sub></emu-eqn>) finite nonzero values are of two kinds:</p>
+        <p>18428729675200069632<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>64<sub>ℝ</sub></sup> - 2<sub>ℝ</sub><sup>54<sub>ℝ</sub></sup></emu-eqn>) of them are normalized, having the form</p>
+        <div class="math-display">
+          _s_ &times; _m_ &times; 2<sup>_e_</sup>
+        </div>
+        <p>where _s_ is +1<sub>ℝ</sub> or -1<sub>ℝ</sub>, _m_ is a positive mathematical integer less than 2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> but not less than 2<sub>ℝ</sub><sup>52<sub>ℝ</sub></sup>, and _e_ is a mathematical integer ranging from -1074<sub>ℝ</sub> to 971<sub>ℝ</sub>, inclusive.</p>
+        <p>The remaining 9007199254740990<sub>ℝ</sub> (that is, <emu-eqn>2<sub>ℝ</sub><sup>53<sub>ℝ</sub></sup> - 2<sub>ℝ</sub></emu-eqn>) values are denormalized, having the form</p>
+        <div class="math-display">
+          _s_ &times; _m_ &times; 2<sup>_e_</sup>
+        </div>
+        <p>where _s_ is +1<sub>ℝ</sub> or -1<sub>ℝ</sub>, _m_ is a positive mathematical integer less than 2<sub>ℝ</sub><sup>52<sub>ℝ</sub></sup>, and _e_ is -1074<sub>ℝ</sub>.</p>
+        <p>Note that all the positive and negative mathematical integers whose magnitude is no greater than 2<sup>53</sup> are representable in the Number type (indeed, the mathematical integer 0 has two representations, *+0* and *-0*).</p>
+        <p>A finite number has an <em>odd significand</em> if it is nonzero and the mathematical integer _m_ used to express it (in one of the two forms shown above) is odd. Otherwise, it has an <em>even significand</em>.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -1095,30 +1095,540 @@ <h1>Well-Known Symbols</h1>
...
+            1. If _x_ is *NaN*, return *NaN*.
+            1. Return the result of negating _x_; that is, compute a Number with the same magnitude but opposite sign.
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-numeric-types-number-bitwiseNOT">
+          <h1>Number::bitwiseNOT (_x_)</h1>
+
+          <emu-alg>
+            1. Let _oldValue_ be ? ToInt32(? _x_).
```
littledan: I think the inner-most `?` is unnecessary here. `_x_` will always be a Number and not an abrupt completion.

annevk:
Should this land after #1135?

littledan:
@annevk Yes.

spec.html
```diff
@@ -25877,6 +26626,144 @@ <h1>Properties of Number Instances</h1>
...
+      <emu-clause id="sec-bigint.prototype">
+        <h1>BigInt.prototype</h1>
+        <p>The initial value of `BigInt.prototype` is the intrinsic object %BigIntPrototype%.</p>
+        <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-properties-of-the-bigint-prototype-object">
+      <h1>Properties of the BigInt Prototype Object</h1>
+      <p>The BigInt prototype object is the intrinsic object <dfn>%BigIntPrototype%</dfn>. The BigInt prototype object is an ordinary object. The BigInt prototype is not a BigInt object; it does not have a [[BigIntData]] internal slot.</p>
```
littledan: Other parts of the specification have moved to a bullet-point format, e.g., see https://tc39.github.io/ecma262/#sec-properties-of-the-number-prototype-object . We should follow that for BigInt as well.

spec.html
```diff
@@ -25877,6 +26626,144 @@ <h1>Properties of Number Instances</h1>
     </emu-clause>
   </emu-clause>
 
+  <emu-clause id="sec-bigint-objects">
+  <h1>BigInt Objects</h1>
+    <emu-clause id="sec-bigint-constructor">
+      <h1>The BigInt Constructor</h1>
```
littledan: Example of a new format to follow: https://tc39.github.io/ecma262/#sec-number-constructor

spec.html
```diff
@@ -1095,30 +1095,540 @@ <h1>Well-Known Symbols</h1>
...
+          <emu-alg>
+            1. Return the result of <emu-xref href="#sec-applying-the-exp-operator" title>Applying the ** operator</emu-xref> with _base_ and _exponent_ as specified in <emu-xref href="#sec-applying-the-exp-operator"></emu-xref>.
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-numeric-types-number-multiply">
+          <h1>Number::multiply (_x_, _y_)</h1>
+
+          <emu-alg>
+            1. Return the result of <emu-xref href="#sec-applying-the-mul-operator" title>Applying the * operator</emu-xref> to _x_ and _y_ as specified in <emu-xref href="#sec-applying-the-mul-operator"></emu-xref>.
```
littledan: These "Applying the ... operator" sections should move here and become the body of these Number:: algorithms.

jmdyck:
I've got a bunch of suggested changes. I think it might be best to submit them as a PR against caiolima:big-int.

littledan:
@jmdyck That's great to hear. Looking forward to seeing your PR.

jmdyck:
Done: https://github.com/caiolima/ecma262/pull/3

spec.html
caiolima: Changed.

spec.html
caiolima: Done.

caiolima:
@littledan FYI, there are new changes from https://github.com/caiolima/ecma262/pull/4.

ljharb:
@caiolima i've merged #1135; can you rebase this?

caiolima:
> @caiolima i've merged #1135; can you rebase this?

@ljharb done.

spec.html
```diff
@@ -3179,7 +4004,7 @@ <h1>IsStrictReference ( _V_ )</h1>
         <h1>HasPrimitiveBase ( _V_ )</h1>
         <emu-alg>
           1. Assert: Type(_V_) is Reference.
-          1. If Type(_V_'s base value component) is Boolean, String, Symbol, or Number, return *true*; otherwise return *false*.
+          1. If Type(_V_'s base value component) is Boolean, String, Symbol, BigInt or Number, return *true*; otherwise return *false*.
```
```suggestion
          1. If Type(_V_'s base value component) is Boolean, String, Symbol, BigInt, or Number, return *true*; otherwise return *false*.
```

spec.html
```diff
@@ -3600,6 +4434,16 @@ <h1>ToBoolean ( _argument_ )</h1>
       </emu-table>
     </emu-clause>
 
+    <emu-clause id="sec-tonumeric" aoid="ToNumeric">
+      <h1>ToNumeric ( _value_ )</h1>
+      <p>The abstract operation ToNumeric returns _value_ converted to a numeric value of type Number or BigInt. This abstract operation functions as follows:</p>
+      <emu-alg>
+        1. Let _primValue_ be ? ToPrimitive(_value_, hint Number).
+        1. If Type(_primValue_) is BigInt, return _primValue_.
+        1. Return ToNumber(_primValue_).
```
```suggestion
        1. Return ? ToNumber(_primValue_).
```

spec.html
```diff
@@ -4411,24 +5407,20 @@ <h1>SameValueZero ( _x_, _y_ )</h1>
...
-        1. If Type(_x_) is Number, then
-          1. If _x_ is *NaN* and _y_ is *NaN*, return *true*.
-          1. If _x_ is *+0* and _y_ is *-0*, return *true*.
-          1. If _x_ is *-0* and _y_ is *+0*, return *true*.
-          1. If _x_ is the same Number value as _y_, return *true*.
-          1. Return *false*.
-        1. Return SameValueNonNumber(_x_, _y_).
+        1. If Type(_x_) is Number or BigInt, then
+          1. Return ! Type(_x_)::sameValueZero(_x_, _y_).
+        1. Return SameValueNonNumeric(_x_, _y_).
```
```suggestion
        1. Return ! SameValueNonNumeric(_x_, _y_).
```

spec.html
```diff
@@ -4461,19 +5453,22 @@ <h1>Abstract Relational Comparison</h1>
...
-          1. If _nx_ and _ny_ are the same Number value, return *false*.
-          1. If _nx_ is *+0* and _ny_ is *-0*, return *false*.
-          1. If _nx_ is *-0* and _ny_ is *+0*, return *false*.
-          1. If _nx_ is *+&infin;*, return *false*.
-          1. If _ny_ is *+&infin;*, return *true*.
-          1. If _ny_ is *-&infin;*, return *false*.
-          1. If _nx_ is *-&infin;*, return *true*.
-          1. If the mathematical value of _nx_ is less than the mathematical value of _ny_&mdash;note that these mathematical values are both finite and not both zero&mdash;return *true*. Otherwise, return *false*.
+          1. If Type(_px_) is BigInt and Type(_py_) is String, then
+            1. Let _ny_ be StringToBigInt(_py_).
```
```suggestion
            1. Let _ny_ be ! StringToBigInt(_py_).
```

spec.html
```diff
@@ -4461,19 +5453,22 @@ <h1>Abstract Relational Comparison</h1>
...
+            1. If _ny_ is *NaN*, return *undefined*.
+            1. Return BigInt::lessThan(_px_, _ny_).
+          1. If Type(_px_) is String and Type(_py_) is BigInt, then
+            1. Let _nx_ be StringToBigInt(_px_).
```
```suggestion
            1. Let _nx_ be ! StringToBigInt(_px_).
```

spec.html
```diff
@@ -3955,6 +4810,118 @@ <h1>ToUint8Clamp ( _argument_ )</h1>
...
+                Throw a *TypeError* exception.
+              </td>
+            </tr>
+            <tr>
+              <td>
+                String
+              </td>
+              <td>
+                <emu-alg>
+                  1. Let _n_ be StringToBigInt(_prim_).
```
```suggestion
                  1. Let _n_ be ! StringToBigInt(_prim_).
```

spec.html
```diff
@@ -3955,6 +4810,118 @@ <h1>ToUint8Clamp ( _argument_ )</h1>
...
+                  1. If _n_ is *NaN*, throw a *SyntaxError* exception.
```
ljharb: Should this use `SameValue(_n_, *NaN*)`? i don't feel strongly either way.

spec.html
```diff
@@ -4096,6 +5071,15 @@ <h1>NumberToString ( _m_ )</h1>
             <a href="http://netlib.sandia.gov/fp/g_fmt.c">http://netlib.sandia.gov/fp/g_fmt.c</a> and may also be found at the various `netlib` mirror sites.</p>
         </emu-note>
       </emu-clause>
+
+      <emu-clause id="sec-biginttostring" aoid="BigIntToString">
+        <h1>BigIntToString ( _i_ )</h1>
+        <p>The abstract operation BigIntToString converts a BigInt _i_ to String format as follows:</p>
+        <emu-alg>
+          1. If _i_ is less than zero, return the string-concatenation of the String `"-"` and ToString(-_i_).
```
ljharb: should `ToString` here be `BigInt::ToString` or `BigIntToString`?

spec.html
```diff
@@ -4393,13 +5393,9 @@ <h1>SameValue ( _x_, _y_ )</h1>
...
-        1. If Type(_x_) is Number, then
-          1. If _x_ is *NaN* and _y_ is *NaN*, return *true*.
-          1. If _x_ is *+0* and _y_ is *-0*, return *false*.
-          1. If _x_ is *-0* and _y_ is *+0*, return *false*.
-          1. If _x_ is the same Number value as _y_, return *true*.
-          1. Return *false*.
-        1. Return SameValueNonNumber(_x_, _y_).
+        1. If Type(_x_) is Number or BigInt, then
+          1. Return ! Type(_x_)::sameValue(_x_, _y_).
+        1. Return SameValueNonNumeric(_x_, _y_).
```
```suggestion
        1. Return ! SameValueNonNumeric(_x_, _y_).
```

spec.html
```diff
@@ -13504,9 +14433,11 @@ <h1>Runtime Semantics: Evaluation</h1>
...
             1. Let _rstr_ be ? ToString(_rprim_).
             1. Return the string-concatenation of _lstr_ and _rstr_.
-          1. Let _lnum_ be ? ToNumber(_lprim_).
-          1. Let _rnum_ be ? ToNumber(_rprim_).
-          1. Return the result of applying the addition operation to _lnum_ and _rnum_. See the Note below <emu-xref href="#sec-applying-the-additive-operators-to-numbers"></emu-xref>.
+          1. Let _lnum_ be ? ToNumeric(_lprim_).
+          1. Let _rnum_ be ? ToNumeric(_rprim_).
+          1. If Type(_lnum_) is different from Type(_rnum_), throw a *TypeError* exception.
+          1. Let _T_ be Type(_lnum_).
+          1. Return _T_::add(_lnum_, _rnum_).
```
ljharb: this seems a bit strange and hand-wavey; is this notation something that's been used before? (ie, using the result of `Type()` as a dynamic reference to a thing?

spec.html
```diff
@@ -26653,7 +27688,9 @@ <h1>Math.min ( _value1_, _value2_, ..._values_ )</h1>
       <emu-clause id="sec-math.pow">
         <h1>Math.pow ( _base_, _exponent_ )</h1>
         <emu-alg>
-          1. Return the result of <emu-xref href="#sec-applying-the-exp-operator" title>Applying the ** operator</emu-xref> with _base_ and _exponent_ as specified in <emu-xref href="#sec-applying-the-exp-operator"></emu-xref>.
+          1. Let _base_ be ? ToNumber(_base_).
+          1. Let _exponent_ be ? ToNumber(_exponent_).
+          1. Return Number::exponentiate(_base_, _exponent_).
```
```suggestion
          1. Return ! Number::exponentiate(_base_, _exponent_).
```

spec.html
```diff
@@ -33070,7 +34107,7 @@ <h1>Properties of Array Iterator Instances</h1>
 
   <emu-clause id="sec-typedarray-objects">
     <h1>TypedArray Objects</h1>
-    <p>_TypedArray_ objects present an array-like view of an underlying binary data buffer (<emu-xref href="#sec-arraybuffer-objects"></emu-xref>). Each element of a _TypedArray_ instance has the same underlying binary scalar data type. There is a distinct _TypedArray_ constructor, listed in <emu-xref href="#table-49"></emu-xref>, for each of the nine supported element types. Each constructor in <emu-xref href="#table-49"></emu-xref> has a corresponding distinct prototype object.</p>
+    <p>_TypedArray_ objects present an array-like view of an underlying binary data buffer (<emu-xref href="#sec-arraybuffer-objects"></emu-xref>). Each element of a _TypedArray_ instance has the same underlying binary scalar data type. There is a distinct _TypedArray_ constructor, listed in <emu-xref href="#table-49"></emu-xref>, for each of the eleven supported element types. Each constructor in <emu-xref href="#table-49"></emu-xref> has a corresponding distinct prototype object.</p>
```
ljharb: let's just remove the explicit count
```suggestion
    <p>_TypedArray_ objects present an array-like view of an underlying binary data buffer (<emu-xref href="#sec-arraybuffer-objects"></emu-xref>). Each element of a _TypedArray_ instance has the same underlying binary scalar data type. There is a distinct _TypedArray_ constructor, listed in <emu-xref href="#table-49"></emu-xref>, for each of the supported element types. Each constructor in <emu-xref href="#table-49"></emu-xref> has a corresponding distinct prototype object.</p>
```

spec.html
```diff
@@ -33745,6 +34823,7 @@ <h1>%TypedArray%.prototype.set ( _typedArray_ [ , _offset_ ] )</h1>
             1. Let _srcLength_ be _typedArray_.[[ArrayLength]].
             1. Let _srcByteOffset_ be _typedArray_.[[ByteOffset]].
             1. If _srcLength_ + _targetOffset_ &gt; _targetLength_, throw a *RangeError* exception.
+            1. If one of _srcType_ and _targetType_ contains the substring `"Big"` and the other does not, throw a *TypeError* exception.
```
ljharb: this seems like a suuuuuper brittle way to handle this. can we come up with a more reliable approach?

spec.html
```diff
@@ -33840,7 +34919,7 @@ <h1>%TypedArray%.prototype.sort ( _comparefn_ )</h1>
         <p>The following version of SortCompare is used by %TypedArray%`.prototype.sort`. It performs a numeric comparison rather than the string comparison used in <emu-xref href="#sec-array.prototype.sort"></emu-xref>. SortCompare has access to the _comparefn_ and _buffer_ values of the current invocation of the `sort` method.</p>
         <p>When the TypedArray SortCompare abstract operation is called with two arguments _x_ and _y_, the following steps are taken:</p>
         <emu-alg>
-          1. Assert: Both Type(_x_) and Type(_y_) is Number.
+          1. Assert: Both Type(_x_) and Type(_y_) is Number or both are BigInt.
```
```suggestion
          1. Assert: Both Type(_x_) and Type(_y_) are Number or both are BigInt.
```

spec.html
```diff
@@ -34029,6 +35108,7 @@ <h1>_TypedArray_ ( _typedArray_ )</h1>
           1. Else,
             1. Let _data_ be ? AllocateArrayBuffer(_bufferConstructor_, _byteLength_).
             1. If IsDetachedBuffer(_srcData_) is *true*, throw a *TypeError* exception.
+            1. If one of _srcType_ and _elementType_ contains the substring `"Big"` and the other does not, throw a *TypeError* exception.
```
ljharb: ditto here

spec.html
```diff
@@ -34132,7 +35212,10 @@ <h1>TypedArraySpeciesCreate ( _exemplar_, _argumentList_ )</h1>
           1. Assert: _exemplar_ is an Object that has a [[TypedArrayName]] internal slot.
           1. Let _defaultConstructor_ be the intrinsic object listed in column one of <emu-xref href="#table-49"></emu-xref> for _exemplar_.[[TypedArrayName]].
           1. Let _constructor_ be ? SpeciesConstructor(_exemplar_, _defaultConstructor_).
-          1. Return ? TypedArrayCreate(_constructor_, _argumentList_).
+          1. Let _result_ be ? TypedArrayCreate(_constructor_, _argumentList_).
+          1. Assert: _result_ has a [[TypedArrayName]] internal slot.
+          1. If _result_.[[TypedArrayName]] contains the substring `"Big"` and _exemplar_.[[TypedArrayName]] does not contain the substring `"Big"`, or vice versa, throw a *TypeError* exception.
```
ljharb: and here - at the very least, let's make an abstract operation instead of repeating this check multiple times

spec.html
```diff
@@ -35274,21 +36358,21 @@ <h1>GetValueFromBuffer ( _arrayBuffer_, _byteIndex_, _type_, _isTypedArray_, _or
           1. If IsSharedArrayBuffer(_arrayBuffer_) is *true*, then
             1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
             1. Let _eventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
-            1. If _isTypedArray_ is *true* and _type_ is `"Int8"`, `"Uint8"`, `"Int16"`, `"Uint16"`, `"Int32"`, or `"Uint32"`, let _noTear_ be *true*; otherwise let _noTear_ be *false*.
+            1. If _isTypedArray_ is *true* and _type_ is `"Int8"`, `"Uint8"`, `"Int16"`, `"Uint16"`, `"Int32"`, or `"Uint32"` or if _type_ is `"BigInt64"` or `"BigUint64"` and _order_ is not `"Init"` or `"Unordered"`, let _noTear_ be *true*; otherwise let _noTear_ be *false*.
```
```suggestion
            1. If _isTypedArray_ is *true* and _type_ is `"Int8"`, `"Uint8"`, `"Int16"`, `"Uint16"`, `"Int32"`, or `"Uint32"`, or if _type_ is `"BigInt64"` or `"BigUint64"` and _order_ is not `"Init"` or `"Unordered"`, let _noTear_ be *true*; otherwise let _noTear_ be *false*.
```
this might be clearer if it was split up - or perhaps moved to an abstract operation.

spec.html
```diff
@@ -35313,15 +36397,15 @@ <h1>SetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _isTypedArra
...
           1. If _isLittleEndian_ is not present, set _isLittleEndian_ to the value of the [[LittleEndian]] field of the surrounding agent's Agent Record.
-          1. Let _rawBytes_ be NumberToRawBytes(_type_, _value_, _isLittleEndian_).
+          1. Let _rawBytes_ be NumericToRawBytes(_type_, _value_, _isLittleEndian_).
           1. If IsSharedArrayBuffer(_arrayBuffer_) is *true*, then
             1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
             1. Let _eventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
-            1. If _isTypedArray_ is *true* and _type_ is `"Int8"`, `"Uint8"`, `"Int16"`, `"Uint16"`, `"Int32"`, or `"Uint32"`, let _noTear_ be *true*; otherwise let _noTear_ be *false*.
+            1. If _isTypedArray_ is *true* and _type_ is `"Int8"`, `"Uint8"`, `"Int16"`, `"Uint16"`, `"Int32"`, or `"Uint32"`, or if _type_ is `"BigInt64"` or `"BigUint64"` and _order_ is not `"Init"` or `"Unordered"`, let _noTear_ be *true*; otherwise let _noTear_ be *false*.
```
ljharb: the repetition of this complex line makes me want an abstract operation for this even more

spec.html
jmdyck: f963efc should have replaced `BigIntToString` with `BigInt::ToString`.

spec.html
jmdyck: That particular notation isn't used in the current spec, but we do have something that's semantically similar, where Environment Record methods are "dynamically dispatched" based on the specific type of record. (Ditto Module Record methods.)

spec.html
caiolima: Oops. Probably I reverted this after rebase. I'll change it to `BigInt::ToString`.

spec.html
caiolima: I'll think in something better.

spec.html
caiolima: I think using `SameValue(_n_, *NaN*)` here would be a unnecessary complex step, but I can change if you think it is better.

spec.html
```diff
@@ -1110,30 +1110,855 @@ <h1>Well-Known Symbols</h1>
...
+      <emu-clause id="sec-ecmascript-language-types-bigint-type">
+        <h1>The BigInt Type</h1>
+        <p>The BigInt type represents a mathematical integer value. The value may be any size and is not limited to a particular bit-width. Generally, where not otherwise noted, operations are designed to return exact mathematically-based answers. For binary operations, BigInts act as two's complement binary strings, with negative numbers treated as having bits set infinitely to the left.</p>
+
+        <p>The BigInt::unit value is *1n*.</p>
+
+        <emu-clause id="sec-numeric-types-bigint-unaryMinus">
+          <h1>BigInt::unaryMinus ( _x_ )</h1>
+          <p>The abstract operation BigInt::unaryMinus with an argument _x_ of type BigInt returns the result of negating _x_.</p>
+          <emu-note>There is only one *0n* value; `-0n` is the same as *0n*.</emu-note>
```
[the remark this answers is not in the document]
ljharb: yes, something like that seems great - most numeric methods have special early-return cases for zeroes, infinities, and NaN.

spec.html
```diff
@@ -34798,6 +34798,22 @@ <h1>%TypedArray%.prototype.set ( _array_ [ , _offset_ ] )</h1>
           </emu-alg>
         </emu-clause>
 
+        <emu-clause id="sec-is-compatible-typed-array-type" aoid="IsCompatibleTypedArrayType">
+          <h1>IsCompatibleTypedArrayType ( _type_ , _otherType_ )</h1>
```
ljharb: what does "is compatible" mean? why would Uint8ClampedArray and Float32Array be "compatible"?

caiolima: I agree this name is not quite good. I'm changing this to `IsCopyableTypedArrayType` and use Types instead of the constructor name.

spec.html
devsnek: I feel like the norm is to use slots, more like lnum.[[Add]](rnum)

caiolima: It is not very clear to me what steps i need to take here.

> I feel like the norm is to use slots, more like `lnum.[[Add]](rnum)`

@devsnek I don't remember seen places in spec where slot is used on Number primitives. Do you mind sharing where this is being used currently?

devsnek: @caiolima not on Number specifically, just in general. `object.[[Get]]()`, `module.[[Evaluate]]()`, etc.

jmdyck: Currently, only objects have internal slots. I think it'd be pretty weird to extend that to numeric values.

For `module.[[Evaluate]]()`, you're probably thinking of `_module_.Evaluate()`, where `Evaluate` is a "concrete method" of module records. So far, only record types have "concrete methods", but I suppose we could extend that to numeric types.

devsnek: @jmdyck yes, thank you for the correction.

spec.html
```diff
@@ -35336,15 +36444,20 @@ <h1>SetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _isTypedArra
...
           1. If IsSharedArrayBuffer(_arrayBuffer_) is *true*, then
             1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
             1. Let _eventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
-            1. If _isTypedArray_ is *true* and _type_ is `"Int8"`, `"Uint8"`, `"Int16"`, `"Uint16"`, `"Int32"`, or `"Uint32"`, let _noTear_ be *true*; otherwise let _noTear_ be *false*.
+            1. If _isTypedArray_ is *true* and _type_ is `"Int8"`, `"Uint8"`, `"Int16"`, `"Uint16"`, `"Int32"`, or `"Uint32"`, then
+              1. Let _noTear_ be *true*.
+            1. Else if _isTypedArray_ is *true* and _type_ is `"BigInt64"` or `"BigUint64"` and _order_ is not `"Init"` or `"Unordered"`, then
+              1. Let _noTear_ be *true*.
+            1. Else
+              1. Let _noTear_ be *false*.
```
ljharb: i'd still like an abstract operation for "noTear" with a meaningful name

spec.html
```diff
@@ -36004,16 +37158,16 @@ <h1>The Atomics Object</h1>
...
+          1. If _waitable_ is not present, set _waitable_ to *false*.
           1. Perform ? RequireInternalSlot(_typedArray_, [[TypedArrayName]]).
           1. Let _typeName_ be _typedArray_.[[TypedArrayName]].
-          1. If _onlyInt32_ is *true*, then
-            1. If _typeName_ is not `"Int32Array"`, throw a *TypeError* exception.
+          1. If _waitable_ is *true*, then
+            1. If _typeName_ is not `"Int32Array"` or `"BigInt64Array"`, throw a *TypeError* exception.
           1. Else,
-            1. If _typeName_ is not `"Int8Array"`, `"Uint8Array"`, `"Int16Array"`, `"Uint16Array"`, `"Int32Array"`, or `"Uint32Array"`, throw a *TypeError* exception.
+            1. If _typeName_ is not `"Int8Array"`, `"Uint8Array"`, `"Int16Array"`, `"Uint16Array"`, `"Int32Array"`, `"Uint32Array"`, `"BigUint64Array"`, or `"BigInt64Array"`, throw a *TypeError* exception.
```
ljharb: is there perhaps a list of these values somewhere so they don't need to be hardcoded in random places?

spec.html
```diff
@@ -33734,15 +34812,33 @@ <h1>%TypedArray%.prototype.set ( _array_ [ , _offset_ ] )</h1>
...
+              1. Perform SetValueInBuffer(_targetBuffer_, _targetByteIndex_, _targetType_, _value_, *true*, `"Unordered"`).
               1. Set _k_ to _k_ + 1.
               1. Set _targetByteIndex_ to _targetByteIndex_ + _targetElementSize_.
             1. Return *undefined*.
           </emu-alg>
         </emu-clause>
 
+        <emu-clause id="sec-is-copyable-typed-array-types" aoid="IsCopyableTypedArrayTypes">
+          <h1>IsCopyableTypedArrayTypes ( _type_ , _otherType_ )</h1>
+          <p>The abstract operation IsCopyableTypedArrayTypes checks if it is possible to copy elements between TypedArray of types _type_ and _otherType_ :</p>
```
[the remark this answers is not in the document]
caiolima: I think I didn't understand your question. This function is checking if the type of elements of the 2 TypedArrays are   compatible. By compatible, I mean that there is no mix of `BigInt` and `Number` on `type` and `otherType`. In such case, we return `false` for every combination of `BigInt` and any other type that isn't `BigInt`. In all other cases, we return `true`. In the case you asked, if `type = Int8Array`, it only returns `false` if `otherType = BigInt64 or BigUint64`. This is required to avoid usage of TyperArrays as a way to have implicit cast of `BigInt` and `Number`

spec.html
```diff
@@ -35297,21 +36401,25 @@ <h1>GetValueFromBuffer ( _arrayBuffer_, _byteIndex_, _type_, _isTypedArray_, _or
           1. If IsSharedArrayBuffer(_arrayBuffer_) is *true*, then
             1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
             1. Let _eventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
-            1. If _isTypedArray_ is *true* and _type_ is `"Int8"`, `"Uint8"`, `"Int16"`, `"Uint16"`, `"Int32"`, or `"Uint32"`, let _noTear_ be *true*; otherwise let _noTear_ be *false*.
+            1. If _isTypedArray_ is *true* and _type_ is `"Int8"`, `"Uint8"`, `"Int16"`, `"Uint16"`, `"Int32"`, or `"Uint32"`, then
```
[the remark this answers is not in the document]
caiolima: I think it is better `IsIntegetTypedArrayType`, since it doesn't include `float` versions of `TypedArray`.

spec.html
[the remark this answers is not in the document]
ljharb: Ah - so the intention here is to ensure that "both types are bigint" or "both types are number"?

In that case, perhaps it would be cleaner to add an internal slot to all typed arrays like `[[ContentType]]`, and set it to "Number" or "BigInt", and then a) the comparison only needs to compare the ContentType slot - that would also make it extensible to, say, Decimals in the future without a bunch of complex logic?

spec.html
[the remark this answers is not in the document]
ljharb: Perfect, sounds great!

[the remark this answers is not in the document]
caiolima: Actually, I think that `IsIntegetTypedArrayType` is also not a good idea, since this list doesn't include `Uint8C` type.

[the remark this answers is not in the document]
ljharb: `IsUnclampedIntegerTypedArrayType`? :-p

spec.html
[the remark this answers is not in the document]
caiolima: > Ah - so the intention here is to ensure that "both types are bigint" or "both types are number"?
Yes.

> In that case, perhaps it would be cleaner to add an internal slot to all typed arrays like [[ContentType]], and set it to "Number" or "BigInt", and then a) the comparison only needs to compare the ContentType slot - that would also make it extensible to, say, Decimals in the future without a bunch of complex logic?

SGTM. I like this approach.

spec.html
caiolima: We can use `IsUnclampedIntegerTypedArrayType` and `IsBigIntTypedArrayType` to perform this check without hardcode this list.

jmdyck:
I've got some more edits, will probably submit them as a PR again.

ljharb:
@jmdyck it'd be great to make them as a PR onto caiolima's fork, or link the commit and i'll pull them in?

jmdyck:
Yup, I meant a PR against caiolima:big-int. (That's what I did before.)

caiolima:
@ljharb 

FYI, this is ready for review again.

jmdyck:
> I've got some more edits, will probably submit them as a PR again.

Done: https://github.com/caiolima/ecma262/pull/6

caiolima:
@ljharb FYi, it is ready for review.

caiolima:
@ljharb I just rebased thi with master. It is ready to be reviewed again.

spec.html
```diff
@@ -1111,30 +1111,857 @@ <h1>Well-Known Symbols</h1>
...
+        <p>Some ECMAScript operators deal only with integers in specific ranges such as <emu-eqn>-2<sup>31</sup></emu-eqn> through <emu-eqn>2<sup>31</sup> - 1</emu-eqn>, inclusive, or in the range 0 through <emu-eqn>2<sup>16</sup> - 1</emu-eqn>, inclusive. These operators accept any value of the Number type but first convert each such value to an integer value in the expected range. See the descriptions of the numeric conversion operations in <emu-xref href="#sec-type-conversion"></emu-xref>.</p>
+
+        <p>The Number::unit value is *1*.</p>
+
+        <emu-clause id="sec-numeric-types-number-unaryMinus">
+          <h1>Number::unaryMinus ( _x_ )</h1>
+          <emu-alg>
+            1. If _x_ is *NaN*, return *NaN*.
+            1. Return the result of negating _x_; that is, compute a Number with the same magnitude but opposite sign.
```
ljharb: does "Infinity" have a magnitude, or should it have special cases here just like NaN?

spec.html
```diff
@@ -1111,30 +1111,857 @@ <h1>Well-Known Symbols</h1>
...
+            1. If _x_ is *-0* and _y_ is *+0*, return *true*.
+            1. Return *false*.
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-numeric-types-number-sameValue">
+          <h1>Number::sameValue ( _x_, _y_ )</h1>
+          <emu-alg>
+            1. If _x_ is *NaN* and _y_ is *NaN*, return *true*.
+            1. If _x_ is *+0* and _y_ is *-0*, return *false*.
```
ljharb: are these needed? if the zero lines are removed from this algorithm, it should fall through to false, no?

spec.html
```diff
@@ -1111,30 +1111,857 @@ <h1>Well-Known Symbols</h1>
...
+
+        <emu-clause id="sec-bigintbitwiseop" aoid="BigIntBitwiseOp">
+          <h1>BigIntBitwiseOp ( _op_, _x_, _y_ )</h1>
+          <emu-alg>
+            1. Let _result_ be *0n*.
+            1. Let _shift_ be 0.
+            1. Repeat, until (_x_ = 0 or _x_ = -1) and (_y_ = 0 or _y_ = -1),
+              1. Let _xDigit_ be _x_ modulo 2.
+              1. Let _yDigit_ be _y_ modulo 2.
+              1. Let _result_ be _result_ + 2<sup>_shift_</sup> &times; _op_(_xDigit_, _yDigit_).
```
ljharb: i'm a little confused how ``` `&amp;` ```, for example, is an abstract operation. Can you elaborate?

spec.html
```diff
@@ -1111,30 +1111,857 @@ <h1>Well-Known Symbols</h1>
...
+          <emu-alg>
+            1. Return BigIntBitwiseOp(`|`, _x_, _y_).
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-numeric-types-bigint-tostring" aoid="BigInt::toString">
+          <h1>BigInt::toString ( _x_ )</h1>
+          <p>The abstract operation BigInt::toString converts a BigInt _x_ to String format as follows:</p>
+          <emu-alg>
+            1. If _x_ is less than zero, return the string-concatenation of the String `"-"` and BigInt::toString(-_x_).
```
```suggestion
            1. If _x_ is less than zero, return the string-concatenation of the String `"-"` and ! BigInt::toString(-_x_).
```

spec.html
```diff
@@ -4030,7 +4991,7 @@ <h1>ToString ( _argument_ )</h1>
               Number
             </td>
             <td>
-              Return NumberToString(_argument_).
+              Return Number::toString(_argument_).
```
ljharb: `!` or `?`?

spec.html
```diff
@@ -4049,6 +5010,14 @@ <h1>ToString ( _argument_ )</h1>
               Throw a *TypeError* exception.
             </td>
           </tr>
+          <tr>
+            <td>
+              BigInt
+            </td>
+            <td>
+              Return BigInt::toString(_argument_).
```
ljharb: `!` or `?`?

spec.html
```diff
@@ -4524,10 +5441,18 @@ <h1>Abstract Equality Comparison</h1>
         1. If _x_ is *undefined* and _y_ is *null*, return *true*.
         1. If Type(_x_) is Number and Type(_y_) is String, return the result of the comparison _x_ == ! ToNumber(_y_).
         1. If Type(_x_) is String and Type(_y_) is Number, return the result of the comparison ! ToNumber(_x_) == _y_.
+        1. If Type(_x_) is BigInt and Type(_y_) is String, then
+          1. Let _n_ be StringToBigInt(_y_).
```
```suggestion
          1. Let _n_ be ! StringToBigInt(_y_).
```

spec.html
```diff
@@ -4524,10 +5441,18 @@ <h1>Abstract Equality Comparison</h1>
...
+          1. If _n_ is *NaN*, return *false*.
+          1. Return the result of the comparison _x_ == _n_.
+        1. If Type(_x_) is String and Type(_y_) is BigInt, return the result of the comparison _y_ == _x_.
         1. If Type(_x_) is Boolean, return the result of the comparison ! ToNumber(_x_) == _y_.
         1. If Type(_y_) is Boolean, return the result of the comparison _x_ == ! ToNumber(_y_).
-        1. If Type(_x_) is either String, Number, or Symbol and Type(_y_) is Object, return the result of the comparison _x_ == ToPrimitive(_y_).
-        1. If Type(_x_) is Object and Type(_y_) is either String, Number, or Symbol, return the result of the comparison ToPrimitive(_x_) == _y_.
+        1. If Type(_x_) is either String, Number, BigInt, or Symbol and Type(_y_) is Object, return the result of the comparison _x_ == ToPrimitive(_y_).
```
ljharb: do we not have an `IsPrimitiveType()` abstract operation to use here, and on the next line?

(we can add this in a followup PR)

spec.html
```diff
@@ -4537,14 +5462,9 @@ <h1>Strict Equality Comparison</h1>
...
-          1. If _x_ is *NaN*, return *false*.
-          1. If _y_ is *NaN*, return *false*.
-          1. If _x_ is the same Number value as _y_, return *true*.
-          1. If _x_ is *+0* and _y_ is *-0*, return *true*.
-          1. If _x_ is *-0* and _y_ is *+0*, return *true*.
-          1. Return *false*.
-        1. Return SameValueNonNumber(_x_, _y_).
+        1. If Type(_x_) is Number or BigInt, then
+          1. Return ! Type(_x_)::equal(_x_, _y_).
+        1. Return SameValueNonNumeric(_x_, _y_).
```
```suggestion
        1. Return ! SameValueNonNumeric(_x_, _y_).
```

spec.html
ljharb: still not convinced by this notation, but i don't have a better suggestion at this time.

spec.html
```diff
@@ -14027,9 +14870,13 @@ <h1>Runtime Semantics: Evaluation</h1>
...
         1. Let _rref_ be the result of evaluating _B_.
         1. Let _rval_ be ? GetValue(_rref_).
-        1. Let _lnum_ be ? ToInt32(_lval_).
-        1. Let _rnum_ be ? ToInt32(_rval_).
-        1. Return the result of applying the bitwise operator @ to _lnum_ and _rnum_. The result is a signed 32-bit integer.
+        1. Let _lnum_ be ? ToNumeric(_lval_).
+        1. Let _rnum_ be ? ToNumeric(_rval_).
+        1. If Type(_lnum_) is different from Type(_rnum_), throw a *TypeError* exception.
+        1. Let _T_ be Type(_lnum_).
+        1. If @ is `&amp;`, return _T_::bitwiseAND(_lnum_, _rnum_).
```
ljharb: this is kind of what i'd expect for the `_op_()` i commented on above

spec.html
```diff
@@ -26011,6 +26858,137 @@ <h1>Properties of Number Instances</h1>
...
+        <emu-alg>
+          1. Let _bits_ be ? ToIndex(_bits_).
+          1. Let _bigint_ be ? ToBigInt(_bigint_).
+          1. Return a BigInt representing _bigint_ modulo 2<sup>_bits_</sup>.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-bigint.prototype">
+        <h1>BigInt.prototype</h1>
+        <p>The initial value of `BigInt.prototype` is the intrinsic object %BigIntPrototype%.</p>
```
ljharb: now that we have the new intrinsics format, let's just remove this intrinsic entirely.
```suggestion
```

spec.html
```diff
@@ -26011,6 +26858,137 @@ <h1>Properties of Number Instances</h1>
...
+        <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-properties-of-the-bigint-prototype-object">
+      <h1>Properties of the BigInt Prototype Object</h1>
+      <p>The BigInt prototype object:</p>
+      <ul>
+        <li>is the intrinsic object <dfn>%BigIntPrototype%</dfn>.</li>
```
```suggestion
```

spec.html
```diff
@@ -26011,6 +26858,137 @@ <h1>Properties of Number Instances</h1>
...
+        <li>is an ordinary object.</li>
+        <li>is not a BigInt object; it does not have a [[BigIntData]] internal slot.</li>
+        <li>has a [[Prototype]] internal slot whose value is the intrinsic object %ObjectPrototype%.</li>
```
```suggestion
        <li>has a [[Prototype]] internal slot whose value is the intrinsic object %Object.prototype%.</li>
```

spec.html
```diff
@@ -35834,6 +36901,46 @@ <h1>DataView.prototype.constructor</h1>
         <p>The initial value of `DataView.prototype.constructor` is %DataView%.</p>
       </emu-clause>
 
+      <emu-clause id="sec-dataview.prototype.getbigint64">
+        <h1>DataView.prototype.getBigInt64 ( _byteOffset_ [ , _littleEndian_ ] )</h1>
+        <p>When the `getBigInt64` method is called with argument _byteOffset_ and optional argument _littleEndian_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _v_ be the *this* value.
+          1. If _littleEndian_ is not present, let _littleEndian_ be *undefined*.
```
ljharb: This line seems unnecessary, since all unprovided arguments are implicitly undefined?
```suggestion
```

(and throughout)

spec.html
caiolima: I'm not 100% sure it does, but from the part of spec explaining the `the Number Value of x`, those values are represented as `2ℝ ** 1024ℝ` and `(-2ℝ) ** 1024ℝ` and they have magnitude.

spec.html
caiolima: I don't think it would be correct, since we have this statement into the spec: `Note that there is both a positive zero and a negative zero. For brevity, these values are also referred to for expository purposes by the symbols +0 and -0, respectively. (Note that these two different zero Number values are produced by the program expressions +0 (or simply 0) and -0.)`.
IMHO, it would very confusing if we rely on `If _x_ is the same Number value as _y_, return *true*.` to handle zero cases, since we don't consider them as the same Number value previously.

spec.html
caiolima: I don't think we have it, at least named as `IsPrimitiveType`. I think it is a good abstract operation to include, though.

spec.html
caiolima: It should be `!`, since we already know that `argument` is a Number.

spec.html
caiolima: Ditto.

spec.html
caiolima: Thanks for pointing out this. At the first time I saw it, I was thinking in prefix notation, but I think I have a better idea for this.

spec.html
caiolima: Yes. I agree. We need to define some Abstract Operations to properly use `_op_()` there.

caiolima:
@ljharb Thank you very much for the review. I've uploaded a version with fixes you pointed and also changed how we are using `_op_` into `BigIntBitwiseOp`.

spec.html
ljharb: i'm fine to keep them; just wanted to confirm that they aren't strictly necessary

spec.html
ljharb: i see another place in the PR/spec where there's an if/else dispatch - ie, it looks at the string, and invokes the right hardcoded abstract op, rather than trying to be metaprogrammy about it

spec.html
ljharb: if it can be added prior to this PR, great; if it should be added as part of this PR, great; if after, great :-)

spec.html
```diff
@@ -1935,29 +1966,29 @@ <h1>BigIntBitwiseOp ( _op_, _x_, _y_ )</h1>
         <emu-clause id="sec-numeric-types-bigint-bitwiseAND">
           <h1>BigInt::bitwiseAND ( _x_, _y_ )</h1>
           <emu-alg>
-            1. Return BigIntBitwiseOp(`&amp;`, _x_, _y_).
+            1. Return BigIntBitwiseOp(BinaryAnd, _x_, _y_).
```
ljharb: this works and is better; but i'm not sure if it's right to pass around abstract ops as first-class values.

spec.html
caiolima: Nice. I'll do in a follow up PR then.

spec.html
caiolima: This is fair. I changed to this approach.

spec.html
jmdyck: There's some precedent for this. 
- The `job` parameter of `EnqueueJob` is (the name of) an abstract operation.

- The `op` parameter of `GetModifySetValueInBuffer` and `AtomicReadModifyWrite` is a "semantic function" or a "combining operation", which appears to be a kind of abstract operation.

- The RegExp runtime semantics pass around various kinds of "internal closures" (MatchResult, Continuation, Matcher, AssertionTester), which are (roughly speaking) dynamically created abstract operations.

- The `steps` parameter of `CreateBuiltinFunction` isn't a abstract operation, but it's pretty close.

- `NumberToRawBytes` binds an abstract operation to an alias, which is treating it as a first-class value.

caiolima:
I added the fixes of static dispatch suggested by @ljharb and rebased the patch.

spec.html
```diff
@@ -35285,9 +36352,46 @@ <h1>CloneArrayBuffer ( _srcBuffer_, _srcByteOffset_, _srcLength_, _cloneConstruc
...
       </emu-clause>
 
-      <emu-clause id="sec-rawbytestonumber" aoid="RawBytesToNumber">
-        <h1>RawBytesToNumber ( _type_, _rawBytes_, _isLittleEndian_ )</h1>
-        <p>The abstract operation RawBytesToNumber takes three parameters, a String _type_, a List _rawBytes_, and a Boolean _isLittleEndian_. This operation performs the following steps:</p>
+      <emu-clause id="sec-isunsignedelementtype" aoid="IsUnsignedElementType">
+        <h1>IsUnsignedElementType ( _type_ )</h1>
+        <p>The abstract operation IsUnsignedElementType verifies if the argument _type_ is an unsigned TypedArray element type. This operation performs the following steps:</p>
+        <emu-alg>
+          1. If the first code unit of _type_ is the code unit 0x0055 (LATIN CAPITAL LETTER U) or _type_ is `"BigUint64"`, return *true*.
```
ljharb: I would prefer something explicit rather than relying on the first letter being "U".

caiolima: Done.

littledan:
I've changed the label from "pending Stage 4" to "has Stage 4" based on the result in the June 2019 TC39 meeting.

What are the next steps for this PR to be able to land?

ljharb:
It can’t actually have stage 4 until this PR is approved by the editors; proposing it for stage 4 in committee without that was premature.

littledan:
Given that you identify the concern as non-blocking, and you haven't made another suggestion, would it make sense to land this PR and be open to other editorial alternatives?

littledan:
Thanks for the approval, BTW!

ljharb:
It’ll need approval from the other editor first. My approval means that yes, it will land once it has that.

spec.html
```diff
@@ -1110,30 +1110,898 @@ <h1>Well-Known Symbols</h1>
...
+        </emu-clause>
+
+        <emu-clause id="sec-numeric-types-bigint-add">
+          <h1>BigInt::add ( _x_, _y_ )</h1>
+          <p>The abstract operation BigInt::add with two arguments _x_ and _y_ of type BigInt returns the BigInt value that represents the sum of _x_ and _y_.</p>
+        </emu-clause>
+
+        <emu-clause id="sec-numeric-types-bigint-subtract">
+          <h1>BigInt::subtract ( _x_, _y_ )</h1>
+          <p>The abstract operation BigInt::subtract with two arguments _x_ and _y_ of type BigInt returns the BigInt representing the difference _x_ minus _y_.</p>
```
```suggestion
          <p>The abstract operation BigInt::subtract with two arguments _x_ and _y_ of type BigInt returns the BigInt value that represents the difference _x_ minus _y_.</p>
```
... for consistency with my last commit. (I missed it because I was only looking for "*a* BigInt representing".)

spec.html
```diff
@@ -35833,6 +36959,25 @@ <h1>DataView.prototype.constructor</h1>
         <p>The initial value of `DataView.prototype.constructor` is %DataView%.</p>
       </emu-clause>
 
+      <emu-clause id="sec-dataview.prototype.getbigint64">
+        <h1>DataView.prototype.getBigInt64 ( _byteOffset_ [ , _littleEndian_ ] )</h1>
+        <p>When the `getBigInt64` method is called with argument _byteOffset_ and optional argument _littleEndian_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _v_ be the *this* value.
```
jmdyck: At this point, insert:
`1. If _littleEndian_ is not present, let _littleEndian_ be *undefined*.`

caiolima:
@zenparsing is there anything else we need to do into this PR to get it merged?

caiolima:
@ljharb and @zenparsing Thanks a lot for the review! @jmdyck and @littledan thx a lot for the fixes and reviews as well.

PR #1516, opened 2019-04-24
Editorial: Remove unnecessary steps before NewPromiseCapability

leobalter:
NewPromiseCapability will immediately check if the argument (C) is an Object using IsConstructor and it will throw a TypeError if it's not.

This patch will just simplify the text with a fewer steps but this won't cause any new observable changes.

```
NewPromiseCapability ( _C_ )

1. If IsConstructor(_C_) is *false*, throw a *TypeError* exception.

IsConstructor ( _argument_ )

1. If Type(_argument_) is not Object, return *false*.
```

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38243,7 +38243,6 @@ THH:mm:ss.sss
         <p>The `all` function returns a new promise which is fulfilled with an array of fulfillment values for the passed promises, or rejects with the reason of the first passed promise that rejects. It resolves all elements of the passed iterable to promises as it runs this algorithm.</p>
         <emu-alg>
           1. Let _C_ be the *this* value.
-          1. If Type(_C_) is not Object, throw a *TypeError* exception.
           1. Let _promiseCapability_ be ? NewPromiseCapability(_C_).
           1. Let _iteratorRecord_ be GetIterator(_iterable_).
           1. IfAbruptRejectPromise(_iteratorRecord_, _promiseCapability_).
@@ -38331,7 +38330,6 @@ THH:mm:ss.sss
         <p>The `race` function returns a new promise which is settled in the same way as the first passed promise to settle. It resolves all elements of the passed _iterable_ to promises as it runs this algorithm.</p>
         <emu-alg>
           1. Let _C_ be the *this* value.
-          1. If Type(_C_) is not Object, throw a *TypeError* exception.
           1. Let _promiseCapability_ be ? NewPromiseCapability(_C_).
           1. Let _iteratorRecord_ be GetIterator(_iterable_).
           1. IfAbruptRejectPromise(_iteratorRecord_, _promiseCapability_).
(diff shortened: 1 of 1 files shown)


leobalter:
@gsathya @mathiasbynens allSettled can follow with the same change if this is accepted.

ljharb:
@leobalter this seems like it introduces brittleness; if `NewPromiseCapability` is changed to not immediately throw on not `ToObject`, then this algorithm would change - it'd be reasonable to say that any change to `NewPromiseCapability` would necessitate checking callers to ensure this, but it also seems like the "redundant" step isn't simplifying - it's just hiding a guard condition inside an abstraction.

leobalter:
> seems like the "redundant" step isn't simplifying - it's just hiding a guard condition inside an abstraction.

I disagree. These operations are very specific and reminds me a lot of [`ValidateTypedArray` in TypedArray prototype methods](https://tc39.github.io/ecma262/#sec-%typedarray%.prototype.copywithin) where `this` is checked as well.

NewPromiseCapability implies the given argument needs to be a valid constructor, therefore an object. Otherwise it will throw a TypeError.

The extra step I'm trying to remove caused me confusion today when I was reading these while reviewing a few tests.

I'm fine if it's rejected, but the proposed simplicity would be helpful to me in any future.

ljharb:
The difference to me is that the TypedArray examples are calling into an abstract operation specifically to *validate* the typed array; whereas this change calls into `NewPromiseCapability` and has the *side effect* of validating that it's an object.

It's certainly true, though, that for example [here](https://tc39.github.io/ecma262/#sec-promise.race), checking `C` locally is only needed to tell you "it will work in `NewPromiseCapability`", as well as "it will work in `PerformPromiseRace`" (which asserts that it's a constructor). So, to apply my intuition would mean that instead of removing this check, it should be checking `IsConstructor` (so that both abstract ops receiving `C` have had it pre-validated) - which would then allow the `? NewPromiseCapability` to become `! NewPromiseCapability`, which seems a bit clearer to me. What do you think about that?

leobalter:
> which would then allow the `? NewPromiseCapability` to become `! NewPromiseCapability`

Not possible, there are more steps there that can return an abrupt completion.

To add extra clarity, it would be very safe to move the "Assert: is Constructor" parts from `PerformPromise{Race,All,AllSettled}` to direct steps into the methods before calling this abstraction. e.g.:

```
          1. Let _C_ be the *this* value.
          1. Let _promiseCapability_ be ? NewPromiseCapability(_C_).
          1. Assert: IsConstructor(_C_) is true.
          1. Let _iteratorRecord_ be GetIterator(_iterable_).
          1. IfAbruptRejectPromise(_iteratorRecord_, _promiseCapability_).
          1. Let _result_ be PerformPromiseRace(_iteratorRecord_, _C_, _promiseCapability_).
```

I don't think this is necessary, same way it validates that `iterable` argument is not `undefined` or `null` and it takes several abstractions deep after `GetIterator` until it is finally done.

ljharb:
ah, that's true, the other place `NewPromiseCapability` can return an abrupt completion is [step 7's Construct](https://tc39.github.io/ecma262/#sec-newpromisecapability).

Given that, the property I'm hoping for here can't really be met anyways (maximizing validations and exceptions done up front), so this change seems fine.

leobalter:
I appreciate that.

leobalter:
I renamed the commit title without any other changes. The original name was too generic.

PR #1526, opened 2019-05-01
Editorial: Improve note on forcing string comparison

shvaikalesh:


spec.html
```diff
@@ -13816,7 +13816,7 @@ <h1>Runtime Semantics: Evaluation</h1>
         <p>Given the above definition of equality:</p>
         <ul>
           <li>
-            String comparison can be forced by: `"" + a == "" + b`.
+            String comparison can be forced by: `\`${a}\` == \`${b}\``.
```
ljharb: the old example is still true; perhaps instead of replacing it, we could include both?

shvaikalesh: Old example is not idiomatic `ToString` conversion: `"" + a` calls [`ToPrimitive`](https://tc39.github.io/ecma262/#sec-toprimitive) without a *hint*, while template literal passes `"string"` *hint*. This is important is `a` is an object with both `valueOf` and `toString`, or defines `@@toPrimitive`.

ljharb: That’s true; i suppose this is indeed the only syntactic way to force ToString (otherwise you’d need `String(a) === String(b)`)

PR #1529, opened 2019-05-08
Editorial: Use IsCallable abstract op

shvaikalesh:


spec.html
```diff
@@ -4815,7 +4815,7 @@ <h1>EnumerableOwnPropertyNames ( _O_, _kind_ )</h1>
       <h1>GetFunctionRealm ( _obj_ )</h1>
       <p>The abstract operation GetFunctionRealm with argument _obj_ performs the following steps:</p>
       <emu-alg>
-        1. Assert: _obj_ is a callable object.
+        1. Assert: IsCallable(_obj_) is *true*.
```
```suggestion
        1. Assert: ! IsCallable(_obj_) is *true*.
```

jmdyck: If we assert that the result of the invocation is `*true*`, then that already implies that it isn't an abrupt completion.

ljharb: i suppose you're right, but it doesn't hurt imo to have the extra assertion. ¯\\\_(ツ)_/¯

ljharb: actually - without the `?` or the `!`, `IsCallable` returns a completion record, which might need to be unpacked - so the `!` seems like it does serve a purpose?

jmdyck: In the world where `IsCallable` always returns a completion record, you can rely on [5.2.3.1](https://tc39.github.io/ecma262/#sec-implicit-completion-values)'s last paragraph to unpack it in this context, so `!` isn't needed for that purpose. But if you think that the extra assertion is good for the reader, fine.

bakkot: FWIW I as a reader find it a lot easier to understand algorithms when all Completion Records are explicitly unwrapped; I support including the `!` for clarity.

shvaikalesh:
Apart from [`IsCallable`](https://tc39.github.io/ecma262/#sec-iscallable) (55 references), there are couple more abstract ops that always return normal completion and are implicitly unpacked:

* [`IsConstructor`](https://tc39.github.io/ecma262/#sec-isconstructor) (27 refs)
* [`IsInteger`](https://tc39.github.io/ecma262/#sec-isinteger) (5 refs)
* [`IsPropertyKey`](https://tc39.github.io/ecma262/#sec-ispropertykey) (36 refs)
* [`IsStringPrefix`](https://tc39.github.io/ecma262/#sec-isstringprefix) (2 refs)
* [`SameValue`](https://tc39.github.io/ecma262/#sec-samevalue) (66 refs)
* [`SameValueZero`](https://tc39.github.io/ecma262/#sec-samevaluezero) (13 refs)
* [`SameValueNonNumber`](https://tc39.github.io/ecma262/#sec-samevaluenonnumber) (3 refs)

I agree that explicitly unwrapping these operations benefits readability: there are a few `Is*` helpers that yield abrupt completions ([`IsArray`](https://tc39.github.io/ecma262/#sec-isarray), [`IsExtensible`](https://tc39.github.io/ecma262/#sec-isextensible-o), [`IsRegExp`](https://tc39.github.io/ecma262/#sec-isregexp)), so a reader can't really tell by name.

Is PR welcome?

ljharb:
Yes, a separate PR for that would be great.

PR #1532, opened 2019-05-09
Editorial: Reference leading/trailing surrogate definitions more

gibson042:


spec.html
```diff
@@ -29107,11 +29108,11 @@ <h1>%StringIteratorPrototype%.next ( )</h1>
...
-            1. If _first_ &lt; 0xD800 or _first_ &gt; 0xDBFF or _position_ + 1 = _len_, let _resultString_ be the String value consisting of the single code unit _first_.
+            1. Let _first_ be the code unit at index _position_ within _s_.
+            1. If _first_ is not a <emu-xref href="#leading-surrogate"></emu-xref> or if _position_ + 1 = _len_, let _resultString_ be the String value consisting of the single code unit _first_.
             1. Else,
-              1. Let _second_ be the numeric value of the code unit at index _position_ + 1 within the String _s_.
-              1. If _second_ &lt; 0xDC00 or _second_ &gt; 0xDFFF, let _resultString_ be the String value consisting of the single code unit _first_.
+              1. Let _second_ be the code unit at index _position_ + 1 within the String _s_.
+              1. If _second_ is not a <emu-xref href="#trailing-surrogate"></emu-xref>, let _resultString_ be the String value consisting of the single code unit _first_.
               1. Else, let _resultString_ be the string-concatenation of the code unit _first_ and the code unit _second_.
```
ljharb: it kind of seems like we could benefit from an abstract operation that takes `s`, and `position`, and returns either the concatenation, or a List of `first, second` - then we could use it both here, and in codePointAt (and potentially other places)

spec.html
```diff
@@ -31010,10 +31011,10 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
             1. If _unicode_ is *false*, return _index_ + 1.
             1. Let _length_ be the number of code units in _S_.
             1. If _index_ + 1 &ge; _length_, return _index_ + 1.
-            1. Let _first_ be the numeric value of the code unit at index _index_ within _S_.
-            1. If _first_ &lt; 0xD800 or _first_ &gt; 0xDBFF, return _index_ + 1.
-            1. Let _second_ be the numeric value of the code unit at index _index_ + 1 within _S_.
-            1. If _second_ &lt; 0xDC00 or _second_ &gt; 0xDFFF, return _index_ + 1.
+            1. Let _first_ be the the code unit at index _index_ within _S_.
```
mathiasbynens: typo: `the the`

gibson042: Fixed.

spec.html
ljharb: It’d be nice to get that in this PR, if possible ;-)

gibson042: Added in 9105de57f9ab1ada4337723125d52db4bc1ce385, though I'm not sure whether the new operation belongs with UTF16Encoding and UTF16Decode under [Section 10.1: Source Text](https://tc39.github.io/ecma262/#sec-source-text) or somewhere else, or even whether all three should move to something like a new "Code Point Operations" section under [Section 7: Abstract Operations](https://tc39.github.io/ecma262/#sec-abstract-operations). Please share your thoughts.

ljharb: Thanks, this is great. i don’t have an opinion on where it lives; alongside the utf16 ops seems fine for now.

spec.html
```diff
@@ -9337,6 +9337,20 @@ <h1>Static Semantics: UTF16Decode ( _lead_, _trail_ )</h1>
         1. Return the code point _cp_.
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-codepointat" aoid="CodePointAt">
+      <h1>Static Semantics: CodePointAt ( _string_, _position_ )</h1>
+      <p>The abstract operation CodePointAt interprets a String _string_ as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, and reads from it a single code point starting with the code unit at index _position_. When called, the following steps are performed:</p>
```
mathiasbynens: Nit: `UTF-16 encoded` → `UTF-16-encoded`

gibson042: This one I will leave for now, because the "UTF-16 encoded" text is common to several other parts of the spec (e.g., [ToNumber](https://tc39.github.io/ecma262/#sec-tonumber-applied-to-the-string-type) and many subsections of [String objects] (https://tc39.github.io/ecma262/#sec-string-objects)).

jmdyck:
I generally like the refactoring here, as it increases the encapsulation of ECMAScript's version of UTF-16 decoding. However, the lines that compare `the numeric value of _cp_` to `0xFFFF` are a bit of a leak in the abstraction.

Also, it seems a bit odd that, because `CodePointAt` calls `UTF16Decode`, `%StringIteratorPrototype%.next` has to then immediately call `UTF16Encoding` because it's only interested in the code units.

I like @ljharb's original suggestion where the abstract op returned the concatenation or a List of the contributing code units. (Not sure what you'd call it, maybe `UTF16CodeUnitsAt`.) We could still have `CodePointAt()`.

gibson042:
The operations have similar but distinct requirements. %StringIteratorPrototype%.next needs the count of code units to slice, AdvanceStringIndex needs the count for index advancement, String.prototype.codePointAt needs a numeric code point value, and Encode needs everything (the count for index advancing and a numeric value for unpaired surrogate checking and UTF-8 encoding). The 1) CodePointAt operation I've just added reports the code point and lets the operations that need count infer it by comparison to the maximum single-code-unit value, which is a bit messy but IMO no worse than 2) $hardToName returning a List of code units and letting the operations that need code unit count get it from the number of elements and the operations that need a code point get it from UTF16Decode. I suppose the other alternatives would be 3) CodePointAt returning a Record with both kinds of data, or 4) $hardToName2 returning just a count of code units. I like option 3, but it seems like overkill to me so I have fallen back on 1. But if there is consensus on another alternative, I'll make the change.

ljharb:
Option 3 does sound nice.

jmdyck:
> Encode needs everything (the count for index advancing and a numeric value for unpaired surrogate checking and UTF-8 encoding).

Yeah, Encode is the real problem, because it's the only one that needs information at multiple "levels".

> I like option 3 [returning a Record], but it seems like overkill to me

I suspect it wouldn't be that bad.

You could also fold in Encode's lone surrogate checking (another leak in the abstraction): a field of the record could indicate whether a properly-encoded code point was found. That would almost completely encapsulate ECMAScript's version of UTF-16 decoding.

spec.html
```diff
@@ -23704,16 +23718,10 @@ <h1>Runtime Semantics: Encode ( _string_, _unescapedSet_ )</h1>
...
-                  1. Let _V_ be the code point with the same numeric value as code unit _C_.
-                1. Else,
-                  1. Increase _k_ by 1.
-                  1. If _k_ equals _strLen_, throw a *URIError* exception.
-                  1. Let _kChar_ be the code unit at index _k_ within _string_.
-                  1. If _kChar_ is not a <emu-xref href="#trailing-surrogate"></emu-xref>, throw a *URIError* exception.
-                  1. Let _V_ be UTF16Decode(_C_, _kChar_).
-                1. Let _Octets_ be the List of octets resulting by applying the UTF-8 transformation to _V_.
+                1. Let _cp_ be ! CodePointAt(_string_, _k_).
+                1. If _cp_ has the same numeric value as a <emu-xref href="#leading-surrogate"></emu-xref> or <emu-xref href="#trailing-surrogate"></emu-xref>, throw a *URIError* exception.
```
michaelficarra: I would word this as

```
1. If the numeric value of _cp_ is a <emu-xref href="#leading-surrogate"></emu-xref> or a <emu-xref href="#trailing-surrogate"></emu-xref>, throw a *URIError* exception.
```

spec.html
```diff
@@ -23704,16 +23718,10 @@ <h1>Runtime Semantics: Encode ( _string_, _unescapedSet_ )</h1>
...
+                1. If the numeric value of _cp_ &gt; 0xFFFF, increase _k_ by 1.
```
michaelficarra: We mix our terminology with "increase _k_ by 1" and "increment _k_ [by 1]" throughout the spec. @tc39/ecma262-editors should we normalise this phrasing? I prefer "increment *k*" (dropping the "by *n*" or "*n* times" for the case where *n* is 1).

ljharb: Let’s use whatever’s most common in this PR; please file a separate issue to track overall normalization.

jmdyck: (See #1266 for some related discussion.)

gibson042:
OK, updated CodePointAt to return a Record: https://github.com/tc39/ecma262/compare/21e0abbdd16321e05a27769f8513d4fb5dea2e93...f2f6a60cee1731b38b9f37522be22dad249700f3

The logic now uses multi-statement conditional blocks to avoid duplicating the unpaired surrogate return value, but could also be re-linearized upon request:
> 1. If _first_ is a trailing surrogate or at the end of the string, return $unpaired.
> 1. Read _second_.
> 1. If _second_ is not a trailing surrogate, return $unpaired.
> 1. Let _cp_ be UTF16Decode(_first_, _second_).
> 1. Return $surrogatePair.

spec.html
```diff
@@ -28295,7 +28306,7 @@ <h1>String.prototype.charCodeAt ( _pos_ )</h1>
       <emu-clause id="sec-string.prototype.codepointat">
         <h1>String.prototype.codePointAt ( _pos_ )</h1>
         <emu-note>
-          <p>Returns a nonnegative integer Number less than 0x110000 that is the code point value of the UTF-16 encoded code point (<emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>) starting at the string element at index _pos_ within the String resulting from converting this object to a String. If there is no element at that index, the result is *undefined*. If a valid UTF-16 <emu-xref href="#surrogate-pair"></emu-xref> does not begin at _pos_, the result is the code unit at _pos_.</p>
+          <p>Returns a nonnegative integer Number less than or equal to 0x10FFFF that is the code point value of the UTF-16 encoded code point (<emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>) starting at the string element at index _pos_ within the String resulting from converting this object to a String. If there is no element at that index, the result is *undefined*. If a valid UTF-16 <emu-xref href="#surrogate-pair"></emu-xref> does not begin at _pos_, the result is the code unit at _pos_.</p>
```
ljharb: Is the “equal to 0x10FFFF” part a normative change, or a correction in the prose?

mathiasbynens: “less than 0x110000” === “less than or equal to 0x10FFFF”, similar to how “less than 3” === “less than or equal to 2”.

It’s just easier to understand/read this way, since Unicode ranges from U+000000 to U+10FFFF. That is, `0x10FFFF` is more easily recognized than `0x110000` in this context.

ljharb: ah, duh, thanks :-) i misread it.

spec.html
```diff
@@ -9337,6 +9337,24 @@ <h1>Static Semantics: UTF16Decode ( _lead_, _trail_ )</h1>
...
+      <emu-alg>
+        1. Let _size_ be the length of _string_.
+        1. Assert: _position_ &ge; 0 and _position_ &lt; _size_.
+        1. Let _first_ be the code unit at index _position_ within _string_.
+        1. Let _cp_ be the code point whose numeric value is that of _first_.
+        1. If _first_ is not a <emu-xref href="#leading-surrogate"></emu-xref> or <emu-xref href="#trailing-surrogate"></emu-xref>, return a new Record { [[CodePoint]]: _cp_, [[CodeUnitCount]]: 1, [[IsUnpairedSurrogate]]: *false* }.
+        1. If _first_ is a <emu-xref href="#leading-surrogate"></emu-xref> and _position_ + 1 &lt; _size_, then
+          1. Let _second_ be the code unit at index _position_ + 1 within _string_.
+          1. If _second_ is a <emu-xref href="#trailing-surrogate"></emu-xref>, then
+            1. Let _cp_ be ! UTF16Decode(_first_, _second_).
```
jmdyck: The previous `_cp_` is still "in scope", so this step should either use "Set" or a different alias.

spec.html
```diff
@@ -29107,14 +29115,9 @@ <h1>%StringIteratorPrototype%.next ( )</h1>
...
-            1. If _first_ &lt; 0xD800 or _first_ &gt; 0xDBFF or _position_ + 1 = _len_, let _resultString_ be the String value consisting of the single code unit _first_.
-            1. Else,
-              1. Let _second_ be the numeric value of the code unit at index _position_ + 1 within the String _s_.
-              1. If _second_ &lt; 0xDC00 or _second_ &gt; 0xDFFF, let _resultString_ be the String value consisting of the single code unit _first_.
-              1. Else, let _resultString_ be the string-concatenation of the code unit _first_ and the code unit _second_.
-            1. Let _resultSize_ be the number of code units in _resultString_.
-            1. Set _O_.[[StringIteratorNextIndex]] to _position_ + _resultSize_.
+            1. Let _cp_ be ! CodePointAt(_s_, _position_).
+            1. Set _O_.[[StringIteratorNextIndex]] to _position_ + _cp_.[[CodeUnitCount]].
+            1. Let _resultString_ be the String value of length _cp_.[[CodeUnitCount]] equal to the substring of _s_ consisting of the code units at indices _position_ (inclusive) through _O_.[[StringIteratorNextIndex]] (exclusive).
```
```suggestion
            1. Let _resultString_ be the String value containing _cp_.[[CodeUnitCount]] consecutive code units from _s_ beginning with the code unit at index _position_.
```
Same effect but a bit simpler.

Also, you could then swap the steps back so that `_O_.[[StringIteratorNextIndex]]` is set last, which I think makes slightly more sense.

jmdyck:
> OK, updated CodePointAt to return a Record: 

Yeah, I like this approach.

> The logic now uses multi-statement conditional blocks to avoid duplicating the unpaired surrogate return value, but could also be re-linearized upon request:

Personally, I think the suggested linearization would make the operation easier to understand. Duplicating the unpaired return doesn't bother me.

One thing that I think might increase readability would be to put each 'return' on its own (sub)step, because then:
- depending on window width, you're more likely to get each Record constructor on a single line, and
- the Record constructors will (mostly) line up vertically, making it easier to compare them.

spec.html
gibson042: Updated.

spec.html
gibson042: Resolved by switching back to a linear approach.

jmdyck:
>  Rename [[CodeUnitCount]] to [[Length]] for better small-window rendering

I've got mixed feelings about this, because `CodeUnitCount` was really clear about its meaning. Could you add a sentence to the preamble for `CodePointAt`, defining `[[Length]]`?

ljharb:
I’m not super concerned about small window rendering, but I’m very concerned about self-documenting names - maybe we could flip that one back?

gibson042:
OK, fair enough. Reverted.

Issue #347, opened 2016-02-02
Integer Indexed Exotic Objects inconsistent wrt Reflect.get/set/has


verwaest:
According to the spec, `[[get]]` (9.4.5.4) and `[[set]`] (9.4.5.5) only ignore CanonicalNumericIndexStrings if `SameValue(O, Receiver)`. `[[HasProperty]]` however does not have such a check since it doesn't even get passed in the Receiver.

```
var o = {__proto__: Int32Array(100)};
Object.prototype[1.3] = 100;
// In 9.4.5.4 we don't go down IntegerIndexedElementGet since
// SameValue(O, Receiver) is false.
assertEquals(100, o[1.3])
// In 9.4.5.2 we do return false since the CanonicalNumericIndexString
// branch isn't guarded by SameValue(O, Receiver).
assertFalse(Reflect.has(o, 1.3));
```

This is inconsistent. I see two options to make this more consistent:
1) Optionally pass in Receiver to Reflect.has, or
2) remove `SameValue(O, Receiver)` from `[[get]]` and `[[set]]`.

I personally prefer the second, since we can't pre-check whether target is receiver in the context of `Reflect.get`. You can easily do `Reflect.get(target, <CanonicalNumericIndexString>, typed_array_in_prototype_chain_of_target)`, causing the `[[get]]` to be blocked once we reach the typed array.

allenwb:
I think the second alternative is probably a good idea.

The receiver check is enforcing that typed array element access do not occur when a typed array is access via prototype lookup. 

When this was specified we were trying to clean up array access corner cases and implementation differences of typed arrays. Apparently at the time I thought that inheriting access to typed array elements was one of the things we wanted to precluded.

However, It isn't clear to me now why this case is important and it does create the inconsistency you have identified. I suspect that at least some implementation have not implemented it which case it is probably pretty safe to change.  However, somebody should test the current status of implementions WRT the receiver check.

rossberg-chromium:
V8 does not implement it yet, which is why this came up.

evilpie:
SpiderMonkey never had this check: https://bugzilla.mozilla.org/show_bug.cgi?id=1216621

PR #1556, opened 2019-05-28
Normative: Guard IntegerIndexedElementSet with receiver check

### LINKED ISSUE #1339: Integer indexed exotic object [[Set]] can not ignore receiver
The change from #347 is not web compatible. I implemented the new behavior of ignoring the receiver in [[Set]] in Firefox and this turns out to be [not web-compatible](https://bugzilla.mozilla.org/show_bug.cgi?id=1502889). No other browser seems to have implemented this behavior, probably because there is _no_ test262 test for it.

A  reduced example that fails with this change:
```
function Vec3(x, y, z) {
  this[0] = x;
  this[1] = y;
  this[2] = z;
}

Vec3.prototype = new Float32Array(3);

Vec3.prototype.set = function(x, y, z) {
  this[0] = x;
  this[1] = y;
  this[2] = z;
}

var vec = new Vec3(1, 2, 3);
var vec2 = new Vec3(4, 5, 6);
console.log(vec[0], vec[1], vec[2]) // 4, 5, 6 (shared prototype changed)
```

### LINKED ISSUE #1541: Typed array instance used as prototype incorrect perform [[Get]]/[[Set]]
Using a Typed Array (or most any exotic object) as as a prototype is weird.  But their specification
 of [\[\[Get\]\]](https://tc39.github.io/ecma262/#sec-integer-indexed-exotic-objects-get-p-receiver) and [\[\[Set\]\]](https://tc39.github.io/ecma262/#sec-integer-indexed-exotic-objects-set-p-v-receiver) does not correctly deal with the case where the _Rceiver_ argument is not the same object as the exotic array.  In this case, a [[Put]] can actually mutate the prototype type object.   Consider:

```js
let o = new Int8Array(1)
Object.setPrototypeOf(Array.prototype, o);
const a = [];  //a inherits indirectly from o
a[0] = 4;  //according to current spec. this will set o[0] to 4.
```

The problem is that [9.4.5.5](https://tc39.github.io/ecma262/#sec-integer-indexed-exotic-objects-set-p-v-receiver) does not check if _O_ is different from _Receiver_ and calls IntegerIndexElementSet to mutate _O_ rather than _Receiver_ (which would also be wrong, but for other reasons.

The fix is to change step 2 to:

> if SameValue(_O_,_Reciever_) and Type(_P_) is String, then

The same change should also be made to [9.4.5.4](https://tc39.github.io/ecma262/#sec-integer-indexed-exotic-objects-get-p-receiver)

shvaikalesh:
Closes #1339.
Closes #1541.

### Recap

In ES6, both [`[[Get]]`](https://262.ecma-international.org/6.0/#sec-integer-indexed-exotic-objects-get-p-receiver) and [`[[Set]]`](https://262.ecma-international.org/6.0/#sec-integer-indexed-exotic-objects-set-p-v-receiver) TypedArray methods had the receiver checks, which were removed in #347 due to consistency with [`[[HasProperty]]`](https://262.ecma-international.org/6.0/#sec-integer-indexed-exotic-objects-hasproperty-p). Removing receiver check from [`[[Set]]`](https://262.ecma-international.org/6.0/#sec-integer-indexed-exotic-objects-set-p-v-receiver) turned out to be [web-incompatible (SM bug)](https://bugzilla.mozilla.org/show_bug.cgi?id=1502889), breaking the 🚀 NASA's WorldWind [web app](https://worldwind.earth/explorer/) and [SDK](https://worldwind.arc.nasa.gov/web/examples/).

I've just checked the links in [Firefox 63 (release binaries)](https://ftp.mozilla.org/pub/firefox/releases/63.0/): the web app and SDK examples still don't work. While there are no other sites that are broken, we can't just fix the WorldWind: it's an educational material that is also distributed offline and viewed via `file:///` protocol.

The breakage was caused by TypedArray instances being in `[[Prototype]]` chain (presumably, for default values) of various objects like [`Matrix`](https://github.com/NASAWorldWind/WebWorldWind/blob/685b44d15afa5ea477d9d071a7846848119ec956/src/geom/Matrix3.js#L70) or [`Vector`](https://github.com/NASAWorldWind/WebWorldWind/blob/685b44d15afa5ea477d9d071a7846848119ec956/src/geom/Vec3.js#L54) instances.

### Proposed change

This PR attempts to specify the most desired behavior while providing the minimum required for web-compatibility: [`OrdinarySet`](https://tc39.es/ecma262/#sec-ordinaryset) is performed for canonical numeric strings that are valid indices only.

* Consistent with [`[[Get]]`](https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects-get-p-receiver) and [`[[HasProperty]]`](https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects-hasproperty-p) TypedArray methods regarding invalid indices: prototype traversal in stopped.
* Consistent with other `[[Set]]` overrides like of [mapped arguments](https://tc39.es/ecma262/#sec-arguments-exotic-objects-set-p-v-receiver) or [legacy platform objects](https://heycam.github.io/webidl/#legacy-platform-object-set) regarding altered receivers: a simple reference check is performed.
* No value coercion is performed for invalid indices if receiver is altered, which is an expected behavior because coercion result would be unused. #2332 proposes to drop value coercion for direct `[[Set]]` as well.

### Implementation status

* V8 follows the proposed spec except that it performs value coercion even for invalid indices.
* SM follows the proposed spec for valid indices, but creates property on receiver for invalid indices, without traversing prototype chain further.
* JSC doesn't currently invoke `[[Set]]` overrides during prototype chain traversal, and gets `Reflect.set` wrong for non-ordinary `[[Set]]`. [This bug](https://bugs.webkit.org/show_bug.cgi?id=217916) will fix non-index `[[Set]]`, including canonical numeric strings that are invalid indices. Indexed `[[Set]]` will be adjusted in a follow-up, as it's quite different and unaware of receiver.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -14648,8 +14648,10 @@
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
-              1. Perform ? IntegerIndexedElementSet(_O_, _numericIndex_, _V_).
-              1. Return *true*.
+              1. If SameValue(_O_, _Receiver_) is *true*, then
+                1. Perform ? IntegerIndexedElementSet(_O_, _numericIndex_, _V_).
+                1. Return *true*.
+              1. If IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *true*.
           1. Return ? OrdinarySet(_O_, _P_, _V_, _Receiver_).
         </emu-alg>
       </emu-clause>


ljharb (dismissed):

shvaikalesh:
Please consider:

```js
let target = [0]
let receiver = [0]
Reflect.set(target, 0, 1, receiver)
console.log({target, receiver})
// => target: [0], receiver: [1]

let typedTarget = new Int8Array(1)
let typedReceiver = new Int8Array(1)
Reflect.set(typedTarget, 0, 1, typedReceiver)
console.log({typedTarget, typedReceiver})
// with this change:
// => typedTarget: [1], typedReceiver: [0]
```

It feels like it is more idiomatic to invoke `IntegerIndexedElement{Get,Set}` on `Receiver` rather than using `SameValue`, though I am yet to came up with more practical example.

ljharb:
With this code:
```js
let o = new Int8Array(1)
Object.setPrototypeOf(Array.prototype, o);
const a = [];  //a inherits indirectly from o
a[0] = 4;  //according to current spec. this will set o[0] to 4.
print(a[0], o[0]);

let target = [0]
let receiver = [0]
Reflect.set(target, 0, 1, receiver)
print(target, receiver)

let typedTarget = new Int8Array(1)
let typedReceiver = new Int8Array(1)
Reflect.set(typedTarget, 0, 1, typedReceiver)
print(typedTarget, typedReceiver)
```
i get these results:
```sh
#### Chakra
4 0
0 1
0 1

#### JavaScriptCore
4 0
0 1
1 0

#### SpiderMonkey
4 0
0 1
0 1

#### V8
4 0
0 1
0 1

#### V8 --harmony
4 0
0 1
0 1
```

so there's actually a variance in the latter case - only JSC returns `1, 0` where the others return `0, 1`. This may need consensus after all.

evilpie:
I am probably misreading the change, but doesn't the return in Step 2.b.ii we never set anything in this case?

akamsmrhill:
I am here!¡!

rkirsling:
Same question as @evilpie: Based on @ljharb's third test case, it seems like we set _either_ `O` or `Receiver`, no?

shvaikalesh:
That's a great question! We don't set anything if receiver is altered, and stop traversing the prototype chain. It diverges from V8 and SM, which currently implement something like:

```diff
    1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
    1. If _numericIndex_ is not *undefined*, then
-     1. Perform ? IntegerIndexedElementSet(_O_, _numericIndex_, _V_).
+     1. If _Receiver_ is an Integer-Indexed exotic object, then
+       1. Perform ? IntegerIndexedElementSet(_Receiver_, _numericIndex_, _V_).
      1. Return *true*.
```

The proposed change is the desired behavior with simpler receiver check, which is present in other `[[Set]]` overrides like [mapped arguments](https://tc39.es/ecma262/#sec-arguments-exotic-objects-set-p-v-receiver) or [legacy platform objects](https://heycam.github.io/webidl/#legacy-platform-object-set). This distinction (`SameValue` vs type check) is only observable via `Reflect.set`, making the proposed change web-compatible.

Also, calling `IntegerIndexedElementSet` on the receiver is quite weird, and it doesn't align with `[[Get]]`:

```js
let t = new Int8Array([0]);
let r = new Int8Array([1]);

Reflect.get(t, 0, r); // => 0 (proposed & current spec / V8 / SM), which is t[0]
Reflect.set(t, 0, 2, r); // => true (proposed spec / V8 / SM)
t[0]; // => 0 (proposed spec / V8 / SM)
r[0]; // => 0 (proposed spec), but 2 (V8 / SM) 
```

spec.html
```diff
@@ -11785,8 +11785,10 @@ <h1>[[Set]] ( _P_, _V_, _Receiver_ )</h1>
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
-              1. Perform ? IntegerIndexedElementSet(_O_, _numericIndex_, _V_).
-              1. Return *true*.
+              1. If ! SameValue(_O_, _Receiver_) is *true*, then
+                1. Perform ? IntegerIndexedElementSet(_O_, _numericIndex_, _V_).
+                1. Return *true*.
+              1. If ! IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *true*.
```
bakkot, line 11791: So if you have an integer index which is in bounds for O, but O and Receiver are different values, you'll now fall through to OrdinarySet. Is that intended?

syg: Honestly I don't remember what the behavior we got consensus for was, but this PR currently doesn't have interop wrt in-bounds/OOB accesses.

In-bounds falls through to OrdinarySet in SM and V8. OOB falls through to OrdinarySet in SM only. JSC currently always sets the value on the target instead of the receiver.

```javascript
// test-iieo-receiver-mismatch.js
let target = new Int8Array(1);
let receiver = [0];

// In-bounds
Reflect.set(target, 0, 1, receiver);
print('after in-bounds set:', target, receiver);

// OOB
Reflect.set(target, 2, 1, receiver);
print('after OOB set:', target, receiver);
```

```
$ eshost ./test-iieo-receiver-mismatch.js
#### ch
after in-bounds set: 0 1
after OOB set: 0 1,,1

#### jsc
after in-bounds set: 1 0
after OOB set: 1 0

#### sm
after in-bounds set: 0 1
after OOB set: 0 1,,1

#### v8
after in-bounds set: 0 1
after OOB set: 0 1

#### xs
after in-bounds set: 1 0
after OOB set: 1 0
```

syg:
I'll try to jog my memory for what the consensus was. Failing that we might need to relitigate this, unfortunately.

shvaikalesh:
@syg The consensus for TypedArray's `[[Set]]` was to invoke `OrdinarySet` if receiver is altered, for all indices, including OOB: https://docs.google.com/presentation/d/1Cg7XDFmq2_Aa7h1ZNdgrmT_UCIm0w02P9Z7Krht46Kw/edit#slide=id.g75ac69d4fb_0_131.

That matches SpiderMonkey behaviour, while this PR matches V8's. Neither proposals have interop, but both are web-compatible, so in this PR I'm suggesting to match TypedArray's `[[Get]]` behaviour to not suppress further prototype lookup for invalid indices.

So formally this PR must receive committee's approval.

@codehag Not only the `Reflect.set`, but also code like `({ __proto__: new Int32Array(10) })[0] = 1`. I can't come up with a use case here, so the goal here is to spec something idiomatic while ensuring NASA's WorldWind [web app](https://worldwind.earth/explorer/), which cares only about in-bounds indices, won't break 😄.

syg (approved):

bakkot (approved):

PR #1559, opened 2019-05-29
Editorial: Make asserts in Else clauses explicit

### LINKED ISSUE #1419: In algorithms, change confusing ”Else foo” into ”Else, Assert: foo”
In the spec, there’s a bunch of instances of:

> 1. Else _x_ is "bar",
>     1. ...

This is confusing, because ”_x_ is "bar"” is an assertion without the word ”assert”, and it can be confused with an assignment without the word ”let”, as it did happen in https://github.com/tc39/proposal-regexp-lookbehind/pull/7#discussion_r151567046.

I suggest to replace it systematically with:

> 1. Else,
>     1. Assert: _x_ is "bar".
>     1. ...

shvaikalesh:
Closes #1419.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3210,7 +3210,8 @@
               1. Assert: In this case, _base_ will never be *undefined* or *null*.
               1. Set _base_ to ! ToObject(_base_).
             1. Return ? _base_.[[Get]](GetReferencedName(_V_), GetThisValue(_V_)).
-          1. Else _base_ must be an Environment Record,
+          1. Else,
+            1. Assert: _base_ is an Environment Record.
             1. Return ? _base_.GetBindingValue(GetReferencedName(_V_), IsStrictReference(_V_)) (see <emu-xref href="#sec-environment-records"></emu-xref>).
         </emu-alg>
         <emu-note>
@@ -3237,7 +3238,8 @@
             1. Let _succeeded_ be ? _base_.[[Set]](GetReferencedName(_V_), _W_, GetThisValue(_V_)).
             1. If _succeeded_ is *false* and IsStrictReference(_V_) is *true*, throw a *TypeError* exception.
             1. Return.
-          1. Else _base_ must be an Environment Record,
+          1. Else,
+            1. Assert: _base_ is an Environment Record.
             1. Return ? _base_.SetMutableBinding(GetReferencedName(_V_), _W_, IsStrictReference(_V_)) (see <emu-xref href="#sec-environment-records"></emu-xref>).
         </emu-alg>
         <emu-note>
@@ -3478,7 +3480,9 @@
         1. If Type(_input_) is Object, then
           1. If _PreferredType_ is not present, let _hint_ be `"default"`.
           1. Else if _PreferredType_ is hint String, let _hint_ be `"string"`.
-          1. Else _PreferredType_ is hint Number, let _hint_ be `"number"`.
+          1. Else,
+            1. Assert: _PreferredType_ is hint Number.
+            1. Let _hint_ be `"number"`.
           1. Let _exoticToPrim_ be ? GetMethod(_input_, @@toPrimitive).
           1. If _exoticToPrim_ is not *undefined*, then
             1. Let _result_ be ? Call(_exoticToPrim_, _input_, &laquo; _hint_ &raquo;).
@@ -4436,7 +4440,8 @@
         1. If the _LeftFirst_ flag is *true*, then
           1. Let _px_ be ? ToPrimitive(_x_, hint Number).
           1. Let _py_ be ? ToPrimitive(_y_, hint Number).
-        1. Else the order of evaluation needs to be reversed to preserve left to right evaluation,
+        1. Else,
+          1. NOTE: The order of evaluation needs to be reversed to preserve left to right evaluation.
           1. Let _py_ be ? ToPrimitive(_y_, hint Number).
           1. Let _px_ be ? ToPrimitive(_x_, hint Number).
         1. If Type(_px_) is String and Type(_py_) is String, then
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -31757,7 +31778,8 @@ <h1>Array.prototype.concat ( ..._arguments_ )</h1>
                   1. Perform ? CreateDataPropertyOrThrow(_A_, ! ToString(_n_), _subElement_).
                 1. Increase _n_ by 1.
                 1. Increase _k_ by 1.
-            1. Else _E_ is added as a single item rather than spread,
+            1. Else,
+              1. Assert: _E_ is added as a single item rather than spread.
```
zenparsing: A couple of these seem like `NOTE`s rather than psuedo-code assertions. @ljharb what do you think?

spec.html
```diff
@@ -12560,10 +12572,11 @@ <h1>Runtime Semantics: EvaluateCall ( _func_, _ref_, _arguments_, _tailPosition_
           1. If Type(_ref_) is Reference, then
             1. If IsPropertyReference(_ref_) is *true*, then
               1. Let _thisValue_ be GetThisValue(_ref_).
-            1. Else the base of _ref_ is an Environment Record,
+            1. Else,
+              1. Assert: the base of _ref_ is an Environment Record.
               1. Let _refEnv_ be GetBase(_ref_).
               1. Let _thisValue_ be _refEnv_.WithBaseObject().
-          1. Else Type(_ref_) is not Reference,
+          1. Else,
```
ljharb: this seems to have been lost:
```suggestion
          1. Else,
              1. Assert: ! Type(_ref_) is not Reference.
```

spec.html
```diff
@@ -22387,7 +22401,8 @@ <h1>ParseModule ( _sourceText_, _realm_, _hostDefined_ )</h1>
                   1. If _ie_.[[ImportName]] is `"*"`, then
                     1. Assert: This is a re-export of an imported module namespace object.
                     1. Append _ee_ to _localExportEntries_.
-                  1. Else this is a re-export of a single name,
+                  1. Else,
+                    1. Assert: This is a re-export of a single name.
```
ljharb: this should be a note

spec.html
ljharb: good call; commented a few more cases like this.

spec.html
```diff
@@ -6951,7 +6960,8 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
               1. If _Desc_.[[Writable]] is present and _Desc_.[[Writable]] is *true*, return *false*.
               1. If _Desc_.[[Value]] is present and SameValue(_Desc_.[[Value]], _current_.[[Value]]) is *false*, return *false*.
               1. Return *true*.
-          1. Else IsAccessorDescriptor(_current_) and IsAccessorDescriptor(_Desc_) are both *true*,
+          1. Else,
+            1. Assert: IsAccessorDescriptor(_current_) and IsAccessorDescriptor(_Desc_) are both *true*.
```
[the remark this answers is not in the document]
shvaikalesh: 3757e77

spec.html
```diff
@@ -6932,7 +6940,8 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
             1. Assert: _extensible_ is *true*.
             1. If IsGenericDescriptor(_Desc_) is *true* or IsDataDescriptor(_Desc_) is *true*, then
               1. If _O_ is not *undefined*, create an own data property named _P_ of object _O_ whose [[Value]], [[Writable]], [[Enumerable]] and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its default value.
-            1. Else _Desc_ must be an accessor Property Descriptor,
+            1. Else,
+              1. Assert: IsAccessorDescriptor(_Desc_) is *true*.
```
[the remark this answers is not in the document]
shvaikalesh: 3757e77

spec.html
shvaikalesh: 4998520

spec.html
```diff
@@ -6596,7 +6602,8 @@ <h1>RunJobs ( )</h1>
       1. In an implementation-dependent manner, obtain the ECMAScript source texts (see clause <emu-xref href="#sec-ecmascript-language-source-code"></emu-xref>) and any associated host-defined values for zero or more ECMAScript scripts and/or ECMAScript modules. For each such _sourceText_ and _hostDefined_, do
         1. If _sourceText_ is the source code of a script, then
           1. Perform EnqueueJob(`"ScriptJobs"`, ScriptEvaluationJob, &laquo; _sourceText_, _hostDefined_ &raquo;).
-        1. Else _sourceText_ is the source code of a module,
+        1. Else,
+          1. Assert: _sourceText_ is the source code of a module.
```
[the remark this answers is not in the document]
shvaikalesh: 4998520

spec.html
shvaikalesh: 4998520

spec.html
shvaikalesh: This is intentional: it is `Else` clause for `If Type(_ref_) is Reference`, so asserting `Type(_ref_) is not Reference` in there seems useless.

We've removed more of those in #1413.

ljharb: aha, makes sense :-)

spec.html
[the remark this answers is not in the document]
jmdyck: @ljharb: I'm curious why you think this one should be a Note rather than an Assert. The "loop header" says "source texts for scripts and/or modules", so if it's not the source of a script, it must be the source of a module.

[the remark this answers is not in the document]
ljharb: Mainly because it seems like prose describing the effect of being the negative case of the if - but i guess if the corresponding prose works in an if, then this prose would work as an assert.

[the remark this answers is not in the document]
ljharb: Manually reverted it back to an assert.

PR #1560, opened 2019-05-29
in NamedEvaluation of ClassExpression, defer to evaluation

michaelficarra:
This fixes the missing `[[SourceText]]` issue for classes created using unnamed class expressions in a name-inferred context, for example

```js
export default (class {});
```

or

```js
let C = class {};
```

See also #1548.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19929,7 +19929,9 @@
       <p>With parameter _name_.</p>
       <emu-grammar>ClassExpression : `class` ClassTail</emu-grammar>
       <emu-alg>
-        1. Return the result of ClassDefinitionEvaluation of |ClassTail| with arguments *undefined* and _name_.
+        1. Let _closure_ be the result of evaluating this |ClassExpression|.
+        1. Perform SetFunctionName(_closure_, _name_).
+        1. Return _closure_.
       </emu-alg>
     </emu-clause>
 


jmdyck:
So it looks to me like it's the same semantics, except for:
- setting [[SourceText]], and
- calling SetFunctionName outside ClassDefinitionEvaluation instead of inside.

The first is a bug fix. For the second, the question is whether the delay in calling SetFunctionName is observable. I think maybe it is. Consider if a ClassElement defined a property named "name":
- In the status quo, SetFunctionName precedes PropertyDefinitionEvaluation, so the "name" property would designate the declared property.
- In the proposed code, PropertyDefinitionEvaluation precedes SetFunctionName, so the "name" property would be the 'external' name.

Does that make sense?

ljharb:
@jmdyck can you provide some code that would behave differently before and after this change?

jmdyck:
Maybe this?: `let A = class { name() { return "B"; } }; A.name`
What does eshost say about it?

ljharb:
```js
#### Chakra
A

#### JavaScriptCore
A

#### SpiderMonkey
A

#### V8
A

#### V8 --harmony
A
```
(even with `get name() {}`)
are you suggesting this would change it to `B`?

michaelficarra:
@ljharb @jmdyck This is the code you're looking for:

```js
let A = class { static get name() { return "B"; } };
A.name; // "B"
```

And, indeed, this PR would change that behaviour.

michaelficarra:
I'll open an alternative PR that leaves the NamedEvaluation of ClassExpression unique and also assigns `[[SourceText]]`.

michaelficarra:
Opened #1569. Sorry for the delay.

PR #1564, opened 2019-05-31
Normative: Use Number arithmetic for constructing dates and times

### LINKED ISSUE #1087: MakeDay should be more explicit how unreasonably large numbers are handled
[20.3.1.12 MakeDay](https://tc39.github.io/ecma262/#sec-makeday) should use a more precise language than "[...] but if this is not possible (because some argument is out of range), return NaN.". For example [20.3.1.11 MakeTime](https://tc39.github.io/ecma262/#sec-maketime) explicitly requires to use IEEE 754-2008 semantics for its operations.

This test prints different results in all four engines (SpiderMonkey and ChakraCore are almost the same, except Chakra prints the year as "-0000" for the second date, but that's probably a different issue) and I blame it one the imprecise language. 😛 
```js
d = new Date(0);
d.setUTCFullYear(Number.MAX_VALUE, Number.MAX_VALUE);
print(d.toUTCString());

d = new Date(0);
d.setUTCFullYear(Number.MAX_VALUE / 12 + 100, -Number.MAX_VALUE);
print(d.toUTCString());
```

gibson042:
Fixes #1087

Also differentiates the range for time value vs. [[DateValue]], cleans up the text for date and time operations (particularly LocalTZA), and consolidates them where possible.

Can be reviewed commit-by-commit if the merged view is overwhelming.


spec.html
```diff
@@ -26777,88 +26777,83 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
         <p>The exact moment of midnight at the beginning of 01 January, 1970 UTC is represented by the time value *+0*.</p>
         <emu-note>
           <p>The 400 year cycle of the Gregorian calendar contains 97 leap years. This yields an average of 365.2425 days per year, or an average of 31,556,952,000 milliseconds per year under the Gregorian calendar. ECMAScript applies a proleptic Gregorian calendar for all time computations.</p>
           <p>As specified by this section, the maximum year range a Number can represent exactly with millisecond precision is approximately -285,426 to 285,426 years relative to midnight at the beginning of 01 January, 1970 UTC.</p>
-          <p>As specified by this section, the maximum year range a time value can represent is approximately -273,790 to 273,790 years relative to midnight at the beginning of 01 January, 1970 UTC.</p>
+          <p>As specified by this section, the maximum year range a [[DateValue]] can represent is approximately -273,790 to 273,790 years relative to midnight at the beginning of 01 January, 1970 UTC.</p>
```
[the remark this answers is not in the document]
gibson042: Right, and it's actually a critically important aspect here that some time values fall outside the valid range of the [[DateValue]] internal slot of Date instances.

For a simple example, consider LocalTime, which returns a time value with UTC date and time of day equal to the local date and time of day at the instant in time specified by its time value argument. This is equivalent to returning the the sum of its argument and the local UTC offset at that instant, which can exceed the maximum valid [[DateValue]] if the argument is already close to it (e.g., if the local UTC offset at the maximum [[DateValue]] of +275760-09-13T00:00Z is +10:00, then LocalTime(8_640_000_000_000_000) returns 8_640_000_036_000_000). And similar effects apply with negative UTC offsets near the minimum [[DateValue]] boundary.

spec.html
```diff
@@ -26777,88 +26777,83 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
+            1. If _day_ &lt; DaysWithinYearBeforeMonth(9, _inLeapYear_), return 8.
+            1. If _day_ &lt; DaysWithinYearBeforeMonth(10, _inLeapYear_), return 9.
+            1. If _day_ &lt; DaysWithinYearBeforeMonth(11, _inLeapYear_), return 10.
+            1. Return 11.
+          </emu-alg>
+        </emu-clause>
         <p>where</p>
         <emu-eqn id="eqn-DayWithinYear" aoid="DayWithinYear">DayWithinYear(_t_) = Day(_t_) - DayFromYear(YearFromTime(_t_))</emu-eqn>
-        <p>A month value of 0 specifies January; 1 specifies February; 2 specifies March; 3 specifies April; 4 specifies May; 5 specifies June; 6 specifies July; 7 specifies August; 8 specifies September; 9 specifies October; 10 specifies November; and 11 specifies December. Note that <emu-eqn>MonthFromTime(0) = 0</emu-eqn>, corresponding to Thursday, 01 January, 1970.</p>
+        <p>Note that <emu-eqn>MonthFromTime(0) = 0</emu-eqn>, corresponding to Thursday, 01 January, 1970.</p>
```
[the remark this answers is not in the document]
gibson042: I think this renders better as-is, but will change to `<emu-note>` if there are strong feelings about it.

spec.html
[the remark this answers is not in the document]
gibson042: I think that would confuse rather than illuminate. There really is just one concept, but the full range of safe integers is only available for internal operations to avoid rounding errors where the spec calls for UTC offset conversion/DST calculation/etc. The bounds are enforced by TimeClip, which is introduced as the last operation in this overview section. Perhaps the range aspects of "Time Values and Time Range" should be pulled into an explicit subsection and TimeClip should be moved there, making it the first operation rather than the last?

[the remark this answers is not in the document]
spectranaut: I like that suggestion, @gibson042.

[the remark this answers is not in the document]

spec.html
```diff
@@ -26901,89 +26901,96 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
-          = 6 if 181 + InLeapYear(_t_) &le; DayWithinYear(_t_) &lt; 212 + InLeapYear(_t_)
-          = 7 if 212 + InLeapYear(_t_) &le; DayWithinYear(_t_) &lt; 243 + InLeapYear(_t_)
-          = 8 if 243 + InLeapYear(_t_) &le; DayWithinYear(_t_) &lt; 273 + InLeapYear(_t_)
-          = 9 if 273 + InLeapYear(_t_) &le; DayWithinYear(_t_) &lt; 304 + InLeapYear(_t_)
-          = 10 if 304 + InLeapYear(_t_) &le; DayWithinYear(_t_) &lt; 334 + InLeapYear(_t_)
-          = 11 if 334 + InLeapYear(_t_) &le; DayWithinYear(_t_) &lt; 365 + InLeapYear(_t_)
+        <p>Months are identified by an integer in the range 0 to 11, inclusive. A month value of 0 specifies January; 1 specifies February; 2 specifies March; 3 specifies April; 4 specifies May; 5 specifies June; 6 specifies July; 7 specifies August; 8 specifies September; 9 specifies October; 10 specifies November; and 11 specifies December.</p>
+        <p>The position of a month _m_ within a year depends upon whether or not the year is a leap year:</p>
+        <emu-eqn id="eqn-DaysWithinYearBeforeMonth" aoid="DaysWithinYearBeforeMonth">DaysWithinYearBeforeMonth(_m_, _inLeapYear_)
```
ljharb: it's strange to have a variable named "in leap year", which implies it's a boolean, but use it in arithmetic. any reason this couldn't be a boolean, and the addition here done with `ToNumber()`?

spec.html
```diff
@@ -27046,58 +27055,48 @@ <h1>Hours, Minutes, Second, and Milliseconds</h1>
...
-          1. Let _ym_ be _y_ + floor(_m_ / 12).
-          1. Let _mn_ be _m_ modulo 12.
-          1. Find a value _t_ such that YearFromTime(_t_) is _ym_ and MonthFromTime(_t_) is _mn_ and DateFromTime(_t_) is 1; but if this is not possible (because some argument is out of range), return *NaN*.
-          1. Return Day(_t_) + _dt_ - 1.
+          1. Let _d_ be ! ToInteger(_date_).
+          1. Let _wholeYears_ be _y_ `+` floor(_m_ / 12), performing the arithmetic according to IEEE 754-2008 rules (that is, as if using the ECMAScript `+` operator).
+          1. If _wholeYears_ is not finite, return *NaN*.
+          1. Let _monthWithinYear_ be _m_ modulo 12.
+          1. Let _inLeapYear_ be InLeapYear(TimeFromYear(_wholeYears_)).
+          1. Let _day_ be DayFromYear(_wholeYears_) `+` DaysWithinYearBeforeMonth(_monthWithinYear_, _inLeapYear_) `+` _d_ `-` 1, performing the arithmetic according to IEEE 754-2008 rules (that is, as if using the ECMAScript operators `+` and `-`).
```
ljharb: if you use the subscript `𝔽`, i don't think you need the "performing …" prose, since that makes them Numbers?

jmdyck: You don't even need the subscript, since it's optional when the operands are Number values, which they presumably are, according to [5.2.5](https://tc39.es/ecma262/#sec-mathematical-operations). And the operators don't need backticks either.

ljharb: even better

spec.html
gibson042: This stems from the [definition of the operation](https://tc39.es/ecma262/#eqn-InLeapYear):
> The leap-year function is 1 for a time within a leap year and otherwise is zero

Would you consider it an improvement to change its argument from a time value to a year and update its name to e.g. LeapDaysInYear, supporting changes like this?
```diff
- 1. Let _inLeapYear_ be InLeapYear(_t_).
+ 1. Let _leapDaysInYear_ be leapDaysInYear(YearFromTime(_t_)).
```

spec.html
gibson042: Cool; I'll update that. Should multiplication of Number values use `*` or `×`?

spec.html
ljharb: that would certainly seem clearer to me :-)

bakkot:
@gibson042 Since there's a normative component to this change it needs consensus from the committee. I don't believe it's been discussed yet. Are you interested in presenting the normative component? I can try to do so it if not, though I'm not confident I can give a full and correct accounting of the change.

spec.html
```diff
@@ -28095,40 +28102,41 @@ <h1>Week Day</h1>
...
+        <p>When _isUTC_ is *false*, <dfn>LocalTZA( _t_, *false* )</dfn> interprets _t_ as a time value representing a date and time of day. It returns the offset of the local time zone from UTC in milliseconds at the smallest (closest to negative infinity) time value _earliestMatch_ for which either the local date and time of day at _earliestMatch_ is equal to the UTC date and time of day at _t_ or the local date and time of day at _earliestMatch_ + 1 is greater than the UTC date and time of day at _t_. When the result is subtracted from _t_, it should yield the smallest time value for which the local date and time of day are equal to the UTC date and time of day at _t_. If no such time value exists because of local times skipped by positive time zone transitions (e.g. when the daylight saving time starts or the time zone adjustment is increased due to a time zone rule change), then subtracting the result from _t_ should instead yield the time value for which local date and time of day in the UTC offset applicable before the transition are equal to the UTC date and time of day at _t_.</p>
```
vmukhachev, line 28110: Can this be given in pseudo code?
1. Get the list of all time zone transitions for a full supported date range (the timestamps, for which LocalTZA(t, true) != LocalTZA(t, true).
2. iterate the list in sorted order, 
2.1. if localtime < transitionTimestamp + max(LocalTZA(transitionTimestamp - 1), LocalTZA(transitionTimestamp))
2.1.1. return localtime - LocalTZA(transitionTimestamp - 1);
3. return localtime - LocalTZA(<the timestamp of the last transition from the list>, true)

gibson042 (commented):
#2120 actually replaces part of this PR, but I think there are still normative changes in MakeDay and MakeDate that will be observable via e.g. `Date.UTC( 1970, 0, 1 - 1e8,   -BIG, BIG * 60 + 1, 1e8 * 86400 * 2 - 61, 61000 )` (where `BIG` is at or near `Number.MAX_SAFE_INTEGER`).

bakkot:
At the last meeting I asked for (and achieved) consensus that all arithmetic on Date objects was to be done with Number semantics, particularly [calling out](https://docs.google.com/presentation/d/1gePsNmlP2u0pYXm0LWO3d7eM4Q_y5Ozx0qXN1zWOv58/edit#slide=id.g8f7c6b6dbb_0_0) the large number cases, so I think the normative parts are good to go unless they entail changes beyond that.

I was planning to make that particular change as a part of https://github.com/tc39/ecma262/pull/2007 (which makes it unambiguous when arithmetic is being done as if on reals vs as if on Numbers). Unfortunately that PR is going to have a bunch of conflicts with this one; when we were working on it I'd forgotten you already made this PR. I'll need to rebase one of them.

gibson042:
@bakkot Please continue to disregard this PR when addressing arithmetic, it's really more about clarifying Date semantics than about any particular types or order of operations. I'll be happy to rebase it (but please do keep #1087 in mind with your changes).

spec.html
Yaffle: @gibson042 
It is the same expression as you are using:
FInd smallest integer `_earliestMatch_` for which `LocalTime(_earliestMatch_) === _t_ || LocalTime(_earliestMatch_ + 1) > _t_`
Return localtime - LocalTime(_earliestMatch_)

where `LocalTime` is the abstract operation mentioned in the spec.

gibson042: Hmm, I'd rather not introduce cyclic references between the text of isUTC and LocalTime.

Yaffle: @gibson042 , it would be better to remove the LocalTZA so and have LocalTime be the thing doing what LocalTZA(t, false) is doing now, but that will be a bigger change.

Yaffle: @gibson042 , it is not cyclic, the LocalTime uses LocalTZA(t, true) branch

michaelficarra:
@gibson042 Is this PR still applicable after #2007?

gibson042:
No, I don't think so. [Simplify MonthFromTime](https://github.com/tc39/ecma262/pull/1564/commits/1dde075353b1eea89225cea063aa229434333f63) and [Dedicate a subsection to time value range](https://github.com/tc39/ecma262/pull/1564/commits/7f6553f8330365313a1bb24a3f9c0abfdfcd8821) and [Refactor InLeapYear(_t_) as LeapDaysInYear(_y_)](https://github.com/tc39/ecma262/pull/1564/commits/e4b9b2d7b664f51e0c6fe1b0e7e713cb907aa770) are editorial changes that could be added independently if deemed desirable, and everything else has already been superseded.

Issue #497, opened 2016-03-26
eliminating implicit conversions to/from Completion Records


jmdyck:
Recap:

Section 6.2.2.2 "Implicit Completion Values" says:

> Unless it is otherwise obvious from the context, an algorithm statement that returns a value that is not a Completion Record, such as: `Return "Infinity"` means the same thing as: `Return NormalCompletion("Infinity")`.

and

> Any reference to a Completion Record value that is in a context that does not explicitly require a complete Completion Record value is equivalent to an explicit reference to the [[Value]] field of the Completion Record value unless the Completion Record is an abrupt completion.

That is, there are implicit conversions to and from Completion Records.

Reaction:

In https://github.com/tc39/ecma262/issues/253#issuecomment-165063648, @rossberg-chromium said:

> The spec should not be optimised for maximum brevity but for maximum clarity. I think the implicit conversion from completion records to their values is actively harmful in that regard.

And in https://github.com/tc39/ecma262/issues/486#issuecomment-201677699, @bterlson said:

> There is a rule in 5.2 [sic] that allows us to treat a completion record as if it were a value but it's something I'd like to get away from.

In https://github.com/tc39/ecma262/issues/253#issuecomment-165134231, I agreed, and said:

> I've thought of an approach that I think would eliminate most of these conversions. I'll raise it in another issue when I have some time.

This post outlines that approach.  (I was hoping to submit a pull request, but now I'm not sure I'll have time in the near future.)

---

The basic idea is to do away with "normal" completion records.

That is, anywhere that an algorithm currently deals with a Completion Record whose [[Type]] is `normal`, instead just use whatever the [[Value]] of that record is. (You can ignore the [[Target]], it's always `empty`.)

Things that would disappear:
- All of 6.2.2.2.
- NormalCompletion()
- Completion()
- Steps of the form `If val is Completion Record, let val be val.[[Value]]` in the definition of ReturnIfAbrupt and "!".

Section 6.2.2, rather than defining "Completion Record", would define "AbruptCompletion Record".

For terminology, you could say something like: a "completion value" is either an "abrupt completion value" (i.e., an AbruptCompletion Record) or a "normal completion value" (i.e., `empty` or an ECMAScript language value).

[...]

PR #1567, opened 2019-06-02
Editorial: Extract IsNonNegativeInteger abstract op

shvaikalesh:
Follow-up of #1534.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -5403,6 +5403,15 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-isnonnegativeinteger" aoid="IsNonNegativeInteger">
+      <h1>IsNonNegativeInteger ( _argument_ )</h1>
+      <p>The abstract operation IsNonNegativeInteger determines if _argument_ is non-negative integer numeric value.</p>
+      <emu-alg>
+        1. If ! IsInteger(_argument_) is *true* and _argument_ &ge; 0, return *true*.
+        1. Otherwise, return *false*.
+      </emu-alg>
+    </emu-clause>
+
     <emu-clause id="sec-ispropertykey" aoid="IsPropertyKey">
       <h1>IsPropertyKey ( _argument_ )</h1>
       <p>The abstract operation IsPropertyKey determines if _argument_, which must be an ECMAScript language value, is a value that may be used as a property key.</p>
@@ -8594,7 +8603,7 @@
       <emu-alg>
         1. Assert: _F_ is an extensible object that does not have a *"length"* own property.
         1. Assert: Type(_length_) is Number.
-        1. Assert: _length_ &ge; 0 and ! IsInteger(_length_) is *true*.
+        1. Assert: ! IsNonNegativeInteger(_length_) is *true*.
         1. Return ! DefinePropertyOrThrow(_F_, *"length"*, PropertyDescriptor { [[Value]]: _length_, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }).
       </emu-alg>
     </emu-clause>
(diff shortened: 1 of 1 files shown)


jmdyck:
Maybe it would be worth defining `IsNonNegativeInteger`.

spec.html
```diff
@@ -32113,9 +32123,9 @@ <h1>FlattenIntoArray ( _target_, _source_, _sourceLen_, _start_, _depth_ [ , _ma
           <emu-alg>
             1. Assert: Type(_target_) is Object.
             1. Assert: Type(_source_) is Object.
-            1. Assert: _sourceLen_ is an integer Number &ge; 0.
-            1. Assert: _start_ is an integer Number &ge; 0.
-            1. Assert: _depth_ is an integer Number, *+&infin;*, or *-&infin;*.
+            1. Assert: ! IsNonNegativeInteger(_sourceLen_) is *true*.
+            1. Assert: ! IsNonNegativeInteger(_start_) is *true*.
+            1. Assert: ! IsNonNegativeInteger(_depth_) is *true*, or _depth_ is either *+&infin;* or *-&infin;*.
```
jmdyck: No, `_depth_` could be a negative integer.

```suggestion
            1. Assert: ! isInteger(_depth_) is *true*, or _depth_ is either *+&infin;* or *-&infin;*.
```

jmdyck: (But with a capital "I": "IsInteger".)

shvaikalesh: Fixed in a0061e5, thank you.

shvaikalesh:
`IsNonNegativeInteger` is indeed tailored for assertions: in algorithms, we mostly deal with values after type coercion and don't need extra checks.

I am neutral regarding it might be an over-abstraction, but I am sure that we should clarify a bit prose like "nonnegative integer" or "integer Number".

Apart from consistency, this change clarifies that a value is number **primitive** and may be larger than 2<sup>32</sup>.

spec.html
```diff
@@ -7620,7 +7630,7 @@ <h1>SetFunctionLength ( _F_, _length_ )</h1>
       <emu-alg>
         1. Assert: _F_ is an extensible object that does not have a `"length"` own property.
         1. Assert: Type(_length_) is Number.
-        1. Assert: _length_ &ge; 0 and ! IsInteger(_length_) is *true*.
+        1. Assert: ! IsNonNegativeInteger(_length_) is *true*.
```
michaelficarra: There's no reason to prefix all of these `IsNonNegativeInteger` uses with `!`. Let's drop them all, please.

shvaikalesh: Thank you for review, please see https://github.com/tc39/ecma262/pull/1529#discussion_r281878951.

If it was up to me, I wouldn't prefix abstract ops that never return abrupt completions with `!`.

michaelficarra: There's no completion records involved here, though. That is, unless you're of the camp that believes all abstract operations return completion records. I guess we'll clear that up with #497, so what we do here doesn't really matter to me. I just think it's ridiculous to litter the spec with `!` when the abstract op being called neither explicitly creates a completion record nor returns a completion record created by another abstract op.

ljharb: It's not a camp; they currently *do* all return completion records - so pending #497 or another change, the ! are needed here.

michaelficarra: I'll just leave this here: https://github.com/tc39/ecma262/issues/496#issuecomment-491603203

spec.html
```diff
@@ -5403,6 +5403,16 @@ <h1>IsInteger ( _argument_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-isnonnegativeinteger" aoid="IsNonNegativeInteger">
+      <h1>IsNonNegativeInteger ( _argument_ )</h1>
+      <p>The abstract operation IsNonNegativeInteger determines if _argument_ is non-negative integer numeric value.</p>
+      <emu-alg>
+        1. If ! IsInteger(_argument_) is *false*, return *false*.
+        1. If _argument_ &ge; 0, return *true*.
+        1. Return *false*.
```
syg: I dislike the three cases here being false, true, false. I'd prefer the following:

1. If IsInteger(_argument_) is true and _argument_ &ge; 0, return true.
1. Otherwise, return false.

If you feel strongly about having 3 steps, then I'd prefer false, false, true, i.e. make step 2 check for < 0.

shvaikalesh: TIL, thank you. Fixed in 8552db7.

ljharb:
@syg If we change our overall convention so that, for example, predicates don't need a prefix, then we can remove them all at that time; at the moment the convention we follow is to always include them in cases like this.

PR #1569, opened 2019-06-04
Normative: set [[SourceText]] in NamedEvaluation of ClassExpression

michaelficarra:
In #1560, we discovered that deferring NamedEvaluation to the evaluation algorithm for ClassExpressions would cause an undesirable change in behaviour. So this PR is the alternative: keep NamedEvaluation of ClassExpression special and set its `[[SourceText]]` slot.

This should be the last missing `[[SourceText]]` slot  :crossed_fingers:.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19991,7 +19991,10 @@
       <p>With parameter _name_.</p>
       <emu-grammar>ClassExpression : `class` ClassTail</emu-grammar>
       <emu-alg>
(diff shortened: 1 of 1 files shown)


jmdyck:
@anba has this change in the first commit of #1490.

ljharb:
That's a larger PR, so it'd be good to land this first :-)

jmdyck:
Okay, wasn't sure about protocol.

PR #1570, opened 2019-06-04
Editorial: remove useless step of ReturnIfAbrupt expansion

michaelficarra:
Since "_hygienicTemp_ is ephemeral and visible only in the steps pertaining to ReturnIfAbrupt", there's no reason to set it in the final step.


spec.html
```diff
@@ -804,7 +804,6 @@ <h1>ReturnIfAbrupt</h1>
         <emu-alg>
           1. Let _hygienicTemp_ be AbstractOperation().
           1. If _hygienicTemp_ is an abrupt completion, return _hygienicTemp_.
-          1. Else if _hygienicTemp_ is a Completion Record, set _hygienicTemp_ to _hygienicTemp_.[[Value]].
```
ljharb: I’m not sure this is accurate; the original algorithm explicitly unwrapped the completion record. See the lines below (817)

jmdyck: Yeah, but in this case, when the result of AbstractOperation() isn't going anywhere (isn't being bound to an alias), the "unwrapping" has no discernible effect.

ljharb: it is, though - `? Foo()` ends up becoming the result of `hygienicTemp`, as if this was a function and there was an implicit `return hygienicTemp`. Perhaps there's better wording we could use here to make that clear?

jmdyck: This is the expansion of the shorthand step
```
1. ReturnIfAbrupt(AbstractOperation()).
```
so the result of AbstractOperation(), if it isn't abrupt, is discarded.

It sounds like you're thinking of a step like
```
1. Let _result_ be ReturnIfAbrupt(AbstractOperation()).
```

michaelficarra: @ljharb Note that this expansion is written in terms of a whole algorithm step, not any part of any algorithm step.

ljharb: ah, i misunderstood. so this isn’t for ?, its for ReturnIfAbrupt?

jmdyck: Right. (The expansions for '?' just replace it with `ReturnIfAbrupt`, so don't get into temporaries or unwrapping.)

ljharb: So then wouldn’t this change the resulting inlining of `?`?

michaelficarra: This is only the case where the value is not used. I think you need to take a look at this section again. The rewrite rules for ReturnIfAbrupt are defined as pattern matches on full algorithm steps, and the rewrite I've changed only applies when the whole step is a call to ReturnIfAbrupt with its argument being an abstract operation. Nobody can be relying on the unobservable setting of a generated temporary variable.

michaelficarra:
This whole thread has got me thinking that it'd probably be worth the effort to look for instances of `ReturnIfAbrupt` that don't pattern match to one of the three cases in [5.2.3.3](https://tc39.github.io/ecma262/#sec-returnifabrupt) or instances of prefix `?` or `!` that don't pattern match to one of the cases in [5.2.3.4](https://tc39.github.io/ecma262/#sec-returnifabrupt-shorthands).

spec.html
jmdyck: It would only affect those occurrences of `?` that expand to a step that says (or is otherwise equivalent to):
```
1. ReturnIfAbrupt(AbstractOperation()).
```
Literally, this would be occurrences of the form:
```
1. ? AbstractOperation()
```
But we have no steps that begin with `?` (outside of the `ReturnIfAbrupt Shorthands` clause).

However, we do have steps of the form:
```
1. Perform ? Foo(...).
```
which presumably expands to
```
1. Perform ReturnIfAbrupt(Foo(...)).
```
which you could argue is "otherwise equivalent" to:
```
1. ReturnIfAbrupt(Foo(...)).
```
whose expansion **is** altered by the proposed change.

But that alteration doesn't change semantics, because, in the original, the result of Foo(), if it isn't abrupt, is discarded anyway.

ljharb: Thanks for explaining. Although it sounds like there might be more refinement possible here, the things I’m concerned about don’t apply, so this change seems good to me!

jmdyck:
> instances of `ReturnIfAbrupt` that don't pattern match to one of the three cases in [5.2.3.3](https://tc39.github.io/ecma262/#sec-returnifabrupt)

It looks like **all** occurrences (outside 5.2.3.3 and 5.2.3.4) match
```
1. ReturnIfAbrupt(_argument_).
```
(The argument is always an alias/metavariable, never an operation-invocation, and the `ReturnIfAbrupt` call is never embedded within some other expression.)

Minor quibble: there's [one occurrence](https://tc39.github.io/ecma262/#sec-ecmascript-function-objects-construct-argumentslist-newtarget) that technically doesn't match, because it's in a one-line `Else`:
```
1. Else, ReturnIfAbrupt(_result_).
```
but of course it would match if we made it an indented sub-block.

----

> or instances of prefix `?` or `!` that don't pattern match to one of the cases in [5.2.3.4](https://tc39.github.io/ecma262/#sec-returnifabrupt-shorthands).

For `?`, it depends on how strictly you want to interpret the cases: both have `?` at the start of the step, which (as stated above) doesn't occur in the spec. However, if you interpret them as talking about those forms occurring anywhere in the step, then it depends on what you think matches `OperationName()` and `_someValue_.OperationName()`.

E.g., There are ~100 occurrences of the form:
```
? _X_.[[InternalMethod]](_args_)
```
So it depends whether you think `[[InternalMethod]]` counts as `OperationName`.

---
For `!`, note that 5.2.3.4 doesn't explicitly say where it's allowed, it just says "Similarly, prefix `!` is used..." and then gives one example of its use. Presumably it means that `!` is allowed in the same contexts where `?` is allowed.

If so, then there are 2 occurrences of the form:
```
! _X_.[[InternalMethod]](_args_)
```
So again, it depends.

jmdyck:
> Actually, I'm no longer sure now whether ReturnIfAbrupt pattern matches full or partial algorithm steps. The wording in that section reads to me like only full algorithm steps apply, but I think the usage throughout the spec is consistent with partial algorithm steps, though I'm not sure exactly how that would work.

Yeah, 5.2.3.{3,4} seem a bit too vague for my taste. I'd prefer, for each construct, a single precise definition that covers all possible usages, rather than a few patterns whose coverage is at best questionable. However, I don't think we have the notation that would allow precise definitions, and I'm doubtful that it'd be worth introducing it.

michaelficarra:
@jmdyck Thanks for doing the initial research! I've confirmed that all of the current usages of `ReturnIfAbrupt` without the shorthand will match the first definition. And many of them are for syntax-directed operations that could use the shorthand as described at the end of 5.2.3.4. I'm betting the authors just didn't know the shorthand was permitted to be used that way.

In my opinion, the way the shorthand notation is defined right now is unacceptably bad. I'd like to see if we could define rewrite rules for the shorthand notation that operate only on a whole step. Something like

> * [*words before*] **?** AbstractOperation([*parameters*]) [*words after*]

gets rewritten to

> * Let *hygienicTemp* be AbstractOperation([*parameters*]).
> * If *hygienicTemp* is an abrupt completion, return *hygienicTemp*.
> * Else if *hygienicTemp* is a Completion Record, set *hygienicTemp* to *hygienicTemp*.[[Value]]. 
> * [*words before*] *hygienicTemp* [*words after*]

@ljharb @bterlson thoughts?

ljharb:
That sounds great to me - essentially, turning `? AbstractOperation(…)` into a do expression :-)

jmdyck:
I was just about to suggest something like that. Except I was going to suggest it for the expansion of `ReturnIfAbrupt`, and retain the idea of `?` being a simple shorthand for `ReturnIfAbrupt`. You're suggesting to sever that connection?

michaelficarra:
Either way works for me.

ljharb:
Seems nice to keep the connection and add better shorthand for ReturnIfAbrupt.

michaelficarra:
I opened a PR for better ReturnIfAbrupt definition in #1573.

michaelficarra:
I think we should merge this without waiting for a resolution on #1573. It's an objective improvement for the time being.

michaelficarra:
Ping @syg @bakkot for a very quick review.

michaelficarra:
Okay that's fair, thanks for the reviews. We'll hold off until #1573 can fix this (literal) nonsense.

PR #1571, opened 2019-06-07
Editorial: normalize wording used to apply syntax-directed operations

michaelficarra:
* Normalised the wording of applying a syntax-directed operation
* Normalised the wording of passing arguments to a syntax-directed operation
* Used `?` shorthand for syntax-directed operations where applicable

I was a bit conflicted between the terse "Operation of |NonTerminal|" and the verbose "the result of applying Operation {for,of,to} |NonTerminal|". I'm happy either way.

I left "the result of evaluating" alone, both for wording normalisation and for usage of `?`. Even though it's just another syntax-directed operation, it is invoked using a special form. If the editors decide that we should include evaluation in this normalisation, I'll make the change.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -724,7 +724,7 @@
       <p>When an algorithm is associated with a production alternative, the alternative is typically shown without any &ldquo;[ ]&rdquo; grammar annotations. Such annotations should only affect the syntactic recognition of the alternative and have no effect on the associated semantics for the alternative.</p>
       <p>Syntax-directed operations are invoked with a parse node and, optionally, other parameters by using the conventions on steps 1, 3, and 4 in the following algorithm:</p>
       <emu-alg>
-        1. Let _status_ be the result of performing SyntaxDirectedOperation of |SomeNonTerminal|.
+        1. Let _status_ be SyntaxDirectedOperation of |SomeNonTerminal|.
         2. Let _someParseNode_ be the parse of some source text.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -10865,13 +10865,13 @@ <h1>Static Semantics: TV and TRV</h1>
             The TRV of <emu-grammar>NotEscapeSequence :: `u` `{` CodePoint [lookahead &lt;! HexDigit] [lookahead != `}`]</emu-grammar> is the sequence consisting of the code unit 0x0075 (LATIN SMALL LETTER U) followed by the code unit 0x007B (LEFT CURLY BRACKET) followed by the code units of the TRV of |CodePoint|.
           </li>
           <li>
-            The TRV of <emu-grammar>DecimalDigit :: one of `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`</emu-grammar> is the SV of the |SourceCharacter| that is that single code point.
+            The TRV of <emu-grammar>DecimalDigit :: one of `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`</emu-grammar> is the SV of |SourceCharacter| that is that single code point.
```
jmdyck: I don't think you meant to delete "the" here. (You didn't in similar contexts on line 10877 and 10898.)

michaelficarra: Yep, that was a mistake. Fixed.

Issue #1572, opened 2019-06-07
! shorthand for completion record [[Value]] extraction is unnecessary


michaelficarra:
In [§5.2.3.4 ReturnIfAbrupt Shorthands](https://tc39.github.io/ecma262/#sec-returnifabrupt-shorthands), we define a shorthand `!` that's not really related to ReturnIfAbrupt. This shorthand asserts that the completion record is a normal completion and extracts its `[[Value]]`. But at the end of [§5.2.3.1 Implicit Completion Values](https://tc39.github.io/ecma262/#sec-implicit-completion-values), we say

> Any reference to a Completion Record value that is in a context that does not explicitly require a complete Completion Record value is equivalent to an explicit reference to the [[Value]] field of the Completion Record value unless the Completion Record is an abrupt completion.

I think we should either

1. remove this clause and use the `!` shorthand or explicit `[[Value]]` extraction throughout the spec, or
1. remove the `!` shorthand, its usages, and any explicit `[[Value]]` extraction

claudepache:
The primary purpose of `!` is to assert that the operation didn’t end abruptly. Without `!`, you don’t know whether an abrupt completion is returned.

Also, some people would love to make obsolete the paragraph at the end of §5.2.3.1 Implicit Completion Values you cited, through the systematic use of the `?`/`!` pair.

ljharb:
I'm a bit unclear what the issue is - indeed, the `!` is critical to make it clear that a completion record is not abrupt. Perhaps the implicit unwrapping overall, means we don't need an explicit unwrap step for `!` - or perhaps we should remove the implicit overall unwrapping, and use explicit unwrapping everywhere - is that what you're suggesting?

jmdyck:
> Also, some people would love to make obsolete the paragraph at the end of §5.2.3.1 Implicit Completion Values you cited, through the systematic use of the `?`/`!` pair.

(And some people would like to make it obsolete via somewhat different means.)

michaelficarra:
@ljharb Yes, we should either fully buy into that implicit unwrapping everywhere or (my preference) drop that paragraph and fix up anywhere that is relying on it. If the latter, we should probably move the `!` shorthand section to the end of the completion values section instead of being lumped in with ReturnIfAbrupt.

@jmdyck Care to elaborate?

ljharb:
I’m fine with the implicit unwrapping in general, but i like it being explicit in ? and !

TimothyGu:
[...]

PR #1573, opened 2019-06-07
change how ReturnIfAbrupt and its shorthands are specified

michaelficarra:
In #1570, we discussed how the way ReturnIfAbrupt and its shorthands are specified is unclear. I've made it clear that the ReturnIfAbrupt equivalence is on sequences of algorithm steps and the shorthand is a simple replacement in any context.

I've updated the `!` shorthand in this PR as well, but in #1572 I propose that we remove it, so we should resolve that first.

I recommend reading the current and proposed spec text separately to review. The diff will not be very helpful.

This will obsolete #1570.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -787,68 +787,97 @@
 
       <emu-clause id="sec-returnifabrupt" aoid="ReturnIfAbrupt">
         <h1>ReturnIfAbrupt</h1>
-        <p>Algorithms steps that say or are otherwise equivalent to:</p>
+        <p>An algorithm step that says:</p>
         <emu-alg>
-          1. ReturnIfAbrupt(_argument_).
+          1. *[before]* ReturnIfAbrupt(_argument_) *[after]*
         </emu-alg>
-        <p>mean the same thing as:</p>
+        <p>is equivalent to the sequence of steps:</p>
         <emu-alg>
           1. If _argument_ is an abrupt completion, return _argument_.
-          1. Else if _argument_ is a Completion Record, set _argument_ to _argument_.[[Value]].
+          1. Else, if _argument_ is a Completion Record, set _argument_ to _argument_.[[Value]].
+          1. *[before]* _argument_ *[after]*
         </emu-alg>
-        <p>Algorithms steps that say or are otherwise equivalent to:</p>
+        <p>where</p>
+        <ul>
+          <li>*[before]*, if present, is any algorithm text that does not contain ReturnIfAbrupt.
+          <li>*[after]*, if present, is any algorithm text.
+          <li>_argument_ is any variable.
+        </ul>
+
+        <p>An algorithm step that says:</p>
         <emu-alg>
-          1. ReturnIfAbrupt(AbstractOperation()).
+          1. *[before]* ReturnIfAbrupt(AbstractOperation(*[arguments]*)) *[after]*
         </emu-alg>
-        <p>mean the same thing as:</p>
+        <p>is equivalent to the sequence of steps:</p>
         <emu-alg>
-          1. Let _hygienicTemp_ be AbstractOperation().
+          1. Let _hygienicTemp_ be AbstractOperation(*[arguments]*).
+          1. If _hygienicTemp_ is an abrupt completion, return _hygienicTemp_.
+          1. Else, if _hygienicTemp_ is a Completion Record, set _hygienicTemp_ to _hygienicTemp_.[[Value]].
+          1. *[before]* _hygienicTemp_ *[after]*
+        </emu-alg>
+        <p>where</p>
+        <ul>
+          <li>*[before]*, if present, is any algorithm text that does not contain ReturnIfAbrupt.
+          <li>*[after]*, if present, is any algorithm text.
+          <li>*[arguments]*, if present, is any algorithm text that provides an argument list to AbstractOperation.
+          <li>AbstractOperation is any abstract operation reference, regardless of the notation used to refer to it.
+          <li>_hygienicTemp_ is ephemeral and visible only in the steps pertaining to ReturnIfAbrupt.
+        </ul>
+
+        <p>For example, an algorithm step that says:</p>
+        <emu-alg>
+          1. Let _result_ be _someValue_ if ReturnIfAbrupt(_someValue_.OperationName(_firstArgument_, _secondArgument_)) is *true*, and _someOtherValue_ otherwise.
+        </emu-alg>
+        <p>is equivalent to the sequence of steps:</p>
+        <emu-alg>
+          1. Let _hygienicTemp_ be _someValue_.OperationName(_firstArgument_, _secondArgument_).
           1. If _hygienicTemp_ is an abrupt completion, return _hygienicTemp_.
           1. Else if _hygienicTemp_ is a Completion Record, set _hygienicTemp_ to _hygienicTemp_.[[Value]].
+          1. Let _result_ be _someValue_ if _hygienicTemp_ is *true*, and _someOtherValue_ otherwise.
         </emu-alg>
-        <p>Where _hygienicTemp_ is ephemeral and visible only in the steps pertaining to ReturnIfAbrupt.</p>
-        <p>Algorithms steps that say or are otherwise equivalent to:</p>
-        <emu-alg>
-          1. Let _result_ be AbstractOperation(ReturnIfAbrupt(_argument_)).
-        </emu-alg>
-        <p>mean the same thing as:</p>
-        <emu-alg>
-          1. If _argument_ is an abrupt completion, return _argument_.
-          1. If _argument_ is a Completion Record, set _argument_ to _argument_.[[Value]].
-          1. Let _result_ be AbstractOperation(_argument_).
-        </emu-alg>
+
+        <p>Note that the resulting steps of ReturnIfAbrupt expansion may themselves contain ReturnIfAbrupt. In these cases, expansion can be applied to the resulting steps until the steps no longer contain ReturnIfAbrupt.</p>
       </emu-clause>
+
       <emu-clause id="sec-returnifabrupt-shorthands">
         <h1>ReturnIfAbrupt Shorthands</h1>
-        <p>Invocations of abstract operations and syntax-directed operations that are prefixed by `?` indicate that ReturnIfAbrupt should be applied to the resulting Completion Record. For example, the step:</p>
+        <p>Invocations of abstract operations may be prefixed by `?` as shorthand for applying the ReturnIfAbrupt expansion.</p>
         <emu-alg>
           1. ? OperationName().
         </emu-alg>
-        <p>is equivalent to the following step:</p>
+        <p>is equivalent to</p>
         <emu-alg>
           1. ReturnIfAbrupt(OperationName()).
         </emu-alg>
-        <p>Similarly, for method application style, the step:</p>
+        <p>wherever it appears. Similarly, prefix `?` may be used to apply the ReturnIfAbrupt expansion to the result of applying a syntax-directed operation.</p>
         <emu-alg>
-          1. ? _someValue_.OperationName().
+          1. Let _result_ be ? SyntaxDirectedOperation of |NonTerminal| with arguments _firstArgument_ and _secondArgument_.
         </emu-alg>
-        <p>is equivalent to:</p>
+        <p>is equivalent to the sequence of steps:</p>
         <emu-alg>
-          1. ReturnIfAbrupt(_someValue_.OperationName()).
+          1. Let _result_ be SyntaxDirectedOperation of |NonTerminal| with arguments _firstArgument_ and _secondArgument_.
+          1. ReturnIfAbrupt(_result_).
         </emu-alg>
-        <p>Similarly, prefix `!` is used to indicate that the following invocation of an abstract or syntax-directed operation will never return an abrupt completion and that the resulting Completion Record's [[Value]] field should be used in place of the return value of the operation. For example, the step:</p>
+
+        <p>Invocations of abstract operations may be prefixed by `!` to indicate that the result will never be an abrupt completion and that the resulting Completion Record's [[Value]] field should be used in place of the return value of the operation.</p>
         <emu-alg>
           1. Let _val_ be ! OperationName().
         </emu-alg>
-        <p>is equivalent to the following steps:</p>
+        <p>is equivalent to the sequence of steps:</p>
         <emu-alg>
           1. Let _val_ be OperationName().
           1. Assert: _val_ is never an abrupt completion.
           1. If _val_ is a Completion Record, set _val_ to _val_.[[Value]].
         </emu-alg>
-        <p>Syntax-directed operations for runtime semantics make use of this shorthand by placing `!` or `?` before the invocation of the operation:</p>
+        <p>Similarly, syntax-directed operations may make use of this shorthand by placing `!` before the invocation of the operation:</p>
         <emu-alg>
-          1. Perform ! SyntaxDirectedOperation of |NonTerminal|.
+          1. Let _result_ be the result of performing ! SyntaxDirectedOperation for |NonTerminal|.
+        </emu-alg>
+        <p>is equivalent to the sequence of steps:</p>
+        <emu-alg>
+          1. Let _result_ be SyntaxDirectedOperation of |NonTerminal|.
+          1. Assert: _result_ is never an abrupt completion.
+          1. If _result_ is a Completion Record, set _result_ to _result_.[[Value]].
         </emu-alg>
       </emu-clause>
     </emu-clause>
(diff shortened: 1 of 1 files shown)


michaelficarra:
@ljharb I've moved the `!` shorthand up to the "Implicit Completion Values" section  for now while we decide its fate in #1572.

spec.html
```diff
@@ -787,68 +787,97 @@ <h1>Throw an Exception</h1>
...
         <emu-alg>
           1. ? OperationName().
         </emu-alg>
-        <p>is equivalent to the following step:</p>
+        <p>is equivalent to</p>
         <emu-alg>
           1. ReturnIfAbrupt(OperationName()).
         </emu-alg>
-        <p>Similarly, for method application style, the step:</p>
+        <p>wherever it appears. Similarly, prefix `?` may be used to apply the ReturnIfAbrupt expansion to the result of applying a syntax-directed operation.</p>
```
jmdyck: Append "For example,"

spec.html
```diff
@@ -787,68 +787,97 @@ <h1>Throw an Exception</h1>
...
         <emu-alg>
-          1. ? _someValue_.OperationName().
+          1. Let _result_ be ? SyntaxDirectedOperation of |NonTerminal| with arguments _firstArgument_ and _secondArgument_.
         </emu-alg>
-        <p>is equivalent to:</p>
+        <p>is equivalent to the sequence of steps:</p>
         <emu-alg>
-          1. ReturnIfAbrupt(_someValue_.OperationName()).
+          1. Let _result_ be SyntaxDirectedOperation of |NonTerminal| with arguments _firstArgument_ and _secondArgument_.
+          1. ReturnIfAbrupt(_result_).
```
jmdyck: This doesn't follow the given expansion. That would be
```
1. Let _result_ be ReturnIfAbrupt(SyntaxDirectedOperation of |NonTerminal| with arguments _firstArgument_ and _secondArgument_).
```

spec.html
```diff
@@ -787,68 +787,97 @@ <h1>Throw an Exception</h1>
...
-          1. Else if _argument_ is a Completion Record, set _argument_ to _argument_.[[Value]].
+          1. Else, if _argument_ is a Completion Record, set _argument_ to _argument_.[[Value]].
+          1. *[before]* _argument_ *[after]*
         </emu-alg>
-        <p>Algorithms steps that say or are otherwise equivalent to:</p>
+        <p>where</p>
+        <ul>
+          <li>*[before]*, if present, is any algorithm text that does not contain ReturnIfAbrupt.
+          <li>*[after]*, if present, is any algorithm text.
+          <li>_argument_ is any variable.
```
jmdyck: The spec calls them "aliases".

spec.html
```diff
@@ -787,68 +787,97 @@ <h1>Throw an Exception</h1>
...
+        </ul>
+
+        <p>An algorithm step that says:</p>
         <emu-alg>
-          1. ReturnIfAbrupt(AbstractOperation()).
+          1. *[before]* ReturnIfAbrupt(AbstractOperation(*[arguments]*)) *[after]*
```
jmdyck: Rather than having 2 definitions, discriminated by the form of the argument, you could have one definition with a generalized argument.

spec.html
```diff
@@ -771,6 +771,27 @@ <h1>Implicit Completion Values</h1>
           1. Return NormalCompletion(*undefined*).
         </emu-alg>
         <p>Any reference to a Completion Record value that is in a context that does not explicitly require a complete Completion Record value is equivalent to an explicit reference to the [[Value]] field of the Completion Record value unless the Completion Record is an abrupt completion.</p>
+
+        <p>Invocations of abstract operations may be prefixed by `!` to indicate that the result will never be an abrupt completion and that the resulting Completion Record's [[Value]] field should be used in place of the return value of the operation.</p>
+        <emu-alg>
+          1. Let _val_ be ! OperationName().
```
jmdyck: As in the status quo, this is an illustrative example, not a real definition. You should maybe give it a proper definition as with 'ReturnIfAbupt'.

spec.html
```diff
@@ -771,6 +771,27 @@ <h1>Implicit Completion Values</h1>
           1. Return NormalCompletion(*undefined*).
         </emu-alg>
         <p>Any reference to a Completion Record value that is in a context that does not explicitly require a complete Completion Record value is equivalent to an explicit reference to the [[Value]] field of the Completion Record value unless the Completion Record is an abrupt completion.</p>
+
+        <p>Invocations of abstract operations may be prefixed by `!` to indicate that the result will never be an abrupt completion and that the resulting Completion Record's [[Value]] field should be used in place of the return value of the operation.</p>
```
jmdyck: I don't think the "!" stuff fits very well in the "Implicit Completion Values" clause. It makes more sense where it is now, in the same clause as "?". Instead of "ReturnIfAbrupt Shorthands", that clause could be called "Shorthands relating to completion records" or just "? and ! Shorthands".

Or split them into 2 clauses: "The ? Shorthand" and "The ! Shorthand".

spec.html
```diff
@@ -787,68 +787,97 @@ <h1>Throw an Exception</h1>
...
+          <li>*[after]*, if present, is any algorithm text.
+          <li>*[arguments]*, if present, is any algorithm text that provides an argument list to AbstractOperation.
+          <li>AbstractOperation is any abstract operation reference, regardless of the notation used to refer to it.
+          <li>_hygienicTemp_ is ephemeral and visible only in the steps pertaining to ReturnIfAbrupt.
+        </ul>
+
+        <p>For example, an algorithm step that says:</p>
         <emu-alg>
-          1. Let _result_ be AbstractOperation(ReturnIfAbrupt(_argument_)).
+          1. Let _result_ be _someValue_ if ReturnIfAbrupt(_someValue_.OperationName(_firstArgument_, _secondArgument_)) is *true*, and _someOtherValue_ otherwise.
```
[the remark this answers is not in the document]
michaelficarra: It's just prose, we can say whatever we like if the meaning is clear. Is there an incorrect interpretation of this prose? Do you have a recommendation for a replacement?

spec.html
michaelficarra: Since the `!` shorthand has no effect on the evaluation, I think it's a little less important to be as rigorous as we are with `ReturnIfAbrupt` expansion. If anything, the `?` shorthand would deserve that level of rigour. I would do it for `?` if the editors think that would help.

spec.html
michaelficarra: Since it's a structural rewrite, I'd like to keep it as simple as I can. Though one could argue that 1 rewrite rule is simpler overall than 2 individually simpler rewrite rules. I'll leave that one up to @tc39/ecma262-editors.

spec.html
michaelficarra: GitHub seems to have lost the position of this, and I can't figure out what you're referring to. Is there an issue in one of my example expansions somewhere?

michaelficarra:
@jmdyck I've addressed your comments and left some responses. Please do another review.

spec.html
ljharb: 1 seems probably simpler, but totes fine to do that in a followup.

spec.html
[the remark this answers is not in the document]
```suggestion
          1. If ReturnIfAbrupt(_someValue_.OperationName(_firstArgument_, _secondArgument_)) is *true*, let _result_ be _someValue_ , else let _result_ be _someOtherValue_.
```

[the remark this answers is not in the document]
jmdyck: Re @ljharb suggestion: comma before "else" should probably be a semicolon.

spec.html
```diff
@@ -787,68 +787,98 @@ <h1>Throw an Exception</h1>
...
-        <p>mean the same thing as:</p>
+        <p>is equivalent to the sequence of steps:</p>
         <emu-alg>
-          1. If _argument_ is an abrupt completion, return _argument_.
-          1. If _argument_ is a Completion Record, set _argument_ to _argument_.[[Value]].
-          1. Let _result_ be AbstractOperation(_argument_).
+          1. Let _hygienicTemp_ be _someValue_.OperationName(_firstArgument_, _secondArgument_).
+          1. If _hygienicTemp_ is an abrupt completion, return _hygienicTemp_.
+          1. Else if _hygienicTemp_ is a Completion Record, set _hygienicTemp_ to _hygienicTemp_.[[Value]].
+          1. Let _result_ be _someValue_ if _hygienicTemp_ is *true*, and _someOtherValue_ otherwise.
```
```suggestion
          1. If _hygienicTemp_ is *true*, let _result_ be _someValue_; else let _result_ be _someOtherValue_.
```
(Similar to @ljharb's suggestion at L829.)

spec.html
```diff
@@ -787,68 +787,98 @@ <h1>Throw an Exception</h1>
...
+          1. Let _hygienicTemp_ be AbstractOperation(*[arguments]*).
           1. If _hygienicTemp_ is an abrupt completion, return _hygienicTemp_.
           1. Else if _hygienicTemp_ is a Completion Record, set _hygienicTemp_ to _hygienicTemp_.[[Value]].
+          1. *[before]* _hygienicTemp_ *[after]*
         </emu-alg>
-        <p>Where _hygienicTemp_ is ephemeral and visible only in the steps pertaining to ReturnIfAbrupt.</p>
-        <p>Algorithms steps that say or are otherwise equivalent to:</p>
+        <p>where</p>
+        <ul>
+          <li>*[before]*, if present, is any algorithm text that does not contain ReturnIfAbrupt.
```
```suggestion
          <li>*[before]*, if present, is any algorithm text that does not contain ReturnIfAbrupt.</li>
```
and similarly for the other 7 `<li>` in this PR. So far, the spec always uses closing tags when they exist. (Not sure why I didn't spot this in my earlier review.)

spec.html
jmdyck: github's position (line 859) is correct, but the example has changed. Updated, my point it that:
```
1. Let _result_ be ? SyntaxDirectedOperation of |NonTerminal|.
```
should (according to the definition) expand to:
```
1. Let _result_ be ReturnIfAbrupt(SyntaxDirectedOperation of |NonTerminal|).
```
as opposed to:
```
1. Let _result_ be SyntaxDirectedOperation of |NonTerminal|.
1. ReturnIfAbrupt(_result_).
```

spec.html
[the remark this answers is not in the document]
michaelficarra: I've just simplified the example to avoid this.

spec.html
michaelficarra: I've reduced the ReturnIfAbrupt definition to a single macro expansion, though it now has the peculiarity that, when used on an alias, as in `ReturnIfAbrupt(_someAlias_)`, we are left with a step at the end that just references the alias. I think that's probably fine, but it may not be clear to everyone that that is meant to be a no-op. Let me know if you want me to switch it back to 2 expansions, I'm not tied to a particular choice.

michaelficarra:
Okay @jmdyck @ljharb ready for another round of reviews.

spec.html
```diff
@@ -787,68 +787,79 @@ <h1>Throw an Exception</h1>
...
       </emu-clause>
+
       <emu-clause id="sec-returnifabrupt-shorthands">
-        <h1>ReturnIfAbrupt Shorthands</h1>
-        <p>Invocations of abstract operations and syntax-directed operations that are prefixed by `?` indicate that ReturnIfAbrupt should be applied to the resulting Completion Record. For example, the step:</p>
+        <h1>Shorthands Relating to Completion Records</h1>
+
+        <p>Invocations of abstract operations may be prefixed by `?` as shorthand for applying the ReturnIfAbrupt expansion.</p>
         <emu-alg>
           1. ? OperationName().
```
ljharb: not sure if we need to call out `Perform` (ie, `Perform ? Op()` or `Perform ! Op()`), or if that's a topic for a different PR.

michaelficarra:
A potential issue with this change as-written: multiple `ReturnIfAbrupt` are permitted within a single algorithm step, but are always evaluated right-to-left. This works fine for instances like

> ... ?A( ?B( ?C() ) ) ...

which I had in mind when writing it, but causes undesirable ordering for instances like

> ... ?A( ?B(), ?C() ) ...

These both would evaluate C, then B, then A. I'm not sure how we'd go about writing a macro expansion that correctly orders the latter case. Is it necessary to support those?

devsnek:
> ... ?A( ?B(), ?C() ) ...

At that point you should probably be using multiple lines just for the sake of clarity.

michaelficarra:
Okay, then to be clear @devsnek, you think it would be okay forbidding that usage? If so, I should add a non-normative note to spec/proposal authors that describes this potentially erroneous usage.

jmdyck:
> A potential issue with this change as-written: multiple `ReturnIfAbrupt` are permitted within a single algorithm step, but are always evaluated right-to-left. This works fine for instances like
> 
> > ... ?A( ?B( ?C() ) ) ...
> 
> which I had in mind when writing it, but causes undesirable ordering for instances like
> 
> > ... ?A( ?B(), ?C() ) ...
> 
> These both would evaluate C, then B, then A. I'm not sure how we'd go about writing a macro expansion that correctly orders the latter case.

Tangent:

I'm guessing you mean that the correct ordering for the latter case is B then C then A, but I don't think the spec actually says that. [5.2.1 Abstract Operations](https://tc39.es/ecma262/#sec-algorithm-conventions-abstract-operations) says:
> Abstract operations are typically referenced using a functional application style such as OperationName(arg1, arg2).

but it *doesn't* then go on to say anything about the order in which the args are 'evaluated'.

I'd be inclined to say that any case where the order is observable is a spec bug. (Detecting such bugs is another matter though.)

devsnek:
You can detect the order by making them both throw

jmdyck:
> You can detect the order by making them both throw

I was talking about detecting (just by examining the spec) places where the order of 'evaluation' of operation arguments could be observed. I think you're saying that, once such a place has been identified, you can construct a test case that will reveal which order an implementation is using.

michaelficarra:
This is blocked until completion record reform work has been completed.

spec.html
bakkot: I like this as a single expansion. I think the "step at the end that just references the alias" thing is unlikely to trip up readers, but if we're worried, we could say "the last line is omitted if `*[before]*` is not present" (or "is empty" or whatever).

bakkot:
> These both would evaluate C, then B, then A. I'm not sure how we'd go about writing a macro expansion that correctly orders the latter case. Is it necessary to support those?

My preference would be to define an evaluation more generally - not rigorously, just say something like "Evaluation of algorithm steps follows a inside-to-outside, left-to-right evaluation order, so that the step `Perform A(B(), C()).` evaluates B, then C, then A" or something like that.

And then in this macro you can say that if there are multiple occurrences of `ReturnIfAbrupt` within one step, they are expanded in evaluation order, with the first to be evaluated being expanded first.

jugglinmike:
@michaelficarra was gh-2547 the completion reform which [blocked this work](https://github.com/tc39/ecma262/pull/1573#issuecomment-811497331)?

michaelficarra:
@jugglinmike Yes, this should be unblocked now.

jugglinmike:
Thanks, Michael! I can't commit to moving this along, but it helps to understand the status.

michaelficarra:
@jugglinmike That's okay, I plan on revisiting it soon(-ish). Let me know if you want to take it over, though. I don't remember what work it needs.

bakkot:
What if we just do away with `ReturnIfAbrupt` entirely, and instead just define `?` directly? That's probably simpler, since it doesn't need to rebind any aliases. And after https://github.com/tc39/ecma262/pull/3268 there's only 4 uses of `ReturnIfAbrupt` left, so getting rid of it is straightforward.

bakkot:
some text I started writing at some point, needs some polish but I think is good:

----

## Evaluation order

When complex expressions appear in algorithm steps, these are to be understood as being evaluated in a left-to-right, inside-to-outside order. For example, the step

1. Return A(B(), C.[[D]]) + E(F()).

is equivalent to

1. Let _tmp1_ be B().
2. Let _tmp2_ be C.[[D]].
3. Let _tmp3_ be A(_tmp1_, _tmp2_).
4. Let _tmp4_ be F().
5. Let _tmp5_ be E(_tmp4_).
6. Let _tmp6_ be _tmp3_ + _tmp5_.
7. Return _tmp6_.

----

## Completion-unwrapping macros

`?` and `!` are used as macros which unwrap completion records. The step

1. Let _foo_ be ? _bar_.

is equivalent to

1. If _bar_ is an abrupt completion, return _bar_.
2. Let _foo_ be _bar_.[[Value]].

The step

1. Let _foo_ be ! _bar_.

is equivalent to

1. Assert: _bar_ is a normal completion.
2. Let _foo_ be _bar_.[[Value]].

When `?` or `!` is used in any other context, first apply the rewrite given in [previous section], then apply this rule.

michaelficarra:
Closing in favour of https://github.com/tc39/ecma262/pull/3705.

PR #1576, opened 2019-06-07
Meta: Use shiny new hostname

mathiasbynens:


spec.html
```diff
@@ -43,15 +43,15 @@
   }
 </style>
 <script>
```
annevk: This entire script block is obsolete I think. I suggest removing it as per #1575 and then I can close that.

mathiasbynens: Done. (Also, sorry, I should have seen your PR earlier.)

michaelficarra:
So... what's with the Spanish TLD?

ljharb:
I’m a bit confused by the new domain name as well; was it discussed during the meeting?

annevk:
Yes and it had consensus.

ljharb:
It’s not on the agenda; maybe i missed it in the draft notes. I’ll take another look.

michaelficarra:
@ljharb It's in the June 4th notes. Not much is mentioned on motivation though. Only comments are from @gesa and @ecmageneva. @ecmageneva requested a Swiss registrar and, Ecma being a swiss organisation, I would think that tc39.ch would be a more appropriate ccTLD. Should we move this discussion to the reflector?

ljharb:
Thanks. Yeah a reflector issue would be best.

mathiasbynens:
Meanwhile, the domain is live and the old domain redirects to it. Let's merge this PR to reflect reality (and save anyone who clicks a redirect)?

mathiasbynens:
@ljharb or @zenparsing, could you also update the URL here?

ljharb:
Done.

azu:
FYI: https://www.ecma-international.org/publications/standards/Ecma-262.htm should also be updated when the new domain has been decided.

michaelficarra:
@ljharb posted to discourse: https://es.discourse.group/t/draft-spec-text-domain/75

PR #1580, opened 2019-06-10
Editorial: Move ToObject out of GetSubstitution

shvaikalesh:
Object coercion is only relevant to `GetSubstitution` call in [`RegExp.prototype[@@replace]`](https://tc39.es/ecma262/#sec-regexp.prototype-@@replace) (step **14.l.i**), so let's move it there. There are no observable steps between current and proposed locations of `ToObject` call.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -29796,7 +29796,7 @@ THH:mm:ss.sss
             1. Let _replStr_ be ? ToString(_replValue_).
           1. Else,
             1. Let _captures_ be a new empty List.
-            1. Let _replStr_ be GetSubstitution(_matched_, _string_, _pos_, _captures_, *undefined*, _replaceValue_).
+            1. Let _replStr_ be ! GetSubstitution(_matched_, _string_, _pos_, _captures_, *undefined*, _replaceValue_).
           1. Let _tailPos_ be _pos_ + the number of code units in _matched_.
           1. Let _newString_ be the string-concatenation of the first _pos_ code units of _string_, _replStr_, and the trailing substring of _string_ starting at index _tailPos_. If _pos_ is 0, the first element of the concatenation will be the empty String.
           1. Return _newString_.
@@ -29819,8 +29819,6 @@ THH:mm:ss.sss
             1. Assert: Type(_replacement_) is String.
             1. Let _tailPos_ be _position_ + _matchLength_.
             1. Let _m_ be the number of elements in _captures_.
-            1. If _namedCaptures_ is not *undefined*, then
-              1. Set _namedCaptures_ to ? ToObject(_namedCaptures_).
             1. Let _result_ be the String value derived from _replacement_ by copying code unit elements from _replacement_ to _result_ while performing replacements as specified in <emu-xref href="#table-45"></emu-xref>. These `$` replacements are done left-to-right, and, once such a replacement is performed, the new replacement text is not subject to further replacements.
             1. Return _result_.
           </emu-alg>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -28657,8 +28657,6 @@ <h1>Runtime Semantics: GetSubstitution ( _matched_, _str_, _position_, _captures
             1. Assert: Type(_replacement_) is String.
             1. Let _tailPos_ be _position_ + _matchLength_.
             1. Let _m_ be the number of elements in _captures_.
-            1. If _namedCaptures_ is not *undefined*, then
-              1. Set _namedCaptures_ to ? ToObject(_namedCaptures_).
```
ljharb: This would be an observable/normative change, since the order the exception is thrown matters.

ljharb: hmm - if `functionalReplace` is false, and `namedCaptures` is null, wouldn't this change be observable?

shvaikalesh: These are the steps between current and proposed location of `ToObject`:

```
1. Assert: Type(_matched_) is String.
1. Let _matchLength_ be the number of code units in _matched_.
1. Assert: Type(_str_) is String.
1. Let _stringLength_ be the number of code units in _str_.
1. Assert: _position_ is a nonnegative integer.
1. Assert: _position_ &le; _stringLength_.
1. Assert: _captures_ is a possibly empty List of Strings.
1. Assert: Type(_replacement_) is String.
1. Let _tailPos_ be _position_ + _matchLength_.
1. Let _m_ be the number of elements in _captures_.
```

None of them seems observable, but I'd rather have throwing exception before all the asserts.

spec.html
```diff
@@ -28634,7 +28634,7 @@ <h1>String.prototype.replace ( _searchValue_, _replaceValue_ )</h1>
             1. Let _replStr_ be ? ToString(_replValue_).
           1. Else,
             1. Let _captures_ be a new empty List.
-            1. Let _replStr_ be GetSubstitution(_matched_, _string_, _pos_, _captures_, *undefined*, _replaceValue_).
+            1. Let _replStr_ be ? GetSubstitution(_matched_, _string_, _pos_, _captures_, *undefined*, _replaceValue_).
```
bakkot: This `?` was necessary before this change, but after this change it looks like `GetSubstitution` can no longer return an abrupt completion. If that's right, this should be `!` (or omitted).

shvaikalesh: That was my first thought too, but please see steps for `$<` in [Replacement Text Symbol Substitutions table](https://tc39.es/ecma262/#table-45): getting named group value may turn abrupt completion.

//cc @ljharb

bakkot: Well, that's not good.

I think that table should be done away with. The right-hand column mixes code points, aliases, declarative prose descriptions of text, and imperative algorithms steps which can return abrupt completions. Probably it is time to just make a big if-else.

ljharb: whoops. I'll convert these back to `?` in a followup commit, and then we can do the larger change via PR.

ljharb: Fixed in 37728d736e9e6a1e2ba951dbeb3fc096914193d6

bakkot: 37728d7 is over-zealous: when `namedCaputres` is `*undefined*`, the operation genuinely cannot throw. So the first invocation of `GetSubstitution` should remain `!`.

shvaikalesh: Perhaps #1732 resolves some of those concerns?

ljharb: 8fd2e014fa52a1c932686904a7b46eb742fe3112

bakkot: #1732 is definitely a step in the right direction, and any further changes to this section would have conflicts with that PR, so that's probably the right place for further discussion.

PR #1583, opened 2019-06-12
Normative: Add Promise.allSettled()

jasonwilliams:
This is a pull request for the **Stage 3** proposal https://github.com/tc39/proposal-promise-allSettled. The proposal is championed by @mathiasbynens.

**test262 tests:** yes https://github.com/tc39/test262/pull/2131 and https://github.com/tc39/test262/pull/2124

## Rationale
A common use case for this combinator is wanting to take an action after multiple requests have completed, regardless of their success or failure. Other Promise combinators can short-circuit, discarding the results of input values that lose the race to reach a certain state. Promise.allSettled is unique in always waiting for all of its input values.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -39611,6 +39611,125 @@ THH:mm:ss.sss
         </emu-clause>
       </emu-clause>
 
+      <emu-clause id="sec-promise.allsettled">
+        <h1>Promise.allSettled ( _iterable_ )</h1>
+        <p>The `allSettled` function returns a promise that is fulfilled with an array of promise state snapshots, but only after all the original promises have settled, i.e. become either fulfilled or rejected. It resolves all elements of the passed iterable to promises as it runs this algorithm.</p>
+        <emu-alg>
+          1. Let _C_ be the *this* value.
+          1. Let _promiseCapability_ be ? NewPromiseCapability(_C_).
+          1. Let _iteratorRecord_ be GetIterator(_iterable_).
+          1. IfAbruptRejectPromise(_iteratorRecord_, _promiseCapability_).
+          1. Let _result_ be PerformPromiseAllSettled(_iteratorRecord_, _C_, _promiseCapability_).
+          1. If _result_ is an abrupt completion, then
+            1. If _iteratorRecord_.[[Done]] is *false*, set _result_ to IteratorClose(_iteratorRecord_, _result_).
+            1. IfAbruptRejectPromise(_result_, _promiseCapability_).
+          1. Return Completion(_result_).
+        </emu-alg>
+        <emu-note>
+          <p>The `allSettled` function requires its *this* value to be a constructor function that supports the parameter conventions of the `Promise` constructor.</p>
+        </emu-note>
+
+        <emu-clause id="sec-performpromiseallsettled" aoid="PerformPromiseAllSettled">
+          <h1>Runtime Semantics: PerformPromiseAllSettled ( _iteratorRecord_, _constructor_, _resultCapability_ )</h1>
+          <p>When the PerformPromiseAllSettled abstract operation is called with arguments _iteratorRecord_, _constructor_, and _resultCapability_, the following steps are taken:</p>
+          <emu-alg>
+            1. Assert: ! IsConstructor(_constructor_) is *true*.
+            1. Assert: _resultCapability_ is a PromiseCapability Record.
+            1. Let _values_ be a new empty List.
+            1. Let _remainingElementsCount_ be a new Record { [[Value]]: 1 }.
+            1. Let _index_ be 0.
+            1. Let _promiseResolve_ be ? Get(_constructor_, `"resolve"`).
+            1. If IsCallable(_promiseResolve_) is *false*, throw a *TypeError* exception.
+            1. Repeat,
+              1. Let _next_ be IteratorStep(_iteratorRecord_).
+              1. If _next_ is an abrupt completion, set _iteratorRecord_.[[Done]] to *true*.
+              1. ReturnIfAbrupt(_next_).
+              1. If _next_ is *false*, then
+                1. Set _iteratorRecord_.[[Done]] to *true*.
+                1. Set _remainingElementsCount_.[[Value]] to _remainingElementsCount_.[[Value]] - 1.
+                1. If _remainingElementsCount_.[[Value]] is 0, then
+                  1. Let _valuesArray_ be ! CreateArrayFromList(_values_).
+                  1. Perform ? Call(_resultCapability_.[[Resolve]], *undefined*, &laquo; _valuesArray_ &raquo;).
+                1. Return _resultCapability_.[[Promise]].
+              1. Let _nextValue_ be IteratorValue(_next_).
+              1. If _nextValue_ is an abrupt completion, set _iteratorRecord_.[[Done]] to *true*.
+              1. ReturnIfAbrupt(_nextValue_).
+              1. Append *undefined* to _values_.
+              1. Let _nextPromise_ be ? Call(_promiseResolve_, _constructor_, &laquo; _nextValue_ &raquo;).
+              1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-promise.allsettled-resolve-element-functions" title></emu-xref>.
+              1. Let _resolveElement_ be ! CreateBuiltinFunction(_steps_, &laquo; [[AlreadyCalled]], [[Index]], [[Values]], [[Capability]], [[RemainingElements]] &raquo;).
+              1. Let _alreadyCalled_ be a new Record { [[Value]]: *false* }.
+              1. Set _resolveElement_.[[AlreadyCalled]] to _alreadyCalled_.
+              1. Set _resolveElement_.[[Index]] to _index_.
+              1. Set _resolveElement_.[[Values]] to _values_.
+              1. Set _resolveElement_.[[Capability]] to _resultCapability_.
+              1. Set _resolveElement_.[[RemainingElements]] to _remainingElementsCount_.
+              1. Let _rejectSteps_ be the algorithm steps defined in <emu-xref href="#sec-promise.allsettled-reject-element-functions" title></emu-xref>.
+              1. Let _rejectElement_ be ! CreateBuiltinFunction(_rejectSteps_, &laquo; [[AlreadyCalled]], [[Index]], [[Values]], [[Capability]], [[RemainingElements]] &raquo;).
+              1. Set _rejectElement_.[[AlreadyCalled]] to _alreadyCalled_.
+              1. Set _rejectElement_.[[Index]] to _index_.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -38326,6 +38326,126 @@ <h1>`Promise.all` Resolve Element Functions</h1>
...
+          1. Let _C_ be the *this* value.
+          1. If Type(_C_) is not Object, throw a *TypeError* exception.
+          1. Let _promiseCapability_ be ? NewPromiseCapability(_C_).
+          1. Let _iteratorRecord_ be GetIterator(_iterable_).
+          1. IfAbruptRejectPromise(_iteratorRecord_, _promiseCapability_).
+          1. Let _result_ be PerformPromiseAllSettled(_iteratorRecord_, _C_, _promiseCapability_).
+          1. If _result_ is an abrupt completion, then
+            1. If _iteratorRecord_.[[Done]] is *false*, set _result_ to IteratorClose(_iteratorRecord_, _result_).
+            1. IfAbruptRejectPromise(_result_, _promiseCapability_).
+          1. Return Completion(_result_).
```
ljharb: why is `Completion()` needed here? we already know it's not an abrupt completion, so isn't it already a normal one?
```suggestion
          1. Return _result_.
```

spec.html
```diff
@@ -38326,6 +38326,126 @@ <h1>`Promise.all` Resolve Element Functions</h1>
...
+          <emu-alg>
+            1. Let _F_ be the active function object.
+            1. Let _alreadyCalled_ be _F_.[[AlreadyCalled]].
+            1. If _alreadyCalled_.[[Value]] is *true*, return *undefined*.
+            1. Set _alreadyCalled_.[[Value]] to *true*.
+            1. Let _index_ be _F_.[[Index]].
+            1. Let _values_ be _F_.[[Values]].
+            1. Let _promiseCapability_ be _F_.[[Capability]].
+            1. Let _remainingElementsCount_ be _F_.[[RemainingElements]].
+            1. Let _obj_ be ! ObjectCreate(%ObjectPrototype%).
```
ljharb: would it be worth (for readability) making an abstract operation here (and below) like `CreateSettledResult(true/false, x)`?

spec.html
mathiasbynens: Per <https://tc39.es/ecma262/#sec-implicit-completion-values>:

> The abstract operation `Completion(completionRecord)` is used to emphasize that a previously computed Completion Record is being returned. The `Completion` abstract operation takes a single argument, `completionRecord`, and performs the following steps: [...]

spec.html
mathiasbynens: That sounds like a potentially beneficial refactoring! Would you like this to happen as part of this PR, or separately as a follow-up?

spec.html
ljharb: As an aside, I’d love to come up with a ?-like pattern around all these IfAbruptRejectPromise combos someday

spec.html
ljharb: I guess it could be either, but it’d be nice to do it as part of this PR :-)

spec.html
```diff
@@ -38326,6 +38326,126 @@ <h1>`Promise.all` Resolve Element Functions</h1>
...
+        </emu-note>
+    
+        <emu-clause id="sec-performpromiseallsettled" aoid="PerformPromiseAllSettled">
+          <h1>Runtime Semantics: PerformPromiseAllSettled ( _iteratorRecord_, _constructor_, _resultCapability_ )</h1>
+          <p>When the PerformPromiseAllSettled abstract operation is called with arguments _iteratorRecord_, _constructor_, and _resultCapability_, the following steps are taken:</p>
+          <emu-alg>
+            1. Assert: ! IsConstructor(_constructor_) is *true*.
+            1. Assert: _resultCapability_ is a PromiseCapability Record.
+            1. Let _values_ be a new empty List.
+            1. Let _remainingElementsCount_ be a new Record { [[Value]]: 1 }.
```
shvaikalesh: Perhaps it would be beneficial to add a note explaining why `remainingElementsCount` is a Record rather than just a Number (to pass it by reference rather than by value)?

spec.html
```diff
@@ -38326,6 +38326,126 @@ <h1>`Promise.all` Resolve Element Functions</h1>
         </emu-clause>
       </emu-clause>
 
+      <emu-clause id="sec-promise.allsettled">
+        <h1>Promise.allSettled ( _iterable_ )</h1>
+        <p>The `allSettled` function returns a promise that is fulfilled with an array of promise state snapshots, but only after all the original promises have settled, i.e. become either fulfilled or rejected. It resolves all elements of the passed iterable to promises as it runs this algorithm.</p>
+        <emu-alg>
+          1. Let _C_ be the *this* value.
+          1. If Type(_C_) is not Object, throw a *TypeError* exception.
```
mathiasbynens: This line can be removed, to match `Promise.race` and `Promise.all`. Doing so does not change observable semantics, since if `Type(C)` is not `Object`, then `NewPromiseCapability` in the next step throws the same exception anyhow.

spec.html
mathiasbynens: This just matches the precedent set by [`PerformPromiseAll`](https://tc39.es/ecma262/#sec-performpromiseall). I'm not opposed to changing this, but maybe we can do it as a separate PR?

spec.html
devsnek: it seems with that change we could also fold these two functions into one

spec.html
ljharb: I find this approach highly distasteful (altho i have no better suggestions atm), and commented such on the proposal repo, but since it does mirror the existing approach i think pursuing changing both in a separate PR is the proper approach (rather than obstructing this proposal).

spec.html
```diff
@@ -38326,6 +38326,126 @@ <h1>`Promise.all` Resolve Element Functions</h1>
...
+            1. If IsCallable(_promiseResolve_) is *false*, throw a *TypeError* exception.
+            1. Repeat,
+              1. Let _next_ be IteratorStep(_iteratorRecord_).
+              1. If _next_ is an abrupt completion, set _iteratorRecord_.[[Done]] to *true*.
+              1. ReturnIfAbrupt(_next_).
+              1. If _next_ is *false*, then
+                1. Set _iteratorRecord_.[[Done]] to *true*.
+                1. Set _remainingElementsCount_.[[Value]] to _remainingElementsCount_.[[Value]] - 1.
+                1. If _remainingElementsCount_.[[Value]] is 0, then
+                  1. Let _valuesArray_ be CreateArrayFromList(_values_).
```
chicoxyzzy:
```suggestion
                  1. Let _valuesArray_ be ! CreateArrayFromList(_values_).
```

spec.html
```diff
@@ -38326,6 +38326,126 @@ <h1>`Promise.all` Resolve Element Functions</h1>
...
+            1. Let _values_ be _F_.[[Values]].
+            1. Let _promiseCapability_ be _F_.[[Capability]].
+            1. Let _remainingElementsCount_ be _F_.[[RemainingElements]].
+            1. Let _obj_ be ! ObjectCreate(%ObjectPrototype%).
+            1. Perform ! CreateDataProperty(_obj_, `"status"`, `"rejected"`).
+            1. Perform ! CreateDataProperty(_obj_, `"reason"`, _x_).
+            1. Set _values_[_index_] to be _obj_.
+            1. Set _remainingElementsCount_.[[Value]] to _remainingElementsCount_.[[Value]] - 1.
+            1. If _remainingElementsCount_.[[Value]] is 0, then
+              1. Let _valuesArray_ be CreateArrayFromList(_values_).
```
chicoxyzzy:
```suggestion
              1. Let _valuesArray_ be ! CreateArrayFromList(_values_).
```

spec.html
```diff
@@ -38326,6 +38326,126 @@ <h1>`Promise.all` Resolve Element Functions</h1>
...
+      <emu-clause id="sec-promise.allsettled">
+        <h1>Promise.allSettled ( _iterable_ )</h1>
+        <p>The `allSettled` function returns a promise that is fulfilled with an array of promise state snapshots, but only after all the original promises have settled, i.e. become either fulfilled or rejected. It resolves all elements of the passed iterable to promises as it runs this algorithm.</p>
+        <emu-alg>
+          1. Let _C_ be the *this* value.
+          1. If Type(_C_) is not Object, throw a *TypeError* exception.
+          1. Let _promiseCapability_ be ? NewPromiseCapability(_C_).
+          1. Let _iteratorRecord_ be GetIterator(_iterable_).
+          1. IfAbruptRejectPromise(_iteratorRecord_, _promiseCapability_).
+          1. Let _result_ be PerformPromiseAllSettled(_iteratorRecord_, _C_, _promiseCapability_).
```
chicoxyzzy:
```suggestion
          1. Let _result_ be ? PerformPromiseAllSettled(_iteratorRecord_, _C_, _promiseCapability_).
```

leobalter: This review feedback is wrong. It should not return if the return is an abrupt completion. 

The current text, without the `?`, is correct.

spec.html
```diff
@@ -38427,6 +38427,125 @@ <h1>`Promise.all` Resolve Element Functions</h1>
...
+          <emu-alg>
+            1. Let _F_ be the active function object.
+            1. Let _alreadyCalled_ be _F_.[[AlreadyCalled]].
+            1. If _alreadyCalled_.[[Value]] is *true*, return *undefined*.
+            1. Set _alreadyCalled_.[[Value]] to *true*.
+            1. Let _index_ be _F_.[[Index]].
+            1. Let _values_ be _F_.[[Values]].
+            1. Let _promiseCapability_ be _F_.[[Capability]].
+            1. Let _remainingElementsCount_ be _F_.[[RemainingElements]].
+            1. Let _obj_ be ! ObjectCreate(%ObjectPrototype%).
```
chicoxyzzy:
```suggestion
            1. Let _obj_ be ! ObjectCreate(%Object.Prototype%).
```

spec.html
```diff
@@ -38427,6 +38427,125 @@ <h1>`Promise.all` Resolve Element Functions</h1>
...
```
chicoxyzzy:
```suggestion
            1. Let _obj_ be ! ObjectCreate(%Object.Prototype%).
```

mathiasbynens:
Is anything still blocking this? It seems we can apply @chicoxyzzy's suggestions and (given that both @ljharb and @zenparsing LGTM'd) get this merged?

mathiasbynens:
Note that @jasonwilliams somehow cannot apply the open suggestions:



Perhaps one of the editors can take care of this before merging?

ljharb:
@mathiasbynens that's a github bug, i believe when the suggester and the PR author both have write access - but either way it can be applied manually on the command line.

I can certainly incorporate any open suggestions prior to merging as well.

mathiasbynens:
Okay, so this seems good to go then!

Issue #1369, opened 2018-12-12
Array.prototpye.sort specification problem


balagge:
The current draft https://tc39.github.io/ecma262/#sec-array.prototype.sort contains confusing wording on the expected result and/or missing constraints on the consistency of the _comparefn_.

(1) the expected result of a consistent _comparefn_ is described thus:

> [...]
> - Then for all nonnegative integers j and k, each less than len, if `SortCompare(old[j], old[k]) < 0` (see SortCompare below), then `new[π(j)] < new[π(k)]`. 

Here the notation `new[π(j)] < new[π(k)]` makes no sense, because the meaning of `<` is unclear in this context. Furthermore, this constraint probably wants to say something about the expected position of the two elements after the sort, and not the elements themselves; so probably `π(j) < π(k)` is intended? In any case, by the previous constraint, `new[π(j)] ` is the same element as `old[j]`, so essentially the current wording means `old[j] < old[k]` (assuming the said properties exist), which is nothing but the definition of _comparefn_, if anything. So the current wording says nothing more about `π`, the result of the sort operation; but clearly, this part is intended to say something about `π`, isn't it?

(2) Missing constraint on  `SortCompare(old[j], old[k]) > 0` (in the expected result of the sort) and/or missing requirement (antisymmetry) on the consistency of _comparefn_ later.

The current wording contains nothing on a positive value returned by the _comparefn_. Therefore, the following _comparefn_ is consistent and any result is acceptable: `( (a,b) => a===b ? 0 : 1)`. This is probably not what is intended by the spec.

I think either a constraint on the result of the sort is missing (dual of the quoted part, this time for `SortCompare(old[j], old[k]) > 0` ), or the consistency requirements should include an additional requirement on the antisymmetry of the _comparefn_. Maybe both.

Update: maybe antisymmetry can be deduced from the existing requirements and then this item (2) is not a problem after all. However, it would make sense to include such a comment in the spec.

(3) The following wording in the consistency requirements is ambiguous:

> A function comparefn is a consistent comparison function for a set of values S if all of the requirements below are met for all values a, b, and c (possibly the same value) ...

Here the terms "set of values" and "same value" is not clear. What is considered to be the "same" in this context? what does "set" mean (note that this is practically important because the reflexivity requirement below implicitly uses the concept of "being the same" by using the variable `a` twice.

claudepache:
> (1) the expected result of a consistent _comparefn_ is described thus:
> 
> > [...]
> > 
[...]

PR #1585, opened 2019-06-14
Normative: A more precise Array.prototype.sort

szuend:
_tl;dr This PR intends to nail down some parts of `Array.prototype.sort` to reduce the amount of cases that result in an implementation-defined sort order._

## Overview

The `Array.prototype.sort` procedure that this PR proposes, can be summarized as follows:

1. Collect all existing, non-`undefined` values in the range of `[0, [[length]])` into a temporary list using `[[Get]]` (let the length of this list be `n`). `undefined`s are counted and not added to this temporary list (let this count be `m`).
1. Sort this temporary list using an implementation-specific sort algorithm.
1. Write back `n` sorted values using `[[Set]]`.
1. Write back `m` `undefined`s.
1. Perform `[[Delete]]` on integer-indexed properties in the range of `[n + m, [[length]])` as holes are moved to the end of the sorting range.

## Advantages

The main advantage is that the following cases **no longer result in an implementation-defined sort order**:

* _sparse_ objects with elements on the prototype chain (e.g. an array with a hole at index `3` when `Object.prototype[3] = 42`)
* objects whose `[[Get]]`, `[[Set]]` behavior for integer-indexed properties inside the sorting range is not the "ordinary" implementation (i.e. a getter or setter or a proxy).
* non-extensible objects (see next paragraph)
* objects with non-configurable or non-writable properties inside the sorting range (see next paragraph)

The reason for this is that `[[Get]]`, `[[Set]]` and `[[Delete]]` are now called in a well-defined order, regardless of the chosen sorting algorithm of any engine. (Previously, non-extensible objects and non-configurable/non-writable properties could result in exceptions at any point in time during sorting, depending on the implementation.)

Additionally, `ToString` operations or comparison functions that throw now leave the object as-is (modulo side-effects and changes caused by `[[Get]]`).

## Web compatibility

We believe this change to be web-compatible, given that it has been (for the most part) shipping since Chrome 74, with the remaining bits shipping since Chrome 76.
- Starting with V8 v7.4 (Chrome 74), V8 copies non-`undefined` values into a temporary list for sorting, as described in this PR. However, compacting all values at the start of the sorting range (removing holes) still happened on the object itself and was observable.

- As of V8 v7.6 (Chrome version 76.0.3806.0) V8's implementation of `Array.prototype.sort` fully behaves as described by this PR.

## Memory concerns

Implementers might be concerned about the added `O(n)` memory requirements of `Array.prototype.sort`. The first version in V8 that applied `Array.prototype.sort` to a temporary copy landed with Chrome 74. Worried about the impact on memory consumption (especially on mobile), we analysed memory usage of some 1000 websites and **did not see any change in peak memory consumption**.

## Variants

This is only a first initial draft of how `Array.prototype.sort` could be specified in a more precise manner. A few variations come to mind:

* Instead of counting `undefined`s, they could be collected as well and passed to `SortCompare`. This would require `SortCompare` to be left as-is and handle `undefined`s appropriately.
* The spec could go further. In the case of an `undefined` comparison function, the initial collection phase could call `ToString` directly. This means the temporary list would contain pairs consisting of the original value and the result of `ToString` for the respective value. This would further reduce the cases of implementation-defined sort order. 

Ref. #302.
@mathiasbynens @ajklein

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -32488,78 +32488,54 @@ THH:mm:ss.sss
       <emu-clause id="sec-array.prototype.sort">
         <h1>Array.prototype.sort ( _comparefn_ )</h1>
         <p>The elements of this array are sorted. The sort must be stable (that is, elements that compare equal must remain in their original order). If _comparefn_ is not *undefined*, it should be a function that accepts two arguments _x_ and _y_ and returns a negative value if _x_ &lt; _y_, zero if _x_ = _y_, or a positive value if _x_ &gt; _y_.</p>
-        <p>Upon entry, the following steps are performed to initialize evaluation of the `sort` function:</p>
+        <p>The following steps are taken:</p>
         <emu-alg>
-          1. If _comparefn_ is not *undefined* and IsCallable(_comparefn_) is *false*, throw a *TypeError* exception.
+          1. [id="step-array-sort-comparefn"] If _comparefn_ is not *undefined* and IsCallable(_comparefn_) is *false*, throw a *TypeError* exception.
           1. Let _obj_ be ? ToObject(*this* value).
-          1. Let _len_ be ? LengthOfArrayLike(_obj_).
-        </emu-alg>
-        <p>Within this specification of the `sort` method, an object, _obj_, is said to be <em>sparse</em> if the following algorithm returns *true*:</p>
-        <emu-alg>
-          1. For each integer _i_ in the range 0 &le; _i_ &lt; _len_, do
-            1. Let _prop_ be ! ToString(𝔽(_i_)).
-            1. Let _elem_ be _obj_.[[GetOwnProperty]](_prop_).
-            1. If _elem_ is *undefined*, return *true*.
-          1. Return *false*.
+          1. [id="step-array-sort-len"] Let _len_ be ? LengthOfArrayLike(_obj_).
+          1. Let _items_ be a new empty List.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Let _kPresent_ be ? HasProperty(_obj_, _Pk_).
+            1. If _kPresent_ is *true*, then
+              1. Let _kValue_ be ? Get(_obj_, _Pk_).
+              1. Append _kValue_ to _items_.
+            1. Set _k_ to _k_ + 1.
+          1. Let _itemCount_ be the number of elements in _items_.
+          1. [id="step-array-sort"] Sort _items_ using an implementation-defined sequence of calls to SortCompare. If any such call returns an abrupt completion, stop before performing any further calls to SortCompare or steps in this algorithm and return that completion.
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _itemCount_,
+            1. Perform ? Set(_obj_, ! ToString(𝔽(_j_)), _items_[_j_], *true*).
+            1. Set _j_ to _j_ + 1.
+          1. Repeat, while _j_ &lt; _len_,
+            1. Perform ? DeletePropertyOrThrow(_obj_, ! ToString(𝔽(_j_))).
+            1. Set _j_ to _j_ + 1.
+          1. Return _obj_.
         </emu-alg>
         <p>The <em>sort order</em> is the ordering, after completion of this function, of the <emu-xref href="#integer-index">integer-indexed</emu-xref> property values of _obj_ whose integer indexes are less than _len_. The result of the `sort` function is then determined as follows:</p>
-        <p>If _comparefn_ is not *undefined* and is not a consistent comparison function for the elements of this array (see below), the sort order is implementation-defined. The sort order is also implementation-defined if _comparefn_ is *undefined* and SortCompare does not act as a consistent comparison function.</p>
-        <p>Let _proto_ be _obj_.[[GetPrototypeOf]](). If _proto_ is not *null* and there exists an integer _j_ such that all of the conditions below are satisfied then the sort order is implementation-defined:</p>
+        <p>The sort order is implementation-defined if any of the following conditions is true:</p>
         <ul>
           <li>
-            _obj_ is sparse
+            If _comparefn_ is not *undefined* and is not a consistent comparison function for the elements of _items_ (see below).
           </li>
           <li>
-            0 &le; _j_ &lt; _len_
-          </li>
-          <li>
-            HasProperty(_proto_, ToString(𝔽(_j_))) is *true*.
-          </li>
-        </ul>
-        <p>The sort order is also implementation-defined if _obj_ is sparse and any of the following conditions are true:</p>
-        <ul>
-          <li>
-            IsExtensible(_obj_) is *false*.
-          </li>
-          <li>
-            Any integer index property of _obj_ whose name is a non-negative integer less than _len_ is a data property whose [[Configurable]] attribute is *false*.
-          </li>
-        </ul>
-        <p>The sort order is also implementation-defined if any of the following conditions are true:</p>
-        <ul>
-          <li>
-            If _obj_ is an exotic object (including Proxy exotic objects) whose behaviour for [[Get]], [[Set]], [[Delete]], and [[GetOwnProperty]] is not the ordinary object implementation of these internal methods.
-          </li>
-          <li>
-            If any index property of _obj_ whose name is a non-negative integer less than _len_ is an accessor property or is a data property whose [[Writable]] attribute is *false*.
-          </li>
-          <li>
-            If _comparefn_ is *undefined* and the application of ToString to any value passed as an argument to SortCompare modifies _obj_ or any object on _obj_'s prototype chain.
+            If _comparefn_ is *undefined* and SortCompare does not act as a consistent comparison function.
           </li>
           <li>
             If _comparefn_ is *undefined* and all applications of ToString, to any specific value passed as an argument to SortCompare, do not produce the same result.
           </li>
         </ul>
-        <p>The following steps are taken:</p>
-        <emu-alg>
-          1. Perform an implementation-defined sequence of calls to the Get, <emu-xref href="#sec-set-o-p-v-throw">Set</emu-xref>, DeletePropertyOrThrow, and HasOwnProperty abstract operation with _obj_ as the first argument, and to SortCompare (described below), such that:
-            * The property key argument for each call to Get, <emu-xref href="#sec-set-o-p-v-throw">Set</emu-xref>, HasOwnProperty, or DeletePropertyOrThrow is the string representation of a non-negative integer less than _len_.
-            * The `Throw` argument for every call to <emu-xref href="#sec-set-o-p-v-throw">Set</emu-xref> is *true*.
-            * The arguments for calls to SortCompare are values returned by a previous call to the Get abstract operation, unless the properties accessed by those previous calls did not exist according to HasOwnProperty. If both prospective arguments to SortCompare correspond to non-existent properties, use *+0*<sub>𝔽</sub> instead of calling SortCompare. If only the first prospective argument is non-existent, use *1*<sub>𝔽</sub>. If only the second prospective argument is non-existent, use *-1*<sub>𝔽</sub>.
-            * If _obj_ is not sparse then DeletePropertyOrThrow must not be called.
-            * If an abrupt completion is returned from any of these operations, it is immediately returned as the value of this function.
-          1. Return _obj_.
-        </emu-alg>
-        <p>Unless the sort order is specified above to be implementation-defined, the returned object must have the following two characteristics:</p>
+        <p>Unless the sort order is specified above to be implementation-defined, _items_ must satisfy all of the following conditions after executing step <emu-xref href="#step-array-sort"></emu-xref> of the algorithm above:</p>
         <ul>
           <li>
-            There must be some mathematical permutation &pi; of the non-negative integers less than _len_, such that for every non-negative integer _j_ less than _len_, if property <emu-eqn>old[_j_]</emu-eqn> existed, then <emu-eqn>new[&pi;(_j_)]</emu-eqn> is exactly the same value as <emu-eqn>old[_j_]</emu-eqn>. But if property <emu-eqn>old[_j_]</emu-eqn> did not exist, then <emu-eqn>new[&pi;(_j_)]</emu-eqn> does not exist.
+            There must be some mathematical permutation &pi; of the non-negative integers less than _itemCount_, such that for every non-negative integer _j_ less than _itemCount_, the element <emu-eqn>old[_j_]</emu-eqn> is exactly the same as <emu-eqn>new[&pi;(_j_)]</emu-eqn>.
           </li>
           <li>
-            Then for all non-negative integers _j_ and _k_, each less than _len_, if <emu-eqn>SortCompare(old[_j_], old[_k_]) &lt; 0</emu-eqn> (see SortCompare below), then <emu-eqn>new[&pi;(_j_)] &lt; new[&pi;(_k_)]</emu-eqn>.
+            Then for all non-negative integers _j_ and _k_, each less than _itemCount_, if <emu-eqn>SortCompare(old[_j_], old[_k_]) &lt; 0</emu-eqn> (see SortCompare below), then <emu-eqn>&pi;(_j_) &lt; &pi;(_k_)</emu-eqn>.
           </li>
         </ul>
-        <p>Here the notation <emu-eqn>old[_j_]</emu-eqn> is used to refer to the hypothetical result of calling <emu-eqn>Get(_obj_, _j_)</emu-eqn> before this function is executed, and the notation <emu-eqn>new[_j_]</emu-eqn> to refer to the hypothetical result of calling <emu-eqn>Get(_obj_, _j_)</emu-eqn> after this function has been executed.</p>
+        <p>Here the notation <emu-eqn>old[_j_]</emu-eqn> is used to refer to <emu-eqn>_items_[_j_]</emu-eqn> before step <emu-xref href="#step-array-sort"></emu-xref> is executed, and the notation <emu-eqn>new[_j_]</emu-eqn> to refer to <emu-eqn>_items_[_j_]</emu-eqn> after step <emu-xref href="#step-array-sort"></emu-xref> has been executed.</p>
         <p>A function _comparefn_ is a consistent comparison function for a set of values _S_ if all of the requirements below are met for all values _a_, _b_, and _c_ (possibly the same value) in the set _S_: The notation <emu-eqn>_a_ &lt;<sub>CF</sub> _b_</emu-eqn> means <emu-eqn>_comparefn_(_a_, _b_) &lt; 0</emu-eqn>; <emu-eqn>_a_ =<sub>CF</sub> _b_</emu-eqn> means <emu-eqn>_comparefn_(_a_, _b_) = 0</emu-eqn> (of either sign); and <emu-eqn>_a_ &gt;<sub>CF</sub> _b_</emu-eqn> means <emu-eqn>_comparefn_(_a_, _b_) &gt; 0</emu-eqn>.</p>
         <ul>
           <li>
@@ -33677,16 +33653,15 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-%typedarray%.prototype.sort">
         <h1>%TypedArray%.prototype.sort ( _comparefn_ )</h1>
-        <p>%TypedArray%`.prototype.sort` is a distinct function that, except as described below, implements the same requirements as those of `Array.prototype.sort` as defined in <emu-xref href="#sec-array.prototype.sort"></emu-xref>. The implementation of the %TypedArray%`.prototype.sort` specification may be optimized with the knowledge that the *this* value is an object that has a fixed length and whose <emu-xref href="#integer-index">integer-indexed</emu-xref> properties are not sparse. The only internal methods of the *this* value that the algorithm may call are [[Get]] and [[Set]].</p>
+        <p>%TypedArray%`.prototype.sort` is a distinct function that, except as described below, implements the same requirements as those of `Array.prototype.sort` as defined in <emu-xref href="#sec-array.prototype.sort"></emu-xref>. The implementation of the %TypedArray%`.prototype.sort` specification may be optimized with the knowledge that the *this* value is an object that has a fixed length and whose <emu-xref href="#integer-index">integer-indexed</emu-xref> properties are not sparse.</p>
         <p>This function is not generic. The *this* value must be an object with a [[TypedArrayName]] internal slot.</p>
-        <p>Upon entry, the following steps are performed to initialize evaluation of the `sort` function. These steps are used instead of the entry steps in <emu-xref href="#sec-array.prototype.sort"></emu-xref>:</p>
+        <p>Upon entry, the following steps are performed to initialize evaluation of the `sort` function. These steps are used instead of steps <emu-xref href="#step-array-sort-comparefn"></emu-xref>–<emu-xref href="#step-array-sort-len"></emu-xref> in <emu-xref href="#sec-array.prototype.sort"></emu-xref>:</p>
         <emu-alg>
           1. If _comparefn_ is not *undefined* and IsCallable(_comparefn_) is *false*, throw a *TypeError* exception.
           1. Let _obj_ be the *this* value.
           1. Let _buffer_ be ? ValidateTypedArray(_obj_).
           1. Let _len_ be _obj_.[[ArrayLength]].
         </emu-alg>
-        <p>The implementation-defined sort order condition for exotic objects is not applied by %TypedArray%`.prototype.sort`.</p>
         <p>The following version of SortCompare is used by %TypedArray%`.prototype.sort`. It performs a numeric comparison rather than the string comparison used in <emu-xref href="#sec-array.prototype.sort"></emu-xref>.</p>
         <p>The abstract operation TypedArraySortCompare takes arguments _x_ and _y_. It also has access to the _comparefn_ and _buffer_ values of the current invocation of the `sort` method. It performs the following steps when called:</p>
         <emu-alg>


spec.html
```diff
@@ -32544,67 +32544,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            If _comparefn_ is not *undefined* and is not a consistent comparison function for the elements of _items_ (see below).
           </li>
           <li>
-            HasProperty(_proto_, ToString(_j_)) is *true*.
-          </li>
-        </ul>
-        <p>The sort order is also implementation-defined if _obj_ is sparse and any of the following conditions are true:</p>
-        <ul>
-          <li>
-            IsExtensible(_obj_) is *false*.
```
szuend: [[Set]] and [[Delete]] is now called in a well-defined order. Additionally, if any ToString or comparison function tries to add properties to the object, the sort-order would still be the same across implementations, as no values have been written back yet.

spec.html
```diff
@@ -32544,67 +32544,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
-          </li>
-          <li>
-            Any integer index property of _obj_ whose name is a nonnegative integer less than _len_ is a data property whose [[Configurable]] attribute is *false*.
```
szuend: Same argument as above.

spec.html
```diff
@@ -32544,67 +32544,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
-          </li>
-        </ul>
-        <p>The sort order is also implementation-defined if any of the following conditions are true:</p>
-        <ul>
-          <li>
-            If _obj_ is an exotic object (including Proxy exotic objects) whose behaviour for [[Get]], [[Set]], [[Delete]], and [[GetOwnProperty]] is not the ordinary object implementation of these internal methods.
```
szuend: Same as above, these are invoked in the same order independent of the concrete sorting algorithm used.

spec.html
```diff
@@ -32544,67 +32544,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
           </li>
         </ul>
-        <p>The following steps are taken:</p>
-        <emu-alg>
-          1. Perform an implementation-dependent sequence of calls to the [[Get]] and [[Set]] internal methods of _obj_, to the DeletePropertyOrThrow and HasOwnProperty abstract operation with _obj_ as the first argument, and to SortCompare (described below), such that:
-            * The property key argument for each call to [[Get]], [[Set]], HasOwnProperty, or DeletePropertyOrThrow is the string representation of a nonnegative integer less than _len_.
-            * The arguments for calls to SortCompare are values returned by a previous call to the [[Get]] internal method, unless the properties accessed by those previous calls did not exist according to HasOwnProperty. If both prospective arguments to SortCompare correspond to non-existent properties, use *+0* instead of calling SortCompare. If only the first prospective argument is non-existent use +1. If only the second prospective argument is non-existent use -1.
```
szuend: The new algorithm above might need some clarification. This is, SortCompare should only be called with values retrieved by ``[[Get]]`` and are part of `items`.

spec.html
```diff
@@ -32544,67 +32544,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
-            * If _obj_ is not sparse then DeletePropertyOrThrow must not be called.
```
szuend: This is implicitly the case, as ``HasProperty(i)`` for any index i in the range of [0, `[[length]]`) always returns true, and thus `|items| + number of undefineds` should be equalt to `[[length]]`.

spec.html
```diff
@@ -32650,9 +32638,6 @@ <h1>Array.prototype.sort ( _comparefn_ )</h1>
           <h1>Runtime Semantics: SortCompare ( _x_, _y_ )</h1>
           <p>The SortCompare abstract operation is called with two arguments _x_ and _y_. It also has access to the _comparefn_ argument passed to the current invocation of the `sort` method. The following steps are taken:</p>
           <emu-alg>
-            1. If _x_ and _y_ are both *undefined*, return *+0*.
-            1. If _x_ is *undefined*, return 1.
-            1. If _y_ is *undefined*, return -1.
```
szuend: `undefined`s are never part of `items`, and thus, `x` and `y` can never be `undefined`.

mathiasbynens:
To make this easier to follow, Simon left inline comments on the diff. We've also prepared a [slide deck](https://docs.google.com/presentation/d/1eFvK__9kRwHnkzZfNEL9FS3d9TJSRtxlXe3NVGmVVAU/edit) that walks through the proposed algorithm in a more visual way, which I'll present at the upcoming TC39 meeting in July.

One question people might have is: to which extent does this PR match implementation reality? We mentioned that V8 implements `Array#sort` as described in the PR, but what do other engines do? Below are some test cases.

### `test-1.js`

Here's the example we walk through in the slide deck. We're sorting an array that contains some `undefined` values and some holes, one of which results in `Object.prototype[2]` being read.

All engines produce the same result here.

#### Code

```js
function log(array) {
  let buf = '';
  for (let index = 0; index < array.length; index++) {
    if (array.hasOwnProperty(index)) {
      buf += String(array[index]);
    } else {
      buf += 'hole';
    }
    if (index < array.length - 1) buf += ',';
  }
  print(buf);
}

/* */

Object.prototype[2] = 4;
const array = [undefined, 3, /*hole*/, 2, undefined, /*hole*/, 1];
array.sort();
log(array);
```

#### Output

```
$ eshost -s test.js
#### Chakra, JavaScriptCore, SpiderMonkey, V8, XS
1,2,3,4,undefined,undefined,hole
```

### `test-2.js`

More interesting version of the same example, with an added getter/setter pair named `2` on `Object.prototype`. When we hit the hole at index 2, we hit the accessor.

Except for JavaScriptCore, all engines have the same behavior: the getter is called once, the setter is called once with value `3`, and then the sorted result is logged. JavaScriptCore instead hits the accessor multiple times, and eventually throws a `TypeError` (which seems like a bug regardless of this proposed change).

#### Code

```js
function log(array) {
  let buf = '';
  for (let index = 0; index < array.length; index++) {
    if (array.hasOwnProperty(index)) {
      buf += String(array[index]);
    } else {
      buf += 'hole';
    }
    if (index < array.length - 1) buf += ',';
  }
  print(buf);
}

/* */

Object.defineProperty(Object.prototype, '2', {
  get() { print('get'); return 4; },
  set(v) { print(`set with ${v}`); }
});
const array = [undefined, 3, /*hole*/, 2, undefined, /*hole*/, 1];
array.sort();
log(array);
```

#### Output

```
$ eshost -s test.js
#### Chakra, SpiderMonkey, V8, XS
get
set with 3
1,2,hole,4,undefined,undefined,hole

#### JavaScriptCore
get
set with 2
get
get
set with [object Object]
get
TypeError: undefined is not an object
```

### `test-3.js`

This example uses a custom comparison function which throws in the middle of sorting (on the third call). The proposed change doesn't modify the original array until the very end, and so in this case, the original array remains unchanged.

Once again, this matches existing implementations, except for JavaScriptCore, which mutates the original array until the exception is thrown.

#### Code

```js
function log(array) {
  let buf = '';
  for (let index = 0; index < array.length; index++) {
    if (array.hasOwnProperty(index)) {
      buf += String(array[index]);
    } else {
      buf += 'hole';
    }
    if (index < array.length - 1) buf += ',';
  }
  print(buf);
}

/* */

Object.defineProperty(Object.prototype, '2', {
  get() { print('get'); return 4; },
  set(v) { print(`set with ${v}`); }
});
const array = [undefined, 3, /*hole*/, 2, undefined, /*hole*/, 1];
let count = 0;
try {
  array.sort((a, b) => {
    if (++count === 3) {
      throw new Error('lolwat');
    }
    return b - a;
  });
} catch (exception) {
  print(exception);
}
log(array);
```

#### Output

```
$ eshost -s test.js
#### Chakra, SpiderMonkey, V8, XS
get
Error: lolwat
undefined,3,hole,2,undefined,hole,1

#### JavaScriptCore
get
set with 2
get
get
set with 4
get
Error: omg
3,4,hole,1,undefined,undefined,hole
```

spec.html
```diff
@@ -32544,67 +32544,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. Let _kPresent_ be ? HasProperty(_obj_, _Pk_).
+            1. If _kPresent_ is *true*, then
+              1. Let _kValue_ be ? Get(_obj_, _Pk_).
+              1. If _kValue_ is *undefined*, then
+                1. Set _undefinedCount_ to _undefinedCount_ + 1.
+              1. Else,
+                1. Set _items_[_itemCount_] to _kValue_.
+                1. Set _itemCount_ to _itemCount_ + 1.
+            1. Set _k_ to _k_ + 1.
+          1. Sort _items_ using an implementation-dependant sequence of calls to SortCompare.
```
```suggestion
          1. Sort _items_ using an implementation-dependent sequence of calls to SortCompare.
```

mathiasbynens: FWIW I also prefer this spelling. We just stuck to "dependant" because that's what the current spec uses.

ljharb: I believe the a spelling only applies in old British English and only when used as a noun; since here it’s an adjective, the e spelling is universally correct.

I’ll audit the other places in the spec and update them in a separate PR.

ljharb: (i can't find any uses of `dependant` in the current spec to follow up with)

spec.html
```diff
@@ -32544,67 +32544,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _undefinedCount_ be 0.
+          1. Repeat, while _k_ &lt; _len_
+            1. Let _Pk_ be ! ToString(_k_).
+            1. Let _kPresent_ be ? HasProperty(_obj_, _Pk_).
+            1. If _kPresent_ is *true*, then
+              1. Let _kValue_ be ? Get(_obj_, _Pk_).
+              1. If _kValue_ is *undefined*, then
+                1. Set _undefinedCount_ to _undefinedCount_ + 1.
+              1. Else,
+                1. Set _items_[_itemCount_] to _kValue_.
```
jmdyck: I'm not positive, but I don't think the spec ever Sets a list element at an index that doesn't exist yet. It would be more idiomatic (and clearer, I think) to simply `Append _kvalue_ to _items_` and then after the loop, `Set _itemCount_ to the length of _items_`.

spec.html
```diff
@@ -32544,67 +32544,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+                1. Set _itemCount_ to _itemCount_ + 1.
+            1. Set _k_ to _k_ + 1.
+          1. Sort _items_ using an implementation-dependent sequence of calls to SortCompare.
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _itemCount_
+            1. Perform ? Set(_obj_, ! ToString(_j_), items[_j_], *true*).
```
```suggestion
            1. Perform ? Set(_obj_, ! ToString(_j_), _items_[_j_], *true*).
```

spec.html
```diff
@@ -32544,67 +32544,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. Set _j_ to _j_ + 1.
+          1. Repeat, while _j_ &lt; _itemCount_ + _undefinedCount_
+            1. Perform ? Set(_obj_, ! ToString(_j_), *undefined*, *true*).
+            1. Set _j_ to _j_ + 1.
+          1. Repeat, while _j_ &lt; len
```
```suggestion
          1. Repeat, while _j_ &lt; _len_
```

spec.html
szuend: Done. Although I moved the declaration of _itemCount_ after the loop, as _itemCount_ is no longer used in the loop. Also addressed the other comments.

spec.html
```diff
@@ -32544,67 +32544,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _itemCount_
+            1. Perform ? Set(_obj_, ! ToString(_j_), items[_j_], *true*).
+            1. Set _j_ to _j_ + 1.
+          1. Repeat, while _j_ &lt; _itemCount_ + _undefinedCount_
+            1. Perform ? Set(_obj_, ! ToString(_j_), *undefined*, *true*).
+            1. Set _j_ to _j_ + 1.
+          1. Repeat, while _j_ &lt; _len_
+            1. Let _Pj_ be ! ToString(_j_).
+            1. Let _jPresent_ be ? HasOwnProperty(_obj_, _Pj_).
```
anba: It should be possible to omit this call to `HasOwnProperty` and unconditionally call `DeletePropertyOrThrow`. (This matches what SpiderMonkey does [here](https://searchfox.org/mozilla-central/rev/da855d65d1fbdd714190cab2c46130f7422f3699/js/src/builtin/Sorting.js#263-268) and [here](https://searchfox.org/mozilla-central/rev/da855d65d1fbdd714190cab2c46130f7422f3699/js/src/builtin/Sorting.js#263-268).)

szuend: I checked the spec of `DeletePropertyOrThrow` and the `delete` syntax (used by SpiderMoneky), and we should indeed be able to remove the `HasOwnProperty` check. Any other opinions?

@ljharb

ljharb: What if `obj` is a proxy - wouldn't removing this "has" check be observable?

It does seem, though, like `obj.[[Delete]](P)` will return `true` if the property was never present to begin with, which suggests the check can be removed.

bakkot:
I'd like to see more investigation of the behavior of existing engines in edge cases. Off the top of my head:

- accessors on the array itself
- accessors on a different object, where the array's prototype has been manually changed to that object
- non-writeable or non-configurable properties on the array
- non-writeable or non-configurable properties on Object.prototype in a position where there is a hole in the array
- what happens when values in the array are mutated as a side effect of an accessor or the comparison function or the `toString` of an array element
- what happens when holes in the array are created or removed as a side effect of an accessor or the comparison function or the `toString` of an array element
- what happens when the length of the array is mutated as a side effect of an accessor or the comparison function or the `toString` of an array element
- what happens for arrays of length 0 or 1, especially when there is a hole or an accessor pair on the array at index 0, or when there is a side effect in the comparison function

I am sure there are more cases to consider, too.

anba:
> I'd like to see more investigation of the behavior of existing engines in edge cases.

SpiderMonkey always first collects all elements, stores them in a temporary object, and then performs the search in that temporary object:
- Self-hosted search algorithm: https://searchfox.org/mozilla-central/rev/40e889be8ff926e32f7567957f4c316f14f6fbef/js/src/builtin/Sorting.js#274-280
- C++ version used in some cases: https://searchfox.org/mozilla-central/rev/40e889be8ff926e32f7567957f4c316f14f6fbef/js/src/builtin/Array.cpp#2227-2287 

The elements in the sorted object are then written back with [[Set]] semantics and holes are deleted through [[Delete]]:
- Self-hosted: https://searchfox.org/mozilla-central/rev/40e889be8ff926e32f7567957f4c316f14f6fbef/js/src/builtin/Sorting.js#264-267
- C++: https://searchfox.org/mozilla-central/rev/40e889be8ff926e32f7567957f4c316f14f6fbef/js/src/builtin/Array.cpp#2342-2344,2347-2360

spec.html
szuend: I removed the `HasOwnProperty` check. The corresponding [V8 CL](https://crrev.com/c/1809366) landed yesterday.

ljharb:
@szuend if you're able to rebase this on latest master, a rendered preview will be generated.

szuend:
> I'd like to see more investigation of the behavior of existing engines in edge cases.

I implemented some of these [here](https://gist.github.com/szuend/05ae15b4e1329b264ab4c9a1cda09242). All tests were run using `eshost -s <file>`. The output of each test case can be found here:

**Accessor on the array itself (`accessors-on-array.js`)**
```
#### chakra, spidermonkey, v8
get [2]
get [3]
set [2] with d
set [3] with undefined
get [2]
get [3]
a,c,d,undefined,undefined,undefined,undefined,hole

#### javascriptcore
get [2]
get [3]
set [2] with d
set [3] with undefined
get [2]
get [2]
set [2] with d
get [2]
get [3]
a,c,d,undefined,undefined,undefined,undefined,hole

#### xs
get [2]
get [3]
set [2] with d
TypeError: ?.set: cannot coerce undefined to object
```

**Accessors on the arrays prototype object with a hole (`accessors-on-object-proto.js`)**
```
#### chakra, spidermonkey, v8, xs
get [2]
set [2] with c
a,b,hole,d,undefined,undefined,undefined,hole

#### javascriptcore
get [2]
set [2] with a
get [2]
get [2]
set [2] with c
a,b,hole,d,undefined,undefined,undefined,hole
```

**non-writeable or non-configurable elements on the array or its prototype**
**`non-configurable-element.js`**
```
#### chakra

TypeError: Object doesn't support this action

#### javascriptcore

TypeError: Unable to delete property.

#### spidermonkey

TypeError: property 7 is non-configurable and can't be deleted

#### v8

TypeError: Cannot delete property '6' of [object Array]

#### xs

TypeError: Array.prototype.sort: delete 6: not configurable
```

**`non-configurable-proto-hole.js`**
```
#### chakra, javascriptcore, spidermonkey, v8, xs
a,b,c,d,foo,undefined,undefined,hole
```

**`non-writeable-element.js`**
```
#### chakra

TypeError: Object doesn't support this action

#### javascriptcore

TypeError: Attempted to assign to readonly property.

#### spidermonkey

TypeError: 1 is read-only

#### v8

TypeError: Cannot assign to read only property '1' of object '[object Array]'

#### xs

TypeError: Array.prototype.sort: C: xsSet 0: not writable

```

**`non-writeable-proto-hole.js`**
```
#### chakra

TypeError: Object doesn't support this action

#### javascriptcore

TypeError: Attempted to assign to readonly property.

#### spidermonkey

TypeError: 2 is read-only

#### v8

TypeError: Cannot assign to read only property '2' of object '[object Array]'

#### xs

TypeError: Array.prototype.sort: C: xsSet 0: not writable
```

**Accessors mutate an element**
**`accessor-sets-predecessor.js`**
```
#### chakra, spidermonkey, v8, xs
a,b,c,d,undefined,undefined,undefined,hole
a,foobar,c,d,undefined,undefined,undefined,hole

#### javascriptcore
a,b,c,d,undefined,undefined,undefined,hole
a,foobar,d,foobar,undefined,undefined,undefined,hole
```

**`accessor-sets-successor.js`**
```
#### chakra, spidermonkey, v8, xs
a,c,d,foobar,undefined,undefined,undefined,hole
a,b,c,d,undefined,undefined,undefined,hole

#### javascriptcore
a,c,foobar,foobar,undefined,undefined,undefined,hole
a,b,c,d,undefined,undefined,undefined,hole
```

**Accessors delete an element**
**`accessor-deletes-predecessor.js`**
```
#### chakra, spidermonkey, v8, xs
a,b,c,d,undefined,undefined,undefined,hole
a,hole,c,d,undefined,undefined,undefined,hole

#### javascriptcore
a,b,c,d,undefined,undefined,undefined,hole
a,hole,d,undefined,undefined,undefined,undefined,hole
```

**`accessor-deletes-successor.js`**
```
#### chakra, javascriptcore, spidermonkey, v8, xs
a,c,d,hole,undefined,undefined,hole,hole
a,b,c,d,undefined,undefined,undefined,hole
```

**Accessor add/remove elements**
**`accessor-adds-two-elements.js`**
```
#### chakra, spidermonkey, v8, xs
a,b,c,d,undefined,undefined,undefined,hole,foo,bar,foo,bar
a,b,c,d,undefined,undefined,undefined,hole,foo,bar

#### javascriptcore
a,b,c,d,undefined,undefined,undefined,hole,foo,bar,foo,bar,foo,bar,foo,bar
a,b,c,d,undefined,undefined,undefined,hole,foo,bar,foo,bar
```

**`accessor-removes-two-elements.js`**
```
#### chakra, spidermonkey, v8, xs
b,c,undefined,undefined
a,b,c,d,undefined,undefined,undefined

#### javascriptcore
b,c,undefined,undefined
a,b,c,undefined
```

**Accessors modify `.length`**
**`accessor-decreases-length.js`**
```
#### chakra, spidermonkey, v8, xs
b,c,undefined,undefined
a,b,c,d,undefined,undefined,undefined

#### javascriptcore
b,c,undefined,undefined
a,b,c,undefined
```

**`accessor-increases-length.js`**
```
#### chakra, spidermonkey, v8, xs
a,b,c,d,undefined,undefined,undefined,hole,hole,hole,hole,hole
a,b,c,d,undefined,undefined,undefined,hole,hole,hole

#### javascriptcore
a,b,c,d,undefined,undefined,undefined,hole,hole,hole,hole,hole,hole,hole,hole,hole
a,b,c,d,undefined,undefined,undefined,hole,hole,hole,hole,hole
```

**0-length and 1-length arrays (`short-arrays.js`)**
```
#### chakra
0-length array:
get [0]
set [0] with undefined
get [0]
undefined
1-length array:
get [0]
set [0] with bar
get [0]
bar,undefined

#### javascriptcore
0-length array:
get [0]
undefined
1-length array:
get [0]
get [0]
set [0] with bar
get [0]
get [0]
set [0] with bar
get [0]
bar,undefined

#### spidermonkey, v8
0-length array:
get [0]
undefined
1-length array:
get [0]
set [0] with bar
get [0]
bar,undefined

#### xs
0-length array:
get [0]
undefined
1-length array:
TypeError: Array.prototype.sort: Cannot coerce to string
```

**Sort on a proxy with a backing array `proxy.js`**
```
#### chakra
get ['length']
get ['0']
get ['1']
get ['2']
get ['3']
get ['4']
get ['5']
get ['6']
get ['7']
set ['0'] = a
set ['1'] = b
set ['2'] = c
set ['3'] = d
set ['4'] = undefined
set ['5'] = undefined
set ['6'] = undefined
set ['7'] = undefined

#### javascriptcore
get ['length']
get ['length']
get ['0']
has ['0']
get ['0']
has ['1']
get ['1']
set ['0'] = c
has ['2']
has ['3']
get ['3']
set ['1'] = b
has ['4']
get ['4']
has ['5']
has ['6']
get ['6']
set ['2'] = a
has ['7']
get ['7']
set ['3'] = d
set ['4'] = undefined
set ['5'] = undefined
delete ['6']
delete ['7']
get ['0']
get ['0']
get ['1']
get ['1']
get ['2']
get ['2']
get ['3']
get ['3']
set ['0'] = a
set ['1'] = b
set ['2'] = c
set ['3'] = d

#### spidermonkey
get ['length']
has ['0']
get ['0']
has ['1']
get ['1']
has ['2']
has ['3']
get ['3']
has ['4']
get ['4']
has ['5']
has ['6']
get ['6']
has ['7']
get ['7']
set ['0'] = a
set ['1'] = b
set ['2'] = c
set ['3'] = d
set ['4'] = undefined
set ['5'] = undefined
delete ['7']
delete ['6']

#### v8, xs
get ['length']
has ['0']
get ['0']
has ['1']
get ['1']
has ['2']
has ['3']
get ['3']
has ['4']
get ['4']
has ['5']
has ['6']
get ['6']
has ['7']
get ['7']
set ['0'] = a
set ['1'] = b
set ['2'] = c
set ['3'] = d
set ['4'] = undefined
set ['5'] = undefined
delete ['6']
delete ['7']
```

**Sort an array with a proxy prototype `proxy-proto.js`**
```
#### chakra
get ['2']
get ['5']
set ['2'] = c
set ['5'] = undefined

#### javascriptcore
has ['2']
has ['5']
set ['2'] = a
set ['5'] = undefined
get ['2']
get ['2']
set ['2'] = c

#### spidermonkey, v8, xs
has ['2']
has ['5']
set ['2'] = c
set ['5'] = undefined
```

The result is not that surprising given that v8 implements basically the same approach now as spidermonkey. Please note that in the `proxy.js` example, the only difference between spidermonkey and v8/xs is the order of `[[Delete]]` calls.

bakkot:
Thanks for the more thorough tests!

It looks like JavaScriptCore disagrees with the proposed spec more often than it agrees, so this normative change mostly amounts to a request that JSC change their implementation of `Array.prototype.sort` (as well as a few smaller changes on the part of ChakraCore, SpiderMonkey, and xs). If JSC is willing to spend the resources necessary to make those changes, I'd be happy for this to be merged. If they are not, it would be a shame to spec something which is not going to match web reality.

mathiasbynens:
> If JSC is willing to spend the resources necessary to make those changes, I'd be happy for this to be merged. If they are not, it would be a shame to spec something which is not going to match web reality.

@kmiller68 @msaboff During the last meeting I had the impression you were willing to change the JSC implementation. Is this still the case? (We could also discuss this during the meeting, but if this turns out to be uncontroversial, then we could all save some valuable meeting time.)

mathiasbynens:
Per the recent TC39 meeting, this patch now has consensus. Here's what still needs to happen before it can be merged:

- @waldemarhorwat found some bugs in the currently-proposed spec text. Waldemar, please let us know what to fix :)
- @kmiller68 wanted to investigate the feasibility of implementing this in JSC, and promised to report back here.

kmiller68:
I’ve filed https://bugs.webkit.org/show_bug.cgi?id=202582 to track the necessary JSC changes. I’ll also update this thread with the results of my investigation.

waldemarhorwat:
Here are a few bugs:

> Unless the sort order is specified above to be implementation-defined, the returned object must have the following two characteristics:

> There must be some mathematical permutation π of the nonnegative integers less than len, such that for every nonnegative integer j less than len, if property old[j] existed, then new[π(j)] is exactly the same value as old[j]. But if property old[j] did not exist, then new[π(j)] does not exist.

That invariant is now broken.

> Then for all nonnegative integers j and k, each less than len, if SortCompare(old[j], old[k]) < 0 (see SortCompare below), then new[π(j)] < new[π(k)].

That invariant is now also broken due to the refactoring of SortCompare to take out the handling of `undefined` and move it to *undefinedCount*. What is the reason for that refactoring? It makes it more awkward to specify such invariants.

> NOTE 1. Because non-existent property values always compare greater than undefined property values, and undefined always compares greater than any other value, undefined property values always sort to the end of the result, followed by non-existent property values.

It's not clear what "compare" means here, but this is in the SortCompare section, so the presumable interpretation is "SortCompare". In that case, this is no longer valid due to the `undefined` refactoring moving it from SortCompare to *undefinedCount*.

szuend:
> Here are a few bugs:
> 
> > Unless the sort order is specified above to be implementation-defined, the returned object must have the following two characteristics:
> 
> > There must be some mathematical permutation π of the nonnegative integers less than len, such that for every nonnegative integer j less than len, if property old[j] existed, then new[π(j)] is exactly the same value as old[j]. But if property old[j] did not exist, then new[π(j)] does not exist.
> 
> That invariant is now broken.

Could you please elaborate here? I don't see how the invariant is broken. 

> > Then for all nonnegative integers j and k, each less than len, if SortCompare(old[j], old[k]) < 0 (see SortCompare below), then new[π(j)] < new[π(k)].
> 
> That invariant is now also broken due to the refactoring of SortCompare to take out the handling of `undefined` and move it to _undefinedCount_. What is the reason for that refactoring? It makes it more awkward to specify such invariants.
> 
> > NOTE 1. Because non-existent property values always compare greater than undefined property values, and undefined always compares greater than any other value, undefined property values always sort to the end of the result, followed by non-existent property values.
> 
> It's not clear what "compare" means here, but this is in the SortCompare section, so the presumable interpretation is "SortCompare". In that case, this is no longer valid due to the `undefined` refactoring moving it from SortCompare to _undefinedCount_.

We can revert the refactoring of `undefind` handling upfront (remove the `undefinedCount`) and handle them in `SortCompare` as in the current version of the spec. Engines are still able to implement `undefined` handling as it is initially specced in this PR as it would be not observable by the user (since the user-provided comparison function is not called if either comparison value is `undefined` in `SortCompare`). Would you agree with this sentiment?

waldemarhorwat:
> > Here are a few bugs:
> > > Unless the sort order is specified above to be implementation-defined, the returned object must have the following two characteristics:
> > 
> > 
> > > There must be some mathematical permutation π of the nonnegative integers less than len, such that for every nonnegative integer j less than len, if property old[j] existed, then new[π(j)] is exactly the same value as old[j]. But if property old[j] did not exist, then new[π(j)] does not exist.
> > 
> > 
> > That invariant is now broken.
> 
> Could you please elaborate here? I don't see how the invariant is broken.

Think about what happens when the prototype has some properties with numeric indices covered up by own properties with the same indices. After sorting the prototype ones can show through the holes punched by deleting properties.

> > > Then for all nonnegative integers j and k, each less than len, if SortCompare(old[j], old[k]) < 0 (see SortCompare below), then new[π(j)] < new[π(k)].
> > 
> > 
> > That invariant is now also broken due to the refactoring of SortCompare to take out the handling of `undefined` and move it to _undefinedCount_. What is the reason for that refactoring? It makes it more awkward to specify such invariants.
> > > NOTE 1. Because non-existent property values always compare greater than undefined property values, and undefined always compares greater than any other value, undefined property values always sort to the end of the result, followed by non-existent property values.
> > 
> > 
> > It's not clear what "compare" means here, but this is in the SortCompare section, so the presumable interpretation is "SortCompare". In that case, this is no longer valid due to the `undefined` refactoring moving it from SortCompare to _undefinedCount_.
> 
> We can revert the refactoring of upfront `undefind` handling (remove the `undefinedCount`) and handle them in `SortCompare` as in the current version of the spec. Engines are still able to implement `undefined` handling as it is initially specced in this PR as it would be not observable by the user (since the user-provided comparison function is not called if either comparison value is `undefined` in `SortCompare`). Would you agree with this sentiment?

Yes, that would be my preferred solution.

waldemarhorwat:
Also, think about how the invariants behave when some of the properties are getters/setters or non-writable. The existing specification just states that the sort order becomes implementation-defined. The new specification attempts to define it, but it fails to satisfy its own invariants.

szuend:
> Think about what happens when the prototype has some properties with numeric indices covered up by own properties with the same indices. After sorting the prototype ones can show through the holes punched by deleting properties.
> 
> Now, you might think you can get around this by stating the invariant in terms of _own_ properties. That will fail in a different way, in that the Has and Get calls in the algorithm can read properties from the prototype visible through holes.
> 

I see two solutions:
1. Re-introduce the notion of "sparseness" and all cases that could result in an "implementation-defined" sort order, like the current spec has
2. Remove the permutation invariant all together
3. Specify the permutation invariant w.r.t. to *items* from the algorithm, instead of the resulting object

I am open to further suggestions and solution preferences. Personally, I prefer 3. My arguments for that are:
- The invariants are used to nail down the behavior of the implementation-specific part of Array#sort. Since we reduced the implementation-specific part to "shuffle elements in *items* around", we could restrict the invariant to apply to that step only
- This would also allow us to keep *undefinedCount*, since the ordering invariant using `SortCompare` would actually reflect what is happening.
- It would keep the spec text shorter while still being well-defined
- It would bring Array#sort closer to how other builtins are defined. All steps are explicitly listed in an algorithm, while the one step that is implementation-specific has (more or less) clear invariants

mathiasbynens:
My thoughts on the three options @szuend outlined are as follows:

> 1. Re-introduce the notion of "sparseness" and all cases that could result in an "implementation-defined" sort order, like the current spec has

This would be a sad resolution. :-1: Reducing the amount of implementation-definedness is a good thing.

> 2. Remove the permutation invariant all together

Other methods like `Array.prototype.reverse` don't seem to have explicit invariants specified. I suppose `sort` is different precisely because of the implementation-defined parts -- @waldemarhorwat, do you know? However, if this patch can nail down the spec to reduce the implementation-definedness, the value of stating the invariant is arguably lessened.

I would personally prefer keeping the invariant, but I don't think removing it is an unreasonable option.

> 3. Specify the permutation invariant w.r.t. to items from the algorithm, instead of the resulting object

This would be my preferred solution. It enables us to keep the spec as precise as possible --- i.e., we wouldn't have to remove the `undefinedCount` logic from the patch, which makes sense given that this is exactly what implementations do. But, the invariant is still listed and scoped correctly. :+1:

szuend:
I updated this PR with two changes:

- As per @waldemarhorwat preference I removed the explicit `undefined` handling from the PR. Engines are still able to implement it in the same way as this PR specced it initially, since its transparent to the user.
- I changed the invariants to explicitly apply to step 8 of the algorithm. Step 8 is the implementation-specific step that does the actual sorting. This brings Array#sort more in line with other builtins. Now Array#sort is specced by a given explicit algorithm and the steps that are implementation-specific are somewhat mathematically nailed down.

waldemarhorwat:
The wording is still a bit broken:

> There must be some mathematical permutation π of the nonnegative integers less than itemCount, such that for every nonnegative integer j less than itemCount, if property old[j] existed, then new[π(j)] is exactly the same value as old[j]. But if property old[j] did not exist, then new[π(j)] does not exist.

The wording of this does not make sense given the changed definition of old and new:

> Here the notation old[j] is used to refer to the hypothetical result of items[j] before step 8 is executed, and the notation new[j] to refer to the hypothetical result of items[j] after step 8 has been executed.

A List has elements, not properties. Also, what does it mean for a property of a List to not exist?

There are a couple other things that are now anachronisms:

> Calling comparefn(a, b) does not modify obj or any object on obj's prototype chain.

Do we still need that condition now that we're using a hidden List? It's fine if it stays, but I'm not sure it's necessary any more.

> Furthermore, Type(v) is Number, and v is not NaN.

Here *v* is the result of *comparefn*. In that case what is the purpose of the ToNumber call inside SortCompare? Should we make *v* be the result of ToNumber(*comparefn*(...)) here?

szuend:
Changed the wording a bit. This is not my strong suit so any suggestion on how to rephrase that Phi is simply a permutation are welcome.

I am also a bit unsure about the second argument:

> Then for all nonnegative integers _j_ and _k_, each less than _itemCount_, if
> SortCompare(old[_j_], old[_k_]) < 0 (see SortCompare below), then new[pi(_j_)] < new[pi(_k_)]

new[pi(j)] is an object, but we use "smaller" here to compare it instead of SortCompare. Shouldn't be the condition here that the index of the resulting permutation is smaller, if elements compare smaller? That is, shouldnt be the second part be pi(_j_) < pi(_k_)?

> Calling comparefn(a, b) does not modify obj or any object on obj's prototype chain.
I would still leave this as-is. The number of comparefn calls is still implementation-specific, so side-effecting compare functions will not behave consistently across engines. No strong preference here, so I am also happy to remove this.

> Furthermore, Type(v) is Number, and v is not NaN.
This part is still the same as the current version of the spec. I would assume that this text is here to define what constitutes a consistent comparison function, that is, the comparison function should return a number consistent with <, > and =.

But since its not guaranteed that actual user code returns a number, implementations have to first use ToNumber on the result of calling comparefn.

One last note: I think the conditions for a consistent comparison function are missing anti-symmetry for smaller and greater, that is, if `a <cf b` then `b <cf a` must be false.

waldemarhorwat:
> One last note: I think the conditions for a consistent comparison function are missing anti-symmetry for smaller and greater, that is, if `a <cf b` holds then `b <cf a` must not hold.

It's not missing. You can derive anti-symmetry from the existing axioms.

Suppose you have `a <cf b`. What can the comparison of `b` to `a` be?

We know that exactly one of the three relations `b <cf a`, `b =cf a`, or `b >cf a` must be true.

- Suppose `b <cf a` holds. Then, by transitivity of `a <cf b` and `b <cf a` we have `a <cf a`, which contradicts reflexivity. So `b <cf a` cannot be true.

- Suppose `b =cf a` holds. Then by symmetry `a =cf b`, which contradicts `a <cf b`. So `b =cf a` cannot be true.

Thus `b >cf a` must be true. QED.

szuend:
Thanks for the proof!

What else does this PR need to address or fix for it to go forward?

spec.html
```diff
@@ -34164,77 +34164,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            There must be some mathematical permutation &pi; of the nonnegative integers less than _itemCount_, such that for every nonnegative integer _j_ less than _itemCount_, the element <emu-eqn>old[_j_]</emu-eqn> is exactly the same as <emu-eqn>new[&pi;(_j_)]</emu-eqn>.
           </li>
           <li>
-            Then for all nonnegative integers _j_ and _k_, each less than _len_, if <emu-eqn>SortCompare(old[_j_], old[_k_]) &lt; 0</emu-eqn> (see SortCompare below), then <emu-eqn>new[&pi;(_j_)] &lt; new[&pi;(_k_)]</emu-eqn>.
+            Then for all nonnegative integers _j_ and _k_, each less than _itemCount_, if <emu-eqn>SortCompare(old[_j_], old[_k_]) &lt; 0</emu-eqn> (see SortCompare below), then <emu-eqn>new[&pi;(_j_)] &lt; new[&pi;(_k_)]</emu-eqn>.
```
claudepache: There is an old bug here, see https://github.com/tc39/ecma262/issues/1369#issuecomment-446678903. It must not be “... then new[&pi;(_j_)] &lt; new[&pi;(_k_)]”, but “... then &pi;(_j_) &lt; &pi;(_k_)”.

spec.html
```diff
@@ -34164,77 +34164,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
           </li>
         </ul>
-        <p>Here the notation <emu-eqn>old[_j_]</emu-eqn> is used to refer to the hypothetical result of calling <emu-eqn>Get(_obj_, _j_)</emu-eqn> before this function is executed, and the notation <emu-eqn>new[_j_]</emu-eqn> to refer to the hypothetical result of calling <emu-eqn>Get(_obj_, _j_)</emu-eqn> after this function has been executed.</p>
+        <p>Here the notation <emu-eqn>old[_j_]</emu-eqn> is used to refer to the hypothetical result of <emu-eqn>_items_[_j_]</emu-eqn> before step 8 is executed, and the notation <emu-eqn>new[_j_]</emu-eqn> to refer to the hypothetical result of <emu-eqn>_items_[_j_]</emu-eqn> after step 8 has been executed.</p>
```
claudepache: Those are no longer “hypothetical result[s] of”, because the value on which it is queried are no longer array-like ECMAScript values, but Lists, for which _items_[_j_], as [defined here](https://tc39.es/ecma262/#sec-list-and-record-specification-type), is not a “result” of some potentially destructive operation.

Just delete the two occurrences of “the hypothetical result of”.

spec.html
```diff
@@ -34164,77 +34164,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
-          </li>
-          <li>
-            0 &le; _j_ &lt; _len_
-          </li>
-          <li>
-            HasProperty(_proto_, ToString(_j_)) is *true*.
-          </li>
-        </ul>
-        <p>The sort order is also implementation-defined if _obj_ is sparse and any of the following conditions are true:</p>
+        <p>The sort order is implementation-defined if any of the following conditions are true:</p>
```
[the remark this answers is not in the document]
szuend: Done

spec.html
```diff
@@ -34164,77 +34164,55 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. Set _k_ to _k_ + 1.
+          1. Let _itemCount_ be the length of _items_.
+          1. Sort _items_ using an implementation-dependent sequence of calls to SortCompare.
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _itemCount_
+            1. Perform ? Set(_obj_, ! ToString(_j_), items[_j_], *true*).
+            1. Set _j_ to _j_ + 1.
+          1. Repeat, while _j_ &lt; _len_
+            1. Let _Pj_ be ! ToString(_j_).
+            1. Perform ? DeletePropertyOrThrow(_obj_, _Pj_).
```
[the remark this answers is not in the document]
szuend: Done

spec.html
szuend: Done

Thanks for linking the bug! I already suspected as much, [see comment in this PR](https://github.com/tc39/ecma262/pull/1585#issuecomment-582798837)

spec.html
szuend: Thanks and done!

claudepache:
Although the spec requires a well-defined behaviour when _comparefn_ is a consistent comparison function (when provided), what we really care here, is whether SortCompare is a consistent comparison function.

My guess is that the consistency requirement was made on _comparefn_ rather than SortCompare in order to state that the user is supposed to provide a well-behaving _comparefn_ if they want Array.prototype.sort to work.

But I think we can made the spec both simpler and tighter by just saying:

> The sort order is implementation-defined if SortCompare is(*) not a consistent comparison function for the elements of _items_.

and following it by a Note giving some sufficient conditions that make SortCompare consistent:

* If _comparefn_ is a consistent comparison function for _S_ (which implies by definition: returns non-NaN Numbers), then SortCompare is(*) a consistent comparison function for _S_.
* If _comparefn_ is undefined and if all applications of ToString to any specific element of _S_ produce the same result and are side-effect free, then SortCompare is(*) a consistent comparison function for _S_.

Moreover, the definition of “consistent comparison function” would benefit to be isolated in its own subsection (and _comparefn_ should be renamed in that definition in order to avoid confusion). 

--- 
(*) or rather, to be pedantic, “acts as” or “behaves as”, since SortCompare is an abstract operation, not a function object. Alternatively, redefine “consistent comparison function” to be either a function object, or an abstract operation taking two arguments.

szuend:
@waldemarhorwat I like the above suggestions, what do you think?

waldemarhorwat:
The wording looks correct as it is now. If you'd prefer to refactor it a bit more by making the consistency requirement on SortCompare instead of *comparefn* as per the above comment, that would be fine too, but that's optional.

spec.html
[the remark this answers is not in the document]
devsnek: we're talking about multiple conditions here, shouldn't it be "are"?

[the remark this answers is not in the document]
ljharb: No, it’s “if any is true”, the plurality of the predicate is irrelevant here.

szuend:
Ok, I would land it as-is for now and do the refactoring as a follow up.

mathiasbynens:
@kmiller68 Did you ever get around to [investigating](https://github.com/tc39/ecma262/pull/1585#issuecomment-538396456) whether these changes are feasible for JSC? Once we get your input, we can add Test262 tests, at which point I think this PR is good to go.

kmiller68:
> @kmiller68 Did you ever get around to [investigating](https://github.com/tc39/ecma262/pull/1585#issuecomment-538396456) whether these changes are feasible for JSC? Once we get your input, we can add Test262 tests, at which point I think this PR is good to go.

I haven't, I'll add it near the top of my todo list. Unfortunately, I'm in bug hell from a recent performance feature I added...

mathiasbynens:
@kmiller68 Any updates on this?

mathiasbynens:
Since @kmiller68 now has a patch over at https://bugs.webkit.org/show_bug.cgi?id=202582, I'm assuming that means it is feasible to implement in JavaScriptCore after all. 👍

kmiller68:
> Since @kmiller68 now has a patch over at https://bugs.webkit.org/show_bug.cgi?id=202582, I'm assuming that means it is feasible to implement in JavaScriptCore after all. 👍

Oh, I just put up a work in progress for someone else to look at / build off. That path definitely doesn't work.

mathiasbynens:
> Oh, I just put up a work in progress for someone else to look at / build off. That patch definitely doesn't work.

@kmiller68 Now that the WebKit bug is RESOLVED FIXED (🎉 ), can you officially confirm this is feasible to implement in JavaScriptCore?

In that case the only blocker for landing this patch is writing some Test262 tests, for which we could re-use the test cases posted earlier in this thread.

mathiasbynens:
Test262 PR: https://github.com/tc39/test262/pull/2939 Passing in V8, JSC, and SpiderMonkey.

@kmiller68 Could you PTAL at https://github.com/tc39/ecma262/pull/1585#issuecomment-739895523 and confirm? That way we can remove the “needs data” label.

kmiller68:
Sorry, I missed the references here. I wish GH had a better @ notification system... 

Anyway, yeah, this is fine from JSC side since we shipped it in the latest Safari (I think).

spec.html
```diff
@@ -34164,77 +34164,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
-            * The arguments for calls to SortCompare are values returned by a previous call to the Get abstract operation, unless the properties accessed by those previous calls did not exist according to HasOwnProperty. If both prospective arguments to SortCompare correspond to non-existent properties, use *+0* instead of calling SortCompare. If only the first prospective argument is non-existent use +1. If only the second prospective argument is non-existent use -1.
-            * If _obj_ is not sparse then DeletePropertyOrThrow must not be called.
-            * If an abrupt completion is returned from any of these operations, it is immediately returned as the value of this function.
-          1. Return _obj_.
-        </emu-alg>
-        <p>Unless the sort order is specified above to be implementation-defined, the returned object must have the following two characteristics:</p>
+        <p>Unless the sort order is specified above to be implementation-defined, _items_ must have the following two characteristics after executing step 8 of the algorithm above:</p>
```
bakkot, line 34205: Editorial nit: rather than giving the explicit number of the step, you can label the step itself:

```
1. [id="step-array-sort"] Sort _items_ using an implementation-dependent sequence of calls to SortCompare.
```
and then reference it:
```
[...] after executing step <emu-xref href="#step-array-sort"></emu-xref> of the algorithm above [...]
```

This ensures the number doesn't get stale.

spec.html
```diff
@@ -34289,7 +34266,7 @@ <h1>Runtime Semantics: SortCompare ( _x_, _y_ )</h1>
             <p>Because non-existent property values always compare greater than *undefined* property values, and *undefined* always compares greater than any other value, *undefined* property values always sort to the end of the result, followed by non-existent property values.</p>
           </emu-note>
           <emu-note>
-            <p>Method calls performed by the ToString abstract operations in steps 5 and 7 have the potential to cause SortCompare to not behave as a consistent comparison function.</p>
+            <p>Method calls performed by the ToString abstract operations in steps 5 and 6 have the potential to cause SortCompare to not behave as a consistent comparison function.</p>
```
ljharb, line 34269: (step numbers here too)

spec.html
szuend: Done

spec.html
szuend: This was addressed upstream and resolved during rebase.

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
-        <p>Within this specification of the `sort` method, an object, _obj_, is said to be <em>sparse</em> if the following algorithm returns *true*:</p>
-        <emu-alg>
-          1. For each integer _i_ in the range 0 &le; _i_ &lt; _len_, do
-            1. Let _prop_ be ! ToString(𝔽(_i_)).
-            1. Let _elem_ be _obj_.[[GetOwnProperty]](_prop_).
-            1. If _elem_ is *undefined*, return *true*.
-          1. Return *false*.
+          1. Let _items_ be a new empty List.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_
```
bakkot, line 32536:
```suggestion
          1. Repeat, while _k_ &lt; _len_,
```

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. Let _Pk_ be ! ToString(_k_).
```
bakkot, line 32537:
```suggestion
            1. Let _Pk_ be ! ToString(𝔽(_k_)).
```

since `_k_` is a mathematical integer, and `ToString` expects a JavaScript Number.

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. Let _kPresent_ be ? HasProperty(_obj_, _Pk_).
+            1. If _kPresent_ is *true*, then
+              1. Let _kValue_ be ? Get(_obj_, _Pk_).
+              1. Append _kValue_ to _items_.
+            1. Set _k_ to _k_ + 1.
+          1. Let _itemCount_ be the length of _items_.
+          1. [id="step-array-sort"] Sort _items_ using an implementation-dependent sequence of calls to SortCompare.
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _itemCount_
```
bakkot, line 32546:
```suggestion
          1. Repeat, while _j_ &lt; _itemCount_,
```

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. Perform ? Set(_obj_, ! ToString(_j_), items[_j_], *true*).
```
bakkot, line 32547:
```suggestion
            1. Perform ? Set(_obj_, ! ToString(𝔽(_j_)), items[_j_], *true*).
```

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. Set _j_ to _j_ + 1.
+          1. Repeat, while _j_ &lt; _len_
```
bakkot, line 32549:
```suggestion
          1. Repeat, while _j_ &lt; _len_,
```

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. Perform ? DeletePropertyOrThrow(_obj_, ! ToString(_j_)).
```
bakkot, line 32550:
```suggestion
            1. Perform ? DeletePropertyOrThrow(_obj_, ! ToString(𝔽(_j_))).
```

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_
+            1. Let _Pk_ be ! ToString(_k_).
+            1. Let _kPresent_ be ? HasProperty(_obj_, _Pk_).
+            1. If _kPresent_ is *true*, then
+              1. Let _kValue_ be ? Get(_obj_, _Pk_).
+              1. Append _kValue_ to _items_.
+            1. Set _k_ to _k_ + 1.
+          1. Let _itemCount_ be the length of _items_.
+          1. [id="step-array-sort"] Sort _items_ using an implementation-dependent sequence of calls to SortCompare.
```
bakkot, line 32544: A nit: `SortCompare` can throw, and that should be called out here, as in

```suggestion
          1. [id="step-array-sort"] Sort _items_ using an implementation-dependent sequence of calls to SortCompare. If any such call returns an abrupt completion, stop before performing any further calls to SortCompare or steps in this algorithm and return that completion.
```

(The above wording matches my understanding of the intent. If it's not what is intended we should of course find another wording.)

bakkot (commented):
A few editorial nits (some of which the linter is complaining about; fixing these should make it pass).

Also, as @jmdyck mentions [above](https://github.com/tc39/ecma262/pull/1585#pullrequestreview-258901783), [%TypedArray%.prototype.sort](https://tc39.es/ecma262/#sec-%typedarray%.prototype.sort) is defined by reference to Array.prototype.sort, and hence needs to be tweaked.

In particular
- it still refers to "the entry steps in 23.1.3.27:", and it's no longer obvious what those are; instead I think we should label steps 1-3 in A.p.sort and refer to "steps 1-3" explicitly (except with xrefs in place of the numbers, of course).
- it refers to "The implementation-defined sort order condition for exotic objects", which has been dropped (and can hence that sentence just be dropped from TA.p.sort, I think).

Looks good to me other than these comments.

spec.html
syg: implementation-dependent -> implementation-defined.

+1 on the extra wording otherwise.

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
-            * The arguments for calls to SortCompare are values returned by a previous call to the Get abstract operation, unless the properties accessed by those previous calls did not exist according to HasOwnProperty. If both prospective arguments to SortCompare correspond to non-existent properties, use *+0*<sub>𝔽</sub> instead of calling SortCompare. If only the first prospective argument is non-existent, use *1*<sub>𝔽</sub>. If only the second prospective argument is non-existent, use *-1*<sub>𝔽</sub>.
-            * If _obj_ is not sparse then DeletePropertyOrThrow must not be called.
-            * If an abrupt completion is returned from any of these operations, it is immediately returned as the value of this function.
-          1. Return _obj_.
-        </emu-alg>
-        <p>Unless the sort order is specified above to be implementation-defined, the returned object must have the following two characteristics:</p>
+        <p>Unless the sort order is specified above to be implementation-defined, _items_ must have the following two characteristics after executing step <emu-xref href="#step-array-sort"></emu-xref> of the algorithm above:</p>
```
syg, line 32567:
```suggestion
        <p>Unless the sort order is specified above to be implementation-defined, _items_ must satisfy all of the following conditions after executing step <emu-xref href="#step-array-sort"></emu-xref> of the algorithm above:</p>
```

syg (approved):
lgtm with nits

mathiasbynens:
Thanks for the thorough reviews! I’ve pushed some changes addressing all the feedback so far.

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _items_ be a new empty List.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Let _kPresent_ be ? HasProperty(_obj_, _Pk_).
+            1. If _kPresent_ is *true*, then
+              1. Let _kValue_ be ? Get(_obj_, _Pk_).
+              1. Append _kValue_ to _items_.
+            1. Set _k_ to _k_ + 1.
+          1. Let _itemCount_ be the length of _items_.
```
jmdyck, line 32543:
```suggestion
          1. Let _itemCount_ be the number of elements in _items_.
```
Generally, when an algorithm refers to "the length of X", X is a String value.

spec.html
```diff
@@ -32526,78 +32526,54 @@ <h1>Array.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. [id="step-array-sort"] Sort _items_ using an implementation-defined sequence of calls to SortCompare. If any such call returns an abrupt completion, stop before performing any further calls to SortCompare or steps in this algorithm and return that completion.
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _itemCount_,
+            1. Perform ? Set(_obj_, ! ToString(𝔽(_j_)), items[_j_], *true*).
```
jmdyck, line 32547:
```suggestion
            1. Perform ? Set(_obj_, ! ToString(𝔽(_j_)), _items_[_j_], *true*).
```

jmdyck (changes requested):

bakkot (approved):

jmdyck (approved):
Looks okay to me.

ljharb (approved):

spec.html
```diff
@@ -33717,14 +33693,13 @@ <h1>%TypedArray%.prototype.some ( _callbackfn_ [ , _thisArg_ ] )</h1>
         <h1>%TypedArray%.prototype.sort ( _comparefn_ )</h1>
         <p>%TypedArray%`.prototype.sort` is a distinct function that, except as described below, implements the same requirements as those of `Array.prototype.sort` as defined in <emu-xref href="#sec-array.prototype.sort"></emu-xref>. The implementation of the %TypedArray%`.prototype.sort` specification may be optimized with the knowledge that the *this* value is an object that has a fixed length and whose <emu-xref href="#integer-index">integer-indexed</emu-xref> properties are not sparse. The only internal methods of the *this* value that the algorithm may call are [[Get]] and [[Set]].</p>
```
devsnek, line 33694: > The only internal methods of the *this* value that the algorithm may call are [[Get]] and [[Set]].

Isn't this violated by these changes?

bakkot: The hole deletion loop doesn't get hit for TypedArrays, as far as I can tell, so we shouldn't have to worry about that part.

bakkot: I think we should remove the

> The only internal methods of the *this* value that the algorithm may call are [[Get]] and [[Set]].

bit, since that was only relevant when implementations had considerably more freedom for how to do the sort. Now that the reading and writing parts are fully specified it's not relevant.

devsnek: Whatever way it goes, it would still be nice to use an abstract op instead of magic step replacement.

mathiasbynens: I've made @bakkot's suggested change. W.r.t. using an abstract operation, I agree, but since it's orthogonal to this PR I'd prefer to tackle it separately.

michaelficarra (approved):

bakkot:
Thanks @szuend and @mathiasbynens for making this happen!

PR #1602, opened 2019-06-25
Editorial: use `TrimString` in `parseFloat` and `parseInt`

ljharb:
Saw an opportunity for some reusability. This should be observably identical.


jmdyck:
TrimString will treat the input string as UTF-16 encoded code points, so this PR should be fine as long as parseFloat and parseInt are supposed to do that too. Presumably they do, though.they don't say so explicitly.

PR #1603, opened 2019-06-25
Editorial: use "be the *this* value" over "be *this* value"

ljharb:
130 occurrences vs 4.


ljharb:
@jmdyck fair fallout :-) i specifically included “be” in the grep to ensure i skipped over those occurrences (which read fine to me)

PR #1607, opened 2019-07-02
Editorial: Add ! and ? before CreateBuiltinFunction and CreateArrayFromList

chicoxyzzy:


chicoxyzzy:
I should revisit some changes in this commit

spec.html
```diff
@@ -8252,7 +8252,7 @@ <h1>MakeArgGetter ( _name_, _env_ )</h1>
           <p>The abstract operation MakeArgGetter called with String _name_ and Environment Record _env_ creates a built-in function object that when executed returns the value bound for _name_ in _env_. It performs the following steps:</p>
           <emu-alg>
             1. Let _steps_ be the steps of an ArgGetter function as specified below.
-            1. Let _getter_ be CreateBuiltinFunction(_steps_, &laquo; [[Name]], [[Env]] &raquo;).
+            1. Let _getter_ be ! CreateBuiltinFunction(_steps_, &laquo; [[Name]], [[Env]] &raquo;).
```
ljharb: GetBindingValue can throw a ReferenceError
```suggestion
            1. Let _getter_ be ? CreateBuiltinFunction(_steps_, &laquo; [[Name]], [[Env]] &raquo;).
```

spec.html
```diff
@@ -38330,7 +38330,7 @@ <h1>Runtime Semantics: PerformPromiseAll ( _iteratorRecord_, _constructor_, _res
               1. Append *undefined* to _values_.
               1. Let _nextPromise_ be ? Call(_promiseResolve_, _constructor_, &laquo; _nextValue_ &raquo;).
               1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-promise.all-resolve-element-functions" title></emu-xref>.
-              1. Let _resolveElement_ be CreateBuiltinFunction(_steps_, &laquo; [[AlreadyCalled]], [[Index]], [[Values]], [[Capability]], [[RemainingElements]] &raquo;).
+              1. Let _resolveElement_ be ! CreateBuiltinFunction(_steps_, &laquo; [[AlreadyCalled]], [[Index]], [[Values]], [[Capability]], [[RemainingElements]] &raquo;).
```
ljharb: This can throw, because it’s calling user code in step 11b of https://tc39.es/ecma262/#sec-promise.all-resolve-element-functions

domenic: I don't believe this is correct. CreateBuiltinFunction just creates a function and assigns those steps to it. It doesn't execute the function it creates.

ljharb: ah, that’s true - in that case, it seems like every call to CreateBuiltinFunction should have `!`?

spec.html
```diff
@@ -32997,7 +32997,7 @@ <h1>%ArrayIteratorPrototype%.next ( )</h1>
             1. If _itemKind_ is `"value"`, let _result_ be _elementValue_.
             1. Else,
               1. Assert: _itemKind_ is `"key+value"`.
-              1. Let _result_ be CreateArrayFromList(&laquo; _index_, _elementValue_ &raquo;).
+              1. Let _result_ be ! CreateArrayFromList(&laquo; _index_, _elementValue_ &raquo;).
```
leobalter: this is misleading for me. When I read the specs for tests I use the `!` as a hint for: "this abstraction could possibly return an abrupt completion, but won't return any for the given arguments". In this same case,  `CreateArrayFromList` won't ever return any abrupt completion, therefore, the `!` is unnecessary and requires me extra time trying to figure out where it could possibly return an abrupt completion until I realize it won't.

This happens every single time I read the specs.

I'm curious what would be the advantage of having the `!`. I believe the specs are way more consistent when they are meaningful.

spec.html
```diff
@@ -37148,7 +37148,7 @@ <h1>AsyncFromSyncIteratorContinuation ( _result_, _promiseCapability_ )</h1>
           1. Let _valueWrapper_ be PromiseResolve(%Promise%, _value_).
           1. IfAbruptRejectPromise(_valueWrapper_, _promiseCapability_).
           1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-async-from-sync-iterator-value-unwrap-functions" title></emu-xref>.
-          1. Let _onFulfilled_ be CreateBuiltinFunction(_steps_, &laquo; [[Done]] &raquo;).
+          1. Let _onFulfilled_ be ! CreateBuiltinFunction(_steps_, &laquo; [[Done]] &raquo;).
```
leobalter: same thing here. `!` is unnecessary and becomes less meaningful to where it actually indicates effect.

spec.html
chicoxyzzy: According to @domenic's [comment in `Promise.allSettled` repo](https://github.com/tc39/proposal-promise-allSettled/issues/32#issue-417895560), `!` or `?` should be used before every abstract operation except if its result is completion value.

spec.html
chicoxyzzy: In this particular case it can fail on `Promise.all` Resolve Element Function

spec.html
chicoxyzzy: I think that you suggestion is valid. I can't commit your change it while conversation is marked as resolved though

spec.html
ljharb: We should be using it every time; every abstract operation could possibly return an abrupt completion, when viewed from the call site.

spec.html
chicoxyzzy: For some reason I can't commit it even when conversation is marked as unresolved. It says that diff has been updated, but it's actually not  ¯\_(ツ)_/¯ 
image

spec.html
chicoxyzzy: It can fail because of previous step (Promise.all Resolve Element Functions in this case)

spec.html
leobalter: that's really unhelpful and I can't rediscuss everything again. I already had mentioned that in another issue but this is circling back.

I can't block anything, but assuming we know our abstractions, using a flag - `!` - only when it's necessary makes it more informative. The referred abstractions are not coming from unknown sources, they are specified in the same document.

Using `!` for everything only makes it vague and redundant. We can be more assertive than that, we already are, no need for changes.

ljharb: We should probably move that discussion to #486; at the moment, every single abstract operation call in the spec should either have its completion record return inspected, or should have a ! or a ? (there’s of plenty of places left to fix, but we should fix them whenever they come up)

leobalter: There is nothing else I can add to the discussion here or there. I can't mandate what individually want and I don't think I have more arguments (as in facts) to convince anyone to change their idea.

It is what it is. At least this one is only editorial.

spec.html
ljharb: Those steps won’t be invoked here tho, per https://github.com/tc39/ecma262/pull/1607#discussion_r299503987 - just creating the function doesn’t seem to ever fail.

chicoxyzzy: How to indicate that 
```
Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-promise.all-resolve-element-functions" title></emu-xref>.
```
may fail?

domenic: That can never fail. That's just letting _steps_ have a certain value.

chicoxyzzy: I've replaced all `?` usages by `!` for `CreateBuiltinFunction`

PR #1608, opened 2019-07-02
Normative: IsSimpleParameterList is false for standalone rest parameters

gibson042:
I noticed in passing that this case was missing.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19205,6 +19205,10 @@
       <emu-alg>
         1. Return *true*.
       </emu-alg>
+      <emu-grammar>FormalParameters : FunctionRestParameter</emu-grammar>
+      <emu-alg>
+        1. Return *false*.
+      </emu-alg>
       <emu-grammar>FormalParameters : FormalParameterList `,` FunctionRestParameter</emu-grammar>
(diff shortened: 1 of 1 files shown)


ljharb:
hm, need this be non simple tho? iow does it create the strict/sloppy parsing difficulties that led to this check?

devsnek:
all the engines throw a syntax error for it (except xs, known bug) :man_shrugging:

ljharb:
Technically i think this is normative.

gibson042:
I wondered about that, but settled on editorial with the presumption that the behavior was understood to be expected but just hadn't been correctly represented in spec text.

jmdyck:
I plugged this hole in [this commit](https://github.com/tc39/ecma262/pull/1301/commits/db3a3b1486b8cdfcf38a595df701683a25156d95) of PR #1301.

ljharb:
I'd consider it a normative spec bug; imo editorial is exclusively for things that aren't observable.

anba:
FWIW this is a regression from 6a06509a7bc1749766f324f47ee5c4baff7bed27

PR #1228, opened 2018-06-15
Editorial: (Exponentiation) Add NaN to exponent values whose behavior deviates from IEEE 754-2008

rwaldron:
This oversight was discovered and reported to me by James Clark (jjc@jclark.com).

PR #1609, opened 2019-07-03
Editorial: Add NaN to exponent values whose behavior deviates from IEEE 754-2008

Ashbringer7077:
…or deviates from IEEE 754-2008

This oversight was discovered and reported to me by James Clark (jjc@jclark.com).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13615,7 +13615,7 @@
         <li>If _base_ &lt; 0 and _base_ is finite and _exponent_ is finite and _exponent_ is not an integer, the result is *NaN*.</li>
       </ul>
       <emu-note>
-        <p>The result of _base_ `**` _exponent_ when _base_ is *1* or *-1* and _exponent_ is *+Infinity* or *-Infinity* differs from IEEE 754-2008. The first edition of ECMAScript specified a result of *NaN* for this operation, whereas later versions of IEEE 754-2008 specified *1*. The historical ECMAScript behaviour is preserved for compatibility reasons.</p>
+        <p>The result of _base_ `**` _exponent_ when _base_ is *1* or *-1* and _exponent_ is *+Infinity*, *-Infinity* or *NaN* differs from IEEE 754-2008. The first edition of ECMAScript specified a result of *NaN* for this operation, whereas later versions of IEEE 754-2008 specified *1*. The historical ECMAScript behaviour is preserved for compatibility reasons.</p>
       </emu-note>
     </emu-clause>
   </emu-clause>


spec.html
```diff
@@ -13615,7 +13615,7 @@ <h1>Applying the `**` Operator</h1>
         <li>If _base_ &lt; 0 and _base_ is finite and _exponent_ is finite and _exponent_ is not an integer, the result is *NaN*.</li>
       </ul>
       <emu-note>
-        <p>The result of _base_ `**` _exponent_ when _base_ is *1* or *-1* and _exponent_ is *+Infinity* or *-Infinity* differs from IEEE 754-2008. The first edition of ECMAScript specified a result of *NaN* for this operation, whereas later versions of IEEE 754-2008 specified *1*. The historical ECMAScript behaviour is preserved for compatibility reasons.</p>
+        <p>The result of _base_ `**` _exponent_ when _base_ is *1* or *-1* and _exponent_ is *+Infinity*, *-Infinity* or *NaN* differs from IEEE 754-2008. The first edition of ECMAScript specified a result of *NaN* for this operation, whereas later versions of IEEE 754-2008 specified *1*. The historical ECMAScript behaviour is preserved for compatibility reasons.</p>
```
```suggestion
        <p>The result of _base_ `**` _exponent_ when _base_ is *1* or *-1* and _exponent_ is *+Infinity*, *-Infinity*, or *NaN* differs from IEEE 754-2008. The first edition of ECMAScript specified a result of *NaN* for this operation, whereas later versions of IEEE 754-2008 specified *1*. The historical ECMAScript behavior is preserved for compatibility reasons.</p>
```

leobalter: the comma in the suggestion is fine, but spec should use Oxford English, so `behaviour` should not be changed.

ljharb: ah, fair enough - there's 125 current uses of the "u" and 1 use without. I'll update the suggestion.

waldemarhorwat:
The new wording is incorrect. IEEE 754-2008 does **not** specify that *-1* ** *NaN* is *1*. The result of that exponentiation is still *NaN* in IEEE 754-2008, which matches what ECMAScript is doing.

apaprocki:
> The new wording is incorrect. IEEE 754-2008 does **not** specify that _-1_ ** _NaN_ is _1_.

Agreed. If my reading of Section 9.2.1 is correct, this would break out the two cases:

```
The result of _base_ `**` _exponent_ 
when _base_ is *1* and _exponent_ is *+Infinity*, *-Infinity* or *NaN*; or
when _base_ is *-1* and _exponent_ is *+Infinity* or *-Infinity* differs ...
```

I'm not sure if there is a cleaner way to write those two conditions with inline lists in one phrase.

chicoxyzzy:
This PR is the the same as #1228 and uses commit from it

ljharb:
Closing as a duplicate of #1228.

PR #1613, opened 2019-07-07
Editorial: Some small formatting tweaks

TimothyGu:
- Make dashes consistent; certain instances in the title of standards are not changed
- Bold the Number value +0

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -120,7 +120,7 @@
 <emu-clause id="sec-normative-references">
   <h1>Normative References</h1>
   <p>The following referenced documents are indispensable for the application of this document. For dated references, only the edition cited applies. For undated references, the latest edition of the referenced document (including any amendments) applies.</p>
-  <p>ISO/IEC 10646 <i>Information Technology &ndash; Universal Multiple-Octet Coded Character Set (UCS) plus Amendment 1:2005, Amendment 2:2006, Amendment 3:2008, and Amendment 4:2008</i>, plus additional amendments and corrigenda, or successor</p>
+  <p>ISO/IEC 10646 <i>Information Technology &mdash; Universal Multiple-Octet Coded Character Set (UCS) plus Amendment 1:2005, Amendment 2:2006, Amendment 3:2008, and Amendment 4:2008</i>, plus additional amendments and corrigenda, or successor</p>
   <p>ECMA-402, <i>ECMAScript 2015 Internationalization API Specification</i>.
     <br>
     <a href="https://ecma-international.org/publications/standards/Ecma-402.htm">https://ecma-international.org/publications/standards/Ecma-402.htm</a></p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -872,10 +872,10 @@ <h1>Mathematical Operations</h1>
       </ul>
 
       <p>In the language of this specification, numerical values and operations (including addition, subtraction, negation, multiplication, division, and comparison) are distinguished among different numeric kinds using subscripts. The subscript <sub><dfn id="𝔽">𝔽</dfn></sub> refers to Numbers, and the subscript <sub><dfn id="ℝ">ℝ</dfn></sub> refers to mathematical values. A subscript is used following each numeric value and operation.</p>
-      <p>For brevity, the <sub>𝔽</sub> subscript can be omitted on Number values--a numeric value with no subscript is interpreted to be a Number. An operation with no subscript is interpreted to be a Number operation, unless one of the parameters has a particular subscript, in which case the operation adopts that subscript. For example, 1<sub>ℝ</sub> + 2<sub>ℝ</sub> = 3<sub>ℝ</sub> is a statement about mathematical values, and 1 + 2 = 3 is a statement about Numbers.</p>
```
ljharb: i have a light preference to include the spaces around em dashes; it makes the raw ecmarkup more readable,

TimothyGu: [Most style guides](https://www.quickanddirtytips.com/education/grammar/em-dash-space) do not include spaces around em-dashes, and the spec currently does not do so either (despite @jmdyck's comment).

TimothyGu:
> * Changing double-hyphen to `&mdash;` is fine with me. Note that [GetWaiterList](https://tc39.es/ecma262/#sec-getwaiterlist) has a couple more of these.

Fixed.

> * The spec also has 3 occurrences of `&ndash;` that should maybe be changed to `&mdash;`.

As I noted above, I did not change occurrences of `&ndash;` in international standard names. However, after checking the actual standard PDFs, they seem to use `&mdash;` with spaces around them, so I did just that.

> * Spaces around dash would be my preference too. Currently, the spec is divided on usage: if you include double-hyphen, mdash, and ndash, there's 8 with spaces vs 7 without.

Out of the 9 uses of em-dashes, 6 have spaces and 1 is in the name of a standard, which as I noted above did not change. En-dashes generally have a different style, so that's not the most relevant comparison.

spec.html
ljharb: Sure, but many style guides also recommend omitting the Oxford comma, or avoiding singular “they” - we can certainly make our own choice here. Either way, for this PR, we should follow whatever pattern is most used for em dashes - which below comments imply is “with spaces”.

jmdyck:
> > * The spec also has 3 occurrences of `&ndash;` that should maybe be changed to `&mdash;`.
> 
> As I noted above, I did not change occurrences of `&ndash;` in international standard names.

Ah, so you did, sorry I missed that.

> However, after checking the actual standard PDFs, they seem to use `&mdash;` with spaces around them, so I did just that.

If ISO were going to explicitly say "use em dash in standard names", I'd expect it to appear [here](https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor137) in their directives. They get very close, down to the level of the elements of a title, but don't say how to separate them. However, all the example titles use U+2014 EM DASH in the HTML source. So changing our 3 occurrences of `&ndash;` to `&mdash;` seems justified.

TimothyGu:
@jmdyck so does this PR LGTM?

jmdyck:
@TimothyGu: I agree with all changes in this PR, apart from the removal of spaces around em-dashes, on which I will accept the decision of the editors.

PR #1614, opened 2019-07-08
Editorial: Tweak ValidateAndApplyPropertyDescriptor

shvaikalesh:


ljharb:
Can you provide more context on why you feel this is an improvement?

spec.html
```diff
@@ -1278,8 +1278,8 @@ <h1>Property Attributes</h1>
             </tbody>
           </table>
         </emu-table>
-        <p>If the initial values of a property's attributes are not explicitly specified by this specification, the default value defined in <emu-xref href="#table-4"></emu-xref> is used.</p>
-        <emu-table id="table-4" caption="Default Attribute Values">
+        <p>If the initial values of a property's attributes are not explicitly specified by this specification, the default value defined in <emu-xref href="#table-default-attribute-values"></emu-xref> is used.</p>
+        <emu-table id="table-default-attribute-values" caption="Default Attribute Values">
```
shvaikalesh: So we can link here "default value(s)" references in `ValidateAndApplyPropertyDescriptor` (and vice versa).

spec.html
```diff
@@ -6949,22 +6949,23 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
...
-              1. If _O_ is not *undefined*, create an own accessor property named _P_ of object _O_ whose [[Get]], [[Set]], [[Enumerable]] and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its default value.
+              1. If _O_ is not *undefined*, create an own accessor property named _P_ of object _O_ whose [[Get]], [[Set]], [[Enumerable]] and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its <emu-xref href="#table-default-attribute-values">default value</emu-xref>.
             1. Return *true*.
           1. If every field in _Desc_ is absent, return *true*.
           1. If _current_.[[Configurable]] is *false*, then
             1. If _Desc_.[[Configurable]] is present and its value is *true*, return *false*.
-            1. If _Desc_.[[Enumerable]] is present and the [[Enumerable]] fields of _current_ and _Desc_ are the Boolean negation of each other, return *false*.
```
shvaikalesh: "Boolean negation of each other" is unnecessarily complex way of saying "is not".
Let's use simple and common prose.

spec.html
```diff
@@ -6949,22 +6949,23 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
...
-          1. If IsGenericDescriptor(_Desc_) is *true*, no further validation is required.
```
shvaikalesh: Considering that name of abstract op is `ValidateAndApplyPropertyDescriptor`, "no further validation is required" may look like "end algorithm here" to infrequent reader.
Let's mark it as note, which it is.

spec.html
```diff
@@ -6949,22 +6949,23 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
...
-          1. Else if IsDataDescriptor(_current_) and IsDataDescriptor(_Desc_) have different results, then
```
shvaikalesh: Please note that w/o `!`, "results" here are completion records, making "different" pretty vague.
Let's unwrap values and compare them in common and consise way.

spec.html
jmdyck:
```suggestion
 <emu-table id="table-default-attribute-values" caption="Default Attribute Values" oldids="table-4">
```
since there might be external links to `table-4`.

shvaikalesh: Perfect, thank you. b575b33

spec.html
ljharb: I don't think it reads super well as "x is not y", nor does that seem idiomatic (in english terms).

spec.html
ljharb: it might be cleaner to instead nest this entire thing in an "if IsGenericDescriptor is false"?

spec.html
```diff
@@ -6949,22 +6949,23 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
...
+            1. If _Desc_.[[Enumerable]] is present and _Desc_.[[Enumerable]] is not _current_.[[Enumerable]], return *false*.
+          1. If ! IsGenericDescriptor(_Desc_) is *true*,
+            1. NOTE: No further validation is required.
+          1. Else if ! IsDataDescriptor(_current_) is not ! IsDataDescriptor(_Desc_), then
```
ljharb: "if x is not y" doesn't read super clearly to me - it seems better to use `SameValue` here?

shvaikalesh: f0e47dc

spec.html
shvaikalesh: Replaced with `SameValue` in f0e47dc

jmdyck: I agree that the status quo is unnecessarily complex. However,
`<field> is not <field>` (or `<field> is <field>` for that matter)
aren't phrasings that the spec currently uses, and they sound a bit odd to me. Instead, `is different from` has some precedent.

spec.html
jmdyck: Similar to above, the spec doesn't currently use the phrasing `<op-invocation> is not <op-invocation>`. Again, I would suggest `is different from`.

spec.html
shvaikalesh: f1805cc

spec.html
shvaikalesh: f1805cc

spec.html
```diff
@@ -6949,22 +6949,23 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
             1. If _extensible_ is *false*, return *false*.
             1. Assert: _extensible_ is *true*.
             1. If IsGenericDescriptor(_Desc_) is *true* or IsDataDescriptor(_Desc_) is *true*, then
-              1. If _O_ is not *undefined*, create an own data property named _P_ of object _O_ whose [[Value]], [[Writable]], [[Enumerable]] and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its default value.
+              1. If _O_ is not *undefined*, create an own data property named _P_ of object _O_ whose [[Value]], [[Writable]], [[Enumerable]] and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its <emu-xref href="#table-default-attribute-values">default value</emu-xref>.
```
```suggestion
              1. If _O_ is not *undefined*, create an own data property named _P_ of object _O_ whose [[Value]], [[Writable]], [[Enumerable]], and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its <emu-xref href="#table-default-attribute-values">default value</emu-xref>.
```

spec.html
```diff
@@ -6949,22 +6949,23 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
...
             1. Else,
               1. Assert: ! IsAccessorDescriptor(_Desc_) is *true*.
-              1. If _O_ is not *undefined*, create an own accessor property named _P_ of object _O_ whose [[Get]], [[Set]], [[Enumerable]] and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its default value.
+              1. If _O_ is not *undefined*, create an own accessor property named _P_ of object _O_ whose [[Get]], [[Set]], [[Enumerable]] and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its <emu-xref href="#table-default-attribute-values">default value</emu-xref>.
```
```suggestion
              1. If _O_ is not *undefined*, create an own accessor property named _P_ of object _O_ whose [[Get]], [[Set]], [[Enumerable]], and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its <emu-xref href="#table-default-attribute-values">default value</emu-xref>.
```

spec.html
```diff
@@ -6949,22 +6949,23 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
...
             1. Return *true*.
           1. If every field in _Desc_ is absent, return *true*.
           1. If _current_.[[Configurable]] is *false*, then
             1. If _Desc_.[[Configurable]] is present and its value is *true*, return *false*.
-            1. If _Desc_.[[Enumerable]] is present and the [[Enumerable]] fields of _current_ and _Desc_ are the Boolean negation of each other, return *false*.
-          1. If IsGenericDescriptor(_Desc_) is *true*, no further validation is required.
-          1. Else if IsDataDescriptor(_current_) and IsDataDescriptor(_Desc_) have different results, then
+            1. If _Desc_.[[Enumerable]] is present and _Desc_.[[Enumerable]] is different from _current_.[[Enumerable]], return *false*.
```
ljharb: This is fine, but i think it’d be clearer to use SameValue:
```suggestion
            1. If _Desc_.[[Enumerable]] is present and ! SameValue(_Desc_.[[Enumerable]], _current_.[[Enumerable]]) is *false*, return *false*.
```

spec.html
```diff
@@ -6949,22 +6949,23 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
...
+          1. If ! IsGenericDescriptor(_Desc_) is *true*,
+            1. NOTE: No further validation is required.
+          1. Else if ! IsDataDescriptor(_current_) is different from ! IsDataDescriptor(_Desc_), then
```
ljharb: Similarly here, it might be better to do this:
```suggestion
          1. Else if ! SameValue(! IsDataDescriptor(_current_), ! IsDataDescriptor(_Desc_)) is *false*, then
```

spec.html
```diff
@@ -6949,22 +6949,23 @@ <h1>ValidateAndApplyPropertyDescriptor ( _O_, _P_, _extensible_, _Desc_, _curren
...
             1. Return *true*.
           1. If every field in _Desc_ is absent, return *true*.
           1. If _current_.[[Configurable]] is *false*, then
             1. If _Desc_.[[Configurable]] is present and its value is *true*, return *false*.
-            1. If _Desc_.[[Enumerable]] is present and the [[Enumerable]] fields of _current_ and _Desc_ are the Boolean negation of each other, return *false*.
-          1. If IsGenericDescriptor(_Desc_) is *true*, no further validation is required.
-          1. Else if IsDataDescriptor(_current_) and IsDataDescriptor(_Desc_) have different results, then
+            1. If _Desc_.[[Enumerable]] is present and _Desc_.[[Enumerable]] is different from _current_.[[Enumerable]], return *false*.
+          1. If ! IsGenericDescriptor(_Desc_) is *true*,
```
```suggestion
          1. If ! IsGenericDescriptor(_Desc_) is *true*, then
```

shvaikalesh: 5b317cb

spec.html
shvaikalesh: e19fdba

spec.html
shvaikalesh: e19fdba

spec.html
shvaikalesh: e51291a

spec.html
shvaikalesh: e51291a

spec.html
shvaikalesh: Personally, I am not a fan of deep nesting.

Another way may be to extract validation logic to distinct abstract op (and early return for generic descriptors), but I don't see how we can do this without adding complexity to this algorithm.

PR #1615, opened 2019-07-08
Editorial: Add note on [[Delete]] in InternalizeJSONProperty

shvaikalesh:


ljharb:
Can you provide more context on why you feel this is an improvement?

spec.html
```diff
@@ -36386,7 +36389,6 @@ <h1>Runtime Semantics: InternalizeJSONProperty ( _holder_, _name_ )</h1>
                   1. Perform ? _val_.[[Delete]](! ToString(_I_)).
                 1. Else,
                   1. Perform ? CreateDataProperty(_val_, ! ToString(_I_), _newElement_).
-                  1. NOTE: This algorithm intentionally does not throw an exception if CreateDataProperty returns *false*.
```
shvaikalesh: This algorithm ignores failures of both `CreateDataProperty` and `[[Delete]]`, but has notes only about ignoring `CreateDataProperty`.

ljharb: Do other places where [[Delete]] return false throw an exception?

shvaikalesh: Yes, two: [`delete` operator](https://tc39.es/ecma262/#sec-delete-operator-runtime-semantics-evaluation) (step 5.b) and [`DeletePropertyOrThrow`](https://tc39.es/ecma262/#sec-deletepropertyorthrow) (step 4).

Handling `false` results of `[[Delete]]` and `[[DefineOwnProperty]]` looks very straightforward in spec with [`DeletePropertyOrThrow`](https://tc39.es/ecma262/#sec-deletepropertyorthrow) and [`DefinePropertyOrThrow`](https://tc39.es/ecma262/#sec-definepropertyorthrow) respectively.

However, some engines (JSC at least) don't have those ops and pass around a "should throw" flag as an argument. A note would be nice, IMO.

raulsebastianmihaila:
I can't think of a case where `CreateDataProperty` would throw inside `InternalizeJSONProperty` because the value is always either an ordinary object or an array that hasn't been touched before.

shvaikalesh:
@raulsebastianmihaila Normally, yes, but consider the following case:

```js
JSON.parse('{"a":1, "b":2}', function(key, value) {
  if (key == "a") Object.freeze(this)
  if (key == "b") return 3
  return value
})

// => {a:1, b:2} // not 3
```

Related JSC issue (with some more test cases in patch):
* [JSON.parse should not modify non-configurable properties](https://bugs.webkit.org/show_bug.cgi?id=163446)

raulsebastianmihaila:
@shvaikalesh Thanks! I started thinking about a similar scenario but then I got confused by the variable names.

PR #1620, opened 2019-07-10
Editorial: `FlattenIntoArray`: add assertions

ljharb:
Add some assertions to clarify the intended usage of this abstract operation.

I wanted to assert that `target` is an Array exotic object, but I think because of `ArraySpeciesCreate` I can't.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -32111,6 +32111,12 @@ THH:mm:ss.sss
         <emu-clause id="sec-flattenintoarray" aoid="FlattenIntoArray">
           <h1>FlattenIntoArray ( _target_, _source_, _sourceLen_, _start_, _depth_ [ , _mapperFunction_, _thisArg_ ] )</h1>
           <emu-alg>
+            1. Assert: Type(_target_) is Object.
+            1. Assert: Type(_source_) is Object.
+            1. Assert: _sourceLen_ is an integer Number &ge; 0.
+            1. Assert: _start_ is an integer Number &ge; 0.
+            1. Assert: _depth_ is an integer Number, *+&infin;*, or *-&infin;*.
+            1. Assert: If _mapperFunction_ is present, then ! IsCallable(_mapperFunction_) is *true*, _thisArg_ is present, and _depth_ is *1*.
             1. Let _targetIndex_ be _start_.
             1. Let _sourceIndex_ be 0.
             1. Repeat, while _sourceIndex_ &lt; _sourceLen_
@@ -32119,7 +32125,6 @@ THH:mm:ss.sss
               1. If _exists_ is *true*, then
                 1. Let _element_ be ? Get(_source_, _P_).
                 1. If _mapperFunction_ is present, then
-                  1. Assert: _thisArg_ is present.
                   1. Set _element_ to ? Call(_mapperFunction_, _thisArg_, &laquo; _element_, _sourceIndex_, _source_ &raquo;).
                 1. Let _shouldFlatten_ be *false*.
                 1. If _depth_ &gt; 0, then
@@ -32143,7 +32148,7 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _O_ be ? ToObject(*this* value).
           1. Let _sourceLen_ be ? LengthOfArrayLike(_O_).
-          1. If IsCallable(_mapperFunction_) is *false*, throw a *TypeError* exception.
+          1. If ! IsCallable(_mapperFunction_) is *false*, throw a *TypeError* exception.
           1. If _thisArg_ is present, let _T_ be _thisArg_; else let _T_ be *undefined*.
           1. Let _A_ be ? ArraySpeciesCreate(_O_, 0).
           1. Perform ? FlattenIntoArray(_A_, _O_, _sourceLen_, 0, 1, _mapperFunction_, _T_).


spec.html
```diff
@@ -32093,6 +32093,12 @@ <h1>Array.prototype.flat ( [ _depth_ ] )</h1>
         <emu-clause id="sec-flattenintoarray" aoid="FlattenIntoArray">
           <h1>FlattenIntoArray ( _target_, _source_, _sourceLen_, _start_, _depth_ [ , _mapperFunction_, _thisArg_ ] )</h1>
           <emu-alg>
+            1. Assert: Type(_target_) is Object.
+            1. Assert: Type(_source_) is Object.
+            1. Assert: _sourceLen_ is an integer Number &ge; 0.
+            1. Assert: _start_ is an integer Number &ge; 0.
+            1. Assert: _depth_ is an integer Number, *+&infin;*, or *-&infin;*.
+            1. Assert: if _mapperFunction_ is present, ! IsCallable(_mapperFunction_) is *true*, _thisArg_ is present, and _depth_ is *1*.
```
```suggestion
            1. Assert: If _mapperFunction_ is present, then ! IsCallable(_mapperFunction_) is *true* and _thisArg_ is present.
```
- Uppercase "I" in "If" because the spec has a strong preference for a capital after "Assert:".
- Insert "then" because otherwise there's a potential garden-path.
(E.g., `If <A>, <B>, ...` could be the start of
`If <A>, <B>, <C>, and <D>.` (meaning `If <A> then <B> and <C> and <D>.`) 
or
`If <A>, <B>, and <C>, <D>.` (meaning `If <A> and <B> and <C> then <D>.`)
It isn't an ambiguity per se, because the position of the "and" tells you where a compound condition ends, but the reader shouldn't have to look ahead and figure it out.)

- Drop the `_depth_ is *1*` condition because, while it happens to be true in the current spec, it's not needed for the correct functioning of this operation. (Consider a variant of `flatMap` that's equivalent to a `map` followed by a `flat` of specified (or infinite) depth: its semantics should be able to invoke `FlattenIntoArray`, and that would work fine, *except* that this `_depth_ is *1*` assertion would fail.)

ljharb: I added the depth assertion because the recursive call doesn’t preserve the mapperFunction or thisArg, so without modification it couldn’t work with that variant anyways.

If that modification is made separately, the depth 1 part of the assertion would thus be removed.

jmdyck: No, I'm saying just a top-level map, followed by X levels of flattening, so the recursive call is just right for that, no modification needed.

ljharb: Maybe we’re talking past each other - if flatMao took a depth (or flat took a mapper), and passed it into the current FlattenIntioArray, it would not work properly, because the mapper wouldn’t be invoked for nested items.

jmdyck: Whether it "works properly" depends on what you think it should do.
- If you're coming from lodash, you would say that it works properly.
- Or consider the quick description of `flatMap`: that it's like a call to `map` followed by a call to `flat`. So if `flatMap` took a depth, then a reasonable expectation would be that it's like a call to `map` followed by a call to `flat(depth)`. So from that point of view, it would also work properly.

I agree that there might be situations in which you'd want the mapper to be applied recursively (whenever it generates an array), and I agree that `FlattenIntoArray` doesn't do that. But the fact that it doesn't support *that* functionality isn't a reason to say that it can't handle a *different* (reasonable, useful) functionality.

ljharb: Fair point. However, the spec currently does not call this operation with both a mapper function and a depth that isn’t 1. So I think the assertion is still appropriate at the moment.

jmdyck: While it's true that the spec only calls `FlattenIntoArray` in ways that satisfy the assertion, those aren't necessarily the only calls -- it's an open spec, and downstream specs can apparently re-use whatever they find useful. An implementation/host might provide (say) `flatMapDepth` and specify it in terms of ECMA-262 operations. If they respect the assertion in question, they'd have to construct their own version of `FlattenIntoArray` that's identical except for omitting the `_depth_ is *1*` condition. Or (more likely?), they'd just use `FlattenIntoArray`, acknowledge the assertion failure, and tell the reader to pretend it doesn't happen. Neither solution is great. So why create unnecessary problems for downstream?

(There's also the case that a future edition of ECMA-262 might wish to add such a function. That's less compelling, because whoever adds the function can remove the `_depth_ is *1*` condition at the same time. But it does still assume that people recognize that the assertion is unnecessary.)

ljharb: I think it's the appropriate thing to do to restrict the possible uses of this abstract operation to what 262 does (modulo known external usages of it) - abstract operations are not a utility library.

jmdyck: > I think it's the appropriate thing to do to restrict the possible uses of this abstract operation to what 262 does (modulo known external usages of it) 

Okay. I disagree, but I'm not going to argue the point further.

> abstract operations are not a utility library.

That's... interesting. I think that deserves a separate issue.

spec.html
```diff
@@ -32093,6 +32093,12 @@ <h1>Array.prototype.flat ( [ _depth_ ] )</h1>
         <emu-clause id="sec-flattenintoarray" aoid="FlattenIntoArray">
           <h1>FlattenIntoArray ( _target_, _source_, _sourceLen_, _start_, _depth_ [ , _mapperFunction_, _thisArg_ ] )</h1>
           <emu-alg>
+            1. Assert: Type(_target_) is Object.
+            1. Assert: Type(_source_) is Object.
+            1. Assert: _sourceLen_ is an integer Number &ge; 0.
```
```suggestion
            1. Assert: _sourceLen_ is an integral Number &ge; 0.
```

spec.html
```diff
@@ -32093,6 +32093,12 @@ <h1>Array.prototype.flat ( [ _depth_ ] )</h1>
...
+            1. Assert: _start_ is an integer Number &ge; 0.
+            1. Assert: _depth_ is an integer Number &ge; 0.
```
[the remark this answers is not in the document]
michaelficarra: > Alternatively, `A.p.flat` could at least map all negatives to (positive) zero.

I like this suggestion a lot.

spec.html
ljharb: hm, i used the same terminology used elsewhere in the spec (plus imo “integral” isn’t unambiguously “integer”)

jmdyck: The current spec has 3 occurrences of "integer Number" and none of "integral Number".

We could change "integer Number" to just "integer", now that it's [defined](https://tc39.es/ecma262/#integer).

As for `&ge; 0` vs `is nonnegative`: the spec uses both, but in algorithm steps, it uses `&ge; 0` more. (I'd say that, in a context where either would fit, you should pick the symbolic, because it's shorter and [it seems to me] easier to understand.)

(Side note: the spec has only one occurrence of "non-negative" with the hyphen. Looks like there was a purge in ES6. Not sure why, because the spec certainly has no problem with the hyphen in terms like non-configurable, non-strict, and non-object.)

ljharb: I'm going to resolve this one for this PR, but please file an issue/PR to make this kind of thing consistent throughout.

PR #1626, opened 2019-07-13
Editorial: fix capitalization of "RegExp String Iterator"

ljharb:
Looks like I forgot to capitalize one of the 'i's.


TimothyGu:
<kbd>Ctrl</kbd>+<kbd>F</kbd> reveals plenty of instances of "iterator object" and "Iterator object" and even "Iterator Object", so I think we should either fix all of them or leave it be.

ljharb:
In this case tho, all RegExp String Iterator Object mentions but this one are capitalized, so i think even fixing only this one is an improvement. We can pursue a holistic change separately.

PR #1633, opened 2019-07-18
Meta: Add Stage 1 proposals list

chicoxyzzy:
Adding missing Stage 1 proposals list to the `README.md`.


chicoxyzzy:
Also I'd suggest to remove Stage 0 proposals list from ECMA262 readme as Stage 0 proposals are not at any of process maturity stages and are not currently in any development by committee. Strawperson proposals could mislead people to think that they are something TC39 works on. I don't have strong opinion on that, just a suggestion.

PR #1645, opened 2019-07-25
Normative: change idempotency for HostImportModuleDynamically

domenic:
This better matches HostResolveImportedModule, which only requires idempotency when it completes normally (which is analogous to the success case here).

Closes https://github.com/tc39/proposal-dynamic-import/issues/80, at least on the ECMAScript side.

---

To be presented at TC39 today; I thought it'd be easier to have a concrete needs-consensus PR rather than trying to page through https://github.com/tc39/proposal-dynamic-import/issues/80.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -22706,7 +22706,7 @@
             </dl>
           </li>
           <li>
-            Every call to HostImportModuleDynamically with the same _referencingScriptOrModule_ and _specifier_ arguments must conform to the <em>same</em> set of requirements above as previous calls do. That is, if the host environment takes the success path once for a given _referencingScriptOrModule_, _specifier_ pair, it must always do so, and the same for the failure path.
+            If the host environment takes the success path once for a given _referencingScriptOrModule_, _specifier_ pair, it must always do so for subsequent calls.
           </li>
           <li>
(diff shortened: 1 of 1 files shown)


ljharb:
Are there test262 tests for this change? (or, is it hard to test/not currently tested?)

Snapstromegon:
@ljharb under language/expressions/dynamic-import are at least the following tests affected by this change:

- double-error-resolution
- double-error-resolution-promise

Since this PR removes required idempotency for double error, I think that those test are no longer needed.

Also ```double-error-resolution_FIXTURE``` is no longer necessary if the tests above are removed.

ljharb:
@Snapstromegon in that case, a merged test262 PR for those changes would be needed here. Want to file it? :-D

fatso83:
Where did this change actually end up? I had a look at the rendered output of https://tc39.es/proposal-dynamic-import/ and it only contained the old text, before this change. Then I tried searching the actual speco of [ecma-262](https://tc39.es/ecma262/), but found hits on either pre or post text.

nicolo-ribaudo:
We unified the HostResolveImportedModule and HostImportModuleDynamically hooks, the requirement is now the second bullet point of https://tc39.es/ecma262/#sec-HostLoadImportedModule.

PR #1652, opened 2019-08-01
Editorial: Simplify AssignmentTargetType values.

### LINKED ISSUE #1643: AssignmentTargetType can be changed back to IsValidSimpleAssignmentTarget
#1132 changed IsValidSimpleAssignmentTarget to AssignmentTargetType to be able to differentiate between ReferenceError and SyntaxErrors when assigning to `eval`/`arguments` in strict-mode. But with #1527 fixed, this change is no longer necessary and AssignmentTargetType can be changed back IsValidSimpleAssignmentTarget. 

-----

AssignmentTargetType returns either `simple`, `invalid`, or `strict`, but since #1527 we only need to differentiate between `simple` and not-`simple`, so a boolean is sufficient, which was exactly the return type of IsValidSimpleAssignmentTarget.

rkirsling:
Resolves #1643.

As expressed there, boolean seems best if `nonsimple` is the best enum value we can come up with. I've also taken the opportunity to simplify the name to `IsSimpleAssignmentTarget` since I'm not sure that "valid" really serves to clarify anything (but feel free to disagree).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12244,7 +12244,7 @@
       <emu-see-also-para op="AssignmentTargetType"></emu-see-also-para>
       <emu-grammar>IdentifierReference : Identifier</emu-grammar>
       <emu-alg>
-        1. If this |IdentifierReference| is contained in strict mode code and StringValue of |Identifier| is *"eval"* or *"arguments"*, return ~strict~.
+        1. If this |IdentifierReference| is contained in strict mode code and StringValue of |Identifier| is *"eval"* or *"arguments"*, return ~invalid~.
         1. Return ~simple~.
       </emu-alg>
       <emu-grammar>IdentifierReference : `yield`</emu-grammar>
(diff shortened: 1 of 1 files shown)


leobalter:
cc @DanielRosenwasser this affects the optional chaining proposal which still uses `IsValidSimpleAssignmentTarget`. 

Thanks for this PR, @rkirsling!

DanielRosenwasser:
Thanks for the heads up! I'll change the proposals and existing PRs to reflect the appropriate changes.

rkirsling:
@ljharb That's fine, but you haven't counter-suggested what the false value should be called. 😕

ljharb:
The op's called "IsSimpleAssignmentTarget" - "simple" vs "nonsimple" or "complex"?

Alternatively we could rename the abstract operation itself but that'd be a harder bikeshed.

rkirsling:
Oh hmm, apparently the return values for `GetGeneratorKind` include `~non-generator~`, so I guess that could be a precedent for `~non-simple~`. Anybody have a strong preference?

rkirsling:
Changed to `~non-simple~` after all. I really think this is the best we can do given that the opposite of simple is _either_ complex or invalid (and we don't actually have a reason to distinguish among those).

rkirsling:
Switched to @zenparsing's suggestion.

rkirsling:
Done!

rkirsling:
(Sorry, misclicked "close and comment"...)

PR #1658, opened 2019-08-07
Normative: Add two missing Early Error rules for productions covering an AssignmentPattern

claudepache:
Refactor and correct Early Error rules for LeftHandSideExpression productions potentially covering an AssignmentPattern:

* Add two missing early error rules  “All Early Error rules for |AssignmentPattern| ... are applied to the |AssignmentPattern| covered by ...”

* Removes an early error rule “... derives a phrase that would produce a Syntax Error ...”, because it is implied by the previous rule.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -15397,18 +15397,25 @@
     <emu-clause id="sec-assignment-operators-static-semantics-early-errors">
       <h1>Static Semantics: Early Errors</h1>
       <emu-grammar>AssignmentExpression : LeftHandSideExpression `=` AssignmentExpression</emu-grammar>
+      <p>If |LeftHandSideExpression| is an |ObjectLiteral| or an |ArrayLiteral|, the following Early Error rules are applied:</p>
       <ul>
         <li>
-          It is a Syntax Error if |LeftHandSideExpression| is either an |ObjectLiteral| or an |ArrayLiteral| and |LeftHandSideExpression| is not covering an |AssignmentPattern|.
+          It is a Syntax Error if |LeftHandSideExpression| is not covering an |AssignmentPattern|.
         </li>
         <li>
-          It is an early Syntax Error if |LeftHandSideExpression| is neither an |ObjectLiteral| nor an |ArrayLiteral| and AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
+          All Early Error rules for |AssignmentPattern| and its derived productions also apply to the |AssignmentPattern| that is covered by |LeftHandSideExpression|.
+        </li>
+      </ul>
+      <p>If |LeftHandSideExpression| is neither an |ObjectLiteral| nor an |ArrayLiteral|, the following Early Error rule is applied:</p>
+      <ul>
+        <li>
+          It is a Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
         </li>
       </ul>
       <emu-grammar>AssignmentExpression : LeftHandSideExpression AssignmentOperator AssignmentExpression</emu-grammar>
       <ul>
         <li>
-          It is an early Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
+          It is a Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
         </li>
       </ul>
     </emu-clause>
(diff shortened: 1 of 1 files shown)


claudepache:
See https://github.com/tc39/ecma262/issues/1643#issuecomment-518844021 for background.

spec.html
```diff
@@ -14144,18 +14144,19 @@ <h2>Syntax</h2>
...
         <li>
-          It is a Syntax Error if |LeftHandSideExpression| is either an |ObjectLiteral| or an |ArrayLiteral| and |LeftHandSideExpression| is not covering an |AssignmentPattern|.
+          It is a Syntax Error if |LeftHandSideExpression| is not covering an |AssignmentPattern|.
         </li>
         <li>
-          It is an early Syntax Error if |LeftHandSideExpression| is neither an |ObjectLiteral| nor an |ArrayLiteral| and AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
+          All Early Error rules for |AssignmentPattern| and its derived productions also apply to the |AssignmentPattern| that is covered by |LeftHandSideExpression|.
         </li>
       </ul>
-      <emu-grammar>AssignmentExpression : LeftHandSideExpression AssignmentOperator AssignmentExpression</emu-grammar>
```
jmdyck: You've deleted the rule for `LeftHandSideExpression AssignmentOperator AssignmentExpression`

jmdyck:
Re the removal of the "recursively applied" rule: In the commit message, it might be clearer to say that the recursion of the removed rule is handled by the fact that AssignmentTargetType recurses through CoverParenthesizedExpressionAndArrowParameterList.

(The "recursively applied" rule was introduced slightly before the forerunner of AssignmentTargetType, so I guess nobody noticed when the former became redundant.)

spec.html
claudepache: Whoops, fixed.

spec.html
```diff
@@ -14144,18 +14144,25 @@ <h2>Syntax</h2>
...
+      <ul>
+        <li>
+          It is a Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
         </li>
       </ul>
       <emu-grammar>AssignmentExpression : LeftHandSideExpression AssignmentOperator AssignmentExpression</emu-grammar>
       <ul>
         <li>
-          It is an early Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
+          It is a Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
```
rkirsling: Seems like there are only three other matches for `early Syntax Error` in the spec—shall we update those right away too?

jmdyck:
Looks good to me.

ljharb:
This seems like a normative change, but also a spec bug - are there any engines implementing the behavior on master, or do they all implement the behavior in this PR?

claudepache:
@ljharb It is a spec bug, and the intention was clear. The “behavior of master” is literally impossible to implement, because there exists a grammatically correct AssignmentPattern which has a derived production that falls into some early error condition of the form “It is a Syntax Error if X is not covering Y”, and whose runtime semantics contains a step saying “Let y be the Y covered by X”. It is necessary to check statically that early error condition for the runtime semantics to make sense.

Without going so far, I’ve checked that at least two implementations treats `try { [ this ] = [ ] } catch (e) { }` as a syntax error (in accordance with the patch of this PR), as `[ this ]` is an AssignmentPattern per the grammar but has an Early Error condition saying that it is a Syntax Error. But, of course, that proves nothing.

spec.html
claudepache: In fact, that one was corrected by accident, because I first mistakingly deleted that rule, then restored it without the redundant adjective.

jmdyck:
@ljharb:
> This seems like a normative change

There are a few things happening in this PR; could you be more specific about which seems like a normative change?

ljharb:
Seems like it’s changing how grammar rules are applied, which would be normative if it produced observable differences.

jmdyck:
Are you particularly concerned about the 2 insertions of the sentence
> All Early Error rules for |AssignmentPattern| and its derived productions also apply to the |AssignmentPattern| that is covered by |LeftHandSideExpression|.

or are there other edits that seem to be changing how rules are applied?

ljharb:
tbh I'm not completely sure; that's why I asked.

syg:
The sense in which this change is "normative" isn't really well-formed per https://github.com/tc39/ecma262/pull/1658#issuecomment-519406975. It's normative in the sense that all spec bugs are just, by definition, normative, but not normative in the sense that we should be on the lookout for introducing unintended behavior.

ljharb:
@claudepache can you rebase this PR on master? there's a conflict i'm uncertain of how to resolve.

bakkot:
Force-pushed to resolve the conflict with e6f8460c.

PR #1660, opened 2019-08-09
Editorial: Remove assertion in InitializeBinding of Object Environment Record

jhnaldo:
In declarative environment records, CreateMutableBinding and CreateImmutableBinding creates uninitialized bindings. Unlike declarative environment records, in object environment records, such internal methods does not create uninitialized bindings. They just creates a new property for the binding object with `undefined` value and there is no mention about uninitialized bindings. However, in InitializeBinding of object environment records, it requires uninitialized bindings. I think that it is better to remove the following highlighted assertion.

PastedGraphic-4

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6535,15 +6535,13 @@
 
         <emu-clause id="sec-object-environment-records-initializebinding-n-v">
           <h1>InitializeBinding ( _N_, _V_ )</h1>
-          <p>The concrete Environment Record method InitializeBinding for object Environment Records is used to set the bound value of the current binding of the identifier whose name is the value of the argument _N_ to the value of argument _V_. An uninitialized binding for _N_ must already exist.</p>
+          <p>The concrete Environment Record method InitializeBinding for object Environment Records is used to set the bound value of the current binding of the identifier whose name is the value of the argument _N_ to the value of argument _V_.</p>
           <emu-alg>
             1. Let _envRec_ be the object Environment Record for which the method was invoked.
-            1. Assert: _envRec_ must have an uninitialized binding for _N_.
-            1. <emu-not-ref>Record</emu-not-ref> that the binding for _N_ in _envRec_ has been initialized.
             1. Return ? _envRec_.SetMutableBinding(_N_, _V_, *false*).
           </emu-alg>
           <emu-note>
-            <p>In this specification, all uses of CreateMutableBinding for object Environment Records are immediately followed by a call to InitializeBinding for the same name. Hence, implementations do not need to explicitly track the initialization state of individual object Environment Record bindings.</p>
+            <p>In this specification, all uses of CreateMutableBinding for object Environment Records are immediately followed by a call to InitializeBinding for the same name. Hence, this specification does not explicitly track the initialization state of bindings in object Environment Records.</p>
           </emu-note>
         </emu-clause>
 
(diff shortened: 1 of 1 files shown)


anba:
See https://tc39.es/archives/bugzilla/1782/ for the motivation to keep this assertion.

syg:
> See https://tc39.es/archives/bugzilla/1782/ for the motivation to keep this assertion.

I'm having trouble extracting the motivation from that thread. If initialization in object environments is a theoretical capability that doesn't manifest in reality **and** deserves a note to explain why there's verbiage about initialization, then... isn't that actually anti-motivation from an editorial clarity point of view?

spec.html
```diff
@@ -5357,7 +5357,6 @@ <h1>InitializeBinding ( _N_, _V_ )</h1>
           <p>The concrete Environment Record method InitializeBinding for object Environment Records is used to set the bound value of the current binding of the identifier whose name is the value of the argument _N_ to the value of argument _V_. An uninitialized binding for _N_ must already exist.</p>
```
syg: The last sentence about "An uninitialized binding ..." should also be removed.

spec.html
```diff
@@ -5357,7 +5357,6 @@ <h1>InitializeBinding ( _N_, _V_ )</h1>
...
           <emu-alg>
             1. Let _envRec_ be the object Environment Record for which the method was invoked.
-            1. Assert: _envRec_ must have an uninitialized binding for _N_.
             1. Record that the binding for _N_ in _envRec_ has been initialized.
```
syg: This step also doesn't make any intuitive sense and should be removed as well.

bakkot:
I've rebased to resolve a merge conflict, and this change still looks good.

However, after looking into this more, I am not convinced that `CreateMutableBinding ` or `InitializeBinding` are ever invoked for object environment records. I certainly can't find a path where it is used, and, thinking about it from the JS side of things, I am pretty sure that `with (o)` cannot create properties on `o` (without referencing `o`, of course).

bakkot:
@allenwb do you know why object environment records have semantics for `CreateMutableBinding`? Your comments in [this bug](https://tc39.es/archives/bugzilla/1782/) imply that you have in mind some place it is used, but I can't find one.

anba:
@bakkot It's used for the [[ObjectRecord]] of the global environment record.

anba:
The note in [InitializeBinding](https://tc39.es/ecma262/#sec-object-environment-records-initializebinding-n-v) should also be removed.

bakkot:
@anba Ah, thank you.

Now that I know to look at [[ObjectRecord]], there's another place where the "Record that the binding has been initialized." step occurs for object records: step 8 of [CreateGlobalFunction](https://tc39.es/ecma262/#sec-createglobalfunctionbinding). I've pushed up another commit which removes it that step (which should be merged with the previous one when merging this PR, or all three commits could be merged).

> The note in InitializeBinding should also be removed.

Why? It seems fine to me.

anba:
The note:
> In this specification, all uses of CreateMutableBinding for object Environment Records are immediately followed by a call to InitializeBinding for the same name. Hence, implementations do not need to explicitly track the initialization state of individual object Environment Record bindings.

is meant to refer to the step:
> Record that the binding for N in envRec has been initialized.

But when that step gets removed, the note should be removed, too.

bakkot:
> But when that step gets removed, the note should be removed, too.

Hm. I think the note still makes sense: it explains why the algorithm does not track initialization for bindings in object records.

---

Incidentally, I think that steps 4-5 of [InitializeBinding](https://tc39.es/ecma262/#sec-global-environment-records-initializebinding-n-v) for global environment records is never hit. But it's a little hard to tell.

bakkot:
I've removed the second sentence of the note, so it just reads

> In this specification, all uses of CreateMutableBinding for object Environment Records are immediately followed by a call to InitializeBinding for the same name.

syg:
> In this specification, all uses of CreateMutableBinding for object Environment Records are immediately followed by a call to InitializeBinding for the same name.

Did you forget to push? I think I'd prefer the note to anticipate the question of initialization and rephrase the second sentence as

> Hence, this specification does not explicitly track the initialization state of bindings in object Environment Records.

bakkot:
> Did you forget to push

Yup.

> I  think I'd prefer the note to anticipate the question of initialization and rephrase the second sentence as

> > Hence, this specification does not explicitly track the initialization state of bindings in object Environment Records.

Done.

Also rebased.

PR #1066, opened 2018-01-12
Normative: add async iteration

domenic:
This supersedes @zbraniecki's #1063. The individual commits are just to show the delta from that.

One potential question for the editor is the ordering of sections within 25 (Control Abstraction Objects). I went with

- Iteration
- GeneratorFunction Objects
- **AsyncGeneratorFunction Objects**
- Generator Objects
- **AsyncGenerator Objects**
- Promise Objects
- AsyncFunction Objects

but I can also see an argument for keeping the two new sections together, perhaps even after the AsyncFunction Objects section.

(Closes #1063)

PR #1661, opened 2019-08-09
Editorial: Resolved ambiguous VarScopedDeclarations for IterationStatement

jhnaldo:
I think that the following two highlighted lines are duplicated. I think the first one is not correct because VarScopedDeclarations collects all declarations with "var". Thus, I propose to remove the first one in the spec.

PastedGraphic-1

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -16689,7 +16689,6 @@
         <emu-grammar>
           IterationStatement :
             `for` `(` LeftHandSideExpression `of` AssignmentExpression `)` Statement
-            `for` `await` `(` `var` ForBinding `of` AssignmentExpression `)` Statement
         </emu-grammar>
         <emu-alg>
           1. Return the VarScopedDeclarations of |Statement|.


anba:
LGTM.

And: Oh wow, [VarDeclaredNames](https://tc39.es/ecma262/#sec-for-in-and-for-of-statements-static-semantics-vardeclarednames) and [VarScopedDeclarations](https://tc39.es/ecma262/#sec-for-in-and-for-of-statements-static-semantics-varscopeddeclarations) are confusingly ordered. That should be clean-ed up at some point.

rkirsling:
Looks like this was introduced in #1066.

Seems like (both here and in the previous section, [13.7.5.7](https://tc39.es/ecma262/#sec-for-in-and-for-of-statements-static-semantics-vardeclarednames)) there are two `<emu-grammar/><emu-alg/>` chunks in a row that say the same thing—shall we merge them together?

jmdyck:
@anba and @rkirsling, you should maybe check that #1284 resolves your additional concerns.

(Maybe if more people review it, it will get merged sooner.)

TimothyGu:
#1284 has been merged, maybe this could be closed now?

ljharb:
ah, thanks. indeed, superseded by #1284.

PR #1662, opened 2019-08-09
Editorial: Added more missing cases for ExpectedArgumentCount for FormalParameters and FormalParameterList

jhnaldo:
FormalParameterList is either 1) “FormalParamer" or 2) "FormalParameterList , FormalParameter”.
However, for ExpectedArgumentCount, only "FormalParameterList , FormalParameter” is defined
but “FormalParameter” is not defined. According to the NOTE, it might be 0 if it has initializer.
Otherwise, it might be 0. I think it also should be formally defined like other semantics.

In a similar way, there is no semantics of ExpectedArgumentCount for "FormalParameters : FormalRestParameter” and “FormalParameters : FormalParameterList , “. Thus, I changed the spec to cover all cases I mentioned above.

PastedGraphic-2

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -18313,17 +18313,28 @@
     <emu-clause id="sec-function-definitions-static-semantics-expectedargumentcount">
       <h1>Static Semantics: ExpectedArgumentCount</h1>
       <emu-see-also-para op="ExpectedArgumentCount"></emu-see-also-para>
-      <emu-grammar>FormalParameters : [empty]</emu-grammar>
+      <emu-grammar>FormalParameters :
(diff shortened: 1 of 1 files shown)


anba:
The addition of `FormalParameters : FormalParameterList ','` is not necessary, because of the [chain production rule](https://tc39.es/ecma262/#sec-algorithm-conventions-syntax-directed-operations). The other additions seem correct and actually where already present in the ECMAScript 6 spec. Somehow they got removed in 6a06509a...

bakkot:
Subsumed by #1662.

Issue #1139, opened 2018-03-20
Missing IsFunctionDefinition for anonymous FunctionExpression


jorendorff:
See [14.1.12](https://tc39.github.io/ecma262/#sec-function-definitions-static-semantics-isfunctiondefinition). It only seems to be specified for the production where the BindingIdentifier is present. Likewise, GeneratorExpression ([14.4.8](https://tc39.github.io/ecma262/#sec-generator-function-definitions-static-semantics-isfunctiondefinition)) and AsyncGeneratorExpression (14.5.8) only specify IsFunctionDefinition for the production with BindingIdentifier.

The corresponding spec for async functions, [14.7.8](https://tc39.github.io/ecma262/#sec-async-function-definitions-static-semantics-IsFunctionDefinition), lists both productions.

I think either the other three sections should be changed to list both productions, like 14.7.8; or else all four should be changed to use `_opt`.

bakkot:
The `?` which should render as "\_opt" is actually present in the source, I think (line 18841, which I can't link to since it's too big).

Pretty sure this is a [bug](https://github.com/bterlson/ecmarkup/issues/108) in ecmarkup. cc @bterlson.

anba:
See #536

Issue #536, opened 2016-04-12
14.1.12 Static Semantics: IsFunctionDefinition


CountOrlok:
I think there is something missing in [IsFunctionDefinition](https://tc39.github.io/ecma262/#sec-function-definitions-static-semantics-isfunctiondefinition) at 14.1.12. The operation as described here will only return true if the given production is a _named_ function definition, but false for _anonymous_ function definitions.

Therefore when calling [IsAnonymousFunctionDefinition](https://tc39.github.io/ecma262/#sec-isanonymousfunctiondefinition) with argument _production_, the first algorithm step would always return false for anonymous function expressions and the operation [HasName](https://tc39.github.io/ecma262/#sec-function-definitions-static-semantics-hasname) would be unreachable.

So it rather should be:

**14.1.12 Static Semantics: IsFunctionDefinition**

FunctionExpression: function (FormalParameters) {FunctionBody}

FunctionExpression: function BindingIdentifier (FormalParameters) {FunctionBody}

Return true. 

[as in ES2015: [IsFunctionDefiniton](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-function-definitions-static-semantics-isfunctiondefinition)]

Or is there a reason for changing this?

GeorgNeis:
I agree. (This also applies to classes and generator functions.)

allenwb:
Right, as noted in the original issue,  this appears to be a regress from the original ES6. The bug also occurs in  [14.4.9](https://tc39.github.io/ecma262/#sec-generator-function-definitions-static-semantics-isfunctiondefinition) and [14.5.8](https://tc39.github.io/ecma262/#sec-class-definitions-static-semantics-isfunctiondefinition)

PR #1663, opened 2019-08-09
Editorial: Added a missing case of IsFunctionDefinition for FunctionExpression

jhnaldo:
According to other rules (e.g. 13.7.4.2 Static Semantics: ContainsDuplicateLabels),
the “opt” subscript is attached to represent both absent and present cases for optional terms.
Thus, the following case does not represent the case when BindingIdentifier is absent.
Then, there is no definition of IsFunctionDefinition for “function ( FormalParameters ) { FunctionBody }” case. I think that the “opt” subscript should be attached into BindingIdentifier to resolve misunderstandings.

Note) I think that there is an error in rewriting system in specification for single line grammar. The problem is resolved when I changed into the single line grammar into multi-line grammar.

PastedGraphic-3

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -18382,7 +18382,10 @@
     <emu-clause id="sec-function-definitions-static-semantics-isfunctiondefinition">
       <h1>Static Semantics: IsFunctionDefinition</h1>
       <emu-see-also-para op="IsFunctionDefinition"></emu-see-also-para>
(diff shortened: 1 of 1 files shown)


bakkot:
See #1139, #536, https://github.com/bterlson/ecmarkup/issues/108.

PR #1664, opened 2019-08-09
Editorial: Fixed error for already unwrapped completion records

jhnaldo:
The Get and Call function always return abrupt completion or non-completion values
because the ReturnIfAbrupt(?) is applied right before return the result in Get and Call.
Thus, in the following cases, _mappedValue_ in Array.from and _k_ and _v_ in AddEntriesFromIterable are not completion but just non-completion values because the abrupt completion cases are already filtered previous conditions. Thus, I suggest to remove access of [[Value]] for _mappedValue_, _k_, and _v_.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -31759,7 +31759,6 @@ THH:mm:ss.sss
               1. If _mapping_ is *true*, then
                 1. Let _mappedValue_ be Call(_mapfn_, _T_, &laquo; _nextValue_, _k_ &raquo;).
                 1. If _mappedValue_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _mappedValue_).
-                1. Set _mappedValue_ to _mappedValue_.[[Value]].
               1. Else, let _mappedValue_ be _nextValue_.
               1. Let _defineStatus_ be CreateDataPropertyOrThrow(_A_, _Pk_, _mappedValue_).
               1. If _defineStatus_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _defineStatus_).
@@ -34285,7 +34284,7 @@ THH:mm:ss.sss
             1. If _k_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _k_).
             1. Let _v_ be Get(_nextItem_, `"1"`).
             1. If _v_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _v_).
-            1. Let _status_ be Call(_adder_, _target_, &laquo; _k_.[[Value]], _v_.[[Value]] &raquo;).
+            1. Let _status_ be Call(_adder_, _target_, &laquo; _k_, _v_ &raquo;).
             1. If _status_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _status_).
         </emu-alg>
         <emu-note>


jmdyck:
Personally, I'm fine with these edits. However, I feel obliged to point out that they go against the current editorial direction (and so I'm puzzled that @ljharb approves).

In more detail:
1. These changes are fine (in fact, they're necessary corrections) in an interpretation where `Call` and `Get` can return both abrupt completions and non-completion values (as mentioned in the initial message above). While this is my preferred interpretation (see issue #497), I don't think it's widely accepted. (But I'll be happy to hear otherwise.)

2. The changes are also valid (though not necessary) in an interpretation where `Call` and `Get` always return a completion record (sometimes abrupt, sometime normal), and you're willing to make use of the rule whereby a normal completion can be implicitly coerced to its [[Value]] (see [5.2.3.1](https://tc39.es/ecma262/#sec-implicit-completion-values)). However, there seems to be a general preference not to use that rule (or at least, no more than the spec already is).

3. So, if `Call` and `Get` always return a completion record, and there's a preference to avoid implicit coercion-to-[[Value]], you instead have to either explicitly 'unwrap' a normal completion via '?' and '!', or explicitly reference its [[Value]] field. The latter is what the existing steps do, and this PR is moving away from that.

So @ljharb, is 3 *not* the current editorial direction, and you're actually okay with 1 or 2?

anba:
I don't think these changes should be applied, because in addition to the reasons outlined above, it'll also create inconsistencies with the rest of the spec. For example compare [IteratorClose](https://tc39.es/ecma262/#sec-iteratorclose), where step 8 will be ambiguous, because it's not clear if `innerResult.[[Type]]` means `innerResult` is an unwrapped normal completion whose `[[Type]]` field is accessed or if `innerResult` is a completion record.

ljharb:
@jmdyck I stamped based on number 1; it seemed like it's a bugfix. I looked at Call, and it appears to indeed return either an abrupt completion, or a non-completion-record. However, it's entirely possible I made a mistake, and the non-completion-record gets implicitly wrapped in a NormalCompletion, which would make master correct as-is.

I agree, option 2 isn't what I'd want to see; now that you state it that way, option 3 seems preferable to me, which would mean this PR shouldn't land.

allenwb:
Note that  as [originally specified in ES6](https://www.ecma-international.org/ecma-262/6.0/index.html#sec-call) \(plus [requirement](https://www.ecma-international.org/ecma-262/6.0/index.html#sec-object-internal-methods-and-internal-slots) that internal methods return completion records) the Call abstract operation always returned a completion record. So, if it  now may return both completion records and non-completion values, that is surely a bug that potentially impacts all uses of Call.

ljharb:
Sounds like this should be closed, then?

jmdyck:
> For example compare [IteratorClose](https://tc39.es/ecma262/#sec-iteratorclose), where step 8 will be ambiguous, because it's not clear if `innerResult.[[Type]]` means `innerResult` is an unwrapped normal completion whose `[[Type]]` field is accessed or if `innerResult` is a completion record.

It could only be ambiguous if a completion record's [[Value]] could be a record (with a [[Type]] field), which is supposedly not allowed. (6.2.3 says that [[Value]] must be an ES language value or empty.)

jhnaldo:
What is the exact rule of implicitly wrapping the values with NormalCompletion?
I think that the implicit conversions to completion records should be occurred only in run-time semantics as described in [Section 5.2.3.1](https://tc39.es/ecma262/#sec-runtime-semantics).
Is Call also run-time semantics? If so, why the last ReturnIfAbrupt(?) is needed?
I think that
image
should be changed into
image

jmdyck:
> What is the exact rule of implicitly wrapping the values with NormalCompletion?

It's the rule at the start of 5.2.3.1.

> I think that the implicit conversions to completion records should be occurred only in run-time semantics as described in [Section 5.2.3.1](https://tc39.es/ecma262/#sec-runtime-semantics).

I believe that's actually the case. 

> Is Call also run-time semantics?

Certainly. (Roughly speaking, runtime semantics is everything that isn't parsing or checking for early errors. So there's lots of stuff that is runtime semantics but isn't explicitly labelled as such.)

> If so, why the last ReturnIfAbrupt(?) is needed?

Normatively, it isn't. But editorially, it reminds the reader that that invocation of `Call` might return an abrupt completion.

> I think that
>   3. Return ? F.[[Call]](V, argumentsList). 
> should be changed into
>   3. Return F.[[Call]](V, argumentsList).

Like I say above, that would be normatively equivalent. So then it's a question of whether the editors prefer to have the '?' there.

allenwb:
> Certainly. (Roughly speaking, runtime semantics is everything that isn't parsing or checking for early errors. So there's lots of stuff that is runtime semantics but isn't explicitly labelled as such.)

"Static semantics" is actually the special case.  Static semantics are for things that came be determined by examining the source code without reference to any runtime (ie, dynamic) state or mechanisms (such as completion records).  They are things that can be determined outside of an ES execution environment. Hence a static semantic routine should never produce a completion record. 

While syntax-linked abstract operations were carefully tagged as either static or runtime semantics, this wasn't the case for other abstraction operations.  The default was runtime.  Any such abstract operations that are used by explicitly tagged static semantics or early errors should probably also be tagged as static semantics.

jhnaldo:
> Like I say above, that would be normatively equivalent. So then it's a question of whether the editors prefer to have the '?' there.

Thank you for kind explanation.
Now, I understand the intention of completion record and its implicitness.
I will close this pull request.

PR #353, opened 2016-02-03
Always update object properties if the property descriptor is applicable

anba:
https://bugs.ecmascript.org/show_bug.cgi?id=3508 contains the background information for this change. And as mentioned in https://bugs.ecmascript.org/show_bug.cgi?id=3508#c8, this change needs to be reviewed very carefully.

PR #1677, opened 2019-08-26
Normative: `ValidateAndApplyPropertyDescriptor`: fix bug that bypasses "apply"

ljharb:
This bug was introduced in https://github.com/tc39/ecma262/pull/353/ - it prevents the change from actually being applied.

I believe the confusion was trying to be consistent with the `return true` above, but that one is in the "current is undefined" section, in which the change is not applied anyways.

I discovered this while implementing `ArraySetLength` in https://npmjs.com/es-abstract, which invoked my implementation of `ValidateAndApplyPropertyDescriptor`.


anba:
In the data-descriptor path (step 7.a.iii), `Return true` must not be removed, because otherwise it's possible to replace a NaN value with a different NaN value for frozen properties. For the accessor-descriptor path (step 8.b.iii), it shouldn't matter much if `Return true` is present or not. I've probably just added it for consistency with step 7.a.iii.
#353 and the original bugzilla issue have more info about the different NaN values case.

PR #1692, opened 2019-09-07
Strengthen Atomics critical section to imply synchronizes-with

### LINKED ISSUE #1680: (memory model, wait/notify) Atomics.wait/notify non-SC behaviour, what is expected?
Currently, as specified, it's possible for Atomics.wait/notify to interact in a non-sequentially-consistent way.

|Thread 1                      | Thread 2 |
|------------------------------|-------------- |
|   (a): Atomics.wait(tA, 0, 0) |  (b): Atomics.store(tA, 0, 1)   |
|       | (c): var x = Atomics.notify(tA, 0, 1) |

Given the following program, it's permitted to have the (operational) interleaving (b) -> (c) -> (a), but for the Atomics.wait of (a) to nevertheless read 0, therefore missing both the store and the notify, and getting stuck.

Is this intended? This feature doesn't seem to be formally specified in any other language, so there's no real "prior art". The closest analogy might be the C/C++11 try_lock operation, which is surprisingly weak. However this isn't a perfect match, and conceptually it seems like Atomics.* operations are expected to act in a reasonably sequential way.

syg:
@conrad-watt @lars-t-hansen 

Fixes #1680.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37250,9 +37250,10 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-getwaiterlist" aoid="GetWaiterList">
         <h1>GetWaiterList ( _block_, _i_ )</h1>
-        <p>A <dfn>WaiterList</dfn> is a semantic object that contains an ordered list of those agents that are waiting on a location (_block_, _i_) in shared memory; _block_ is a Shared Data Block and _i_ a byte offset into the memory of _block_.</p>
+        <p>A <dfn>WaiterList</dfn> is a semantic object that contains an ordered list of those agents that are waiting on a location (_block_, _i_) in shared memory; _block_ is a Shared Data Block and _i_ a byte offset into the memory of _block_. A WaiterList object also optionally contains a Synchronize event denoting the previous leaving of its critical section.</p>
+        <p>Initially a WaiterList object has an empty list and no Synchronize event.</p>
         <p>The agent cluster has a store of WaiterList objects; the store is indexed by (_block_, _i_). WaiterLists are agent-independent: a lookup in the store of WaiterLists by (_block_, _i_) will result in the same WaiterList object in any agent in the agent cluster.</p>
-        <p>Each WaiterList has a <dfn>critical section</dfn> that controls exclusive access to that WaiterList during evaluation. Only a single agent may enter a WaiterList's critical section at one time. Entering and leaving a WaiterList's critical section is controlled by the abstract operations EnterCriticalSection and LeaveCriticalSection. Operations on a WaiterList&mdash;adding and removing waiting agents, traversing the list of agents, suspending and notifying agents on the list&mdash;may only be performed by agents that have entered the WaiterList's critical section.</p>
+        <p>Each WaiterList has a <dfn>critical section</dfn> that controls exclusive access to that WaiterList during evaluation. Only a single agent may enter a WaiterList's critical section at one time. Entering and leaving a WaiterList's critical section is controlled by the abstract operations EnterCriticalSection and LeaveCriticalSection. Operations on a WaiterList&mdash;adding and removing waiting agents, traversing the list of agents, suspending and notifying agents on the list, setting and retrieving the Synchronize event&mdash;may only be performed by agents that have entered the WaiterList's critical section.</p>
         <p>The abstract operation GetWaiterList takes two arguments, a Shared Data Block _block_ and a nonnegative integer _i_. It performs the following steps:</p>
         <emu-alg>
           1. Assert: _block_ is a Shared Data Block.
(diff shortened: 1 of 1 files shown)


ljharb:
@syg do you think this is something that can have test262 test coverage?

syg:
@ljharb Unfortunately not. :(

spec.html
```diff
@@ -36075,6 +36075,16 @@ <h1>EnterCriticalSection ( _WL_ )</h1>
...
           1. Wait until no agent is in the critical section for _WL_, then enter the critical section for _WL_ (without allowing any other agent to enter).
+          1. If there was an agent in the critical section for _WL_ prior to entry, then
+            1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
+            1. Let _eventsRecord_ be the Agent Events Record in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
+            1. Let _entererEventList_ be _eventsRecord_.[[EventList]].
+            1. Let _L_ be the Agent Signifier of the agent in the critical section immediately preceding the surrounding agent's entry.
+            1. Let _leaverEventList_ be the [[EventList]] field of the element in _execution_.[[EventRecords]] whose [[AgentSignifier]] is _L_.
+            1. Let _enterEvent_ and _leaveEvent_ be new Synchronize events.
+            1. Append _enterEvent_ to _entererEventList_.
+            1. Append _leaveEvent_ to _leaverEventList_.
```
conrad-watt: Morally, this will create a synchronization in the wrong place.

The leaver's (unpaired) synchronization event should be inserted into the event list at the moment of leaving the critical section, and with subsequent entries into the critical section creating the paired synchronization event. Otherwise, you allow executions where the leaver's synchronization event shows up in the agent-order _after_ other events which are program-order _before_ it (i.e. the notifying thread could have continued execution before the waiter gets a chance to re-enter the critical section). This is potentially un-observable, but it's un-intuitive and might cause inconsistencies later on if other features which expose operational interleavings are added.

I'm also a little iffy about the prose identification of the "immediately preceding" agent. This seems like the sort of thing that should be explicitly tracked as state in a semantic object.

The minimum patch would be to make the WaiterList keep a record of the last exiting synchronization event, updated with every LeaveCriticalSection, and synchronized on with every EnterCriticalSection.

spec.html
```diff
@@ -36144,15 +36154,6 @@ <h1>NotifyWaiter ( _WL_, _W_ )</h1>
         <emu-alg>
           1. Assert: The calling agent is in the critical section for _WL_.
           1. Assert: _W_ is on the list of waiters in _WL_.
-          1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
```
conrad-watt: Deleting these lines has an interesting interaction with the current Atomics.waitAsync proposal. Not something that should interfere overly with this PR though.

Currently, Atomics.wait will re-enter the critical section after being notified. This provides synchronization from the critical section exit of the previous Atomics.notify, and makes these lines redundant. However, if I'm reading the proposed spec correctly, Atomics.waitAsync does not enter the critical section after being notified, so deleting this section loses a synchronization edge that should probably be present, and will need to be added somewhere else in the proposal.

spec.html
syg: Thank you as always, @conrad-watt. These are all good suggestions, and I will try to incorporate them.

spec.html
syg: Yes, I realized this when writing this. Since it'll most likely fall on me to fix that spec text, I figured it's easiest to do the minimum delta for the current version of the spec so as to not raise "why is this here?" kind of questions, and fix it up in the waitAsync proposal.

lars-t-hansen:
Awaiting more code before reviewing, but really inclined to simply accept the judgement of @conrad-watt on this patch.

syg:
@conrad-watt Is this more or less what you had in mind?

spec.html
```diff
@@ -36057,9 +36057,10 @@ <h1>ValidateAtomicAccess ( _typedArray_, _requestIndex_ )</h1>
 
       <emu-clause id="sec-getwaiterlist" aoid="GetWaiterList">
         <h1>GetWaiterList ( _block_, _i_ )</h1>
-        <p>A <dfn>WaiterList</dfn> is a semantic object that contains an ordered list of those agents that are waiting on a location (_block_, _i_) in shared memory; _block_ is a Shared Data Block and _i_ a byte offset into the memory of _block_.</p>
+        <p>A <dfn>WaiterList</dfn> is a semantic object that contains an ordered list of those agents that are waiting on a location (_block_, _i_) in shared memory; _block_ is a Shared Data Block and _i_ a byte offset into the memory of _block_. A WaiterList object also optionally contains a Synchronize event denoting the previous leaving of its critical section.</p>
```
ljharb: I see the term "critical section" already in use, but there seems to be no definition of it in the spec. That doesn't have to block this PR, but it'd probably be good to define it in a separate PR, if not this one :-)

spec.html
```diff
@@ -36083,6 +36093,12 @@ <h1>LeaveCriticalSection ( _WL_ )</h1>
         <p>The abstract operation LeaveCriticalSection takes one argument, a WaiterList _WL_. It performs the following steps:</p>
         <emu-alg>
           1. Assert: The calling agent is in the critical section for _WL_.
+          1. Let _execution_ be the [[CandidateExecution]] field of the calling surrounding's Agent Record.
+          1. Let _eventsRecord_ be the Agent Events Record in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
+          1. Let _leaverEventList_ be _eventsRecord_.[[EventList]].
+          1. Let _leaveEvent_ be a new Synchronize event.
+          1. Append _leaveEvent_ to _leaverEventList_.
+          1. Set the Synchronize event in _WL_ to _leaveEvent_.
```
ljharb: Should this be an internal slot, instead of a vague concept of being "in" a WaiterList?

spec.html
syg: I'll do a separate editorial PR.

spec.html
syg: I don't recall exactly the history of why WaiterList is a "semantic object" instead of being a Record with fields. Perhaps it was to distinguish it as a special spec object that can be touched by different agents and needs to be locked by the critical section to be used? That way we can keep thinking of Records as objects that are intra-agent and don't need any extra care for use.

This also affects how I write the `Atomics.waitAsync` spec, since that extends the WaiterList to include more information that are accessed from multiple agents.

My preference is to keep it as a semantic object for this PR. I'm open to having some specialization of Record to distinguish WaiterList from other Records -- any thoughts there?

ljharb: I'm more confused by the whole concept of "semantic object" - the spec doesn't have anything else like that afaik.

It's fine to leave as-is, for the reasons you stated, but it seems like it'd be clearer to me in the future to have it be a proper Record; or to define some kind of more explicit syntax for interacting with "semantic objects" besides prose.

syg:
Rebased.

Issue #568, opened 2016-05-13
The '?' and '!' syntax in algorithms is very poorly chosen


bzbarsky:
There are several problems:

1. The choice to optimize for brevity at the expense of readability means the sigils are actually quite easy to miss.  Especially '!', which is quite thin.
2. '!' has existing meanings as "not" that confuse people (e.g. when people see `! IsCallable(obj)` in an algorithm they think that it means "obj is not callable".

Especially as people try to use these sigils in other specifications, with less-expert audiences than ecma262, this is causing real problems.  I see the value in using a single algorithm-definition language in ecma262 and other specifications that define things in terms of EcmaScript objects, but if we want to do that we should seriously consider a more understandable syntax.

As a specific suggestion, '!' could be replaced with "unbox" or something along those lines.  It would retain the clarity of extracting a value from the completion record, but make it even more clear that something nontrivial is going on, and for people who have never seen the syntax before make them stop and think about what "unbox" actually means instead of assuming that there's just a stray character in the spec draft, which is what they do now when they see the '?' and '!'.

bterlson:
"unbox" doesn't say what "!" does which is both an unboxing and an assertion that the completion record is never an abrupt completion. The latter is really the most important part.

I'm open to changing (and agree with point 2 above) but I'd need to see a good alternative. We did discuss alternatives for some time so I don't anticipate this will be an easy task :)

bzbarsky:
I'm not tied to "unbox", though it's common for "unbox" operations to assert that the unboxing is ok....

Is the existing discussion archived somewhere, in full or summary form?  What alternatives were considered?

bterlson:
#105 and #129 have some discussion. Not sure where the bulk of the syntax bike shedding happened, although much of it was IRC/elsewhere.

claudepache:
I have to make some effort in order to not read "!" as "not"...

The best I'm currently thinking of are "unbox-or-throw" and "nothrow-unbox".

ljharb:
`✓’?     ᕕ( ᐛ )ᕗ

allenwb:
! -> 👌
? -> 🤔

bzbarsky:
I recomend using something people might actually have the fonts for.... ;)

bterlson:
I am not against using Unicode for this in case folks are joking. It's not hard to set up your editor to make things easy to author and we have precedent (eg. &laquo; &raquo;).

[...]

PR #1693, opened 2019-09-10
Editorial: Typo—removed ! (logical NOT) from ToNumber()

singhArmani:
**Summary of the changes:** 
- Removed `!` from the definition of `Abstract Equality Comparison`. 

**Explanation**
It seems like a typo as it was correct in the [ECMAScript 2017 specs](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-abstract-equality-comparison). And then in rest of the specification onwards, it has this typo issue.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4521,10 +4521,10 @@
           1. Return the result of performing Strict Equality Comparison _x_ === _y_.
         1. If _x_ is *null* and _y_ is *undefined*, return *true*.
         1. If _x_ is *undefined* and _y_ is *null*, return *true*.
-        1. If Type(_x_) is Number and Type(_y_) is String, return the result of the comparison _x_ == ! ToNumber(_y_).
-        1. If Type(_x_) is String and Type(_y_) is Number, return the result of the comparison ! ToNumber(_x_) == _y_.
-        1. If Type(_x_) is Boolean, return the result of the comparison ! ToNumber(_x_) == _y_.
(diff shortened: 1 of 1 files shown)


shvaikalesh:
It is not a typo. Please see a description of `!` in [5.2.3.4ReturnIfAbrupt Shorthands](https://tc39.es/ecma262/#sec-returnifabrupt-shorthands):

> Similarly, prefix `!` is used to indicate that the following invocation of an abstract or syntax-directed operation will never return an abrupt completion and that the resulting Completion Record's `[[Value]]` field should be used in place of the return value of the operation.

loganfsmyth:
As mentioned, that is not a logical NOT. You can see the change that introduced this explicitly in https://github.com/tc39/ecma262/pull/921.

singhArmani:
> It is not a typo. Please see a description of `!` in [5.2.3.4 ReturnIfAbrupt Shorthands](https://tc39.es/ecma262/#sec-returnifabrupt-shorthands):
> 
> > Similarly, prefix `!` is used to indicate that the following invocation of an abstract or syntax-directed operation will never return an abrupt completion and that the resulting Completion Record's `[[Value]]` field should be used in place of the return value of the operation.

@shvaikalesh , @shvaikalesh : thank you guys. I entirely missed that part. Appreciate. I am closing this PR. 👏

claudepache:
See also Issue #568.

PR #1713, opened 2019-10-01
Normative: add RegExp Match Indices

rbuckton:
This adds the specification text from https://github.com/tc39/proposal-regexp-match-indices, currently at Stage 3.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -35030,7 +35030,10 @@ THH:mm:ss.sss
             A <em>CharSet</em> is a mathematical set of characters. When the _Unicode_ flag is *true*, &ldquo;all characters&rdquo; means the CharSet containing all code point values; otherwise &ldquo;all characters&rdquo; means the CharSet containing all code unit values.
           </li>
           <li>
-            A <em>State</em> is an ordered pair (_endIndex_, _captures_) where _endIndex_ is an integer and _captures_ is a List of _NcapturingParens_ values. States are used to represent partial match states in the regular expression matching algorithms. The _endIndex_ is one plus the index of the last input character matched so far by the pattern, while _captures_ holds the results of capturing parentheses. The _n_<sup>th</sup> element of _captures_ is either a List of characters that represents the value obtained by the _n_<sup>th</sup> set of capturing parentheses or *undefined* if the _n_<sup>th</sup> set of capturing parentheses hasn't been reached yet. Due to backtracking, many States may be in use at any time during the matching process.
+            A <dfn variants="Ranges">Range</dfn> is an ordered pair (_startIndex_, _endIndex_) that represents the range of characters included in a capture, where _startIndex_ is an integer representing the start index (inclusive) of the range within _Input_, and _endIndex_ is an integer representing the end index (exclusive) of the range within _Input_. For any Range, these indices must satisfy the invariant that _startIndex_ &le; _endIndex_.
+          </li>
+          <li>
+            A <em>State</em> is an ordered pair (_endIndex_, _captures_) where _endIndex_ is an integer and _captures_ is a List of _NcapturingParens_ values. States are used to represent partial match states in the regular expression matching algorithms. The _endIndex_ is one plus the index of the last input character matched so far by the pattern, while _captures_ holds the results of capturing parentheses. The _n_<sup>th</sup> element of _captures_ is either a Range representing the range of characters captured by the _n_<sup>th</sup> set of capturing parentheses, or *undefined* if the _n_<sup>th</sup> set of capturing parentheses hasn't been reached yet. Due to backtracking, many States may be in use at any time during the matching process.
           </li>
           <li>
             A <em>MatchResult</em> is either a State or the special token ~failure~ that indicates that the match failed.
@@ -35051,21 +35054,20 @@ THH:mm:ss.sss
         <emu-grammar>Pattern :: Disjunction</emu-grammar>
         <emu-alg>
           1. Let _m_ be CompileSubpattern of |Disjunction| with argument ~forward~.
-          1. Return a new Abstract Closure with parameters (_str_, _index_) that captures _m_ and performs the following steps when called:
-            1. Assert: Type(_str_) is String.
-            1. Assert: _index_ is a non-negative integer which is &le; the length of _str_.
-            1. If _Unicode_ is *true*, let _Input_ be StringToCodePoints(_str_). Otherwise, let _Input_ be a List whose elements are the code units that are the elements of _str_. _Input_ will be used throughout the algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref>. Each element of _Input_ is considered to be a character.
+          1. Return a new Abstract Closure with parameters (_inputChars_, _index_) that captures _m_ and performs the following steps when called:
+            1. Assert: _inputChars_ is a List of characters.
+            1. Assert: _index_ is a non-negative integer which is &le; the number of characters in _inputChars_.
+            1. Let _Input_ be _inputChars_. This alias will be used throughout the algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref>.
             1. Let _InputLength_ be the number of characters contained in _Input_. This alias will be used throughout the algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref>.
-            1. Let _listIndex_ be the index into _Input_ of the character that was obtained from element _index_ of _str_.
             1. Let _c_ be a new Continuation with parameters (_y_) that captures nothing and performs the following steps when called:
               1. Assert: _y_ is a State.
               1. Return _y_.
             1. Let _cap_ be a List of _NcapturingParens_ *undefined* values, indexed 1 through _NcapturingParens_.
-            1. Let _x_ be the State (_listIndex_, _cap_).
+            1. Let _x_ be the State (_index_, _cap_).
             1. Return _m_(_x_, _c_).
         </emu-alg>
         <emu-note>
-          <p>A Pattern compiles to an Abstract Closure value. RegExpBuiltinExec can then apply this procedure to a String and an offset within the String to determine whether the pattern would match starting at exactly that offset within the String, and, if it does match, what the values of the capturing parentheses would be. The algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref> are designed so that compiling a pattern may throw a *SyntaxError* exception; on the other hand, once the pattern is successfully compiled, applying the resulting Abstract Closure to find a match in a String cannot throw an exception (except for any implementation-defined exceptions that can occur anywhere such as out-of-memory).</p>
+          <p>A Pattern compiles to an Abstract Closure value. RegExpBuiltinExec can then apply this procedure to a List of characters and an offset within that List to determine whether the pattern would match starting at exactly that offset within the List, and, if it does match, what the values of the capturing parentheses would be. The algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref> are designed so that compiling a pattern may throw a *SyntaxError* exception; on the other hand, once the pattern is successfully compiled, applying the resulting Abstract Closure to find a match in a List of characters cannot throw an exception (except for any implementation-defined exceptions that can occur anywhere such as out-of-memory).</p>
         </emu-note>
       </emu-clause>
 
@@ -35468,12 +35470,12 @@ THH:mm:ss.sss
               1. Let _ye_ be _y_'s _endIndex_.
               1. If _direction_ is ~forward~, then
                 1. Assert: _xe_ &le; _ye_.
-                1. Let _s_ be a List whose elements are the characters of _Input_ at indices _xe_ (inclusive) through _ye_ (exclusive).
+                1. Let _r_ be the Range (_xe_, _ye_).
               1. Else,
                 1. Assert: _direction_ is ~backward~.
                 1. Assert: _ye_ &le; _xe_.
-                1. Let _s_ be a List whose elements are the characters of _Input_ at indices _ye_ (inclusive) through _xe_ (exclusive).
-              1. Set _cap_[_parenIndex_ + 1] to _s_.
+                1. Let _r_ be the Range (_ye_, _xe_).
+              1. Set _cap_[_parenIndex_ + 1] to _r_.
               1. Let _z_ be the State (_ye_, _cap_).
               1. Return _c_(_z_).
             1. Return _m_(_x_, _d_).
@@ -35558,15 +35560,17 @@ THH:mm:ss.sss
               1. Assert: _x_ is a State.
               1. Assert: _c_ is a Continuation.
               1. Let _cap_ be _x_'s _captures_ List.
-              1. Let _s_ be _cap_[_n_].
-              1. If _s_ is *undefined*, return _c_(_x_).
+              1. Let _r_ be _cap_[_n_].
+              1. If _r_ is *undefined*, return _c_(_x_).
               1. Let _e_ be _x_'s _endIndex_.
-              1. Let _len_ be the number of elements in _s_.
+              1. Let _rs_ be _r_'s _startIndex_.
+              1. Let _re_ be _r_'s _endIndex_.
+              1. Let _len_ be _re_ - _rs_.
               1. If _direction_ is ~forward~, let _f_ be _e_ + _len_.
               1. Else, let _f_ be _e_ - _len_.
               1. If _f_ &lt; 0 or _f_ &gt; _InputLength_, return ~failure~.
               1. Let _g_ be min(_e_, _f_).
-              1. If there exists an integer _i_ between 0 (inclusive) and _len_ (exclusive) such that Canonicalize(_s_[_i_]) is not the same character value as Canonicalize(_Input_[_g_ + _i_]), return ~failure~.
+              1. If there exists an integer _i_ between 0 (inclusive) and _len_ (exclusive) such that Canonicalize(_Input_[_rs_ + _i_]) is not the same character value as Canonicalize(_Input_[_g_ + _i_]), return ~failure~.
               1. Let _y_ be the State (_f_, _cap_).
               1. Return _c_(_y_).
           </emu-alg>
@@ -35922,7 +35926,7 @@ THH:mm:ss.sss
             1. Else, let _P_ be ? ToString(_pattern_).
             1. If _flags_ is *undefined*, let _F_ be the empty String.
             1. Else, let _F_ be ? ToString(_flags_).
-            1. If _F_ contains any code unit other than *"g"*, *"i"*, *"m"*, *"s"*, *"u"*, or *"y"* or if it contains the same code unit more than once, throw a *SyntaxError* exception.
+            1. If _F_ contains any code unit other than *"d"*, *"g"*, *"i"*, *"m"*, *"s"*, *"u"*, or *"y"* or if it contains the same code unit more than once, throw a *SyntaxError* exception.
             1. If _F_ contains *"u"*, let _u_ be *true*; else let _u_ be *false*.
             1. If _u_ is *true*, then
               1. Let _patternText_ be StringToCodePoints(_P_).
@@ -36087,16 +36091,20 @@ THH:mm:ss.sss
             1. Let _flags_ be _R_.[[OriginalFlags]].
             1. If _flags_ contains *"g"*, let _global_ be *true*; else let _global_ be *false*.
             1. If _flags_ contains *"y"*, let _sticky_ be *true*; else let _sticky_ be *false*.
+            1. If _flags_ contains *"d"*, let _hasIndices_ be *true*; else let _hasIndices_ be *false*.
             1. If _global_ is *false* and _sticky_ is *false*, set _lastIndex_ to 0.
             1. Let _matcher_ be _R_.[[RegExpMatcher]].
             1. If _flags_ contains *"u"*, let _fullUnicode_ be *true*; else let _fullUnicode_ be *false*.
             1. Let _matchSucceeded_ be *false*.
+            1. If _fullUnicode_ is *true*, let _input_ be StringToCodePoints(_S_). Otherwise, let _input_ be a List whose elements are the code units that are the elements of _S_.
+            1. NOTE: Each element of _input_ is considered to be a character.
             1. Repeat, while _matchSucceeded_ is *false*,
               1. If _lastIndex_ &gt; _length_, then
                 1. If _global_ is *true* or _sticky_ is *true*, then
                   1. Perform ? Set(_R_, *"lastIndex"*, *+0*<sub>𝔽</sub>, *true*).
                 1. Return *null*.
-              1. Let _r_ be _matcher_(_S_, _lastIndex_).
+              1. Let _inputIndex_ be the index into _input_ of the character that was obtained from element _lastIndex_ of _S_.
+              1. Let _r_ be _matcher_(_input_, _inputIndex_).
               1. If _r_ is ~failure~, then
                 1. If _sticky_ is *true*, then
                   1. Perform ? Set(_R_, *"lastIndex"*, *+0*<sub>𝔽</sub>, *true*).
@@ -36106,9 +36114,7 @@ THH:mm:ss.sss
                 1. Assert: _r_ is a State.
                 1. Set _matchSucceeded_ to *true*.
             1. Let _e_ be _r_'s _endIndex_ value.
-            1. If _fullUnicode_ is *true*, then
-              1. _e_ is an index into the _Input_ character list, derived from _S_, matched by _matcher_. Let _eUTF_ be the smallest index into _S_ that corresponds to the character at element _e_ of _Input_. If _e_ is greater than or equal to the number of elements in _Input_, then _eUTF_ is the number of code units in _S_.
-              1. Set _e_ to _eUTF_.
+            1. If _fullUnicode_ is *true*, set _e_ to GetStringIndex(_S_, _e_).
             1. If _global_ is *true* or _sticky_ is *true*, then
               1. Perform ? Set(_R_, *"lastIndex"*, 𝔽(_e_), *true*).
             1. Let _n_ be the number of elements in _r_'s _captures_ List. (This is the same value as <emu-xref href="#sec-notation"></emu-xref>'s _NcapturingParens_.)
@@ -36117,27 +36123,43 @@ THH:mm:ss.sss
             1. Assert: The mathematical value of _A_'s *"length"* property is _n_ + 1.
             1. Perform ! CreateDataPropertyOrThrow(_A_, *"index"*, 𝔽(_lastIndex_)).
             1. Perform ! CreateDataPropertyOrThrow(_A_, *"input"*, _S_).
-            1. Let _matchedSubstr_ be the substring of _S_ from _lastIndex_ to _e_.
+            1. Let _match_ be the Match Record { [[StartIndex]]: _lastIndex_, [[EndIndex]]: _e_ }.
+            1. Let _indices_ be a new empty List.
+            1. Let _groupNames_ be a new empty List.
+            1. Append _match_ to _indices_.
+            1. Let _matchedSubstr_ be GetMatchString(_S_, _match_).
             1. Perform ! CreateDataPropertyOrThrow(_A_, *"0"*, _matchedSubstr_).
             1. If _R_ contains any |GroupName|, then
               1. Let _groups_ be OrdinaryObjectCreate(*null*).
+              1. Let _hasGroups_ be *true*.
             1. Else,
               1. Let _groups_ be *undefined*.
+              1. Let _hasGroups_ be *false*.
             1. Perform ! CreateDataPropertyOrThrow(_A_, *"groups"*, _groups_).
             1. For each integer _i_ such that _i_ &ge; 1 and _i_ &le; _n_, in ascending order, do
               1. Let _captureI_ be _i_<sup>th</sup> element of _r_'s _captures_ List.
-              1. If _captureI_ is *undefined*, let _capturedValue_ be *undefined*.
-              1. Else if _fullUnicode_ is *true*, then
-                1. Assert: _captureI_ is a List of code points.
-                1. Let _capturedValue_ be CodePointsToString(_captureI_).
+              1. If _captureI_ is *undefined*, then
+                1. Let _capturedValue_ be *undefined*.
+                1. Append *undefined* to _indices_.
               1. Else,
-                1. Assert: _fullUnicode_ is *false*.
-                1. Assert: _captureI_ is a List of code units.
-                1. Let _capturedValue_ be the String value consisting of the code units of _captureI_.
+                1. Let _captureStart_ be _captureI_'s _startIndex_.
+                1. Let _captureEnd_ be _captureI_'s _endIndex_.
+                1. If _fullUnicode_ is *true*, then
+                  1. Set _captureStart_ to GetStringIndex(_S_, _captureStart_).
+                  1. Set _captureEnd_ to GetStringIndex(_S_, _captureEnd_).
+                1. Let _capture_ be the Match Record { [[StartIndex]]: _captureStart_, [[EndIndex]]: _captureEnd_ }.
+                1. Let _capturedValue_ be GetMatchString(_S_, _capture_).
+                1. Append _capture_ to _indices_.
               1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(𝔽(_i_)), _capturedValue_).
               1. If the _i_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
                 1. Let _s_ be the CapturingGroupName of the corresponding |RegExpIdentifierName|.
                 1. Perform ! CreateDataPropertyOrThrow(_groups_, _s_, _capturedValue_).
+                1. Append _s_ to _groupNames_.
+              1. Else,
+                1. Append *undefined* to _groupNames_.
+            1. If _hasIndices_ is *true*, then
+              1. Let _indicesArray_ be MakeMatchIndicesIndexPairArray(_S_, _indices_, _groupNames_, _hasGroups_).
+              1. Perform ! CreateDataPropertyOrThrow(_A_, *"indices"*, _indicesArray_).
             1. Return _A_.
           </emu-alg>
         </emu-clause>
@@ -36159,6 +36181,115 @@ THH:mm:ss.sss
             1. If _index_ + 1 &ge; _length_, return _index_ + 1.
             1. Let _cp_ be CodePointAt(_S_, _index_).
             1. Return _index_ + _cp_.[[CodeUnitCount]].
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-getstringindex" type="abstract operation">
+          <h1>
+            GetStringIndex (
+              _S_: a String,
+              _e_: a non-negative integer,
+            ): a non-negative integer
+          </h1>
+          <dl class="header">
+          </dl>
+          <emu-alg>
+            1. If _S_ is the empty String, return 0.
+            1. Let _codepoints_ be StringToCodePoints(_S_).
+            1. Let _eUTF_ be the smallest index into _S_ that corresponds to the character at element _e_ of _codepoints_. If _e_ is greater than or equal to the number of elements in _codepoints_, then _eUTF_ is the number of code units in _S_.
+            1. Return _eUTF_.
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-match-records">
+          <h1>Match Records</h1>
+          <p>A <dfn variants="Match Records">Match Record</dfn> is a Record value used to encapsulate the start and end indices of a regular expression match or capture.</p>
+          <p>Match Records have the fields listed in <emu-xref href="#table-match-record"></emu-xref>.</p>
+          <emu-table id="table-match-record" caption="Match Record Fields">
+            <table>
+              <tr>
+                <th>Field Name</th>
+                <th>Value</th>
+                <th>Meaning</th>
+              </tr>
+              <tr>
+                <td>[[StartIndex]]</td>
+                <td>a non-negative integer</td>
+                <td>The number of code units from the start of a string at which the match begins (inclusive).</td>
+              </tr>
+              <tr>
+                <td>[[EndIndex]]</td>
+                <td>an integer &ge; [[StartIndex]]</td>
+                <td>The number of code units from the start of a string at which the match ends (exclusive).</td>
+              </tr>
+            </table>
+          </emu-table>
+        </emu-clause>
+
+        <emu-clause id="sec-getmatchstring" type="abstract operation">
+          <h1>
+            GetMatchString (
+              _S_: a String,
+              _match_: a Match Record,
+            ): a String
+          </h1>
+          <dl class="header">
+          </dl>
+          <emu-alg>
+            1. Assert: _match_.[[StartIndex]] is a non-negative integer less than or equal to the length of _S_.
+            1. Assert: _match_.[[EndIndex]] is an integer between _match_.[[StartIndex]] and the length of _S_, inclusive.
+            1. Return the substring of _S_ from _match_.[[StartIndex]] to _match_.[[EndIndex]].
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-getmatchindexpair" type="abstract operation">
+          <h1>
+            GetMatchIndexPair (
+              _S_: a String,
+              _match_: a Match Record,
+            ): an Array
+          </h1>
+          <dl class="header">
+          </dl>
+          <emu-alg>
+            1. Assert: _match_.[[StartIndex]] is a non-negative integer less than or equal to the length of _S_.
+            1. Assert: _match_.[[EndIndex]] is an integer between _match_.[[StartIndex]] and the length of _S_, inclusive.
+            1. Return CreateArrayFromList(&laquo; 𝔽(_match_.[[StartIndex]]), 𝔽(_match_.[[EndIndex]]) &raquo;).
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-makematchindicesindexpairarray" type="abstract operation">
+          <h1>
+            MakeMatchIndicesIndexPairArray (
+              _S_: a String,
+              _indices_: a List of either Match Records or *undefined*,
+              _groupNames_: a List of either Strings or *undefined*,
+              _hasGroups_: a Boolean,
+            ): an Array
+          </h1>
+          <dl class="header">
+          </dl>
+          <emu-alg>
+            1. Let _n_ be the number of elements in _indices_.
+            1. Assert: _n_ &lt; 2<sup>32</sup> - 1.
+            1. Assert: _groupNames_ has _n_ - 1 elements.
+            1. NOTE: The _groupNames_ List contains elements aligned with the _indices_ List starting at _indices_[1].
+            1. Let _A_ be ! ArrayCreate(_n_).
+            1. If _hasGroups_ is *true*, then
+              1. Let _groups_ be OrdinaryObjectCreate(*null*).
+            1. Else,
+              1. Let _groups_ be *undefined*.
+            1. Perform ! CreateDataPropertyOrThrow(_A_, *"groups"*, _groups_).
+            1. For each integer _i_ starting with 0 such that _i_ &lt; _n_, in ascending order, do
+              1. Let _matchIndices_ be _indices_[_i_].
+              1. If _matchIndices_ is not *undefined*, then
+                1. Let _matchIndexPair_ be GetMatchIndexPair(_S_, _matchIndices_).
+              1. Else,
+                1. Let _matchIndexPair_ be *undefined*.
+              1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(𝔽(_i_)), _matchIndexPair_).
+              1. If _i_ &gt; 0 and _groupNames_[_i_ - 1] is not *undefined*, then
+                1. Perform ! CreateDataPropertyOrThrow(_groups_, _groupNames_[_i_ - 1], _matchIndexPair_).
+            1. Return _A_.
           </emu-alg>
         </emu-clause>
       </emu-clause>
@@ -36200,6 +36331,8 @@ THH:mm:ss.sss
           1. Let _R_ be the *this* value.
           1. If Type(_R_) is not Object, throw a *TypeError* exception.
           1. Let _result_ be the empty String.
+          1. Let _hasIndices_ be ToBoolean(? Get(_R_, *"hasIndices"*)).
+          1. If _hasIndices_ is *true*, append the code unit 0x0064 (LATIN SMALL LETTER D) as the last code unit of _result_.
           1. Let _global_ be ToBoolean(? Get(_R_, *"global"*)).
           1. If _global_ is *true*, append the code unit 0x0067 (LATIN SMALL LETTER G) as the last code unit of _result_.
           1. Let _ignoreCase_ be ToBoolean(? Get(_R_, *"ignoreCase"*)).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -32203,6 +32252,71 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
...
+  
+        <emu-clause id="sec-getmatchindicesarray" aoid="GetMatchIndicesArray">
+          <h1>GetMatchIndicesArray ( _S_, _match_ )</h1>
+          <p>The abstract operation GetMatchIndicesArray with arguments _S_ and _match_ performs the following steps:</p>
+          <emu-alg>
+            1. Assert: Type(_S_) is String.
+            1. Assert: _match_ is a Match Record.
+            1. Assert: _match_.[[StartIndex]] is an integer value &ge; 0 and &lt; the length of _S_.
+            1. Assert: _match_.[[EndIndex]] is an integer value &ge; _match_.[[StartIndex]] and &le; the length of _S_.
+            1. Return CreateArrayFromList(&laquo; _match_.[[StartIndex]], _match_.[[EndIndex]] &raquo;).
```
```suggestion
            1. Return ! CreateArrayFromList(&laquo; _match_.[[StartIndex]], _match_.[[EndIndex]] &raquo;).
```

spec.html
```diff
@@ -32203,6 +32252,71 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
...
+            1. Assert: _groupNames_ is a List or is *undefined*.
+            1. Let _n_ be the number of elements in _indices_.
+            1. Assert: _n_ &lt; 2<sup>32</sup>-1.
+            1. Set _A_ to ! ArrayCreate(_n_).
+            1. Assert: The value of _A_'s `"length"` property is _n_.
+            1. If _groupNames_ is not *undefined*, then
+              1. Let _groups_ be ! ObjectCreate(*null*).
+            1. Else,
+              1. Let _groups_ be *undefined*.
+            1. Perform ! CreateDataProperty(_A_, `"groups"`, _groups_).
```
```suggestion
            1. Perform ! CreateDataPropertyOrThrow(_A_, `"groups"`, _groups_).
```

spec.html
```diff
@@ -32203,6 +32252,71 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
...
+            1. For each integer _i_ such that _i_ &ge; 0 and _i_ &lt; _n_, do
+              1. Let _matchIndices_ be _indices_[_i_].
+              1. If _matchIndices_ is not *undefined*, then
+                1. Let _matchIndicesArray_ be ! GetMatchIndicesArray(_S_, _matchIndices_).
+              1. Else,
+                1. Let _matchIndicesArray_ be *undefined*.
+              1. Perform ! CreateDataProperty(_A_, ! ToString(_i_), _matchIndicesArray_).
```
```suggestion
              1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(_i_), _matchIndicesArray_).
```

spec.html
```diff
@@ -30941,7 +30941,10 @@ <h1>Notation</h1>
...
+          </li>
+          <li>
+            A <em>State</em> is an ordered pair (_endIndex_, _captures_) where _endIndex_ is an integer and _captures_ is a List of _NcapturingParens_ values. States are used to represent partial match states in the regular expression matching algorithms. The _endIndex_ is one plus the index of the last input character matched so far by the pattern, while _captures_ holds the results of capturing parentheses. The _n_<sup>th</sup> element of _captures_ is either a List that represents the <em>Range</em> obtained by the _n_<sup>th</sup> set of capturing parentheses or *undefined* if the _n_<sup>th</sup> set of capturing parentheses hasn't been reached yet. Due to backtracking, many States may be in use at any time during the matching process.
```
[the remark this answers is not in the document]
rbuckton: _NcapturingParens_ is the name of a variable, per https://tc39.es/ecma262/#sec-notation.

spec.html
```diff
@@ -32164,27 +32198,42 @@ <h1>Runtime Semantics: RegExpBuiltinExec ( _R_, _S_ )</h1>
...
             1. Perform ! CreateDataProperty(_A_, `"index"`, _lastIndex_).
             1. Perform ! CreateDataProperty(_A_, `"input"`, _S_).
-            1. Let _matchedSubstr_ be the matched substring (i.e. the portion of _S_ between offset _lastIndex_ inclusive and offset _e_ exclusive).
-            1. Perform ! CreateDataProperty(_A_, `"0"`, _matchedSubstr_).
-            1. If _R_ contains any |GroupName|, then
+            1. Let _indices_ be a new empty List.
+            1. Let _match_ be the Match { [[StartIndex]]: _lastIndex_, [[EndIndex]]: _e_ }.
+            1. Add _match_ as the last element of _indices_.
+            1. Let _matchedValue_ be ! GetMatchString(_S_, _match_).
+            1. Perform ! CreateDataProperty(_A_, `"0"`, _matchedValue_).
```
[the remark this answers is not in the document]
rbuckton: Is this to align with https://github.com/tc39/ecma262/pull/1676?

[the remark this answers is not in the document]
rbuckton: I've reverted this line (and changing the name of the variable on the line above to match), since it is likely already addressed by #1676 and will be one less merge conflict.

rbuckton:
Test262 Tests have already been merged: https://github.com/tc39/test262/pull/2309

spec.html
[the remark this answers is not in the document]
ljharb: ah, thanks, what a confusing variable name

jmdyck:
I've submitted an editorial [PR](https://github.com/rbuckton/ecma262/pull/1) against this PR's branch.

rbuckton:
NOTE: I still need to update the specification text based on the changes in https://github.com/tc39/proposal-regexp-match-indices/pull/49

jmdyck (changes requested):

rbuckton:
@ljharb, I know you approved this prior to the addition of the `d` flag. Can you take another look to make sure I haven't missed any thing editorially in the process of updating this PR?

spec.html
```diff
@@ -31047,6 +31067,101 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
...
+
+        <emu-clause id="sec-getmatchstring" aoid="GetMatchString">
+          <h1>GetMatchString ( _S_, _match_ )</h1>
+          <p>The abstract operation GetMatchString with arguments _S_ and _match_ performs the following steps:</p>
+          <emu-alg>
+            1. Assert: Type(_S_) is String.
+            1. Assert: _match_ is a Match Record.
+            1. Assert: _match_.[[StartIndex]] is an integer value &ge; 0 and &le; the length of _S_.
+            1. Assert: _match_.[[EndIndex]] is an integer value &ge; _match_.[[StartIndex]] and &le; the length of _S_.
+            1. Return the portion of _S_ between offset _match_.[[StartIndex]] inclusive and offset _match_.[[EndIndex]] exclusive.
```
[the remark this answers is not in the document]
rbuckton: Does the term "substring" imply that the start is inclusive and end is exclusive, or are we just relying on the description in https://arai-a.github.io/ecma262-compare/history/PR/1713/48224e631a962e8a86c613bb50daff4d280a9b1a/index.html#sec-match-records?

spec.html
```diff
@@ -31047,6 +31067,101 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
...
+
+        <emu-clause id="sec-getmatchindicesarray" aoid="GetMatchIndicesArray">
+          <h1>GetMatchIndicesArray ( _S_, _match_ )</h1>
+          <p>The abstract operation GetMatchIndicesArray with arguments _S_ and _match_ performs the following steps:</p>
+          <emu-alg>
+            1. Assert: Type(_S_) is String.
+            1. Assert: _match_ is a Match Record.
+            1. Assert: _match_.[[StartIndex]] is an integer value &ge; 0 and &lt; the length of _S_.
+            1. Assert: _match_.[[EndIndex]] is an integer value &ge; _match_.[[StartIndex]] and &le; the length of _S_.
+            1. Return ! CreateArrayFromList(&laquo; _match_.[[StartIndex]], _match_.[[EndIndex]] &raquo;).
```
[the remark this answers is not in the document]
rbuckton: The issue with characters like `𝔽` is that they are UTF-8, but the spec does not include the UTF-8 preamble so some tools like Beyond Compare show it as garbled ASCII unless you explicitly set the encoding for the document. Apparently Beyond Compare won't interpret the `<meta charset="utf-8" >` at the top of the file. I had assumed that's why we use `&laquo;` instead of &laquo;. Although, I suppose `𝔽` is easier to read in source than `&#120125;`.

spec.html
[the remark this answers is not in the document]
jmdyck: > Does the term "substring" imply that the start is inclusive and end is exclusive,

Yes, see https://tc39.es/ecma262/#substring (introduced by PR #2021).

spec.html
```diff
@@ -31010,34 +31014,50 @@ <h1>RegExpBuiltinExec ( _R_, _S_ )</h1>
...
+              1. Let _indicesArray_ be ! MakeIndicesArray(_S_, _indices_, _groupNames_, _hasGroups_).
+              1. Perform ! CreateDataPropertyOrThrow(_A_, *"indices"*, _indicesArray_).
             1. Return _A_.
           </emu-alg>
         </emu-clause>
 
         <emu-clause id="sec-advancestringindex" aoid="AdvanceStringIndex">
           <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
-          <p>The abstract operation AdvanceStringIndex takes arguments _S_ (a String), _index_ (a non-negative integer), and _unicode_ (a Boolean). It performs the following steps when called:</p>
+          <p>The abstract operation GetStringIndex takes arguments _S_, _Input_, and _e_. It performs the following steps when called:</p>
```
jmdyck, line 31060:
```suggestion
          <p>The abstract operation AdvanceStringIndex takes arguments _S_ (a String), _index_ (a non-negative integer), and _unicode_ (a Boolean). It performs the following steps when called:</p>
```

rbuckton: Oops, thanks

spec.html
```diff
@@ -31047,6 +31067,101 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
...
+            </table>
+          </emu-table>
+        </emu-clause>
+
+        <emu-clause id="sec-getmatchstring" aoid="GetMatchString">
+          <h1>GetMatchString ( _S_, _match_ )</h1>
+          <p>The abstract operation GetMatchString takes arguments _S_ and _match_. It performs the following steps when called:</p>
+          <emu-alg>
+            1. Assert: Type(_S_) is String.
+            1. Assert: _match_ is a Match Record.
```
ljharb, line 31116:
```suggestion
          <p>The abstract operation GetMatchString takes arguments _S_ (a String) and _match_ (a Match Record). It performs the following steps when called:</p>
          <emu-alg>
```

spec.html
```diff
@@ -31047,6 +31067,101 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
...
+            1. Return the substring of _S_ from _match_.[[StartIndex]] to _match_.[[EndIndex]].
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-getmatchindicesarray" aoid="GetMatchIndicesArray">
+          <h1>GetMatchIndicesArray ( _S_, _match_ )</h1>
+          <p>The abstract operation GetMatchIndicesArray takes arguments _S_ and _match_. It performs the following steps when called:</p>
+          <emu-alg>
+            1. Assert: Type(_S_) is String.
+            1. Assert: _match_ is a Match Record.
```
ljharb, line 31128:
```suggestion
          <p>The abstract operation GetMatchIndicesArray takes arguments _S_ (a String) and _match_ (a Match Record). It performs the following steps when called:</p>
          <emu-alg>
```

spec.html
```diff
@@ -31047,6 +31067,101 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
...
+            1. Return ! CreateArrayFromList(&laquo; 𝔽(_match_.[[StartIndex]]), 𝔽(_match_.[[EndIndex]]) &raquo;).
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-makeindicesarray" aoid="MakeIndicesArray">
+          <h1>MakeIndicesArray ( _S_, _indices_, _groupNames_, _hasGroups_ )</h1>
+          <p>The abstract operation MakeIndicesArray takes arguments _S_, _indices_, _groupNames_, and _hasGroups_. It performs the following steps when called:</p>
+          <emu-alg>
+            1. Assert: Type(_S_) is String.
+            1. Assert: _indices_ is a List.
```
ljharb, line 31140:
```suggestion
          <p>The abstract operation MakeIndicesArray takes arguments _S_ (a String), _indices_ (a List), _groupNames_ (a List), and _hasGroups_ (a Boolean). It performs the following steps when called:</p>
          <emu-alg>
```

spec.html
```diff
@@ -31047,6 +31067,101 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
...
+            1. Let _n_ be the number of elements in _indices_.
+            1. Assert: _n_ &lt; 2<sup>32</sup> - 1.
+            1. Assert: _groupNames_ is a List with _n_ - 1 elements.
+            1. NOTE: The _groupNames_ List contains elements aligned with the _indices_ List starting at _indices_[1].
+            1. Assert: Type(_hasGroups_) is Boolean.
```
ljharb, line 31145:
```suggestion
            1. Assert: _groupNames_ has _n_ - 1 elements.
            1. NOTE: The _groupNames_ List contains elements aligned with the _indices_ List starting at _indices_[1].
```

ljharb (approved):
still LGTM

spec.html
```diff
@@ -31134,15 +31129,12 @@ <h1>GetMatchIndicesArray ( _S_, _match_ )</h1>
...
           <h1>MakeIndicesArray ( _S_, _indices_, _groupNames_, _hasGroups_ )</h1>
-          <p>The abstract operation MakeIndicesArray takes arguments _S_, _indices_, _groupNames_, and _hasGroups_. It performs the following steps when called:</p>
+          <p>The abstract operation MakeIndicesArray takes arguments _S_ (a String), _indices_ (a List), _groupNames_ (a List), and _hasGroups_ (a Boolean). It performs the following steps when called:</p>
           <emu-alg>
-            1. Assert: Type(_S_) is String.
-            1. Assert: _indices_ is a List.
             1. Let _n_ be the number of elements in _indices_.
             1. Assert: _n_ &lt; 2<sup>32</sup> - 1.
-            1. Assert: _groupNames_ is a List with _n_ - 1 elements.
+            1. Assert: _groupNames_ with _n_ - 1 elements.
```
jmdyck, line 31136:
```suggestion
            1. Assert: _groupNames_ has _n_ - 1 elements.
```
?

rbuckton: For whatever reason, github is not allowing me to commit suggestions. I'll push a new commit shortly.

ljharb: whenever the suggester has write access, github doesn't allow you to commit the suggestion. it's a long-standing bug.

jmdyck (changes requested):

spec.html
```diff
@@ -31047,6 +31067,93 @@ <h1>AdvanceStringIndex ( _S_, _index_, _unicode_ )</h1>
             1. Return _index_ + _cp_.[[CodeUnitCount]].
           </emu-alg>
         </emu-clause>
+
+        <emu-clause id="sec-getstringindex" aoid="GetStringIndex">
+          <h1>GetStringIndex ( _S_, _Input_, _e_ )</h1>
+          <p>The abstract operation GetStringIndex takes arguments _S_ (a String), _Input_ (a List), and _e_. It performs the following steps when called:</p>
+          <emu-alg>
+            1. Assert: _Input_ contains the code points of _S_ interpreted as a UTF-16 encoded string.
```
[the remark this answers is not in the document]
rbuckton: IIRC, this was pretty much a direct copy of spec text at the time the proposal was drafted, but the underlying spec text has changed since. The ES2016 spec has the following text in [`sec-pattern`](https://262.ecma-international.org/7.0/#sec-pattern):

```
sequence of code points of _str_ interpreted as a UTF-16 encoded (<emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>) Unicode string
```

I appear to have dropped the word `Unicode` in my transcription. That said, if the editors believe this is a better interpretation I'm perfectly fine with it.

ljharb:
@rbuckton would you mind rebasing this? it'd be great to get it in for ES2022.

rbuckton:
I've rebased and updated the new algorithms to use structured headers. Please let me know if I missed anything.

spec.html
```diff
@@ -36114,6 +36247,21 @@ <h1>get RegExp.prototype.global</h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-get-regexp.prototype.hasIndices">
+        <h1>get RegExp.prototype.hasIndices</h1>
+        <p>`RegExp.prototype.hasIndices` is an accessor property whose set accessor function is *undefined*. Its get accessor function performs the following steps:</p>
+        <emu-alg>
+          1. Let _R_ be the *this* value.
+          1. If Type(_R_) is not Object, throw a *TypeError* exception.
+          1. If _R_ does not have an [[OriginalFlags]] internal slot, then
```
ljharb, line 36256: I think this now needs to call into RegExpHasFlag, like the other accessors have been updated to do?

spec.html
```diff
@@ -35994,9 +36000,7 @@ <h1>
                 1. Assert: _r_ is a State.
                 1. Set _matchSucceeded_ to *true*.
             1. Let _e_ be _r_'s _endIndex_ value.
-            1. If _fullUnicode_ is *true*, then
-              1. _e_ is an index into the _Input_ character list, derived from _S_, matched by _matcher_. Let _eUTF_ be the smallest index into _S_ that corresponds to the character at element _e_ of _Input_. If _e_ is greater than or equal to the number of elements in _Input_, then _eUTF_ is the number of code units in _S_.
-              1. Set _e_ to _eUTF_.
+            1. If _fullUnicode_ is *true*, set _e_ to ! GetStringIndex(_S_, _Input_, _e_).
```
ljharb, line 36003: I don’t think Inout is defined anywhere now - and it seems like it should be derived from S inside StringGetIndex rather than passing it in?

rbuckton: When was that changed?

spec.html
```diff
@@ -36049,6 +36069,117 @@ <h1>
...
+            GetStringIndex (
+              _S_: a String,
+              _Input_: a List,
+              _e_: a non-negative integer,
+            )
+          </h1>
+          <dl class="header">
+          </dl>
+          <emu-alg>
+            1. Assert: _Input_ contains the code points of _S_ interpreted as a UTF-16 encoded string.
```
jmdyck, line 36084:
```suggestion
            1. Assert: _Input_ contains the code points of StringToCodePoints(_S_).
```

spec.html
```diff
@@ -36049,6 +36069,117 @@ <h1>
...
+          <emu-table id="table-match-record" caption="Match Record Fields">
+            <table>
+              <tr>
+                <th>Field Name</th>
+                <th>Value</th>
+                <th>Meaning</th>
+              </tr>
+              <tr>
+                <td>[[StartIndex]]</td>
+                <td>An integer &ge; 0.</td>
```
jmdyck, line 36104:
```suggestion
                <td>a non-negative integer</td>
```

spec.html
```diff
@@ -36049,6 +36069,117 @@ <h1>
...
+                <td>The number of code units from the start of a string at which the match begins (inclusive).</td>
+              </tr>
+              <tr>
+                <td>[[EndIndex]]</td>
+                <td>An integer &ge; [[StartIndex]].</td>
```
jmdyck, line 36109:
```suggestion
                <td>an integer &ge; [[StartIndex]]</td>
```

spec.html
```diff
@@ -36049,6 +36069,117 @@ <h1>
...
+            1. Assert: _match_.[[EndIndex]] is an integer value &ge; _match_.[[StartIndex]] and &le; the length of _S_.
+            1. Return ! CreateArrayFromList(&laquo; 𝔽(_match_.[[StartIndex]]), 𝔽(_match_.[[EndIndex]]) &raquo;).
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-makeindicesarray" type="abstract operation">
+          <h1>
+            MakeIndicesArray (
+              _S_: a String,
+              _indices_: a List,
```
jmdyck, line 36152:
```suggestion
              _indices_: a List, each of whose elements is a Match Record or *undefined*,
```

spec.html
```diff
@@ -36049,6 +36069,117 @@ <h1>
...
+              _groupNames_: a List or *undefined*,
```
jmdyck, line 36153: `_groupNames_` can't be `*undefined*`: step 3 asserts that it has elements.

```suggestion
              _groupNames_: a List, each of whose elements is a String or *undefined*,
```

jmdyck (changes requested):

spec.html
rbuckton: Isn't it defined in https://tc39.es/ecma262/#sec-pattern-semantics?

spec.html
rbuckton: Nice catch, I accidentally overwrote that in the rebase.

spec.html
ljharb: I think it’s defined there for the *other* places you use it, but i think this one doesn’t use that, which is why the existing prose defines it inline.

rbuckton: Would it be sufficient to change the algorithm steps of GetStringIndex to:

```
1. If _S_ is the empty String, return 0.
1. Let _Input_ be StringToCodePoints(_S_).
1. Let _eUTF_ be the smallest index into _S_ that corresponds to the character at element _e_ of _input_. If _e_ is greater than or equal to the number of elements in _input_, then _eUTF_ is the number of code units in _S_.
1. Return _eUTF_.
```

and let implementations optimize away the repeated invocations of `StringToCodePoints(_S_)`? It feels very hand-wavy (then again, so did the previous wording this replaces).

Something less hand-wavy might have created a List of code points and a List of the original index in `_S_` at which each code point was found, and passed them both to `GetStringIndex` (or a single list containing tuples of `(codePoint, index)`).

rbuckton: Or should `GetStringIndex` just contain a loop that repeated calls `AdvanceStringIndex` until enough code points have been consumed?

ljharb: i mean, there's also `StringIndexOf` already, so it's possible nothing new is needed here.

In particular it just seems silly to pass around `S` and `Input` when both are, via assertion, always matching and derivable from each other, and when the spec isn't a programming language so implementations can trivially optimize it (meaning, we shouldn't be worried about the perf of the spec algorithm, only about the perf of resulting implementations)

rbuckton: > I think it’s defined there for the other places you use it, but i think this one doesn’t use that, which is why the existing prose defines it inline.

The original prose doesn't explain how _Input_ is derived from _S_, only that it **is** derived from _S_. https://tc39.es/ecma262/#sec-pattern-semantics is the only place that describes how _Input_ is to be derived.

rbuckton: _e_ refers to the position in _Input_ per the pattern semantics. The point of this operation is to convert the ordinal index of _e_ in _Input_ to an actual string index in _S_.

rbuckton: Part of the problem with this section is that _Input_ is defined in https://tc39.es/ecma262/#sec-pattern-semantics, and initialized in https://tc39.es/ecma262/#sec-compilepattern, step 2.c, and is "used throughout the algorithms in 22.2.2", but isn't defined as an output of the matcher created by CompilePattern, so RegExpBuiltinExec has always been referencing something not exactly "reachable" by the algorithm.

rbuckton: > i mean, there's also `StringIndexOf` already, so it's possible nothing new is needed here.

StringIndexOf doesn't do anything like what this is trying to do.

rbuckton: The simplest thing may be just to revert this line, duplicate the original steps for _captureStart_ and _captureEnd_, and drop GetStringIndex altogether. We could then rework RegExpBuiltinExec and the Regular expression pattern semantics later.

spec.html
```diff
@@ -36005,27 +36009,43 @@ <h1>
...
+                1. Append *undefined* to _indices_.
               1. Else,
-                1. Assert: _fullUnicode_ is *false*.
-                1. Assert: _captureI_ is a List of code units.
-                1. Let _capturedValue_ be the String value consisting of the code units of _captureI_.
+                1. Let _captureStart_ be _captureI_'s _startIndex_.
+                1. Let _captureEnd_ be _captureI_'s _endIndex_.
+                1. If _fullUnicode_ is *true*, then
+                  1. Set _captureStart_ to ! GetStringIndex(_S_, _Input_, _captureStart_).
+                  1. Set _captureEnd_ to ! GetStringIndex(_S_, _Input_, _captureEnd_).
```
rbuckton, line 36035: In reference to possibly dropping GetStringIndex, we could just do this instead:

```suggestion
                  1. _captureStart_ is an index into the _Input_ character list, derived from _S_, matched by _matcher_. Let _captureStartUTF_ be the smallest index into _S_ that corresponds to the character at element _captureStart_ of _Input_. If _captureStart_ is greater than or equal to the number of elements in _Input_, then _captureStartUTF_ is the number of code units in _S_.
                  1. Set _captureStart_ to _captureStartUTF_.
                  1. _captureEnd_ is an index into the _Input_ character list, derived from _S_, matched by _matcher_. Let _captureStartUTF_ be the smallest index into _S_ that corresponds to the character at element _captureEnd_ of _Input_. If _captureEnd_ is greater than or equal to the number of elements in _Input_, then _captureStartUTF_ is the number of code units in _S_.
                  1. Set _captureEnd_ to _captureEndUTF_.
```

spec.html
rbuckton: I'm not a fan of how wordy it is, but the change needed to drop GetStringIndex would look something like this: https://github.com/tc39/ecma262/pull/1713#pullrequestreview-853492024

rbuckton: Alternatively, the steps that define _Input_ and _listIndex_ in CompilePattern should be pulled out and placed in RegExpBuiltinExec instead, and passed as arguments to _matcher_ instead of _S_ and _lastIndex_, so that it can be referred to correctly.

rbuckton: > Alternatively, the steps that define _Input_ and _listIndex_ in CompilePattern should be pulled out and placed in RegExpBuiltinExec instead, and passed as arguments to _matcher_ instead of _S_ and _lastIndex_, so that it can be referred to correctly.

At first glance, this seems to be a cleaner approach and removes some of the ambiguity around _Input_. I'll amend this PR with this change and we can discuss it further.

spec.html
rbuckton: I've changed this a bit after I moved the step to derive _Input_ out of CompilePattern and into RegExpBuiltinExec.

spec.html
```diff
@@ -34920,26 +34923,25 @@ <h1>Notation</h1>
...
         </dl>
         <emu-grammar>Pattern :: Disjunction</emu-grammar>
         <emu-alg>
           1. Let _m_ be CompileSubpattern of |Disjunction| with argument ~forward~.
-          1. Return a new Abstract Closure with parameters (_str_, _index_) that captures _m_ and performs the following steps when called:
-            1. Assert: Type(_str_) is String.
-            1. Assert: _index_ is a non-negative integer which is &le; the length of _str_.
-            1. If _Unicode_ is *true*, let _Input_ be ! StringToCodePoints(_str_). Otherwise, let _Input_ be a List whose elements are the code units that are the elements of _str_. _Input_ will be used throughout the algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref>. Each element of _Input_ is considered to be a character.
+          1. Return a new Abstract Closure with parameters (_input_, _index_) that captures _m_ and performs the following steps when called:
+            1. Assert: Type(_input_) is a List of characters.
```
jmdyck, line 34932:
```suggestion
            1. Assert: _input_ is a List of characters.
```

spec.html
```diff
@@ -34920,26 +34923,25 @@ <h1>Notation</h1>
...
+            1. Assert: _index_ is a non-negative integer which is &le; the length of _input_.
```
jmdyck, line 34933:
```suggestion
            1. Assert: _index_ is a non-negative integer which is &le; the number of characters in _input_.
```
Generally, when an algorithm refers to `the length of _X_`, `_X_` is a String. For Lists, we usually use `the number of elements in _X_`, but in this case it can be more specific.

spec.html
```diff
@@ -35975,16 +35979,19 @@ <h1>
             1. Let _flags_ be _R_.[[OriginalFlags]].
             1. If _flags_ contains *"g"*, let _global_ be *true*; else let _global_ be *false*.
             1. If _flags_ contains *"y"*, let _sticky_ be *true*; else let _sticky_ be *false*.
+            1. If _flags_ contains *"d"*, let _hasIndices_ be *true*; else let _hasIndices_ be *false*.
             1. If _global_ is *false* and _sticky_ is *false*, set _lastIndex_ to 0.
             1. Let _matcher_ be _R_.[[RegExpMatcher]].
             1. If _flags_ contains *"u"*, let _fullUnicode_ be *true*; else let _fullUnicode_ be *false*.
             1. Let _matchSucceeded_ be *false*.
+            1. If _fullUnicode_ is *true*, let _input_ be ! StringToCodePoints(_S_). Otherwise, let _input_ be a List whose elements are the code units that are the elements of _S_. Each element of _input_ is considered to be a character.
```
jmdyck, line 35987:
```suggestion
            1. If _fullUnicode_ is *true*, let _input_ be ! StringToCodePoints(_S_). Otherwise, let _input_ be a List whose elements are the code units that are the elements of _S_.
            1. NOTE: Each element of _input_ is considered to be a character.
```
It's nice to separate the commentary from the 'executable' step. Doing so also makes it a bit clearer that the commentary applies to both possibilities, not just the "Otherwise" part.

spec.html
jmdyck: Are you saying those changes have rendered my suggestion invalid?

spec.html
```diff
@@ -36049,6 +36070,117 @@ <h1>
...
+          <emu-table id="table-match-record" caption="Match Record Fields">
+            <table>
+              <tr>
+                <th>Field Name</th>
+                <th>Value</th>
+                <th>Meaning</th>
+              </tr>
+              <tr>
+                <td>[[StartIndex]]</td>
+                <td>a non-negative integer.</td>
```
jmdyck, line 36105:
```suggestion
                <td>a non-negative integer</td>
```

spec.html
```diff
@@ -36049,6 +36070,117 @@ <h1>
...
+                <td>The number of code units from the start of a string at which the match begins (inclusive).</td>
+              </tr>
+              <tr>
+                <td>[[EndIndex]]</td>
+                <td>an integer &ge; [[StartIndex]].</td>
```
jmdyck, line 36110:
```suggestion
                <td>an integer &ge; [[StartIndex]]</td>
```

jmdyck (dismissed):

spec.html
rbuckton: More that I wasn't certain if you needed to revisit at all, or if I should change the wording in the header:

> `_Input_: a List of characters derived from _S_,`

spec.html
```diff
@@ -34920,26 +34923,25 @@ <h1>Notation</h1>
...
+          <p>A Pattern compiles to an Abstract Closure value. RegExpBuiltinExec can then apply this procedure to a List of characters and an offset within that list to determine whether the pattern would match starting at exactly that offset within the list, and, if it does match, what the values of the capturing parentheses would be. The algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref> are designed so that compiling a pattern may throw a *SyntaxError* exception; on the other hand, once the pattern is successfully compiled, applying the resulting Abstract Closure to find a match in a list of characters cannot throw an exception (except for any implementation-defined exceptions that can occur anywhere such as out-of-memory).</p>
```
michaelficarra, line 34944: Please capitalise List when referring to the spec data type.

bakkot (approved):
Some minor editorial stuff, but looks good overall.

@rbuckton if you'd like, the editors can address these and any other editorial issues ourselves before merging, since I think the intent is clear throughout. You're welcome to continue to address them yourself if you'd prefer (or get to it before we do), of course.

spec.html
```diff
@@ -36049,6 +36071,117 @@ <h1>
...
+          <emu-alg>
+            1. Assert: _match_.[[StartIndex]] is an integer value &ge; 0 and &le; the length of _S_.
+            1. Assert: _match_.[[EndIndex]] is an integer value &ge; _match_.[[StartIndex]] and &le; the length of _S_.
+            1. Return the substring of _S_ from _match_.[[StartIndex]] to _match_.[[EndIndex]].
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-getmatchindicesarray" type="abstract operation">
+          <h1>
+            GetMatchIndicesArray (
```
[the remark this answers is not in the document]
jmdyck: Just to keep the names straight, maybe use "IndexPair" for the array that's a start/end pair? So:
- `GetMatchIndicesArray` -> `GetMatchIndexPair`
- `_matchIndicesArray_` -> `_matchIndexPair_`

spec.html
```diff
@@ -36049,6 +36071,117 @@ <h1>
...
+            1. Assert: The value of _A_'s *"length"* property is 𝔽(_n_).
+            1. If _hasGroups_ is *true*, then
+              1. Let _groups_ be ! OrdinaryObjectCreate(*null*).
+            1. Else,
+              1. Let _groups_ be *undefined*.
+            1. Perform ! CreateDataPropertyOrThrow(_A_, *"groups"*, _groups_).
+            1. For each integer _i_ such that _i_ &ge; 0 and _i_ &lt; _n_, do
+              1. Let _matchIndices_ be _indices_[_i_].
+              1. If _matchIndices_ is not *undefined*, then
+                1. Let _matchIndicesArray_ be ! GetMatchIndicesArray(_S_, _matchIndices_).
```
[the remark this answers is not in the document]
jmdyck: `_range_` might be confusing, because of the definition of `Range`.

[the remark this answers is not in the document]
syg: Unhelpfully, V8's implementation just calls these "subarrays". English doesn't really have a notion of higher dimensions of plurality. Maybe `MakeMatchIndicesIndexPairArray`?

syg (approved):
lgtm % some naming nits

spec.html
```diff
@@ -36049,6 +36071,117 @@ <h1>
...
+          <emu-alg>
+            1. Assert: _Input_ contains the code points of StringToCodePoints(_S_).
+            1. If _S_ is the empty String, return 0.
+            1. Let _eUTF_ be the smallest index into _S_ that corresponds to the character at element _e_ of _Input_. If _e_ is greater than or equal to the number of elements in _Input_, then _eUTF_ is the number of code units in _S_.
+            1. Return _eUTF_.
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-match-records">
+          <h1>Match Records</h1>
```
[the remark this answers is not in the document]
rbuckton: I created a separate definition primarily because the only place where term "ordered pair" introduces some kind of record-like value in the spec is specific to regular expression pattern semantics. I felt that the creation of the regexp result object was far enough away from the pattern semantics to justify using spec mechanics used throughout the rest of the specification.

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
             1. Return _index_ + _cp_.[[CodeUnitCount]].
           </emu-alg>
         </emu-clause>
+
+        <emu-clause id="sec-getstringindex" type="abstract operation">
+          <h1>
+            GetStringIndex (
+              _S_: a String,
+              _e_: a non-negative integer,
+            )
```
michaelficarra, line 36080:
```suggestion
            ): a non-negative integer
```

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+            </table>
+          </emu-table>
+        </emu-clause>
+
+        <emu-clause id="sec-getmatchstring" type="abstract operation">
+          <h1>
+            GetMatchString (
+              _S_: a String,
+              _match_: a Match Record,
+            )
```
michaelficarra, line 36122:
```suggestion
            ): a String
```

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+          </h1>
+          <dl class="header">
+          </dl>
+          <emu-alg>
+            1. Assert: _match_.[[StartIndex]] is an integer value &ge; 0 and &lt; the length of _S_.
+            1. Assert: _match_.[[EndIndex]] is an integer value &ge; _match_.[[StartIndex]] and &le; the length of _S_.
+            1. Return ! CreateArrayFromList(&laquo; 𝔽(_match_.[[StartIndex]]), 𝔽(_match_.[[EndIndex]]) &raquo;).
```
michaelficarra, line 36145: This operation no longer fails. You should rebase and correct any errors like this that ecmarkup would now catch.

```suggestion
            1. Return CreateArrayFromList(&laquo; 𝔽(_match_.[[StartIndex]]), 𝔽(_match_.[[EndIndex]]) &raquo;).
```

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+          <h1>
+            GetMatchString (
+              _S_: a String,
+              _match_: a Match Record,
+            )
+          </h1>
+          <dl class="header">
+          </dl>
+          <emu-alg>
+            1. Assert: _match_.[[StartIndex]] is an integer value &ge; 0 and &le; the length of _S_.
```
michaelficarra, line 36127:
```suggestion
            1. Assert: _match_.[[StartIndex]] is a non-negative integer value &le; the length of _S_.
```

I don't really like using &ge; and &le; in ways that are not binary infix. These usages are kind of unary-y and it's weird. This change will at least reduce it. We could also spell everything out with words, as in `less than or equal to the length of _S_`.

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+            1. Return the substring of _S_ from _match_.[[StartIndex]] to _match_.[[EndIndex]].
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-getmatchindexpair" type="abstract operation">
+          <h1>
+            GetMatchIndexPair (
+              _S_: a String,
+              _match_: a Match Record,
+            )
```
michaelficarra, line 36138:
```suggestion
            ): an Array
```

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+            1. Assert: The value of _A_'s *"length"* property is 𝔽(_n_).
+            1. If _hasGroups_ is *true*, then
+              1. Let _groups_ be ! OrdinaryObjectCreate(*null*).
+            1. Else,
+              1. Let _groups_ be *undefined*.
+            1. Perform ! CreateDataPropertyOrThrow(_A_, *"groups"*, _groups_).
+            1. For each integer _i_ starting with 0 such that _i_ &lt; _n_, in ascending order, do
+              1. Let _matchIndices_ be _indices_[_i_].
+              1. If _matchIndices_ is not *undefined*, then
+                1. Let _matchIndexPair_ be ! GetMatchIndexPair(_S_, _matchIndices_).
```
michaelficarra, line 36175:
```suggestion
                1. Let _matchIndexPair_ be GetMatchIndexPair(_S_, _matchIndices_).
```

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+          </dl>
+          <emu-alg>
+            1. Let _n_ be the number of elements in _indices_.
+            1. Assert: _n_ &lt; 2<sup>32</sup> - 1.
+            1. Assert: _groupNames_ has _n_ - 1 elements.
+            1. NOTE: The _groupNames_ List contains elements aligned with the _indices_ List starting at _indices_[1].
+            1. Let _A_ be ! ArrayCreate(_n_).
+            1. Assert: The value of _A_'s *"length"* property is 𝔽(_n_).
+            1. If _hasGroups_ is *true*, then
+              1. Let _groups_ be ! OrdinaryObjectCreate(*null*).
```
michaelficarra, line 36168:
```suggestion
              1. Let _groups_ be OrdinaryObjectCreate(*null*).
```

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+        </emu-clause>
+
+        <emu-clause id="sec-makematchindicesindexpairarray" type="abstract operation">
+          <h1>
+            MakeMatchIndicesIndexPairArray (
+              _S_: a String,
+              _indices_: a List, each of whose elements is a Match Record or *undefined*,
+              _groupNames_: a List, each of whose elements is a String or *undefined*,
+              _hasGroups_: a Boolean,
+            )
```
michaelficarra, line 36156:
```suggestion
            ): an Array
```

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+            1. Return ! CreateArrayFromList(&laquo; 𝔽(_match_.[[StartIndex]]), 𝔽(_match_.[[EndIndex]]) &raquo;).
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-makematchindicesindexpairarray" type="abstract operation">
+          <h1>
+            MakeMatchIndicesIndexPairArray (
+              _S_: a String,
+              _indices_: a List, each of whose elements is a Match Record or *undefined*,
+              _groupNames_: a List, each of whose elements is a String or *undefined*,
```
michaelficarra, line 36154:
```suggestion
              _groupNames_: a List of either Strings or *undefined*,
```

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+            1. Assert: _match_.[[EndIndex]] is an integer value &ge; _match_.[[StartIndex]] and &le; the length of _S_.
+            1. Return ! CreateArrayFromList(&laquo; 𝔽(_match_.[[StartIndex]]), 𝔽(_match_.[[EndIndex]]) &raquo;).
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-makematchindicesindexpairarray" type="abstract operation">
+          <h1>
+            MakeMatchIndicesIndexPairArray (
+              _S_: a String,
+              _indices_: a List, each of whose elements is a Match Record or *undefined*,
```
michaelficarra, line 36153:
```suggestion
              _indices_: a List of either Match Records or *undefined*,
```

spec.html
```diff
@@ -34902,7 +34902,10 @@ <h1>Notation</h1>
...
-            A <em>State</em> is an ordered pair (_endIndex_, _captures_) where _endIndex_ is an integer and _captures_ is a List of _NcapturingParens_ values. States are used to represent partial match states in the regular expression matching algorithms. The _endIndex_ is one plus the index of the last input character matched so far by the pattern, while _captures_ holds the results of capturing parentheses. The _n_<sup>th</sup> element of _captures_ is either a List of characters that represents the value obtained by the _n_<sup>th</sup> set of capturing parentheses or *undefined* if the _n_<sup>th</sup> set of capturing parentheses hasn't been reached yet. Due to backtracking, many States may be in use at any time during the matching process.
+            A <em>Range</em> is an ordered pair (_startIndex_, _endIndex_) that represents the range of characters included in a capture, where _startIndex_ is an integer representing the start index (inclusive) of the range within _Input_, and _endIndex_ is an integer representing the end index (exclusive) of the range within _Input_. For any <em>Range</em>, these indices must satisfy the invariant that _startIndex_ &le; _endIndex_.
```
michaelficarra, line 34905: Can we at least `<dfn>` this so it links instead of adding another item to the https://github.com/tc39/ecma262/issues/2486 list?

spec.html
```diff
@@ -36049,6 +36071,116 @@ <h1>
...
+          </h1>
+          <dl class="header">
+          </dl>
+          <emu-alg>
+            1. Let _n_ be the number of elements in _indices_.
+            1. Assert: _n_ &lt; 2<sup>32</sup> - 1.
+            1. Assert: _groupNames_ has _n_ - 1 elements.
+            1. NOTE: The _groupNames_ List contains elements aligned with the _indices_ List starting at _indices_[1].
+            1. Let _A_ be ! ArrayCreate(_n_).
+            1. Assert: The value of _A_'s *"length"* property is 𝔽(_n_).
```
michaelficarra, line 36166: Why do we care about this assertion?

spec.html
```diff
@@ -36088,6 +36220,8 @@ <h1>get RegExp.prototype.flags</h1>
           1. Let _R_ be the *this* value.
           1. If Type(_R_) is not Object, throw a *TypeError* exception.
           1. Let _result_ be the empty String.
+          1. Let _hasIndices_ be ! ToBoolean(? Get(_R_, *"hasIndices"*)).
```
michaelficarra, line 36223:
```suggestion
          1. Let _hasIndices_ be ToBoolean(? Get(_R_, *"hasIndices"*)).
```

spec.html
```diff
@@ -36088,6 +36220,8 @@ <h1>get RegExp.prototype.flags</h1>
...
+          1. If _hasIndices_ is *true*, append the code unit 0x0064 (LATIN SMALL LETTER D) as the last code unit of _result_.
```
michaelficarra, line 36224:
```suggestion
          1. If _hasIndices_ is *true*, append the code unit 0x0064 (LATIN SMALL LETTER D) to _result_.
```

michaelficarra (commented):
LGTM otherwise.

spec.html
```diff
@@ -34920,26 +34923,25 @@ <h1>Notation</h1>
...
         <emu-alg>
           1. Let _m_ be CompileSubpattern of |Disjunction| with argument ~forward~.
-          1. Return a new Abstract Closure with parameters (_str_, _index_) that captures _m_ and performs the following steps when called:
-            1. Assert: Type(_str_) is String.
-            1. Assert: _index_ is a non-negative integer which is &le; the length of _str_.
-            1. If _Unicode_ is *true*, let _Input_ be ! StringToCodePoints(_str_). Otherwise, let _Input_ be a List whose elements are the code units that are the elements of _str_. _Input_ will be used throughout the algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref>. Each element of _Input_ is considered to be a character.
+          1. Return a new Abstract Closure with parameters (_input_, _index_) that captures _m_ and performs the following steps when called:
+            1. Assert: _input_ is a List of characters.
+            1. Assert: _index_ is a non-negative integer which is &le; the number of characters in _input_.
+            1. Let _Input_ be _input_. This alias will be used throughout the algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref>.
```
michaelficarra, line 34934: I really don't like that these aliases are the same other than casing. Can we choose better names?

michaelficarra:
@rbuckton I've rebased your PR and applied my editorial changes.

michaelficarra (approved):

spec.html
```diff
@@ -36247,25 +36247,25 @@ <h1>
...
-            )
+            ): an Array
           </h1>
           <dl class="header">
           </dl>
           <emu-alg>
-            1. Assert: _match_.[[StartIndex]] is an integer value &ge; 0 and &lt; the length of _S_.
-            1. Assert: _match_.[[EndIndex]] is an integer value &ge; _match_.[[StartIndex]] and &le; the length of _S_.
-            1. Return ! CreateArrayFromList(&laquo; 𝔽(_match_.[[StartIndex]]), 𝔽(_match_.[[EndIndex]]) &raquo;).
+            1. Assert: _match_.[[StartIndex]] is a non-negative integer less than or equal to the length of _S_.
```
michaelficarra, line 36255: This was totally on purpose and not an accident.

syg (approved):

spec.html
```diff
@@ -35054,10 +35054,10 @@ <h1>Runtime Semantics: CompilePattern ( ): an Abstract Closure that takes a Stri
...
         <emu-alg>
           1. Let _m_ be CompileSubpattern of |Disjunction| with argument ~forward~.
-          1. Return a new Abstract Closure with parameters (_input_, _index_) that captures _m_ and performs the following steps when called:
-            1. Assert: _input_ is a List of characters.
-            1. Assert: _index_ is a non-negative integer which is &le; the number of characters in _input_.
-            1. Let _Input_ be _input_. This alias will be used throughout the algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref>.
+          1. Return a new Abstract Closure with parameters (_inputChars_, _index_) that captures _m_ and performs the following steps when called:
+            1. Assert: _inputChars_ is a List of characters.
+            1. Assert: _index_ is a non-negative integer which is &le; the number of characters in _inputChars_.
+            1. Let _Input_ be _inputChars_. This alias will be used throughout the algorithms in <emu-xref href="#sec-pattern-semantics"></emu-xref>.
```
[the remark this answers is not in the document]
syg: I retract this suggestion given that it is the style today that RegExp aliases are capitalized.

PR #1716, opened 2019-10-02
Normative: make String.prototype.matchAll throw for non-global regular expressions

mathiasbynens:
Per October 2 2019 TC39 meeting consensus.

#### Tracking issues

- [V8](https://bugs.chromium.org/p/v8/issues/detail?id=9800)


ljharb:
Tests are merged: https://github.com/tc39/test262/pull/2389

ljharb:
Confirmed in person with @msaboff that the PR is good to go.

PR #1717, opened 2019-10-02
Update README.md

Sousa125:


ljharb:
Please don't spam.

jridgewell:
I don't think this is spam, it's just a stylistic choice. I definitely prefer the underscored headers to match the length of the text.

PR #1724, opened 2019-10-07
Editorial: Use consistent notation for code points.

rkirsling:
Re-evaluating all cases of <code>\`"foo"\`</code> throughout the spec (so as to switch to <code>\*"foo"\*</code> in one fell swoop) reveals that our notation for code ~units~ points isn't entirely consistent.

I'm starting with the maximal patch here for discussion's sake, but I'm happy to drop any (or all) of these changes; my only goal is to avoid blindly upgrading to asterisks without adequate consideration.

(CC: @jmdyck)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4758,7 +4758,7 @@
               The MV of <emu-grammar>StrUnsignedDecimalLiteral ::: DecimalDigits ExponentPart</emu-grammar> is the MV of |DecimalDigits| times 10<sub>ℝ</sub><sup>_e_</sup>, where _e_ is the MV of |ExponentPart|.
             </li>
           </ul>
-          <p>Once the exact MV for a String numeric literal has been determined, it is then rounded to a value of the Number type. If the MV is 0, then the rounded value is *+0* unless the first non white space code point in the String numeric literal is `"-"`, in which case the rounded value is *-0*. Otherwise, the rounded value must be the Number value for the MV (in the sense defined in <emu-xref href="#sec-ecmascript-language-types-number-type"></emu-xref>), unless the literal includes a |StrUnsignedDecimalLiteral| and the literal has more than 20 significant digits, in which case the Number value may be either the Number value for the MV of a literal produced by replacing each significant digit after the 20th with a 0 digit or the Number value for the MV of a literal produced by replacing each significant digit after the 20th with a 0 digit and then incrementing the literal at the 20th digit position. A digit is significant if it is not part of an |ExponentPart| and</p>
+          <p>Once the exact MV for a String numeric literal has been determined, it is then rounded to a value of the Number type. If the MV is 0, then the rounded value is *+0* unless the first non white space code point in the String numeric literal is `-`, in which case the rounded value is *-0*. Otherwise, the rounded value must be the Number value for the MV (in the sense defined in <emu-xref href="#sec-ecmascript-language-types-number-type"></emu-xref>), unless the literal includes a |StrUnsignedDecimalLiteral| and the literal has more than 20 significant digits, in which case the Number value may be either the Number value for the MV of a literal produced by replacing each significant digit after the 20th with a 0 digit or the Number value for the MV of a literal produced by replacing each significant digit after the 20th with a 0 digit and then incrementing the literal at the 20th digit position. A digit is significant if it is not part of an |ExponentPart| and</p>
           <ul>
             <li>
               it is not `0`; or
@@ -13002,7 +13002,7 @@
             It is a Syntax Error if BodyText of |RegularExpressionLiteral| cannot be recognized using the goal symbol |Pattern| of the ECMAScript RegExp grammar specified in <emu-xref href="#sec-patterns"></emu-xref>.
           </li>
           <li>
-            It is a Syntax Error if FlagText of |RegularExpressionLiteral| contains any code points other than `"g"`, `"i"`, `"m"`, `"s"`, `"u"`, or `"y"`, or if it contains the same code point more than once.
+            It is a Syntax Error if FlagText of |RegularExpressionLiteral| contains any code points other than `g`, `i`, `m`, `s`, `u`, or `y`, or if it contains the same code point more than once.
           </li>
         </ul>
       </emu-clause>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -12996,7 +12996,7 @@ <h1>Static Semantics: Early Errors</h1>
             It is a Syntax Error if BodyText of |RegularExpressionLiteral| cannot be recognized using the goal symbol |Pattern| of the ECMAScript RegExp grammar specified in <emu-xref href="#sec-patterns"></emu-xref>.
           </li>
           <li>
-            It is a Syntax Error if FlagText of |RegularExpressionLiteral| contains any code points other than `"g"`, `"i"`, `"m"`, `"s"`, `"u"`, or `"y"`, or if it contains the same code point more than once.
+            It is a Syntax Error if FlagText of |RegularExpressionLiteral| contains any code points other than `g`, `i`, `m`, `s`, `u`, or `y`, or if it contains the same code point more than once.
```
jmdyck: These are code points, not code units. (So the status quo is kind of wrong as well, but this PR probably shouldn't propagate that.)

spec.html
```diff
@@ -30939,7 +30939,7 @@ <h1>Static Semantics: StringValue</h1>
...
+      <p>A |Pattern| is either a BMP pattern or a Unicode pattern depending upon whether or not its associated flags contain a `u`. A BMP pattern matches against a String interpreted as consisting of a sequence of 16-bit values that are Unicode code points in the range of the Basic Multilingual Plane. A Unicode pattern matches against a String interpreted as consisting of Unicode code points encoded using UTF-16. In the context of describing the behaviour of a BMP pattern &ldquo;character&rdquo; means a single 16-bit Unicode BMP code point. In the context of describing the behaviour of a Unicode pattern &ldquo;character&rdquo; means a UTF-16 encoded code point (<emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>). In either context, &ldquo;character value&rdquo; means the numeric value of the corresponding non-encoded code point.</p>
```
jmdyck: This is (arguably) a code point, not a code unit.

spec.html
rkirsling: Agh, right, I am conflating the two (though in my defense, it's not just me—compare lines 25051 and 25076 here 😅). Is your vote to just worry about the code _point_ cases and let (sequences of) code units be represented as strings?

jmdyck:
To denote a code unit, we currently have a long way (e.g. `[the code unit] 0x002E (FULL STOP)`) and a short way (e.g. ``` `"."` ```). The latter notation also denotes a string, but context allows us to infer "the sole code unit of the string ...".

This PR proposes a different short notation, i.e. the current notation minus the quotes. But that is already the notation we use for terminal symbols and ES code snippets, i.e. source text, i.e. Unicode code points. I don't think it would be a good idea to use the same notation for code units and code points.

Instead, I think it would be okay for these cases to just change the backticks to asterisks as part of the one fell swoop, with the possible exception of the two cases noted.

spec.html
```diff
@@ -25044,11 +25044,11 @@ <h1>decodeURI ( _encodedURI_ )</h1>
         <p>The `decodeURI` function is the <dfn>%decodeURI%</dfn> intrinsic object. When the `decodeURI` function is called with one argument _encodedURI_, the following steps are taken:</p>
         <emu-alg>
           1. Let _uriString_ be ? ToString(_encodedURI_).
-          1. Let _reservedURISet_ be a String containing one instance of each code unit valid in |uriReserved| plus `"#"`.
+          1. Let _reservedURISet_ be a String containing one instance of each code unit valid in |uriReserved| plus `#`.
           1. Return ? Decode(_uriString_, _reservedURISet_).
         </emu-alg>
         <emu-note>
-          <p>The code point `"#"` is not decoded from escape sequences even though it is not a reserved URI code point.</p>
+          <p>The code point `#` is not decoded from escape sequences even though it is not a reserved URI code point.</p>
```
rkirsling: Seems like we should update `code point` → `code unit` here.

spec.html
jmdyck: Yup, that'd be my vote.

spec.html
jmdyck: It's a bit tricky. Although the input and output of `Decode` are just String values, they are different ways of encoding a ("true") URI consisting of Unicode code points. So it's somewhat ambiguous whether "decoded" in this sentence is talking about the whole effect of `Decode`, or a particular point within it at which an escape sequence could be (but isn't) decoded into a Unicode code point.

Given the placement of the note, it seems more plausible that it would be talking about the whole effect of `Decode`, rather than an internal detail, in which case changing the first occurrence of "code point" to "code unit" (as you suggest) would be appropriate. (Though rewording the sentence to not have that ambiguity would be a bigger improvement, I think.)

spec.html
```diff
@@ -32049,8 +32049,8 @@ <h1>Runtime Semantics: RegExpInitialize ( _obj_, _pattern_, _flags_ )</h1>
             1. Else, let _P_ be ? ToString(_pattern_).
             1. If _flags_ is *undefined*, let _F_ be the empty String.
             1. Else, let _F_ be ? ToString(_flags_).
-            1. If _F_ contains any code unit other than `"g"`, `"i"`, `"m"`, `"s"`, `"u"`, or `"y"` or if it contains the same code unit more than once, throw a *SyntaxError* exception.
-            1. If _F_ contains `"u"`, let _BMP_ be *false*; else let _BMP_ be *true*.
+            1. If _F_ contains any code unit other than `g`, `i`, `m`, `s`, `u`, or `y` or if it contains the same code unit more than once, throw a *SyntaxError* exception.
+            1. If _F_ contains `u`, let _BMP_ be *false*; else let _BMP_ be *true*.
```
rkirsling: Comparing the two lines you deemed worth changing to the two lines here, I find myself asking: _even when a distinction is being made, is it being made meaningfully?_

I wonder if it would be better to close this PR entirely and create an issue to re-evaluate usage of "code unit" and "code point" throughout the spec...?

rkirsling: Note: If we do keep the status quo of referring to RegExp flags with strings across the board, then we should instead fix the singular place where we're not doing that: the `y` in [this note](https://tc39.es/ecma262/#sec-assertion).

jmdyck: Well, there are two ways in which RegExp flags manifest:
- as code points in source text; and
- as code units in String values (arising either from the stringification of the above source text, or via the 'flags' parameter to the RegExp constructor).

Lines 12999 and 30942 are talking about flags as source text; lines 32052-3 are talking about flags in a String value.

> I wonder if it would be better to close this PR entirely and create an issue to re-evaluate usage of "code unit" and "code point" throughout the spec...?

If the point of the issue would be to ask whether the distinction between "code unit" and "code point" is meaningful, then I suspect you'll just get the answer "yes, it is". Don't let me stop you, though.

spec.html
rkirsling: Alternatively, maybe it's the parallel sentence below that should change in the opposite direction? Re-reading the sentence as a whole, it seems like it we could argue that it is making a note about the code point `#`, whereas the algorithm step is building a string from code units including **"#"**.

jmdyck: > Alternatively, maybe it's the parallel sentence below that should change in the opposite direction?

That is, on line 25076, change "code unit" to "code point". 

> Re-reading the sentence as a whole, it seems we could argue that the note is purely about the code point `#`,

Yup, it's debatable.

> whereas the algorithm step is building a string from code units including **"#"**.

The algorithm is clearly building a string, although the phrase `each code unit valid in |uriReserved|` is slightly problematic, if we assume (as with the spec's other lexical grammars) that the terminal symbols are code points, not code units. In effect, the 'uri' grammar is not there to specify a syntax (nothing is parsed with `|uri|` as the goal symbol), but merely to define  `|uriReserved|` and `|uriUnescaped|` as sets of code units. But that's mostly beside the point.

jmdyck:
The net effect of these two commits looks good to me.

rkirsling:
Added feedback from #1733 here.

jmdyck:
Yup, looks good.

spec.html
```diff
@@ -31793,7 +31793,7 @@ <h1>CharacterClassEscape</h1>
         <p>The production <emu-grammar>UnicodePropertyValueExpression :: LoneUnicodePropertyNameOrValue</emu-grammar> evaluates as follows:</p>
         <emu-alg>
           1. Let _s_ be SourceText of |LoneUnicodePropertyNameOrValue|.
-          1. If ! UnicodeMatchPropertyValue(`"General_Category"`, _s_) is identical to a List of Unicode code points that is the name of a Unicode general category or general category alias listed in the &ldquo;Property value and aliases&rdquo; column of <emu-xref href="#table-unicode-general-category-values"></emu-xref>, then
+          1. If ! UnicodeMatchPropertyValue(`General_Category`, _s_) is identical to a List of Unicode code points that is the name of a Unicode general category or general category alias listed in the &ldquo;Property value and aliases&rdquo; column of <emu-xref href="#table-unicode-general-category-values"></emu-xref>, then
```
rkirsling: Added one more case noticed in #1733.

jmdyck:
(Note that the PR was merged with the original title involving "code units" rather than the revised title involving "code points", which might be a source of confusion.)

ljharb:
Oops, thanks. Hopefully the linked PR will resolve any potential future confusion.

PR #1725, opened 2019-10-07
Editorial: Use tildes for spec-internal values wherever possible.

rkirsling:
Follow-up to #1302, prerequisite for the <code>\`"foo"\`</code> → <code>\*"foo"\*</code> update.

This patch aims to change all non-observable magic strings into `~foo~`-style spec-internal values.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -891,6 +891,11 @@
         <p><emu-eqn>floor<sub>_t_</sub>(_x_) = _x_ -<sub>_t_</sub> (_x_ modulo<sub>_t_</sub> 1<sub>_t_</sub>)</emu-eqn>.</p>
       </emu-note>
     </emu-clause>
+    <emu-clause id="sec-value-notation">
+      <h1>Value Notation</h1>
+      <p>In this specification, ECMAScript language values are displayed in *bold*. Examples include *null*, *true*, or *"hello"*. These are distinguished from longer ECMAScript code sequences such as `Function.prototype.apply` or `let n = 42;`.</p>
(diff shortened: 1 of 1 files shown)


ljharb:
This seems good, but more eyes is better. Also, can we add something about tilde-d things to a new section in https://tc39.es/ecma262/#sec-notational-conventions, either a 5.3, or a 5.2.6?

rkirsling:
We absolutely need such a section (not just for tildes but also asterisks, etc.), but I was hoping to address this in a separate PR. Is that okay?

ljharb:
I suppose, but I worry that without the section, the drift will continue - is there a reason it has to be delayed, or a rush to merge this one?

rkirsling:
Sorry, I didn't mean to imply any sort of chronology or delay; I just thought it would be deserving of its own review thread.

ljharb:
Hopefully it would be no more controversial than this very PR :-)

rkirsling:
Added a section 5.2.6.
(I suppose 5.2.5 and 5.2.6 could be turned into 5.3.1 and 5.3.2 but I'm not sure what 5.3's title would be.)

spec.html
```diff
@@ -890,6 +890,11 @@ <h1>Mathematical Operations</h1>
         <p><emu-eqn>floor<sub>_t_</sub>(_x_) = _x_ -<sub>_t_</sub> (_x_ modulo<sub>_t_</sub> 1<sub>_t_</sub>)</emu-eqn>.</p>
       </emu-note>
     </emu-clause>
+    <emu-clause id=sec-value-notation>
```
```suggestion
    <emu-clause id="sec-value-notation">
```

rkirsling:
Thanks for noticing that, @jmdyck! I've gone ahead with your "forward" suggestion.

jmdyck:
Yup, this looks good to go now.

PR #1727, opened 2019-10-08
Editorial: Formally disambiguate the non-Annex-B grammar

### LINKED ISSUE #969: Editorial: nested surrogate pairs?
In the ['Patterns'](https://tc39.github.io/ecma262/#sec-patterns) clause, the Syntax section contains the paragraph:
> Each `\u TrailSurrogate` for which the choice of associated `u LeadSurrogate` is ambiguous shall be associated with the nearest possible `u LeadSurrogate` that would otherwise have no corresponding `\u TrailSurrogate`.

This strikes me as very odd. The wording is exactly parallel to that for resolving the dangling else problem, suggesting that surrogate pairs *nest* somehow. I.e., given
```
\u LeadSurrogate \u LeadSurrogate \u TrailSurrogate \u TrailSurrogate
```
the last `TrailSurrogate` should be 'associated' with the first `LeadSurrogate`. But what would that even mean? (Note that the relevant semantics make no mention of "associated" or "corresponding" Surrogates.)

Now, granted, the grammar *is* formally ambiguous here (`Alternative[+U]` derives `\u LeadSurrogate \u TrailSurrogate` in two distinct ways), and so requires some disambiguation. But rather than the quoted paragraph, I'd prefer one of these approaches:

- If some part of the source text matches `\u LeadSurrogate \u TrailSurrogate`, it must be parsed as a single `Atom` rather than two.
- In cases of ambiguity, RegExpUnicodeEscapeSequence's first RHS is preferred over its second and third.
- Change the second RHS to `[+U] u LeadSurrogate [lookahead != \u TrailSurrogate]`

(One objection to the last approach is that `TrailSurrogate` is not a terminal symbol, and so this is not a 'legal' lookahead sequence. However, this doesn't bother me much: we could enlarge the definition of lookahead sequences if we wanted.)

gibson042:
@waldemarhorwat recently expressed objections to moving the Annex B regular expression syntax into the main grammar because of its order-dependent productions. However, I found an example of the same already in the main grammar, for surrogate pairs in codepoint-based "Unicode" regular expressions. This PR inserts a lookahead assertion to correct that ambiguity, and also applies the same treatment to replace some NumericLiteral prose with a formal assertion.

Closes #969.


spec.html
```diff
@@ -30493,24 +30492,20 @@ <h2>Syntax</h2>
           &lt;ZWJ&gt;
 
         RegExpUnicodeEscapeSequence[U] ::
-          [+U] `u` LeadSurrogate `\u` TrailSurrogate
-          [+U] `u` LeadSurrogate
-          [+U] `u` TrailSurrogate
-          [+U] `u` NonSurrogate
+          [+U] RegExpUnicodeSurrogatePair
+          [+U] [lookahead &notin; RegExpUnicodeSurrogatePair] `u` Hex4Digits
```
jmdyck: `RegExpUnicodeSurrogatePair` generates a set of terminal sequences, each of length 11. This doesn't fit with the current definition of lookahead-restrictions.

bakkot: `lookahead &notin; RegExpUnicodeSurrogatePair` seems a little suspicious to me, since `RegExpUnicodeSurrogatePair` describes a set of sequences of terminals rather than just a set of terminals. As far as I know the only other place where there are sequences of length greater than 1 in a lookahead-restriction-set is the `async function` restriction in [ExpressionStatement](https://tc39.es/ecma262/#sec-expression-statement), and there the sequence is of length precisely two (and also it's written out more explicitly).

Could this instead be written as

```
[+U] `u` LeadSurrogate [lookahead ≠ `\u` TrailSurrogate]
[+U] `u` LeadSurrogate `\u` TrailSurrogate
[+U] `u` TrailSurrogate
[+U] `u` NonSurrogate
```
? That feels clearer to me, if it's equivalent. (And if it's not, I'm confused.)

gibson042: That is equivalent, though still a lookahead of six code points. And not more clear to me, but I would be willing to switch to it if there's consensus.

gibson042:
There are currently three flavors of negative lookahead assertions, all defined in [Grammar Notation](https://tc39.es/ecma262/#sec-grammar-notation) (emphasis mine):
* [lookahead ∉ _set_], in which "_set_ can be written as a comma separated list of **one or two element terminal sequences** enclosed in curly brackets"
* [lookahead ∉ _set_], in which—"for convenience"—_set_ can be "written as a nonterminal, in which case it represents the set of all terminals to which that nonterminal could expand"
* [lookahead ≠ _terminal_]

So ``` `u` LeadSurrogate [lookahead ≠ `\u` TrailSurrogate]``` would not be valid, because the ≠ notation is only allowed with a single terminal input element.

But addressing the more substantive point, I agree with @jmdyck that the intent is to bound lookahead, in particular limiting it to two terminals (corresponding with recognition by an LR(2) parser). However, that is already not the case for two reasons. One of them has to do with the preservation of 
_LineTerminator_ input elements, such that a strict reading of [section 5.1.2](https://tc39.es/ecma262/#sec-lexical-and-regexp-grammars) requires **unbounded** lookahead for ```[lookahead ≠ `let [`]``` (because `let` and `[` could be separated by an arbitrary amount of _LineTerminator_-replaced _MultiLineComment_ sequences), and even a loose reading in which consecutive _LineTerminator_ elements were collapsed would still require a lookahead of up to three elements (`let`, _LineTerminator_, `[`). And the other reason why lookahead is not actually bounded at two or even three terminals is the very section that I am changing... `\uD834\uDF06` in a Unicode regular expression **must** be parsed as a single U+1D306 TETRAGRAM FOR CENTRE code point expanded from ```RegExpUnicodeEscapeSequence_U :: `u` LeadSurrogate `\u` TrailSurrogate``` rather than as a U+D834 code point expanded from ```RegExpUnicodeEscapeSequence_U :: `u` LeadSurrogate``` followed by a U+DF06 code point expanded from ```RegExpUnicodeEscapeSequence_U :: `u` TrailSurrogate```, and the content enforcing that restriction is currently prose rather than formal lookahead semantics, even though an actual implementation is not permitted to recognize a ```RegExpUnicodeEscapeSequence_U :: `u` LeadSurrogate``` expansion without confirming that the following six code points of source text do not match ``` `\u` TrailSurrogate```—which is **exactly** equivalent to a lookahead assertion!

It is worth noting, but not necessarily compelling, that this applies to the Regular Expression grammar rather than to the syntactic grammar. Nevertheless, I believe that these particulars should be formally expressed, even if that means admitting that ECMAScript is not as friendly to parse as it might otherwise appear to be (:roll_eyes:). I am willing to update the Grammar Notation section accordingly if that is the consensus. But an alternative does exist to expressing this requirement with a lookahead—introduction of a "longest expansion" rule analogous to the "longest input element" rule for [lexical scanning](https://tc39.es/ecma262/#sec-ecmascript-language-lexical-grammar). I'm not sure what exactly that would look like, but would be willing to try something out.

jmdyck:
> But an alternative does exist to expressing this requirement with a lookahead—introduction of a "longest expansion" rule analogous to the "longest input element" rule for [lexical scanning](https://tc39.es/ecma262/#sec-ecmascript-language-lexical-grammar). I'm not sure what exactly that would look like, but would be willing to try something out.

I have a feeling that would be a bad idea: too much chance of unintended consequences. (Might depend on the exact formulation, though.) Currently, I think using a lookahead-restriction is the best option, along with any necessary changes to the Grammar Notation section.

bakkot:
My own preference, revisiting this, is to relax the length bound on lookahead restrictions from being bounded at 2 to being bounded at 4 or (if 4 does not suffice) 6, ideally with that relaxation scoped specifically to the RegExp grammar.

This might look like 

> **In the RegExp grammar, the set can consist of nonempty sequences of terminals of length at most four. In the Syntactic grammar, the set can consist of nonempty sequences of terminals of length at most two. In the Lexical grammar, the set can consist of sequences of terminals of length exactly one. Lookaheads are not used in the Numeric String grammar.** The set can be written as a comma separated list of ~one or two element~ terminal sequences enclosed in curly brackets. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all **sequences of** terminals to which that nonterminal could expand.

(Additions in **bold**, removals ~struck through.) This also has the advantage of being explicit that the restriction on length of lookaheads is a property of all lookaheads, not just a particular way of writing them.

Thoughts?

spec.html
jmdyck: Currently, in the ES grammars, a lookahead-constraint either:
(a) occurs at the end of a right-hand-side, or
(b) occurs before a nonterminal, where that nonterminal derives phrases that begin with the disallowed sequences (i.e. the constraint is 'scoped' to that nonterminal).

So the right-hand-side:
```
[lookahead &notin; RegExpUnicodeSurrogatePair] `u` Hex4Digits
```
would be quite unusual, in that the lookahead-constraint has to "look through" the terminal `u` and nonterminal _HexDigits_, and then look past them to a potential `u` _Hex4Digits_ following. For this reason, I prefer @bakkot's suggestion:
```
LeadSurrogate [lookahead != `\u` TrailSurrogate]
```
as it eliminates the "look through" and winds up with a fairly standard end-of-RHS constraint. (But yes, the nature of its lookahead-sequence would require tweaking 5.1.5 Grammar Notation.)

Also, I think I'd prefer it to come *after* the Lead+Trail right-hand-side:
```
[+U] `u` LeadSurrogate `\u` TrailSurrogate
[+U] `u` LeadSurrogate [lookahead != `\u` TrailSurrogate]
```
(The other thing I like about this solution is that just those two lines make it really obvious why the lookahead-constraint is needed.)

gibson042:
Rebased and simplified to minimize churn.

nicolo-ribaudo:
Would it be possible to handle this by _not_ having a single term contain the two halves of a surrogate pair, but instead parse it as two separate terms and then doing the merging in the algorithms at some higher level?

gibson042:
> Would it be possible to handle this by _not_ having a single term contain the two halves of a surrogate pair, but instead parse it as two separate terms and then doing the merging in the algorithms at some higher level?

Discussion in today's editor call was overall negative on attempting such a change in this PR—it might not be at all convenient, and would certainly extend the scope to currently unchanged operations.

gibson042:
> a strict reading of [section 5.1.2](https://tc39.es/ecma262/#sec-lexical-and-regexp-grammars) requires **unbounded** lookahead for `` [lookahead ≠ `let [`] `` (because `let` and `[` could be separated by an arbitrary amount of _LineTerminator_-replaced _MultiLineComment_ sequences)

The latest commit closes this gap by specifying that consecutive line terminators in the stream of input elements for the syntactic grammar are collapsed.

jmdyck:
> a strict reading of [section 5.1.2](https://tc39.es/ecma262/#sec-lexical-and-regexp-grammars) requires **unbounded** lookahead for `` [lookahead ≠ `let [`] `` (because `let` and `[` could be separated by an arbitrary amount of _LineTerminator_-replaced _MultiLineComment_ sequences)

Well, that *would* be an unbounded amount of source text, and an unbounded number of input elements, but it's still only 2 tokens. And lookahead restrictions are supposedly defined on token sequences. On the other hand, that stopped being true when we added lookahead restrictions involving "[no LineTerminator here]".

PR #1728, opened 2019-10-08
Editorial: Reference DecimalDigit rather than duplicating its RHS

gibson042:
Simple improvement noticed while working on #1727.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -30549,16 +30549,7 @@ THH:mm:ss.sss
 
         UnicodePropertyValueCharacter ::
           UnicodePropertyNameCharacter
-          `0`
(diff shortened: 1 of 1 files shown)


ljharb:
@gibson042 is it also worth collapsing more things, so that `DecimalDigit` is composed of `OctalDigit` and `8` and `9`; and `HexDigit` is composed of `DecimalDigit` and a-F; and perhaps `OctalDigit` is composed of `ZeroToThree` and `FourToSeven`?

gibson042:
I'd be open to that, but I do think there's value in each of the "Digit" productions having a complete and independent enumeration.

PR #1733, opened 2019-10-11
Editorial: Stylize strings as values.

rkirsling:
#1725 included a new section on Value Notation; this PR enacts it by updating <code>\`"foo"\`</code> → <code>\*"foo"\*</code> throughout the spec. This means that all ECMAScript language values shall now be stylized as such.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -162,14 +162,14 @@
 
     <emu-clause id="sec-objects">
       <h1>Objects</h1>
-      <p>Even though ECMAScript includes syntax for class definitions, ECMAScript objects are not fundamentally class-based such as those in C++, Smalltalk, or Java. Instead objects may be created in various ways including via a literal notation or via <em>constructors</em> which create objects and then execute code that initializes all or part of them by assigning initial values to their properties. Each constructor is a function that has a property named `"prototype"` that is used to implement <em>prototype-based inheritance</em> and <em>shared properties</em>. Objects are created by using constructors in <b>new</b> expressions; for example, `new Date(2009, 11)` creates a new Date object. Invoking a constructor without using <b>new</b> has consequences that depend on the constructor. For example, `Date()` produces a string representation of the current date and time rather than an object.</p>
-      <p>Every object created by a constructor has an implicit reference (called the object's <em>prototype</em>) to the value of its constructor's `"prototype"` property. Furthermore, a prototype may have a non-null implicit reference to its prototype, and so on; this is called the <em>prototype chain</em>. When a reference is made to a property in an object, that reference is to the property of that name in the first object in the prototype chain that contains a property of that name. In other words, first the object mentioned directly is examined for such a property; if that object contains the named property, that is the property to which the reference refers; if that object does not contain the named property, the prototype for that object is examined next; and so on.</p>
+      <p>Even though ECMAScript includes syntax for class definitions, ECMAScript objects are not fundamentally class-based such as those in C++, Smalltalk, or Java. Instead objects may be created in various ways including via a literal notation or via <em>constructors</em> which create objects and then execute code that initializes all or part of them by assigning initial values to their properties. Each constructor is a function that has a property named *"prototype"* that is used to implement <em>prototype-based inheritance</em> and <em>shared properties</em>. Objects are created by using constructors in <b>new</b> expressions; for example, `new Date(2009, 11)` creates a new Date object. Invoking a constructor without using <b>new</b> has consequences that depend on the constructor. For example, `Date()` produces a string representation of the current date and time rather than an object.</p>
+      <p>Every object created by a constructor has an implicit reference (called the object's <em>prototype</em>) to the value of its constructor's *"prototype"* property. Furthermore, a prototype may have a non-null implicit reference to its prototype, and so on; this is called the <em>prototype chain</em>. When a reference is made to a property in an object, that reference is to the property of that name in the first object in the prototype chain that contains a property of that name. In other words, first the object mentioned directly is examined for such a property; if that object contains the named property, that is the property to which the reference refers; if that object does not contain the named property, the prototype for that object is examined next; and so on.</p>
       <emu-figure id="figure-1" caption="Object/Prototype Relationships">
         <object data="img/figure-1.svg" height="354" type="image/svg+xml" width="719"> <img alt="An image of lots of boxes and arrows." height="354" src="img/figure-1.png" width="719"> </object>
       </emu-figure>
       <p>In a class-based object-oriented language, in general, state is carried by instances, methods are carried by classes, and inheritance is only of structure and behaviour. In ECMAScript, the state and methods are carried by objects, while structure, behaviour, and state are all inherited.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -42817,10 +42817,10 @@ <h1>Corrections and Clarifications in ECMAScript 2015 with Possible Compatibilit
...
   <p><emu-xref href="#sec-date.prototype.toisostring"></emu-xref>: If the year cannot be represented using the Date Time String Format specified in <emu-xref href="#sec-date-time-string-format"></emu-xref> a RangeError exception is thrown. Previous editions did not specify the behaviour for that case.</p>
   <p><emu-xref href="#sec-date.prototype.tostring"></emu-xref>: Previous editions did not specify the value returned by `Date.prototype.toString` when this time value is *NaN*. ECMAScript 2015 specifies the result to be the String value *"Invalid Date"*.</p>
-  <p><emu-xref href="#sec-regexp-pattern-flags"></emu-xref>, <emu-xref href="#sec-escaperegexppattern"></emu-xref>: Any LineTerminator code points in the value of the `"source"` property of a RegExp instance must be expressed using an escape sequence. Edition 5.1 only required the escaping of `"/"`.</p>
+  <p><emu-xref href="#sec-regexp-pattern-flags"></emu-xref>, <emu-xref href="#sec-escaperegexppattern"></emu-xref>: Any LineTerminator code points in the value of the *"source"* property of a RegExp instance must be expressed using an escape sequence. Edition 5.1 only required the escaping of `/`.</p>
```
rkirsling: Note: The removal of quotes around the `/` here is just to (hopefully?) simplify conflict resolution with #1724.

spec.html
```diff
@@ -31624,7 +31624,7 @@ <h1>Runtime Semantics: Canonicalize ( _ch_ )</h1>
...
+            <p>In case-insignificant matches when _Unicode_ is *true*, all characters are implicitly case-folded using the simple mapping provided by the Unicode standard immediately before they are compared. The simple mapping always maps to a single code point, so it does not map, for example, *"&szlig;"* (U+00DF) to *"SS"*. It may however map a code point outside the Basic Latin range to a character within, for example, *"&#x17f;"* (U+017F) to *"s"*. Such characters are not mapped if _Unicode_ is *false*. This prevents Unicode code points such as U+017F and U+212A from matching regular expressions such as `/[a-z]/i`, but they will match `/[a-z]/ui`.</p>
```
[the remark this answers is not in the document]
rkirsling: This particular line is some remaining confusion from #1724, as I thought we decided not to remove the quotes, but I ought to have asked that more explicitly over there.

spec.html
```diff
@@ -19025,7 +19025,7 @@ <h2>Syntax</h2>
...
       <h1>Directive Prologues and the Use Strict Directive</h1>
       <p>A <dfn id="directive-prologue">Directive Prologue</dfn> is the longest sequence of |ExpressionStatement|s occurring as the initial |StatementListItem|s or |ModuleItem|s of a |FunctionBody|, a |ScriptBody|, or a |ModuleBody| and where each |ExpressionStatement| in the sequence consists entirely of a |StringLiteral| token followed by a semicolon. The semicolon may appear explicitly or may be inserted by automatic semicolon insertion. A Directive Prologue may be an empty sequence.</p>
-      <p>A <dfn id="use-strict-directive">Use Strict Directive</dfn> is an |ExpressionStatement| in a Directive Prologue whose |StringLiteral| is either the exact code unit sequences `"use strict"` or `'use strict'`. A Use Strict Directive may not contain an |EscapeSequence| or |LineContinuation|.</p>
+      <p>A <dfn id="use-strict-directive">Use Strict Directive</dfn> is an |ExpressionStatement| in a Directive Prologue whose |StringLiteral| is either the exact code unit sequences *"use strict"* or *'use strict'*. A Use Strict Directive may not contain an |EscapeSequence| or |LineContinuation|.</p>
```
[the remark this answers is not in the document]
rkirsling: Sounds good; this is the this I was most conflicted over myself.

spec.html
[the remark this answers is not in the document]
jmdyck: I don't think we made an explicit decision on this case in #1724, I just failed to flag it as code points there.

jmdyck:
The net effect is good now, but I think, for clarity, it would be better to restructure the PR as two commits:
1. Handling the 3 code point exceptions. (The commit message could describe it as a belated addition to #1724.)
2. All the backtick-to-asterisk cases.

rkirsling:
PRs are generally squash-merged though, right? So in that sense, it might be better to move the third commit here _over to_ #1724 (since the merge button hasn't been hit on it yet) and ensure that those lines aren't touched in this PR.

jmdyck:
Ah! I totally forgot that #1724 was still open. Moving those cases there makes sense.

rkirsling:
Moved third commit to #1724 and dropped the problematic lines from the first commit here.

spec.html
```diff
@@ -31793,7 +31793,7 @@ <h1>CharacterClassEscape</h1>
         <p>The production <emu-grammar>UnicodePropertyValueExpression :: LoneUnicodePropertyNameOrValue</emu-grammar> evaluates as follows:</p>
         <emu-alg>
           1. Let _s_ be SourceText of |LoneUnicodePropertyNameOrValue|.
-          1. If ! UnicodeMatchPropertyValue(`"General_Category"`, _s_) is identical to a List of Unicode code points that is the name of a Unicode general category or general category alias listed in the &ldquo;Property value and aliases&rdquo; column of <emu-xref href="#table-unicode-general-category-values"></emu-xref>, then
+          1. If ! UnicodeMatchPropertyValue(*"General_Category"*, _s_) is identical to a List of Unicode code points that is the name of a Unicode general category or general category alias listed in the &ldquo;Property value and aliases&rdquo; column of <emu-xref href="#table-unicode-general-category-values"></emu-xref>, then
```
jmdyck: Found another! `UnicodeMatchPropertyValue` "takes two parameters ..., each of which is a List of Unicode code points". So here, "General_Category" represents a List of code points, not a String value. (Ditto the occurrence on the next line.) This case should probably be handled in #1724.

rkirsling: Good catch! I don't think the next line applies though, as it's referring to [this](https://unicode.org/reports/tr44/#UnicodeData.txt).

rkirsling:
Removed those two changes from this PR.

spec.html
jmdyck: >  I don't think the next line applies though

Okay, I can go along with that.

Ms2ger:
Are there plans to fix up all the active proposals?

rkirsling:
@Ms2ger That is a very good question. It seems quite a hurdle to address all proposal repos at all stages at once. Perhaps we should enforce this just at the point of the actual 262 PR?

syg:
> Are there plans to fix up all the active proposals?

Surely not for proposal spec text. The old convention doesn't impede readability and understanding. But of course we should take care to apply the new convention when merging proposals into master.

PR #1736, opened 2019-10-14
Editorial: Stylize "this object" consistently.

rkirsling:
Currently we have 7 cases of '`this` object' and 19 cases of '**this** object' in the spec.

This appears to be short for '**this** value (which is an object)', so let's unify them accordingly.


jmdyck:
Makes sense to me.

PR #1739, opened 2019-10-17
Normative: function implementation hiding

michaelficarra:
This PR implements the currently stage 2 [function implementation hiding](https://github.com/tc39/proposal-function-implementation-hiding) proposal. I am opening it now to get reviewer feedback and editor sign-off, and to start having a linkable spec text rendering. It is my intention to present this proposal for stage 3 at the June 2020 TC39 meeting.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -490,7 +490,7 @@
       <h1>The Syntactic Grammar</h1>
       <p>The <em>syntactic grammar</em> for ECMAScript is given in clauses 11, 12, 13, 14, and 15. This grammar has ECMAScript tokens defined by the lexical grammar as its terminal symbols (<emu-xref href="#sec-lexical-and-regexp-grammars"></emu-xref>). It defines a set of productions, starting from two alternative goal symbols |Script| and |Module|, that describe how sequences of tokens form syntactically correct independent components of ECMAScript programs.</p>
       <p>When a stream of code points is to be parsed as an ECMAScript |Script| or |Module|, it is first converted to a stream of input elements by repeated application of the lexical grammar; this stream of input elements is then parsed by a single application of the syntactic grammar. The input stream is syntactically in error if the tokens in the stream of input elements cannot be parsed as a single instance of the goal nonterminal (|Script| or |Module|), with no tokens left over.</p>
-      <p>When a parse is successful, it constructs a <em>parse tree</em>, a rooted tree structure in which each node is a <dfn>Parse Node</dfn>. Each Parse Node is an <em>instance</em> of a symbol in the grammar; it represents a span of the source text that can be derived from that symbol. The root node of the parse tree, representing the whole of the source text, is an instance of the parse's goal symbol. When a Parse Node is an instance of a nonterminal, it is also an instance of some production that has that nonterminal as its left-hand side. Moreover, it has zero or more <em>children</em>, one for each symbol on the production's right-hand side: each child is a Parse Node that is an instance of the corresponding symbol.</p>
+      <p>When a parse is successful, it constructs a <em>parse tree</em>, a rooted tree structure in which each node is a <dfn>Parse Node</dfn>. Each Parse Node is an <em>instance</em> of a symbol in the grammar; it represents a span of the source text that can be derived from that symbol. The root node of the parse tree, representing the whole of the source text, is an instance of the parse's goal symbol. When a Parse Node is an instance of a nonterminal, it is also an instance of some production that has that nonterminal as its left-hand side. Moreover, it has zero or more <em>children</em>, one for each symbol on the production's right-hand side: each child is a Parse Node that is an instance of the corresponding symbol. A Parse Node _A_ occurs within another Parse Node _B_ if _A_ is the child of _B_ or if _A_ is the child of some Parse Node _C_ where _C_ occurs within _B_.</p>
       <p>New Parse Nodes are instantiated for each invocation of the parser and never reused between parses even of identical source text. Parse Nodes are considered <dfn>the same Parse Node</dfn> if and only if they represent the same span of source text, are instances of the same grammar symbol, and resulted from the same parser invocation.</p>
       <emu-note>
         <p>Parsing the same String multiple times will lead to different Parse Nodes. For example, consider:</p>
@@ -8357,6 +8357,17 @@
         </tr>
         <tr>
           <td>
+            [[HasSourceTextAvailable]]
+          </td>
+          <td>
+            Boolean
+          </td>
+          <td>
+            When *false*, this function will be rendered as a |NativeFunction| by `Function.prototype.toString` (<emu-xref href="#sec-function.prototype.tostring"></emu-xref>) and corresponding stack trace entries will have no incidental attribution or position information, as defined in <emu-xref href="#sec-error-objects"></emu-xref>.
+          </td>
+        </tr>
+        <tr>
+          <td>
             [[HomeObject]]
           </td>
           <td>
@@ -8503,6 +8514,7 @@
         1. Set _F_.[[FormalParameters]] to _ParameterList_.
         1. Set _F_.[[ECMAScriptCode]] to _Body_.
         1. If the source text matching _Body_ is strict mode code, let _Strict_ be *true*; else let _Strict_ be *false*.
+        1. Set _F_.[[HasSourceTextAvailable]] to HasSourceTextAvailable of _Body_.
         1. Set _F_.[[Strict]] to _Strict_.
         1. If _thisMode_ is ~lexical-this~, set _F_.[[ThisMode]] to ~lexical~.
         1. Else if _Strict_ is *true*, set _F_.[[ThisMode]] to ~strict~.
@@ -13620,6 +13632,18 @@
         </emu-grammar>
         <emu-alg>
           1. Return ~invalid~.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-callexpression-static-semantics-hassourcetextavailable">
+        <h1>Static Semantics: HasSourceTextAvailable</h1>
+        <emu-see-also-para op="HasSourceTextAvailable"></emu-see-also-para>
+        <emu-grammar>CallExpression : CoverCallExpressionAndAsyncArrowHead</emu-grammar>
+        <emu-alg>
+          1. If |CallExpression| occurs within a |ScriptBody|, |ModuleBody|, or |FunctionBody| that has a Directive Prologue that contains a Hidden Implementation Directive, return *false*.
+          1. If the source text matched by |CallExpression| is eval code resulting from a direct eval, then
+            1. Return HasSourceTextAvailable of the |CallExpression| whose evaluation is the direct eval.
+          1. Return *true*.
         </emu-alg>
       </emu-clause>
     </emu-clause>
@@ -19426,12 +19450,14 @@
     </emu-grammar>
 
     <emu-clause id="sec-directive-prologues-and-the-use-strict-directive">
-      <h1>Directive Prologues and the Use Strict Directive</h1>
+      <h1>Directive Prologues</h1>
       <p>A <dfn id="directive-prologue">Directive Prologue</dfn> is the longest sequence of |ExpressionStatement|s occurring as the initial |StatementListItem|s or |ModuleItem|s of a |FunctionBody|, a |ScriptBody|, or a |ModuleBody| and where each |ExpressionStatement| in the sequence consists entirely of a |StringLiteral| token followed by a semicolon. The semicolon may appear explicitly or may be inserted by automatic semicolon insertion. A Directive Prologue may be an empty sequence.</p>
       <p>A <dfn id="use-strict-directive">Use Strict Directive</dfn> is an |ExpressionStatement| in a Directive Prologue whose |StringLiteral| is either of the exact code point sequences `"use strict"` or `'use strict'`. A Use Strict Directive may not contain an |EscapeSequence| or |LineContinuation|.</p>
-      <p>A Directive Prologue may contain more than one Use Strict Directive. However, an implementation may issue a warning if this occurs.</p>
+      <p>A <dfn id="hidden-implementation-directive">Hidden Implementation Directive</dfn> is an |ExpressionStatement| in a Directive Prologue whose |StringLiteral| is either of the exact code point sequences `"hide source"` or `'hide source'`. A Hidden Implementation Directive may not contain an |EscapeSequence| or |LineContinuation|.</p>
+      <p>A <dfn id="built-in-directive">Built-in Directive</dfn> is an |ExpressionStatement| in a Directive Prologue that is either a Use Strict Directive or a Hidden Implementation Directive.</p>
+      <p>A Directive Prologue may contain duplicate Built-in Directives. However, an implementation may issue a warning if this occurs.</p>
       <emu-note>
-        <p>The |ExpressionStatement|s of a Directive Prologue are evaluated normally during evaluation of the containing production. Implementations may define implementation specific meanings for |ExpressionStatement|s which are not a Use Strict Directive and which occur in a Directive Prologue. If an appropriate notification mechanism exists, an implementation should issue a warning if it encounters in a Directive Prologue an |ExpressionStatement| that is not a Use Strict Directive and which does not have a meaning defined by the implementation.</p>
+        <p>The |ExpressionStatement|s of a Directive Prologue are evaluated normally during evaluation of the containing production. Implementations may define implementation-specific meanings for |ExpressionStatement|s in a Directive Prologue which are not a Built-in Directive. If an appropriate notification mechanism exists, an implementation should issue a warning if it encounters in a Directive Prologue an |ExpressionStatement| that is not a Built-in Directive and which does not have a meaning defined by the implementation.</p>
       </emu-note>
     </emu-clause>
 
@@ -19614,6 +19640,19 @@
       <emu-grammar>FunctionBody : FunctionStatementList</emu-grammar>
       <emu-alg>
         1. If the Directive Prologue of |FunctionBody| contains a Use Strict Directive, return *true*; otherwise, return *false*.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-function-definitions-static-semantics-hassourcetextavailable">
+      <h1>Static Semantics: HasSourceTextAvailable</h1>
+      <emu-see-also-para op="HasSourceTextAvailable"></emu-see-also-para>
+      <emu-grammar>FunctionBody : FunctionStatementList</emu-grammar>
+      <emu-alg>
+        1. If the Directive Prologue of |FunctionBody| contains a Hidden Implementation Directive, return *false*.
+        1. If |FunctionBody| occurs within a |ScriptBody|, |ModuleBody|, or |FunctionBody| that has a Directive Prologue that contains a Hidden Implementation Directive, return *false*.
+        1. If the source text matched by |FunctionBody| is eval code resulting from a direct eval, then
+          1. Return HasSourceTextAvailable of the |CallExpression| whose evaluation is the direct eval.
+        1. Return *true*.
       </emu-alg>
     </emu-clause>
 
@@ -20014,6 +20053,18 @@
       <emu-grammar>ConciseBody : ExpressionBody</emu-grammar>
       <emu-alg>
         1. Return *false*.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-arrow-function-definitions-static-semantics-hassourcetextavailable">
+      <h1>Static Semantics: HasSourceTextAvailable</h1>
+      <emu-see-also-para op="HasSourceTextAvailable"></emu-see-also-para>
+      <emu-grammar>ExpressionBody : AssignmentExpression</emu-grammar>
+      <emu-alg>
+        1. If |ExpressionBody| occurs within a |ScriptBody|, |ModuleBody|, or |FunctionBody| that has a Directive Prologue that contains a Hidden Implementation Directive, return *false*.
+        1. If the source text matched by |ExpressionBody| is eval code resulting from a direct eval, then
+          1. Return HasSourceTextAvailable of the |CallExpression| whose evaluation is the direct eval.
+        1. Return *true*.
       </emu-alg>
     </emu-clause>
 
@@ -26394,7 +26445,7 @@
         <p>When the `toString` method is called, the following steps are taken:</p>
         <emu-alg>
           1. Let _func_ be the *this* value.
-          1. If _func_ is a <emu-xref href="#sec-bound-function-exotic-objects">bound function exotic object</emu-xref> or a <emu-xref href="#sec-built-in-function-objects">built-in function object</emu-xref>, then return an implementation-dependent String source code representation of _func_. The representation must have the syntax of a |NativeFunction|. Additionally, if _func_ is a <emu-xref href="#sec-well-known-intrinsic-objects">Well-known Intrinsic Object</emu-xref> and is not identified as an anonymous function, the portion of the returned String that would be matched by |PropertyName| must be the initial value of the *"name"* property of _func_.
+          1. If _func_ is a <emu-xref href="#sec-bound-function-exotic-objects">bound function exotic object</emu-xref> or a <emu-xref href="#sec-built-in-function-objects">built-in function object</emu-xref> or has a [[HasSourceTextAvailable]] slot with the value *false*, then return an implementation-dependent String source code representation of _func_. The representation must have the syntax of a |NativeFunction|. Additionally, if _func_ is a <emu-xref href="#sec-well-known-intrinsic-objects">Well-known Intrinsic Object</emu-xref> and is not identified as an anonymous function, the portion of the returned String that would be matched by |PropertyName| must be the initial value of the *"name"* property of _func_.
           1. If Type(_func_) is Object and _func_ has a [[SourceText]] internal slot and _func_.[[SourceText]] is a sequence of Unicode code points and ! HostHasSourceTextAvailable(_func_) is *true*, then
             1. Return ! UTF16Encode(_func_.[[SourceText]]).
           1. If Type(_func_) is Object and IsCallable(_func_) is *true*, then return an implementation-dependent String source code representation of _func_. The representation must have the syntax of a |NativeFunction|.
@@ -26830,6 +26881,7 @@
   <emu-clause id="sec-error-objects">
     <h1>Error Objects</h1>
     <p>Instances of Error objects are thrown as exceptions when runtime errors occur. The Error objects may also serve as base objects for user-defined exception classes.</p>
+    <p>Stack trace information exposed to the running program through implementation-defined accessors such as the de facto `Error.prototype.stack` must not include any incidental attribution or position information related to functions whose [[HasSourceTextAvailable]] slot has a value of *false*. Examples of incidental attribution information are filenames, module specifiers, and URIs. Examples of incidental position information are line numbers and column numbers.</p>
 
     <emu-clause id="sec-error-constructor">
       <h1>The Error Constructor</h1>
--- a/spec.html
+++ b/spec.html
@@ -8357,6 +8357,17 @@
         </tr>
         <tr>
           <td>
+            [[PresentInStackTraces]]
+          </td>
+          <td>
+            Boolean
+          </td>
+          <td>
+            When *false*, this function must not have a corresponding entry in stack traces.
+          </td>
+        </tr>
+        <tr>
+          <td>
             [[HasSourceTextAvailable]]
           </td>
           <td>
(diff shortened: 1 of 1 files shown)


michaelficarra:
FYI I would still like feedback on alternative directive names at https://github.com/tc39/proposal-function-implementation-hiding/issues/3. I don't really love the current ones, but haven't heard anything better.

spec.html
```diff
@@ -8498,6 +8536,12 @@ <h1>FunctionInitialize ( _F_, _kind_, _ParameterList_, _Body_, _Scope_ )</h1>
         1. Let _len_ be the ExpectedArgumentCount of _ParameterList_.
         1. Perform ! SetFunctionLength(_F_, _len_).
         1. If the source text matching _Body_ is strict mode code, let _Strict_ be *true*; else let _Strict_ be *false*.
+        1. Set _F_.[[AppearsInStackTraces]] to OmittedFromStackTraces of _Body_.
```
claudepache: ??? I’d expect [[AppearsInStackTraces]] to be the logical negation of OmittedInStackTraces.

michaelficarra: Thanks. I was trying to avoid "negative" Boolean-valued slot names, but it seems to be more trouble than it's worth. Renamed to `OmittedFromStackTraces` for simplicity.

spec.html
```diff
@@ -8340,6 +8356,28 @@ <h1>ECMAScript Function Objects</h1>
...
+          <td>
+            Boolean
+          </td>
+          <td>
+            When *true*, this function must not have a corresponding entry in stack traces.
+          </td>
+        </tr>
+        <tr>
+          <td>
+            [[ImplementationHidden]]
```
ljharb: Maybe something more like the slot for toString, `[[HasSourceTextAvailable]]`?

I’d also expect one slot not to apply the other; iow, I’d expect both to be required to hide both.

I’ll leave a comment on the proposal issue about the bikeshed; i don’t think we’ve landed on the best pair of wordings quite yet :-)

spec.html
```diff
@@ -19023,12 +19067,15 @@ <h2>Syntax</h2>
...
       <p>A <dfn id="use-strict-directive">Use Strict Directive</dfn> is an |ExpressionStatement| in a Directive Prologue whose |StringLiteral| is either of the exact code point sequences `"use strict"` or `'use strict'`. A Use Strict Directive may not contain an |EscapeSequence| or |LineContinuation|.</p>
-      <p>A Directive Prologue may contain more than one Use Strict Directive. However, an implementation may issue a warning if this occurs.</p>
+      <p>A <dfn id="sensitive-directive">Sensitive Directive</dfn> is an |ExpressionStatement| in a Directive Prologue whose |StringLiteral| is either of the exact code point sequences `"sensitive"` or `'sensitive'`. A Sensitive Directive may not contain an |EscapeSequence| or |LineContinuation|.</p>
```
ljharb: why are any of these dfn ids necessary?

jmdyck: When ecmarkup auto-links occurrences of defined terms:
- For a `<dfn>` with an `id`, the link will go to the `<dfn>`.
- For a `<dfn>` without an `id`, the link will go to the `<dfn>`'s containing clause. (I.e., ecmarkup doesn't fabricate an id.)

So whether the ids are necessary depends on how precise the links need to be. (Personally, I'm okay with the latter if the `<dfn>` is close to the start of the clause.)

michaelficarra:
Ping @wycats @waldemarhorwat @mikesamuel. You are the stage 2 reviewers for this proposal.

michaelficarra:
Ping @wycats @waldemarhorwat @mikesamuel once more. Please review before the upcoming meeting. I would like this proposal to advance to stage 3.

erights:
Hi @michaelficarra , has anything changed since our discussion? IOW, is it essentially the same as what I approved in that discussion?

Can I see a rendered form of this? I do not want to evaluate this by reading ecmarkup source.

michaelficarra:
@erights The only changes were the ones we talked about in our last call. See [the most recently closed issues](https://github.com/tc39/proposal-function-implementation-hiding/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), particularly https://github.com/tc39/proposal-function-implementation-hiding/issues/32, https://github.com/tc39/proposal-function-implementation-hiding/issues/31, and https://github.com/tc39/proposal-function-implementation-hiding/issues/30.

You can see a rendering of any PRs to ecma262, including this one, by clicking "Details" next to the words "Deploy preview ready!" in the CI checks below.

erights:
> You can see a rendering of any PRs to ecma262, including this one, by clicking "Details" next to the words "Deploy preview ready!" in the CI checks below.

Good to know, thanks!

waldemarhorwat:
I'm having trouble reviewing what _GetDirectiveContext_ is doing. It seems to make some unstated assumptions about what happens to be on the execution context stack at the time it is called. I can't tell whether that matches the lexical scope for all of the possible call paths (and note that _Scope_ is passed as a separate parameter and strict mode scoping is done differently). Furthermore, more call paths may be added in the future, so relying on this kind of a hidden side effect seems brittle.

waldemarhorwat:
If I understand it correctly, _GetDirectiveContext_ seems broken. Just filed [an issue](https://github.com/tc39/proposal-function-implementation-hiding/issues/36) for it.

michaelficarra:
@waldemarhorwat You're right, I think I had a bug. I've removed `GetDirectiveContext` in https://github.com/tc39/ecma262/pull/1739/commits/19687c88f35222d0903137beb88561d8b5c5f77d since it wasn't really helpful anymore. Please see if that resolves your issues. If we don't use the execution context stack to track the directive state here, what alternative would you recommend? /cc @allenwb

spec.html
michaelficarra: I'm sorry @ljharb but I don't understand any part of that comment.

ljharb: Sure, I'll rephrase: your current implementation allows these two slots to function independently. This means that, although the proposal only has three states: "true and true", "true and false", or "false and false", four states are possible.

Perhaps an enum, rather than two booleans, is a better choice; but if it's going to be two booleans, than I'd prefer to see the implementation explicitly prevent "false and true" from having any effect (or better, an assertion that this is not the case).

michaelficarra: The proposal has two directives, each of which may be used independently. Four states is the exact number needed to represent these combinations. What is "false and true" and why is it not meaningful?

michaelficarra: I've renamed `ImplementationHidden` to `HasSourceTextAvailable` and renamed `OmittedFromStackTraces` to `PresentInStackTraces`.

ljharb: maybe I’m confused; i thought there were two pragmas in the proposal; is the intention to require both to be present, or does one subsume the other?

michaelficarra: There are two directives and two features. The `hide source` directive enables one feature, and the `sensitive` directive enables both features. So you could say that `sensitive` subsumes `hide source`.

ljharb: Right - so there's only three states:
1. no features
1. one feature (hide source)
1. both features (hide source, hide from stacks)

Two booleans represents four states, and I'd prefer to see either three (an enum) or an explicit assertion that makes it impossible for "hide from stacks but show source" to be a thing.

allenwb: Why are hide source and hide from stack coupled at all?  I have actually built a debugger (for another language) where the frames of  certain implementation generated functions were (by default) hidden from stack traces.  This was done to make a simpler presentation that avoided burdening the user will seeing those implementation details.  But hiding the stack frame had nothing to do with whether or not source code of such functions was accessible.  That should be an orthogonal option leaving open the possible that a function's source can be accessed even if it has hidden stack frames.

I think you should have two boolean valued slots associated with each function: [[HiddenSource]] and [[HiddenStack]] generated four states, each of which is meaningful.  Also isHiddenSpirce and isHiddenStack abstract operations. 

BTW,  the setting and access of these should exactly parallel the pattern used for [[Strict]]/isStrict.

ljharb: If the current proposal allows users to have that use case, then I agree that two booleans is the right way to implement it; however, it currently does not. When/if such a use case is supported, we can refactor the spec to use two booleans at that time.

allenwb: That sounds silly.  Fix the proposal to do the rational thing.

michaelficarra: @allenwb I'd appreciate your feedback on the proposal repo. We arrived at this design through careful consideration of the use cases we care about. See https://github.com/tc39/proposal-function-implementation-hiding/issues/35.

michaelficarra: @ljharb I've added an assertion, please check it out.

ljharb: Thanks, that resolves my concern!

ljharb (approved):
Although I think the pragma contents will still need some bikeshedding, from an editor standpoint, I'm good!

michaelficarra:
I'm happy to bikeshed names during plenary, though sad to waste committee time on it. Haven't gotten much feedback from other delegates on https://github.com/tc39/proposal-function-implementation-hiding/issues/3, unfortunately, so it would be irresponsible not to at least bring it up.

spec.html
```diff
@@ -8505,6 +8527,21 @@ <h1>OrdinaryFunctionCreate ( _functionPrototype_, _ParameterList_, _Body_, _this
         1. Set _F_.[[FormalParameters]] to _ParameterList_.
         1. Set _F_.[[ECMAScriptCode]] to _Body_.
         1. If the source text matching _Body_ is strict mode code, let _Strict_ be *true*; else let _Strict_ be *false*.
+        1. Set _F_.[[PresentInStackTraces]] to PresentInStackTraces of _Body_.
```
mikesamuel: Is the idea here to mirror the inheritance rules implicit in "is strict mode" but in a way that's simpler since these directives don't affect where future reserved words are allowed?

mikesamuel (commented):
Sorry for the delay.  Looks mostly good.  Some questions below.

spec.html
```diff
@@ -8505,6 +8527,21 @@ <h1>OrdinaryFunctionCreate ( _functionPrototype_, _ParameterList_, _Body_, _this
...
+          1. If _runningFunc_ is not *null*,
+            1. If _F_.[[PresentInStackTraces]] is *true*, set _F_.[[PresentInStackTraces]] to _runningFunc_.[[PresentInStackTraces]].
+            1. If _F_.[[HasSourceTextAvailable]] is *true*, set _F_.[[HasSourceTextAvailable]] to _runningFunc_.[[HasSourceTextAvailable]].
+          1. Let _scriptOrModule_ be _ec_'s ScriptOrModule.
+          1. If _scriptOrModule_ is not *null*,
+            1. If _F_.[[PresentInStackTraces]] is *true*, set _F_.[[PresentInStackTraces]] to PresentInStackTraces of _scriptOrModule_.[[ECMAScriptCode]].
+            1. If _F_.[[HasSourceTextAvailable]] is *true*, set _F_.[[HasSourceTextAvailable]] to HasSourceTextAvailable of _scriptOrModule_.[[ECMAScriptCode]].
+        1. Assert: If _F_.[[PresentInStackTraces]] is *false*, _F_.[[HasSourceTextAvailable]] is *false*.
+        1. Set _F_.[[Strict]] to _Strict_.
+        1. Set _F_.[[Environment]] to _Scope_.
```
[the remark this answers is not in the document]
michaelficarra: This is a mistake from a rebase. I'll fix it.

spec.html
michaelficarra: That's correct. Since we're not introducing any parse errors or early errors, we can implement it using the already-existing spec machinery for runtime behaviour. From [the README](https://github.com/tc39/proposal-function-implementation-hiding#the-solution):

> Similar to the "use strict" directive, these new directives apply "inclusively downward", so that everything within the scope, plus the function itself when in function scope, gets hidden.

spec.html
```diff
@@ -24459,7 +24585,7 @@ <h1>Runtime Semantics: PerformEval ( _x_, _callerRealm_, _strictCaller_, _direct
           1. If _strictEval_ is *true*, set _varEnv_ to _lexEnv_.
           1. If _runningContext_ is not already suspended, suspend _runningContext_.
           1. Let _evalContext_ be a new ECMAScript code execution context.
-          1. Set _evalContext_'s Function to *null*.
+          1. Set _evalContext_'s Function to _ctx_'s Function.
```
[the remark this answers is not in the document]
michaelficarra: It is necessary because we access this field here: https://github.com/tc39/ecma262/pull/1739/files#diff-3540caefa502006d8a33cb1385720803R8534

So eval code that creates a function should inherit the `"hide source"` and `"sensitive"` directives of the containing function:

```js
function f() {
  'hide source';
  return eval('(function g(){})').toString();
}
f(); // should be a string that matches NativeFunction
```

There may have been other spec bugs that were caused due to this field being set to `null`. I believe it was always meant to be inherited. But I haven't done an audit of all uses of the `Function` field of an execution context.

spec.html
```diff
@@ -26357,6 +26483,7 @@ <h1>Properties of Symbol Instances</h1>
   <emu-clause id="sec-error-objects">
     <h1>Error Objects</h1>
     <p>Instances of Error objects are thrown as exceptions when runtime errors occur. The Error objects may also serve as base objects for user-defined exception classes.</p>
+    <p>Stack trace information exposed to the running program through implementation-defined accessors such as the de facto `Error.prototype.stack` must not indicate the presence of functions whose [[PresentInStackTraces]] slot has a value of *false*. Additionally, stack frames from these accessors which refer to functions whose [[HasSourceTextAvailable]] slot has a value of *false* must not include any attribution or position information related to the function.</p>
```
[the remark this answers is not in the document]
michaelficarra: @mikesamuel I've added some examples in the latest commit. Do you think that's sufficient to clarify? Remember that this is not just about emulating a `NativeFunction`, but also about avoiding refactoring hazards.

spec.html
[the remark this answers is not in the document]
mikesamuel: Thanks for explaining.

spec.html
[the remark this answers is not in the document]
mikesamuel: Thanks.  That clarifies.

mikesamuel:
+1.  As a reviewer, I sign off on this for stage 3.

waldemarhorwat:
> @waldemarhorwat You're right, I think I had a bug. I've removed `GetDirectiveContext` in [19687c8](https://github.com/tc39/ecma262/commit/19687c88f35222d0903137beb88561d8b5c5f77d) since it wasn't really helpful anymore. Please see if that resolves your issues. If we don't use the execution context stack to track the directive state here, what alternative would you recommend? /cc @allenwb

As far as I can tell, the bug is fixed. The thing that makes me uneasy is that the use of a _running_ execution context to implement lexical scoping makes it difficult to follow what's happening here — I had to do a deep trace of code throughout the spec to find all possible things that could be on the running execution context stack when _OrdinaryFunctionCreate_ is called. Module and function lexical scopes push onto the execution context, but class lexical scopes don't, which is weird and not apparent without looking at the implementation. I'm with @allenwb here.

erights:
@waldemarhorwat , I would also like to see lexical concepts avoid needless dependencies on dynamic concepts. How should this be stated?

michaelficarra:
@waldemarhorwat Do you think we can avoid using the running execution context even while supporting code like this?

```js
function f(x) {
  'hide source';
  return eval('(function g() {\
    return x;\
  })');
}

f(0).toString(); // should match NativeFunction
```

allenwb:
> Do you think we can avoid using the running execution context even while supporting code like this?

Yes, in exactly the same way (and place) strict mode is handled for [direct eval](https://tc39.es/ecma262/#sec-function-calls-runtime-semantics-evaluation) 6.a.iv-vi

As I explained elsewhere on github recently (and which I now can not longer find, probably lost as a comment on another PR) there should be no significant difference in the handling of these new directives from how the strict mode directly is currently handled.  You can either replicated what is done for strict mode or you can generalize the strict mode pseudo-code to handle a larger set of directives.  But there is should be no new semantics here and the semantics of these new directives should not diverge from the semantics of the strict mode directive.

michaelficarra:
@waldemarhorwat @erights @allenwb I've update this PR to use static information for deriving `[[HasSourceTextAvailable]]` and `[[PresentInStackTraces]]`, as opposed to tracking that information at run time via the Execution Context stack. I think this is a nice improvement. Of course, `eval` still uses the running Execution Context, but I'm pretty sure that's inevitable. Let me know if this satisfies your concerns.

allenwb (changes requested):
In general, I think this update has move in the right direction.  However, eval code is still dependent upon the running execution context rather than being lexically based.  As I said in https://github.com/tc39/ecma262/pull/1739#issuecomment-561001106 "...there should be no significant difference in the handling of these new directives from how the strict mode direct[ive] is currently handled." But that isn't the case in this updated PR.

You have to take into account that `eval` is really two different things depending upon whether it is a direct eval or an indirect eval. Direct eval and indirect eval code have different lexical contexts (ignoring here for simplicity the additional differences  between strict and non-strict eval code) that are statically knowable without reference to the current execution environment (notice that while it is not always knowable that a particular `eval`-like expression is a direct or indirect eval (or an eval at all). its possible direct and indirect eval lexical contexts are statically knowable so all that needs to be determine at execution time is whether or not the `eval` is direct as is done in https://tc39.es/ecma262/#sec-function-calls-runtime-semantics-evaluation 6.a.iv-vi .

bakkot:
@allenwb 

> "...there should be no significant difference in the handling of these new directives from how the strict mode direct[ive] is currently handled." 

I disagree with this. The strict mode directive is currently handled by defining a new kind of code, in prose, in a way which has [caused pain in the past](https://github.com/tc39/ecma262/pull/1158). I do not think the appropriate way to introduce these directives is to add two new kinds of code.

I'm fine with changing this PR to be more static; I think that should be doable by introducing definitions for `PresentInStackTraces` and `HasSourceTextAvailable` of `CallExpression : CoverCallExpressionAndAsyncArrowHead` for use in the function call semantics for direct evals. But I don't think we should try to precisely match how `"use strict"` works.

allenwb:
> > "...there should be no significant difference in the handling of these new directives from how the strict mode direct[ive] is currently handled."
> 
> I disagree with this. The strict mode directive is currently handled by defining a new kind of code, in prose, in a way which has [caused pain in the past](https://github.com/tc39/ecma262/pull/1158). I do not think the appropriate way to introduce these directives is to add two new kinds of code.

Sorry, I should clarify. I wasn't talking about how you approach specifying the determination and semantics of the different kinds of directives.  I was talking about the rules (that a human needs to learn) concerning how the extent of the affect of directive is determined.  The extent of affect of a lexically enclosing ```'use strict'``` directive  upon an direct or indirect eval should be the same as the extent of affect of a lexically enclosing ```'hide source'``` (etc.) directive  on the same eval.  In other words, what a human needs to know to determining whether a ```'use strict'``` applies should be the same as what they need to know to determine if a ```'hide source'```applies. And we know, that for ```'use strict'``` this doesn't require any knowledge of the execution environment (other than determining whether an apparent direct eval really is one).

I don't care strongly about how exactly this is expressed in the specification but I agree that the kinds of code prose may be outliving its usefulness.

bakkot:
> In other words, what a human needs to know to determining whether a `'use strict'` applies should be the same as what they need to know to determine if a `'hide source'` applies. And we know, that for `'use strict'` this doesn't require any knowledge of the execution environment (other than determining whether an apparent direct eval really is one).

Oh, I see. That's always been true: in an earlier version of this PR the behavior was specified using knowledge of the execution environment, but that was just for editorial convenience; the actual semantics have always been static and have not changed across revisions of this PR. I had thought your concern was just about the editorial mechanism by which it was specified.

Anyway, sounds like we're on the same page.

michaelficarra:
@allenwb Thanks for the quick feedback. @Bakkot helped me to remove all usage of execution contexts. Can you review again to see if this is acceptable now?

allenwb (approved):

michaelficarra:
@waldemarhorwat or @erights Can you review the spec text here before the next meeting? Note that I will be asking to split the proposal and only advance the `"hide source"` directive.

erights:
I have just reviewed with this split in mind. The portions of this relating just to hiding source text, as well as the resulting suppression of position and source location within a stack frame but no more, LGTM.

This is still a Draft so this comment is adequate, yes?

michaelficarra:
@erights Yes that is perfect. Thank you!

erights:
I see at https://github.com/tc39/ecma262/pull/1739/files#r351043373 that @allenwb also prefers orthogonal "hide source" and "hide stack" directives. The resulting thread asks for a use case.

Here's one: When computing across a membrane, the stack frames of the membrane mechanism itself are often a distraction to the debugging experience. Debugger often do have better ways of handling the suppression of such unwanted detail. But for so-called printf debugging, looking at stacks dumped to logs, this could reduce the verbosity a lot. The Error stacks proposal will give better knobs for this as well.

My point is that I can imagine wanting to suppress stack traces for reasons that would not imply that I also want to suppress source.

Altogether, I am happy to see this go forward with only `"hide source";`. We should revisit the hiding of stack frames as part of the error stacks proposal.

spec.html
```diff
@@ -8353,6 +8353,28 @@ <h1>ECMAScript Function Objects</h1>
...
+        </tr>
+        <tr>
+          <td>
+            [[HasSourceTextAvailable]]
+          </td>
+          <td>
+            Boolean
+          </td>
+          <td>
+            When *false*, this function will be rendered as a |NativeFunction| by `Function.prototype.toString` (<emu-xref href="#sec-function.prototype.tostring"></emu-xref>) and corresponding stack trace entries will have no attribution or position information.
```
syg, line 8375: Not an editorial review comment, but, I'd prefer "have no attribution or position information" to be replaced with "have implementation-defined non-incidental attribution or position information", using the definition of incidental given below.

michaelficarra:
FYI I've split the proposal into two commits in this PR to illustrate the intended proposal split more clearly.

spec.html
```diff
@@ -13625,6 +13650,31 @@ <h1>Static Semantics: AssignmentTargetType</h1>
...
       </emu-clause>
+
+      <emu-clause id="sec-callexpression-static-semantics-presentinstacktraces">
+        <h1>Static Semantics: PresentInStackTraces</h1>
+        <emu-see-also-para op="PresentInStackTraces"></emu-see-also-para>
+        <emu-grammar>CallExpression : CoverCallExpressionAndAsyncArrowHead</emu-grammar>
+        <emu-alg>
+          1. If |CallExpression| occurs within a |ScriptBody|, |ModuleBody|, or |FunctionBody| that has a Directive Prologue that contains a Sensitive Directive, return *false*.
+          1. If the source text matched by |CallExpression| is eval code resulting from a direct eval, then
+            1. Return PresentInStackTraces of the |CallExpression| whose evaluation is the direct eval.
```
syg, line 13661: I think I get what you're trying to do here, but magically rematerializing the instance of the ancestor |CallExpression| that resulted in the particular direct eval that led to the instance of this |CallExpression| is harder to think about than I'd like. I wonder if there's a way to write this in a more obvious top-down manner, but the interleaving of static and run-time semantics definitely makes that very challenging.

One option is to have an argument-taking Static Semantics where direct eval can pass in the inherited "is sensitive" or "hide source" state. Not sure if that's a net editorial improvement...

spec.html
michaelficarra: How about just "... will have no incidental attribution or position information, as defined in (section reference)."?

spec.html
bakkot: > One option is to have an argument-taking Static Semantics where direct eval can pass in the inherited "is sensitive" or "hide source" state. Not sure if that's a net editorial improvement...

The problem is that we need the relevant bit at this point, during runtime semantics. Having it at the point that you are doing the `eval` call doesn't really help unless you propagate that information on some runtime construct like lexical environments.

spec.html
syg: That sgtm.

spec.html
syg: Hm, I see. If we can't come up with more obvious verbiage here, I think that sentence warrants some note that all this is doing is trying to reflect the "lexical" nesting of the directives across (space and?) time, due to direct evals.

spec.html
```diff
@@ -19616,6 +19642,19 @@ <h1>Static Semantics: ContainsUseStrict</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-function-definitions-static-semantics-hassourcetextavailable">
+      <h1>Static Semantics: HasSourceTextAvailable</h1>
+      <emu-see-also-para op="HasSourceTextAvailable"></emu-see-also-para>
+      <emu-grammar>FunctionBody : FunctionStatementList</emu-grammar>
+      <emu-alg>
+        1. If the Directive Prologue of |FunctionBody| contains a Sensitive Directive or a Hidden Implementation Directive, return *false*.
+        1. If |FunctionBody| occurs within a |ScriptBody|, |ModuleBody|, or |FunctionBody| that has a Directive Prologue that contains a Sensitive Directive or a Hidden Implementation Directive, return *false*.
```
```suggestion
        1. If the Directive Prologue of |FunctionBody| contains a Hidden Implementation Directive, return *false*.
        1. If |FunctionBody| occurs within a |ScriptBody|, |ModuleBody|, or |FunctionBody| that has a Directive Prologue that contains a Hidden Implementation Directive, return *false*.
```
also moved to the second commit

spec.html
```diff
@@ -20016,6 +20055,18 @@ <h1>Static Semantics: ContainsUseStrict</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-arrow-function-definitions-static-semantics-hassourcetextavailable">
+      <h1>Static Semantics: HasSourceTextAvailable</h1>
+      <emu-see-also-para op="HasSourceTextAvailable"></emu-see-also-para>
+      <emu-grammar>ExpressionBody : AssignmentExpression</emu-grammar>
+      <emu-alg>
+        1. If |ExpressionBody| occurs within a |ScriptBody|, |ModuleBody|, or |FunctionBody| that has a Directive Prologue that contains a Sensitive Directive or a Hidden Implementation Directive, return *false*.
```
```suggestion
        1. If |ExpressionBody| occurs within a |ScriptBody|, |ModuleBody|, or |FunctionBody| that has a Directive Prologue that contains a Hidden Implementation Directive, return *false*.
```
also moved to the second commit

ljharb (commented):
first commit LGTM otherwise

spec.html
bakkot: We could add a NOTE reading something like

> NOTE: Sensitive Directives apply to nested functions, including to code within a direct call to `eval`.

I'm not in love with that wording but don't have a better one off the top of my head.

spec.html
```diff
@@ -490,7 +490,7 @@ <h1>The Numeric String Grammar</h1>
...
+      <p>When a parse is successful, it constructs a <em>parse tree</em>, a rooted tree structure in which each node is a <dfn>Parse Node</dfn>. Each Parse Node is an <em>instance</em> of a symbol in the grammar; it represents a span of the source text that can be derived from that symbol. The root node of the parse tree, representing the whole of the source text, is an instance of the parse's goal symbol. When a Parse Node is an instance of a nonterminal, it is also an instance of some production that has that nonterminal as its left-hand side. Moreover, it has zero or more <em>children</em>, one for each symbol on the production's right-hand side: each child is a Parse Node that is an instance of the corresponding symbol. A Parse Node _A_ <dfn>occurs within</dfn> another Parse Node _B_ if _A_ is the child of _B_ or if _A_ is the child of some Parse Node _C_ where _C_ occurs within _B_.</p>
```
bakkot, line 493: There are 13 existing uses of the phrase "occurs within", none of which match this definition but all of which get linked as a consequence of defining it here (e.g. "[Every other finite time value t is defined relative to the greatest preceding time value s that is such a multiple, and represents the instant that **occurs within** the same UTC day as s but follows it by t − s milliseconds](https://ci.tc39.es/preview/tc39/ecma262/sha/9c3c9cc8554487aa24d49c12c78e1bb976a6a6e1/#sec-time-values-and-time-range)".)

I think it's probably fine not to mark this as a definition.

spec.html
```diff
@@ -19426,12 +19450,14 @@ <h2>Syntax</h2>
...
       <p>A <dfn id="use-strict-directive">Use Strict Directive</dfn> is an |ExpressionStatement| in a Directive Prologue whose |StringLiteral| is either of the exact code point sequences `"use strict"` or `'use strict'`. A Use Strict Directive may not contain an |EscapeSequence| or |LineContinuation|.</p>
-      <p>A Directive Prologue may contain more than one Use Strict Directive. However, an implementation may issue a warning if this occurs.</p>
+      <p>A <dfn id="hidden-implementation-directive">Hidden Implementation Directive</dfn> is an |ExpressionStatement| in a Directive Prologue whose |StringLiteral| is either of the exact code point sequences `"hide source"` or `'hide source'`. A Hidden Implementation Directive may not contain an |EscapeSequence| or |LineContinuation|.</p>
+      <p>A <dfn id="built-in-directive">Built-in Directive</dfn> is an |ExpressionStatement| in a Directive Prologue that is either a Use Strict Directive, or a Hidden Implementation Directive.</p>
```
```suggestion
      <p>A <dfn id="built-in-directive">Built-in Directive</dfn> is an |ExpressionStatement| in a Directive Prologue that is either a Use Strict Directive or a Hidden Implementation Directive.</p>
```

spec.html
michaelficarra: I would like our new use of it to be linked though. How do I do that?

spec.html
michaelficarra: Fixed.

spec.html
jmdyck: Well, you could use an explicit `<emu-xref>` on the occurrences you want to link, or an `<emu-not-ref>` on the ones you don't want to link, but I think both of those approaches are sub-optimal. It would be better to come up with a different (distinct) phrase for this sense of "occurs within". E.g., "is a descendant of" or "descends from": either is a natural extension of the "child" terminology already established for parse trees, and neither occurs anywhere else in the spec.

PR #1740, opened 2019-10-17
rename cxt/ctx variables to context

michaelficarra:
We were using both `cxt` and `ctx` as a shortening of "context" in variable names. To avoid this inconsistency, I just replaced them with `context`.


michaelficarra:
I pushed a commit that removes some weird "the"s as well. I pushed it here since a second PR would be horribly conflicting, but I can split it out if that's preferred.

PR #1745, opened 2019-10-17
add missing ContainsUseStrict definition for AsyncConciseBody

michaelficarra:
At first, I noticed that there was a missing `ContainsUseStrict` definition for `AsyncConciseBody`, which was used in an early error, so I added the missing definition. ~But then I noticed that there was no way for the early error condition to be `true`, so I removed the early error and the `ContainsUseStrict` definitions became dead, so I removed those, too.~

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -21287,6 +21287,15 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-async-arrow-function-definitions-static-semantics-containsusestrict">
+      <h1>Static Semantics: ContainsUseStrict</h1>
+      <emu-see-also-para op="ContainsUseStrict"></emu-see-also-para>
+      <emu-grammar>AsyncConciseBody : AssignmentExpression</emu-grammar>
+      <emu-alg>
+        1. Return *false*.
+      </emu-alg>
+    </emu-clause>
+
     <emu-clause id="sec-async-arrow-function-definitions-static-semantics-ExpectedArgumentCount">
       <h1>Static Semantics: ExpectedArgumentCount</h1>
       <emu-grammar>


spec.html
```diff
@@ -19527,9 +19527,6 @@ <h1>Static Semantics: Early Errors</h1>
         <li>
           It is a Syntax Error if |ArrowParameters| Contains |AwaitExpression| is *true*.
         </li>
-        <li>
-          It is a Syntax Error if ContainsUseStrict of |ConciseBody| is *true* and IsSimpleParameterList of |ArrowParameters| is *false*.
```
bakkot: How is this dead?  [`ContainsUseStrict` of `FunctionBody`](https://tc39.es/ecma262/#sec-function-definitions-static-semantics-containsusestrict) can return `true`, and [`ConciseBody` can produce `{ FunctionBody }`](https://tc39.es/ecma262/#prod-ConciseBody). By the [chain rule](https://tc39.es/ecma262/#sec-algorithm-conventions-syntax-directed-operations), then, `ContainsUseStrict of |ConciseBody|` can return `true`.

Same for the other early error, except with one more intermediate step in the chain rule (`AsyncFunctionBody` -> `FunctionBody`).

michaelficarra: Ah, I saw `ConciseBody : AssignmentExpression` and thought that was the only alternative, but forgot about the `{ FunctionBody }` case. Yeah, the chain rule should make the early errors not dead. I guess I'll just drop the second commit then and have this PR add the `ContainsUseStrict` definition for `AsyncConciseBody : AssignmentExpression`.

ljharb:
Can you elaborate on why ContainsUseStrict would be always false in the following Script code?:
```js
var obj = {
  async foo() {
    return this === globalThis;
  }
};
(0, obj.foo)().then(x => assert(x === true))
```

michaelficarra:
@ljharb It is always false for `ConciseBody` and `AsyncConciseBody` nodes that parse as `AssignmentExpression`s. This is because they do not contain a `FunctionBody` with a Directive Prologue that contains a Use Strict Directive.

The code you posted contains a single example of a `ConciseBody`: `assert(x === true)`. This example does not contain a Use Strict Directive.

ljharb:
Ah, thanks for the clarification. So specifically, `const f = async () => this; (await f()) === window` will always be true, because there's no way to add "use strict" in it?

So is this PR, then, adding the async analogue for https://tc39.es/ecma262/#sec-arrow-function-definitions-static-semantics-containsusestrict?

michaelficarra:
Yes, it is doing exactly that @ljharb. It looks like the definition of `ContainsUseStrict` was just overlooked when adding the async arrow feature.

michaelficarra:
@syg Haha I had thought the same thing at first, and even pushed up a commit that did that as part of this PR. See https://github.com/tc39/ecma262/pull/1745#discussion_r336108469.

syg:
> @syg Haha I had thought the same thing at first, and even pushed up a commit that did that as part of this PR. See [#1745 (comment)](https://github.com/tc39/ecma262/pull/1745#discussion_r336108469).

Humbled by the chain rule!

ljharb:
I think this is normative, but also an obvious spec bug, so it doesn't need consensus.

bakkot:
(In this case I would have called it editorial because the previous specification literally did not mean anything. This doesn't change the meaning of anything, rather it makes it meaningful in the first place.)

ljharb:
It's a grey area to be sure; ether way, nobody could possibly have implemented it a wrong way, so it's unobservable.

PR #1746, opened 2019-10-17
find Directive Prologue in appropriate node in IsStrict/ContainsUsestrict

michaelficarra:
Two things:

1. Define `IsStrict` for `Script`, since that's how it's applied.
1. Refer to the directive prologue of a `ScriptBody`/`FunctionBody` since that's how it's defined.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19147,7 +19147,7 @@
       <emu-see-also-para op="ContainsUseStrict"></emu-see-also-para>
       <emu-grammar>FunctionBody : FunctionStatementList</emu-grammar>
       <emu-alg>
-        1. If the Directive Prologue of |FunctionStatementList| contains a Use Strict Directive, return *true*; otherwise, return *false*.
+        1. If the Directive Prologue of |FunctionBody| contains a Use Strict Directive, return *true*; otherwise, return *false*.
       </emu-alg>
     </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -21851,9 +21851,9 @@ <h1>Static Semantics: Early Errors</h1>
 
     <emu-clause id="sec-static-semantics-isstrict">
       <h1>Static Semantics: IsStrict</h1>
-      <emu-grammar>ScriptBody : StatementList</emu-grammar>
+      <emu-grammar>Script : ScriptBody</emu-grammar>
```
```suggestion
      <emu-grammar>Script : ScriptBody?</emu-grammar>
```
(I assume that's what you meant, otherwise the alg wouldn't be checking if `ScriptBody` is present.)

michaelficarra: Sure, though I think it'll be rendered the same.

spec.html
```diff
@@ -19180,7 +19180,7 @@ <h1>Static Semantics: ContainsUseStrict</h1>
       <emu-see-also-para op="ContainsUseStrict"></emu-see-also-para>
       <emu-grammar>FunctionBody : FunctionStatementList</emu-grammar>
       <emu-alg>
-        1. If the Directive Prologue of |FunctionStatementList| contains a Use Strict Directive, return *true*; otherwise, return *false*.
+        1. If the Directive Prologue of |FunctionBody| contains a Use Strict Directive, return *true*; otherwise, return *false*.
```
ljharb: just to document my understanding: https://tc39.es/ecma262/#directive-prologue indicates that only FunctionBody, ScriptBody, or ModuleBody can have a Directive Prologue, so this is a bug fix.

michaelficarra: Correct.

PR #1748, opened 2019-10-17
Revert "Editorial: remove context copy from AsyncFunctionStart (#1685)"

syg:
This reverts commit e9faaa18253f2d6dc063f4cb3c4cf481e8f4f849.

I now think my analysis in #1685 of removing the copy was incorrect. After discussion with @devsnek on IRC, the issue comes down to this step:

4. Resume the suspended evaluation of _asyncContext_. Let _result_ be the value returned by the resumed computation.

That seems to assume that _asyncContext_ can be "executed" and its _result_ observed without overwriting the current "execution" of AsyncFunctionStart. If _asyncContext_ isn't a copied context but instead **is** the current context, I think that step is ill-formed, it'd be like a jump or tail call: there's no way to get the _result_ as required by that step.

All in all it seems clearer to me to revert.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -40125,6 +40125,7 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _runningContext_ be the running execution context.
           1. Let _asyncContext_ be a copy of _runningContext_.
+          1. NOTE: Copying the execution state is required for the step below to resume its execution. It is ill-defined to resume a currently executing context.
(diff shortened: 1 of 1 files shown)


syg:
For completeness, I think it's possible to avoid a copy here and use a single execution context, but I'm afraid it'll be even less clear than duplication.

I think that would involve some conditional popping of the async execution context depending on whether the function ran to completion after awaiting something (i.e. if [[Call]] already popped the execution context for the initial invocation), or without awaiting anything (i.e. if [[Call]]  is going to pop the initial execution context).

PR #1752, opened 2019-10-24
Editorial: Refactor index checking for Integer-Indexed exotic objects

syg:
This refactor was prompted by the confusing **-0** check in Integer-Indexed exotic objects' [[DefineOwnProperty]] trap. Usually the index validity checking is done in IntegerIndexedElementGet and IntegerIndexedElementSet, but that trap does some extra checks because it might not fall through to IntegerIndexedElementSet due to a non-data property descriptor.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9354,7 +9354,7 @@
         <p>When the [[GetOwnProperty]] internal method of an Integer-Indexed exotic object _O_ is called with property key _P_, the following steps are taken:</p>
         <emu-alg>
           1. Assert: IsPropertyKey(_P_) is *true*.
-          1. Assert: _O_ is an Object that has a [[ViewedArrayBuffer]] internal slot.
+          1. Assert: _O_ is an Integer-Indexed exotic object.
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
@@ -9370,16 +9370,13 @@
         <p>When the [[HasProperty]] internal method of an Integer-Indexed exotic object _O_ is called with property key _P_, the following steps are taken:</p>
         <emu-alg>
           1. Assert: IsPropertyKey(_P_) is *true*.
-          1. Assert: _O_ is an Object that has a [[ViewedArrayBuffer]] internal slot.
+          1. Assert: _O_ is an Integer-Indexed exotic object.
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
               1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
               1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
-              1. If IsInteger(_numericIndex_) is *false*, return *false*.
-              1. If _numericIndex_ = *-0*, return *false*.
-              1. If _numericIndex_ &lt; 0, return *false*.
-              1. If _numericIndex_ &ge; _O_.[[ArrayLength]], return *false*.
+              1. If ! IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *false*.
               1. Return *true*.
           1. Return ? OrdinaryHasProperty(_O_, _P_).
         </emu-alg>
@@ -9390,15 +9387,11 @@
         <p>When the [[DefineOwnProperty]] internal method of an Integer-Indexed exotic object _O_ is called with property key _P_, and Property Descriptor _Desc_, the following steps are taken:</p>
         <emu-alg>
           1. Assert: IsPropertyKey(_P_) is *true*.
-          1. Assert: _O_ is an Object that has a [[ViewedArrayBuffer]] internal slot.
+          1. Assert: _O_ is an Integer-Indexed exotic object.
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
-              1. If IsInteger(_numericIndex_) is *false*, return *false*.
-              1. If _numericIndex_ = *-0*, return *false*.
-              1. If _numericIndex_ &lt; 0, return *false*.
-              1. Let _length_ be _O_.[[ArrayLength]].
-              1. If _numericIndex_ &ge; _length_, return *false*.
+              1. If ! IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *false*.
               1. If IsAccessorDescriptor(_Desc_) is *true*, return *false*.
               1. If _Desc_ has a [[Configurable]] field and if _Desc_.[[Configurable]] is *true*, return *false*.
               1. If _Desc_ has an [[Enumerable]] field and if _Desc_.[[Enumerable]] is *false*, return *false*.
@@ -9442,7 +9435,7 @@
         <p>When the [[OwnPropertyKeys]] internal method of an Integer-Indexed exotic object _O_ is called, the following steps are taken:</p>
         <emu-alg>
           1. Let _keys_ be a new empty List.
-          1. Assert: _O_ is an Object that has [[ViewedArrayBuffer]], [[ArrayLength]], [[ByteOffset]], [[ContentType]], and [[TypedArrayName]] internal slots.
+          1. Assert: _O_ is an Integer-Indexed exotic object.
           1. Let _len_ be _O_.[[ArrayLength]].
           1. For each integer _i_ starting with 0 such that _i_ &lt; _len_, in ascending order, do
             1. Add ! ToString(_i_) as the last element of _keys_.
@@ -9473,18 +9466,28 @@
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-isvalidintegerindex" aoid="IsValidIntegerIndex">
+        <h1>IsValidIntegerIndex ( _O_, _index_ )</h1>
+        <p>The abstract operation IsValidIntegerIndex with arguments _O_ and _index_ performs the following steps:</p>
+        <emu-alg>
+          1. Assert: _O_ is an Integer-Indexed exotic object.
+          1. Assert: Type(_index_) is Number.
+          1. If ! IsInteger(_index_) is *false*, return *false*.
+          1. If _index_ is *-0*, return *false*.
+          1. If _index_ &lt; 0 or _index_ &ge; _O_.[[ArrayLength]], return *false*.
+          1. Return *true*.
+        </emu-alg>
+      </emu-clause>
+
       <emu-clause id="sec-integerindexedelementget" aoid="IntegerIndexedElementGet">
         <h1>IntegerIndexedElementGet ( _O_, _index_ )</h1>
         <p>The abstract operation IntegerIndexedElementGet with arguments _O_ and _index_ performs the following steps:</p>
         <emu-alg>
+          1. Assert: _O_ is an Integer-Indexed exotic object.
           1. Assert: Type(_index_) is Number.
-          1. Assert: _O_ is an Object that has [[ViewedArrayBuffer]], [[ArrayLength]], [[ByteOffset]], [[ContentType]], and [[TypedArrayName]] internal slots.
           1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
           1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
-          1. If IsInteger(_index_) is *false*, return *undefined*.
-          1. If _index_ = *-0*, return *undefined*.
-          1. Let _length_ be _O_.[[ArrayLength]].
-          1. If _index_ &lt; 0 or _index_ &ge; _length_, return *undefined*.
+          1. If ! IsValidIntegerIndex(_O_, _index_) is *false*, return *undefined*.
           1. Let _offset_ be _O_.[[ByteOffset]].
           1. Let _arrayTypeName_ be the String value of _O_.[[TypedArrayName]].
           1. Let _elementSize_ be the Element Size value specified in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -9480,18 +9473,27 @@ <h1>IntegerIndexedObjectCreate ( _prototype_, _internalSlotsList_ )</h1>
...
+          1. If _index_ &lt; 0 or _index_ &ge; _O_.[[ArrayLength]], return *false*.
+          1. Return *true*.
+        </emu-alg>
+      </emu-clause>
+
       <emu-clause id="sec-integerindexedelementget" aoid="IntegerIndexedElementGet">
         <h1>IntegerIndexedElementGet ( _O_, _index_ )</h1>
         <p>The abstract operation IntegerIndexedElementGet with arguments _O_ and _index_ performs the following steps:</p>
         <emu-alg>
-          1. Assert: Type(_index_) is Number.
```
ljharb: why is this assert no longer valid? (ideally all argument types are asserted at the top of every abstract operation)

spec.html
```diff
@@ -9505,16 +9507,12 @@ <h1>IntegerIndexedElementGet ( _O_, _index_ )</h1>
         <h1>IntegerIndexedElementSet ( _O_, _index_, _value_ )</h1>
         <p>The abstract operation IntegerIndexedElementSet with arguments _O_, _index_, and _value_ performs the following steps:</p>
         <emu-alg>
-          1. Assert: Type(_index_) is Number.
```
ljharb: why is this assert no longer valid?

spec.html
```diff
@@ -9401,18 +9398,14 @@ <h1>[[DefineOwnProperty]] ( _P_, _Desc_ )</h1>
...
-              1. Let _length_ be _O_.[[ArrayLength]].
-              1. If _numericIndex_ &ge; _length_, return *false*.
               1. If IsAccessorDescriptor(_Desc_) is *true*, return *false*.
               1. If _Desc_ has a [[Configurable]] field and if _Desc_.[[Configurable]] is *true*, return *false*.
               1. If _Desc_ has an [[Enumerable]] field and if _Desc_.[[Enumerable]] is *false*, return *false*.
               1. If _Desc_ has a [[Writable]] field and if _Desc_.[[Writable]] is *false*, return *false*.
               1. If _Desc_ has a [[Value]] field, then
                 1. Let _value_ be _Desc_.[[Value]].
                 1. Return ? IntegerIndexedElementSet(_O_, _numericIndex_, _value_).
+              1. If IsIntegerIndexValid(_O_, _numericIndex_) is *false*, return *false*.
```
ljharb: per @anba's comment, i think this needs to move up to line 9401?

syg:
The asserts are still valid indeed, I'll reinstate them.

spec.html
syg: Yes, will move this back up.

syg:
Addressed review.

spec.html
```diff
@@ -9376,10 +9376,7 @@ <h1>[[HasProperty]] ( _P_ )</h1>
             1. If _numericIndex_ is not *undefined*, then
               1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
               1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
-              1. If IsInteger(_numericIndex_) is *false*, return *false*.
-              1. If _numericIndex_ = *-0*, return *false*.
-              1. If _numericIndex_ &lt; 0, return *false*.
-              1. If _numericIndex_ &ge; _O_.[[ArrayLength]], return *false*.
+              1. If IsIntegerIndexValid(_O_, _numericIndex_) *false*, return *false*.
```
```suggestion
              1. If IsIntegerIndexValid(_O_, _numericIndex_) is *false*, return *false*.
```

spec.html
```diff
@@ -9473,6 +9466,19 @@ <h1>IntegerIndexedObjectCreate ( _prototype_, _internalSlotsList_ )</h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-integerindexedelementget" aoid="IntegerIndexedElementGet">
+        <h1>IsIntegerIndexValid ( _O_, _index_ )</h1>
+        <p>The abstract operation IsIntegerIndexValid with arguments _O_ and _index_ performs the following steps:</p>
+        <emu-alg>
+          1. Assert: Type(_index_) is Number.
+          1. Assert: _O_ is an Object that has an [[ArrayLength]] internal slot.
```
ljharb: is this a shorthand for:
```suggestion
          1. Assert: _O_ is an Integer-Indexed exotic object.
```
?

if so, can we use that definition?

spec.html
```diff
@@ -9473,6 +9466,19 @@ <h1>IntegerIndexedObjectCreate ( _prototype_, _internalSlotsList_ )</h1>
...
+          1. If IsInteger(_index_) is *false*, return *false*.
+          1. If _index_ = *-0*, return *false*.
```
```suggestion
          1. If _index_ is *-0*, return *false*.
```
perhaps?

syg: There is a well-established precedence for using `=` to check for mathematical values, though *-0* is kind of a corner case. My preference is to keep the `=`.

spec.html
syg: I can't commit the suggestion for some reason, will force-push.

spec.html
syg: It's a more relaxed check than asserting something is an Integer-Indexed exotic object, which I take to mean it would assert that the object has *all* the internal slots as well as the Integer-Indexed internal methods.

The spec has a convention to only do the brand check that it needs, and I think that flexibility is nice for writing future specs. I prefer to keep it to just the [[ArrayLength]] assert.

ljharb: To me it seems more brittle to future changes. Is there anything that would have that internal slot and not have all the others?

Additionally, the auto linking to the definition would likely be helpful.

syg: I don't understand the brittleness concern. Isn't it more brittle (less flexible) to assert that it is an IIEO? Do we have precedence for that kind of assertion over checking for specific slots?

jmdyck: Some precendent, yes. E.g., checking whether something is:
- a Bound Function exotic object (vs having `[[BoundTargetFunction]]`, `[[BoundThis]]`, `[[BoundArguments]]` slots),
- a Proxy [exotic] object (vs having `[[ProxyHandler]]`, `[[ProxyTarget]]` slots),
- an AsyncGenerator instance (vs having `[[AsyncGenerator{State,Context,Queue}]]` slots),
- an ECMAScript function [object] (vs having the slots of Table 29)

jmdyck: (However, checking for the presence/absence of particular slots is about 5 times more popular.)

ljharb: I read these assertions like type annotations; imo we should be checking for a nominal type here, not a structural one.

spec.html
ljharb: There's 5 cases of `= -0` in the spec right now, and 81 of `is -0`, so precedent seems to land on the side of the suggestion. I'd be happy to clean up the wayward 5.

spec.html
syg: Good to know there's precedence for both.

@ljharb and @jmdyck, do you think all exotic object checking should be implemented via "is a Foo exotic object"?

In general I like the explicit asserting of slots approach better, because all "exotic object" means is that it has special behavior for one of the essential internal methods. That it might have a set of distinctive internal slots is probably an implementation necessity for the exotic behavior, not an intrinsic property of being an exotic object. What do you think?

spec.html
syg: Sure, sounds fine to me.

spec.html
ljharb: I think that if the special behavior appears on multiple kinds of exotic object, it's definitely appropriate to assert on that special behavior alone. However, if there's only one kind of thing with that behavior, it's clearer to assert that you have a special kind of thing everywhere, so that if in the future new things are added with the behavior, the refactorings of all relevant usage sites are forced and explicit.

syg: > However, if there's only one kind of thing with that behavior, it's clearer to assert that you have a special kind of thing everywhere, so that if in the future new things are added with the behavior, the refactorings of all relevant usage sites are forced and explicit.

I see editorial arguments both ways. Doing slot checks (and structural checking) is more in line with the ethos of having generic abstract operations that "just work" for future things, i.e. better compositionality. Doing "is Foo exotic object" checks is more in line with the ethos of narrower operations, which may guard against undesirable genericity.

For this particular PR, TypedArray operations are most definitely not generic, so I'm fine with changing the asserts in that section to "is an IIEO".

ljharb: Happy to continue considering case by case, for sure.

spec.html
```diff
@@ -9473,18 +9466,28 @@ <h1>IntegerIndexedObjectCreate ( _prototype_, _internalSlotsList_ )</h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-integerindexedelementget" aoid="IntegerIndexedElementGet">
+        <h1>IsIntegerIndexValid ( _O_, _index_ )</h1>
```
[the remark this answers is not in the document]
syg: Sure, why not.

spec.html
```diff
@@ -9473,18 +9466,28 @@ <h1>IntegerIndexedObjectCreate ( _prototype_, _internalSlotsList_ )</h1>
...
+        <p>The abstract operation IsIntegerIndexValid with arguments _O_ and _index_ performs the following steps:</p>
+        <emu-alg>
+          1. Assert: Type(_index_) is Number.
+          1. Assert: _O_ is an Integer-Indexed exotic object.
```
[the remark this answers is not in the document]
syg: I'll reorder all the relevant asserts.

spec.html
```diff
@@ -9473,18 +9466,28 @@ <h1>IntegerIndexedObjectCreate ( _prototype_, _internalSlotsList_ )</h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-isvalidintegerindex" aoid="IsValidIntegerIndex">
+        <h1>IsValidIntegerIndex ( _O_, _index_ )</h1>
+        <p>The abstract operation IsValidIntegerIndex with arguments _O_ and _index_ performs the following steps:</p>
+        <emu-alg>
+          1. Assert: _O_ is an Integer-Indexed exotic object.
+          1. Assert: Type(_index_) is Number.
+          1. If IsInteger(_index_) is *false*, return *false*.
```
```suggestion
          1. If ! IsInteger(_index_) is *false*, return *false*.
```

spec.html
```diff
@@ -9473,18 +9466,28 @@ <h1>IntegerIndexedObjectCreate ( _prototype_, _internalSlotsList_ )</h1>
...
+          1. Assert: _O_ is an Integer-Indexed exotic object.
           1. Assert: Type(_index_) is Number.
-          1. Assert: _O_ is an Object that has [[ViewedArrayBuffer]], [[ArrayLength]], [[ByteOffset]], [[ContentType]], and [[TypedArrayName]] internal slots.
           1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
           1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
-          1. If IsInteger(_index_) is *false*, return *undefined*.
-          1. If _index_ = *-0*, return *undefined*.
-          1. Let _length_ be _O_.[[ArrayLength]].
-          1. If _index_ &lt; 0 or _index_ &ge; _length_, return *undefined*.
+          1. If IsValidIntegerIndex(_O_, _index_) is *false*, return *undefined*.
```
```suggestion
          1. If ! IsValidIntegerIndex(_O_, _index_) is *false*, return *undefined*.
```

spec.html
```diff
@@ -9498,16 +9501,13 @@ <h1>IntegerIndexedElementGet ( _O_, _index_ )</h1>
...
-          1. Assert: _O_ is an Object that has [[ViewedArrayBuffer]], [[ArrayLength]], [[ByteOffset]], [[ContentType]], and [[TypedArrayName]] internal slots.
           1. If _O_.[[ContentType]] is ~BigInt~, let _numValue_ be ? ToBigInt(_value_).
           1. Otherwise, let _numValue_ be ? ToNumber(_value_).
           1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
           1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
-          1. If IsInteger(_index_) is *false*, return *false*.
-          1. If _index_ = *-0*, return *false*.
-          1. Let _length_ be _O_.[[ArrayLength]].
-          1. If _index_ &lt; 0 or _index_ &ge; _length_, return *false*.
+          1. If IsValidIntegerIndex(_O_, _index_) is *false*, return *false*.
```
```suggestion
          1. If ! IsValidIntegerIndex(_O_, _index_) is *false*, return *false*.
```

spec.html
```diff
@@ -9390,15 +9387,11 @@ <h1>[[DefineOwnProperty]] ( _P_, _Desc_ )</h1>
...
+          1. Assert: _O_ is an Integer-Indexed exotic object.
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
-              1. If IsInteger(_numericIndex_) is *false*, return *false*.
-              1. If _numericIndex_ = *-0*, return *false*.
-              1. If _numericIndex_ &lt; 0, return *false*.
-              1. Let _length_ be _O_.[[ArrayLength]].
-              1. If _numericIndex_ &ge; _length_, return *false*.
+              1. If IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *false*.
```
```suggestion
              1. If ! IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *false*.
```

spec.html
```diff
@@ -9370,16 +9370,13 @@ <h1>[[HasProperty]] ( _P_ )</h1>
...
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
               1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
               1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
-              1. If IsInteger(_numericIndex_) is *false*, return *false*.
-              1. If _numericIndex_ = *-0*, return *false*.
-              1. If _numericIndex_ &lt; 0, return *false*.
-              1. If _numericIndex_ &ge; _O_.[[ArrayLength]], return *false*.
+              1. If IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *false*.
```
```suggestion
              1. If ! IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *false*.
```

syg:
@ljharb You know, in general I don't agree with the editorial direction that ! be prefixed on invocations of AOs that cannot throw because they do not return completion values. ! and ? telegraph to me that the AO returns a completion value, since those are shorthands for unwrapping, and I'd prefer to use ! only for an AO that *can* throw, but *will not* for that particular invocation.

That said, there is so much prior use of ! in the way I disagree with, I'll incorporate your desired changes here.

ljharb:
Every abstract op always returns a completion value, implicitly.

The drive for using ! or ? at callsites is that, otherwise, you can't possibly *know* if an abstract op call "can throw" or not, so the `!` gives you the only info you need: either that it can't throw, or that it won't throw with these arguments. Whether it can throw in other cases isn't helpful information.

PR #1757, opened 2019-10-26
spec: corrects the TypeError description as its name represent

yorkie:
The `TypeError` should own itself purpose just like the title described, and as for default
case, the `Error` is better choice.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -26471,7 +26471,7 @@
 
       <emu-clause id="sec-native-error-types-used-in-this-standard-typeerror">
         <h1>TypeError</h1>
-        <p>TypeError is used to indicate an unsuccessful operation when none of the other _NativeError_ objects are an appropriate indication of the failure cause.</p>
+        <p>TypeError is used to indicate that a value is not of the expected type.</p>
       </emu-clause>
 
       <emu-clause id="sec-native-error-types-used-in-this-standard-urierror">


bathos:
This is accurate as currently written. Intrinsic APIs and platform APIs do not throw `Error`, they only throw _NativeError_ objects (or, in the case of platform APIs, DOMException, etc).

It does generally work out that the remainder of cases not covered by the other _NativeErrors_ are related in some way to type expectations (though otoh it’s not always cut and dry what constitutes ‘related to type expectations’ in ES).

yorkie:
> This is accurate as currently written. Intrinsic APIs and platform APIs do not throw Error, they only throw NativeError objects (or, in the case of platform APIs, DOMException, etc).

Using `TypeError` as the default is incorrect in some ways, I think `NativeError` or a new `UnknownError` should be introduced here and let the `TypeError` only accept the type exceptions.

yorkie:
And see [`TypeError` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError), it's worse a bit for developers that's different between MDN and ECMA262 Spec :(

ljharb:
@yorkie that sounds like a new feature proposal, and this repo isn't the place for those - see https://github.com/tc39/ecma262/blob/master/CONTRIBUTING.md#creating-a-new-proposal.

yorkie:
Thank you @ljharb, I have created a topic at https://es.discourse.group/t/nativeerror-introduce-a-new-unknownerror-instead-of-current-typeerror/114, looking forward any update :)

yorkie:
And closing this PR because moved to discourse, thanks all.

bathos:
> different between MDN and ECMA262

The text in ECMA262 is describing the situation for implementors rather than prescribing usage for end-users of the language. Perhaps there is an editorial change (as opposed to feature req) that would be useful here if that was not clear.

yorkie:
Other spec like [web-bluetooth-spec](https://github.com/WebBluetoothCG/web-bluetooth) would make use of the ECMA262 definitions, this may be spread to user-land, for example, the end-user have to handle `TypeError`s when calling those exposed APIs.

bathos:
Yes. There’s a layer in between, though, since it’s Web IDL that defines how [exceptions are used within platform APIs](https://heycam.github.io/webidl/#idl-exceptions) and provides the definition of DOMException (which isn’t part of ES) and its extensive [name enumeration](https://heycam.github.io/webidl/#idl-DOMException-error-names).

PR #1769, opened 2019-11-02
Editorial: Add BigInt to the list of primitive values

chicoxyzzy:
[BigInt](https://github.com/tc39/ecma262/pull/1515) is missing in the list of primitive values

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -154,7 +154,7 @@
   <emu-clause id="sec-ecmascript-overview">
     <h1>ECMAScript Overview</h1>
     <p>The following is an informal overview of ECMAScript&mdash;not all parts of the language are described. This overview is not part of the standard proper.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -332,6 +332,21 @@ <h1>Number object</h1>
...
+    </emu-clause>
+
+    <emu-clause id="sec-terms-and-definitions-bigint-type">
+      <h1>BigInt type</h1>
+      <p>set of all possible BigInt values</p>
+    </emu-clause>
+
+    <emu-clause id="sec-bigint-object">
+      <h1>BigInt object</h1>
+      <p>member of the Object type that is an instance of the standard built-in `BigInt` constructor</p>
```
bakkot: "String object" and "Number object" have NOTEs under them like

> A String object is created by using the String constructor in a new expression, supplying a String value as an argument. The resulting object has an internal slot whose value is the String value. A String object can be coerced to a String value by calling the String constructor as a function (21.1.1.1).

Is there a reason to omit a similar note here?

spec.html
```diff
@@ -332,6 +332,21 @@ <h1>Number object</h1>
       </emu-note>
     </emu-clause>
 
+    <emu-clause id="sec-terms-and-definitions-bigint-value">
+      <h1>BigInt value</h1>
```
bakkot: This placement separates the Number type from the various instances of the Number type (like Infinity). I think it would be better to place this after the definition of NaN.

spec.html
chicoxyzzy: BigInt can't be created via `new`, same as Symbol which doesn't have such note. I'm not sure what text should be in the note for BigInt.

PR #1770, opened 2019-11-06
Editorial: Update IEEE 754 standard version

### LINKED ISSUE #1650: A new edition of IEEE-754 has been published
Note that  IEEE 754-2019 has been published.

Changes summarized in http://754r.ucbtest.org/background/ 

Probably nothing that impacts ECMA-262 but we may want to update any normative references.

chicoxyzzy:
Closes #1650

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -326,7 +326,7 @@
 
     <emu-clause id="sec-terms-and-definitions-number-value">
       <h1>Number value</h1>
-      <p>primitive value corresponding to a double-precision 64-bit binary format IEEE 754-2008 value</p>
+      <p>primitive value corresponding to a double-precision 64-bit binary format IEEE 754-2019 value</p>
       <emu-note>
         <p>A Number value is a member of the Number type and is a direct representation of a number.</p>
       </emu-note>
@@ -352,7 +352,7 @@
 
     <emu-clause id="sec-terms-and-definitions-nan">
       <h1>NaN</h1>
-      <p>Number value that is an IEEE 754-2008 &ldquo;Not-a-Number&rdquo; value</p>
+      <p>Number value that is an IEEE 754-2019 &ldquo;Not-a-Number&rdquo; value</p>
     </emu-clause>
 
     <emu-clause id="sec-terms-and-definitions-bigint-value">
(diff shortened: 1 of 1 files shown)


PrimerMediaGroup:
Rename Spacebookclub.com so we don't lose our profile pictures and information

michaelficarra:
If there's nothing that affects observable behaviour of ECMAScript programs, what's the motivation to update the normative reference? Sell more copies of IEEE-754?

apaprocki:
To the contrary, the observable behavior is the fact that the reference has updated with no change, signaling to readers that the spec is and was compatible with both versions. Otherwise, someone could ask themselves “wait, why the *old* standard?” and spend some amount of time finding the answer.

michaelficarra:
@apaprocki Then, for the same reason, we should mention that either will do.

apaprocki:
What do you think about normatively referencing the 2019 standard, and then providing a "conformance note" (those words) that a 2008 conformant implementation is also acceptable?

Whether or not that extends to say something along the lines of "there are no normative changes between IEEE-754:2008 and IEEE-754:2019 that affect the ECMA-262 specification" is up to the editors.

chicoxyzzy:
Should this note be added to Bibliography?

@ljharb IPR check is failing. Do I need to do something with this?

michaelficarra:
@apaprocki I would strongly prefer the latter wording.

ljharb:
@chicoxyzzy no; it's not a required check. it currently fails on every PR because there's no way to get encrypted credentials safely into that script. I'm working on removing it from PRs (but leaving it on master) until that's solved; ignore it in the meantime.

chicoxyzzy:
@michaelficarra where in spec do you think we should put a note about IEEE 754-2008? It seems to me that the Bibliography section is not the proper place.

syg:
@chicoxyzzy I think a sensible place to put the note that either 2008 or 2019 versions are sufficient is in the bibliography. To make that discoverable, I was thinking existing mentions of IEEE 754 could be linked to jump to the biblio entry. What do you think?

chicoxyzzy:
@syg sounds good. I'll update the PR on weekend.

chicoxyzzy:
- [x] add a note
- [x] add links to bibliography

spec.html
```diff
@@ -43232,8 +43232,11 @@ <h1>Colophon</h1>
 <emu-annex id="sec-bibliography">
   <h1>Bibliography</h1>
   <ol>
-    <li>
-      IEEE Std 754-2008: <i>IEEE Standard for Floating-Point Arithmetic</i>. Institute of Electrical and Electronic Engineers, New York (2008)
+    <li id="sec-bibliography-ieee754">
```
ljharb: i wonder if instead of an li id, we could have a `<dfn>` around `IEEE Std 754-2019` and then it would get auto-linked?

chicoxyzzy: I didn't know about `<dfn>`. Fixed!

PR #1773, opened 2019-11-11
Editorial: Add BigInt to the list of reference base value types and the language types intro text

bathos:
Similar to https://github.com/tc39/ecma262/pull/1769 — BigInt was absent from the list of types that the base value component of a Reference could be.

Note that it is already included and handled as a possible base value in the [HasPrimitiveBase abstract op](https://tc39.es/ecma262/#sec-hasprimitivebase).

I was unsure if there was any particular logic to the ordering of types, so I just slotted it in the ‘most alphabetic’ seeming part of the list. Alternatively, the sentence could be reworded to avoid listing them all if it’s felt that this is becoming an update hazard, e.g.

> The base value component is either an Environment Record or a language type other than Null.

Or, if it’s desired to draw additional attention to undefined’s distinct significance here:

> The base value component is either an Environment Record or a language type other than Null (but including *undefined*).

(ref #1515)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -921,7 +921,7 @@
 
   <emu-clause id="sec-ecmascript-language-types">
     <h1>ECMAScript Language Types</h1>
-    <p>An <dfn>ECMAScript language type</dfn> corresponds to values that are directly manipulated by an ECMAScript programmer using the ECMAScript language. The ECMAScript language types are Undefined, Null, Boolean, String, Symbol, Number, and Object. An <dfn>ECMAScript language value</dfn> is a value that is characterized by an ECMAScript language type.</p>
+    <p>An <dfn>ECMAScript language type</dfn> corresponds to values that are directly manipulated by an ECMAScript programmer using the ECMAScript language. The ECMAScript language types are Undefined, Null, Boolean, String, Symbol, Number, BigInt, and Object. An <dfn>ECMAScript language value</dfn> is a value that is characterized by an ECMAScript language type.</p>
 
     <emu-clause id="sec-ecmascript-language-types-undefined-type">
       <h1>The Undefined Type</h1>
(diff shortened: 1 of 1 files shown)


claudepache:
> Alternatively, this could be reworded to avoid listing them all, e.g.
> 
> > The base value component is either an Environment Record or a language type other than Null.

`undefined` plays a special role which could have better been modelled as `empty`. IMO, a better rewording would be:

> The base value component is either *undefined*, or an [ECMAScript language type] other than Undefined and Null, or an Environment Record.

[ECMAScript language type]: https://tc39.es/ecma262/#sec-ecmascript-language-types

claudepache:
BTW, another place where BigInt is still missing, is in the list of types found in the introductory pragraph of Section 6.1 [ECMAScript language type](https://tc39.es/ecma262/#sec-ecmascript-language-types).

bathos:
Thanks @claudepache, I’ve added the missing one you mentioned. For now I’ve left the full enumeration in the base value list to keep the change minimal, but will change it if it turns out there’s general agreement that a briefer version would be better.

spec.html
```diff
@@ -921,7 +921,7 @@ <h1>ECMAScript Data Types and Values</h1>
 
   <emu-clause id="sec-ecmascript-language-types">
     <h1>ECMAScript Language Types</h1>
-    <p>An <dfn>ECMAScript language type</dfn> corresponds to values that are directly manipulated by an ECMAScript programmer using the ECMAScript language. The ECMAScript language types are Undefined, Null, Boolean, String, Symbol, Number, and Object. An <dfn>ECMAScript language value</dfn> is a value that is characterized by an ECMAScript language type.</p>
+    <p>An <dfn>ECMAScript language type</dfn> corresponds to values that are directly manipulated by an ECMAScript programmer using the ECMAScript language. The ECMAScript language types are Undefined, Null, BigInt, Boolean, String, Symbol, Number, and Object. An <dfn>ECMAScript language value</dfn> is a value that is characterized by an ECMAScript language type.</p>
```
```suggestion
    <p>An <dfn>ECMAScript language type</dfn> corresponds to values that are directly manipulated by an ECMAScript programmer using the ECMAScript language. The ECMAScript language types are Undefined, Null, Boolean, String, Symbol, Number, BigInt, and Object. An <dfn>ECMAScript language value</dfn> is a value that is characterized by an ECMAScript language type.</p>
```
minor, but this seems the better sorting to me :-)

spec.html
```diff
@@ -4101,7 +4101,7 @@ <h1>The Reference Specification Type</h1>
...
       </emu-note>
-      <p>A <dfn>Reference</dfn> is a resolved name or property binding. A Reference consists of three components, the base value component, the referenced name component, and the Boolean-valued strict reference flag. The base value component is either *undefined*, an Object, a Boolean, a String, a Symbol, a Number, or an Environment Record. A base value component of *undefined* indicates that the Reference could not be resolved to a binding. The referenced name component is a String or Symbol value.</p>
+      <p>A <dfn>Reference</dfn> is a resolved name or property binding. A Reference consists of three components, the base value component, the referenced name component, and the Boolean-valued strict reference flag. The base value component is either *undefined*, an Object, a BigInt, a Boolean, a String, a Symbol, a Number, or an Environment Record. A base value component of *undefined* indicates that the Reference could not be resolved to a binding. The referenced name component is a String or Symbol value.</p>
```
```suggestion
      <p>A <dfn>Reference</dfn> is a resolved name or property binding. A Reference consists of three components, the base value component, the referenced name component, and the Boolean-valued strict reference flag. The base value component is either *undefined*, an Object, a Boolean, a String, a Symbol, a Number, a BigInt, or an Environment Record. A base value component of *undefined* indicates that the Reference could not be resolved to a binding. The referenced name component is a String or Symbol value.</p>
```

spec.html
bakkot: @ljharb's proposed order matches the order under [ECMAScript Language Types](https://tc39.es/ecma262/#sec-ecmascript-language-types), which seems like a desirable property.

bakkot:
It is also missing from the first step of [CreateListFromArrayLike](https://tc39.es/ecma262/#sec-createlistfromarraylike). I've checked the rest of the spec and found no other omissions.

caiolima:
> BTW, another place where BigInt is still missing, is in the list of types found in the introductory paragraph of Section 6.1 [ECMAScript language type](https://tc39.es/ecma262/#sec-ecmascript-language-types).

This is being defined as sub-section of [Numeric Types](https://tc39.es/ecma262/#sec-numeric-types). More precisely, it is defined  in [The BigInt Type](https://tc39.es/ecma262/#sec-ecmascript-language-types-bigint-type).

caiolima:
Thank you very much for this PR!

PR #1780, opened 2019-11-14
Editorial: Tweak casing of "empty String" and "Number value"

shvaikalesh:
There is minor inconsistency with casing: 63 occurrences of "empty **S**tring", 15 of "empty **s**tring". Should we uppercase all the "string"s in this PR to avoid any possible doubt on primitive vs object wrapper?

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -328,18 +328,18 @@
       <h1>Number object</h1>
       <p>member of the Object type that is an instance of the standard built-in `Number` constructor</p>
       <emu-note>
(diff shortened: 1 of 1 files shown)


ljharb:
If we're removing doubt, we'd lowercase them all, since `string` says "primitive" to me more than `String` does.

ljharb:
An alternative would be replacing all the "empty string"s with `*""*`.

michaelficarra:
I prefer "the empty String".

shvaikalesh:
> An alternative would be replacing all the "empty string"s with `*""*`.

I am recalling we were replacing `""` with "empty string": https://github.com/tc39/ecma262/pull/1411#discussion_r248525913.

Regarding casing: we should be consistent with other primitives and `Type` helper (e.g. `Assert: Type(x) is String`).

ljharb:
Fair enough - let's change all of them to `the empty String`, then, in case there's any more missing.

PR #1791, opened 2019-11-23
Normative: specify for-in enumeration order in more cases

bakkot:
This PR implements the [for-in-order](https://github.com/tc39/proposal-for-in-order) proposal, currently at stage 3.

I intend to request stage 4 at the December 2019 meeting.

[Here](https://github.com/tc39/test262/pull/2432) is the PR adding tests to test262.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3052,6 +3052,16 @@
             </tr>
             <tr>
               <td>
+                %ForInIteratorPrototype%
+              </td>
+              <td>
+              </td>
+              <td>
+                The prototype of For-In iterator objects (<emu-xref href="#sec-for-in-iterator-objects"></emu-xref>)
+              </td>
+            </tr>
+            <tr>
+              <td>
                 %Function%
               </td>
               <td>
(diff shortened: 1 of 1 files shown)


ljharb:
("has tests" applies after the test262 tests are merged)

spec.html
```diff
@@ -17739,8 +17749,17 @@ <h1>EnumerateObjectProperties ( _O_ )</h1>
...
         <p>Enumerating the properties of the target object includes enumerating properties of its prototype, and the prototype of the prototype, and so on, recursively; but a property of a prototype is not processed if it has the same name as a property that has already been processed by the iterator's `next` method. The values of [[Enumerable]] attributes are not considered when determining if a property of a prototype object has already been processed. The enumerable property names of prototype objects must be obtained by invoking EnumerateObjectProperties passing the prototype object as the argument. EnumerateObjectProperties must obtain the own property keys of the target object by calling its [[OwnPropertyKeys]] internal method. Property attributes of the target object must be obtained by calling its [[GetOwnProperty]] internal method.</p>
+        <p>In addition, if neither _O_ nor any object in its prototype chain is a Proxy exotic object, Integer-Indexed exotic object, Module Namespace exotic object, or implementation provided exotic object, then the iterator must behave as would the iterator given by CreateForInIterator ( _O_ ) until one of the following occurs:</p>
```
```suggestion
        <p>In addition, if neither _O_ nor any object in its prototype chain is a Proxy exotic object, Integer-Indexed exotic object, Module Namespace exotic object, or implementation-provided exotic object, then the iterator must behave as would the iterator given by CreateForInIterator ( _O_ ) until one of the following occurs:</p>
```
not sure tho.

spec.html
```diff
@@ -17760,6 +17779,121 @@ <h1>EnumerateObjectProperties ( _O_ )</h1>
             }
           </code></pre>
         </emu-note>
+        <emu-note>
+          The list of exotic objects for which hosts are not required to match CreateForInIterator was chosen because hosts historically differed in behavior for those cases, and agreed in all others.
+        </emu-note>
+      </emu-clause>
+
+      <emu-clause id="sec-for-in-iterator-objects">
+        <h1>For In Iterator Objects</h1>
```
ljharb: would it be worth saying "For-In" for clarity that it's a single token? i feel like i've commonly seen this representation when describing the construct, to disambiguate from prose.

spec.html
bakkot: Existing uses of "implementation provided" ([e.g.](https://tc39.es/ecma262/#sec-invariants-of-the-essential-internal-methods)) do not hyphenate, so this PR just matches them.

spec.html
```diff
@@ -17760,6 +17779,121 @@ <h1>EnumerateObjectProperties ( _O_ )</h1>
...
+                1. Set _O_.[[ObjectWasVisited]] to *false*.
+                1. If _object_ is *null*, return CreateIterResultObject(*undefined*, *true*).
+            </emu-alg>
+          </emu-clause>
+        </emu-clause>
+
+        <emu-clause id="sec-properties-of-for-in-iterator-instances">
+          <h1>Properties of For In Iterator Instances</h1>
+          <p>For In Iterator instances are ordinary objects that inherit properties from the %ForInIteratorPrototype% intrinsic object. For In Iterator instances are initially created with the internal slots listed in <emu-xref href="#table-for-in-iterator-instance-slots"></emu-xref>.</p>
+              <emu-table id="table-for-in-iterator-instance-slots" caption="Internal Slots of For In Iterator Instances">
```
bakkot: Note to self: there's extra indentation here.

spec.html
```diff
@@ -17760,6 +17779,121 @@ <h1>EnumerateObjectProperties ( _O_ )</h1>
             }
           </code></pre>
         </emu-note>
+        <emu-note>
+          The list of exotic objects for which hosts are not required to match CreateForInIterator was chosen because hosts historically differed in behavior for those cases, and agreed in all others.
```
syg: "hosts" -> "implementations and host environments". The list above has both exotic objects that an implementation must provide independent of hosts, and host-provided.

spec.html
```diff
@@ -17760,6 +17779,121 @@ <h1>EnumerateObjectProperties ( _O_ )</h1>
...
+                <td>
+                  A list of String values which have been emitted by this iterator thus far.
+                </td>
+              </tr>
+              <tr>
+                <td>
+                  [[RemainingKeys]]
+                </td>
+                <td>
+                  A list of String values remaining to be emitted for the current object, before iterating those on its prototype (if its prototype is not *null*).
```
syg: "before iterating the properties of its prototype"? The antecedent of "those" reads like it's String values, which could be clearer.

spec.html
bakkot: Hm. I am happy to change to "implementations", but I think it's confusing to have "implementations and host environments": at least in my taxonomy, implementations are what provide the semantics of AOs like `EnumerateObjectProperties`, even when they are invoked on host-provided objects.

bakkot:
Addressed comments; thanks for the feedback.

spec.html
syg: Just "implementations" sgtm.

spec.html
```diff
@@ -17739,8 +17749,17 @@ <h1>EnumerateObjectProperties ( _O_ )</h1>
...
+          <li>the value of the [[Prototype]] internal slot of _O_ or an object in its prototype chain changes,</li>
+          <li>a property is removed from _O_ or an object in its prototype chain,</li>
+          <li>a property is added to an object in _O_'s prototype chain, or</li>
+          <li>the value of the [[Enumerable]] attribute of a property of _O_ or an object in its prototype chain changes.</li>
+        </ul>
+
         <emu-note>
-          <p>The following is an informative definition of an ECMAScript generator function that conforms to these rules:</p>
+          <p>Hosts are not required to implement the algorithm in <emu-xref href="#sec-%foriniteratorprototype%.next"></emu-xref> directly. They may choose any implementation whose behavior will not deviate from that algorithm unless one of the constraints in the previous paragraph is violated.</p>
+          <p>The following is an informative definition of an ECMAScript generator function that conforms to these rules:</ins></p>
```
```suggestion
          <p>The following is an informative definition of an ECMAScript generator function that conforms to these rules:</p>
```

PR #1797, opened 2019-11-28
Editorial: remove AssertionTester from RexExp semantics

### LINKED ISSUE #1795: RegExp Lookahead and lookbehind routines have the wrong signature
I'm currently looking at the regexp specifications.

`Term :: Assertion` evaluates Assertion to get an AssertionTester: https://tc39.es/ecma262/#sec-term

The beginning-of-input and end-of-input and word boundary Assertions do exactly that.

The lookahead and lookbehind Assertions, however, return Matchers. (https://tc39.es/ecma262/#sec-assertion)

This seems .... wrong. Some kind of translation needs to happen to move from states to booleans? (Probably: not-failure -> True, failure -> False), though a Matcher takes a continuation and an AssertionTester doesn't so there's a bit more that ought to change there.

Or am I just missing something?

Relevant spec portions:
> The production Term::Assertion evaluates as follows:
>
> 1. Return an internal Matcher closure that takes two arguments, a State x and a Continuation c, and performs the following steps:
> a. Evaluate Assertion to obtain an AssertionTester t.
> b. Call t(x) and let r be the resulting Boolean value.
> c. If r is false, return failure.
> d Call c(x) and return its result.

> The production Assertion::$ evaluates as follows:
>
> 1. Return an internal AssertionTester closure that takes a State argument x and performs the following steps:
> a. Let e be x's endIndex.
> b. If e is equal to InputLength, return true.
> c. If Multiline is false, return false.
> d. If the character Input[e] is one of LineTerminator, return true.
> e. Return false.

> The production Assertion::(?=Disjunction) evaluates as follows:
> 
> 1. Evaluate Disjunction with +1 as its direction argument to obtain a Matcher m.
> 2. Return an internal Matcher closure that takes two arguments, a State x and a Continuation c, and performs the following steps:
> a. Let d be a Continuation that always returns its State argument as a successful MatchResult.
> b. Call m(x, d) and let r be its result.
> c. If r is failure, return failure.
> d. Let y be r's State.
> e. Let cap be y's captures List.
> f. Let xe be x's endIndex.
> g. Let z be the State (xe, cap).
> h. Call c(z) and return its result.

bakkot:
Fixes #1795. cc @scole66, @claudepache, @mathiasbynens.

Marked as editorial because the previous spec text was nonsensical as written.

The first commit is all that is strictly necessary to resolve the issue. The second is just an improvement in clarity (to my eyes).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -31409,9 +31409,6 @@ THH:mm:ss.sss
           <li>
             A <em>Matcher</em> procedure is an internal closure that takes two arguments&mdash;a State and a Continuation&mdash;and returns a MatchResult result. A Matcher attempts to match a middle subpattern (specified by the closure's already-bound arguments) of the pattern against _Input_, starting at the intermediate state given by its State argument. The Continuation argument should be a closure that matches the rest of the pattern. After matching the subpattern of a pattern to obtain a new State, the Matcher then calls Continuation on that new State to test if the rest of the pattern can match as well. If it can, the Matcher returns the State returned by Continuation; if not, the Matcher may try different choices at its choice points, repeatedly calling Continuation until it either succeeds or all possibilities have been exhausted.
           </li>
-          <li>
-            An <em>AssertionTester</em> procedure is an internal closure that takes a State argument and returns a Boolean result. The assertion tester tests a specific condition (specified by the closure's already-bound arguments) against the current place in _Input_ and returns *true* if the condition matched or *false* if not.
-          </li>
         </ul>
       </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -31100,44 +31093,42 @@ <h1>Runtime Semantics: RepeatMatcher ( _m_, _min_, _max_, _greedy_, _x_, _c_, _p
...
+          1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps:
             1. Let _e_ be _x_'s _endIndex_.
             1. Call IsWordChar(_e_ - 1) and let _a_ be the Boolean result.
             1. Call IsWordChar(_e_) and let _b_ be the Boolean result.
-            1. If _a_ is *true* and _b_ is *false*, return *true*.
-            1. If _a_ is *false* and _b_ is *true*, return *true*.
-            1. Return *false*.
+            1. If _a_ is *true* and _b_ is *false*, or if _a_ is *false* and _b_ is *true*, then
+              1. Call _c_(_x_) and return its result.
+            1. Return *failure*.
```
claudepache: Should be: `~failure~`

spec.html
```diff
@@ -31100,44 +31093,42 @@ <h1>Runtime Semantics: RepeatMatcher ( _m_, _min_, _max_, _greedy_, _x_, _c_, _p
...
+          1. Return an internal Matcher closure that takes two arguments, a State _x_ and a Continuation _c_, and performs the following steps:
             1. Let _e_ be _x_'s _endIndex_.
             1. Call IsWordChar(_e_ - 1) and let _a_ be the Boolean result.
             1. Call IsWordChar(_e_) and let _b_ be the Boolean result.
-            1. If _a_ is *true* and _b_ is *false*, return *false*.
-            1. If _a_ is *false* and _b_ is *true*, return *false*.
-            1. Return *true*.
+            1. If _a_ is *true* and _b_ is *true*, or if _a_ is *false* and _b_ is *false*, then
+              1. Call _c_(_x_) and return its result.
+            1. Return *failure*.
```
claudepache: Ditto, should be: `~failure~`

spec.html
bakkot: Thanks, fixed.

PR #1799, opened 2019-12-02
WIP: Meta: add script to land PRs

ljharb:
This is a beginning attempt.

In the future I'll add more checks; this should ensure that all commit messages either have a valid prefix, or (if not the first commit), `squash: ` or `fixup: `.

### DIFF
--- a/.github/workflows/check-commit.yml
+++ b/.github/workflows/check-commit.yml
@@ -1,0 +1,17 @@
+name: 262 Commit Checker
+
+on: pull_request
+
+jobs:
+   test:
+    name: Check commit messages
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+      with:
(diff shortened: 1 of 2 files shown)


ljharb:
Failing: https://github.com/tc39/ecma262/pull/1799/checks?check_run_id=330098116
Passing: https://github.com/tc39/ecma262/pull/1799/checks?check_run_id=330132746
(altho the failing one no longer shows up; suffice to say it works)

bakkot:
I would strongly prefer we not do this.

ljharb:
I'm confused as to why, but we can discuss it during the next editor call.

ljharb:
As discussed; I'll repurpose this PR to contain a "land a PR" script, which will run these validations instead of doing it on each PR.

(My understanding of the objection, which I realize I agree with, is that PR status checks should only fail when the PR author can take action to correct them)

ljharb:
TODO: Document git procedure for merging here

PR #1800, opened 2019-12-03
Editorial: adding ids to individual algorithms of equality semantics

brabalan:
This PR adds ids to the different `emu-alg` of the [equality semantics](https://tc39.es/ecma262/#sec-equality-operators-runtime-semantics-evaluation). This does not change the way the specification is rendered.

I have two questions about this PR:
- Is it a good approach to have links to algorithms when there are several in a section?
- What should be the naming convention for these links?


michaelficarra:
We spoke about this in the editor call. We'd prefer we just add support to ecmarkup for this so we don't continue adding one-off IDs and so we maintain a consistent naming convention.

PR #1062, opened 2018-01-11
Editorial: Add Interesting Cases of Automatic Semicolon Insertion

bmeck:
Unsure on how we want to put @littledan as original author.

PR #1808, opened 2019-12-12
Editorial: Improve readability in 11.10.

rkirsling:
Just a small wordsmithing follow-up to #1062:

We have two sentences in a row saying effectively the same thing in different words, namely, the definition of "interesting". In fact, the previous paragraph already sets up the same concept (surprising reparses) for a different purpose (_new_ productions instead of existing ones) so in addition to eliminating a sentence here, we can also make a cleaner transition into the new paragraph.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12260,7 +12260,7 @@
 
     <p>As new syntactic features are added to ECMAScript, additional grammar productions could be added that cause lines relying on automatic semicolon insertion preceding them to change grammar productions when parsed.</p>
 
-    <p>The interesting cases of automatic semicolon insertion are places where a semicolon may or may not be inserted depending on preceding source text according to the rules above. These places are considered interesting if changes outside of an existing source text grammar production could change the grammar production of the existing source text depending on automatic semicolon insertion. The rest of this section describes a number of interesting cases of automatic semicolon insertion in this version of ECMAScript.</p>
+    <p>For the purposes of this section, a case of automatic semicolon insertion is considered interesting if it is a place where a semicolon may or may not be inserted, depending on the source text which precedes it. The rest of this section describes a number of interesting cases of automatic semicolon insertion in this version of ECMAScript.</p>
     <emu-clause id="sec-asi-interesting-cases-in-statement-lists">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -12123,7 +12123,7 @@ <h1>Interesting Cases of Automatic Semicolon Insertion</h1>
 
     <p>As new syntactic features are added to ECMAScript, additional grammar productions could be added that cause lines relying on automatic semicolon insertion preceding them to change grammar productions when parsed.</p>
 
-    <p>The interesting cases of automatic semicolon insertion are places where a semicolon may or may not be inserted depending on preceding source text according to the rules above. These places are considered interesting if changes outside of an existing source text grammar production could change the grammar production of the existing source text depending on automatic semicolon insertion. The rest of this section describes a number of interesting cases of automatic semicolon insertion in this version of ECMAScript.</p>
+    <p>The interesting cases of automatic semicolon insertion are thus places where a semicolon may or may not be inserted, depending on the source text which precedes it. The rest of this section describes a number of interesting cases of automatic semicolon insertion in this version of ECMAScript.</p>
```
ljharb: one possibility:
```suggestion
    <p>Places where a semicolon may or may not be inserted depending on preceding source text according to the rules above are considered interesting if changes outside of an existing source text grammar production could change the grammar production of the existing source text depending on automatic semicolon insertion. The rest of this section describes a number of interesting cases of automatic semicolon insertion in this version of ECMAScript.</p>
```

bakkot: The previous wording was effectively defining "interesting", which I would prefer to keep. For example:
```suggestion
    <p>For the purposes of this section a case of automatic semicolon insertion is considered interesting if it is a place where a semicolon may or may not be inserted, depending on the source text which precedes it. The rest of this section describes a number of interesting cases of automatic semicolon insertion in this version of ECMAScript.</p>
```

syg: +1 to this suggestion.

rkirsling: I was hoping the single word "thus" sufficed to convey that, but more explicit is good. :smile:

rkirsling: Added.

PR #1821, opened 2019-12-27
Editorial: Add aoid for CaseBlockEvaluation so it gets auto-linked / ref-counted

bathos:
CaseBlockEvaluation is referenced in step 6 of the [eval algorithm for switch statements](https://tc39.es/ecma262/#sec-switch-statement-runtime-semantics-evaluation), but was unlinked. I think this is the right way to fix this?

image

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -18282,7 +18282,7 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-runtime-semantics-caseblockevaluation">
+    <emu-clause id="sec-runtime-semantics-caseblockevaluation" aoid="CaseBlockEvaluation">
       <h1>Runtime Semantics: CaseBlockEvaluation</h1>
       <p>With parameter _input_.</p>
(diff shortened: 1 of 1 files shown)


jmdyck:
(Note that this approach doesn't work for all syntax-directed operations, because over half of them are defined by more than one `<emu-clause>`, and it wouldn't really make sense to pick just one of those clauses to get the `aoid` attribute. Still, about 40 of them *are* defined in a single clause, so handling them would probably be worth the effort.)

bathos:
@jmdyck ah, I see what you mean.

It would be nice if there were a way to address it even in multiple-definitions cases. If multiple emu-clauses provide same-named definitions, I would think that linking out from them via the References popup would still be possible. Links to them also seems plausible with a bit of indirection: if more than one def exists, the same UI pattern as is used for the References lists could be used. That’d probably be a sizable effort though, and I’m not sure if others rely on the reference links as much as I do, so I don’t know if it’d be worth it.

bakkot:
I think we'll do this via #1950 instead (when we get to it).

PR #1828, opened 2020-01-03
Editorial: Lookahead single-element shorthand permits terminal sequences

bathos:
Currently, the [Grammar Notation](https://tc39.es/ecma262/#sec-grammar-notation) section describes when the `≠` shorthand is permitted for `∉` as follows:

image

(With similar text following for `=` / `∈`.)

The shorthand for single members is used with terminal sequences, though:

image

That makes sense, given the set notation permits sequences, but it means the description in Grammar Notation is incorrect. It may just be a typo in origin.

This change would align the descriptions with how this notation is used in practice later in the spec. It preserves the existing constraint that the sequences in question be composed of terminals only.

---

(Regarding markup: I found I needed to write `_terminal_ _sequence_` instead of `_terminal sequence_`, and it’s maybe a bit odd that these italicized terms should be the ‘click to highlight’ sort since this isn’t algorithmic text. Should they be marked up using `<em>` or `<i>` instead maybe?)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -671,7 +671,7 @@
           `9`
       </emu-grammar>
       <p>If the phrase &ldquo;[empty]&rdquo; appears as the right-hand side of a production, it indicates that the production's right-hand side contains no terminals or nonterminals.</p>
-      <p>If the phrase &ldquo;[lookahead &notin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may not be used if the immediately following input token sequence is a member of the given _set_. The _set_ can be written as a comma separated list of one or two element terminal sequences enclosed in curly brackets. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all terminals to which that nonterminal could expand. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead &ne; _terminal_]&rdquo; may be used.</p>
+      <p>If the phrase &ldquo;[lookahead &notin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may not be used if the immediately following input token sequence is a member of the given _set_. The _set_ can be written as a comma separated list of one or two element terminal sequences enclosed in curly brackets. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all terminals to which that nonterminal could expand. If the _set_ consists of a single terminal sequence the phrase &ldquo;[lookahead &ne; _terminal_ _sequence_]&rdquo; may be used.</p>
       <p>For example, given the definitions:</p>
       <emu-grammar type="example">
         DecimalDigit :: one of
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -671,7 +671,7 @@ <h1>Grammar Notation</h1>
...
+      <p>If the phrase &ldquo;[lookahead &notin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may not be used if the immediately following input token sequence is a member of the given _set_. The _set_ can be written as a comma separated list of one or two element terminal sequences enclosed in curly brackets. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all terminals to which that nonterminal could expand. If the _set_ consists of a single terminal sequence the phrase &ldquo;[lookahead &ne; _terminal_ _sequence_]&rdquo; may be used.</p>
```
rkirsling: I wonder if it should be _terminalSequence_ instead? Or is there a precedent for the space?

bathos: It said ‘terminal sequences’ in the first paragraph already when describing set members. I don’t think it occurs anywhere else so it’s not much of a precedent. Camel cased, I’d kind of expect it to be a defined term — currently it’s just directly descriptive, meaning just ‘a sequence of terminals’.

rkirsling: I just mean that `_this_` in ecmarkup denotes a [variable](https://bterlson.github.io/ecmarkup/#emd-overview). Sure, this one's ad hoc and meant to be self-explanatory, but it doesn't seem right to stick a space inside it.

bathos: Makes sense & I agree. I’ll change it to a camelcase name if you think that’s the best course. Personally, I tend to think these aren’t vars in the usual sense and shouldn’t be marked up with `_underscores_` at all (see last paragraph of first message) but that seemed too opinionated of a change for me to make, so I figured I was stuck with the two words because that’s how it’s described in the preceding sentences.

rkirsling: Sure. I'll defer to what others think, but assuming we keep the underscores, anything without a space is okay by me.

syg: Agree with @rkirsling. The easiest mental model seems to be that a single variable expands to a sequence of a terminals.

spec.html
```diff
@@ -688,7 +688,7 @@ <h1>Grammar Notation</h1>
           DecimalDigit [lookahead &lt;! DecimalDigit]
       </emu-grammar>
       <p>matches either the letter `n` followed by one or more decimal digits the first of which is even, or a decimal digit not followed by another decimal digit.</p>
-      <p>Similarly, if the phrase &ldquo;[lookahead &isin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the immediately following input token sequence is a member of the given _set_. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead = _terminal_]&rdquo; may be used.</p>
+      <p>Similarly, if the phrase &ldquo;[lookahead &isin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the immediately following input token sequence is a member of the given _set_. If the _set_ consists of a single terminal sequence the phrase &ldquo;[lookahead = _terminal_ _sequence_]&rdquo; may be used.</p>
```
syg: Ditto on single variable.

bakkot:
Thanks for raising this! I believe I've addressed it (alongside other changes) as part of #2254, so I'm closing this PR. If you see further issues please open a new issue or PR.

PR #1830, opened 2020-01-04
Editorial: Update About this Specification wording

sideshowbarker:
This change updates the wording of the first sentence in the "About this Specification" section to read:

> This document at https://tc39.es/ecma262/ is the most accurate and up-to-date ECMAScript specification.

Without this change, it just reads “the most accurate up-to-date”, without the word “and”.


ljharb:
Can you elaborate on why this change? "accurate and up-to-date" sounds better to me than "accurate up-to-date" (altho both are correct, i think).

ljharb:
aha, i see you're *adding* "and". In that case I'm on board, as it does read more clearly.

PR #1831, opened 2020-01-04
Editorial: Clean up some inline notes.

rkirsling:
This PR cleans up a couple of notes which are inlined into algorithms:
- Move notes that share a line with a normal algorithm step to their own line. (x3)
- Un-inline a final note which refers to the algorithm as a whole. (x1)
- Fix grammar. (x2)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1990,7 +1990,8 @@
               1. Assert: _op_ is *"^"*.
               1. Let _tmp_ be BinaryXor(_x_ modulo 2, _y_ modulo 2).
             1. If _tmp_ &ne; 0, then
-              1. Set _result_ to _result_ - 2<sup>_shift_</sup>. NOTE: This extends the sign.
+              1. Set _result_ to _result_ - 2<sup>_shift_</sup>.
+              1. NOTE: This extends the sign.
             1. Return _result_.
           </emu-alg>
         </emu-clause>
@@ -5451,8 +5452,10 @@
         1. Assert: Type(_p_) is String.
         1. Assert: Type(_q_) is String.
         1. If _q_ can be the string-concatenation of _p_ and some other String _r_, return *true*. Otherwise, return *false*.
-        1. NOTE: Any String is a prefix of itself, because _r_ may be the empty String.
       </emu-alg>
+      <emu-note>
+        <p>Any String is a prefix of itself, because _r_ may be the empty String.</p>
+      </emu-note>
     </emu-clause>
 
     <emu-clause id="sec-samevalue" aoid="SameValue">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -12642,9 +12646,11 @@ <h1>Runtime Semantics: ArrayAccumulation</h1>
         <emu-alg>
           1. Let _len_ be _nextIndex_ + 1.
           1. Perform ? Set(_array_, *"length"*, _len_, *true*).
-          1. NOTE: The above Set throws if _len_ exceeds 2<sup>32</sup>-1.
           1. Return _len_.
         </emu-alg>
+        <emu-note>
+          <p>The above Set throws if _len_ exceeds 2<sup>32</sup>-1.</p>
```
rkirsling: I think this one might be the most subjective. Technically it describes a single line so I'm happy to leave it as-is, but then the algorithm as a whole is tiny and I think endnotes tend to be cleaner than inline notes (though part of that is because we currently lack special styling for inline notes).

spec.html
```diff
@@ -8720,7 +8724,7 @@ <h1>FunctionDeclarationInstantiation ( _func_, _argumentsList_ )</h1>
               1. Else,
                 1. Let _initialValue_ be ! _envRec_.GetBindingValue(_n_, *false*).
               1. Call _varEnvRec_.InitializeBinding(_n_, _initialValue_).
-              1. NOTE: vars whose names are the same as a formal parameter, initially have the same value as the corresponding initialized parameter.
+              1. NOTE: Vars whose names are the same as a formal parameter initially have the same value as the corresponding initialized parameter.
```
ljharb: "vars" seems right to me; if capitalized i'd expect "Variables". not entirely sure why.

spec.html
ljharb: this note seems more appropriate inline, as it was.

spec.html
rkirsling: Yeah, I wasn't totally sure. It seems "vars" is only used three times total and this is the only sentence-initial case. [13.3.2](https://tc39.es/ecma262/#sec-variable-statement) does have "Var variables" at the beginning of a sentence though. In general, we seem to be inconsistent with regard to whether backticks are used, e.g. "var declaration" vs. "`var` declaration", but it's only really noticeable when you'd otherwise expect capitalization.

rkirsling: As a matter of fact, the other two instances of "vars" are from the very same algorithm, in this note which appears twice:
> NOTE: Only a single lexical environment is needed for the parameters and top-level vars.

I think if we're going to use "vars" like this then it should be fine to capitalize it, but another approach would be to make the note here singular, such that it'd start with "A var whose...".

rkirsling:
@syg Yeah, I guess they're basically like line comments, except that we can't just stick a newline in an emu-alg for readability... We could follow other specs and have the inline NOTE style match the out-of-line NOTE style, though this doesn't address the "refers to N steps" concern you mentioned. I kind of like your visual bracketing idea. Would be worth creating an issue for. :smile:

PR #1833, opened 2020-01-05
Editorial: Remove erroneous emu-grammar attribute.

rkirsling:
See https://github.com/bterlson/ecmarkup/issues/126#issuecomment-570834393.

A certain `emu-grammar` element is currently marked `type="example"` in spite of clearly being no such thing. This has no visible effect at the moment but it will soon.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -15128,7 +15128,7 @@
 
     <emu-clause id="sec-binary-bitwise-operators-runtime-semantics-evaluation">
       <h1>Runtime Semantics: Evaluation</h1>
-      <p>The production <emu-grammar type="example">A : A @ B</emu-grammar>, where @ is one of the bitwise operators in the productions above, is evaluated as follows:</p>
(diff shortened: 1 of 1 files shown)


jmdyck:
It's true that it isn't an example, but it certainly doesn't recap an existing production either. It's sort of a pattern that indirectly refers to a set of productions, so it should maybe have a `type` value all its own. (But I'm okay with this PR too.)

ljharb:
It seems like an example to me given that it's an exemplar of a pattern.

rkirsling:
"Example" here seems to refer strictly to `emu-grammar` instances in Section 5, where they aren't actually legitimate spec productions. This is an "abstract reference" in that it's referring to a bunch of real productions at once.

rkirsling:
Just a note that with https://github.com/bterlson/ecmarkup/pull/166 merged, the next update will make this issue visible.

To bring the relevant screenshot over from the comment linked in the description:

image

ljharb:
We'll have to rebase this after ecmarkup is published so we can see the visual difference.

rkirsling:
Rebased for visibility.

Issue #1834, opened 2020-01-06
referencing a const or let variable with this keyword is undefined


moonformeli:
**Description:** 
I'm not sure if this is an intended feature. 
When x is called with `this` keyword, it gives you an undefined.
I've tried to find the reason for this in ecma-interantional.org but don't know where I could find it.
Could someone provide me a link I can convince myself for why this isn't a bug?

**Output:**

```
const x = 1;

console.log(this.x); // undefined
```

ljharb:
This isn't a bug because variables aren't the same as properties (except when you're in the global scope of a Script).

bathos:
> Could someone provide me a link I can convince myself for why this isn't a bug?

It might help to reframe the question mentally if you’re still trying to see where this does/doesn’t occur. Rather than trying to figure out why it doesn’t happen here, you may have an easier time picking through the spec to figure out why it _does_ happen under specific conditions involving `var` and `function` declared bindings in scripts or undeclared bindings in sloppy mode scripts.

Unfortunately I don’t think there’s a single key part of the spec one can point to that explains all this in depth. The behaviors arise from a number of algorithms working together. Some key bits to look at:

1. The [Reference][0] specification type
2. The related [PutValue][1] operation (esp. step 5, where a sloppy unresolvable reference can obtain the global this as the value to which a property gets assigned)
3. The [GetIdentifierReference][2] operation, which is recusrive and can bottom out with an unresolvable reference
4. The [runtime semantics][3] (eval rules) of `VariableDeclaration : BindingIdentifier Initializer[opt]`, which use GetIdentifierReference indirectly through ResolveBinding and end with a PutValue call.

TBH it’s a bit confusing (and a note in the last link seems to reference a non-existent sixth step...), so I’m not sure I’ve nailed the right/best spots to look at, but hopefully it gives you a handhold for further investigation.

[0]: https://tc39.es/ecma262/#sec-reference-specification-type
[1]: https://tc39.es/ecma262/#sec-putvalue
[2]: https://tc39.es/ecma262/#sec-getidentifierreference
[3]: https://tc39.es/ecma262/#_ref_12315

rkirsling:
> a note in the second-to-last item’s linked op seems to reference a non-existent sixth step

Interesting, looks like the note was correct in ES2018 but became invalid in ES2019. And it's not alone. Will create a PR to fix, but this makes referring to step numbers seem like a really dangerous practice...

PR #1835, opened 2020-01-07
Editorial: Fix outdated Step # references.

rkirsling:
Noticed by @bathos in https://github.com/tc39/ecma262/issues/1834#issuecomment-571075603.

I'm shocked by how many of these there are, but it makes sense because there's nothing to enforce their correctness. We really could use a safer approach here, but in the mean time, this should fix what's no longer accurate.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -15387,7 +15387,7 @@
         1. Return _r_.
       </emu-alg>
       <emu-note>
-        <p>When an assignment occurs within strict mode code, it is a runtime error if _lref_ in step 1.f of the first algorithm or step 7 of the second algorithm it is an unresolvable reference. If it is, a *ReferenceError* exception is thrown. The |LeftHandSideExpression| also may not be a reference to a data property with the attribute value { [[Writable]]: *false* }, to an accessor property with the attribute value { [[Set]]: *undefined* }, nor to a non-existent property of an object for which the IsExtensible predicate returns the value *false*. In these cases a *TypeError* exception is thrown.</p>
+        <p>When an assignment occurs within strict mode code, it is a runtime error if _lref_ in step 1.e of the first algorithm or step 7 of the second algorithm is an unresolvable reference. If it is, a *ReferenceError* exception is thrown. The |LeftHandSideExpression| also may not be a reference to a data property with the attribute value { [[Writable]]: *false* }, to an accessor property with the attribute value { [[Set]]: *undefined* }, nor to a non-existent property of an object for which the IsExtensible predicate returns the value *false*. In these cases a *TypeError* exception is thrown.</p>
       </emu-note>
     </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -31484,7 +31484,7 @@ <h1>Runtime Semantics: RepeatMatcher ( _m_, _min_, _max_, _greedy_, _x_, _c_, _p
             <p>because each iteration of the outermost `*` clears all captured Strings contained in the quantified |Atom|, which in this case includes capture Strings numbered 2, 3, 4, and 5.</p>
           </emu-note>
           <emu-note>
-            <p>Step 1 of the RepeatMatcher's _d_ closure states that, once the minimum number of repetitions has been satisfied, any more expansions of |Atom| that match the empty character sequence are not considered for further repetitions. This prevents the regular expression engine from falling into an infinite loop on patterns such as:</p>
+            <p>Step a of the RepeatMatcher's _d_ closure states that, once the minimum number of repetitions has been satisfied, any more expansions of |Atom| that match the empty character sequence are not considered for further repetitions. This prevents the regular expression engine from falling into an infinite loop on patterns such as:</p>
```
rkirsling: This one's kind of weird. Step 2 defines the closure, so Step 2.a would therefore be "Step a of the ... closure".

The thing is that in [ES2015](https://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-repeatmatcher-abstract-operation), we actually ensured that this Step 2 in particular had Arabic numeral substeps, perhaps for this very purpose. But this was lost as of [ES2016](https://www.ecma-international.org/ecma-262/7.0/#sec-runtime-semantics-repeatmatcher-abstract-operation).

If "Step a" is too weird I can try to rephrase such that Step 2.a will be appropriate.

ljharb: this one is kind of weird. should we perhaps force the closure steps to be numeric again?

rkirsling: Do you know if there's a simple way to do that in ecmarkup? Seems like just adding an extra `<emu-alg>` layer in there doesn't help.

ljharb: hm, doesn't look like ecmarkup has that option - cc @rbuckton?

jmdyck: I think "Step 2.a of RepeatMatcher" would be fine. (The fact that it's a step of a closure created by RepeatMatcher isn't vital to what the Note is saying.)

syg:
> Long term, perhaps we should come up with a way in ecmarkup to add an ID to a step, and then reference that step elsewhere, so the numbering/lettering can be automated?

I know @bterlson has thought about this a lot in the past and was met with difficulties.

bterlson:
The best way I could figure to solve the immediate problem is by manually setting an ID using some syntax on steps that need inbound links. Unfortunately this doesn't address the use case where spec _readers_ want to link to a particular step as well.

Considered auto-generating IDs based on a hash of the step contents so it's stable between builds and re-orderings, but this is guaranteed to be frustrating. Among other problems, fixing a typo breaks inbound links.

Another solution I considered was auto-generating IDs to start, encoding them in the SOURCE document, and then the ids are stable until an editor decides to remove the step or change it significantly or whatever. In which case they might then delete the step ID and let ecmarkup generate a new one. I never went with this approach because the tool modifying the source document felt a little weird.

Anyway, just a dump of thoughts I've had over the years :)

PR #1842, opened 2020-01-12
Editorial: Give emu-notes some color.

### LINKED ISSUE #1825: Styling for notes ought to be less subdued
# Problem

Other specs tend to style notes in a way that calls attention with color while retaining adequate subtlety, yet we style them as if they were ~being consumed by encroaching fog~ blockquotes in GitHub Markdown.

## WASM
image
## HTML
image
## CSS
image
## ECMAScript
image

But notes are meant to be read—they may be "inessential" from the view of the spec as an implementation in itself, but they are often of key importance to the human reader, and I question whether we really mean to visually "background" them as such. 

# Solution(s)

I think we can do better, with a bit of Ecma orange! Here are a few possibilities.

## 1. Higher-impact approach
image

Specifically:
- Make border full Ecma orange (like the [metadata box](https://tc39.es/ecma262/#metadata-block)), remove grey text color
- Add background color which is Ecma orange with lightness bumped to 95%
  (also make `<code>` background transparent and `<td>` background white)
- Add 10px padding on the other three sides

Notes:
- I say high-_er_ impact because we could go further still and restyle the "NOTE" indicator if we wanted.
- Since the border is full-colored, the background lightness here could even be bumped up further.

## 2. Medium-impact approach
image

Like the first approach, but without a border:
- Remove grey border and its padding, remove grey text color
- Add background color which is Ecma orange with lightness bumped to 95%
  (also make `<code>` background transparent and `<td>` background white)
- Add 10px padding on all sides

Notes:
- This might be slightly less good for the printed version, where I believe background color would be omitted but border and padding would be preserved.

## 3. Lowest-impact approach
image

Basically the complement of the second approach:
- Make border full Ecma orange (like the [metadata box](https://tc39.es/ecma262/#metadata-block))...and that's it.

Notes:
- I don't care for this since I really want to do something about the grey text color too, but if the note text is black-on-white just like the body text, I fear the border color alone isn't enough of a visual distinction.

-----

Please note that I'm raising this discussion here, but the PR would need to take place in the ecmarkup repo, assuming we don't want it to be restricted to ECMA-262.

rkirsling:
Resolves #1825.
(_Well, sort of. This should eventually be done in ecmarkup itself, but ecmarkup is tricky to get changes into at the moment, and it's nice to have a concrete 262 snapshot to critique._)

This PR implements the "green notes without colored NOTE indicator" approach of https://github.com/tc39/ecma262/issues/1825#issuecomment-570971879.

Specifically:
- Borrow green border and background colors from Bikeshed
    (_easier than inventing something new, at least as a starting point_)
- Remove grey text color
- Add 10px padding to the three non-border sides
- Make `<code>` background transparent and `<td>` background white
    (_existing issues that only become visible when sitting on top of a non-white background_)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -41,6 +41,19 @@
     padding-left: 0;
     list-style: none;
   }
+  /* the following overrides need to be moved into ecmarkup itself */
+  emu-note {
+    color: inherit;
+    border-left: 5px solid #52E052;
+    background: #E9FBE9;
+    padding: 10px 10px 10px 0;
+  }
+  pre code.hljs {
+    background: transparent;
+  }
+  emu-table td {
+    background: #fff;
+  }
 </style>
 <pre class=metadata>
   title: ECMAScript&reg; 2020 Language&nbsp;Specification


rkirsling:
Seems like our troubles with updating ecmarkup may be resolved now, so if we'd prefer to do the visual review here and the code merge over there, that'd work too.

ljharb:
I think I'd (lightly) rather land this here, and then upstream it to ecmarkup once we know it works, and then we can remove the changes here with a dep update.

PR #1844, opened 2020-01-12
Editorial: Let code wrap in tables-in-notes.

### LINKED ISSUE #1843: Table 48 doesn't fit on the page
Table 48 is having a bad time at the moment:
image
It doesn't fit on the page, and scrolling is disabled so there isn't even a way to see the rest of the content.

If it weren't inside a note, it just fits _visually_ on my screen, but it still spills out of the containing `<figure>`:
image

The only way I can see to address this without removing a column is by ensuring that the code can wrap:
image
This is achievable with `emu-table td code { white-space: normal; }`.
(I don't think it needs to be specific to tables-in-notes?)

rkirsling:
Resolves #1843.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -41,6 +41,9 @@
     padding-left: 0;
     list-style: none;
   }
+  emu-note emu-table td code {
+    white-space: normal;
+  } 
 </style>
 <pre class=metadata>
   title: ECMAScript&reg; 2020 Language&nbsp;Specification


ljharb:
This certainly seems to solve it; i'd prefer, however, to only wrap code on table 48 instead of on all tables.

rkirsling:
I can do that, it just leads to various concerns:

- Won't it be hard to remember that we have special CSS for "`#table-42` which is actually Table 48"?

- If we make a Table 48-specific solution, is this the right-sized Table 48-specific solution to make? (Should we pull the table out of the note in addition / instead?)

- Is there a concrete drawback to a generic solution here? It should only apply when we've actually run out of width to work with, and has the nice benefit of being responsive if your window is a bit too narrow.

ljharb:
> It should only apply when we've actually run out of width to work with, and has the nice benefit of being responsive if your window is a bit too narrow.

This is a really good point; maybe it's fine as-is then.

rkirsling:
Should I move this PR to ecmarkup, if we're okay with a general solution?

rkirsling:
Moved to https://github.com/bterlson/ecmarkup/pull/186.

Issue #1649, opened 2019-07-29
document adherence to Oxford English Dictionary spelling


michaelficarra:
I believe it started with ES2015, but at some point we started to follow Oxford English Dictionary spelling conventions. The most noticeable change for me was the renaming of "object initialisers" to "object initializers". @allenwb should know more about this decision. Either way, I think we should document this within the spec, possibly within a new "editorial conventions" section.

allenwb:
I don't have the actual references at hand, but what we discovered back then was that the ISO guidelnes permitted standards to follow the OED conventions and that was it was also acceptable to Ecma. So we adopted those conventions.

jmdyck:
I agree that we should document editorial preferences & decisions, but I'm doubtful that we should do so *within the spec*. Certainly, it makes sense to state an editorial choice in the spec if doing so would help a reader understand the spec. But I don't think that's true of a lot of "style guide" choices.

For things that are less for someone *reading* the spec and more for someone *writing* it (i.e., new proposals. PRs), perhaps we could have a new document in the how-we-work repo.

apaprocki:
I'll just leave this, since it is sufficiently meta -- the British Standards Institution (BSI) publishes a standard, BS 0, the "_standard for standards_", where it literally covers only these 3 items, on page 1, in the _introduction_ (they must come up _really_ often!):

> Certain topics regularly arise during any revision of BS 0, and it was decided:
> • to retain the “z” spelling usage (e.g. “standardization”) for consistency with the preferred variant listed in the Shorter Oxford English Dictionary, noting also that it is the traditional English usage rather than an Americanism;
> • to retain the decimal point in preference to the decimal comma, noting that it reflects the overwhelmingly predominant practice in the UK and is inconsistent with many international and European standards that are adopted without change by BSI; and
> • to retain the “f” spelling of “sulfur” and its derivatives, noting that it is the agreed variant adopted by the International Union of Pure and Applied Chemistry (IUPAC) and most analogous professional bodies in the UK, and reflects the general usage of most of those in the UK scientific community.

PR #1856, opened 2020-01-29
Editorial: Unify spelling of 'behaviour'.

rkirsling:
The spec uses `behaviour` 125 times and `behavior` 4 times, so let's unify upon the former.
(_I looked for any other US/UK spelling inconsistencies I could find, but didn't identify any._)

Also fix a misspelling on one of these lines (`soley` → `solely`).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -18068,7 +18068,7 @@
         </ul>
 
         <emu-note>
-          <p>Hosts are not required to implement the algorithm in <emu-xref href="#sec-%foriniteratorprototype%.next"></emu-xref> directly. They may choose any implementation whose behavior will not deviate from that algorithm unless one of the constraints in the previous paragraph is violated.</p>
+          <p>Hosts are not required to implement the algorithm in <emu-xref href="#sec-%foriniteratorprototype%.next"></emu-xref> directly. They may choose any implementation whose behaviour will not deviate from that algorithm unless one of the constraints in the previous paragraph is violated.</p>
           <p>The following is an informative definition of an ECMAScript generator function that conforms to these rules:</p>
           <pre><code class="javascript">
             function* EnumerateObjectProperties(obj) {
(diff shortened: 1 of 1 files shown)


rkirsling:
Yeah, if we had reason to shift, then that would be fine too, but consistency first, I figure.

claudepache:
I think that the convention is to use OED spelling. See also #1649.

rkirsling:
@claudepache Thanks! I knew I'd seen some sort of issue or PR about it but was having trouble finding it.

spec.html
```diff
@@ -18065,13 +18065,13 @@ <h1>EnumerateObjectProperties ( _O_ )</h1>
...
         </emu-note>
         <emu-note>
-          The list of exotic objects for which implementations are not required to match CreateForInIterator was chosen because implementations historically differed in behavior for those cases, and agreed in all others.
+          The list of exotic objects for which implementations are not required to match CreateForInIterator was chosen because implementations historically differed in behaviour for those cases, and agreed in all others.
         </emu-note>
       </emu-clause>
 
       <emu-clause id="sec-for-in-iterator-objects">
         <h1>For-In Iterator Objects</h1>
-        <p>A For-In Iterator is an object that represents a specific iteration over some specific object. For-In Iterator objects are never directly accessible to ECMAScript code; they exist soley to illustrate the behavior of EnumerateObjectProperties.</p>
```
syg: lol soley

claudepache:
BTW, I’ve noticed that “colour” is misspelled [here](https://tc39.es/ecma262/#sec-string.prototype.fontcolor). Should I open a PR? 🙃

> #### String.prototype.fontcolor ( <del>_color_</del> <ins>_colour_</ins> )
> When the **fontcolor** method is called with argument <del>_color_</del> <ins>_colour_</ins>, the following steps are taken:

rkirsling:
Hahah, I did notice that, but it seemed like it'd be a pretty confusing change. 😅

ljharb:
No, I think `color` is appropriate since the method is named `fontcolor`, and identifiers aren't prose.

PR #1860, opened 2020-01-31
Editorial: Improve consistency of backtick use for built-ins.

rkirsling:
### Problem

All cases of `<built-in> {object, type, value, instance, prototype}` throughout the spec have the built-in name consistently unstylized. `<built-in> constructor`, however, varies: usually the name is unstylized, but sometimes it's written with backticks.

I originally thought to do away with _all_ of these backticks, but I ended up feeling like perhaps the backticks are meant to distinguish "the concrete <code>\`&lt;built-in&gt;\`</code> constructor function" from "the conceptual `<built-in>` constructor", in which case some of them could be legitimate.

For instance:
- <code>The AsyncGeneratorFunction constructor ... is a standard built-in function object that inherits from the \`Function\` constructor</code> certainly feels broken, but...
- <code>A Number object is created by using the \`Number\` constructor in a \`new\` expression</code> might be concrete enough to warrant backticks.

### Solution

- Remove backticks around cases of `<built-in> constructor` ~that aren't clearly referring to a concrete function~
- Remove backticks for all cases of `<built-in> behaviour`

(Also, as a drive-by fix: add missing backticks for a few function names.)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -308,9 +308,9 @@
 
     <emu-clause id="sec-boolean-object">
       <h1>Boolean object</h1>
-      <p>member of the Object type that is an instance of the standard built-in `Boolean` constructor</p>
+      <p>member of the Object type that is an instance of the standard built-in Boolean constructor</p>
       <emu-note>
-        <p>A Boolean object is created by using the `Boolean` constructor in a `new` expression, supplying a Boolean value as an argument. The resulting object has an internal slot whose value is the Boolean value. A Boolean object can be coerced to a Boolean value.</p>
+        <p>A Boolean object is created by using the Boolean constructor in a `new` expression, supplying a Boolean value as an argument. The resulting object has an internal slot whose value is the Boolean value. A Boolean object can be coerced to a Boolean value.</p>
       </emu-note>
     </emu-clause>
 
@@ -329,9 +329,9 @@
 
     <emu-clause id="sec-string-object">
       <h1>String object</h1>
-      <p>member of the Object type that is an instance of the standard built-in `String` constructor</p>
+      <p>member of the Object type that is an instance of the standard built-in String constructor</p>
       <emu-note>
-        <p>A String object is created by using the `String` constructor in a `new` expression, supplying a String value as an argument. The resulting object has an internal slot whose value is the String value. A String object can be coerced to a String value by calling the `String` constructor as a function (<emu-xref href="#sec-string-constructor-string-value"></emu-xref>).</p>
+        <p>A String object is created by using the String constructor in a `new` expression, supplying a String value as an argument. The resulting object has an internal slot whose value is the String value. A String object can be coerced to a String value by calling the String constructor as a function (<emu-xref href="#sec-string-constructor-string-value"></emu-xref>).</p>
       </emu-note>
     </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spectranaut:
I think your reasoning here is good... but when people edit and contribute to the spec in the future, will they be able to pick up on the subtlety of this distinction by observation? For simplicity's sake I might still prefer (a) all backticks removed, or (b) all references to built-ins given backticks.

With a slight preference for (a) because it is the smaller change. But I'm curious if others have an opinion.

rkirsling:
@spectranaut Yeah, I do share your worry. I'd be more than happy to switch to (a) if our editors agree.

jmdyck:
>  to distinguish "the concrete `` `<built-in>` `` constructor function" from "the conceptual `<built-in>` constructor",

I don't understand this distinction. What's a "conceptual" constructor?

rkirsling:
@jmdyck Hmm, perhaps 'abstract' is a better word; I guess I mean "a spec entity and not a userland one"? But I can't guarantee that this is ultimately coherent.

The thought process that me led here was like this:
1. I hadn't realized that stuff like "Date object" is consistently unstylized; I kind of expected that `Date` would be backticked everywhere.
2. Since "`Date` constructor" seems like the odd one out, I tried removing backticks for all such cases, but the more 'concrete' cases (which refer to `new` or `super` or what have you) struck me as more intentional and made me start to second-guess myself (with respect to what the editors might prefer).

That said, I myself would strongly prefer not to make a distinction among the "`<built-in>` constructor" cases. Let me add a commit to drop the rest of the backticks and we can see if folks have any objections.

spec.html
```diff
@@ -39078,7 +39078,7 @@ <h1>The AsyncGeneratorFunction Constructor</h1>
...
+        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified AsyncGeneratorFunction behaviour must include a `super` call to the AsyncGeneratorFunction constructor to create and initialize subclass instances with the internal slots necessary for built-in AsyncGeneratorFunction behaviour. All ECMAScript syntactic forms for defining async generator function objects create direct instances of `AsyncGeneratorFunction`. There is no syntactic means to create instances of `AsyncGeneratorFunction` subclasses.</li>
```
michaelficarra, line 39081: We should remove backticks from all occurrences of `AsyncGeneratorFunction` here, right?

```suggestion
        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified AsyncGeneratorFunction behaviour must include a `super` call to the AsyncGeneratorFunction constructor to create and initialize subclass instances with the internal slots necessary for built-in AsyncGeneratorFunction behaviour. All ECMAScript syntactic forms for defining async generator function objects create direct instances of AsyncGeneratorFunction. There is no syntactic means to create instances of AsyncGeneratorFunction subclasses.</li>
```

rkirsling: Yeah, I can deal with all four of these paragraphs while we're at it (one for each function type).

michaelficarra: @rkirsling Sounds like a PR opportunity to me.

PR #1879, opened 2020-02-22
Editorial: Update InternalizeJSONProperty to accept reviver explicitly

### LINKED ISSUE #1878: InternalizeJSONProperty missing reviver argument
At https://tc39.es/ecma262/#sec-json.parse JSON.parse calls InternalizeJSONProperty but does not pass the reviver as an argument. InternalizeJSONProperty does not have it as a parameter. But then InternalizeJSONProperty uses the reviver that JSON.parse should have passed to it.

Attn @FUDCo

gibson042:
Fixes #1878

(copied from [JSON.parse source text access
 proposal](https://tc39.es/proposal-json-parse-with-source/#sec-internalizejsonproperty))

No effects on downstream dependencies:
* [Web IDL](https://heycam.github.io/webidl/)
* [HTML Standard](https://html.spec.whatwg.org/)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38133,7 +38133,7 @@ THH:mm:ss.sss
           1. Let _root_ be OrdinaryObjectCreate(%Object.prototype%).
           1. Let _rootName_ be the empty String.
           1. Perform ! CreateDataPropertyOrThrow(_root_, _rootName_, _unfiltered_).
-          1. Return ? InternalizeJSONProperty(_root_, _rootName_).
+          1. Return ? InternalizeJSONProperty(_root_, _rootName_, _reviver_).
         1. Else,
           1. Return _unfiltered_.
       </emu-alg>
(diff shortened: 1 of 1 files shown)


bakkot:
If we're changing this, we should also change

- [`SerializeJSONProperty`](https://tc39.es/ecma262/#sec-serializejsonproperty), which has a similar thing with `ReplacerFunction`,
- [`SerializeJSONObject`](https://tc39.es/ecma262/#sec-serializejsonobject), which has a similar thing with `stack`, `indent`, `gap`, and `PropertyList`, and
- [`SerializeJSONArray`](https://tc39.es/ecma262/#sec-serializejsonarray), which has a similar thing with `stack`, `indent`, `gap`, and `PropertyList`.

If we're passing these arguments explicitly, all three of these functions will need access to all five values, since they are mutually recursive.

ljharb:
An alternative might be, passing a single Record that holds on to all of the relevant values used in the operation?

Issue #1295, opened 2018-08-23
`spread` is mistakenly called "an operator" in the introduction section of ES2018


zerkms:
> It also includes rest parameter and **spread operator** support for object properties.

(highlight is mine) https://www.ecma-international.org/ecma-262/9.0/index.html#sec-intro 

Syntactically it's not an operator, so I think the wording should be changed to "spread syntax" or something like that.

fkling:
Since this is specific to objects, maybe it should be "rest properties" and "spread properties" or "object rest" and "object spread".

Both "rest parameter" and "spread operator" are confusing at best and plain wrong at worst in the context of objects (what is a parameter in the context of objects??).

The terminology used in the spec, even in non-normative sections, should be as precise as possible IMO.

PR #1881, opened 2020-02-23
Editorial: fix misleading uses of 'operator'

uhyo:
Hello, there are two uses of the word 'operator' that refer to those not defined to be operators. This PR proposes to replace them with other appropriate words.

- “`new` or `super` operators”
  - While `new` is actually an operator ([12.3.5 The new Operator](https://tc39.es/ecma262/#sec-new-operator)), `super` is not ([12.3.7 The super Keyword](https://tc39.es/ecma262/#sec-super-keyword)). Following @ljharb's suggestion, this PR rephrases it to “`new` operator or a `super` call`”.
- “optional chaining operator”
  - The production that starts from `?.` token is simply called optional chains ([12.3.9 Optional Chains](https://tc39.es/ecma262/#sec-optional-chains)). Since the paragraph in question discusses syntactic analysis, I chose the word 'syntax'.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2444,7 +2444,7 @@
                 (a List of <em>any</em>, Object) <b>&rarr;</b> Object
               </td>
               <td>
-                Creates an object. Invoked via the `new` or `super` operators. The first argument to the internal method is a list containing the arguments of the operator. The second argument is the object to which the `new` operator was initially applied. Objects that implement this internal method are called <em>constructors</em>. A function object is not necessarily a constructor and such non-constructor function objects do not have a [[Construct]] internal method.
+                Creates an object. Invoked via the `new` operator or a `super` call. The first argument to the internal method is a list containing the arguments of the constructor invocation or the `super` call. The second argument is the object to which the `new` operator was initially applied. Objects that implement this internal method are called <em>constructors</em>. A function object is not necessarily a constructor and such non-constructor function objects do not have a [[Construct]] internal method.
               </td>
             </tr>
             </tbody>
@@ -13465,7 +13465,7 @@
             a?.b
             `c`
           </code></pre>
-          <p>so that it would be interpreted as two valid statements. The purpose is to maintain consistency with similar code without optional chaining operator:</p>
+          <p>so that it would be interpreted as two valid statements. The purpose is to maintain consistency with similar code without optional chaining:</p>
           <pre><code class="javascript">
             a.b
             `c`


spec.html
```diff
@@ -2444,7 +2444,7 @@ <h1>Object Internal Methods and Internal Slots</h1>
                 (a List of <em>any</em>, Object) <b>&rarr;</b> Object
               </td>
               <td>
-                Creates an object. Invoked via the `new` or `super` operators. The first argument to the internal method is a list containing the arguments of the operator. The second argument is the object to which the `new` operator was initially applied. Objects that implement this internal method are called <em>constructors</em>. A function object is not necessarily a constructor and such non-constructor function objects do not have a [[Construct]] internal method.
+                Creates an object. Invoked via the `new` or `super` expressions. The first argument to the internal method is a list containing the arguments of the expression. The second argument is the object to which the `new` operator was initially applied. Objects that implement this internal method are called <em>constructors</em>. A function object is not necessarily a constructor and such non-constructor function objects do not have a [[Construct]] internal method.
```
ljharb: I’m not sure it makes sense to say an expression has arguments, but i don’t have a better suggestion.

uhyo: Thank you! Another thought of mine is `call`:  “Invoked via the `new` or `super` calls.” and “...a list containing the arguments of the call.”
Which looks better?

ljharb: call to me implies a function, so I’m not sure that makes sense either for new.

Maybe:
```suggestion
                Creates an object. Invoked via the `new` operator or a `super` call. The first argument to the internal method is a list containing the arguments of the constructor invocation or the `super` call. The second argument is the object to which the `new` operator was initially applied. Objects that implement this internal method are called <em>constructors</em>. A function object is not necessarily a constructor and such non-constructor function objects do not have a [[Construct]] internal method.
```

uhyo: Thank you! Yours looks totally fine to me. I will adopt your suggestion. (It seems that GitHub doesn't allow me to adopt a suggestion that is created too recently.)

uhyo: Adopted.

bakkot:
See also https://github.com/tc39/ecma262/issues/1295

michaelficarra:
If static member access (`.`) can be called an operator (its RHS is not an expression), then `?.` is an operator, too.

spec.html
```diff
@@ -13572,7 +13572,7 @@ <h1>Static Semantics: Early Errors</h1>
             a?.b
             `c`
           </code></pre>
-          <p>so that it would be interpreted as two valid statements. The purpose is to maintain consistency with similar code without optional chaining operator:</p>
+          <p>so that it would be interpreted as two valid statements. The purpose is to maintain consistency with similar code without optional chaining syntax:</p>
```
```suggestion
          <p>so that it would be interpreted as two valid statements. The purpose is to maintain consistency with similar code without optional chaining:</p>
```

syg:
```suggestion
          <p>so that it would be interpreted as two valid statements. The purpose is to maintain consistency with similar code without optional chaining:</p>
```

uhyo: Adopted; sorry that I should have made an issue first for this.

ljharb:
@uhyo would you mind signing the contribution agreement? https://github.com/tc39/ecma262/blob/master/CONTRIBUTING.md#required-legal-agreements

uhyo:
@ljharb done, thank you.

PR #1885, opened 2020-02-27
Editorial: Add subscripts to BindingIdentifier static semantics.

marjakh:
Adding subscripts here makes sense, because the semantics specifically talk
about the presence of a parameter.

Also, the productions below, where the semantics also talk about the presence
of a parameter, already have the subscripts. So it's inconsistent to not have
them here.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12252,7 +12252,7 @@
         </li>
       </ul>
       <emu-grammar>
-        BindingIdentifier : `yield`
+        BindingIdentifier[Yield, Await] : `yield`
       </emu-grammar>
       <ul>
         <li>
@@ -12260,7 +12260,7 @@
         </li>
       </ul>
       <emu-grammar>
-        BindingIdentifier : `await`
+        BindingIdentifier[Yield, Await] : `await`
       </emu-grammar>
       <ul>
         <li>


bakkot:
Note that currently none of these render because https://github.com/bterlson/ecmarkup/pull/165 has not been upstreamed.

jmdyck:
> Note that currently none of these render because [bterlson/ecmarkup#165](https://github.com/bterlson/ecmarkup/pull/165) has not been upstreamed.

I think the effects of that PR were visible shortly after it was merged. Certainly they're visible now. For example, we can see the `[Yield, Await]` parameters in the productions immediately following the ones targeted by this PR.

bakkot:
> Certainly they're visible now.

Ah, let me amend that to "none of these render in PR snapshots". It didn't occur to me that snapshots might be built with a different version of ecmarkup.

ljharb:
@marjakh would you mind registering as a contributor? https://github.com/tc39/ecma262/blob/master/CONTRIBUTING.md#required-legal-agreements

marjakh:
I work for Google and I was under the impression we're covered somehow. Isn't that the case? This is my second contribution and I don't remember what i did the first time, whether I registered to something or not. It was a while back. I can try to search my e-mail for history, if needed.

syg:
> I work at Google and I was under the impression we're covered somehow. Isn't that the case? This is my second contribution and I don't remember what i did the first time, whether I registered to something or not. It was a while back. I can try to search my e-mail for history, if needed.

This is a good point, though unfortunately there's no easy way to check for employer. The best approximation we came up with is check for public membership in https://github.com/googlers. @marjakh Could you please join that at go/github, then go to https://github.com/googlers, search for your handle and set your membership to Public?

marjakh:
Done!

littledan:
Yes, people who work for member companies are generally covered, and should not sign the contributor form. Glad this has been sorted out.

PR #1890, opened 2020-03-06
Editorial: pass necessary arguments to SerializeJSON*

bakkot:
`SerializeJSONProperty `, `SerializeJSONObject`, and `SerializeJSONArray` refer to variables which they have not been passed using notation like ``SerializeJSONProperty has access to _ReplacerFunction_ from the invocation of the `stringify` method``.

I find this sketchy (see #1884).

This PR makes all the necessary values be passed explicitly, wrapped up in a Record.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38120,7 +38120,8 @@ THH:mm:ss.sss
           1. Let _gap_ be the empty String.
         1. Let _wrapper_ be OrdinaryObjectCreate(%Object.prototype%).
         1. Perform ! CreateDataPropertyOrThrow(_wrapper_, the empty String, _value_).
-        1. Return ? SerializeJSONProperty(the empty String, _wrapper_).
+        1. Let _state_ be the Record { [[ReplacerFunction]]: _ReplacerFunction_, [[Stack]]: _stack_, [[Indent]]: _indent_, [[Gap]]: _gap_, [[PropertyList]]: _PropertyList_ }.
+        1. Return ? SerializeJSONProperty(_state_, the empty String, _wrapper_).
       </emu-alg>
       <p>This function is the <dfn>%JSONStringify%</dfn> intrinsic object.</p>
       <p>The *"length"* property of the `stringify` function is 3.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -38264,16 +38265,16 @@ <h1>JSON.stringify ( _value_ [ , _replacer_ [ , _space_ ] ] )</h1>
       </emu-note>
 
       <emu-clause id="sec-serializejsonproperty" aoid="SerializeJSONProperty">
-        <h1>Runtime Semantics: SerializeJSONProperty ( _key_, _holder_ )</h1>
-        <p>The abstract operation SerializeJSONProperty with arguments _key_, and _holder_ has access to _ReplacerFunction_ from the invocation of the `stringify` method. Its algorithm is as follows:</p>
+        <h1>Runtime Semantics: SerializeJSONProperty ( _state_, _key_, _holder_ )</h1>
+        <p>The abstract operation SerializeJSONProperty with arguments _state_, _key_, and _holder_ performs the following steps:</p>
         <emu-alg>
           1. Let _value_ be ? Get(_holder_, _key_).
```
ljharb: perhaps?
```suggestion
          1. Assert: _state_ is a Record.
          1. Let _value_ be ? Get(_holder_, _key_).
```

spec.html
```diff
@@ -38426,56 +38427,56 @@ <h1>Runtime Semantics: UnicodeEscape ( _C_ )</h1>
...
-          1. If _stack_ contains _value_, throw a *TypeError* exception because the structure is cyclical.
-          1. Append _value_ to _stack_.
-          1. Let _stepback_ be _indent_.
-          1. Set _indent_ to the string-concatenation of _indent_ and _gap_.
-          1. If _PropertyList_ is not *undefined*, then
-            1. Let _K_ be _PropertyList_.
+        <h1>Runtime Semantics: SerializeJSONObject ( _state_, _value_ )</h1>
+        <p>The abstract operation SerializeJSONObject with arguments _state_ and _value_ serializes an object. It performs the following steps:</p>
+        <emu-alg>
+          1. If _state_.[[Stack]] contains _value_, throw a *TypeError* exception because the structure is cyclical.
```
ljharb: and also?
```suggestion
          1. Assert: _state_ is a Record.
          1. If _state_.[[Stack]] contains _value_, throw a *TypeError* exception because the structure is cyclical.
```

spec.html
```diff
@@ -38426,56 +38427,56 @@ <h1>Runtime Semantics: UnicodeEscape ( _C_ )</h1>
...
-        <h1>Runtime Semantics: SerializeJSONArray ( _value_ )</h1>
-        <p>The abstract operation SerializeJSONArray with argument _value_ serializes an array. It has access to the _stack_, _indent_, and _gap_ values of the current invocation of the `stringify` method.</p>
+        <h1>Runtime Semantics: SerializeJSONArray ( _state_, _value_ )</h1>
+        <p>The abstract operation SerializeJSONArray with arguments _state_ and _value_ serializes an array. It performs the following steps:</p>
         <emu-alg>
-          1. If _stack_ contains _value_, throw a *TypeError* exception because the structure is cyclical.
-          1. Append _value_ to _stack_.
-          1. Let _stepback_ be _indent_.
-          1. Set _indent_ to the string-concatenation of _indent_ and _gap_.
+          1. If _state_.[[Stack]] contains _value_, throw a *TypeError* exception because the structure is cyclical.
```
ljharb: and?
```suggestion
          1. Assert: _state_ is a Record.
          1. If _state_.[[Stack]] contains _value_, throw a *TypeError* exception because the structure is cyclical.
```

bakkot:
Hm. Asserting that something is a Record isn't all that useful, since that tells you nothing about the fields in that record, which are just as important. So I think I'd prefer to omit those assertions, though it's not a strong preference.

ljharb:
In general I prefer abstract ops to assert about their arguments as strongly as possible; certainly we could make a JSONStateRecord type and assert it's one of those (which would include the slots), but that feels a bit overkill here.

bakkot:
I agree that it feels like overkill to make such an assertion, but at that point I don't really see the value of making any assertion at all. There's nowhere else in the spec we just assert something is a Record, as far as I know.

spec.html
syg: Hm, is it that helpful since it's not an assert on the specific fields in the record?

ljharb: https://github.com/tc39/ecma262/pull/1890#issuecomment-595904356 suggests it's not

PR #1907, opened 2020-03-20
Editorial: Use abstract closures instead of semantic functions

### LINKED ISSUE #1895: Use abstract closures for Atomics semantic functions
Now that we have abstract closures we should stop using the "semantic functions" used in the Atomics section.

syg:
Refactor semantic functions as used by atomic RMW operations to be
a subclass of abstract closures that are pure.

The remaining use of "semantic function" for reads-bytes-from is
changed to say "mathematical function". It is used in the axiomatic
event semantics and do not have algorithm steps.

Fixes #1895.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4435,7 +4435,7 @@
 
     <emu-clause id="sec-data-blocks">
       <h1>Data Blocks</h1>
-      <p>The <dfn>Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
+      <p>The <dfn>Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A <dfn>byte value</dfn> is an integer value in the range 0 through 255, inclusive. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
       <p>For notational convenience within this specification, an array-like syntax can be used to access the individual bytes of a Data Block value. This notation presents a Data Block value as a 0-origined integer-indexed sequence of bytes. For example, if _db_ is a 5 byte Data Block value then _db_[2] can be used to access its 3<sup>rd</sup> byte.</p>
       <p>A data block that resides in memory that can be referenced from multiple agents concurrently is designated a <dfn>Shared Data Block</dfn>. A Shared Data Block has an identity (for the purposes of equality testing Shared Data Block values) that is <em>address-free</em>: it is tied not to the virtual addresses the block is mapped to in any process, but to the set of locations in memory that the block represents. Two data blocks are equal only if the sets of the locations they contain are equal; otherwise, they are not equal and the intersection of the sets of locations they contain is empty.  Finally, Shared Data Blocks can be distinguished from Data Blocks.</p>
       <p>The semantics of Shared Data Blocks is defined using Shared Data Block events by the memory model. Abstract operations below introduce Shared Data Block events and act as the interface between evaluation semantics and the event semantics of the memory model. The events form a candidate execution, on which the memory model acts as a filter. Please consult the memory model for full semantics.</p>
@@ -36762,6 +36762,23 @@ THH:mm:ss.sss
   <emu-clause id="sec-arraybuffer-objects">
     <h1>ArrayBuffer Objects</h1>
 
+    <emu-clause id="sec-arraybuffer-notation">
+      <h1>Notation</h1>
+      <p>The descriptions below in this section, <emu-xref href="#sec-atomics-object"></emu-xref>, and <emu-xref href="#sec-memory-model"></emu-xref> use the read-modify-write modification function internal data structure.</p>
+      <p>A <dfn>read-modify-write modification function</dfn> is a mathematical function that is notationally represented as an abstract closure that takes two Lists of byte values as arguments and returns a List of byte values. These abstract closures satisfy all of the following properties:</p>
+      <ul>
+        <li>They perform all their algorithm steps atomically.</li>
+        <li>Their individual algorithm steps are not observable.</li>
+      </ul>
+      <emu-note>
+        <p>To aid verifying that a read-modify-write modification function's algorithm steps constitute a pure, mathematical function, the following editorial conventions are recommended:</p>
+        <ul>
+          <li>They do not access, directly or transitively via invoked abstract operations and abstract closures, any language or specification values except their parameters and captured values.</li>
+          <li>They do not return completion values.</li>
+        </ul>
+      </emu-note>
+    </emu-clause>
+
     <emu-clause id="sec-abstract-operations-for-arraybuffer-objects">
       <h1>Abstract Operations For ArrayBuffer Objects</h1>
 
@@ -36948,7 +36965,7 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-getmodifysetvalueinbuffer" aoid="GetModifySetValueInBuffer">
         <h1>GetModifySetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _op_ [ , _isLittleEndian_ ] )</h1>
-        <p>The abstract operation GetModifySetValueInBuffer takes arguments _arrayBuffer_ (a SharedArrayBuffer object), _byteIndex_ (a non-negative integer), _type_ (a TypedArray element type), _value_ (a Number or a BigInt), and _op_ (a semantic function) and optional argument _isLittleEndian_ (a Boolean). It performs the following steps when called:</p>
+        <p>The abstract operation GetModifySetValueInBuffer takes arguments _arrayBuffer_ (a SharedArrayBuffer object), _byteIndex_ (a non-negative integer), _type_ (a TypedArray element type), _value_ (a Number or a BigInt), and _op_ (a read-modify-write modification function) and optional argument _isLittleEndian_ (a Boolean). It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: IsSharedArrayBuffer(_arrayBuffer_) is *true*.
           1. Assert: There are sufficient bytes in _arrayBuffer_ starting at _byteIndex_ to represent a value of _type_.
@@ -37785,7 +37802,7 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-atomicreadmodifywrite" aoid="AtomicReadModifyWrite">
         <h1>AtomicReadModifyWrite ( _typedArray_, _index_, _value_, _op_ )</h1>
-        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a pure combining operation). _op_ takes two List of byte values arguments and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps when called:</p>
+        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a read-modify-write modification function). _op_ takes two List of byte values arguments and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps when called:</p>
         <emu-alg>
           1. Let _buffer_ be ? ValidateSharedIntegerTypedArray(_typedArray_).
           1. Let _i_ be ? ValidateAtomicAccess(_typedArray_, _index_).
@@ -37814,23 +37831,63 @@ THH:mm:ss.sss
           1. Return GetValueFromBuffer(_buffer_, _indexedPosition_, _elementType_, *true*, ~SeqCst~).
         </emu-alg>
       </emu-clause>
+
+      <emu-clause id="sec-bytelistbitwiseop" aoid="ByteListBitwiseOp">
+        <h1>ByteListBitwiseOp( _op_, _xBytes_, _yBytes_ )</h1>
+        <p>The abstract operation ByteListBitwiseOp takes arguments _op_ (a read-modify-write modification function), _xBytes_ (a List of byte values), and _yBytes_ (a List of byte values). The operation atomically performs a bitwise operation on all byte values of the arguments and returns a List of byte values. It performs the following steps when called:</p>
+        <emu-alg>
+          1. Assert: _xBytes_ and _yBytes_ have the same number of elements.
+          1. Let _result_ be a new empty List.
+          1. Let _i_ be 0.
+          1. For each element _xByte_ of _xBytes_, do
+            1. Let _yByte_ be _yBytes_[_i_].
+            1. Let _resultByte_ be the result of applying the bitwise operator _op_ to _xByte_ and _yByte_.
+            1. Set _i_ to _i_ + 1.
+            1. Append _resultByte_ to the end of _result_.
+          1. Return _result_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-bytelistequal" aoid="ByteListEqual">
+        <h1>ByteListEqual( _xBytes_, _yBytes_ )</h1>
+        <p>The abstract operation ByteListEqual takes arguments _xBytes_ (a List of byte values) and _yBytes_ (a List of byte values). It performs the following steps when called:</p>
+        <emu-alg>
+          1. If _xBytes_ and _yBytes_ do not have the same number of elements, return *false*.
+          1. Let _i_ be 0.
+          1. For each element _xByte_ of _xBytes_, do
+            1. Let _yByte_ be _yBytes_[_i_].
+            1. If _xByte_ is not equal to _yByte_, return *false*.
+            1. Set _i_ to _i_ + 1.
+          1. Return *true*.
+        </emu-alg>
+      </emu-clause>
     </emu-clause>
 
     <emu-clause id="sec-atomics.add">
       <h1>Atomics.add ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `add` denote a semantic function of two List of byte values arguments that applies the addition operation to the Number values corresponding to the List of byte values arguments and returns a List of byte values corresponding to the result of that operation.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `add`).
+        1. Let _type_ be the Element Type value in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _typedArray_.[[TypedArrayName]].
+        1. Let _isLittleEndian_ be the value of the [[LittleEndian]] field of the surrounding agent's Agent Record.
+        1. Let _add_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures _type_ and _isLittleEndian_ and performs the following steps atomically when called:
+          1. Let _x_ be RawBytesToNumeric(_type_, _xBytes_, _isLittleEndian_).
+          1. Let _y_ be RawBytesToNumeric(_type_, _yBytes_, _isLittleEndian_).
+          1. Let _T_ be Type(_x_).
+          1. Let _sum_ be _T_::add(_x_, _y_).
+          1. Let _sumBytes_ be NumericToRawBytes(_type_, _sum_, _isLittleEndian_).
+          1. Assert: _sumBytes_, _xBytes_, and _yBytes_ have the same number of elements.
+          1. Return _sumBytes_.
+        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, _add_).
       </emu-alg>
     </emu-clause>
 
     <emu-clause id="sec-atomics.and">
       <h1>Atomics.and ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `and` denote a semantic function of two List of byte values arguments that applies the bitwise-and operation element-wise to the two arguments and returns a List of byte values corresponding to the result of that operation.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `and`).
+        1. Let _and_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures nothing and performs the following steps atomically when called:
+          1. Return ByteListBitwiseOp(`&`, _xBytes_, _yBytes_).
+        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, _and_).
       </emu-alg>
     </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -37915,10 +37956,16 @@ <h1>Atomics.store ( _typedArray_, _index_, _value_ )</h1>
 
     <emu-clause id="sec-atomics.sub">
       <h1>Atomics.sub ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `subtract` denote a semantic function of two List of byte values arguments that applies the subtraction operation to the Number values corresponding to the List of byte values arguments and returns a List of byte values corresponding to the result of that operation.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `subtract`).
+        1. Let _subtract_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures _type_ and _isLittleEndian_ and performs the following steps atomically when called:
```
bakkot: This presumably needs the same

```
        1. Let _type_ be the Element Type value in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _typedArray_.[[TypedArrayName]].
        1. Let _isLittleEndian_ be the value of the [[LittleEndian]] field of the surrounding agent's Agent Record.
```
steps that `Atomics.add` has.

spec.html
```diff
@@ -36755,6 +36755,17 @@ <h1>Structured Data</h1>
   <emu-clause id="sec-arraybuffer-objects">
     <h1>ArrayBuffer Objects</h1>
 
+    <emu-clause id="sec-arraybuffer-notation">
+      <h1>Notation</h1>
+      <p>The descriptions below in this section, <emu-xref href="#sec-atomics-object"></emu-xref>, and <emu-xref href="#sec-memory-model"></emu-xref> use the read-modify-write modification function internal data structure.</p>
+      <p>A <em>read-modify-write modification function</em> is a mathematical function that is notationally represented as an abstract closure that takes two Lists of bytes as arguments returns a List of bytes. These abstract closures satisfy all of the following properties:</p>
+      <ul>
+        <li>They do not access any language or specification values except their parameters, captured values, and the [[LittleEndian]] field of the surrounding agent's Agent Record.</li>
```
bakkot: They don't ever refer to `[[LittleEndian]]`, that I can see.

spec.html
```diff
@@ -37846,17 +37882,21 @@ <h1>Atomics.compareExchange ( _typedArray_, _index_, _expectedValue_, _replaceme
...
 
     <emu-clause id="sec-atomics.exchange">
       <h1>Atomics.exchange ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `second` denote a semantic function of two List of byte values arguments that returns its second argument.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `second`).
+        1. Let _compareExchange_ be a new read-modify-write modification function with parameters (_oldBytes_, _newBytes_) that captures nothing and performs the following steps atomically when called:
+          1. Return _newBytes_.
+        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, _second_).
```
```suggestion
        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, _compareExchange_).
```

spec.html
```diff
@@ -37807,23 +37818,48 @@ <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
...
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `add`).
+        1. Let _type_ be the Element Type value in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _typedArray_.[[TypedArrayName]].
+        1. Let _isLittleEndian_ be the value of the [[LittleEndian]] field of the surrounding agent's Agent Record.
+        1. Let _add_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures _type_ and _isLittleEndian_ and performs the following steps atomically when called:
+          1. Assert: _xBytes_ and _yBytes_ have equal length.
+          1. Let _x_ be RawBytesToNumeric(_type_, _xBytes_, _isLittleEndian_).
+          1. Let _y_ be RawBytesToNumeric(_type_, _yBytes_, _isLittleEndian_).
+          1. Let _T_ be Type(_x_).
+          1. Let _sum_ be _T_::add(_x_, _y_).
+          1. Return NumericToRawBytes(_type_, _sumBytes_, _isLittleEndian_).
```
bakkot: `_sumBytes_` does not appear to be defined. Perhaps you meant `_sum_`?

Also, you maybe need to worry about ensuring you return a list of the same length as was input.

spec.html
```diff
@@ -36755,6 +36755,17 @@ <h1>Structured Data</h1>
   <emu-clause id="sec-arraybuffer-objects">
     <h1>ArrayBuffer Objects</h1>
 
+    <emu-clause id="sec-arraybuffer-notation">
+      <h1>Notation</h1>
+      <p>The descriptions below in this section, <emu-xref href="#sec-atomics-object"></emu-xref>, and <emu-xref href="#sec-memory-model"></emu-xref> use the read-modify-write modification function internal data structure.</p>
+      <p>A <em>read-modify-write modification function</em> is a mathematical function that is notationally represented as an abstract closure that takes two Lists of bytes as arguments returns a List of bytes. These abstract closures satisfy all of the following properties:</p>
```
```suggestion
      <p>A <em>read-modify-write modification function</em> is a mathematical function that is notationally represented as an abstract closure that takes two Lists of bytes as arguments and returns a List of bytes. These abstract closures satisfy all of the following properties:</p>
```
also, what is a "byte" in terms of spec or JS values? It doesn't seem defined in the current spec, but that seems like a big omission.

spec.html
```diff
@@ -36755,6 +36755,17 @@ <h1>Structured Data</h1>
...
+      <ul>
+        <li>They do not access any language or specification values except their parameters, captured values, and the [[LittleEndian]] field of the surrounding agent's Agent Record.</li>
+        <li>They do not return completion values and do not invoke other abstract closures or operations that return completion values.</li>
```
ljharb: Shouldn't "they do not return completion values" be deferred until we do overall completion value reform?

spec.html
```diff
@@ -37807,23 +37818,48 @@ <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
...
     </emu-clause>
 
     <emu-clause id="sec-atomics.and">
       <h1>Atomics.and ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `and` denote a semantic function of two List of byte values arguments that applies the bitwise-and operation element-wise to the two arguments and returns a List of byte values corresponding to the result of that operation.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `and`).
+        1. Let _and_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures nothing and performs the following steps atomically when called:
+          1. Return ByteListBitwiseOp(`&`, _xBytes_, _yBytes_).
```
```suggestion
          1. Return ! ByteListBitwiseOp(`&`, _xBytes_, _yBytes_).
```

spec.html
```diff
@@ -37807,23 +37818,48 @@ <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
...
       <h1>Atomics.add ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `add` denote a semantic function of two List of byte values arguments that applies the addition operation to the Number values corresponding to the List of byte values arguments and returns a List of byte values corresponding to the result of that operation.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `add`).
+        1. Let _type_ be the Element Type value in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _typedArray_.[[TypedArrayName]].
+        1. Let _isLittleEndian_ be the value of the [[LittleEndian]] field of the surrounding agent's Agent Record.
+        1. Let _add_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures _type_ and _isLittleEndian_ and performs the following steps atomically when called:
+          1. Assert: _xBytes_ and _yBytes_ have equal length.
+          1. Let _x_ be RawBytesToNumeric(_type_, _xBytes_, _isLittleEndian_).
```
```suggestion
          1. Let _x_ be ! RawBytesToNumeric(_type_, _xBytes_, _isLittleEndian_).
```

spec.html
```diff
@@ -37807,23 +37818,48 @@ <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
...
+          1. Let _y_ be RawBytesToNumeric(_type_, _yBytes_, _isLittleEndian_).
```
```suggestion
          1. Let _y_ be ! RawBytesToNumeric(_type_, _yBytes_, _isLittleEndian_).
```

spec.html
```diff
@@ -37807,23 +37818,48 @@ <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
...
+          1. Let _T_ be Type(_x_).
+          1. Let _sum_ be _T_::add(_x_, _y_).
```
```suggestion
          1. Let _sum_ be ! _T_::add(_x_, _y_).
```

spec.html
```suggestion
          1. Return ! NumericToRawBytes(_type_, _sum_, _isLittleEndian_).
```

spec.html
```diff
@@ -37846,17 +37882,21 @@ <h1>Atomics.compareExchange ( _typedArray_, _index_, _expectedValue_, _replaceme
         1. Let _elementSize_ be the Element Size value specified in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
         1. Let _offset_ be _typedArray_.[[ByteOffset]].
         1. Let _indexedPosition_ be (_i_ &times; _elementSize_) + _offset_.
-        1. Let `compareExchange` denote a semantic function of two List of byte values arguments that returns the second argument if the first argument is element-wise equal to _expectedBytes_.
-        1. Return GetModifySetValueInBuffer(_buffer_, _indexedPosition_, _elementType_, _replacement_, `compareExchange`).
+        1. Let _compareExchange_ be a new read-modify-write modification function with parameters (_oldBytes_, _newBytes_) that captures _expectedBytes_ and performs the following steps atomically when called:
+          1. If _oldBytes_ is element-wise equal to _expectedBytes_, then
```
ljharb: the phrase "element-wise equal" only appears in the context of Atomics; I'd missed this in prior review. What does this mean? This seems like something that should be defined instead of being inferred from prose.

spec.html
```diff
@@ -37846,17 +37882,21 @@ <h1>Atomics.compareExchange ( _typedArray_, _index_, _expectedValue_, _replaceme
...
       </emu-alg>
     </emu-clause>
 
     <emu-clause id="sec-atomics.exchange">
       <h1>Atomics.exchange ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `second` denote a semantic function of two List of byte values arguments that returns its second argument.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `second`).
+        1. Let _compareExchange_ be a new read-modify-write modification function with parameters (_oldBytes_, _newBytes_) that captures nothing and performs the following steps atomically when called:
```
ljharb: can "that captures nothing" be the default state, so we don't have to state it explicitly? (or do we prefer stating it explicitly)

spec.html
```diff
@@ -37888,10 +37928,11 @@ <h1>Atomics.load ( _typedArray_, _index_ )</h1>
 
     <emu-clause id="sec-atomics.or">
       <h1>Atomics.or ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `or` denote a semantic function of two List of byte values arguments that applies the bitwise-or operation element-wise to the two arguments and returns a List of byte values corresponding to the result of that operation.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `or`).
+        1. Let _or_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures nothing and performs the following steps atomically when called:
+          1. Return ByteListBitwiseOp(`|`, _xBytes_, _yBytes_).
```
```suggestion
          1. Return ! ByteListBitwiseOp(`|`, _xBytes_, _yBytes_).
```

spec.html
```diff
@@ -37915,10 +37956,16 @@ <h1>Atomics.store ( _typedArray_, _index_, _value_ )</h1>
...
     <emu-clause id="sec-atomics.sub">
       <h1>Atomics.sub ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `subtract` denote a semantic function of two List of byte values arguments that applies the subtraction operation to the Number values corresponding to the List of byte values arguments and returns a List of byte values corresponding to the result of that operation.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `subtract`).
+        1. Let _subtract_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures _type_ and _isLittleEndian_ and performs the following steps atomically when called:
+          1. Assert: _xBytes_ and _yBytes_ have equal length.
+          1. Let _x_ be RawBytesToNumeric(_type_, _xBytes_, _isLittleEndian_).
+          1. Let _y_ be RawBytesToNumeric(_type_, _yBytes_, _isLittleEndian_).
```
```suggestion
          1. Let _y_ be ! RawBytesToNumeric(_type_, _yBytes_, _isLittleEndian_).
```

spec.html
```diff
@@ -37915,10 +37956,16 @@ <h1>Atomics.store ( _typedArray_, _index_, _value_ )</h1>
 
     <emu-clause id="sec-atomics.sub">
       <h1>Atomics.sub ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `subtract` denote a semantic function of two List of byte values arguments that applies the subtraction operation to the Number values corresponding to the List of byte values arguments and returns a List of byte values corresponding to the result of that operation.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `subtract`).
+        1. Let _subtract_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures _type_ and _isLittleEndian_ and performs the following steps atomically when called:
+          1. Assert: _xBytes_ and _yBytes_ have equal length.
+          1. Let _x_ be RawBytesToNumeric(_type_, _xBytes_, _isLittleEndian_).
```
```suggestion
          1. Let _x_ be ! RawBytesToNumeric(_type_, _xBytes_, _isLittleEndian_).
```

spec.html
```diff
@@ -37915,10 +37956,16 @@ <h1>Atomics.store ( _typedArray_, _index_, _value_ )</h1>
...
+          1. Let _y_ be RawBytesToNumeric(_type_, _yBytes_, _isLittleEndian_).
+          1. Let _T_ be Type(_x_).
+          1. Let _difference_ be _T_::subtract(_x_, _y_).
+          1. Return NumericToRawBytes(_type_, _differenceBytes_, _isLittleEndian_).
```
```suggestion
          1. Return ! NumericToRawBytes(_type_, _differenceBytes_, _isLittleEndian_).
```

spec.html
```diff
@@ -37989,10 +38036,11 @@ <h1>Atomics.notify ( _typedArray_, _index_, _count_ )</h1>
 
     <emu-clause id="sec-atomics.xor">
       <h1>Atomics.xor ( _typedArray_, _index_, _value_ )</h1>
-      <p>Let `xor` denote a semantic function of two List of byte values arguments that applies the bitwise-xor operation element-wise to the two arguments and returns a List of byte values corresponding to the result of that operation.</p>
       <p>The following steps are taken:</p>
       <emu-alg>
-        1. Return ? AtomicReadModifyWrite(_typedArray_, _index_, _value_, `xor`).
+        1. Let _xor_ be a new read-modify-write modification function with parameters (_xBytes_, _yBytes_) that captures nothing and performs the following steps atomically when called:
+          1. Return ByteListBitwiseOp(`^`, _xBytes_, _yBytes_).
```
```suggestion
          1. Return ! ByteListBitwiseOp(`^`, _xBytes_, _yBytes_).
```

spec.html
syg: It is a byte. I think that's a reasonable omission though, given prevalence of the shared understanding of "byte" in computing.

spec.html
syg: Well, I think it's important they're obviously "pure" from the start, and returning explicit Completion Records would confuse that. Calling out something like "do not return explicit completion values and do not invoke other [...] that return explicit completion values" is also kind of confusing IMO. Any ideas here?

spec.html
syg: See above for not returning completion values or call other non-completion value returning ops so as to appear obviously pure. Lack of ?/! is intentional here.

syg:
I'm fine with requiring either that no ?/! be used anywhere in the RMW modification functions, or that only ! be used. I prefer the former, since I think it's more obvious to me that the function is "pure" that way. @ljharb I assume prefers the other way. @bakkot tie-breaker?

spec.html
bakkot: "byte" or at least "byte-sized" is defined at the top of the section on [Data Blocks](https://deploy-preview-1907--ecma262-snapshots.netlify.com/#sec-data-blocks).

(FWIW there are a number of platforms with byte sizes other than 8 bits which are still in use. This comes up a lot on e.g. [the LLVM developers list](http://lists.llvm.org/pipermail/llvm-dev/2019-May/132080.html), where being imprecise about whether or not that is a shared assumption has led to a fair bit of pain. For a specification, it's worth mentioning. But I don't think we need to do more than we already have.)

spec.html
syg: Ack.

spec.html
syg: Oops, will remove.

spec.html
syg: Actually _second_ is right, I named the function itself wrong.

bakkot:
My preference is also to avoid `!`. I hope to address https://github.com/tc39/ecma262/issues/1796 this week, at which point avoiding `!` will be unambiguously the right thing.

spec.html
syg: Yes I wondered about overflow here, but it is how the rest of TA works (and `|0`, I assume) already, so an assert seems fine here.

spec.html
syg: Hm, okay, I can add a bit to the section on Lists.

spec.html
syg: Both @bakkot and I prefer stating it explicitly.

spec.html
ljharb: I don't see how that would confuse it at all - if you wanted to constrain it to only return *normal* completion values, that seems fine, but with the current state of completion values, having some of them *not* return one seems very confusing to me.

spec.html
syg: Added a sentence to the Data Blocks section.

ljharb: A byte can be a string of 0s and 1s, or an integer (mathematical, Number, BigInt), a char, or something else perhaps. I think it's important to be specific.

ljharb:
Prior to #1796, the reigning approach is to always use the `!`. To me, purity has nothing to do with completion record vs return value (https://github.com/tc39/ecma262/pull/1907#discussion_r396663286).

syg:
> [...] but with the current state of completion values, having some of them not return one seems very confusing to me.

The current state is not consistent, as we've discussed at length before, and besides @bakkot will PR that.

>  To me, purity has nothing to do with completion record vs return value (#1907 (comment))

This is the wrong intuition and this confusion is precisely why @bakkot, @michaelficarra, and I want to do the reform. Completion Records are about *reifying control* in observable program evaluation, with `normal` meaning "next step", `return` meaning early returns, `break` meaning labeled statement breaks, and `continue` meaning loop continues.

1. `! Op()` means "Op can have effects on evaluation control flow. For this particular callsite, the only effect it can have is a normal continuation, i.e. evaluate the next thing." 
1. `? Op()` means "Op can have effects on evaluation control flow. For this particular callsite, abrupt values are propagated immediately."
1. `Op()` means "Op cannot have effects on evaluation control flow."

1 and 3 really should not be conflated. 3 tells me a lot about no observability on control flow, which is a necessary (but not sufficient) condition for an operation being observable at all. 1 doesn't tell me that.

spec.html
```diff
@@ -4409,7 +4409,7 @@ <h1>The Abstract Closure Specification Type</h1>
 
     <emu-clause id="sec-data-blocks">
       <h1>Data Blocks</h1>
-      <p>The <dfn>Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
+      <p>The <dfn>Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A <em>byte</em> is an integer value in the range 0 through 255, inclusive. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
```
```suggestion
      <p>The <dfn>Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A <dfn>byte</dfn> is an integer value in the range 0 through 255, inclusive. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
```

so it auto-links?

spec.html
ljharb: this one isn't an abstract closure.

syg: That's true, that's worth a note saying none of the steps performed by ::add here are observable because _x_ and _y_ are guaranteed to be numbers.

ljharb:
The problem I see is that step 2 is one that's easy to forget, and the omnipresence of `!` helps ensure that a `?` is not omitted. This is something we should really discuss re the overall reform topic tho; however pending that, i don't think we should be changing the way callable things work in the spec so far. (it is pretty consistent already, i think, just not desirable in the long term)

spec.html
ljharb: which means that "doesn't call other abstract operations" isn't actually necessary to indicate or guarantee that something is "pure" - so why is that part of the definition of these functions?

spec.html
syg: I didn't want the auto-linking, it seems excessive. My thought was byte could be used in many different contexts where the auto-linking has dubious value.

spec.html
```diff
@@ -36755,6 +36755,23 @@ <h1>Structured Data</h1>
   <emu-clause id="sec-arraybuffer-objects">
     <h1>ArrayBuffer Objects</h1>
 
+    <emu-clause id="sec-arraybuffer-notation">
+      <h1>Notation</h1>
+      <p>The descriptions below in this section, <emu-xref href="#sec-atomics-object"></emu-xref>, and <emu-xref href="#sec-memory-model"></emu-xref> use the read-modify-write modification function internal data structure.</p>
+      <p>A <em>read-modify-write modification function</em> is a mathematical function that is notationally represented as an abstract closure that takes two Lists of bytes as arguments and returns a List of bytes. These abstract closures satisfy all of the following properties:</p>
```
```suggestion
      <p>A <dfn>read-modify-write modification function</dfn> is a mathematical function that is notationally represented as an abstract closure that takes two Lists of bytes as arguments and returns a List of bytes. These abstract closures satisfy all of the following properties:</p>
```
this one, i think, should be linked.

spec.html
```diff
@@ -36755,6 +36755,23 @@ <h1>Structured Data</h1>
...
+      <ul>
+        <li>They perform all their algorithm steps atomically.</li>
+        <li>Their individual algorithm steps are not observable.</li>
+      </ul>
+      <emu-note>
+        <p>To aid verifying that a read-modify-write modification function's algorithm steps constitute a pure, mathematical function, the following editorial conventions are recommended:</p>
```
ljharb: thanks, this is much clearer to me.

spec.html
ljharb: what about if you call it a "byte value" instead of just a "byte"? then the few times it's used, it would usefully link, but prose discussion of bytes wouldn't be noisy.

spec.html
syg: Ack.

syg:
Addressed review, squashed, rebased.

spec.html
```diff
@@ -36754,6 +36754,23 @@ <h1>Structured Data</h1>
   <emu-clause id="sec-arraybuffer-objects">
     <h1>ArrayBuffer Objects</h1>
 
+    <emu-clause id="sec-arraybuffer-notation">
+      <h1>Notation</h1>
+      <p>The descriptions below in this section, <emu-xref href="#sec-atomics-object"></emu-xref>, and <emu-xref href="#sec-memory-model"></emu-xref> use the read-modify-write modification function internal data structure.</p>
+      <p>A <dfn>read-modify-write modification function</dfn> is a mathematical function that is notationally represented as an abstract closure that takes two Lists of bytes as arguments and returns a List of bytes. These abstract closures satisfy all of the following properties:</p>
```
```suggestion
      <p>A <dfn>read-modify-write modification function</dfn> is a mathematical function that is notationally represented as an abstract closure that takes two Lists of byte values as arguments and returns a List of byte values. These abstract closures satisfy all of the following properties:</p>
```

syg:
Rebased; PTAL.

spec.html
```diff
@@ -37785,7 +37802,7 @@ <h1>NotifyWaiter ( _WL_, _W_ )</h1>
 
       <emu-clause id="sec-atomicreadmodifywrite" aoid="AtomicReadModifyWrite">
         <h1>AtomicReadModifyWrite ( _typedArray_, _index_, _value_, _op_ )</h1>
-        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a pure combining operation). _op_ takes two List of byte values arguments and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps when called:</p>
+        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (an read-modify-write modification function). _op_ takes two List of byte values arguments and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps when called:</p>
```
bakkot, line 37805:
```suggestion
        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a read-modify-write modification function). _op_ takes two List of byte values arguments and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps when called:</p>
```

PR #1908, opened 2020-03-20
Normative: Allow Atomics methods on ArrayBuffers

syg:
This is to align with [wasm allowing atomics to work on non-shared memory](https://github.com/WebAssembly/threads/issues/144).

Most methods work in the obvious deterministic way on ArrayBuffers, with the exceptions of:

- `Atomics.wait` still throws on `ArrayBuffer`s.
- `Atomics.notify` always returns 0 on `ArrayBuffer`s.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37437,7 +37437,7 @@ THH:mm:ss.sss
             1. Let _readEvent_ be ReadSharedMemory { [[Order]]: _order_, [[NoTear]]: _noTear_, [[Block]]: _block_, [[ByteIndex]]: _byteIndex_, [[ElementSize]]: _elementSize_ }.
             1. Append _readEvent_ to _eventList_.
             1. Append Chosen Value Record { [[Event]]: _readEvent_, [[ChosenValue]]: _rawValue_ } to _execution_.[[ChosenValues]].
-          1. Else, let _rawValue_ be a List of _elementSize_ containing, in order, the _elementSize_ sequence of bytes starting with _block_[_byteIndex_].
+          1. Else, let _rawValue_ be a List of size _elementSize_ containing the sequence of _elementSize_ bytes starting with _block_[_byteIndex_].
           1. If _isLittleEndian_ is not present, set _isLittleEndian_ to the value of the [[LittleEndian]] field of the surrounding agent's Agent Record.
           1. Return RawBytesToNumeric(_type_, _rawValue_, _isLittleEndian_).
         </emu-alg>
@@ -37487,9 +37487,9 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-getmodifysetvalueinbuffer" aoid="GetModifySetValueInBuffer">
         <h1>GetModifySetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _op_ [ , _isLittleEndian_ ] )</h1>
-        <p>The abstract operation GetModifySetValueInBuffer takes arguments _arrayBuffer_ (a SharedArrayBuffer object), _byteIndex_ (a non-negative integer), _type_ (a TypedArray element type), _value_ (a Number or a BigInt), and _op_ (a read-modify-write modification function) and optional argument _isLittleEndian_ (a Boolean). It performs the following steps when called:</p>
+        <p>The abstract operation GetModifySetValueInBuffer takes arguments _arrayBuffer_ (an ArrayBuffer object or a SharedArrayBuffer object), _byteIndex_ (a non-negative integer), _type_ (a TypedArray element type), _value_ (a Number or a BigInt), and _op_ (a read-modify-write modification function) and optional argument _isLittleEndian_ (a Boolean). It performs the following steps when called:</p>
         <emu-alg>
-          1. Assert: IsSharedArrayBuffer(_arrayBuffer_) is *true*.
+          1. Assert: IsDetachedBuffer(_arrayBuffer_) is *false*.
           1. Assert: There are sufficient bytes in _arrayBuffer_ starting at _byteIndex_ to represent a value of _type_.
           1. Assert: ! IsNonNegativeInteger(_byteIndex_) is *true*.
           1. Assert: Type(_value_) is BigInt if ! IsBigIntElementType(_type_) is *true*; otherwise, Type(_value_) is Number.
@@ -37497,13 +37497,18 @@ THH:mm:ss.sss
           1. Let _elementSize_ be the Element Size value specified in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for Element Type _type_.
           1. If _isLittleEndian_ is not present, set _isLittleEndian_ to the value of the [[LittleEndian]] field of the surrounding agent's Agent Record.
           1. Let _rawBytes_ be NumericToRawBytes(_type_, _value_, _isLittleEndian_).
-          1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
-          1. Let _eventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
-          1. Let _rawBytesRead_ be a List of length _elementSize_ of nondeterministically chosen byte values.
-          1. NOTE: In implementations, _rawBytesRead_ is the result of a load-link, of a load-exclusive, or of an operand of a read-modify-write instruction on the underlying hardware. The nondeterminism is a semantic prescription of the memory model to describe observable behaviour of hardware with weak consistency.
-          1. Let _rmwEvent_ be ReadModifyWriteSharedMemory { [[Order]]: ~SeqCst~, [[NoTear]]: *true*, [[Block]]: _block_, [[ByteIndex]]: _byteIndex_, [[ElementSize]]: _elementSize_, [[Payload]]: _rawBytes_, [[ModifyOp]]: _op_ }.
-          1. Append _rmwEvent_ to _eventList_.
-          1. Append Chosen Value Record { [[Event]]: _rmwEvent_, [[ChosenValue]]: _rawBytesRead_ } to _execution_.[[ChosenValues]].
+          1. If IsSharedArrayBuffer(_arrayBuffer_) is *true*, then
+            1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
+            1. Let _eventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
+            1. Let _rawBytesRead_ be a List of length _elementSize_ of nondeterministically chosen byte values.
+            1. NOTE: In implementations, _rawBytesRead_ is the result of a load-link, of a load-exclusive, or of an operand of a read-modify-write instruction on the underlying hardware. The nondeterminism is a semantic prescription of the memory model to describe observable behaviour of hardware with weak consistency.
+            1. Let _rmwEvent_ be ReadModifyWriteSharedMemory { [[Order]]: ~SeqCst~, [[NoTear]]: *true*, [[Block]]: _block_, [[ByteIndex]]: _byteIndex_, [[ElementSize]]: _elementSize_, [[Payload]]: _rawBytes_, [[ModifyOp]]: _op_ }.
+            1. Append _rmwEvent_ to _eventList_.
+            1. Append Chosen Value Record { [[Event]]: _rmwEvent_, [[ChosenValue]]: _rawBytesRead_ } to _execution_.[[ChosenValues]].
+          1. Else,
+            1. Let _rawBytesRead_ be a List of size _elementSize_ containing the sequence of _elementSize_ bytes starting with _block_[_byteIndex_].
+            1. Let _rawBytesModified_ be _op_(_rawBytesRead_, _rawBytes_).
+            1. Store the individual bytes of _rawBytesModified_ into _block_, in order, starting at _block_[_byteIndex_].
           1. Return RawBytesToNumeric(_type_, _rawBytesRead_, _isLittleEndian_).
         </emu-alg>
       </emu-clause>
@@ -38176,21 +38181,18 @@ THH:mm:ss.sss
     <emu-clause id="sec-abstract-operations-for-atomics">
       <h1>Abstract Operations for Atomics</h1>
 
-      <emu-clause id="sec-validatesharedintegertypedarray" aoid="ValidateSharedIntegerTypedArray">
-        <h1>ValidateSharedIntegerTypedArray ( _typedArray_ [ , _waitable_ ] )</h1>
-        <p>The abstract operation ValidateSharedIntegerTypedArray takes argument _typedArray_ and optional argument _waitable_ (a Boolean). It performs the following steps when called:</p>
+      <emu-clause id="sec-validateintegertypedarray" aoid="ValidateIntegerTypedArray" oldid="sec-validatesharedintegertypedarray">
+        <h1>ValidateIntegerTypedArray ( _typedArray_ [ , _waitable_ ] )</h1>
+        <p>The abstract operation ValidateIntegerTypedArray takes argument _typedArray_ and optional argument _waitable_ (a Boolean). It performs the following steps when called:</p>
         <emu-alg>
           1. If _waitable_ is not present, set _waitable_ to *false*.
-          1. Perform ? RequireInternalSlot(_typedArray_, [[TypedArrayName]]).
+          1. Let _buffer_ be ? ValidateTypedArray(_typedArray_).
           1. Let _typeName_ be _typedArray_.[[TypedArrayName]].
           1. Let _type_ be the Element Type value in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _typeName_.
           1. If _waitable_ is *true*, then
             1. If _typeName_ is not *"Int32Array"* or *"BigInt64Array"*, throw a *TypeError* exception.
           1. Else,
             1. If ! IsUnclampedIntegerElementType(_type_) is *false* and ! IsBigIntElementType(_type_) is *false*, throw a *TypeError* exception.
-          1. Assert: _typedArray_ has a [[ViewedArrayBuffer]] internal slot.
-          1. Let _buffer_ be _typedArray_.[[ViewedArrayBuffer]].
-          1. If IsSharedArrayBuffer(_buffer_) is *false*, throw a *TypeError* exception.
           1. Return _buffer_.
         </emu-alg>
       </emu-clause>
@@ -38200,8 +38202,8 @@ THH:mm:ss.sss
         <p>The abstract operation ValidateAtomicAccess takes arguments _typedArray_ and _requestIndex_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: _typedArray_ is an Object that has a [[ViewedArrayBuffer]] internal slot.
-          1. Let _accessIndex_ be ? ToIndex(_requestIndex_).
           1. Let _length_ be _typedArray_.[[ArrayLength]].
+          1. Let _accessIndex_ be ? ToIndex(_requestIndex_).
           1. Assert: _accessIndex_ &ge; 0.
           1. If _accessIndex_ &ge; _length_, throw a *RangeError* exception.
           1. Return _accessIndex_.
@@ -38324,11 +38326,13 @@ THH:mm:ss.sss
         <h1>AtomicReadModifyWrite ( _typedArray_, _index_, _value_, _op_ )</h1>
         <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a read-modify-write modification function). _op_ takes two List of byte values arguments and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps when called:</p>
         <emu-alg>
-          1. Let _buffer_ be ? ValidateSharedIntegerTypedArray(_typedArray_).
+          1. Let _buffer_ be ? ValidateIntegerTypedArray(_typedArray_).
           1. Let _i_ be ? ValidateAtomicAccess(_typedArray_, _index_).
           1. Let _arrayTypeName_ be _typedArray_.[[TypedArrayName]].
           1. If _typedArray_.[[ContentType]] is ~BigInt~, let _v_ be ? ToBigInt(_value_).
           1. Otherwise, let _v_ be ? ToInteger(_value_).
+          1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
+          1. NOTE: The above check is not redundant with the check in ValidateIntegerTypedArray because the call to ToBigInt or ToInteger on the preceding lines can have arbitrary side effects, which could cause the buffer to become detached.
           1. Let _elementSize_ be the Element Size value specified in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
           1. Let _elementType_ be the Element Type value in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
           1. Let _offset_ be _typedArray_.[[ByteOffset]].
@@ -38341,8 +38345,10 @@ THH:mm:ss.sss
         <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
         <p>The abstract operation AtomicLoad takes arguments _typedArray_ and _index_. It atomically loads a value and returns the loaded value. It performs the following steps when called:</p>
         <emu-alg>
-          1. Let _buffer_ be ? ValidateSharedIntegerTypedArray(_typedArray_).
+          1. Let _buffer_ be ? ValidateIntegerTypedArray(_typedArray_).
           1. Let _i_ be ? ValidateAtomicAccess(_typedArray_, _index_).
+          1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
+          1. NOTE: The above check is not redundant with the check in ValidateIntegerTypedArray because the call to ValidateAtomicAccess on the preceding line can have arbitrary side effects, which could cause the buffer to become detached.
           1. Let _arrayTypeName_ be _typedArray_.[[TypedArrayName]].
           1. Let _elementSize_ be the Element Size value specified in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
           1. Let _elementType_ be the Element Type value in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -36941,23 +36952,27 @@ <h1>SetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _isTypedArra
 
       <emu-clause id="sec-getmodifysetvalueinbuffer" aoid="GetModifySetValueInBuffer">
         <h1>GetModifySetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _op_ [ , _isLittleEndian_ ] )</h1>
-        <p>The abstract operation GetModifySetValueInBuffer takes six parameters, a SharedArrayBuffer _arrayBuffer_, a nonnegative integer _byteIndex_, a TypedArray element type _type_, a Number or BigInt _value_, a semantic function _op_, and optionally a Boolean _isLittleEndian_. This operation performs the following steps:</p>
+        <p>The abstract operation GetModifySetValueInBuffer takes six parameters, a SharedArrayBuffer _arrayBuffer_, a nonnegative integer _byteIndex_, a TypedArray element type _type_, a Number or BigInt _value_, a read-modify-write modification function _op_, and optionally a Boolean _isLittleEndian_. This operation performs the following steps:</p>
```
bakkot: > a SharedArrayBuffer _arrayBuffer_

Presumably this is now wrnog.

spec.html
```diff
@@ -36941,23 +36952,27 @@ <h1>SetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _isTypedArra
...
-          1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
-          1. Let _eventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
-          1. Let _rawBytesRead_ be a List of length _elementSize_ of nondeterministically chosen byte values.
-          1. NOTE: In implementations, _rawBytesRead_ is the result of a load-link, of a load-exclusive, or of an operand of a read-modify-write instruction on the underlying hardware. The nondeterminism is a semantic prescription of the memory model to describe observable behaviour of hardware with weak consistency.
-          1. Let _rmwEvent_ be ReadModifyWriteSharedMemory { [[Order]]: ~SeqCst~, [[NoTear]]: *true*, [[Block]]: _block_, [[ByteIndex]]: _byteIndex_, [[ElementSize]]: _elementSize_, [[Payload]]: _rawBytes_, [[ModifyOp]]: _op_ }.
-          1. Append _rmwEvent_ to _eventList_.
-          1. Append Chosen Value Record { [[Event]]: _rmwEvent_, [[ChosenValue]]: _rawBytesRead_ } to _execution_.[[ChosenValues]].
+          1. Let _rawBytesRead_ be a new empty List.
```
bakkot: Nit: I don't think we usually worry about pre-declaring variables. I recall there's at least some other place in the spec where we have a `Let x be foo` in both branches of an `if` and then we refer to `x` outside of the `if`.

michaelficarra: I'd prefer to lint those away if we could.

jmdyck: I'd like to join this discussion, but it seems like it should be in a separate issue.

spec.html
```diff
@@ -37657,7 +37657,6 @@ <h1>ValidateSharedIntegerTypedArray ( _typedArray_ [ , _waitable_ ] )</h1>
             1. If ! IsUnclampedIntegerElementType(_type_) is *false* and ! IsBigIntElementType(_type_) is *false*, throw a *TypeError* exception.
           1. Assert: _typedArray_ has a [[ViewedArrayBuffer]] internal slot.
           1. Let _buffer_ be _typedArray_.[[ViewedArrayBuffer]].
-          1. If IsSharedArrayBuffer(_buffer_) is *false*, throw a *TypeError* exception.
```
ljharb: this part is normative, but not part of the rename; can it be separated into its own commit?

spec.html
```diff
@@ -37643,9 +37643,9 @@ <h1>The Atomics Object</h1>
     <emu-clause id="sec-abstract-operations-for-atomics">
       <h1>Abstract Operations for Atomics</h1>
 
-      <emu-clause id="sec-validatesharedintegertypedarray" aoid="ValidateSharedIntegerTypedArray">
-        <h1>ValidateSharedIntegerTypedArray ( _typedArray_ [ , _waitable_ ] )</h1>
-        <p>The abstract operation ValidateSharedIntegerTypedArray takes one argument _typedArray_ and an optional Boolean _waitable_. It performs the following steps:</p>
+      <emu-clause id="sec-validateintegertypedarray" aoid="ValidateIntegerTypedArray" oldid="sec-validatesharedintegertypedarray">
+        <h1>ValidateIntegerTypedArray ( _typedArray_ [ , _waitable_ ] )</h1>
+        <p>The abstract operation ValidateIntegerTypedArray takes one argument _typedArray_ and an optional Boolean _waitable_. It performs the following steps:</p>
```
ljharb: The rename alone is editorial.

spec.html
syg: Ack.

spec.html
syg: I have no opinions here, we have precedents for both, but for pre-declaration it's mostly for variables that are actually mutable, like an array index that gets incremented.

I'll change this to "let" in both branches for now?

spec.html
syg: Preferably not because the rename would be misleading without changing this line. The only reason it's renamed is because the AO now applies to all TypedArrays, not just SAB-backed ones.

spec.html
syg: See comment above.

syg:
Addressed review and rebased.

spec.html
ljharb: Sure, but the new name isn't wrong even if it only applies to shared array buffers.

syg: I don't understand why I would land a renaming PR that is less clear, separately.

ljharb: right, but it makes the commit separable, even if it's in the same PR. iow, this commit is normative, not editorial, as it's currently labelled.

syg: Are you saying you'd prefer that if this were a separate commit at all, if that separate commit must be normative, it should just be folded into the other one? That's fine with me. It ended up being a separate commit because it was a logical chunk of work.

ljharb: mainly just that i'd prefer individual commits be atomic and labelled properly; if the intention is to fold them all up, then it doesn't matter.

rwaldron:
@syg I'm going to add this to the end of the priority testing queue

syg:
@erights raised the following good question: "Are different optimization allowed between a single-agent program that uses Atomics on a non-shared SAB and a single-agent program that uses Atomics on an AB?"

If one takes different optimizations to mean different observable behavior, then the answer is no. On paper, since Atomics on ABs do not introduce SeqCst events (or, indeed, any events), an implementation is allowed to do things to Atomics on ABs that it's not allowed to do to Atomics on SABs, like coalesce writes, reorder them, etc. But it's only allowed to do that if it can prove that such changes are non-observable inside that single agent. SABs need extra guidance because observability is extended to the agent cluster -- proving single agent observability is insufficient.

However, if one takes different optimizations to mean can different codegen be observed by something outside of the spec's normative control like devtools, then the answer is yes. This is something the spec does not legislate anyhow.

In practice, I reckon the answer to both interpretations above is no. I see no reason why compiler authors would go the length of writing optimizations for Atomics used on ABs instead of reusing the same optimizations as Atomics used on SABs.

Cc @tlivley and @lars-t-hansen for thoughts from the wasm side.

tlively:
I would expect different (semantics-preserving) optimizations to be allowed for the reasons you mention. I assume doing going out of the way to do extra optimizations of unshared atomic operations would be low priority for implementors since that is a rather niche situation. On the other hand, it might not be difficult to arrange for a compiler to notice that the atomics are not shared early in the pipeline and lower them to nonatomic operations that can subsequently be optimized using the same code paths as normal reads and writes.

erights:
To be clear, I was asking, not advocating. If the observable behavior is already equivalent, then we're done.

I do *not* think that the spec for AB should be loosened in any way, to enable optimizations that we're not currently doing for AB. Such equivalence of allowed behavior is not attractive to me, even if everyone else were happy with it.

Let me ask a different question: If the allowed behavior is not equivalent: Is the allowed observable behavior on AB a subset of the allowed observable behavior on SAB? If not, I'd like to understand the difference. Until I do understand the difference, I'm still not yet advocating that anything be changed.

syg:
> Let me ask a different question: If the allowed behavior is not equivalent: Is the allowed observable behavior on AB a subset of the allowed observable behavior on SAB? If not, I'd like to understand the difference.

The allowed observable single-thread behavior of AB and the allowed observable behavior of non-shared SAB are the same. What differs is what codegen is allowed, but since there are no concurrent ways to observe the SAB since it's non-shared, the differently generated code have the same observable behavior.

The _only_ way you _might_ observe a difference is a non-normative host-implemented observation utilities like a debugger.

tl;dr The observable behavior is already equivalent given how people currently understand how to optimize single-threaded JS, so I think we're done.

lars-t-hansen:
> I would expect different (semantics-preserving) optimizations to be allowed for the reasons you mention. I assume doing going out of the way to do extra optimizations of unshared atomic operations would be low priority for implementors since that is a rather niche situation. On the other hand, it might not be difficult to arrange for a compiler to notice that the atomics are not shared early in the pipeline and lower them to nonatomic operations that can subsequently be optimized using the same code paths as normal reads and writes.

I agree.

spec.html
```diff
@@ -37972,6 +38045,7 @@ <h1>Atomics.notify ( _typedArray_, _index_, _count_ )</h1>
         1. Let _arrayTypeName_ be _typedArray_.[[TypedArrayName]].
         1. Let _elementSize_ be the Element Size value specified in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
         1. Let _indexedPosition_ be (_i_ &times; _elementSize_) + _offset_.
+        1. If IsSharedArrayBuffer(_buffer_) is *false*, return 0.
```
rwaldron, line 38048: Is there any particular reason that this couldn't occur as Step 2?

syg: I wanted to keep all the observable behavior (like the calls to ValidateAtomicAccess and ToInteger(_count_)) between the the 0-because-of-`ArrayBuffer` special case and the 0-because-no-one-is-waiting case the same.

rwaldron: Ok, but that changes the order in which validation of the arguments occur. In the current spec, validation of the  `index` value will not be reached if `ValidateSharedIntegerTypedArray()` results in a thrown exception. If "return 0" replaces "throw a TypeError exception from ValidateSharedIntegerTypedArray", then "return 0" needs to be the new step 2; if it's not, then ValidateAtomicAccess is called and `index` is evaluated, which means that an additional, unintended observable behavior has been created.

```
const i64a = new BigInt64Array(
  new ArrayBuffer(BigInt64Array.BYTES_PER_ELEMENT * 8)
);

const poisonedIndex = {
  valueOf() {
    throw new Error('evaluating poisonedIndex resulted in this new observable behavior');
  }
};

// This ok... 
Atomics.notify(i64a, 0, 0) === 0;

// But this will fail because poisonedIndex is evaluated. 
Atomics.notify(i64a, poisonedIndex, 1) === 0;
```

syg: It was my intention to make index validation observable in symmetry with the SAB case.

Are you saying that since `Atomics.notify` is hardcoded to return 0 on ABs anyways, there's no reason it should validate the index? That's weird to me, since `Atomics.notify(someSAB, poisonedIndex, 1)` currently throws because of `poisonedIndex`, regardless of there being waiters or not. I think the simplest mental model for `Atomics.notify` on ArrayBuffers is something like "there are always zero waiters, since waiting is impossible". In that mental model, I don't see any reason to skip index validation.

rwaldron: > Are you saying that since Atomics.notify is hardcoded to return 0 on ABs anyways, there's no reason it should validate the index?

Yes, that's right. 

> That's weird to me, since Atomics.notify(someSAB, poisonedIndex, 1) currently throws because of poisonedIndex,

Sure, but only because `someSAB` is a `SharedArrayBuffer`. In the current design, if it were an `ArrayBuffer`, it would fail before reaching evaluating `poisonedIndex`, and I'm suggesting that we continue to enforce the "does not reach evaluation of `index`" behavior, by placing the new special case handling of `ArrayBuffer` before the call to `ValidateAtomicAccess`.

syg: > Sure, but only because someSAB is a SharedArrayBuffer. In the current design, if it were an ArrayBuffer, it would fail before reaching evaluating poisonedIndex, and I'm suggesting that we continue to enforce the "does not reach evaluation of index" behavior, by placing the new special case handling of ArrayBuffer before the call to ValidateAtomicAccess.

But why? This would make it different with every other Atomics method that is relaxed by this PR to also work on ABs.

rwaldron: > This would make it different from every other Atomics method that is relaxed by this PR to also work on ABs

I think I'm convinced by this: the operations that _are_ changing, will all change in a consistent way. Thanks!

anba:
This seems to be missing detached array buffer checks.

anba:
The `Set` part of `GetModifySetValueInBuffer` for non-shared buffers seems to be missing, too. And as currently written, it returns the result after applying `op` for non-shared buffers, whereas for shared buffers, the original read value is returned.

syg:
Yeah this PR is very wrong. Thanks @anba, will fix and rebase.

syg:
Rebased and fixed the issues @anba pointed out.

spec.html
```diff
@@ -37695,7 +37700,8 @@ <h1>ValidateSharedIntegerTypedArray ( _typedArray_ [ , _waitable_ ] )</h1>
             1. If ! IsUnclampedIntegerElementType(_type_) is *false* and ! IsBigIntElementType(_type_) is *false*, throw a *TypeError* exception.
           1. Assert: _typedArray_ has a [[ViewedArrayBuffer]] internal slot.
           1. Let _buffer_ be _typedArray_.[[ViewedArrayBuffer]].
-          1. If IsSharedArrayBuffer(_buffer_) is *false*, throw a *TypeError* exception.
+          1. Assert: If IsSharedArrayBuffer(_buffer_) is *true*, then IsDetachedBuffer(_buffer_) is *false*.
```
ljharb, line 37703: would it make more sense to have this assertion inside `IsDetachedBuffer`?

spec.html
```diff
@@ -37681,9 +37686,9 @@ <h1>The Atomics Object</h1>
     <emu-clause id="sec-abstract-operations-for-atomics">
       <h1>Abstract Operations for Atomics</h1>
 
-      <emu-clause id="sec-validatesharedintegertypedarray" aoid="ValidateSharedIntegerTypedArray">
-        <h1>ValidateSharedIntegerTypedArray ( _typedArray_ [ , _waitable_ ] )</h1>
-        <p>The abstract operation ValidateSharedIntegerTypedArray takes argument _typedArray_ and optional argument _waitable_ (a Boolean). It performs the following steps when called:</p>
+      <emu-clause id="sec-validateintegertypedarray" aoid="ValidateIntegerTypedArray" oldid="sec-validatesharedintegertypedarray">
+        <h1>ValidateIntegerTypedArray ( _typedArray_ [ , _waitable_ ] )</h1>
+        <p>The abstract operation ValidateIntegerTypedArray takes argument _typedArray_ and optional argument _waitable_ (a Boolean). It performs the following steps when called:</p>
```
ljharb, line 37691: it seems weird for me (both before and after this change) that an abstract operation called "validate" returns something. is there a better name for this abstract op that conveys that?

spec.html
syg: Sure.

spec.html
syg: I'd consider that outside the scope of this PR.

One option is to tack on another verb, like `ValidateIntegerTypedArrayAndGetBuffer` and `ValidateAtomicAccessAndGetIndex`, that's pretty verbose.

Another option is to `ToValidIntegerTypedArrayBuffer`, but it might be hard to understand the intent of the AO without reading what it does.

In any case, we should do those as a follow-on PR.

ljharb: Maybe `GetSharedIntegerTypedArrayBuffer`?

I'm not sure it's outside the scope; you're renaming the op in this PR, so any concerns about the name seem thus in scope of the PR.

spec.html
syg: I removed this assert because it turned out to be encoding useless information. IsSharedArrayBuffer already returned *false* of a `null` [[ArrayBufferData]].

spec.html
syg: It no longer applies to shared buffers only, which is why this renames that op to delete the word "shared". It doesn't follow to me that a complete rename is also in scope.

Also `GetIntegerTypedArrayBuffer` doesn't capture most of what the AO does: validating the kind of TypedArray the argument is depending on the parameters.

anba:
We need to have additional detached buffer checks:
- AtomicReadModifyWrite, after ToBigInt/ToInteger.
- Atomics.compareExchange, after ToBigInt/ToInteger.
- Atomics.store, after ToBigInt/ToInteger.
- AtomicLoad, after ValidateAtomicAccess.
- (I'm not sure if we also want to check for detached buffers in Atomics.notify, because we're returning 0 anyway.)

ValidateAtomicAccess also needs to a slight modification: Step 3 ("Let length be typedArray.[[ArrayLength]]") should be moved before step 2 ("Let accessIndex be ? ToIndex(requestIndex)."), because we don't want to read [[ArrayLength]] on a typed array with a detached buffer. I'm not sure that rule is documented anywhere, but IIRC Allen made sure this never happens when he wrote the ES6 spec. And it also makes things a bit easier for implementors, because some implementations return 0 for [[ArrayLength]] with a detached buffer, but per spec the original length should be returned.

syg:
> - AtomicReadModifyWrite, after ToBigInt/ToInteger.
> - Atomics.compareExchange, after ToBigInt/ToInteger.
> - Atomics.store, after ToBigInt/ToInteger.
> - AtomicLoad, after ValidateAtomicAccess.

For these, I'll delay the detached check to be as late as possible. I don't see a good reason to check detachment early, do you, @anba?

> - (I'm not sure if we also want to check for detached buffers in Atomics.notify, because we're returning 0 anyway.)

I think, out of consistency of errors, yes. Trying to access a detached buffer is an egregious enough behavior that we shouldn't hide it even if we could return an answer for Atomics.notify.

> ValidateAtomicAccess also needs to a slight modification: Step 3 ("Let length be typedArray.[[ArrayLength]]") should be moved before step 2 ("Let accessIndex be ? ToIndex(requestIndex)."), because we don't want to read [[ArrayLength]] on a typed array with a detached buffer. I'm not sure that rule is documented anywhere, but IIRC Allen made sure this never happens when he wrote the ES6 spec. And it also makes things a bit easier for implementors, because some implementations return 0 for [[ArrayLength]] with a detached buffer, but per spec the original length should be returned.

It looks like this invariant is already broken in the following place. [IIEO [[DefineProperty]]](https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects-defineownproperty-p-desc) calls IsValidIntegerIndex, which gets [[ArrayLength]] without checking if the underlying buffer is detached. It's not until step 3.b.vi.2, when IntegerIndexedElementSet is called, that the detachment check is done. I think the property you said holds everywhere else.

Noted about some implementations returning 0 for [[ArrayLength]] for views with detached buffers. (V8 is incidentally buggy non-compliant for [[DefineProperty]]: we throw a RangeError instead of a TypeError.)

anba:
> For these, I'll delay the detached check to be as late as possible. I don't see a good reason to check detachment early, do you, @anba?

TypedArray built-in functions normally (*) check for detached buffers twice (**): Once when validating the input is indeed a TypedArray (through [ValidateTypedArray](https://tc39.es/ecma262/#sec-validatetypedarray)) and then later one when actually performing the work on the underlying data. 

(*) There are exceptions like in [%TypedArray%.prototype.set ( array [ , offset ] )](https://tc39.es/ecma262/#sec-%typedarray%.prototype.set-array-offset), which doesn't call `ValidateTypedArray`. But all *modern* TypedArray built-ins perform multiple checks.
(**) Can be more than twice when it happens through Integer-Indexed exotic object internal methods.

anba:
> It looks like this invariant is already broken in the following place. [...]

This doesn't mean we should break in more places, right? :-)

I prefer to move the access to [[ArrayLength]], because it makes life easier for implementers, because then implementations can implement the spec step by step, instead of having to annotate why the implementation is doing steps out of order.

syg:
I'm convinced by the consistency argument that most AB methods already do multiple detach checks. I've added a second detach check before [[ArrayLength]] where appropriate. Notably, AtomicLoad, Atomics.notify, and Atomics.wait only have a single check.

anba:
I'd prefer to have the first detached array buffer check in `ValidateIntegerTypedArray` instead of `ValidateAtomicAccess`. Similar to how `ValidateTypedArray` works.

---

We could even call `ValidateTypeArray` here, if we want to reduce some duplicate steps:
1. If waitable is not present, set waitable to false.
1. Let buffer be ? ValidateTypedArray(typedArray).
1. Let typeName be typedArray.[[TypedArrayName]].
1. Let type be the Element Type value in Table 61 for typeName.
 1. If waitable is true, then
    1. If typeName is not "Int32Array" or "BigInt64Array", throw a TypeError exception.
1. Else,
     1. If ! IsUnclampedIntegerElementType(type) is false and ! IsBigIntElementType(type) is false, throw a TypeError exception.
1. Return buffer.

syg:
I thought about that, but then we would have 2-3 detach checks for some Atomics on ABs:

1. In ValidateIntegerTypedArray
1. In ValidateAtomicAccess, after ? ToIndex, to preserve the property you want of not getting [[ArrayLength]] on detached TAs.
1. Before doing the actual operation, depending on the op.

syg:
> That one can be changed without a detached check by simply moving the access to [[ArrayLength]] before calling ToIndex. Basically what implementations like SpiderMonkey or V8 already have to do.

That seems fine, I'll do that.

I think misunderstood your original implementation-freedom point: you were saying even though [[ArrayLength]] isn't zeroed on buffer detaching, implementations might want to do that. I interpreted that to mean you wanted ValidateAtomicAccess to throw a TypeError if the buffer were detached. While what you were saying was it's fine to throw a RangeError on detached buffer, as long as it's explicit in the spec algorithm that the value of [[ArrayLength]] is manually cached in a local before possible detachment.

anba:
Yes, exactly that. Sorry if I couldn't make it clear enough!

spec.html
```diff
@@ -37705,8 +37707,8 @@ <h1>ValidateAtomicAccess ( _typedArray_, _requestIndex_ )</h1>
         <p>The abstract operation ValidateAtomicAccess takes arguments _typedArray_ and _requestIndex_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: _typedArray_ is an Object that has a [[ViewedArrayBuffer]] internal slot.
-          1. Let _accessIndex_ be ? ToIndex(_requestIndex_).
           1. Let _length_ be _typedArray_.[[ArrayLength]].
+          1. Let _accessIndex_ be ? ToIndex(_requestIndex_).
```
ljharb, line 37711: is there any reason to flip these? it wouldn't be observable either way, ofc.

syg: This is what @anba and I have been talking about in the thread. The invariant is that [[ArrayLength]] shouldn't be accessed on a TA whose buffer is detached, which ToIndex can cause to happen. One benefit of doing that is an implementation can implement getting a TA's length as returning 0 when its buffer is detached without thinking hard about if it is observable.

The reordering is strictly speaking not observable but the intent is clearer to have it come first, to lessen the chances a naive implementation is wrong.

spec.html
```diff
@@ -38101,11 +38103,12 @@ <h1>NotifyWaiter ( _WL_, _W_ )</h1>
         <h1>AtomicReadModifyWrite ( _typedArray_, _index_, _value_, _op_ )</h1>
         <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a read-modify-write modification function). _op_ takes two List of byte values arguments and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps when called:</p>
         <emu-alg>
-          1. Let _buffer_ be ? ValidateSharedIntegerTypedArray(_typedArray_).
+          1. Let _buffer_ be ? ValidateIntegerTypedArray(_typedArray_).
           1. Let _i_ be ? ValidateAtomicAccess(_typedArray_, _index_).
           1. Let _arrayTypeName_ be _typedArray_.[[TypedArrayName]].
           1. If _typedArray_.[[ContentType]] is ~BigInt~, let _v_ be ? ToBigInt(_value_).
           1. Otherwise, let _v_ be ? ToInteger(_value_).
+          1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
```
bakkot, line 38111: Isn't this redundant? `ValidateIntegerTypedArray` already calls `ValidateTypedArray`, and that performs the detached buffer check.

Ditto most of the other new `IsDetachedBuffer` checks.

syg: Ah, no, it's because ? ToInteger() calls ToNumber, which could call ToPrimitive, which could call arbitrary code. :(

bakkot: Uuuugh.

That's worth adding `NOTE`s for, I think, especially since an implementation which thinks the same thing I did could plausibly end up with a security bug. E.g. after each of these surprisingly-necessary steps, add a

> 1. NOTE: The above check is not redundant with the check in ValidateIntegerTypedArray because the call to `ToInteger` on the preceding line can have arbitrary side effects, which could cause the buffer to become detached after the call to ValidateIntegerTypedArray.

spec.html
```diff
@@ -37266,23 +37266,28 @@ <h1>SetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _isTypedArra
...
+            1. Let _eventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
+            1. Let _rawBytesRead_ be a List of length _elementSize_ of nondeterministically chosen byte values.
+            1. NOTE: In implementations, _rawBytesRead_ is the result of a load-link, of a load-exclusive, or of an operand of a read-modify-write instruction on the underlying hardware. The nondeterminism is a semantic prescription of the memory model to describe observable behaviour of hardware with weak consistency.
+            1. Let _rmwEvent_ be ReadModifyWriteSharedMemory { [[Order]]: ~SeqCst~, [[NoTear]]: *true*, [[Block]]: _block_, [[ByteIndex]]: _byteIndex_, [[ElementSize]]: _elementSize_, [[Payload]]: _rawBytes_, [[ModifyOp]]: _op_ }.
+            1. Append _rmwEvent_ to _eventList_.
+            1. Append Chosen Value Record { [[Event]]: _rmwEvent_, [[ChosenValue]]: _rawBytesRead_ } to _execution_.[[ChosenValues]].
+          1. Else,
+            1. Let _rawBytesRead_ be a List of _elementSize_ containing, in order, the _elementSize_ sequence of bytes starting with _block_[_byteIndex_].
```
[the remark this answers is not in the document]
syg: Will change to that.

PR #1912, opened 2020-03-22
Normative: Promise.{all,allSettled,race} should check resolve is a function before opening their iterable

### LINKED ISSUE #1902: Promise.{all,allSettled,race} should check resolve is a function before opening their iteratable
Currently, the spec for Promise.{all,allSettled,race} requires calling the first arguments iterator before checking if this.resolve is a function. Since this.resolve is hoisted out of the loop it makes writing the function in JS much more complicated.

If I were going to write a polyfill for Promise.{all,allSettled,race} I would expect it to look something like: 

```js
function all(iterable) {
    ...
    var promiseResolve = this.resolve;
    if (typeof promiseResolve !== "function")
        #throwTypeError("Promise resolve is not a function");
    for (let i of iterable)
        ...
```

Instead you need something like:

```js
function all(iterable) {
    ...
    var iterator = iterable.#SymbolIterator();

    try {
        var promiseResolve = this.resolve;
        if (typeof promiseResolve !== "function")
            #throwTypeError("Promise.resolve is not a function");
    } catch (error) {
        try {
            iterator.return();
        } catch {
            // Inner error is ignored: see steps 6-7 of https://tc39.es/ecma262/#sec-iteratorclose
        }
        throw error;
    }
    // Since for-of loop once more looks up the @@iterator property of a given iterable,
    // it could be observable if the user defines a getter for @@iterator.
    // To avoid this situation, we define a wrapper object that @@iterator just returns a given iterator.
    var wrapper = {};
    wrapper.@@iterator = function() { return iterator; };

    for (var value of wrapper) {
          ...
```

Alternatively, you could iterate the iterator by hand but getting that right is actually also tricky...

Given that at one point every engine implemented this incorrectly, (https://test262.report/browse/built-ins/Promise/all/invoke-resolve-get-error-close.js?date=2019-07-06) it seems like it might be reasonable to change the spec to match the first implementation. Thoughts?

kmiller68:
This change makes promise.{all,allSettled,race} behave more like a canonical JS implementation.  In particular, the resolve property is now checked to be callable before beginning iteration on the first argument.

Closes #1902.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -40100,9 +40100,11 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _C_ be the *this* value.
           1. Let _promiseCapability_ be ? NewPromiseCapability(_C_).
+          1. Let _promiseResolve_ be GetPromiseResolve(_C_).
+          1. IfAbruptRejectPromise(_promiseResolve_, _promiseCapability_).
           1. Let _iteratorRecord_ be GetIterator(_iterable_).
           1. IfAbruptRejectPromise(_iteratorRecord_, _promiseCapability_).
-          1. Let _result_ be PerformPromiseAll(_iteratorRecord_, _C_, _promiseCapability_).
+          1. Let _result_ be PerformPromiseAll(_iteratorRecord_, _C_, _promiseCapability_, _promiseResolve_).
           1. If _result_ is an abrupt completion, then
             1. If _iteratorRecord_.[[Done]] is *false*, set _result_ to IteratorClose(_iteratorRecord_, _result_).
             1. IfAbruptRejectPromise(_result_, _promiseCapability_).
@@ -40113,16 +40115,26 @@ THH:mm:ss.sss
           <p>The `all` function requires its *this* value to be a constructor function that supports the parameter conventions of the Promise constructor.</p>
         </emu-note>
 
+        <emu-clause id="sec-getpromiseresolve" aoid="GetPromiseResolve">
+          <h1>Runtime Semantics: GetPromiseResolve ( _promiseConstructor_ )</h1>
+          <p>The abstract operation GetPromiseResolve takes argument _promiseConstructor_. It performs the following steps when called:</p>
+          <emu-alg>
+            1. Assert: IsConstructor(_promiseConstructor_) is *true*.
+            1. Let _promiseResolve_ be ? Get(_promiseConstructor_, *"resolve"*).
+            1. If IsCallable(_promiseResolve_) is *false*, throw a *TypeError* exception.
+            1. Return _promiseResolve_.
+          </emu-alg>
+        </emu-clause>
+
         <emu-clause id="sec-performpromiseall" aoid="PerformPromiseAll">
-          <h1>Runtime Semantics: PerformPromiseAll ( _iteratorRecord_, _constructor_, _resultCapability_ )</h1>
-          <p>The abstract operation PerformPromiseAll takes arguments _iteratorRecord_, _constructor_, and _resultCapability_. It performs the following steps when called:</p>
+          <h1>Runtime Semantics: PerformPromiseAll ( _iteratorRecord_, _constructor_, _resultCapability_, _promiseResolve_ )</h1>
+          <p>The abstract operation PerformPromiseAll takes arguments _iteratorRecord_, _constructor_, _resultCapability_, and _promiseResolve_. It performs the following steps when called:</p>
           <emu-alg>
             1. Assert: IsConstructor(_constructor_) is *true*.
             1. Assert: _resultCapability_ is a PromiseCapability Record.
+            1. Assert: IsCallable(_promiseResolve_) is *true*.
             1. Let _values_ be a new empty List.
             1. Let _remainingElementsCount_ be the Record { [[Value]]: 1 }.
-            1. Let _promiseResolve_ be ? Get(_constructor_, *"resolve"*).
-            1. If ! IsCallable(_promiseResolve_) is *false*, throw a *TypeError* exception.
             1. Let _index_ be 0.
             1. Repeat,
               1. Let _next_ be IteratorStep(_iteratorRecord_).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -40003,16 +40007,25 @@ <h1>Promise.all ( _iterable_ )</h1>
           <p>The `all` function requires its *this* value to be a constructor function that supports the parameter conventions of the `Promise` constructor.</p>
         </emu-note>
 
+        <emu-clause id="sec-getpromiseresolve" aoid="GetPromiseResolve">
+          <h1>Runtime Semantics: GetPromiseResolve ( _promiseCapability_ )</h1>
+          <p>When the GetPromiseResolve abstract operation is call with argument _promiseCapability_, the following steps are taken:</p>
+          <emu-alg>
+            1. Let _promiseResolve_ be ? Get(_C_, *"resolve"*).
+            1. If IsCallable(_promiseResolve_) is *false*, throw a *TypeError* exception.
+            1. return _promiseResolve_.
```
```suggestion
            1. Return _promiseResolve_.
```

spec.html
```diff
@@ -40003,16 +40007,25 @@ <h1>Promise.all ( _iterable_ )</h1>
           <p>The `all` function requires its *this* value to be a constructor function that supports the parameter conventions of the `Promise` constructor.</p>
         </emu-note>
 
+        <emu-clause id="sec-getpromiseresolve" aoid="GetPromiseResolve">
+          <h1>Runtime Semantics: GetPromiseResolve ( _promiseCapability_ )</h1>
+          <p>When the GetPromiseResolve abstract operation is call with argument _promiseCapability_, the following steps are taken:</p>
+          <emu-alg>
+            1. Let _promiseResolve_ be ? Get(_C_, *"resolve"*).
```
```suggestion
            1. Assert: IsConstructor(_C_) is *true*.
            1. Let _promiseResolve_ be ? Get(_C_, *"resolve"*).
```

spec.html
```diff
@@ -40003,16 +40007,25 @@ <h1>Promise.all ( _iterable_ )</h1>
           <p>The `all` function requires its *this* value to be a constructor function that supports the parameter conventions of the `Promise` constructor.</p>
         </emu-note>
 
+        <emu-clause id="sec-getpromiseresolve" aoid="GetPromiseResolve">
+          <h1>Runtime Semantics: GetPromiseResolve ( _promiseCapability_ )</h1>
+          <p>When the GetPromiseResolve abstract operation is call with argument _promiseCapability_, the following steps are taken:</p>
```
ljharb: this should be called with `C` as well.

spec.html
```diff
@@ -39990,9 +39990,13 @@ <h1>Promise.all ( _iterable_ )</h1>
         <emu-alg>
           1. Let _C_ be the *this* value.
           1. Let _promiseCapability_ be ? NewPromiseCapability(_C_).
+          1. Let _promiseResolve_ be GetPromiseResolve(_promiseCapability_).
+          1. If _promiseResolve_ is an abrupt completion, then
+            1. IfAbruptRejectPromise(_promiseResolve_, _promiseCapability_).
+            1. Return Completion(_promiseResolve_).
```
```suggestion
            1. Return _promiseResolve_.
```
it's already an abrupt completion.

spec.html
```diff
@@ -40073,9 +40086,13 @@ <h1>Promise.allSettled ( _iterable_ )</h1>
...
```
```suggestion
            1. Return _promiseResolve_.
```
doesn't this mean the promise is rejected, but the function also throws? shouldn't it not throw in this case?

spec.html
```diff
@@ -40198,9 +40214,13 @@ <h1>Promise.race ( _iterable_ )</h1>
...
```
```suggestion
            1. Return _promiseResolve_.
```

spec.html
bakkot: This step actually don't make sense - [IfAbruptRejectPromise](https://tc39.es/ecma262/#sec-ifabruptrejectpromise) is a macro, and it causes the invoking function to return if passed an abrupt completion. So control can never reach here.

I think this step can be omitted entirely.

ljharb: That makes much more sense.

kmiller68: Ah, I didn't realize it was macro! That does indeed make so much more sense...

spec.html
kmiller68: Done

spec.html
kmiller68: It worked because IfAbruptRejectPromise is a macro rather than an operation. It's not revenant anymore because I changed the entire if-branch to be a single IfAbruptRejectPromise.

spec.html
kmiller68: Added.

kmiller68:
Looks, like @shvaikalesh graciously updated test262 for me with https://github.com/tc39/test262/pull/2648. (Thanks!)

I rebased against the current spec and fixed my summaries to match the newer style. I noticed that we use the phrasing `The abstract operation <Op> takes argument <Arg>.`, which seems odd to me but I kept it for consistency.

spec.html
```diff
@@ -40313,9 +40326,11 @@ <h1>Promise.race ( _iterable_ )</h1>
         <emu-alg>
           1. Let _C_ be the *this* value.
           1. Let _promiseCapability_ be ? NewPromiseCapability(_C_).
+          1. Let _promiseResolve_ be GetPromiseResolve(_C_).
+          1. IfAbruptRejectPromise(_promiseResolve_, _promiseCapability_).
           1. Let _iteratorRecord_ be GetIterator(_iterable_).
           1. IfAbruptRejectPromise(_iteratorRecord_, _promiseCapability_).
-          1. Let _result_ be PerformPromiseRace(_iteratorRecord_, _C_, _promiseCapability_).
+          1. Let _result_ be PerformPromiseRace(_iteratorRecord_, _C_, _promiseCapability_, _promiseResolve_).
```
ljharb, line 40333: PerformPromiseRace still only takes 3 arguments

kmiller68: Ah... sorry. I think that was a bad rebase fixed.

spec.html
```diff
@@ -40118,16 +40120,26 @@ <h1>Promise.all ( _iterable_ )</h1>
...
+            1. If IsCallable(_promiseResolve_) is *false*, throw a *TypeError* exception.
+            1. Return _promiseResolve_.
+          </emu-alg>
+        </emu-clause>
+
         <emu-clause id="sec-performpromiseall" aoid="PerformPromiseAll">
-          <h1>Runtime Semantics: PerformPromiseAll ( _iteratorRecord_, _constructor_, _resultCapability_ )</h1>
-          <p>The abstract operation PerformPromiseAll takes arguments _iteratorRecord_, _constructor_, and _resultCapability_. It performs the following steps when called:</p>
+          <h1>Runtime Semantics: PerformPromiseAll ( _iteratorRecord_, _constructor_, _resultCapability_, _promiseResolve_ )</h1>
+          <p>The abstract operation PerformPromiseAll takes arguments _iteratorRecord_, _constructor_, _resultCapability_, and _promiseResolve. It performs the following steps when called:</p>
```
bakkot, line 40136:
```suggestion
          <p>The abstract operation PerformPromiseAll takes arguments _iteratorRecord_, _constructor_, _resultCapability_, and _promiseResolve_. It performs the following steps when called:</p>
```

(The linter doesn't yet know how to catch these, sorry. Someday!)

PR #1914, opened 2020-03-24
Editorial: use consistent wording for abstract operation preambles

bakkot:
This unifies the wording used to introduce abstract operations. The intent is that the new wording should provide all and only the information in the current wording (except in a couple cases where the existing wording is obviously wrong). This is a step towards #545 (structured headers could be used to generate this text) and #1796 (which wants a consistent place to describe the return type of abstract operations).

We're wildly inconsistent about what information we bother to include in these descriptions. This is more obvious after this PR because the wording is now consistent, but since it doesn't actually make the information included any less consistent, I would prefer we not try to address that here.

~This PR also removes abstract operation preambles entirely when they contain no information other than the name of the operation and its arguments (though I am open to going the other direction and such adding introductions everywhere they are currently absent).~ Edit: went the other direction, this PR now add such introductions to abstract operations which were missing them.

After we land this I will try to make a PR to ecmarkup which enforces this style, at least to some extent.

Thanks to @jmdyck for [ecmaspeak-py](https://github.com/jmdyck/ecmaspeak-py), which has hardcoded all the many, many forms of preamble currently in use and allowed me to extract the information I needed from them.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1751,6 +1751,7 @@
 
         <emu-clause id="sec-numberbitwiseop" aoid="NumberBitwiseOp">
           <h1>NumberBitwiseOp ( _op_, _x_, _y_ )</h1>
+          <p>The abstract operation NumberBitwiseOp takes arguments _op_, _x_, and _y_. It performs the following steps when called:</p>
           <emu-alg>
             1. Let _lnum_ be ! ToInt32(_x_).
             1. Let _rnum_ be ! ToUint32(_y_).
@@ -1962,6 +1963,7 @@
 
         <emu-clause id="sec-binaryand" aoid="BinaryAnd">
           <h1>BinaryAnd ( _x_, _y_ )</h1>
+          <p>The abstract operation BinaryAnd takes arguments _x_ and _y_. It performs the following steps when called:</p>
           <emu-alg>
             1. Assert: _x_ is 0 or 1.
             1. Assert: _y_ is 0 or 1.
@@ -1972,6 +1974,7 @@
 
         <emu-clause id="sec-binaryor" aoid="BinaryOr">
           <h1>BinaryOr ( _x_, _y_ )</h1>
+          <p>The abstract operation BinaryOr takes arguments _x_ and _y_. It performs the following steps when called:</p>
           <emu-alg>
             1. Assert: _x_ is 0 or 1.
             1. Assert: _y_ is 0 or 1.
@@ -1982,6 +1985,7 @@
 
         <emu-clause id="sec-binaryxor" aoid="BinaryXor">
           <h1>BinaryXor ( _x_, _y_ )</h1>
+          <p>The abstract operation BinaryXor takes arguments _x_ and _y_. It performs the following steps when called:</p>
           <emu-alg>
             1. Assert: _x_ is 0 or 1.
             1. Assert: _y_ is 0 or 1.
@@ -1993,6 +1997,7 @@
 
         <emu-clause id="sec-bigintbitwiseop" aoid="BigIntBitwiseOp">
           <h1>BigIntBitwiseOp ( _op_, _x_, _y_ )</h1>
+          <p>The abstract operation BigIntBitwiseOp takes arguments _op_, _x_, and _y_. It performs the following steps when called:</p>
           <emu-alg>
             1. Assert: _op_ is *"&amp;"*, *"|"*, or *"^"*.
             1. Let _result_ be *0n*.
@@ -4069,10 +4074,9 @@
 
         <emu-clause id="await-fulfilled">
           <h1>Await Fulfilled Functions</h1>
-
           <p>An Await fulfilled function is an anonymous built-in function that is used as part of the Await specification device to deliver the promise fulfillment value to the caller as a normal completion. Each Await fulfilled function has an [[AsyncContext]] internal slot.</p>
-
           <p>When an Await fulfilled function is called with argument _value_, the following steps are taken:</p>
+
 
           <emu-alg>
             1. Let _F_ be the active function object.
@@ -4090,10 +4094,9 @@
 
         <emu-clause id="await-rejected">
           <h1>Await Rejected Functions</h1>
-
           <p>An Await rejected function is an anonymous built-in function that is used as part of the Await specification device to deliver the promise rejection reason to the caller as an abrupt throw completion. Each Await rejected function has an [[AsyncContext]] internal slot.</p>
-
           <p>When an Await rejected function is called with argument _reason_, the following steps are taken:</p>
+
 
           <emu-alg>
             1. Let _F_ be the active function object.
@@ -4136,7 +4139,7 @@
 
       <emu-clause id="sec-updateempty" aoid="UpdateEmpty">
         <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
-        <p>The abstract operation UpdateEmpty with arguments _completionRecord_ and _value_ performs the following steps:</p>
+        <p>The abstract operation UpdateEmpty takes arguments _completionRecord_ and _value_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: If _completionRecord_.[[Type]] is either ~return~ or ~throw~, then _completionRecord_.[[Value]] is not ~empty~.
           1. If _completionRecord_.[[Value]] is not ~empty~, return Completion(_completionRecord_).
@@ -4156,6 +4159,7 @@
 
       <emu-clause id="sec-getbase" aoid="GetBase" oldids="ao-getbase">
         <h1>GetBase ( _V_ )</h1>
+        <p>The abstract operation GetBase takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: Type(_V_) is Reference.
           1. Return the base value component of _V_.
@@ -4164,6 +4168,7 @@
 
       <emu-clause id="sec-getreferencedname" aoid="GetReferencedName" oldids="ao-getreferencedname">
         <h1>GetReferencedName ( _V_ )</h1>
+        <p>The abstract operation GetReferencedName takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: Type(_V_) is Reference.
           1. Return the referenced name component of _V_.
@@ -4172,6 +4177,7 @@
 
       <emu-clause id="sec-isstrictreference" aoid="IsStrictReference" oldids="ao-isstrictreference">
         <h1>IsStrictReference ( _V_ )</h1>
+        <p>The abstract operation IsStrictReference takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: Type(_V_) is Reference.
           1. Return the strict reference flag of _V_.
@@ -4180,6 +4186,7 @@
 
       <emu-clause id="sec-hasprimitivebase" aoid="HasPrimitiveBase" oldids="ao-hasprimitivebase">
         <h1>HasPrimitiveBase ( _V_ )</h1>
+        <p>The abstract operation HasPrimitiveBase takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: Type(_V_) is Reference.
           1. If Type(_V_'s base value component) is Boolean, String, Symbol, BigInt, or Number, return *true*; otherwise return *false*.
@@ -4188,6 +4195,7 @@
 
       <emu-clause id="sec-ispropertyreference" aoid="IsPropertyReference" oldids="ao-ispropertyreference">
         <h1>IsPropertyReference ( _V_ )</h1>
+        <p>The abstract operation IsPropertyReference takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: Type(_V_) is Reference.
           1. If either the base value component of _V_ is an Object or HasPrimitiveBase(_V_) is *true*, return *true*; otherwise return *false*.
@@ -4196,6 +4204,7 @@
 
       <emu-clause id="sec-isunresolvablereference" aoid="IsUnresolvableReference" oldids="ao-isunresolvablereference">
         <h1>IsUnresolvableReference ( _V_ )</h1>
+        <p>The abstract operation IsUnresolvableReference takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: Type(_V_) is Reference.
           1. If the base value component of _V_ is *undefined*, return *true*; otherwise return *false*.
@@ -4204,6 +4213,7 @@
 
       <emu-clause id="sec-issuperreference" aoid="IsSuperReference" oldids="ao-issuperreference">
         <h1>IsSuperReference ( _V_ )</h1>
+        <p>The abstract operation IsSuperReference takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: Type(_V_) is Reference.
           1. If _V_ has a thisValue component, return *true*; otherwise return *false*.
@@ -4212,6 +4222,7 @@
 
       <emu-clause id="sec-getvalue" aoid="GetValue">
         <h1>GetValue ( _V_ )</h1>
+        <p>The abstract operation GetValue takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. ReturnIfAbrupt(_V_).
           1. If Type(_V_) is not Reference, return _V_.
@@ -4233,6 +4244,7 @@
 
       <emu-clause id="sec-putvalue" aoid="PutValue">
         <h1>PutValue ( _V_, _W_ )</h1>
+        <p>The abstract operation PutValue takes arguments _V_ and _W_. It performs the following steps when called:</p>
         <emu-alg>
           1. ReturnIfAbrupt(_V_).
           1. ReturnIfAbrupt(_W_).
@@ -4261,6 +4273,7 @@
 
       <emu-clause id="sec-getthisvalue" aoid="GetThisValue">
         <h1>GetThisValue ( _V_ )</h1>
+        <p>The abstract operation GetThisValue takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: IsPropertyReference(_V_) is *true*.
           1. If IsSuperReference(_V_) is *true*, then
@@ -4271,6 +4284,7 @@
 
       <emu-clause id="sec-initializereferencedbinding" aoid="InitializeReferencedBinding">
         <h1>InitializeReferencedBinding ( _V_, _W_ )</h1>
+        <p>The abstract operation InitializeReferencedBinding takes arguments _V_ and _W_. It performs the following steps when called:</p>
         <emu-alg>
           1. ReturnIfAbrupt(_V_).
           1. ReturnIfAbrupt(_W_).
@@ -4291,7 +4305,7 @@
 
       <emu-clause id="sec-isaccessordescriptor" aoid="IsAccessorDescriptor">
         <h1>IsAccessorDescriptor ( _Desc_ )</h1>
-        <p>When the abstract operation IsAccessorDescriptor is called with Property Descriptor _Desc_, the following steps are taken:</p>
+        <p>The abstract operation IsAccessorDescriptor takes argument _Desc_ (a Property Descriptor or *undefined*). It performs the following steps when called:</p>
         <emu-alg>
           1. If _Desc_ is *undefined*, return *false*.
           1. If both _Desc_.[[Get]] and _Desc_.[[Set]] are absent, return *false*.
@@ -4301,7 +4315,7 @@
 
       <emu-clause id="sec-isdatadescriptor" aoid="IsDataDescriptor">
         <h1>IsDataDescriptor ( _Desc_ )</h1>
-        <p>When the abstract operation IsDataDescriptor is called with Property Descriptor _Desc_, the following steps are taken:</p>
+        <p>The abstract operation IsDataDescriptor takes argument _Desc_ (a Property Descriptor or *undefined*). It performs the following steps when called:</p>
         <emu-alg>
           1. If _Desc_ is *undefined*, return *false*.
           1. If both _Desc_.[[Value]] and _Desc_.[[Writable]] are absent, return *false*.
@@ -4311,7 +4325,7 @@
 
       <emu-clause id="sec-isgenericdescriptor" aoid="IsGenericDescriptor">
         <h1>IsGenericDescriptor ( _Desc_ )</h1>
-        <p>When the abstract operation IsGenericDescriptor is called with Property Descriptor _Desc_, the following steps are taken:</p>
+        <p>The abstract operation IsGenericDescriptor takes argument _Desc_ (a Property Descriptor or *undefined*). It performs the following steps when called:</p>
         <emu-alg>
           1. If _Desc_ is *undefined*, return *false*.
           1. If IsAccessorDescriptor(_Desc_) and IsDataDescriptor(_Desc_) are both *false*, return *true*.
@@ -4321,7 +4335,7 @@
 
       <emu-clause id="sec-frompropertydescriptor" aoid="FromPropertyDescriptor">
         <h1>FromPropertyDescriptor ( _Desc_ )</h1>
-        <p>When the abstract operation FromPropertyDescriptor is called with Property Descriptor _Desc_, the following steps are taken:</p>
+        <p>The abstract operation FromPropertyDescriptor takes argument _Desc_ (a Property Descriptor or *undefined*). It performs the following steps when called:</p>
         <emu-alg>
           1. If _Desc_ is *undefined*, return *undefined*.
           1. Let _obj_ be OrdinaryObjectCreate(%Object.prototype%).
@@ -4344,7 +4358,7 @@
 
       <emu-clause id="sec-topropertydescriptor" aoid="ToPropertyDescriptor">
         <h1>ToPropertyDescriptor ( _Obj_ )</h1>
-        <p>When the abstract operation ToPropertyDescriptor is called with object _Obj_, the following steps are taken:</p>
+        <p>The abstract operation ToPropertyDescriptor takes argument _Obj_. It performs the following steps when called:</p>
         <emu-alg>
           1. If Type(_Obj_) is not Object, throw a *TypeError* exception.
           1. Let _desc_ be a new Property Descriptor that initially has no fields.
@@ -4382,7 +4396,7 @@
 
       <emu-clause id="sec-completepropertydescriptor" aoid="CompletePropertyDescriptor">
         <h1>CompletePropertyDescriptor ( _Desc_ )</h1>
-        <p>When the abstract operation CompletePropertyDescriptor is called with Property Descriptor _Desc_, the following steps are taken:</p>
+        <p>The abstract operation CompletePropertyDescriptor takes argument _Desc_ (a Property Descriptor). It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: _Desc_ is a Property Descriptor.
           1. Let _like_ be the Record { [[Value]]: *undefined*, [[Writable]]: *false*, [[Get]]: *undefined*, [[Set]]: *undefined*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.
@@ -4430,7 +4444,7 @@
 
       <emu-clause id="sec-createbytedatablock" aoid="CreateByteDataBlock">
         <h1>CreateByteDataBlock ( _size_ )</h1>
-        <p>When the abstract operation CreateByteDataBlock is called with integer argument _size_, the following steps are taken:</p>
+        <p>The abstract operation CreateByteDataBlock takes argument _size_ (an integer). It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: _size_ &ge; 0.
           1. Let _db_ be a new Data Block value consisting of _size_ bytes. If it is impossible to create such a Data Block, throw a *RangeError* exception.
@@ -4441,7 +4455,7 @@
 
       <emu-clause id="sec-createsharedbytedatablock" aoid="CreateSharedByteDataBlock">
         <h1>CreateSharedByteDataBlock ( _size_ )</h1>
-        <p>When the abstract operation CreateSharedByteDataBlock is called with integer argument _size_, the following steps are taken:</p>
+        <p>The abstract operation CreateSharedByteDataBlock takes argument _size_ (an integer). It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: _size_ &ge; 0.
           1. Let _db_ be a new Shared Data Block value consisting of _size_ bytes. If it is impossible to create such a Shared Data Block, throw a *RangeError* exception.
@@ -4456,7 +4470,7 @@
 
       <emu-clause id="sec-copydatablockbytes" aoid="CopyDataBlockBytes">
         <h1>CopyDataBlockBytes ( _toBlock_, _toIndex_, _fromBlock_, _fromIndex_, _count_ )</h1>
-        <p>When the abstract operation CopyDataBlockBytes is called, the following steps are taken:</p>
+        <p>The abstract operation CopyDataBlockBytes takes arguments _toBlock_, _toIndex_, _fromBlock_, _fromIndex_, and _count_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: _fromBlock_ and _toBlock_ are distinct Data Block or Shared Data Block values.
           1. Assert: _fromIndex_, _toIndex_, and _count_ are integer values &ge; 0.
@@ -4501,7 +4515,7 @@
 
     <emu-clause id="sec-toprimitive" aoid="ToPrimitive" oldids="table-9">
       <h1>ToPrimitive ( _input_ [ , _PreferredType_ ] )</h1>
-      <p>The abstract operation ToPrimitive takes an _input_ argument and an optional argument _PreferredType_. The abstract operation ToPrimitive converts its _input_ argument to a non-Object type. If an object is capable of converting to more than one primitive type, it may use the optional hint _PreferredType_ to favour that type. Conversion occurs according to the following algorithm:</p>
+      <p>The abstract operation ToPrimitive takes argument _input_ and optional argument _PreferredType_. It converts its _input_ argument to a non-Object type. If an object is capable of converting to more than one primitive type, it may use the optional hint _PreferredType_ to favour that type. It performs the following steps when called:</p>
       <emu-alg>
         1. Assert: _input_ is an ECMAScript language value.
         1. If Type(_input_) is Object, then
@@ -4525,7 +4539,7 @@
 
       <emu-clause id="sec-ordinarytoprimitive" aoid="OrdinaryToPrimitive">
         <h1>OrdinaryToPrimitive ( _O_, _hint_ )</h1>
-        <p>When the abstract operation OrdinaryToPrimitive is called with arguments _O_ and _hint_, the following steps are taken:</p>
+        <p>The abstract operation OrdinaryToPrimitive takes arguments _O_ and _hint_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: Type(_O_) is Object.
           1. Assert: Type(_hint_) is String and its value is either *"string"* or *"number"*.
@@ -4545,7 +4559,7 @@
 
     <emu-clause id="sec-toboolean" aoid="ToBoolean">
       <h1>ToBoolean ( _argument_ )</h1>
-      <p>The abstract operation ToBoolean converts _argument_ to a value of type Boolean according to <emu-xref href="#table-10"></emu-xref>:</p>
+      <p>The abstract operation ToBoolean takes argument _argument_. It converts _argument_ to a value of type Boolean according to <emu-xref href="#table-10"></emu-xref>:</p>
       <emu-table id="table-10" caption="ToBoolean Conversions">
         <table>
           <tbody>
@@ -4628,7 +4642,7 @@
 
     <emu-clause id="sec-tonumeric" aoid="ToNumeric">
       <h1>ToNumeric ( _value_ )</h1>
-      <p>The abstract operation ToNumeric returns _value_ converted to a numeric value of type Number or BigInt. This abstract operation functions as follows:</p>
+      <p>The abstract operation ToNumeric takes argument _value_. It returns _value_ converted to a numeric value of type Number or BigInt. It performs the following steps when called:</p>
       <emu-alg>
         1. Let _primValue_ be ? ToPrimitive(_value_, hint Number).
         1. If Type(_primValue_) is BigInt, return _primValue_.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -4124,7 +4122,6 @@ <h1>ThrowCompletion</h1>
 
       <emu-clause id="sec-updateempty" aoid="UpdateEmpty">
         <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
-        <p>The abstract operation UpdateEmpty with arguments _completionRecord_ and _value_ performs the following steps:</p>
```
ljharb: i think i'd prefer to always include a preamble; but if you wouldn't want to make it correct/consistent in this PR, then i'm also fine landing this first, and then adding correct preambles in a followup.

spec.html
```diff
@@ -4309,7 +4306,7 @@ <h1>IsGenericDescriptor ( _Desc_ )</h1>
 
       <emu-clause id="sec-frompropertydescriptor" aoid="FromPropertyDescriptor">
         <h1>FromPropertyDescriptor ( _Desc_ )</h1>
-        <p>When the abstract operation FromPropertyDescriptor is called with Property Descriptor _Desc_, the following steps are taken:</p>
+        <p>The abstract operation FromPropertyDescriptor takes argument _Desc_ (a Property Descriptor or *undefined*). It performs the following steps:</p>
```
ljharb: I notice that in some places, we have explicit asserts for any preamble-declared types, but eg in this one, we don't. Is that an example of the inconsistency you don't want to handle in this PR?

spec.html
```diff
@@ -5322,7 +5317,7 @@ <h1>Testing and Comparison Operations</h1>
 
     <emu-clause id="sec-requireobjectcoercible" aoid="RequireObjectCoercible">
       <h1>RequireObjectCoercible ( _argument_ )</h1>
-      <p>The abstract operation RequireObjectCoercible throws an error if _argument_ is a value that cannot be converted to an Object using ToObject. It is defined by <emu-xref href="#table-14"></emu-xref>:</p>
+      <p>The abstract operation RequireObjectCoercible takes argument _argument_. It throws an error if _argument_ is a value that cannot be converted to an Object using ToObject. It is defined by <emu-xref href="#table-14"></emu-xref>:</p>
```
```suggestion
      <p>The abstract operation RequireObjectCoercible takes argument _argument_ (an ECMAScript language value). It throws an error if _argument_ is a value that cannot be converted to an Object using ToObject. It is defined by <emu-xref href="#table-14"></emu-xref>:</p>
```
per some of the below ones where this parenthetical exists

spec.html
```diff
@@ -5303,7 +5298,7 @@ <h1>CanonicalNumericIndexString ( _argument_ )</h1>
 
     <emu-clause id="sec-toindex" aoid="ToIndex">
       <h1>ToIndex ( _value_ )</h1>
-      <p>The abstract operation ToIndex returns _value_ argument converted to a non-negative integer if it is a valid integer index value. This abstract operation functions as follows:</p>
+      <p>The abstract operation ToIndex takes argument _value_. It returns _value_ argument converted to a non-negative integer if it is a valid integer index value. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation ToIndex takes argument _value_ (an ECMAScript language value). It returns _value_ argument converted to a non-negative integer if it is a valid integer index value. It performs the following steps:</p>
```

spec.html
```diff
@@ -5290,7 +5285,7 @@ <h1>ToLength ( _argument_ )</h1>
 
     <emu-clause id="sec-canonicalnumericindexstring" aoid="CanonicalNumericIndexString">
       <h1>CanonicalNumericIndexString ( _argument_ )</h1>
-      <p>The abstract operation CanonicalNumericIndexString returns _argument_ converted to a Number value if it is a String representation of a Number that would be produced by ToString, or the string *"-0"*. Otherwise, it returns *undefined*. This abstract operation functions as follows:</p>
+      <p>The abstract operation CanonicalNumericIndexString takes argument _argument_. It returns _argument_ converted to a Number value if it is a String representation of a Number that would be produced by ToString, or the string *"-0"*. Otherwise, it returns *undefined*. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation CanonicalNumericIndexString takes argument _argument_ (a String). It returns _argument_ converted to a Number value if it is a String representation of a Number that would be produced by ToString, or the string *"-0"*. Otherwise, it returns *undefined*. It performs the following steps:</p>
```

spec.html
```diff
@@ -5448,7 +5442,7 @@ <h1>IsExtensible ( _O_ )</h1>
 
     <emu-clause id="sec-isinteger" aoid="IsInteger">
       <h1>IsInteger ( _argument_ )</h1>
-      <p>The abstract operation IsInteger determines if _argument_ is a finite integer Number value.</p>
+      <p>The abstract operation IsInteger takes argument _argument_. It determines if _argument_ is a finite integer Number value. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation IsInteger takes argument _argument_ (an ECMAScript language value). It determines if _argument_ is a finite integer Number value. It performs the following steps:</p>
```

spec.html
bakkot: It's easy to add them in this PR. I actually had them all in originally and then decided to take them all out instead, so it should just be a matter of reverting the second commit.

spec.html
bakkot: > Is that an example of the inconsistency you don't want to handle in this PR?

Yes. I agree those would also be good to address, I just don't want to consider them in scope for this.

bakkot:
@ljharb

> I want to make sure that these kinds of tweaks are something you consider in scope for this PR (ie, adding correct parentheticals on each argument to indicate their type)

I would prefer to consider those out of scope for this PR, even though I also would like to have those added at some point. My intent was to have this PR be a very mechanical rewording of the existing prose, rather than introducing any new information.

That said, if you notice any obvious ones as you're reading through and you want to note them as you go rather than someone having to re-discover them all later, feel free to leave comments like the ones above and just mark them as resolved, and I'll collect them into a followup PR once this lands.

spec.html
bakkot: That said, they're a lot of noise, so I will maybe wait to push that commit so the more substantive preambles can be reviewed first.

michaelficarra:
I would prefer "It performs the following steps *when called*:" to "It performs the following steps:".

jmdyck:
It seems to me that the "performs the following steps" wording in operation/function preambles is superfluous. (Note that several preambles don't have it, and people don't appear to be confused about them.) If we just dropped it, would any readers not understand when the following steps were performed?

(Or, put it this way: in a real programming language, would you ever write the comment: "When called, this function executes the following statements"?)

spec.html
```diff
@@ -4626,7 +4621,7 @@ <h1>ToNumeric ( _value_ )</h1>
 
     <emu-clause id="sec-tonumber" aoid="ToNumber">
       <h1>ToNumber ( _argument_ )</h1>
-      <p>The abstract operation ToNumber converts _argument_ to a value of type Number according to <emu-xref href="#table-11"></emu-xref>:</p>
+      <p>The abstract operation ToNumber takes argument _argument_. It converts _argument_ to a value of type Number according to <emu-xref href="#table-11"></emu-xref>:</p>
```
michaelficarra: How do you feel about this kind of wording for operations that take a single effectively-unnamed argument?

```suggestion
      <p>The abstract operation ToNumber takes one argument. It converts its argument to a value of type Number according to <emu-xref href="#table-11"></emu-xref>:</p>
```

bakkot: Would does that apply to, say, IsExtensible?

> `
The abstract operation IsExtensible takes argument _O_ (an Object) and returns a completion record which, if its [[Type]] is ~normal~, has a [[Value]] which is a Boolean. It is used to determine whether additional properties can be added to _O_. It performs the following steps:
`

or IsAccessorDescriptor?

>`The abstract operation IsAccessorDescriptor takes argument _Desc_ (a Property Descriptor or *undefined*). It performs the following steps:`

I think I would mildly prefer not to do this just on consistency grounds; it's not obvious to me why we'd omit the name of the argument when there's exactly one but not when there's exactly two, especially in cases where there's no description beyond "it performs the following steps".

spec.html
```diff
@@ -5490,7 +5483,7 @@ <h1>IsRegExp ( _argument_ )</h1>
 
     <emu-clause id="sec-isstringprefix" aoid="IsStringPrefix">
       <h1>IsStringPrefix ( _p_, _q_ )</h1>
-      <p>The abstract operation IsStringPrefix determines if String _p_ is a prefix of String _q_.</p>
+      <p>The abstract operation IsStringPrefix takes arguments _p_ (a String) and _q_ (a String). It determines if _p_ is a prefix of _q_ It performs the following steps:</p>
```
michaelficarra: We should look into giving aliases like these more meaningful names. This is not a generic enough operation to warrant completely meaningless names. @tc39/ecma262-editors What do you think?

spec.html
michaelficarra: I think they're fine.

> `The abstract operation IsExtensible takes one argument (an Object) and returns a completion record which, if its [[Type]] is ~normal~, has a [[Value]] which is a Boolean. It is used to determine whether additional properties can be added to its argument. It performs the following steps:`

> `The abstract operation IsAccessorDescriptor takes one argument (a Property Descriptor or *undefined*). It performs the following steps:`

> it's not obvious to me why we'd omit the name of the argument when there's exactly one but not when there's exactly two

We do this all the time in English. When there's no ambiguity, we don't bother to call out things by name. We just use "the" or "it", etc.

bakkot: > When there's no ambiguity, we don't bother to call out things by name. We just use "the" or "it", etc.

My point is that in a lot of cases we _just_ name the arguments and don't later refer to any of them, even as "it".  And in those cases it seems weird to avoid naming them when there's one but not when there's two.

spec.html
ljharb: I think that we should apply that sensibility to new abstract ops, but I don't think the churn is often likely to be worth it.

spec.html
michaelficarra: Oh I'm fine with not listing them if we don't otherwise refer to them. I assumed every argument would be referred to at least once to describe its type.

bakkot: Many preambles do not currently describe the types of the arguments.

syg: I prefer naming all parameters, regardless of the operation's arity.

spec.html
```diff
@@ -4986,7 +4981,7 @@ <h1>ToUint8Clamp ( _argument_ )</h1>
 
     <emu-clause id="sec-tobigint" aoid="ToBigInt">
       <h1>ToBigInt ( _argument_ )</h1>
-      <p>The abstract operation ToBigInt converts its argument _argument_ to a BigInt value, or throws if an implicit conversion from Number would be required.</p>
+      <p>The abstract operation ToBigInt takes argument _argument_. It converts its argument _argument_ to a BigInt value, or throws if an implicit conversion from Number would be required. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation ToBigInt takes argument _argument_. It converts _argument_ to a BigInt value, or throws if an implicit conversion from Number would be required. It performs the following steps:</p>
```

spec.html
```diff
@@ -5459,7 +5453,7 @@ <h1>IsInteger ( _argument_ )</h1>
 
     <emu-clause id="sec-isnonnegativeinteger" aoid="IsNonNegativeInteger">
       <h1>IsNonNegativeInteger ( _argument_ )</h1>
-      <p>The abstract operation IsNonNegativeInteger determines if _argument_ is non-negative integer Number value.</p>
+      <p>The abstract operation IsNonNegativeInteger takes argument _argument_. It determines if _argument_ is non-negative integer Number value. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation IsNonNegativeInteger takes argument _argument_. It determines if _argument_ is a non-negative integral Number value. It performs the following steps:</p>
```

spec.html
```suggestion
      <p>The abstract operation IsStringPrefix takes arguments _p_ (a String) and _q_ (a String). It determines if _p_ is a prefix of _q_. It performs the following steps:</p>
```

spec.html
```diff
@@ -5862,7 +5855,7 @@ <h1>TestIntegrityLevel ( _O_, _level_ )</h1>
 
     <emu-clause id="sec-createarrayfromlist" aoid="CreateArrayFromList">
       <h1>CreateArrayFromList ( _elements_ )</h1>
-      <p>The abstract operation CreateArrayFromList is used to create an Array object whose elements are provided by a List. This abstract operation performs the following steps:</p>
+      <p>The abstract operation CreateArrayFromList takes argument _elements_. It is used to create an Array object whose elements are provided by a List. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation CreateArrayFromList takes argument _elements_ (a List). It is used to create an Array object whose elements are provided by _elements_. It performs the following steps:</p>
```

spec.html
```diff
@@ -5876,7 +5869,7 @@ <h1>CreateArrayFromList ( _elements_ )</h1>
 
     <emu-clause id="sec-lengthofarraylike" aoid="LengthOfArrayLike">
       <h1>LengthOfArrayLike ( _obj_ )</h1>
-      <p>The abstract operation LengthOfArrayLike returns the value of the *"length"* property of an array-like object.</p>
+      <p>The abstract operation LengthOfArrayLike takes argument _obj_. It returns the value of the *"length"* property of an array-like object. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation LengthOfArrayLike takes argument _obj_ (an array-like object). It returns the value of the *"length"* property of _obj_. It performs the following steps:</p>
```

bakkot: There's no actual requirement that _obj_ be array-like, contrary to what the existing prose says, so I don't want to add this one.

michaelficarra: There are abstract ops where we say "(an Object)" but then in the algorithm steps we throw if it's observed not to be an object. What's the difference here?

bakkot: We should get rid of those too. I got rid of several as I noticed them.

michaelficarra: Okay are you going to get rid of them in this PR?

bakkot: Sure.

spec.html
```diff
@@ -5892,7 +5885,7 @@ <h1>LengthOfArrayLike ( _obj_ )</h1>
 
     <emu-clause id="sec-createlistfromarraylike" aoid="CreateListFromArrayLike">
       <h1>CreateListFromArrayLike ( _obj_ [ , _elementTypes_ ] )</h1>
-      <p>The abstract operation CreateListFromArrayLike is used to create a List value whose elements are provided by the indexed properties of an array-like object, _obj_. The optional argument _elementTypes_ is a List containing the names of ECMAScript Language Types that are allowed for element values of the List that is created. This abstract operation performs the following steps:</p>
+      <p>The abstract operation CreateListFromArrayLike takes argument _obj_ (an Object) and optional argument _elementTypes_ (a List of names of ECMAScript Language Types). It is used to create a List value whose elements are provided by the indexed properties of _obj_. _elementTypes_ contains the names of ECMAScript Language Types that are allowed for element values of the List that is created. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation CreateListFromArrayLike takes argument _obj_ and optional argument _elementTypes_ (a List of names of ECMAScript Language Types). It is used to create a List value whose elements are provided by the indexed properties of _obj_. _elementTypes_ contains the names of ECMAScript Language Types that are allowed for element values of the List that is created. It performs the following steps:</p>
```

spec.html
```diff
@@ -5923,7 +5916,7 @@ <h1>Invoke ( _V_, _P_ [ , _argumentsList_ ] )</h1>
 
     <emu-clause id="sec-ordinaryhasinstance" aoid="OrdinaryHasInstance">
       <h1>OrdinaryHasInstance ( _C_, _O_ )</h1>
-      <p>The abstract operation OrdinaryHasInstance implements the default algorithm for determining if an object _O_ inherits from the instance object inheritance path provided by constructor _C_. This abstract operation performs the following steps:</p>
+      <p>The abstract operation OrdinaryHasInstance takes arguments _C_ (an ECMAScript language value) and _O_ (an Object). It implements the default algorithm for determining if _O_ inherits from the instance object inheritance path provided by _C_. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation OrdinaryHasInstance takes arguments _C_ (an ECMAScript language value) and _O_. It implements the default algorithm for determining if _O_ inherits from the instance object inheritance path provided by _C_. It performs the following steps:</p>
```

spec.html
```diff
@@ -8152,7 +8129,7 @@ <h1>OrdinaryOwnPropertyKeys ( _O_ )</h1>
 
     <emu-clause id="sec-ordinaryobjectcreate" aoid="OrdinaryObjectCreate" oldids="sec-objectcreate">
       <h1>OrdinaryObjectCreate ( _proto_ [ , _additionalInternalSlotsList_ ] )</h1>
-      <p>The abstract operation OrdinaryObjectCreate with argument _proto_ (an object or *null*) is used to specify the runtime creation of new ordinary objects. The optional argument _additionalInternalSlotsList_ is a List of the names of additional internal slots that must be defined as part of the object, beyond [[Prototype]] and [[Extensible]]. If the list is not provided, a new empty List is used. This abstract operation performs the following steps:</p>
+      <p>The abstract operation OrdinaryObjectCreate takes argument _proto_ (an Object or *null*) and optional argument _additionalInternalSlotsList_ (a List of names of internal slots). It is used to specify the runtime creation of new ordinary objects. _additionalInternalSlotsList_ contains the names of additional internal slots that must be defined as part of the object, beyond [[Prototype]] and [[Extensible]]. If the list is not provided, a new empty List is used. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation OrdinaryObjectCreate takes argument _proto_ (an Object or *null*) and optional argument _additionalInternalSlotsList_ (a List of names of internal slots). It is used to specify the runtime creation of new ordinary objects. _additionalInternalSlotsList_ contains the names of additional internal slots that must be defined as part of the object, beyond [[Prototype]] and [[Extensible]]. If _additionalInternalSlotsList_ is not provided, a new empty List is used. It performs the following steps:</p>
```

spec.html
```diff
@@ -8168,7 +8145,7 @@ <h1>OrdinaryObjectCreate ( _proto_ [ , _additionalInternalSlotsList_ ] )</h1>
...
-      <p>The abstract operation OrdinaryCreateFromConstructor creates an ordinary object whose [[Prototype]] value is retrieved from a constructor's *"prototype"* property, if it exists. Otherwise the intrinsic named by _intrinsicDefaultProto_ is used for [[Prototype]]. The optional _internalSlotsList_ is a List of the names of additional internal slots that must be defined as part of the object. If the list is not provided, a new empty List is used. This abstract operation performs the following steps:</p>
+      <p>The abstract operation OrdinaryCreateFromConstructor takes arguments _constructor_ and _intrinsicDefaultProto_ and optional argument _internalSlotsList_ (a List of names of internal slots). It creates an ordinary object whose [[Prototype]] value is retrieved from a constructor's *"prototype"* property, if it exists. Otherwise the intrinsic named by _intrinsicDefaultProto_ is used for [[Prototype]]. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. If the list is not provided, a new empty List is used. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation OrdinaryCreateFromConstructor takes arguments _constructor_ and _intrinsicDefaultProto_ and optional argument _internalSlotsList_ (a List of names of internal slots). It creates an ordinary object whose [[Prototype]] value is retrieved from a constructor's *"prototype"* property, if it exists. Otherwise the intrinsic named by _intrinsicDefaultProto_ is used for [[Prototype]]. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. If the _internalSlotsList_ is not provided, a new empty List is used. It performs the following steps:</p>
```

spec.html
```diff
@@ -10364,15 +10334,15 @@ <h1>Static Semantics: UTF16Encoding ( _cp_ )</h1>
 
     <emu-clause id="sec-utf16encode" aoid="UTF16Encode">
       <h1>Static Semantics: UTF16Encode ( _text_ )</h1>
-      <p>This abstract operation converts _text_, a sequence of Unicode code points, into a String value, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>.</p>
+      <p>The abstract operation UTF16Encode takes argument _text_ (a sequence of unicode code points). It converts _text_ into a String value, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>. It performs the following steps:</p>
```
```suggestion
      <p>The abstract operation UTF16Encode takes argument _text_ (a sequence of Unicode code points). It converts _text_ into a String value, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>. It performs the following steps:</p>
```

spec.html
```diff
@@ -22724,7 +22684,7 @@ <h1>Static Semantics: ImportEntries</h1>
 
       <emu-clause id="sec-importedlocalnames" aoid="ImportedLocalNames">
         <h1>Static Semantics: ImportedLocalNames ( _importEntries_ )</h1>
-        <p>The abstract operation ImportedLocalNames with argument _importEntries_ creates a List of all of the local name bindings defined by a List of ImportEntry Records (see <emu-xref href="#table-39"></emu-xref>). ImportedLocalNames performs the following steps:</p>
+        <p>The abstract operation ImportedLocalNames takes argument _importEntries_. It creates a List of all of the local name bindings defined by a List of ImportEntry Records (see <emu-xref href="#table-39"></emu-xref>). It performs the following steps:</p>
```
```suggestion
        <p>The abstract operation ImportedLocalNames takes argument _importEntries_ (a List of ImportEntry Records (see <emu-xref href="#table-39"></emu-xref>)). It creates a List of all of the local name bindings defined by _importEntries_. It performs the following steps:</p>
```

spec.html
```diff
@@ -23181,10 +23136,8 @@ <h1>Evaluate ( ) Concrete Method</h1>
 
           <emu-clause id="sec-innermoduleevaluation" aoid="InnerModuleEvaluation">
             <h1>InnerModuleEvaluation ( _module_, _stack_, _index_ )</h1>
+            <p>The abstract operation InnerModuleEvaluation takes arguments _module_, _stack_, and _index_. It is used by Evaluate to perform the actual evaluation process for the Source Text Module Record _module_, as well as recursively on all other modules in the dependency graph. The _stack_ and _index_ parameters, as well as _module_'s [[DFSIndex]] and [[DFSAncestorIndex]] fields, are used the same way as in InnerModuleLinking. It performs the following steps:</p>
```
```suggestion
            <p>The abstract operation InnerModuleEvaluation takes arguments _module_ (a Source Text Module Record), _stack_, and _index_. It is used by Evaluate to perform the actual evaluation process for _module_, as well as recursively on all other modules in the dependency graph. The _stack_ and _index_ parameters, as well as _module_'s [[DFSIndex]] and [[DFSAncestorIndex]] fields, are used the same way as in InnerModuleLinking. It performs the following steps:</p>
```

spec.html
```diff
@@ -23759,7 +23712,7 @@ <h1>Source Text Module Records</h1>
 
         <emu-clause id="sec-parsemodule" aoid="ParseModule">
           <h1>ParseModule ( _sourceText_, _realm_, _hostDefined_ )</h1>
-          <p>The abstract operation ParseModule with arguments _sourceText_, _realm_, and _hostDefined_ creates a Source Text Module Record based upon the result of parsing _sourceText_ as a |Module|. ParseModule performs the following steps:</p>
+          <p>The abstract operation ParseModule takes arguments _sourceText_, _realm_, and _hostDefined_. It creates a Source Text Module Record based upon the result of parsing _sourceText_ as a |Module|. It performs the following steps:</p>
```
```suggestion
          <p>The abstract operation ParseModule takes arguments _sourceText_ (ECMAScript source text), _realm_, and _hostDefined_. It creates a Source Text Module Record based upon the result of parsing _sourceText_ as a |Module|. It performs the following steps:</p>
```

spec.html
```diff
@@ -25136,7 +25080,7 @@ <h2>Runtime Semantics</h2>
 
         <emu-clause id="sec-encode" aoid="Encode">
           <h1>Runtime Semantics: Encode ( _string_, _unescapedSet_ )</h1>
-          <p>The encoding and escaping process is described by the abstract operation Encode taking two String arguments _string_ and _unescapedSet_.</p>
+          <p>The abstract operation Encode takes arguments _string_ (a String) and _unescapedSet_ (a String). It describes the encoding and escaping process. It performs the following steps:</p>
```
```suggestion
          <p>The abstract operation Encode takes arguments _string_ (a String) and _unescapedSet_ (a String). It performs URI encoding and escaping. It performs the following steps:</p>
```

spec.html
```diff
@@ -25162,7 +25106,7 @@ <h1>Runtime Semantics: Encode ( _string_, _unescapedSet_ )</h1>
 
         <emu-clause id="sec-decode" aoid="Decode">
           <h1>Runtime Semantics: Decode ( _string_, _reservedSet_ )</h1>
-          <p>The unescaping and decoding process is described by the abstract operation Decode taking two String arguments _string_ and _reservedSet_.</p>
+          <p>The abstract operation Decode takes arguments _string_ (a String) and _reservedSet_ (a String). It describes the unescaping and decoding process. It performs the following steps:</p>
```
```suggestion
          <p>The abstract operation Decode takes arguments _string_ (a String) and _reservedSet_ (a String). It performs URI unescaping and decoding. It performs the following steps:</p>
```

spec.html
```diff
@@ -25857,7 +25800,7 @@ <h1>Object.getOwnPropertySymbols ( _O_ )</h1>
 
         <emu-clause id="sec-getownpropertykeys" aoid="GetOwnPropertyKeys">
           <h1>Runtime Semantics: GetOwnPropertyKeys ( _O_, _type_ )</h1>
-          <p>The abstract operation GetOwnPropertyKeys is called with arguments _O_ and _type_ where _O_ is an Object and _type_ is one of the ECMAScript specification types String or Symbol. The following steps are taken:</p>
+          <p>The abstract operation GetOwnPropertyKeys takes arguments _O_ and _type_ (either String or Symbol). It performs the following steps:</p>
```
michaelficarra: Elsewhere we replaced the union of String and Symbol with property key. Should we do that everywhere?

spec.html
```diff
@@ -26131,7 +26074,7 @@ <h1>Function ( _p1_, _p2_, &hellip; , _pn_, _body_ )</h1>
 
         <emu-clause id="sec-createdynamicfunction" aoid="CreateDynamicFunction">
           <h1>Runtime Semantics: CreateDynamicFunction ( _constructor_, _newTarget_, _kind_, _args_ )</h1>
-          <p>The abstract operation CreateDynamicFunction is called with arguments _constructor_, _newTarget_, _kind_, and _args_. _constructor_ is the constructor function that is performing this action, _newTarget_ is the constructor that `new` was initially applied to, _kind_ is either ~normal~, ~generator~, ~async~, or ~asyncGenerator~, and _args_ is a List containing the actual argument values that were passed to _constructor_. The following steps are taken:</p>
+          <p>The abstract operation CreateDynamicFunction takes arguments _constructor_ (a constructor), _newTarget_ (a constructor), _kind_ (either ~normal~, ~generator~, ~async~, or ~asyncGenerator~), and _args_ (a List of ECMAScript language values). _constructor_ is the constructor function that is performing this action. _newTarget_ is the constructor that `new` was initially applied to. _args_ is a List containing the actual argument values that were passed to _constructor_. It performs the following steps:</p>
```
```suggestion
          <p>The abstract operation CreateDynamicFunction takes arguments _constructor_ (a constructor), _newTarget_ (a constructor), _kind_ (either ~normal~, ~generator~, ~async~, or ~asyncGenerator~), and _args_ (a List of ECMAScript language values). _constructor_ is the constructor function that is performing this action. _newTarget_ is the constructor that `new` was initially applied to. _args_ is the argument values that were passed to _constructor_. It performs the following steps:</p>
```

We describe its structure in the first sentence. This sentence should just describe what it represents.

spec.html
```diff
@@ -31468,7 +31397,7 @@ <h1>Term</h1>
 
         <emu-clause id="sec-runtime-semantics-repeatmatcher-abstract-operation" aoid="RepeatMatcher">
           <h1>Runtime Semantics: RepeatMatcher ( _m_, _min_, _max_, _greedy_, _x_, _c_, _parenIndex_, _parenCount_ )</h1>
-          <p>The abstract operation RepeatMatcher takes eight parameters, a Matcher _m_, an integer _min_, an integer (or &infin;) _max_, a Boolean _greedy_, a State _x_, a Continuation _c_, an integer _parenIndex_, and an integer _parenCount_, and performs the following steps:</p>
+          <p>The abstract operation RepeatMatcher takes arguments _m_ (a Matcher), _min_ (an integer), _max_ (an integer or a &infin;), _greedy_ (a Boolean), _x_ (a State), _c_ (a Continuation), _parenIndex_ (an integer), and _parenCount_ (an integer). It performs the following steps:</p>
```
```suggestion
          <p>The abstract operation RepeatMatcher takes arguments _m_ (a Matcher), _min_ (an integer), _max_ (an integer or &infin;), _greedy_ (a Boolean), _x_ (a State), _c_ (a Continuation), _parenIndex_ (an integer), and _parenCount_ (an integer). It performs the following steps:</p>
```

spec.html
bakkot: Note that this isn't _a_ String or Symbol. It's literally String or Symbol. This parameter takes the _name_ of a type. (Open to suggestions for how to word this better; I guess I could put it back closer to what it was before.)

michaelficarra: Oh, should it be `~string~ or ~symbol~`?

spec.html
```diff
@@ -35799,7 +35715,7 @@ <h1>Map ( [ _iterable_ ] )</h1>
 
       <emu-clause id="sec-add-entries-from-iterable" aoid="AddEntriesFromIterable">
         <h1>AddEntriesFromIterable ( _target_, _iterable_, _adder_ )</h1>
-        <p>The abstract operation AddEntriesFromIterable accepts a _target_ object, an _iterable_ of entries, and an _adder_ function to be invoked, with _target_ as the receiver.</p>
+        <p>The abstract operation AddEntriesFromIterable takes arguments _target_, _iterable_, and _adder_ (a function object). _adder_ will be invoked, with _target_ as the receiver It performs the following steps:</p>
```
```suggestion
        <p>The abstract operation AddEntriesFromIterable takes arguments _target_, _iterable_, and _adder_ (a function object). _adder_ will be invoked, with _target_ as the receiver. It performs the following steps:</p>
```

spec.html
bakkot: Then the implementation would have to change. Currently this parameter is used as `If Type(_nextKey_) is _type_`.

spec.html
```diff
@@ -37777,7 +37689,7 @@ <h1>NotifyWaiter ( _WL_, _W_ )</h1>
 
       <emu-clause id="sec-atomicreadmodifywrite" aoid="AtomicReadModifyWrite">
         <h1>AtomicReadModifyWrite ( _typedArray_, _index_, _value_, _op_ )</h1>
-        <p>The abstract operation AtomicReadModifyWrite takes four arguments, _typedArray_, _index_, _value_, and a pure combining operation _op_. The pure combining operation _op_ takes two List of byte values arguments and returns a List of byte values. The operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps:</p>
+        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a pure combining operation). It atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps:</p>
```
```suggestion
        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a pure combining operation). _op_ takes two List of byte values arguments and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps:</p>
```

spec.html
```diff
@@ -38200,8 +38111,7 @@ <h1>Runtime Semantics: SerializeJSONProperty ( _state_, _key_, _holder_ )</h1>
 
       <emu-clause id="sec-quotejsonstring" aoid="QuoteJSONString">
         <h1>Runtime Semantics: QuoteJSONString ( _value_ )</h1>
-        <p>The abstract operation QuoteJSONString with argument _value_ wraps a String value in QUOTATION MARK code units and escapes certain other code units within it.</p>
-        <p>This operation interprets a String value as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>.</p>
+        <p>The abstract operation QuoteJSONString takes argument _value_. It wraps a String value in QUOTATION MARK code units and escapes certain other code units within it. This operation interprets a String value as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>. It performs the following steps:</p>
```
```suggestion
        <p>The abstract operation QuoteJSONString takes argument _value_. It wraps _value_ in 0x0022 (QUOTATION MARK) code units and escapes certain other code units within it. This operation interprets _value_ as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>. It performs the following steps:</p>
```

spec.html
```diff
@@ -39788,7 +39687,7 @@ <h1>FulfillPromise ( _promise_, _value_ )</h1>
 
       <emu-clause id="sec-newpromisecapability" aoid="NewPromiseCapability">
         <h1>NewPromiseCapability ( _C_ )</h1>
-        <p>The abstract operation NewPromiseCapability takes a constructor function, and attempts to use that constructor function in the fashion of the built-in `Promise` constructor to create a Promise object and extract its resolve and reject functions. The promise plus the resolve and reject functions are used to initialize a new PromiseCapability Record which is returned as the value of this abstract operation.</p>
+        <p>The abstract operation NewPromiseCapability takes argument _C_. It attempts to use _C_ as a constructor in the fashion of the built-in `Promise` constructor to create a Promise object and extract its resolve and reject functions. The promise plus the resolve and reject functions are used to initialize a new PromiseCapability Record. It performs the following steps:</p>
```
```suggestion
        <p>The abstract operation NewPromiseCapability takes argument _C_. It attempts to use _C_ as a constructor in the fashion of the built-in `Promise` constructor to create a Promise object and extract its resolve and reject functions. The Promise object plus the `resolve` and `reject` functions are used to initialize a new PromiseCapability Record. It performs the following steps:</p>
```

spec.html
```diff
@@ -40392,7 +40286,7 @@ <h1>Promise.prototype.then ( _onFulfilled_, _onRejected_ )</h1>
 
         <emu-clause id="sec-performpromisethen" aoid="PerformPromiseThen">
           <h1>PerformPromiseThen ( _promise_, _onFulfilled_, _onRejected_ [ , _resultCapability_ ] )</h1>
-          <p>The abstract operation PerformPromiseThen performs the &ldquo;then&rdquo; operation on _promise_ using _onFulfilled_ and _onRejected_ as its settlement actions. If _resultCapability_ is passed, the result is stored by updating _resultCapability_'s promise. (If it is not passed, then PerformPromiseThen is being called by a specification-internal operation where the result does not matter.)</p>
+          <p>The abstract operation PerformPromiseThen takes arguments _promise_, _onFulfilled_, and _onRejected_ and optional argument _resultCapability_. It performs the &ldquo;then&rdquo; operation on _promise_ using _onFulfilled_ and _onRejected_ as its settlement actions. If _resultCapability_ is passed, the result is stored by updating _resultCapability_'s promise. (If it is not passed, then PerformPromiseThen is being called by a specification-internal operation where the result does not matter.) It performs the following steps:</p>
```
```suggestion
          <p>The abstract operation PerformPromiseThen takes arguments _promise_, _onFulfilled_, and _onRejected_ and optional argument _resultCapability_. It performs the &ldquo;then&rdquo; operation on _promise_ using _onFulfilled_ and _onRejected_ as its settlement actions. If _resultCapability_ is passed, the result is stored by updating _resultCapability_'s promise. If it is not passed, then PerformPromiseThen is being called by a specification-internal operation where the result does not matter. It performs the following steps:</p>
```

spec.html
michaelficarra: Easy enough: https://github.com/tc39/ecma262/pull/1918

spec.html
bakkot: Added the missing "a", but leaving this as "integer" rather than "integral" to match other operations. Feel free to submit a followup PR fixing all of them.

spec.html
```diff
@@ -37689,7 +37689,7 @@ <h1>NotifyWaiter ( _WL_, _W_ )</h1>
 
       <emu-clause id="sec-atomicreadmodifywrite" aoid="AtomicReadModifyWrite">
         <h1>AtomicReadModifyWrite ( _typedArray_, _index_, _value_, _op_ )</h1>
-        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a pure combining operation). It atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps:</p>
+        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a pure combining operation). _op_ takes two List of byte values arguments and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps:</p>
```
```suggestion
        <p>The abstract operation AtomicReadModifyWrite takes arguments _typedArray_, _index_, _value_, and _op_ (a pure combining operation). _op_ takes two arguments, each a List of byte values, and returns a List of byte values. This operation atomically loads a value, combines it with another value, and stores the result of the combination. It returns the loaded value. It performs the following steps:</p>
```

michaelficarra:
Needs to resolve conflicts after https://github.com/tc39/ecma262/pull/1918.

bakkot:
> i think i'd prefer to always include a preamble

Done.

> I would prefer "It performs the following steps when called:" to "It performs the following steps:".

Done.

> Needs to resolve conflicts after #1918.

Done.

spec.html
```diff
@@ -4083,11 +4082,10 @@ <h1>Await Fulfilled Functions</h1>
 
         <emu-clause id="await-rejected">
           <h1>Await Rejected Functions</h1>
-
           <p>An Await rejected function is an anonymous built-in function that is used as part of the Await specification device to deliver the promise rejection reason to the caller as an abrupt throw completion. Each Await rejected function has an [[AsyncContext]] internal slot.</p>
-
           <p>When an Await rejected function is called with argument _reason_, the following steps are taken:</p>
 
+
```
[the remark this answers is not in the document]
bakkot: The ones which got changed here were an artifact of the process I was using. Whitespace is pretty inconsistent in general, so I figured it would be fine, but I can back those changes out if you'd prefer. I'd be happy to make it consistent in a separate PR, though I think would want to wait for the ability to enforce it (cf https://github.com/bterlson/ecmarkup/issues/173).

bakkot:
> ensure all of those types are matched with either an explicit list of assertions, or, with a notational convention that implicitly asserts this for all AOs

There is not yet agreement on exactly this one of these two possible alternatives, just that we would like it to be consistent somehow. There are other possible ways that could happen - for example, it would also be consistent to just leave out the assertions entirely.

spec.html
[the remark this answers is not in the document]
ljharb: I'm fine waiting on it; i'd just prefer to end up never having two blank lines in a row :-)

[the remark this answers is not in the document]
bakkot: Added to my [list](https://github.com/bterlson/ecmarkup/issues/173) of desiderata for linting the spec.

jmdyck:
> > ensure all of those types are matched with either an explicit list of assertions, or, with a notational convention that implicitly asserts this for all AOs
> 
> [...] There are other possible ways that could happen - for example, it would also be consistent to just leave out the assertions entirely.

Isn't that the same as @ljharb's second alternative? That is: don't `Assert` parameter types in the `<emu-alg>`, just have parenthesized parameter types in the preamble, and then say in 5.2.1 Abstract Operations that the latter have the same effect as Asserts. (If that *isn't* the suggestion, please clarify.)

bakkot:
> If that isn't the suggestion, please clarify.

The suggestion (well, not a suggestion _per se_, just a statement of possibility) is to omit the modification to 5.2.1, or to modify it in such a way that it does not make mention of `Assert` in explaining the interpretation of these type parameters.

PR #1916, opened 2020-03-25
Editorial: added prefix ! when calling OrdinaryGetOwnProperty

jhnaldo:
I added prefix ! when calling [`OrdinaryGetOwnProperty`](https://tc39.es/ecma262/#sec-ordinarygetownproperty) in [`[[GetOwnProperty]]`](https://tc39.es/ecma262/#sec-arguments-exotic-objects-getownproperty-p) of arguments exotic object.

[`OrdinaryGetOwnProperty`](https://tc39.es/ecma262/#sec-ordinarygetownproperty) is a runtime semantics thus it always return a completion record (Actually, in this case `OrdinaryGetOwnProperty` always return normal completion). In step 2 of [`[[GetOwnProperty]]`](https://tc39.es/ecma262/#sec-arguments-exotic-objects-getownproperty-p) of arguments exotic object, _desc_ points to a completion record. I know there exists a implicit coercion rule for `[[Value]]` of completion record in [5.2.3.1](https://tc39.es/ecma262/#sec-implicit-completion-values) through #1664.

However, in step 6-a, it tries to access `[[Value]]` field of _desc_. It seems to access the `[[Value]]` of completion record but the actually meaning is to access the `[[Value]]` field of a data property in _desc_.`[[Value]]` (simply _desc_.`[[Value]]`.`[[Value]`).

I think it is ambiguous thus I think the prefix `!` should be added for function call of [`OrdinaryGetOwnProperty`](https://tc39.es/ecma262/#sec-ordinarygetownproperty) in [`[[GetOwnProperty]]`](https://tc39.es/ecma262/#sec-arguments-exotic-objects-getownproperty-p) of arguments exotic object.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9101,7 +9101,7 @@
         <p>The [[GetOwnProperty]] internal method of an arguments exotic object when called with a property key _P_ performs the following steps:</p>
         <emu-alg>
           1. Let _args_ be the arguments object.
-          1. Let _desc_ be OrdinaryGetOwnProperty(_args_, _P_).
+          1. Let _desc_ be ! OrdinaryGetOwnProperty(_args_, _P_).
           1. If _desc_ is *undefined*, return _desc_.
           1. Let _map_ be _args_.[[ParameterMap]].
           1. Let _isMapped_ be ! HasOwnProperty(_map_, _P_).


jmdyck:
Strictly speaking, step 6.a is "a context that does not explicitly require a complete Completion Record value", so the 5.2.3.1 rule applies, and you're obliged to interpret `_desc_` as a reference to the completion record's `[[Value]]` field (i.e., the property descriptor), and `_desc_.[[Value]]` as a reference to the property descriptor's `[[Value]]` field.

So technically, there isn't an ambiguity, but the disambiguation hinges on an awfully fine point, so it could lead to reader confusion. And inserting a '!' makes that go away.

It looks like the same could be said of the calls to `OrdinaryGetOwnProperty` in:
- [`Array's [[DefineOwnProperty]]`](https://tc39.es/ecma262/#sec-array-exotic-objects-defineownproperty-p-desc)
- [`ArraySetLength`](https://tc39.es/ecma262/#sec-arraysetlength)

(Mind you, issue #1796 would presumably change `OrdinaryGetOwnProperty` to not return a completion record, and therefore remove the '!' from any call to it.)

PR #1919, opened 2020-03-26
Editorial: sunk note for null

wentout:
Summary of facts caused this PR.
Most of modern javascript engines can give us the following checking capabilities:

1. object has no inherited ancestor via:

```javascript
// if it returns null, then there is no inheritance
Object.getPrototypeOf(object_we_are_checking);
```

2. `Object.prototype` is `null`:

```javascript
Object.getPrototypeOf(Object.prototype); // null
```
3. `typeof` operator applied to `null` value returns `"object"`:

```javascript
typeof null === 'object'; // true
```

Combining the facts someone can get a conclusion all of them are connected.

But, the events that triggered these additives as standard were random and not related. Despite the fact that someone may find it useful to have such a connection, its careless use can lead to the creation of [sunk cost](https://en.wikipedia.org/wiki/Sunk_cost). This PR is an attempt to leave a warning and continues the discussion begun in #1913.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -209,6 +209,9 @@
       <emu-note>
         <p>A primitive value is a datum that is represented directly at the lowest level of the language implementation.</p>
       </emu-note>
+      <emu-note>
+        <p>The value of Null type has additional usage besides being one of primitive values: <emu-xref href="#sec-ecmascript-language-types-null-type"></emu-xref></p>
+      </emu-note>
     </emu-clause>
 
     <emu-clause id="sec-terms-and-definitions-object">
@@ -943,6 +946,9 @@
     <emu-clause id="sec-ecmascript-language-types-null-type">
       <h1>The Null Type</h1>
       <p>The Null type has exactly one value, called *null*.</p>
+      <emu-note>
+        <p>Value *null* used as a root of prototype chain if there is no inheritance. Keep in mind *typeof* operator also returns "object" for *null* value: <emu-xref href="#sec-typeof-operator"></emu-xref>. A careless combination of these facts can be dangerous.</p>
+      </emu-note>
     </emu-clause>
 
     <emu-clause id="sec-ecmascript-language-types-boolean-type">
@@ -14438,6 +14444,11 @@
           </table>
         </emu-table>
       </emu-clause>
+
+      <emu-note>
+        <p>Backing to the history the result of typeof *null* is "object" was mostly used for backward compatibility requirements. Meanwhile other additions allowed to express some parts of inheritance model throgh the value *null*. Today *null* used as a root of prototype chain for objects not being inherited from anything else. And since we maintain connection between *null* and *object*, we caution you against looking for some deep meaning in this behavior. Although it may in fact satisfy the reasons you need, careless combination of these capabilities might cause sunk cost.</p>
+      </emu-note>
+
     </emu-clause>
 
 


spec.html
```diff
@@ -209,6 +209,9 @@ <h1>primitive value</h1>
       <emu-note>
         <p>A primitive value is a datum that is represented directly at the lowest level of the language implementation.</p>
       </emu-note>
+      <emu-note>
+        <p>The value of Null type has additional usage besides being one of primitive values: <emu-xref href="#sec-ecmascript-language-types-null-type"></emu-xref></p>
```
ljharb: "Being a primitive value" isn't a usage, it's what it is, so this link isn't an additional use of that.

spec.html
```diff
@@ -943,6 +946,9 @@ <h1>The Undefined Type</h1>
     <emu-clause id="sec-ecmascript-language-types-null-type">
       <h1>The Null Type</h1>
       <p>The Null type has exactly one value, called *null*.</p>
+      <emu-note>
+        <p>Value *null* used as a root of prototype chain if there is no inheritance. Keep in mind *typeof* operator also returns "object" for *null* value: <emu-xref href="#sec-typeof-operator"></emu-xref>. A careless combination of these facts can be dangerous.</p>
```
ljharb: Words like "careless" and "dangerous" do not belong in a specification, and there's no benefit in restating how `typeof` works, since it's already defined in this document.

spec.html
```diff
@@ -14438,6 +14444,11 @@ <h1>Runtime Semantics: Evaluation</h1>
           </table>
         </emu-table>
       </emu-clause>
+
+      <emu-note>
+        <p>Backing to the history the result of typeof *null* is "object" was mostly used for backward compatibility requirements. Meanwhile other additions allowed to express some parts of inheritance model throgh the value *null*. Today *null* used as a root of prototype chain for objects not being inherited from anything else. And since we maintain connection between *null* and *object*, we caution you against looking for some deep meaning in this behavior. Although it may in fact satisfy the reasons you need, careless combination of these capabilities might cause sunk cost.</p>
```
ljharb: As was [explained](https://github.com/tc39/ecma262/pull/1913#issuecomment-603895260) in #1913, the spec isn't the appropriate place for this kind of information. I'd suggest you look into MDN instead of making this change here.

(separately, i have no idea what you mean by "cause sunk cost")

wentout: I need more data or information between.
If there is an ability to discuss it more...

spec.html
wentout: I thought about it in first. But then thought about a chain between 1.2.3. and leaved it untouched.
I need more understanding, there must be the other way to explain.

spec.html
wentout: Good way is to try to figure out is to get rid of retrospection `typeof null === 'object` is a bad thing.
Because not it causes Catch-22
https://en.wikipedia.org/wiki/Catch-22

And this causes Sunk cost
https://en.wikipedia.org/wiki/Sunk_cost

And as I'm unable to explain this properly right now, it might be better to close both PRs
there should be the other way to explain, it might be spec as a source of truth, but obviously this should be done the other way, not by a note but by something else

wentout: wrong link for Catch-22, the correct is
https://en.wikipedia.org/wiki/Catch-22_(logic)

So, yes, if it causes catch 22 or if it not causes -- it is the thing which is not applicable to the catch itself, because if there is a catch it doesn't matter, therefore we shoud get rid of any opinion about the sort of event caused
`typeof null === 'object'`

event itself is important, but giving opinion it was bad or good causes catch-22
and as I can see from Game theory - 
https://en.wikipedia.org/wiki/Game_theory

it is better to give good opinion if there is a way to give good
and it is better to insist on bad opinion if there is no other way

It is as simple as Nash equilibrium
https://en.wikipedia.org/wiki/Nash_equilibrium

That is it... but it is not technical, but it is how it works

PR #1920, opened 2020-03-27
Normative: Generically forbid 402 method extensions in Forbidden Extensions

rkirsling:
[Forbidden Extensions](https://tc39.es/ecma262/#sec-forbidden-extensions) has a list of methods which are specified in ECMA-402 instead, but a few are missing; this PR rectifies that.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -24737,7 +24737,7 @@
         Neither mapped nor unmapped arguments objects may be created with an own property named *"caller"*.
       </li>
       <li>
-        The behaviour of the following methods must not be extended except as specified in ECMA-402: `Object.prototype.toLocaleString`, `Array.prototype.toLocaleString`, `Number.prototype.toLocaleString`, `Date.prototype.toLocaleDateString`, `Date.prototype.toLocaleString`, `Date.prototype.toLocaleTimeString`, `String.prototype.localeCompare`, %TypedArray%`.prototype.toLocaleString`.
+        The behaviour of built-in methods which are specified in ECMA-402, such as those named `toLocaleString`, must not be extended except as specified in ECMA-402.
       </li>
       <li>
         The RegExp pattern grammars in <emu-xref href="#sec-patterns"></emu-xref> and <emu-xref href="#sec-regular-expressions-patterns"></emu-xref> must not be extended to recognize any of the source characters A-Z or a-z as |IdentityEscape[+U]| when the <sub>[U]</sub> grammar parameter is present.


ljharb:
Given that these are actually forbidden, this seems like a normative change to me, not an editorial one (altho not one I suspect needs consensus).

rkirsling:
My understanding is that the normative part is in each of the respective sections (e.g. [BigInt.prototype.toLocaleString](https://tc39.es/ecma262/#sec-bigint.prototype.tolocalestring)), while this bullet point is just being helpful by collecting the full list in one spot.

bakkot:
Hm, I'd lean towards considering this normative. `BigInt.prototype.toLocaleString` [says](https://tc39.es/ecma262/#sec-bigint.prototype.tolocalestring)

> An ECMAScript implementation that includes the ECMA-402 Internationalization API must implement the BigInt.prototype.toLocaleString method as specified in the ECMA-402 specification. If an ECMAScript implementation does not include the ECMA-402 API the following specification of the toLocaleString method is used.

which reads to me as a less strict requirement than the "The behaviour of the following methods must not be extended except as specified in ECMA-402:" requirement which appears in [Forbidden Extensions](https://tc39.es/ecma262/#sec-forbidden-extensions).

michaelficarra:
If we ask for consensus for this, the consensus we get should be for fast-tracking all 402 methods into 262 forbidden extensions, without requiring further approval from the committee.

rkirsling:
Here's an attempt at a generic phrasing.

spec.html
```diff
@@ -24747,7 +24747,7 @@ <h1>Forbidden Extensions</h1>
         Neither mapped nor unmapped arguments objects may be created with an own property named *"caller"*.
       </li>
       <li>
-        The behaviour of the following methods must not be extended except as specified in ECMA-402: `Object.prototype.toLocaleString`, `Array.prototype.toLocaleString`, `Number.prototype.toLocaleString`, `Date.prototype.toLocaleDateString`, `Date.prototype.toLocaleString`, `Date.prototype.toLocaleTimeString`, `String.prototype.localeCompare`, %TypedArray%`.prototype.toLocaleString`.
+        The behaviour of built-in methods which are specified in ECMA-402, such as `toLocaleString`, must not be extended except as specified in ECMA-402.
```
leobalter: I'm not finding a better solution but I wish the current text could be more explicit in saying it refers to many `toLocaleString`. Currently it gives an impression there is only method with this name.

Perhaps: 

```suggestion
        The behaviour of built-in methods which are specified in ECMA-402, such as those named `toLocaleString`, must not be extended except as specified in ECMA-402.
```

rkirsling: I like it!

PR #1921, opened 2020-03-30
fix auto-linking of LocalTZA abstract op

michaelficarra:
`LocalTZA` is currently listed as having 0 references, and not auto-linked.


michaelficarra:
🎉 Woohoo, made the 2020 cut!

PR #1924, opened 2020-03-31
Editorial: Tweak Array exotic object internal methods

shvaikalesh:
This change replaces prose in normative steps with [`IsDataDescriptor`](https://tc39.es/ecma262/#sec-isdatadescriptor) abstract op, explicitly marks a `NOTE:` and result of `OrdinaryDefineOwnProperty` call.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -8893,7 +8893,8 @@
             1. Return ? ArraySetLength(_A_, _Desc_).
           1. Else if _P_ is an array index, then
             1. Let _oldLenDesc_ be OrdinaryGetOwnProperty(_A_, *"length"*).
-            1. Assert: _oldLenDesc_ will never be *undefined* or an accessor descriptor because Array objects are created with a length data property that cannot be deleted or reconfigured.
+            1. Assert: ! IsDataDescriptor(_oldLenDesc_) is *true*.
+            1. Assert: _oldLenDesc_.[[Configurable]] is *false*.
             1. Let _oldLen_ be _oldLenDesc_.[[Value]].
             1. Assert: IsNonNegativeInteger(_oldLen_) is *true*.
             1. Let _index_ be ! ToUint32(_P_).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -8976,7 +8978,8 @@ <h1>ArraySetLength ( _A_, _Desc_ )</h1>
               1. Perform ! OrdinaryDefineOwnProperty(_A_, *"length"*, _newLenDesc_).
               1. Return *false*.
           1. If _newWritable_ is *false*, then
-            1. Return OrdinaryDefineOwnProperty(_A_, *"length"*, PropertyDescriptor { [[Writable]]: *false* }). This call will always return *true*.
+            1. Let _succeeded_ be ! OrdinaryDefineOwnProperty(_A_, *"length"*, PropertyDescriptor { [[Writable]]: *false* }).
+            1. Assert: _succeeded_ is *true*.
```
michaelficarra: I don't know why we care to assert this. We can just call `OrdinaryDefineOwnProperty` for its effects and be done with it.

ljharb: It returns `false` if it fails, so without the assertion, a failure would be silently allowed and ignored.

michaelficarra: But it won't fail.

ljharb: Sure - that's what the note was meant to convey, and what this assert conveys. It seems important to preserve the intent that this call should never be able to be made to fail in the future, as well.

shvaikalesh: The assert might be taking an extra step, but I find it helpful because descriptors merging logic in [`ValidateAndApplyPropertyDescriptor`](https://tc39.es/ecma262/#sec-validateandapplypropertydescriptor) is a bit hard to follow. There is another assert like this one in Array's [`[[DefineOwnProperty]]`](https://tc39.es/ecma262/#sec-array-exotic-objects-defineownproperty-p-desc).

PR #1927, opened 2020-03-31
Editorial: fix auto-linking of TemplateStrings abstract op

bakkot:
Same problem as https://github.com/tc39/ecma262/pull/1921.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13099,7 +13099,7 @@
         </ul>
       </emu-clause>
 
-      <emu-clause id="sec-static-semantics-templatestrings">
+      <emu-clause id="sec-static-semantics-templatestrings" aoid="TemplateStrings">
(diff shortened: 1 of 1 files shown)


jmdyck:
The difference with PR #1921 is that `LocalTZA` is an (implementation-defined) abstract operation, whereas `TemplateStrings` is a syntax-directed operation. So this case is actually more like PR #1821, which attempts to achieve auto-linking for the SDO `CaseBlockEvaluation`.

As I pointed out there, when an SDO has a single clause that contains all of its definitions, it is tempting to put an `aoid` on that clause, but that approach can't be applied to all SDOs. @syg said he'd rather not add `aoid`s to SDOs.

michaelficarra:
Ah good point @jmdyck.

bakkot:
Closing; should be fixed by #1950.

PR #1928, opened 2020-03-31
Editorial: List module namespace exports as new in the 11th edition

mathiasbynens:
I think this was an oversight, but I might be missing something.


spec.html
```diff
@@ -105,7 +105,7 @@ <h1>Introduction</h1>
...
+  <p>This specification, the 11<sup>th</sup> edition, introduces the `matchAll` method for Strings, to produce an iterator for all match objects generated by a global regular expression; `import()`, a syntax to asynchronously import Modules with a dynamic specifier; `BigInt`, a new number primitive for working with arbitrary precision integers; `Promise.allSettled`, a new Promise combinator that does not short-circuit; `globalThis`, a universal way to access the global `this` value; dedicated `import * as ns from 'module'` syntax for use within modules; increased standardization of `for-in` enumeration order, as well as adding two new syntax features to improve working with “nullish” values (`null` or `undefined`): nullish coalescing, a value selection operator; and optional chaining, a property access and function invocation operator that short-circuits if the value to access/invoke is nullish.</p>
```
```suggestion
  <p>This specification, the 11<sup>th</sup> edition, introduces the `matchAll` method for Strings, to produce an iterator for all match objects generated by a global regular expression; `import()`, a syntax to asynchronously import Modules with a dynamic specifier; `BigInt`, a new number primitive for working with arbitrary precision integers; `Promise.allSettled`, a new Promise combinator that does not short-circuit; `globalThis`, a universal way to access the global `this` value; dedicated `export * as ns from 'module'` syntax for use within modules; increased standardization of `for-in` enumeration order, as well as adding two new syntax features to improve working with “nullish” values (`null` or `undefined`): nullish coalescing, a value selection operator; and optional chaining, a property access and function invocation operator that short-circuits if the value to access/invoke is nullish.</p>
```

mathiasbynens: Good catch! I previously had something about symmetry with `import *` but removed it, which is how this snuck in.

I couldn't click "commit suggestion" so I committed it locally and force-pushed.

PR #1932, opened 2020-04-02
Normative: allow non-BMP characters in capture group names

bakkot:
This is a followup to https://github.com/tc39/ecma262/pull/1869. Per consensus, `/(?<𝒜>.)/` should be legal. But the rules for parsing non-`u` patterns are such that the source text is parsed by treating each half of a surrogate pair as an individual code point. So the rules for `RegExpIdentifierStart` and `RegExpIdentifierPart` need to be tweaked to allow surrogate pairs for that case.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -30890,7 +30890,7 @@ THH:mm:ss.sss
           `.`
           `\` AtomEscape[?U, ?N]
           CharacterClass[?U]
-          `(` GroupSpecifier Disjunction[?U, ?N] `)`
+          `(` GroupSpecifier[?U] Disjunction[?U, ?N] `)`
           `(` `?` `:` Disjunction[?U, ?N] `)`
 
         SyntaxCharacter :: one of
@@ -30903,7 +30903,7 @@ THH:mm:ss.sss
           DecimalEscape
           CharacterClassEscape[?U]
           CharacterEscape[?U]
-          [+N] `k` GroupName
+          [+N] `k` GroupName[?U]
 
         CharacterEscape[U] ::
           ControlEscape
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -31091,13 +31099,25 @@ <h1>Static Semantics: Early Errors</h1>
...
         <ul>
           <li>
-            It is a Syntax Error if the CharacterValue of |RegExpUnicodeEscapeSequence| is none of *"$"*, or *"_"*, or the UTF16Encoding of either &lt;ZWNJ&gt; or &lt;ZWJ&gt;, or the UTF16Encoding of a Unicode code point that would be matched by the |UnicodeIDContinue| lexical grammar production.
+            It is a Syntax Error if the CharacterValue of |RegExpUnicodeEscapeSequence| is not the code point value of *"$"*,  *"_"*, &lt;ZWNJ&gt;, &lt;ZWJ&gt;, or some code point matched by the |UnicodeIDContinue| lexical grammar production.
+          </li>
+        </ul>
+        <emu-grammar>RegExpIdentifierPart[U] :: UnicodeLeadSurrogate UnicodeTrailSurrogate</emu-grammar>
+        <ul>
+          <li>
+            It is a Syntax Error if the result of performing UTF16DecodeSurrogatePair on the two code points matched by UnicodeLeadSurrogate and UnicodeTrailSurrogate respectively is not the code point value of some code point matched by the |UnicodeIDContinue| lexical grammar production.
```
michaelficarra, line 31120: Maybe we could phrase these new early errors like this?

```suggestion
            It is a Syntax Error if the result of performing UTF16DecodeSurrogatePair on the two code points matched by UnicodeLeadSurrogate and UnicodeTrailSurrogate respectively is not matched by the |UnicodeIDContinue| lexical grammar production.
```

PR #1933, opened 2020-04-02
Editorial: delete some SDO cases that are covered by the chain production rule

michaelficarra:
@bakkot You may want to consider linting for these in https://github.com/bterlson/ecmarkup/issues/173.


michaelficarra:
At today's editor call, we decided that we will try to reduce unnecessary SDO definitions, as we do in this PR, relying on the chain rule wherever applicable.

PR #1938, opened 2020-04-06
Editorial: Early return & throw in Number.prototype.to* methods

shvaikalesh:
This change makes handling of non-finite `this` values more concise (by using [`Number::toString`](https://tc39.es/ecma262/#sec-numeric-types-number-tostring)) and moves it earlier in steps, as well as argument range checks. This makes observable steps grouped and order of observable steps easier to read (especially if a runtime uses something like `std::isfinite` as JSC does).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27210,14 +27210,12 @@
           1. Let _x_ be ? thisNumberValue(*this* value).
           1. Let _f_ be ? ToInteger(_fractionDigits_).
           1. Assert: If _fractionDigits_ is *undefined*, then _f_ is 0.
-          1. If _x_ is *NaN*, return the String *"NaN"*.
+          1. If _x_ is not finite, return ! Number::toString(_x_).
+          1. If _f_ &lt; 0 or _f_ &gt; 100, throw a *RangeError* exception.
           1. Let _s_ be the empty String.
           1. If _x_ &lt; 0, then
             1. Set _s_ to *"-"*.
             1. Set _x_ to -_x_.
-          1. If _x_ = *+&infin;*, then
-            1. Return the string-concatenation of _s_ and *"Infinity"*.
-          1. If _f_ &lt; 0 or _f_ &gt; 100, throw a *RangeError* exception.
           1. If _x_ = 0, then
             1. Let _m_ be the String value consisting of _f_ + 1 occurrences of the code unit 0x0030 (DIGIT ZERO).
             1. Let _e_ be 0.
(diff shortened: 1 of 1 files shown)


shvaikalesh:
With this change, it is easier to spot a discrepancy in handling of non-finite `this` values:

```js
0..toExponential(-1) // RangeError
0..toFixed(-1) // RangeError
0..toPrecision(-1) // RangeError

NaN.toExponential(-1) // "NaN"
NaN.toFixed(-1) // RangeError
NaN.toPrecision(-1) // "NaN"
```

Would you consider a normative PR to align it?

We can either a) move range check after finiteness check in `toFixed` or b) move range checks before finiteness checks in both `toExponential` and `toPrecision`.

a) is most certainly web-compatible, and it makes sense to perform all checks on `this` value before arguments.

spec.html
```diff
@@ -27229,7 +27227,7 @@ <h1>Number.prototype.toFixed ( _fractionDigits_ )</h1>
           1. Let _f_ be ? ToInteger(_fractionDigits_).
           1. Assert: If _fractionDigits_ is *undefined*, then _f_ is 0.
           1. If _f_ &lt; 0 or _f_ &gt; 100, throw a *RangeError* exception.
-          1. If _x_ is *NaN*, return the String *"NaN"*.
+          1. If _x_ is not finite, return ! Number::toString(_x_).
```
shvaikalesh, line 27230: `±Infinity` is handled a few steps further:

```
8. If x ≥ 10^21, then
  a. Let m be ! ToString(x).
```

ljharb: if `x` can only be finite or `NaN` here, why is it clearer to use "is not finite", rather than, say, leaving the `is NaN` check but adding `Assert: _x_ is finite` after it?

shvaikalesh: `x` can be anything here, `±Infinity` is handled 2 steps below, making `Number.prototype.toFixed` change observably equivalent.

shvaikalesh: @ljharb I would also appreciate your feedback on https://github.com/tc39/ecma262/pull/1938#issuecomment-610051371.

ljharb: ah, i see, so this is basically "hoisting" the infinities handling up to this line, this seems good.

ljharb:
@shvaikalesh regarding https://github.com/tc39/ecma262/pull/1938#issuecomment-610051371, it seems like you're suggesting either a) making `NaN.toFixed(-1)` return NaN instead of throwing a RangeError, or b) making `NaN.toExponential(-1)` and `NaN.toPrecision(-1)` both throw a RangeError.

The latter seems unlikely to be web compatible, and the former does not seem desirable. Can you elaborate?

shvaikalesh:
> The latter seems unlikely to be web compatible, and the former does not seem desirable. Can you elaborate?

My only argument for the normative change is consistency, but it doesn't justify adding undesirable behavior. While all built-in methods check `this` value before arguments, they never return early for a subset of values.

PR #1944, opened 2020-04-13
Editorial: Clarify notes on async functions in CreateDynamicFunction

### LINKED ISSUE #1701: Note in CreateDynamicFunction unclear
At https://www.ecma-international.org/ecma-262/10.0/index.html#sec-createdynamicfunction is the note

> A prototype property is created for every non-async function created using CreateDynamicFunction to provide for the possibility that the function will be used as a constructor.

Unclear if "non-async functions" is intended to include async generator functions.

shvaikalesh:
Resolves #1701.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -26316,14 +26316,14 @@
               1. Let _prototype_ be OrdinaryObjectCreate(%AsyncGenerator.prototype%).
               1. Perform DefinePropertyOrThrow(_F_, *"prototype"*, PropertyDescriptor { [[Value]]: _prototype_, [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *false* }).
             1. Else if _kind_ is ~normal~, perform MakeConstructor(_F_).
-            1. NOTE: Async functions are not constructible and do not have a [[Construct]] internal method or a *"prototype"* property.
+            1. NOTE: Functions whose _kind_ is ~async~ are not constructible and do not have a [[Construct]] internal method or a *"prototype"* property.
             1. Let _prefix_ be the prefix associated with _kind_ in <emu-xref href="#table-dynamic-function-sourcetext-prefixes"></emu-xref>.
             1. Let _sourceString_ be the string-concatenation of _prefix_, *" anonymous("*, _P_, 0x000A (LINE FEED), *") {"*, _bodyString_, and *"}"*.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -26304,14 +26304,14 @@ <h1>Runtime Semantics: CreateDynamicFunction ( _constructor_, _newTarget_, _kind
               1. Let _prototype_ be OrdinaryObjectCreate(%AsyncGenerator.prototype%).
               1. Perform DefinePropertyOrThrow(_F_, *"prototype"*, PropertyDescriptor { [[Value]]: _prototype_, [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *false* }).
             1. Else if _kind_ is ~normal~, perform MakeConstructor(_F_).
-            1. NOTE: Async functions are not constructable and do not have a [[Construct]] internal method or a *"prototype"* property.
+            1. NOTE: Functions whose _kind_ is ~async~ are not constructable and do not have a [[Construct]] internal method nor a *"prototype"* property.
```
syg, line 26307: This should remain an "or". "nor" tends to be paired with "neither", or only used when preceding a verb phrase. In this case we have two noun phrases.

spec.html
```diff
@@ -26304,14 +26304,14 @@ <h1>Runtime Semantics: CreateDynamicFunction ( _constructor_, _newTarget_, _kind
...
             1. Let _prefix_ be the prefix associated with _kind_ in <emu-xref href="#table-dynamic-function-sourcetext-prefixes"></emu-xref>.
             1. Let _sourceString_ be the string-concatenation of _prefix_, *" anonymous("*, _P_, 0x000A (LINE FEED), *") {"*, _bodyString_, and *"}"*.
             1. Set _F_.[[SourceText]] to ! UTF16DecodeString(_sourceString_).
             1. Return _F_.
           </emu-alg>
           <emu-note>
-            <p>A *"prototype"* property is created for every non-async function created using CreateDynamicFunction to provide for the possibility that the function will be used as a constructor.</p>
+            <p>CreateDynamicFunction defines a *"prototype"* property on any function whose _kind_ is not ~async~ to provide for the possibility that the function will be used as a constructor.</p>
```
syg, line 26314:
```suggestion
            <p>CreateDynamicFunction defines a *"prototype"* property on any function it creates whose _kind_ is not ~async~ to provide for the possibility that the function will be used as a constructor.</p>
```

PR #1955, opened 2020-04-19
Normative: Fixed bug in Number::unsignedRightShift

jhnaldo:
In the previous version ES2019, [The Unsigned Right Shift Operator(`>>>`)](https://www.ecma-international.org/ecma-262/#sec-unsigned-right-shift-operator-runtime-semantics-evaluation) utilize [`ToUint32`](https://www.ecma-international.org/ecma-262/#sec-touint32) for both left and right expressions. However, the current version apply [`ToUint32`](https://tc39.es/ecma262/#sec-touint32) for only the right number value but use [`ToInt32`](https://tc39.es/ecma262/#sec-toint32) for the left number value in [`Number::unsignedRightShift`](https://tc39.es/ecma262/#sec-numeric-types-number-unsignedRightShift). I think it breaks the backward compatibility thus I revised the semantics of [`Number::unsignedRightShift`](https://tc39.es/ecma262/#sec-numeric-types-number-unsignedRightShift) to use [`ToUint32`](https://tc39.es/ecma262/#sec-touint32) for both left and right number value.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1692,7 +1692,7 @@
         <emu-clause id="sec-numeric-types-number-unsignedRightShift">
           <h1>Number::unsignedRightShift ( _x_, _y_ )</h1>
           <emu-alg>
-            1. Let _lnum_ be ! ToInt32(_x_).
+            1. Let _lnum_ be ! ToUint32(_x_).
             1. Let _rnum_ be ! ToUint32(_y_).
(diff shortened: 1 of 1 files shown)


caiolima:
Sorry for late comment here. I confirm this was a bug introduced with BigInt PR. Thanks for fixing it.

PR #1959, opened 2020-04-20
Editorial: Ambiguous notation `≠` in ArraySetLength

jhnaldo:
In the [ArraySetLength](https://tc39.es/ecma262/#sec-arraysetlength) algorithm, the operator `≠` is ambiguous. According to the following test262 tests,

- [test262/test/built-ins/Object/defineProperties/15.2.3.7-6-a-127.js](https://github.com/tc39/test262/blob/master/test/built-ins/Object/defineProperties/15.2.3.7-6-a-127.js)
- [test262/test/built-ins/Object/defineProperty/15.2.3.6-4-131.js](https://github.com/tc39/test262/blob/master/test/built-ins/Object/defineProperty/15.2.3.6-4-131.js)

the algorithm should not throw the RangeError exception when `newLen` is +0 and `numberLen` is -0. However, the ArraySetLength algorithm throws a RangeError exception when `newLen ≠ numberLen`. According, to the [`Number::equal`](https://tc39.es/ecma262/#sec-numeric-types-number-equal) algorithm, +0 and -0 is not the same Number value. Therefore, I revised the fifth line to use [`Number::equal`](https://tc39.es/ecma262/#sec-numeric-types-number-equal) instead fo the ambiguous notation `≠` for clarity.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -8975,7 +8975,7 @@
           1. Let _newLenDesc_ be a copy of _Desc_.
           1. Let _newLen_ be ? ToUint32(_Desc_.[[Value]]).
           1. Let _numberLen_ be ? ToNumber(_Desc_.[[Value]]).
-          1. If _newLen_ &ne; _numberLen_, throw a *RangeError* exception.
+          1. If Number::equal(_newLen_, _numberLen_) is *false*, throw a *RangeError* exception.
           1. Set _newLenDesc_.[[Value]] to _newLen_.
           1. Let _oldLenDesc_ be OrdinaryGetOwnProperty(_A_, *"length"*).
           1. Assert: ! IsDataDescriptor(_oldLenDesc_) is *true*.


devsnek:
I think in this context, `a ≠ b` already means that it should use the floating point definition of equality (where -0 == 0), thanks to https://tc39.es/ecma262/#sec-mathematical-operations.

ljharb:
I agree, it's not ambiguous as-is.

That said, I'm also +0 on this change.

jhnaldo:
> I think in this context, `a ≠ b` already means that it should use the floating point definition of equality (where -0 == 0), thanks to https://tc39.es/ecma262/#sec-mathematical-operations.

So, you mean that `-0 is not the same Number value as +0` but `-0 = +0`
because the operators `=` and `≠` are  IEEE 754-2019 double-precision floating point operators, right?

jhnaldo:
image
Then, how about this one?
According to your interpretation, I think that the step 7 in [`StringGetOwnProperty`](https://tc39.es/ecma262/#sec-stringgetownproperty) return **undefined** not only if *index* is -0 but also if *index* is +0 because the equality operator should be the IEEE 754-2019 double-precision floating point operator.

devsnek:
It should be changed to say "If index is -0"

jhnaldo:
> It should be changed to say "If index is -0"

Thank you for the explanation. I understand now! Thus, I will close this pull request.

PR #1963, opened 2020-04-22
Editorial: Using `is -0` instead of `= -0` in StringGetOwnProperty

jhnaldo:
I revised a small part in [`StringGetOwnProperty`](https://tc39.es/ecma262/#sec-stringgetownproperty) to use `is -0` instead of `= -0` to correctly represent the semantics. The expression `index = -0` denotes that `index` is `-0` or `+0` because the `=` operator is floating point equality operator. Thus, instead, I think that it should be revised to `index is -0`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9075,7 +9075,7 @@
           1. Let _index_ be ! CanonicalNumericIndexString(_P_).
           1. If _index_ is *undefined*, return *undefined*.
           1. If IsInteger(_index_) is *false*, return *undefined*.
-          1. If _index_ = *-0*, return *undefined*.
+          1. If _index_ is *-0*, return *undefined*.
           1. Let _str_ be _S_.[[StringData]].
           1. Assert: Type(_str_) is String.
           1. Let _len_ be the length of _str_.


ljharb:
There's a few other cases of `= 0`; we should probably change all of them at once to be "is 0".

(Also, note the overlap with #1086)

jhnaldo:
> There's a few other cases of `= 0`; we should probably change all of them at once to be "is 0".
> 
> (Also, note the overlap with #1086)

I'm confusing to understand the conversations in #1086.
So, in consequence, what is the exact meaning of `= 0` and `is 0`?
I believe that the meaning is as follows:

- `x = 0` denotes `x` is ES Number values `+0`, `-0`, or the mathematical value `0`
- `x is 0` denotes `x` should be mathematical value `0`.

Am I right?

ljharb:
In general, there's a lot of implicitness around Number and mathematical values in the spec; this seems like it may be one of those cases.

jhnaldo:
@ljharb I think that `= -0` and `= 0` cases are slightly different. I don't know whether few cases of `= 0` are correct or not, or even if they are incorrect, I don't know how to revise them.

However, I believe that the `= -0` equality check in [`StringGetOwnProperty`](https://tc39.es/ecma262/#sec-stringgetownproperty) should revised into `is -0`.
According to #1752, it revised `= -0` equality check to use instead `is 0` and it was accepted and merged. Is there other issues that only this `= -0` equality check should be considered with other cases of `= 0`?

bakkot:
I am in favor of this change. Specifically, for the case of comparing ECMAScript values to the ECMAScript values `0` and `-0` I would prefer that we always say `is +0` or `is -0` rather than `= 0`, `= +0`, `= -0`, or `is 0`.

PR #1976, opened 2020-04-28
Editorial: add prefix '?' on calling ToPrimitive in Abstract Equality Comparison

h2oche:
In step 10 and step 11 of [`Abstract Equality Comparison`](https://tc39.es/ecma262/#sec-abstract-equality-comparison) operation, return value of [`ToPrimitive`](https://tc39.es/ecma262/#sec-toprimitive) operation is directly used. However, [`ToPrimitive`](https://tc39.es/ecma262/#sec-toprimitive) operation may return abrupt completion record and adding prefix '?' on calling [`ToPrimitive`](https://tc39.es/ecma262/#sec-toprimitive) operation seems to be proper way of handle this problem.


ljharb:
https://ci.tc39.es/ipr/tc39/ecma262/pull/1976 tells me the IPR form is signed.

PR #1939, opened 2020-04-07
Normative: Update RegExp property aliases per Unicode v13

mathiasbynens:
Tests: https://github.com/tc39/test262/pull/2526

Ref. #1897 & #1896.

PR #1992, opened 2020-05-13
Editorial: use real table rows instead of ULs in Unicode tables

michaelficarra:
Compare [current](https://tc39.es/ecma262/#table-binary-unicode-properties) with [updated](https://ci.tc39.es/preview/tc39/ecma262/pull/1992/#table-binary-unicode-properties).

I also removed a few duplicate entries from Table 58: `Cham`, `Lisu`, `Modi`, `Newa`, and `Thai`.

FYI @mathiasbynens this would require a rebase of #1896 and #1939. I'm happy to do it for you if we merge this first.

### DIFF
--- a/table-binary-unicode-properties.html
+++ b/table-binary-unicode-properties.html
@@ -12,22 +12,18 @@
       <td><a href="https://unicode.org/reports/tr18/#General_Category_Property">`ASCII`</a></td>
     </tr>
     <tr>
-      <td>
-        <ul>
-          <li>`ASCII_Hex_Digit`</li>
-          <li>`AHex`</li>
-        </ul>
-      </td>
-      <td><a href="https://unicode.org/reports/tr44/#ASCII_Hex_Digit">`ASCII_Hex_Digit`</a></td>
+      <td>`ASCII_Hex_Digit`</td>
+      <td rowspan="2"><a href="https://unicode.org/reports/tr44/#ASCII_Hex_Digit">`ASCII_Hex_Digit`</a></td>
     </tr>
     <tr>
-      <td>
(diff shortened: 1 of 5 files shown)


table-binary-unicode-properties.html
```diff
@@ -12,22 +12,19 @@
...
     </tr>
     <tr>
-      <td>
-        <ul>
-          <li>`ASCII_Hex_Digit`</li>
-          <li>`AHex`</li>
-        </ul>
+      <td>`ASCII_Hex_Digit`</td>
+      <td>`AHex`</td>
       </td>
```
bakkot, line 17: Seems stray?

michaelficarra: Figures, the only one I did by hand is messed up 🙄. Ugh.

table-binary-unicode-properties.html
```diff
@@ -12,22 +12,18 @@
...
     <tr>
-      <td>
-        <ul>
-          <li>`ASCII_Hex_Digit`</li>
-          <li>`AHex`</li>
-        </ul>
-      </td>
-      <td><a href="https://unicode.org/reports/tr44/#ASCII_Hex_Digit">`ASCII_Hex_Digit`</a></td>
+      <td>`ASCII_Hex_Digit`</td>
+      <td rowspan=2><a href="https://unicode.org/reports/tr44/#ASCII_Hex_Digit">`ASCII_Hex_Digit`</a></td>
```
ljharb, line 16: i have a mild preference for always quoting all html property values

michaelficarra: Done.

PR #1995, opened 2020-05-16
Meta: bump ecmarkup, enforce linting

bakkot:
This enforces some simple linting rules for the specification added in https://github.com/tc39/ecmarkup/pull/199. Specifically, during `npm run build-master`, it checks:

- grammarkdown's built-in validity checks for the grammar
- the productions used in the definition of each early error and SDO are defined in the main grammar (at least when it is able to identify the early errors and SDOs)
- those productions do not include `[no LineTerminator here]` restrictions or `[+flag]` gating
- line endings in algorithm steps are consistent with the current style; [see here](https://github.com/tc39/ecmarkup/blob/7930f0baaea557db600023d33427e8eefdfffa97/src/lint/rules/algorithm-line-endings.ts#L5-L19) for precisely what that means

If any lint rule fails, the build fails. Hopefully we'll find out, because there's one place which currently fails, which I'll fix in a second commit.

Note that there's [a bunch more things we could reasonably enforce](https://github.com/tc39/ecmarkup/issues/173) ; PRs (to ecmarkup) welcome!

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "build-spec": "YEAR=2020 && git checkout --quiet \"es${YEAR}\" && mkdir -p \"out/${YEAR}\" && cp -R img \"out/${YEAR}\" && ecmarkup --verbose spec.html \"out/${YEAR}/index.html\" --css \"out/${YEAR}/ecmarkup.css\" --js \"out/${YEAR}/ecmarkup.js\" && git checkout --quiet travis-origin/test-travis",
     "postbuild-spec": "git remote rm travis-origin",
     "prebuild-master": "npm run clean && mkdir out && cp -R img out",
-    "build-master": "npm run wipe-es6biblio && ecmarkup --verbose spec.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js",
+    "build-master": "npm run wipe-es6biblio && ecmarkup --lint-spec --verbose spec.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js",
     "build": "npm run build-master",
     "build-for-pdf": "npm run build-master -- --old-toc",
     "build-travis": "npm run build-master && npm run build-spec",
@@ -26,7 +26,7 @@
   "license": "SEE LICENSE IN https://tc39.es/ecma262/#sec-copyright-and-software-license",
   "homepage": "https://tc39.es/ecma262/",
   "dependencies": {
-    "ecmarkup": "^3.19.0"
+    "ecmarkup": "^3.21.0"
   },
   "devDependencies": {
     "@alrra/travis-scripts": "^2.1.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,70 @@
       "integrity": "sha1-vL50jgHjxZ6IT22jxrHSLCg3ChE=",
       "dev": true
     },
+    "@babel/code-frame": {
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
+      "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
+      "requires": {
+        "@babel/highlight": "^7.8.3"
+      }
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "build-spec": "YEAR=2020 && git checkout --quiet \"es${YEAR}\" && mkdir -p \"out/${YEAR}\" && cp -R img \"out/${YEAR}\" && ecmarkup --verbose spec.html \"out/${YEAR}/index.html\" --css \"out/${YEAR}/ecmarkup.css\" --js \"out/${YEAR}/ecmarkup.js\" && git checkout --quiet travis-origin/test-travis",
     "postbuild-spec": "git remote rm travis-origin",
     "prebuild-master": "npm run clean && mkdir out && cp -R img out",
-    "build-master": "npm run wipe-es6biblio && ecmarkup --verbose spec.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js",
+    "build-master": "npm run wipe-es6biblio && ecmarkup --lint-spec --verbose spec.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js",
```
ljharb, line 12: rather than add this in `build-master`, perhaps this should be added as a travis job, so it can run in parallel with the build process?

package.json
```diff
@@ -26,7 +26,7 @@
   "license": "SEE LICENSE IN https://tc39.es/ecma262/#sec-copyright-and-software-license",
   "homepage": "https://tc39.es/ecma262/",
   "dependencies": {
-    "ecmarkup": "^3.19.0"
+    "ecmarkup": "3.20.0"
```
ljharb, line 29:
```suggestion
    "ecmarkup": "^3.20.0"
```

package.json
bakkot: The linting isn't really divorced from the building, since they do much of the same processing. And if you pass `--lint-spec` and there are no linting errors it just builds. (Think of it like `-Wall -Werror`, if you're familiar with C++ compilers.) I don't think there's much to be gained from splitting it out as a separate job.

ljharb: i was thinking clearer error messages - ie, the lint failed but the build passed tells you more than "the build failed"

package.json
bakkot: I'd personally prefer pinning a precise version, so that the build output only changes when a human explicitly makes the decision to change it.

(I feel like we've had this conversation already, but I can't find it. Maybe it was with someone else.)

@michaelficarra @syg thoughts?

ljharb: That's what the lockfile already achieves, for those interested in using it, like ecma262. The `^` indicates that ecmarkup follows semver. In other words, pinning it here adds no determinism, because we already have a lockfile.

bakkot: Ah, yeah, lockfile is good enough. Will fix.

bakkot:
(force-pushed to update the version of ecmarkup)

package.json
michaelficarra: @bakkot Thoughts on having a `--lint-only` that exits after linting, regardless of the number of errors?

bakkot: @michaelficarra Some of the [things I want to enforce](https://github.com/tc39/ecmarkup/issues/173) (e.g. the output is valid HTML) could not feasibly done without running the full build, so I'm not sure there would be much point.

michaelficarra: @bakkot That sounds like it could be the job of a separate, more generic tool.

bakkot: True.

I'm fine with adding `--lint-only`, I guess. I am still worried that we may want to add linting checks which are integrated with later steps of the compilation, which would make it less useful.

bakkot: @michaelficarra For a concrete example of a late-stage linting error, ecmarkup currently `console.log`s a warning when [it finds a duplicate id](https://github.com/tc39/ecmarkup/blob/c7e4349b54f7100009cdbdc5f81779e30be2b0e6/src/Builder.ts#L29) or [it can't link an xref](https://github.com/tc39/ecmarkup/blob/b27317bc9a78f67a04b2ace5babd0ba05e930f68/src/Xref.ts#L121). It seems like these should be linting errors instead. But building and xref linking happen relatively late in the compilation process.

michaelficarra: It'd still probably be useful to be able to do linting without having to output unwanted artifacts to the filesystem.

Issue #385, opened 2016-02-12
Add <dfn>s for all the well-known symbols (and make sure they link)


domenic:
Similarly do what we did for %intrinsics%, we should do the same for @@wellKnownSymbols.

PR #1996, opened 2020-05-17
Editorial: Add <dfn> to well‑known symbols

ExE-Boss:
Fixes <https://github.com/tc39/ecma262/issues/385>

---

[Preview](https://ci.tc39.es/preview/tc39/ecma262/pull/1996) ([#sec-well-known-symbols](https://ci.tc39.es/preview/tc39/ecma262/pull/1996#sec-well-known-symbols))

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1006,7 +1006,7 @@
             </tr>
             <tr>
               <td>
-                @@asyncIterator
+                <dfn>@@asyncIterator</dfn>
               </td>
               <td>
                 *"Symbol.asyncIterator"*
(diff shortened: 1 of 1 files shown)


ExE-Boss:
The downside of moving the `<dfn>`s to the [table of well‑known symbols](https://ci.tc39.es/preview/tc39/ecma262/pull/1996#sec-well-known-symbols) is that the symbol auto‑links now point to the section as a whole instead of the specific symbol.

bakkot:
I'm fine with that. For someone who doesn't know what the `@@` notation means, the context in that section is useful. And it's easy to find the particular symbol you're looking for in the table.

ExE-Boss:
It’s possible to fix that by simply adding the `id` attribute to the `<dfn>` (or the enclosing `<td>`/`<tr>`) elements.

Issue #1989, opened 2020-05-10
Editorial: ByteListBitwiseOp parameter _op_


jmdyck:
ByteListBitwiseOp's parameter `_op_` is declared to be a read-modify-write modification function. However, that's a closure that takes two *Lists* of byte values, whereas `_op_` is applied to merely two byte values. (If `_op_` really *did* operate on two Lists of byte values,  you wouldn't need ByteListBitwiseOp.)

PR #1998, opened 2020-05-19
Editorial: make `NumberBitwiseOp`, `BigIntBitwiseOp`, `ByteListBitwiseOp` consistent

ljharb:
Ensure `op` is a String in all cases, and ensure all three operations assert on the contents of `op`.

Previously, `NumberBitwiseOp` and `ByteListBitwiseOp ` did not assert on `op`, and they were passing an operator as if it was a first-class value. Instead, I made all three use the approach of `BigIntBitwiseOp`.

An alternative to strings would be to use `~^~` etc; happy to switch them all to that form if that's preferred.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1766,8 +1766,9 @@
 
         <emu-clause id="sec-numberbitwiseop" aoid="NumberBitwiseOp">
           <h1>NumberBitwiseOp ( _op_, _x_, _y_ )</h1>
-          <p>The abstract operation NumberBitwiseOp takes arguments _op_, _x_, and _y_. It performs the following steps when called:</p>
+          <p>The abstract operation NumberBitwiseOp takes arguments _op_ (a sequence of Unicode code points), _x_, and _y_. It performs the following steps when called:</p>
           <emu-alg>
+            1. Assert: _op_ is `&amp;`, `^`, or `|`.
             1. Let _lnum_ be ! ToInt32(_x_).
             1. Let _rnum_ be ! ToInt32(_y_).
             1. Return the result of applying the bitwise operator _op_ to _lnum_ and _rnum_. The result is a signed 32-bit integer.
@@ -2019,26 +2020,26 @@
 
         <emu-clause id="sec-bigintbitwiseop" aoid="BigIntBitwiseOp">
           <h1>BigIntBitwiseOp ( _op_, _x_, _y_ )</h1>
-          <p>The abstract operation BigIntBitwiseOp takes arguments _op_, _x_, and _y_. It performs the following steps when called:</p>
+          <p>The abstract operation BigIntBitwiseOp takes arguments _op_ (a sequence of Unicode code points), _x_, and _y_. It performs the following steps when called:</p>
           <emu-alg>
-            1. Assert: _op_ is *"&amp;"*, *"|"*, or *"^"*.
+            1. Assert: _op_ is `&amp;`, `^`, or `|`.
             1. Let _result_ be *0n*.
             1. Let _shift_ be 0.
             1. Repeat, until (_x_ = 0 or _x_ = -1) and (_y_ = 0 or _y_ = -1),
               1. Let _xDigit_ be _x_ modulo 2.
               1. Let _yDigit_ be _y_ modulo 2.
-              1. If _op_ is *"&amp;"*, set _result_ to _result_ + 2<sup>_shift_</sup> &times; BinaryAnd(_xDigit_, _yDigit_).
-              1. Else if _op_ is *"|"*, set _result_ to _result_ + 2<sup>_shift_</sup> &times; BinaryOr(_xDigit_, _yDigit_).
+              1. If _op_ is `&amp;`, set _result_ to _result_ + 2<sup>_shift_</sup> &times; BinaryAnd(_xDigit_, _yDigit_).
+              1. Else if _op_ is `|`, set _result_ to _result_ + 2<sup>_shift_</sup> &times; BinaryOr(_xDigit_, _yDigit_).
               1. Else,
-                1. Assert: _op_ is *"^"*.
+                1. Assert: _op_ is `^`.
                 1. Set _result_ to _result_ + 2<sup>_shift_</sup> &times; BinaryXor(_xDigit_, _yDigit_).
               1. Set _shift_ to _shift_ + 1.
               1. Set _x_ to (_x_ - _xDigit_) / 2.
               1. Set _y_ to (_y_ - _yDigit_) / 2.
-            1. If _op_ is *"&amp;"*, let _tmp_ be BinaryAnd(_x_ modulo 2, _y_ modulo 2).
-            1. Else if _op_ is *"|"*, let _tmp_ be BinaryOr(_x_ modulo 2, _y_ modulo 2).
+            1. If _op_ is `&amp;`, let _tmp_ be BinaryAnd(_x_ modulo 2, _y_ modulo 2).
+            1. Else if _op_ is `|`, let _tmp_ be BinaryOr(_x_ modulo 2, _y_ modulo 2).
             1. Else,
-              1. Assert: _op_ is *"^"*.
+              1. Assert: _op_ is `^`.
               1. Let _tmp_ be BinaryXor(_x_ modulo 2, _y_ modulo 2).
             1. If _tmp_ &ne; 0, then
               1. Set _result_ to _result_ - 2<sup>_shift_</sup>.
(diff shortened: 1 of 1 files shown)


michaelficarra:
I prefer the token to the string. It feels like a bit less indirection to me.

ljharb:
Like `~|~`?

michaelficarra:
Yeah.

ljharb:
@michaelficarra updated!

spec.html
```diff
@@ -2021,7 +2022,7 @@ <h1>BinaryXor ( _x_, _y_ )</h1>
           <h1>BigIntBitwiseOp ( _op_, _x_, _y_ )</h1>
           <p>The abstract operation BigIntBitwiseOp takes arguments _op_, _x_, and _y_. It performs the following steps when called:</p>
           <emu-alg>
-            1. Assert: _op_ is *"&amp;"*, *"|"*, or *"^"*.
+            1. Assert: _op_ is ~&amp;~, ~|~, or ~^~.
```
jmdyck, line 2025: In the subsequent tests on `_op_`, you need to change the value that it's compared to.

(Alternatively, maybe BigIntBitwiseOp can just talk about `the result of applying _op_ to ...`, the same as NumberBitwiseOp and ByteListBitwiseOp do. And drop BinaryAnd, BinaryOr, BinaryXor.)

spec.html
```diff
@@ -37871,6 +37872,7 @@ <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
         <h1>ByteListBitwiseOp( _op_, _xBytes_, _yBytes_ )</h1>
         <p>The abstract operation ByteListBitwiseOp takes arguments _op_ (a read-modify-write modification function), _xBytes_ (a List of byte values), and _yBytes_ (a List of byte values). The operation atomically performs a bitwise operation on all byte values of the arguments and returns a List of byte values. It performs the following steps when called:</p>
         <emu-alg>
+          1. Assert: _op_ is ~&amp;~, ~|~, or ~^~.
```
jmdyck, line 37875: The preamble's description of `_op_` should be modified likewise (which might take care of Issue #1989).

spec.html
```diff
@@ -1778,23 +1779,23 @@ <h1>NumberBitwiseOp ( _op_, _x_, _y_ )</h1>
...
-            1. Return NumberBitwiseOp(`&amp;`, _x_, _y_).
+            1. Return NumberBitwiseOp(~&amp;~, _x_, _y_).
           </emu-alg>
         </emu-clause>
 
-        <emu-clause id="sec-numeric-types-number-bitwiseXOR">
-          <h1>Number::bitwiseXOR ( _x_, _y_ )</h1>
-          <p>The abstract operation Number::bitwiseXOR takes arguments _x_ (a Number) and _y_ (a Number). It performs the following steps when called:</p>
+        <emu-clause id="sec-numeric-types-number-bitwiseOR">
+          <h1>Number::bitwiseOR ( _x_, _y_ )</h1>
```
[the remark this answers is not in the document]
ljharb: I want all three to be in the same order; I don't care what the order is. I'll make them all be `&`, `^`, `|` for consistency.

spec.html
```diff
@@ -1768,6 +1768,7 @@ <h1>Number::sameValueZero ( _x_, _y_ )</h1>
           <h1>NumberBitwiseOp ( _op_, _x_, _y_ )</h1>
           <p>The abstract operation NumberBitwiseOp takes arguments _op_, _x_, and _y_. It performs the following steps when called:</p>
           <emu-alg>
+            1. Assert: _op_ is ~&amp;~, ~|~, or ~^~.
```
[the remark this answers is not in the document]
ljharb: At the moment, I don't think there's actually anything that describes the action except `result of applying the bitwise operator _op_`. So, for now, I think using the actual operator tokens is clearest.

In a future PR, i think it'd be fine to explore an alternative, perhaps with some kind of explicit lookup.

spec.html
```diff
@@ -37868,9 +37869,10 @@ <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
       </emu-clause>
 
       <emu-clause id="sec-bytelistbitwiseop" aoid="ByteListBitwiseOp">
-        <h1>ByteListBitwiseOp( _op_, _xBytes_, _yBytes_ )</h1>
-        <p>The abstract operation ByteListBitwiseOp takes arguments _op_ (a read-modify-write modification function), _xBytes_ (a List of byte values), and _yBytes_ (a List of byte values). The operation atomically performs a bitwise operation on all byte values of the arguments and returns a List of byte values. It performs the following steps when called:</p>
+        <h1>ByteListBitwiseOp ( _op_, _xBytes_, _yBytes_ )</h1>
```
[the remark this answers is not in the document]
bakkot: That seems like the sort of thing a linter ought to check! Created https://github.com/tc39/ecmarkup/pull/205.

michaelficarra:
@ljharb In 12.15.6 (Runtime Semantics: EvaluateStringOrNumericBinaryExpression), we describe the `opText` parameter as a sequence of Unicode code points. Do we need to change that as well?

I also don't like how (as far as I can tell) we don't explicitly say anywhere that, for example, `^` means the bitwise xor operation. In `NumberBitwiseOp`, we just say

> Return the result of applying the bitwise operator *op* to *lnum* and *rnum*.

Do we actually say anywhere that `^` means xor or are we just supposed to infer that from the names of productions and AOs?

ljharb:
I believe we’re just supposed to infer it, which I’m not happy with either.

I’ll update 12.15.16 for sure; in the meantime, I’d either be happy to take a suggestion for how to specify this stuff better in this PR, or also to do it in a followup so that this one can at least remove a bunch of inconsistency.

ljharb:
@michaelficarra hmm, i'm actually not sure how to update 12.15.16 without solving that larger problem. Would you be OK with flipping this PR back to using strings, and then in a followup, we can tackle the larger issue of how these are defined and specified?

michaelficarra:
@ljharb Yeah, let's go back to strings for now. I'll open an issue to track being more explicit with numeric operations.

spec.html
```diff
@@ -2030,15 +2031,15 @@ <h1>BigIntBitwiseOp ( _op_, _x_, _y_ )</h1>
               1. If _op_ is *"&amp;"*, set _result_ to _result_ + 2<sup>_shift_</sup> &times; BinaryAnd(_xDigit_, _yDigit_).
               1. Else if _op_ is *"|"*, set _result_ to _result_ + 2<sup>_shift_</sup> &times; BinaryOr(_xDigit_, _yDigit_).
               1. Else,
-                1. Assert: _op_ is *"^"*.
+                1. Assert: _op_ is *"^"*,
```
syg, line 2034: Typo: , instead of .

spec.html
```diff
@@ -2030,15 +2031,15 @@ <h1>BigIntBitwiseOp ( _op_, _x_, _y_ )</h1>
...
                 1. Set _result_ to _result_ + 2<sup>_shift_</sup> &times; BinaryXor(_xDigit_, _yDigit_).
               1. Set _shift_ to _shift_ + 1.
               1. Set _x_ to (_x_ - _xDigit_) / 2.
               1. Set _y_ to (_y_ - _yDigit_) / 2.
             1. If _op_ is *"&amp;"*, let _tmp_ be BinaryAnd(_x_ modulo 2, _y_ modulo 2).
             1. Else if _op_ is *"|"*, let _tmp_ be BinaryOr(_x_ modulo 2, _y_ modulo 2).
             1. Else,
-              1. Assert: _op_ is *"^"*.
+              1. Assert: _op_ is *"^"*,
```
syg, line 2042: Typo: , instead of . Global search and replace snafu?

spec.html
```diff
@@ -37868,9 +37869,10 @@ <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
       </emu-clause>
 
       <emu-clause id="sec-bytelistbitwiseop" aoid="ByteListBitwiseOp">
-        <h1>ByteListBitwiseOp( _op_, _xBytes_, _yBytes_ )</h1>
-        <p>The abstract operation ByteListBitwiseOp takes arguments _op_ (a read-modify-write modification function), _xBytes_ (a List of byte values), and _yBytes_ (a List of byte values). The operation atomically performs a bitwise operation on all byte values of the arguments and returns a List of byte values. It performs the following steps when called:</p>
+        <h1>ByteListBitwiseOp ( _op_, _xBytes_, _yBytes_ )</h1>
+        <p>The abstract operation ByteListBitwiseOp takes arguments _op_ (a sequence of Unicode code points), _xBytes_ (a List of byte values), and _yBytes_ (a List of byte values). The operation atomically performs a bitwise operation on all byte values of the arguments and returns a List of byte values. It performs the following steps when called:</p>
```
syg, line 37873: Oops, good catch and thanks for fixing.

PR #2007, opened 2020-05-22
Editorial: Make numeric values and operations default to mathematical values

caiolima:
As reported on #1964, the fact that we are using Number as default representation of numeric literals and operations creates uncounted bugs throughout the spec. This PR is proposing to rollback the default value to mathematical value. Since the major intention of PR #1135 was to avoid implicit conversions between numeric versions, we are keeping this change but it necessary to properly cast those numeric types when they are manipulated, since mixing different numeric types is not allowed.

The goals we want to achieve here are:

1. Resolve bugs on current version of spec like #1960 or #1229.
1. Mathematical values should never be leaked and are internal concept of ECMA262 specification. If this happens, it should be considered as editorial error.
1. Avoid the maximum of explicit conversions between Number, BigInt and Mathematical values.

Cc. @littledan @waldemarhorwat @ljharb @syg @michaelficarra @bakkot 

PS. It's still a WIP PR, but I think we have enough changes to discuss some changes made before I keep moving forward.

Issue #1964, opened 2020-04-22
Arithmetic broken throughout the spec


waldemarhorwat:
After doing more analysis of bugs #1960 and #1962, I've found hundreds of other bugs in the spec introduced by pull request #1135 and have come to the conclusion that we need to revert #1135. It's impractical to find and fix them all without a large effort, and approximately no one reading or writing the spec seems to realize the consequences of what this did:

* Formulas in the spec were commonly understood to be mathematical formulas before #1135 and were written in common mathematical notation. #1135 broke that: A formula like a + b + c in the spec behaves differently depending on whether it's evaluated as (a + b) + c or a + (b + c), and this was not intended. Such formulas were meant to be mathematical formulas, and those generally rely on the standard mathematical identities being valid, which they do not if they are interpreted as Number arithmetic.
* Algorithms sometimes use integer counters for various invisible and expository spec purposes. If, per #1135, these are interpreted as Numbers instead of true integers then those algorithms break when they overflow.
* What operators like a<sup>b</sup> do in the spec is well-defined only for mathematical formulas. When, as #1135 changed the spec to do, that notation is interpreted as Number arithmetic (which, incidentally, the notation a<sup>b</sup> is never actually defined for), it's almost always wrong: given an integer n, 10<sup>n</sup> is not always a power of 10, 2<sup>n</sup> is not always a power of 2, etc. This created a huge spec bug farm and most of the spec authors are unaware of it even for new parts of the spec.
* The confusion will get even worse with the addition of Decimal.

The correct approach is to revert #1135 and restore the commonly understood mathematical meaning of formulas. Number, Decimal, etc. arithmetic should be done explicitly and only one operation at a time; in most cases it's just a single addition or subtraction. We should never use notation such as a<sup>b</sup> on Numbers.

ljharb:
cc @littledan @caiolima

littledan:
@waldemarhorwat I'd like to make sure these bugs are fixed. You've noted a number of changes that were the broad strokes of the spec change, which were on purpose. You say you've found hundreds of bugs--could you note some of these concretely so we could go about fixing them? We could always fix a particular error by using the R subscript when appropriate--the change specifically left that possibility open.

littledan:
Also, if you have other ideas for how we should phrase things in the specification, then I'd be interested to hear them, so we can fix all errors while keeping BigInt in the specification.

waldemarhorwat:
[...]

PR #2009, opened 2020-05-24
Normative: Add String.prototype.replaceAll

mathiasbynens:
Proposal: https://github.com/tc39/proposal-string-replaceall

Tests: https://github.com/tc39/test262/pull/2423

Note that the changes to `String.prototype.matchAll` that were part of this proposal have already been merged here: https://github.com/tc39/ecma262/pull/1716.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -10452,6 +10452,27 @@
     </emu-clause>
   </emu-clause>
 
+  <emu-clause id="sec-stringindexof" aoid="StringIndexOf">
+    <h1>Runtime Semantics: StringIndexOf ( _string_, _searchValue_, _fromIndex_ )</h1>
+    <p>The abstract operation StringIndexOf takes arguments _string_ (a String), _searchValue_ (a String), and _fromIndex_ (a non-negative integer). It performs the following steps when called:</p>
+    <emu-alg>
+      1. Assert: Type(_string_) is String.
+      1. Assert: Type(_searchValue_) is String.
+      1. Assert: ! IsNonNegativeInteger(_fromIndex_) is *true*.
+      1. Let _len_ be the length of _string_.
+      1. If _searchValue_ is the empty String and _fromIndex_ &le; _len_, return _fromIndex_.
+      1. Let _searchLen_ be the length of _searchValue_.
+      1. If there exists any integer _k_ such that _fromIndex_ &le; _k_ &le; _len_ - _searchLen_ and for all nonnegative integers _j_ less than _searchLen_, the code unit at index _k_ + _j_ within _string_ is the same as the code unit at index _j_ within _searchValue_, let _pos_ be the smallest (closest to *-&infin;*) such integer. Otherwise, let _pos_ be -1.
+      1. Return _pos_.
+    </emu-alg>
+    <emu-note>
+      <p>If _searchValue_ is empty and _fromIndex_ is less than or equal to the length of _string_, this algorithm returns _fromIndex_. An empty _searchValue_ is effectively found at every position within a string, including after the last code unit.</p>
+    </emu-note>
+    <emu-note>
+      <p>This algorithm always returns -1 if _fromIndex_ &gt; the length of _string_.</p>
+    </emu-note>
+  </emu-clause>
+
   <emu-clause id="sec-types-of-source-code">
     <h1>Types of Source Code</h1>
     <p>There are four types of ECMAScript code:</p>
@@ -29987,7 +30008,9 @@ THH:mm:ss.sss
           1. Let _len_ be the length of _S_.
           1. Let _start_ be min(max(_pos_, 0), _len_).
           1. Let _searchLen_ be the length of _searchStr_.
-          1. If there exists any integer _k_ not smaller than _start_ such that _k_ + _searchLen_ is not greater than _len_, and for all nonnegative integers _j_ less than _searchLen_, the code unit at index _k_ + _j_ within _S_ is the same as the code unit at index _j_ within _searchStr_, return *true*; but if there is no such integer _k_, return *false*.
+          1. Let _index_ be ! StringIndexOf(_S_, _searchStr_, _start_).
+          1. If _index_ is not -1, return *true*.
+          1. Return *false*.
         </emu-alg>
         <emu-note>
           <p>If _searchString_ appears as a substring of the result of converting this object to a String, at one or more indices that are greater than or equal to _position_, return *true*; otherwise, returns *false*. If _position_ is *undefined*, 0 is assumed, so as to search all of the String.</p>
@@ -30015,7 +30038,7 @@ THH:mm:ss.sss
           1. Let _len_ be the length of _S_.
           1. Let _start_ be min(max(_pos_, 0), _len_).
           1. Let _searchLen_ be the length of _searchStr_.
-          1. Return the smallest possible integer _k_ not smaller than _start_ such that _k_ + _searchLen_ is not greater than _len_, and for all nonnegative integers _j_ less than _searchLen_, the code unit at index _k_ + _j_ within _S_ is the same as the code unit at index _j_ within _searchStr_; but if there is no such integer _k_, return the value -1.
+          1. Return ! StringIndexOf(_S_, _searchStr_, _start_).
         </emu-alg>
         <emu-note>
           <p>The `indexOf` function is intentionally generic; it does not require that its *this* value be a String object. Therefore, it can be transferred to other kinds of objects for use as a method.</p>
@@ -30366,6 +30389,50 @@ THH:mm:ss.sss
             </table>
           </emu-table>
         </emu-clause>
+      </emu-clause>
+
+      <emu-clause id="sec-string.prototype.replaceall">
+        <h1>String.prototype.replaceAll ( _searchValue_, _replaceValue_ )</h1>
+        <p>When the `replaceAll` method is called with arguments _searchValue_ and _replaceValue_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be ? RequireObjectCoercible(*this* value).
+          1. If _searchValue_ is neither *undefined* nor *null*, then
+            1. Let _isRegExp_ be ? IsRegExp(_searchValue_).
+            1. If _isRegExp_ is *true*, then
+              1. Let _flags_ be ? Get(_searchValue_, *"flags"*).
+              1. Perform ? RequireObjectCoercible(_flags_).
+              1. If ? ToString(_flags_) does not contain *"g"*, throw a *TypeError* exception.
+            1. Let _replacer_ be ? GetMethod(_searchValue_, @@replace).
+            1. If _replacer_ is not *undefined*, then
+              1. Return ? Call(_replacer_, _searchValue_, « _O_, _replaceValue_ »).
+          1. Let _string_ be ? ToString(_O_).
+          1. Let _searchString_ be ? ToString(_searchValue_).
+          1. Let _functionalReplace_ be IsCallable(_replaceValue_).
+          1. If _functionalReplace_ is *false*, then
+            1. Set _replaceValue_ to ? ToString(_replaceValue_).
+          1. Let _searchLength_ be the length of _searchString_.
+          1. Let _advanceBy_ be max(1, _searchLength_).
+          1. Let _matchPositions_ be a new empty List.
+          1. Let _position_ be ! StringIndexOf(_string_, _searchString_, 0).
+          1. Repeat, while _position_ is not -1,
+            1. Append _position_ to the end of _matchPositions_.
+            1. Set _position_ to ! StringIndexOf(_string_, _searchString_, _position_ + _advanceBy_).
+          1. Let _endOfLastMatch_ be 0.
+          1. Let _result_ be the empty String.
+          1. For each _position_ in _matchPositions_, do
+            1. If _functionalReplace_ is *true*, then
+              1. Let _replacement_ be ? ToString(? Call(_replaceValue_, *undefined*, « _searchString_, _position_, _string_ »)).
+            1. Else,
+              1. Assert: Type(_replaceValue_) is String.
+              1. Let _captures_ be a new empty List.
+              1. Let _replacement_ be ! GetSubstitution(_searchString_, _string_, _position_, _captures_, *undefined*, _replaceValue_).
+            1. Let _stringSlice_ be the substring of _string_ consisting of the code units from _endOfLastMatch_ (inclusive) up to _position_ (exclusive).
+            1. Set _result_ to the string-concatenation of _result_, _stringSlice_, and _replacement_.
+            1. Set _endOfLastMatch_ to _position_ + _searchLength_.
+          1. If _endOfLastMatch_ &lt; the length of _string_, then
+            1. Set _result_ to the string-concatenation of _result_ and the substring of _string_ consisting of the code units from _endOfLastMatch_ (inclusive) up through the final code unit of _string_ (inclusive).
+          1. Return _result_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-string.prototype.search">


spec.html
```diff
@@ -10440,6 +10440,25 @@ <h1>Static Semantics: CodePointAt ( _string_, _position_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-stringindexof">
+      <h1>Static Semantics: StringIndexOf ( _string_, _searchValue_, _fromIndex_ )</h1>
+      <p>The abstract operation _StringIndexOf_ performs the following steps:</p>
+      <!-- Note: OOB fromIndex values (fromIndex >= string.length) are accepted by the algorithm. -->
```
mathiasbynens, line 10446: I've kept these comments for now since they seem useful, but let me know if you'd rather have them removed.

ljharb: would this be useful as a non-normative note, rather than a comment?

spec.html
```diff
@@ -30384,6 +30404,50 @@ <h1>Runtime Semantics: GetSubstitution ( _matched_, _str_, _position_, _captures
...
+            1. If _isRegExp_ is true, then
+              1. Let _flags_ be ? Get(_searchValue_, *"flags"*).
+              1. Perform ? RequireObjectCoercible(_flags_).
+              1. If ? ToString(_flags_) does not contain *"g"*, throw a *TypeError* exception.
+            1. Let _replacer_ be ? GetMethod(_searchValue_, @@replace).
+            1. If _replacer_ is not *undefined*, then
+              1. Return ? Call(_replacer_, _searchValue_, « _O_, _replaceValue_ »).
+          1. Let _string_ be ? ToString(_O_).
+          1. Let _searchString_ be ? ToString(_searchValue_).
+          1. Let _functionalReplace_ be IsCallable(_replaceValue_).
```
ljharb, line 30423:
```suggestion
          1. Let _functionalReplace_ be ! IsCallable(_replaceValue_).
```

spec.html
```diff
@@ -30384,6 +30404,50 @@ <h1>Runtime Semantics: GetSubstitution ( _matched_, _str_, _position_, _captures
...
+            1. Let _position_ be ! StringIndexOf(_string_, _searchString_, _position_ + _advanceBy_).
+          1. Let _endOfLastMatch_ be 0.
+          1. Let _result_ be the empty string value.
+          1. For each _position_ in _matchPositions_, do
+            1. If _functionalReplace_ is *true*, then
+                1. Let _replacement_ be ? ToString(? Call(_replaceValue_, *undefined*, « _searchString_, _position_, _string_ »).
+            1. Else,
+                1. Assert: Type(_replaceValue_) is String.
+                1. Let _captures_ be a new empty List.
+                1. Let _replacement_ be GetSubstitution(_searchString_, _string_, _position_, _captures_, *undefined*, _replaceValue_).
```
ljharb, line 30441:
```suggestion
              1. Assert: Type(_replaceValue_) is String.
              1. Let _captures_ be a new empty List.
              1. Let _replacement_ be GetSubstitution(_searchString_, _string_, _position_, _captures_, *undefined*, _replaceValue_).
```

spec.html
```diff
@@ -10440,6 +10440,25 @@ <h1>Static Semantics: CodePointAt ( _string_, _position_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-stringindexof">
```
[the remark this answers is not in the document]
mathiasbynens: As a note, I'd love to just click "commit suggestion" instead of applying all these changes locally but GitHub won't let me, instead showing the "This diff has recently been updated. Refresh and try again." error. If there is some trick to making this work, please let me know.

spec.html
mathiasbynens: Done.

mathiasbynens:
Thanks for the reviews! I've addressed the feedback.

spec.html
```diff
@@ -30031,7 +30052,8 @@ <h1>String.prototype.indexOf ( _searchString_ [ , _position_ ] )</h1>
           1. Let _len_ be the length of _S_.
           1. Let _start_ be min(max(_pos_, 0), _len_).
           1. Let _searchLen_ be the length of _searchStr_.
-          1. Return the smallest possible integer _k_ not smaller than _start_ such that _k_ + _searchLen_ is not greater than _len_, and for all nonnegative integers _j_ less than _searchLen_, the code unit at index _k_ + _j_ within _S_ is the same as the code unit at index _j_ within _searchStr_; but if there is no such integer _k_, return the value -1.
+          1. Let _index_ be ! StringIndexOf(_S_, _searchStr_, _start_).
+          1. Return _index_.
```
ljharb, line 30056:
```suggestion
          1. Return ! StringIndexOf(_S_, _searchStr_, _start_).
```

spec.html
```diff
@@ -30112,7 +30134,7 @@ <h1>String.prototype.matchAll ( _regexp_ )</h1>
           1. Let _O_ be ? RequireObjectCoercible(*this* value).
           1. If _regexp_ is neither *undefined* nor *null*, then
             1. Let _isRegExp_ be ? IsRegExp(_regexp_).
-            1. If _isRegExp_ is *true*, then
+            1. If _isRegExp_ is true, then
```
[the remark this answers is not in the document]
ljharb: cc @bakkot for the linter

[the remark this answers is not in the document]
bakkot: Added a desiderata to [the list](https://github.com/tc39/ecmarkup/issues/173).

That one is a little tricky to lint for, because we say stuff like "the substeps are only applied if the predicate is true" or "if one or more of the following conditions is true:".

spec.html
```diff
@@ -30384,6 +30404,50 @@ <h1>Runtime Semantics: GetSubstitution ( _matched_, _str_, _position_, _captures
...
       </emu-clause>
 
+      <emu-clause id="sec-string.prototype.replaceall">
+        <h1>String.prototype.replaceAll ( _searchValue_, _replaceValue_ )</h1>
+        <p>When the `replaceAll` method is called with arguments _searchValue_ and _replaceValue_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be ? RequireObjectCoercible(*this* value).
+          1. If _searchValue_ is neither *undefined* nor *null*, then
+            1. Let _isRegExp_ be ? IsRegExp(_searchValue_).
+            1. If _isRegExp_ is true, then
```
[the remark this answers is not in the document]
jmdyck: (It looks kind of like, instead of adding asterisks to the `true` in S.p.replaceAll, you removed them from the `*true*` in S.p.matchAll, which is then what @ljharb commented on ~8 hours ago. So this one still needs to be fixed.)

spec.html
```diff
@@ -10440,6 +10440,27 @@ <h1>Static Semantics: CodePointAt ( _string_, _position_ )</h1>
...
+        1. If _searchValue_ is the empty String, and _fromIndex_ &le; _len_, return _fromIndex_.
+        1. Let _searchLen_ be the length of _searchValue_.
+        1. If there exists any integer _k_ such that _fromIndex_ &le; _k_ &le; _len_ - _searchLen_ and for all nonnegative integers _j_ less than _searchLen_, the code unit at index _k_ + _j_ within _string_ is the same as the code unit at index _j_ within _searchValue_, let _pos_ be the smallest (closest to *-&infin;*) such integer. Otherwise, let _pos_ be -1.
+        1. Return _pos_.
+      </emu-alg>
+      <emu-note>
+        <p>If _searchValue_ is empty and _fromIndex_ is less than or equal to the length of _string_, this algorithm returns _fromIndex_. An empty _searchValue_ is effectively found at every position within a string, including after the last code unit.</p>
+      </emu-note>
+      <emu-note>
+        <p>The above algorithm accepts out-of-bounds _fromIndex_ values (where _fromIndex_ &ge; the length of _string_).</p>
```
jmdyck, line 10460: I think this note would be more useful if it said what happened when `_fromIndex_` is out-of-bounds. But you can't just say it always returns -1, because it doesn't in the case covered by Note 1. It'd be simpler if this Note avoided talking about the "=" case, so something like:
> This algorithm always returns -1 if _fromIndex_ > the length of _string_.

spec.html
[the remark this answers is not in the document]
jmdyck: It appears to be a [known bug](https://github.community/t5/How-to-use-Git-and-GitHub/Cannot-apply-suggestion-in-PR-as-collaborator-with-Write-access/m-p/45917)

I considered making a PR against your branch, but the number of edits didn't seem quite big enough to warrant the bother.

spec.html
mathiasbynens: I like it! Done.

spec.html
```diff
@@ -10440,6 +10440,25 @@ <h1>Static Semantics: CodePointAt ( _string_, _position_ )</h1>
...
     </emu-clause>
 
+    <emu-clause id="sec-stringindexof">
+      <h1>Static Semantics: StringIndexOf ( _string_, _searchValue_, _fromIndex_ )</h1>
+      <p>The abstract operation _StringIndexOf_ performs the following steps:</p>
+      <!-- Note: OOB fromIndex values (fromIndex >= string.length) are accepted by the algorithm. -->
+      <emu-alg>
+        1. Assert: Type(_string_) is String.
+        1. Assert: Type(_searchValue_) is String.
+        1. Assert: _fromIndex_ is a nonnegative integer.
```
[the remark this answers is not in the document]
bakkot: This is technically OK as the spec is currently written, but see https://github.com/tc39/ecma262/issues/1964 and https://github.com/tc39/ecma262/pull/2007.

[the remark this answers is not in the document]
jmdyck: Yup, "nonnegative integer" is currently OK, but #1567 was a strong indication to use `IsNonNegativeInteger()` instead, at least in assertions.

spec.html
```diff
@@ -10440,6 +10440,27 @@ <h1>Static Semantics: CodePointAt ( _string_, _position_ )</h1>
...
+        1. If _searchValue_ is the empty String, and _fromIndex_ &le; _len_, return _fromIndex_.
+        1. Let _searchLen_ be the length of _searchValue_.
+        1. If there exists any integer _k_ such that _fromIndex_ &le; _k_ &le; _len_ - _searchLen_ and for all nonnegative integers _j_ less than _searchLen_, the code unit at index _k_ + _j_ within _string_ is the same as the code unit at index _j_ within _searchValue_, let _pos_ be the smallest (closest to *-&infin;*) such integer. Otherwise, let _pos_ be -1.
+        1. Return _pos_.
+      </emu-alg>
+      <emu-note>
+        <p>If _searchValue_ is empty and _fromIndex_ is less than or equal to the length of _string_, this algorithm returns _fromIndex_. An empty _searchValue_ is effectively found at every position within a string, including after the last code unit.</p>
+      </emu-note>
+      <emu-note>
+        <p>This algorithm always returns -1 if _fromIndex_ > the length of _string_.</p>
```
syg, line 10460: `&gt;` instead of >?

spec.html
```diff
@@ -32848,7 +32913,7 @@ <h1>RegExp.prototype [ @@replace ] ( _string_, _replaceValue_ )</h1>
           1. If Type(_rx_) is not Object, throw a *TypeError* exception.
           1. Let _S_ be ? ToString(_string_).
           1. Let _lengthS_ be the number of code unit elements in _S_.
-          1. Let _functionalReplace_ be IsCallable(_replaceValue_).
+          1. Let _functionalReplace_ be ! IsCallable(_replaceValue_).
```
syg, line 32916: I prefer this left as is, without `!`.

spec.html
```diff
@@ -10440,6 +10440,27 @@ <h1>Static Semantics: CodePointAt ( _string_, _position_ )</h1>
...
 
+    <emu-clause id="sec-stringindexof" aoid="StringIndexOf">
+      <h1>Static Semantics: StringIndexOf ( _string_, _searchValue_, _fromIndex_ )</h1>
+      <p>The abstract operation StringIndexOf takes arguments _string_ (a String), _searchValue_ (a String), and _fromIndex_ (a non-negative integer). It performs the following steps when called:</p>
+      <emu-alg>
+        1. Assert: Type(_string_) is String.
+        1. Assert: Type(_searchValue_) is String.
+        1. Assert: ! IsNonNegativeInteger(_fromIndex_) is *true*.
+        1. Let _len_ be the length of _string_.
+        1. If _searchValue_ is the empty String, and _fromIndex_ &le; _len_, return _fromIndex_.
```
michaelficarra, line 10451:
```suggestion
        1. If _searchValue_ is the empty String and _fromIndex_ &le; _len_, return _fromIndex_.
```

spec.html
```diff
@@ -30384,6 +30405,50 @@ <h1>Runtime Semantics: GetSubstitution ( _matched_, _str_, _position_, _captures
...
+            1. Set _replaceValue_ to ? ToString(_replaceValue_).
+          1. Let _searchLength_ be the length of _searchString_.
+          1. Let _advanceBy_ be max(1, _searchLength_).
+          1. Let _matchPositions_ be a new empty List.
+          1. Let _position_ be ! StringIndexOf(_string_, _searchString_, 0).
+          1. Repeat, while _position_ is not -1,
+            1. Append _position_ to the end of _matchPositions_.
+            1. Set _position_ to ! StringIndexOf(_string_, _searchString_, _position_ + _advanceBy_).
+          1. Let _endOfLastMatch_ be 0.
+          1. Let _result_ be the empty String value.
```
michaelficarra, line 30435:
```suggestion
          1. Let _result_ be the empty String.
```

I see there's another case of this in §21.2.5.10 RegExp.prototype[@@replace]. We should fix that one, too.

spec.html
ljharb: Until we have #1796, most every abstract operation should be using the prefix. Why would it not need the prefix?

spec.html
mathiasbynens: Happy to make the change, but note that it doesn't really matter. `foo > bar` is valid HTML (and valid ecmarkup).

spec.html
```diff
@@ -32869,7 +32934,7 @@ <h1>RegExp.prototype [ @@replace ] ( _string_, _replaceValue_ )</h1>
                   1. Let _thisIndex_ be ? ToLength(? Get(_rx_, *"lastIndex"*)).
                   1. Let _nextIndex_ be AdvanceStringIndex(_S_, _thisIndex_, _fullUnicode_).
                   1. Perform ? Set(_rx_, *"lastIndex"*, _nextIndex_, *true*).
-          1. Let _accumulatedResult_ be the empty String value.
+          1. Let _accumulatedResult_ be the empty String.
```
michaelficarra, line 32937: Oops. I meant as part of a separate PR. I'll send a PR real quick.

spec.html
ljharb: i suppose we do only have 5 of them prefixed at this point; although that's just an oversight from previous editorial direction, we can skip the prefix here for now

spec.html
```diff
@@ -10440,6 +10440,27 @@ <h1>Static Semantics: CodePointAt ( _string_, _position_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-stringindexof" aoid="StringIndexOf">
+      <h1>Static Semantics: StringIndexOf ( _string_, _searchValue_, _fromIndex_ )</h1>
```
bakkot, line 10444:
```suggestion
      <h1>Runtime Semantics: StringIndexOf ( _string_, _searchValue_, _fromIndex_ )</h1>
```

mathiasbynens: Thanks. I’ve made this change. Should we also move it to a different section now? Let me know if you have any pointers.

spec.html
```diff
@@ -30031,7 +30052,7 @@ <h1>String.prototype.indexOf ( _searchString_ [ , _position_ ] )</h1>
           1. Let _len_ be the length of _S_.
           1. Let _start_ be min(max(_pos_, 0), _len_).
           1. Let _searchLen_ be the length of _searchStr_.
-          1. Return the smallest possible integer _k_ not smaller than _start_ such that _k_ + _searchLen_ is not greater than _len_, and for all nonnegative integers _j_ less than _searchLen_, the code unit at index _k_ + _j_ within _S_ is the same as the code unit at index _j_ within _searchStr_; but if there is no such integer _k_, return the value -1.
+          1. Return ! StringIndexOf(_S_, _searchStr_, _start_).
```
gibson042, line 30055: It would be nice to see similar use of StringIndexOf for String.prototype.includes. String.prototype.lastIndexOf would be the final piece of the full closure, but that's sufficiently different that I can see leaving it out.

spec.html
```diff
@@ -30384,6 +30405,50 @@ <h1>Runtime Semantics: GetSubstitution ( _matched_, _str_, _position_, _captures
...
+          1. Let _endOfLastMatch_ be 0.
+          1. Let _result_ be the empty String.
+          1. For each _position_ in _matchPositions_, do
+            1. If _functionalReplace_ is *true*, then
+              1. Let _replacement_ be ? ToString(? Call(_replaceValue_, *undefined*, « _searchString_, _position_, _string_ »)).
+            1. Else,
+              1. Assert: Type(_replaceValue_) is String.
+              1. Let _captures_ be a new empty List.
+              1. Let _replacement_ be ! GetSubstitution(_searchString_, _string_, _position_, _captures_, *undefined*, _replaceValue_).
+            1. Let _stringSlice_ be the substring of _string_ consisting of the code units from _endOfLastMatch_ (inclusive) up to _position_ (exclusive).
```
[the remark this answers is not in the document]
michaelficarra: Note that the notation in #2021 is still in flux and no longer matches the suggestion.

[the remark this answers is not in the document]
mathiasbynens: What should I do here? Happy to make any changes to get this to a land-able state.

spec.html
```diff
@@ -30384,6 +30405,50 @@ <h1>Runtime Semantics: GetSubstitution ( _matched_, _str_, _position_, _captures
...
+            1. Set _result_ to the string-concatenation of _result_, _stringSlice_, and _replacement_.
+            1. Set _endOfLastMatch_ to _position_ + _searchLength_.
+          1. If _endOfLastMatch_ &lt; the length of _string_, then
+            1. Set _result_ to the string-concatenation of _result_ and the substring of _string_ consisting of the code units from _endOfLastMatch_ (inclusive) up through the final code unit of _string_ (inclusive).
```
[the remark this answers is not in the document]
mathiasbynens: Per the above comment, I'm not sure if I should make this change now. Maybe we can do it separately from this patch, once the discussion in #2021 is resolved?

spec.html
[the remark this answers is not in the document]
michaelficarra: If we're going to land it before #2021, it doesn't matter what we do here.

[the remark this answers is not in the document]
ljharb: Seems like this one will be able to land before #2021, but I think whichever one lands first should rebase and ensure this prose is consistent?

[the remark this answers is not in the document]
michaelficarra: @ljharb This PR is consistent with existing forms currently.

spec.html
mathiasbynens: How about this?

```diff
diff --git a/spec.html b/spec.html
index c10bf37..c305fcd 100644
--- a/spec.html
+++ b/spec.html
@@ -30024,7 +30024,9 @@ THH:mm:ss.sss
           1. Let _len_ be the length of _S_.
           1. Let _start_ be min(max(_pos_, 0), _len_).
           1. Let _searchLen_ be the length of _searchStr_.
-          1. If there exists any integer _k_ not smaller than _start_ such that _k_ + _searchLen_ is not greater than _len_, and for all nonnegative integers _j_ less than _searchLen_, the code unit at index _k_ + _j_ within _S_ is the same as the code unit at index _j_ within _searchStr_, return *true*; but if there is no such integer _k_, return *false*.
+          1. Let _index_ be ! StringIndexOf(_S_, _searchStr_, _start_).
+          1. If _index_ &ge; 0, return *true*.
+          1. Return *false*.
         </emu-alg>
         <emu-note>
           <p>If _searchString_ appears as a substring of the result of converting this object to a String, at one or more indices that are greater than or equal to _position_, return *true*; otherwise, returns *false*. If _position_ is *undefined*, 0 is assumed, so as to search all of the String.</p>
```

spec.html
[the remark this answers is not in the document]
ljharb: Seems like it should stay that way then, until #2021 is merged.

spec.html
ljharb: Looks great to me!

mathiasbynens: Pushed up that change.

spec.html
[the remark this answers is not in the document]
gibson042: That seems fine. My suggested phrasing is the current predominant pattern, but it looks like everything will be simplified by #2021.

spec.html
```diff
@@ -30384,6 +30407,50 @@ <h1>Runtime Semantics: GetSubstitution ( _matched_, _str_, _position_, _captures
...
+          1. Let _string_ be ? ToString(_O_).
+          1. Let _searchString_ be ? ToString(_searchValue_).
+          1. Let _functionalReplace_ be IsCallable(_replaceValue_).
+          1. If _functionalReplace_ is *false*, then
+            1. Set _replaceValue_ to ? ToString(_replaceValue_).
+          1. Let _searchLength_ be the length of _searchString_.
+          1. Let _advanceBy_ be max(1, _searchLength_).
+          1. Let _matchPositions_ be a new empty List.
+          1. Let _position_ be ! StringIndexOf(_string_, _searchString_, 0).
+          1. Repeat, while _position_ is not -1,
```
gibson042, line 30433: Your test in String.prototype.includes is `If _index_ &ge; 0`. I don't have a preference for "≥ 0" vs. "is not -1", but I do have a preference for them to match.

mathiasbynens: I've made the change. This is a little awkward though:

```
If index is not -1, return true.
Return false.
```

In code you'd probably write it as:

```
If index is -1, return false.
Return true.
```

🤷

Issue #1863, opened 2020-02-01
editorial nit: having both `UTF16Encode` and `UTF16Encoding` is weird


bakkot:
Pulling this out of [#1552](https://github.com/tc39/ecma262/pull/1552#discussion_r373762709) now that it's merged:

I feel somewhat negatively about having both [`UTF16Encoding`](https://tc39.es/ecma262/#sec-utf16encoding) and [`UTF16Encode`](https://tc39.es/ecma262/#sec-utf16encode) (because the names are too similar). Unfortunately I don't have an obviously superior name to suggest. `UTF16EncodeText` instead of `UTF16Encoding`, perhaps?

jmdyck:
To me, `UTF16EncodeText` suggests the semantics of `UTF16Encode` (i.e., it operates on _text_, a sequence of Unicode code points), so would be a confusing rename for `UTF16Encoding`.

Instead, how about renaming `UTF16Encoding` as `UTF16EncodeCodePoint` (since it operates on a single code point)?

Note that any renaming of `UTF16Encoding` will likely not be as simple as a search-and-replace, because there are uses of the form "the UTF16Encoding of <some code point>", which will sound odd we rename it from a noun (gerund) to a verb. (Mind you, invoking an abstract operation via that syntax is unusual to begin with, so it would be nice to make those more conventional.)

bakkot:
> To me, `UTF16EncodeText` suggests the semantics of `UTF16Encode` (i.e., it operates on text, a sequence of Unicode code points), so would be a confusing rename for `UTF16Encoding`.

Sorry, I meant to suggest it as a name for `UTF16Encode`, yes.

> Instead, how about renaming `UTF16Encoding` as `UTF16EncodeCodePoint` (since it operates on a single code point)?

Works for me. I am happy with also changing the invocation style for the operation.

domenic:
/cc @annevk who I thought had raised this before.

In general in web browser software at least, "encode" in the context of text refers to going from code points to bytes. This isn't really what is done in these operations, which go from code points to code units.

The names I would suggest are UTF16CodePointToCodeUnits and UTF16CodePointsToCodeUnits.

Issue #828, opened 2017-02-27
Editorial: "sequence of code units" != "String" ?


jmdyck:
The spec defines SV, TV, and TRV to return a sequence of code units. But a String value is a sequence of code units. So why not define SV, TV, and TRV to return a String value?

Note that TemplateStrings appears to assume that TV and TRV do indeed return Strings. E.g., it says
```
Let _string_ be the TV of |TemplateMiddle|.
```
and not
```
Let _string_ be the String value whose code units are the TV of |TemplateMiddle|.
```

allenwb:
The intent was to distinguish between abstract sequences of code units and actual string values which are ECMAScript language runtime values. You should have to have the full semantics of ECMAScript strings available in order to lex/parse ES code or to validate static semantic rules.

From that perspective, the direct use of TV  sequences as string values like shown above is a spec. bug. The fix would be language like that shown as the alternative.

jmdyck:
> You should have to have the full semantics of ECMAScript strings available in order to lex/parse ES code or to validate static semantic rules.

(Presumably you mean "You *shouldn't*".) That seems like a plausible reason, except:
- The "abstract sequence of code units" approach is confined to the definitions of SV, TV, and TRV. Elsewhere, those sequences are 'converted' into String values, *and those values are used to validate static semantic rules*. E.g., StringValue performs such a conversion, and returns a String value, and it is used in Early Error rules and in other Static Semantics operations. So keeping ES Strings out of SV/TV/TRV doesn't keep them out of static semantics.
- What's the problem with having "the full semantics" of ES Strings available at 'static' time? Presumably there's no normative effect. I.e., if SV/TV/TRV were defined in terms of Strings, implementations wouldn't be required to do anything differently. Do you think it would mislead implementers into doing something incorrect or inefficient? Do you think it would be harder to understand? (I'm doubtful on both counts.)
- How are the semantics of ES Strings any different from the semantics of sequences of code units? It's just different terminology for the same (or isomorphic) data models. How is the distinction helpful?

bterlson:
I am also curious, it seems fine to me to just always use string values (as long as it's clear that implementations don't need to use actual JS strings, which I think it is, given the existing loosyness @jmdyck points out).

bakkot:
The current editor group talked about it and is weakly split. @michaelficarra and @syg mildly prefer the status quo, @ljharb mildly prefers getting rid of the distinction. Does anyone have a strong reason to prefer one or the other?

[...]

PR #2014, opened 2020-05-25
Editorial: Rename Unicode encoding/decoding operations

bakkot:
Fixes https://github.com/tc39/ecma262/issues/1863 following @domenic's [suggestion](https://github.com/tc39/ecma262/issues/1863#issuecomment-597291191) there. I've also renamed `UTF16DecodeSurrogatePair` to `UTF16SurrogatePairToCodePoint` and `UTF16DecodeString` to `UTF16CodeUnitsToCodePoints`. I'm open to other names; these seemed the clearest.

After just renaming, some of the wording became a little strange: "the UTF16CodePointToCodeUnits of a Unicode code point" reads more awkwardly to me than "the UTF16Encoding of a Unicode code point" did. So I rephrased those usages.

cc @jmdyck

EDIT: names [changed](https://github.com/tc39/ecma262/pull/2014#issuecomment-633823196).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -10684,9 +10684,9 @@
       <p>ECMAScript differs from the Java programming language in the behaviour of Unicode escape sequences. In a Java program, if the Unicode escape sequence `\\u000A`, for example, occurs within a single-line comment, it is interpreted as a line terminator (Unicode code point U+000A is LINE FEED (LF)) and therefore the next code point is not part of the comment. Similarly, if the Unicode escape sequence `\\u000A` occurs within a string literal in a Java program, it is likewise interpreted as a line terminator, which is not allowed within a string literal&mdash;one must write `\\n` instead of `\\u000A` to cause a LINE FEED (LF) to be part of the String value of a string literal. In an ECMAScript program, a Unicode escape sequence occurring within a comment is never interpreted and therefore cannot contribute to termination of the comment. Similarly, a Unicode escape sequence occurring within a string literal in an ECMAScript program always contributes to the literal and is never interpreted as a line terminator or as a code point that might terminate the string literal.</p>
     </emu-note>
 
-    <emu-clause id="sec-utf16encoding" aoid="UTF16Encoding">
-      <h1>Static Semantics: UTF16Encoding ( _cp_ )</h1>
-      <p>The abstract operation UTF16Encoding takes argument _cp_ (a numeric code point value). It performs the following steps when called:</p>
+    <emu-clause id="sec-codepointtoutf16codeunits" aoid="CodePointToUTF16CodeUnits" oldids="sec-utf16encoding">
+      <h1>Static Semantics: CodePointToUTF16CodeUnits ( _cp_ )</h1>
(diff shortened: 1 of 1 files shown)


jmdyck:
Hm. The names `UTF16CodePoint[s]ToCodeUnits` suggest that there's such a thing as a "UTF16 code point", which there isn't. Maybe `CodePoint[s]ToUTF16CodeUnits`.

bakkot:
Ah, I'd intended it to read as "the UTF-16 version of code point to code unit translation", but I can see the ambiguity there.

jmdyck:
(It's tempting to moreover suggest replacing `UTF16CodeUnits` with just `String`:
- `CodePointToString`
- `CodePointsToString`
- `StringToCodePoints`

but I haven't actually determined if that's a good suggestion. It does seem more readable.)

mathiasbynens:
Came here to say what @jmdyck already said: `UTF16CodePointToCodeUnits` → `CodePoint[s]ToUTF16CodeUnits`, at a minimum. That said, I like the suggestion to just use `String` even better.

bakkot:
I am concerned about saying `String` instead of `CodeUnits` because the spec currently makes a distinction between those: [StringValue](https://tc39.es/ecma262/#sec-string-literals-static-semantics-stringvalue) of StringLiteral, for example, says

> Return the String value whose code units are the [SV](https://tc39.es/ecma262/#sec-static-semantics-sv) of this StringLiteral.

---

Actually, I guess that means `UTF16CodePointsToCodeUnits` and `UTF16CodeUnitsToCodePoints` are bad names, because those _do_ return and take a String value respectively.


So, thoughts on

- [`UTF16Encoding`](https://tc39.es/ecma262/#sec-utf16encoding) -> `CodePointToUTF16CodeUnits`
- [`UTF16Encode`](https://tc39.es/ecma262/#sec-utf16encode) -> `CodePointsToString`
- [`UTF16DecodeSurrogatePair`](https://tc39.es/ecma262/#sec-utf16decodesurrogatepair) -> `UTF16SurrogatePairToCodePoint`
- [`UTF16DecodeString`](https://tc39.es/ecma262/#sec-utf16decodestring) -> `StringToCodePoints`

?

jmdyck:
> I am concerned about saying `String` instead of `CodeUnits` because the spec currently makes a distinction between those: [StringValue](https://tc39.es/ecma262/#sec-string-literals-static-semantics-stringvalue) of StringLiteral, for example, says
> 
> > Return the String value whose code units are the [SV](https://tc39.es/ecma262/#sec-static-semantics-sv) of this StringLiteral.

Yeah, but I think that's a pointless distinction. See Issue #828.

(I started a PR to dissolve that distinction, but something stopped me. I'll have another look.)

bakkot:
Force-pushed commits using the names above.

@jmdyck

> Yeah, but I think that's a pointless distinction.

Arguably, but if we do away with it we should do so holistically. We can rename `CodePointToUTF16CodeUnits` to `CodePointToString` at that point.

PR #2015, opened 2020-05-26
Editorial: deduplicate some work in atomics

bakkot:
Replaces https://github.com/tc39/ecma262/pull/835. This includes two of the three commits in that PR; the third, https://github.com/tc39/ecma262/pull/835/commits/8e56f87a235130a319a8adcc4b99bd4acc336aef, I am not inclined to take.

I'm neutral on the second commit, but included it for completeness.

(I opened a separate PR rather than just updating #835 because I wanted to drop https://github.com/tc39/ecma262/pull/835/commits/8e56f87a235130a319a8adcc4b99bd4acc336aef but didn't want to force-push it away.)

cc @anba @syg @lars-t-hansen

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37609,20 +37609,6 @@ THH:mm:ss.sss
         </emu-alg>
       </emu-clause>
 
-      <emu-clause id="sec-atomicload" aoid="AtomicLoad">
-        <h1>AtomicLoad ( _typedArray_, _index_ )</h1>
-        <p>The abstract operation AtomicLoad takes arguments _typedArray_ and _index_. It atomically loads a value and returns the loaded value. It performs the following steps when called:</p>
-        <emu-alg>
-          1. Let _buffer_ be ? ValidateIntegerTypedArray(_typedArray_).
-          1. Let _indexedPosition_ be ? ValidateAtomicAccess(_typedArray_, _index_).
-          1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
-          1. NOTE: The above check is not redundant with the check in ValidateIntegerTypedArray because the call to ValidateAtomicAccess on the preceding line can have arbitrary side effects, which could cause the buffer to become detached.
-          1. Let _arrayTypeName_ be _typedArray_.[[TypedArrayName]].
-          1. Let _elementType_ be the Element Type value in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -37838,33 +37841,15 @@ <h1>AtomicReadModifyWrite ( _typedArray_, _index_, _value_, _op_ )</h1>
...
           1. Otherwise, let _v_ be ? ToInteger(_value_).
-          1. Let _elementSize_ be the Element Size value specified in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
           1. Let _elementType_ be the Element Type value in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _arrayTypeName_.
-          1. Let _offset_ be _typedArray_.[[ByteOffset]].
-          1. Let _indexedPosition_ be (_i_ &times; _elementSize_) + _offset_.
           1. Return GetModifySetValueInBuffer(_buffer_, _indexedPosition_, _elementType_, _v_, _op_).
         </emu-alg>
       </emu-clause>
 
-      <emu-clause id="sec-atomicload" aoid="AtomicLoad">
```
jmdyck, line 37853: Removing this will un-define the id "sec-atomicload". Should it become an oldid somewhere?

bakkot: I think it's moderately unlikely that anyone will have linked it, and I think it'll be pretty easy to figure out what to look at instead of so, so I'm not too worried about it.

ljharb: I'd prefer to always keep oldids if there's a reasonable place to link them; crawlers will have linked it.

bakkot: There is no corresponding abstract operation anymore because this operation is getting removed entirely. `Atomics.load` is the closest thing, but it's a built-in function, not an AO, so it's a little weird to tag it.

I don't really care if crawlers will have linked it if no human is ever going to click on those links, and I don't really expect a human to ever click on a such a crawled link.

ljharb: URLs should all work forever :-) but if there's no sensible place to link it, then it sounds right not to use an oldid.

bakkot: Basically it comes down to whether or not we want to add it to `Atomics.load`, which does the same thing (prior to this PR its definition was just "call AtomicLoad")  but is a built-in function instead of an abstract operation. I'd lean towards no, but I don't feel strongly.

syg: Yeah why not we can add an oldid.

bakkot: Done.

bakkot:
Rebased. @syg would you take a quick look to make sure I correctly rebased on top of #1908?

syg (approved):
lgtm, thanks for rebasing

PR #2021, opened 2020-05-29
Editorial: Use consistent language for extracting substrings

gibson042:
- [x] "substring of _x_ consisting of…" → "String value equal to the substring of _x_ consisting of…"
- [x] "substring of _x_ that starts at…" → "String value equal to the substring of _x_ consisting of…"
- [x] "substring of _x_ from…" → "String value equal to the substring of _x_ consisting of…"
- [x] "sequence of code units of _x_ starting at…" → "String value equal to the substring of _x_ consisting of…"
- [x] "String value containing _N_ consecutive code units from _x_…" →"String value equal to the substring of _x_ consisting of…"
- [x] "code units from…" → "code units at indices…"
- [x] "trailing substring of _x_" → "trailing code units of _x_" ("substring" is not normatively defined)
- [x] "substring of _x_… up through the final code unit" → "trailing code units of _x_…" (and I'm considering converting these to [start, end) as well)
- [x] "String value whose length is…, containing code units from _x_…" → "String value equal to the substring of _x_ consisting of…" (String.prototype.substring)
- [x] "portion of _x_ between…" → "String value equal to the substring of _x_ consisting of…" (RegExpBuiltinExec)
- [x] "String value consisting of the first…" → "String value equal to the substring of…" (JSON.stringify)

Also up for consideration is defining and using "the <dfn>string-slice</dfn> of <var>S</var>...", similar to [string-concatenation](https://tc39.es/ecma262/#sec-ecmascript-language-types-string-type).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1016,6 +1016,7 @@
         <p>The rationale behind this design was to keep the implementation of Strings as simple and high-performing as possible. If ECMAScript source text is in Normalized Form C, string literals are guaranteed to also be normalized, as long as they do not contain any Unicode escape sequences.</p>
       </emu-note>
       <p>In this specification, the phrase "the <dfn id="string-concatenation">string-concatenation</dfn> of _A_, _B_, ..." (where each argument is a String value, a code unit, or a sequence of code units) denotes the String value whose sequence of code units is the concatenation of the code units (in order) of each of the arguments (in order).</p>
+      <p>The phrase "the <dfn id="substring">substring</dfn> of _S_ from _inclusiveStart_ to _exclusiveEnd_" (where _S_ is a String value or a sequence of code units and _inclusiveStart_ and _exclusiveEnd_ are integers) denotes the String value consisting of the consecutive code units of _S_ beginning at index _inclusiveStart_ and ending immediately before index _exclusiveEnd_ (which is the empty String when _inclusiveStart_ = _exclusiveEnd_). If the "to" suffix is omitted, the length of _S_ is used as the value of _exclusiveEnd_.</p>
 
       <emu-clause id="sec-stringindexof" aoid="StringIndexOf">
         <h1>Runtime Semantics: StringIndexOf ( _string_, _searchValue_, _fromIndex_ )</h1>
@@ -25547,7 +25548,8 @@
           1. If the length of _S_ is at least 2 and the first two code units of _S_ are either *"0x"* or *"0X"*, then
             1. Remove the first two code units from _S_.
             1. Set _R_ to 16.
-        1. If _S_ contains a code unit that is not a radix-_R_ digit, let _Z_ be the substring of _S_ consisting of all code units before the first such code unit; otherwise, let _Z_ be _S_.
+        1. If _S_ contains a code unit that is not a radix-_R_ digit, let _end_ be the index within _S_ of the first such code unit; otherwise, let _end_ be the length of _S_.
+        1. Let _Z_ be the substring of _S_ from 0 to _end_.
         1. If _Z_ is empty, return *NaN*.
         1. Let _mathInt_ be the mathematical integer value that is represented by _Z_ in radix-_R_ notation, using the letters <b>A</b>-<b>Z</b> and <b>a</b>-<b>z</b> for digits with values 10 through 35. (However, if _R_ is 10 and _Z_ contains more than 20 significant digits, every significant digit after the 20th may be replaced by a 0 digit, at the option of the implementation; and if _R_ is not 2, 4, 8, 10, 16, or 32, then _mathInt_ may be an implementation-approximated value representing the mathematical integer value that is represented by _Z_ in radix-_R_ notation.)
         1. If _mathInt_ = 0<sub>ℝ</sub>, then
@@ -25662,7 +25664,7 @@
                   1. If _C_ is not in _reservedSet_, then
                     1. Let _S_ be the String value containing only the code unit _C_.
                   1. Else,
-                    1. Let _S_ be the substring of _string_ from index _start_ to index _k_ inclusive.
+                    1. Let _S_ be the substring of _string_ from _start_ to _k_ + 1.
                 1. Else,
                   1. Assert: the most significant bit in _B_ is 1.
                   1. Let _n_ be the smallest nonnegative integer such that (_B_ &lt;&lt; _n_) &amp; 0x80 is equal to 0.
@@ -30506,10 +30508,11 @@ THH:mm:ss.sss
           1. If _endPosition_ is *undefined*, let _pos_ be _len_; else let _pos_ be ? ToInteger(_endPosition_).
           1. Let _end_ be min(max(_pos_, 0), _len_).
           1. Let _searchLength_ be the length of _searchStr_.
+          1. If _searchLength_ = 0, return *true*.
           1. Let _start_ be _end_ - _searchLength_.
-          1. If _start_ is less than 0, return *false*.
-          1. If the sequence of code units of _S_ starting at _start_ of length _searchLength_ is the same as the full code unit sequence of _searchStr_, return *true*.
-          1. Otherwise, return *false*.
+          1. If _start_ &lt; 0, return *false*.
+          1. Let _substring_ be the substring of _S_ from _start_ to _end_.
+          1. Return ! SameValueNonNumeric(_substring_, _searchStr_).
         </emu-alg>
         <emu-note>
           <p>Returns *true* if the sequence of code units of _searchString_ converted to a String is the same as the corresponding code units of this object (converted to a String) starting at _endPosition_ - length(this). Otherwise returns *false*.</p>
@@ -30541,7 +30544,7 @@ THH:mm:ss.sss
           1. Return *false*.
         </emu-alg>
         <emu-note>
-          <p>If _searchString_ appears as a substring of the result of converting this object to a String, at one or more indices that are greater than or equal to _position_, return *true*; otherwise, returns *false*. If _position_ is *undefined*, 0 is assumed, so as to search all of the String.</p>
+          <p>If _searchString_ appears as a <emu-not-ref>substring</emu-not-ref> of the result of converting this object to a String, at one or more indices that are greater than or equal to _position_, return *true*; otherwise, returns *false*. If _position_ is *undefined*, 0 is assumed, so as to search all of the String.</p>
         </emu-note>
         <emu-note>
           <p>Throwing an exception if the first argument is a RegExp is specified in order to allow future editions to define extensions that allow such argument values.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -25062,7 +25062,7 @@ <h1>parseInt ( _string_, _radix_ )</h1>
           1. If the length of _S_ is at least 2 and the first two code units of _S_ are either *"0x"* or *"0X"*, then
             1. Remove the first two code units from _S_.
             1. Set _R_ to 16.
-        1. If _S_ contains a code unit that is not a radix-_R_ digit, let _Z_ be the substring of _S_ consisting of all code units before the first such code unit; otherwise, let _Z_ be _S_.
+        1. If _S_ contains a code unit that is not a radix-_R_ digit, let _Z_ be the String value equal to the substring of _S_ consisting of all code units before the first such code unit; otherwise, let _Z_ be _S_.
```
michaelficarra, line 25065: Why is this phrasing preferable? I like it how it was.

jmdyck:
Note there's also:
- 2 x `the sequence of code units of _S_ starting at _start_ of length _searchLength_` (in String.prototype.{ends,starts}With)
- `the String value whose length is _to_ - _from_, containing code units from _S_, namely the code units with indices _from_ through _to_ - 1, in ascending order` (in String.prototype.substring)
- `the portion of _S_ between offset _lastIndex_ inclusive and offset _e_ exclusive` (in RegExpBuiltinExec)
- `the String value consisting of the first 10 code units of _space_` (in JSON.stringify)

(and possibly others as this is based on an old search).

spec.html
gibson042: Mostly because "the String value equal to the substring of <var>x</var> consisting of…" is the predominant phrasing in the spec, but a little bit because it is more correct. Alternatively, we could define "<dfn>the substring of</dfn> <var>S</var>..." similarly to how [string-concatenation](https://tc39.es/ecma262/#sec-ecmascript-language-types-string-type) is defined and then use the shorter phrasing.

gibson042:
Thanks @jmdyck, I will update to cover those as well.

spec.html
```diff
@@ -29929,10 +29929,11 @@ <h1>String.prototype.endsWith ( _searchString_ [ , _endPosition_ ] )</h1>
...
           1. Let _end_ be min(max(_pos_, 0), _len_).
           1. Let _searchLength_ be the length of _searchStr_.
+          1. If _searchLength_ is 0, return *true*.
           1. Let _start_ be _end_ - _searchLength_.
-          1. If _start_ is less than 0, return *false*.
-          1. If the sequence of code units of _S_ starting at _start_ of length _searchLength_ is the same as the full code unit sequence of _searchStr_, return *true*.
-          1. Otherwise, return *false*.
+          1. If _start_ &lt; 0, return *false*.
+          1. Let _substring_ be the String value equal to the substring of _S_ consisting of the code units at indices _start_ (inclusive) through _end_ (exclusive).
+          1. Return ! SameValueNonNumeric(_substring_, _searchStr_).
```
gibson042, line 29936: How do people feel about this? It seemed more natural than the duplication, and more parsimonious than defining and using something like "<dfn>string-equals</dfn>".

spec.html
michaelficarra: I would prefer that, yes.

gibson042: Done; introduced as string-slice for targeted auto-linking.

spec.html
```diff
@@ -982,6 +982,7 @@ <h1>The String Type</h1>
         <p>The rationale behind this design was to keep the implementation of Strings as simple and high-performing as possible. If ECMAScript source text is in Normalized Form C, string literals are guaranteed to also be normalized, as long as they do not contain any Unicode escape sequences.</p>
       </emu-note>
       <p>In this specification, the phrase "the <dfn>string-concatenation</dfn> of _A_, _B_, ..." (where each argument is a String value, a code unit, or a sequence of code units) denotes the String value whose sequence of code units is the concatenation of the code units (in order) of each of the arguments (in order).</p>
+      <p>The phrase "the <dfn id="string-slice">string-slice</dfn> of _S_ from _inclusiveStart_ to _exclusiveEnd_" (where _S_ is a String value or a sequence of code units and _inclusiveStart_ and _exclusiveEnd_ are integers) denotes the String value consisting of the consecutive code units of _S_ beginning at index _inclusiveStart_ and ending immediately before index _exclusiveEnd_ (which is the empty String when _inclusiveStart_ = _exclusiveEnd_). If the "to" suffix is omitted, the length of _S_ is used as the value of _exclusiveEnd_.</p>
```
ljharb, line 985: why use prose for this instead of an abstract operation?

michaelficarra: I kind of prefer this prose. We could make string concatenation an AO as well, but for something so straightforward, it seems like overkill.

gibson042: Just following suit with string-concatenation and (to a lesser extent) [mathematical operations](https://tc39.es/ecma262/#sec-mathematical-operations) like abs/min/max/modulo. Do you feel strongly about it being a more formal abstract operation instead? And if so, where would it go? A new subsection under [7 Abstract Operations](https://tc39.es/ecma262/#sec-abstract-operations)?

jmdyck: (I considered making string-concatenation an AO back when I introduced it. One reason I didn't was probably that it was unclear how to allow the bullet-list invocations (e.g., in [Number::toString](https://tc39.es/ecma262/#sec-numeric-types-number-tostring)) as an alternative to the standard prefix-parenthesized format.)

ljharb: String concatenation seems like a simple operation to me; slicing seems more complex and thus would benefit from being an AO.

If we made an AO, I'd expect it would indeed have to be a 7.5, or defined here, under the String Type.

@bakkot @syg do you have any thoughts about prose vs an AO for this?

spec.html
```diff
@@ -30489,7 +30493,8 @@ <h1>String.prototype.substring ( _start_, _end_ )</h1>
           1. Let _finalEnd_ be min(max(_intEnd_, 0), _len_).
           1. Let _from_ be min(_finalStart_, _finalEnd_).
           1. Let _to_ be max(_finalStart_, _finalEnd_).
-          1. Return the String value whose length is _to_ - _from_, containing code units from _S_, namely the code units with indices _from_ through _to_ - 1, in ascending order.
+          1. If _from_ = _to_, return the empty String.
```
bakkot, line 30496: Why include this step?

spec.html
bakkot: I'm happier with prose, personally.

spec.html
gibson042: I think it made more sense in an earlier iteration; removing.

gibson042:
@bakkot You might be overestimating the novelty of the term; "slice" is already the name of the canonical function in this language and is standard vocabulary in [Python](https://docs.python.org/3/reference/expressions.html#slicings), [Pandas](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.str.slice.html), [Ruby](https://ruby-doc.org/core-2.7.1/String.html#method-i-slice), [Go](https://golang.org/ref/spec#Slice_types), etc. It's good for the operation to have a unique name within the spec, for autolinking but also for more primitive searches—and I think "string-slice" in particular has a nice correspondence with "string-concatenation".

bakkot:
@gibson042 "slice" is a common enough term, but "string-slice" is not a concept I see in common use elsewhere. Since "substring" is the common term for this operation, I (and other editors) still prefer to define "substring", and annotate the few places where the usage doesn't accord with this definition. Are you OK with that?

gibson042:
Works for me, then.

gibson042:
Updated.

spec.html
```diff
@@ -31111,7 +31120,7 @@ <h1>String.prototype.substring ( _start_, _end_ )</h1>
           1. Let _finalEnd_ be min(max(_intEnd_, 0), _len_).
           1. Let _from_ be min(_finalStart_, _finalEnd_).
           1. Let _to_ be max(_finalStart_, _finalEnd_).
-          1. Return the String value whose length is _to_ - _from_, containing code units from _S_, namely the code units with indices _from_ through _to_ - 1, in ascending order.
+          1. Return the substring of _S_ from  _from_ to _to_.
```
michaelficarra, line 31123:
```suggestion
          1. Return the substring of _S_ from _from_ to _to_.
```

spec.html
```diff
@@ -30753,19 +30756,23 @@ <h1>String.prototype.replace ( _searchValue_, _replaceValue_ )</h1>
...
           1. Else,
             1. Let _captures_ be a new empty List.
-            1. Let _replStr_ be ! GetSubstitution(_matched_, _string_, _pos_, _captures_, *undefined*, _replaceValue_).
-          1. Let _tailPos_ be _pos_ + the number of code units in _matched_.
-          1. Let _newString_ be the string-concatenation of the first _pos_ code units of _string_, _replStr_, and the trailing substring of _string_ starting at index _tailPos_. If _pos_ is 0, the first element of the concatenation will be the empty String.
-          1. Return _newString_.
+            1. Let _replStr_ be ! GetSubstitution(_searchString_, _string_, _pos_, _captures_, *undefined*, _replaceValue_).
+          1. Let _tailPos_ be _pos_ + _searchLen_.
+          1. Return the string-concatenation of:
+            * the substring of _string_ from 0 to _pos_
```
michaelficarra, line 30773: I don't think we need to create a list here for this. There's not a huge number of items (3) and they're not that long. Compare with this step below:

>  1. Set _accumulatedResult_ to the string-concatenation of the current value of _accumulatedResult_, the substring of _S_ from _nextSourcePosition_ to _position_, and _replacement_.

spec.html
```diff
@@ -33493,10 +33503,10 @@ <h1>RegExp.prototype [ @@replace ] ( _string_, _replaceValue_ )</h1>
               1. Let _replacement_ be ? GetSubstitution(_matched_, _S_, _position_, _captures_, _namedCaptures_, _replaceValue_).
             1. If _position_ &ge; _nextSourcePosition_, then
               1. NOTE: _position_ should not normally move backwards. If it does, it is an indication of an ill-behaving RegExp subclass or use of an access triggered side-effect to change the global flag or other characteristics of _rx_. In such cases, the corresponding substitution is ignored.
-              1. Set _accumulatedResult_ to the string-concatenation of the current value of _accumulatedResult_, the substring of _S_ consisting of the code units from _nextSourcePosition_ (inclusive) up to _position_ (exclusive), and _replacement_.
+              1. Set _accumulatedResult_ to the string-concatenation of the current value of _accumulatedResult_, the substring of _S_ from _nextSourcePosition_ to _position_, and _replacement_.
```
michaelficarra, line 33506: This is the only use of "the current value of".

```suggestion
              1. Set _accumulatedResult_ to the string-concatenation of _accumulatedResult_, the substring of _S_ from _nextSourcePosition_ to _position_, and _replacement_.
```

spec.html
```diff
@@ -33543,8 +33553,8 @@ <h1>get RegExp.prototype.source</h1>
...
+          <p>Returns an Array object into which substrings of the result of converting _string_ to a String have been stored. The substrings are determined by searching from left to right for matches of the *this* value regular expression; these occurrences are not part of any string in the returned array, but serve to divide up the String value.</p>
```
michaelficarra, line 33556: Should we capitalise String here?

```suggestion
          <p>Returns an Array object into which substrings of the result of converting _string_ to a String have been stored. The substrings are determined by searching from left to right for matches of the *this* value regular expression; these occurrences are not part of any String in the returned array, but serve to divide up the String value.</p>
```

ljharb:
I've updated this to resolve all of the comments except https://github.com/tc39/ecma262/pull/2021#pullrequestreview-468881785 - @gibson042, thoughts? lmk if you want to update based on that comment, or if i should review as-is.

gibson042:
I think it's good as-is... the links suppressed by \<emu-not-ref> would not be harmful, but they would be superfluous.

bakkot:
Works for me.

PR #2026, opened 2020-05-31
Editorial: Add <dfn>s for native error constructors

ExE-Boss:
This makes auto‑linking work.

---

[Preview](https://ci.tc39.es/preview/tc39/ecma262/pull/2026) ([#sec-well-known-intrinsic-objects](https://ci.tc39.es/preview/tc39/ecma262/pull/2026#sec-well-known-intrinsic-objects))

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -26927,31 +26927,37 @@
 
       <emu-clause id="sec-native-error-types-used-in-this-standard-evalerror">
         <h1>EvalError</h1>
+        <p>The EvalError constructor is the intrinsic object <dfn>%EvalError%</dfn>.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -26924,31 +26924,37 @@ <h1>Native Error Types Used in This Standard</h1>
 
       <emu-clause id="sec-native-error-types-used-in-this-standard-evalerror">
         <h1>EvalError</h1>
+        <p>Is the intrinsic object <dfn>%EvalError%</dfn>.</p>
```
michaelficarra, line 26927: We don't typically do this as a sentence/paragraph on its own. We do it in an unordered list. We could either switch to an unordered list here or (my preference) make this a complete sentence and tack it on to the following paragraph.

PR #2027, opened 2020-05-31
Editorial: Add specific id for string-concatenation

gibson042:
Fixes suboptimal linking. Compare the latest draft to the preview build of this PR.
* Draft: string-concatenation links (e.g., from https://tc39.es/ecma262/#sec-utf16encode) point at #sec-ecmascript-language-types-string-type (6.1.4 The String Type).
* This PR: string-concatenation links (e.g., from https://ci.tc39.es/preview/tc39/ecma262/sha/e80669a328d044fed714e6975844ce1bdc4b2e9f/#sec-utf16encode) point at #string-concatenation (the specific definition of "string-concatenation").

PR #2029, opened 2020-05-31
Editorial: Simplify spec use of bitwise operations

gibson042:


spec.html
```diff
@@ -1679,7 +1679,7 @@ <h1>Number::leftShift ( _x_, _y_ )</h1>
           <emu-alg>
             1. Let _lnum_ be ! ToInt32(_x_).
             1. Let _rnum_ be ! ToUint32(_y_).
-            1. Let _shiftCount_ be the result of masking out all but the least significant 5 bits of _rnum_, that is, compute _rnum_ &amp; 0x1F.
+            1. Let _shiftCount_ be _rnum_ modulo 32.
```
ljharb, line 1682: `modulo` is an abstract operation
```suggestion
            1. Let _shiftCount_ be modulo(_rnum_, 32).
```

spec.html
```diff
@@ -1690,7 +1690,7 @@ <h1>Number::signedRightShift ( _x_, _y_ )</h1>
...
```
ljharb, line 1693:
```suggestion
            1. Let _shiftCount_ be modulo(_rnum_, 32).
```

spec.html
```diff
@@ -1701,7 +1701,7 @@ <h1>Number::unsignedRightShift ( _x_, _y_ )</h1>
           <emu-alg>
             1. Let _lnum_ be ! ToUint32(_x_).
             1. Let _rnum_ be ! ToUint32(_y_).
-            1. Let _shiftCount_ be the result of masking out all but the least significant 5 bits of _rnum_, that is, compute _rnum_ &amp; 0x1F.
+            1. Let _shiftCount_ be _rnum_ modulo 32.
```
ljharb, line 1704:
```suggestion
            1. Let _shiftCount_ be modulo(_rnum_, 32).
```

spec.html
gibson042: Is that incoming? There are currently 35 uses of modulo, all of which look like this ([example](https://tc39.es/ecma262/#sec-bigintbitwiseop)).

ljharb: oh oops, you're totally right; disregard this.

spec.html
```diff
@@ -25247,7 +25246,7 @@ <h1>Runtime Semantics: Decode ( _string_, _reservedSet_ )</h1>
                     1. If the code unit at index _k_ within _string_ is not the code unit 0x0025 (PERCENT SIGN), throw a *URIError* exception.
                     1. If the code units at index (_k_ + 1) and (_k_ + 2) within _string_ do not represent hexadecimal digits, throw a *URIError* exception.
                     1. Let _B_ be the 8-bit value represented by the two hexadecimal digits at index (_k_ + 1) and (_k_ + 2).
-                    1. If the two most significant bits in _B_ are not 10, throw a *URIError* exception.
+                    1. If the two most significant bits of _B_ ordered by descending significance are not 1 and 0, throw a *URIError* exception.
```
michaelficarra, line 25249:
```suggestion
                    1. If the two most significant bits of _B_ in descending significance are not 1 and 0 respectively, throw a *URIError* exception.
```

spec.html
```diff
@@ -25228,17 +25228,16 @@ <h1>Runtime Semantics: Decode ( _string_, _reservedSet_ )</h1>
...
                   1. Else,
-                    1. Let _S_ be the substring of _string_ from index _start_ to index _k_ inclusive.
+                    1. Let _S_ be the String value equal to the substring of _string_ consisting of the code units at indices _start_ (inclusive) through _k_ (inclusive).
                 1. Else,
-                  1. Assert: the most significant bit in _B_ is 1.
-                  1. Let _n_ be the smallest nonnegative integer such that (_B_ &lt;&lt; _n_) &amp; 0x80 is equal to 0.
-                  1. If _n_ equals 1 or _n_ is greater than 4, throw a *URIError* exception.
-                  1. Let _Octets_ be a List of 8-bit integers of size _n_.
+                  1. If _n_ = 1 or _n_ &gt; 4, throw a *URIError* exception.
+                  1. Let _Octets_ be a List of 8-bit integers of length _n_.
```
michaelficarra, line 25240: Spec lists can't have holes, so what are the values in this list? Is it even useful to initially size/populate this list?

gibson042: Good point; refactored to avoid that.

spec.html
```diff
@@ -25228,29 +25228,28 @@ <h1>Runtime Semantics: Decode ( _string_, _reservedSet_ )</h1>
...
                 1. Set _k_ to _k_ + 2.
-                1. If the most significant bit in _B_ is 0, then
+                1. Let _n_ be the number of leading 1 bits in _B_.
+                1. If _n_ = 0, then
                   1. Let _C_ be the code unit whose value is _B_.
                   1. If _C_ is not in _reservedSet_, then
                     1. Let _S_ be the String value containing only the code unit _C_.
                   1. Else,
-                    1. Let _S_ be the substring of _string_ from index _start_ to index _k_ inclusive.
+                    1. Let _S_ be the String value equal to the substring of _string_ consisting of the code units at indices _start_ (inclusive) through _k_ (inclusive).
```
michaelficarra, line 25237: Might want to update this to match https://github.com/tc39/ecma262/pull/2027.

gibson042: I'm ready to resolve the conflicts for whichever of those PRs loses the race.

spec.html
```diff
@@ -25659,29 +25659,28 @@ <h1>Runtime Semantics: Decode ( _string_, _reservedSet_ )</h1>
...
                   1. If _k_ + (3 &times; (_n_ - 1)) is greater than or equal to _strLen_, throw a *URIError* exception.
+                  1. Let _Octets_ be a new List containing _B_.
                   1. Let _j_ be 1.
                   1. Repeat, while _j_ &lt; _n_,
                     1. Set _k_ to _k_ + 1.
                     1. If the code unit at index _k_ within _string_ is not the code unit 0x0025 (PERCENT SIGN), throw a *URIError* exception.
                     1. If the code units at index (_k_ + 1) and (_k_ + 2) within _string_ do not represent hexadecimal digits, throw a *URIError* exception.
                     1. Let _B_ be the 8-bit value represented by the two hexadecimal digits at index (_k_ + 1) and (_k_ + 2).
-                    1. If the two most significant bits in _B_ are not 10, throw a *URIError* exception.
+                    1. If the two most significant bits of _B_ in descending significance are not 1 and 0 respectively, throw a *URIError* exception.
```
bakkot, line 25679:
```suggestion
                    1. If the two most significant bits of _B_ in order of descending significance are not 1 and 0 respectively, throw a *URIError* exception.
```

bakkot: I see @michaelficarra suggested the current wording; I'll try to resolve this definitively at today's editor call.

michaelficarra: WFM

ljharb: i lightly prefer @bakkot's suggestion

gibson042: Accepted. It also occurs to me that this step could be removed completely, because there's a later check that throws a URIError if the final value of _Octets_ "does not contain a valid UTF-8 encoding of a Unicode code point", which is just as true for **« 0xC0, 0x00 »** as it is for **« 0xC0, 0x80 »**.

michaelficarra: @gibson042 sure, that sounds even better.

gibson042: Done.

spec.html
```diff
@@ -25659,29 +25659,27 @@ <h1>Runtime Semantics: Decode ( _string_, _reservedSet_ )</h1>
...
                     1. Let _S_ be the substring of _string_ from _start_ to _k_ + 1.
                 1. Else,
-                  1. Assert: the most significant bit in _B_ is 1.
-                  1. Let _n_ be the smallest nonnegative integer such that (_B_ &lt;&lt; _n_) &amp; 0x80 is equal to 0.
-                  1. If _n_ equals 1 or _n_ is greater than 4, throw a *URIError* exception.
-                  1. Let _Octets_ be a List of 8-bit integers of size _n_.
-                  1. Set _Octets_[0] to _B_.
+                  1. If _n_ = 1 or _n_ &gt; 4, throw a *URIError* exception.
                   1. If _k_ + (3 &times; (_n_ - 1)) is greater than or equal to _strLen_, throw a *URIError* exception.
+                  1. Let _Octets_ be a new List containing _B_.
```
ljharb, line 25672:
```suggestion
                  1. Let _Octets_ be &laquo; _B_ &raquo;.
```
but it's fine either way

PR #2030, opened 2020-06-02
Normative: Add Logical Assignment Operators

jridgewell:
https://github.com/tc39/proposal-logical-assignment

This includes a few new changes:
- `NamedEvaluation` is used, given today's tentative consensus
- `LogicalAssignmentOperator` has been removed, and instead each `@@=` operator is defined individually

- - -

<details>

<summary>Outdated discussion</summary>

I need help getting it to compile correctly, though. When I run `build`, I get:

```
Could not find a production matching RHS in syntax-directed operation (undefined-nonterminal)
```

This is likely from defining the grammar as `LeftHandSideExpression ShortCircuitAssignmentOperator AssignmentExpression`, but defining individual algorithms as ```AssignmentExpression : LeftHandSideExpression `||=` AssignmentExpression```. We could define each grammar individually, or do a bit of metaprogramming for the algorithms, like `AssignmentOperator` did. I just don't know how to metaprogram in spec-ese.

</details>

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -11145,6 +11145,7 @@
         `&amp;&amp;` `||` `??`
         `?` `:`
         `=` `+=` `-=` `*=` `%=` `**=` `&lt;&lt;=` `&gt;&gt;=` `&gt;&gt;&gt;=` `&amp;=` `|=` `^=`
+        `&amp;&amp;=` `||=` `??=`
         `=&gt;`
 
       DivPunctuator ::
@@ -15332,6 +15333,9 @@
         AsyncArrowFunction[?In, ?Yield, ?Await]
         LeftHandSideExpression[?Yield, ?Await] `=` AssignmentExpression[?In, ?Yield, ?Await] #assignment
         LeftHandSideExpression[?Yield, ?Await] AssignmentOperator AssignmentExpression[?In, ?Yield, ?Await]
+        LeftHandSideExpression[?Yield, ?Await] `&amp;&amp;=` AssignmentExpression[?In, ?Yield, ?Await]
+        LeftHandSideExpression[?Yield, ?Await] `||=` AssignmentExpression[?In, ?Yield, ?Await]
+        LeftHandSideExpression[?Yield, ?Await] `??=` AssignmentExpression[?In, ?Yield, ?Await]
 
       AssignmentOperator : one of
         `*=` `/=` `%=` `+=` `-=` `&lt;&lt;=` `&gt;&gt;=` `&gt;&gt;&gt;=` `&amp;=` `^=` `|=` `**=`
@@ -15355,7 +15359,13 @@
           It is a Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
         </li>
       </ul>
-      <emu-grammar>AssignmentExpression : LeftHandSideExpression AssignmentOperator AssignmentExpression</emu-grammar>
+      <emu-grammar>
+        AssignmentExpression :
+          LeftHandSideExpression AssignmentOperator AssignmentExpression
+          LeftHandSideExpression `&amp;&amp;=` AssignmentExpression
+          LeftHandSideExpression `||=` AssignmentExpression
+          LeftHandSideExpression `??=` AssignmentExpression
+      </emu-grammar>
       <ul>
         <li>
           It is a Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
@@ -15379,6 +15389,9 @@
           YieldExpression
           LeftHandSideExpression `=` AssignmentExpression
           LeftHandSideExpression AssignmentOperator AssignmentExpression
+          LeftHandSideExpression `&amp;&amp;=` AssignmentExpression
+          LeftHandSideExpression `||=` AssignmentExpression
+          LeftHandSideExpression `??=` AssignmentExpression
       </emu-grammar>
       <emu-alg>
         1. Return *false*.
@@ -15395,6 +15408,9 @@
           AsyncArrowFunction
           LeftHandSideExpression `=` AssignmentExpression
           LeftHandSideExpression AssignmentOperator AssignmentExpression
+          LeftHandSideExpression `&amp;&amp;=` AssignmentExpression
+          LeftHandSideExpression `||=` AssignmentExpression
+          LeftHandSideExpression `??=` AssignmentExpression
       </emu-grammar>
       <emu-alg>
         1. Return ~invalid~.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -15274,6 +15279,12 @@ <h1>Static Semantics: Early Errors</h1>
           It is a Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
         </li>
       </ul>
+      <emu-grammar>AssignmentExpression : LeftHandSideExpression ShortCircuitAssignmentOperator AssignmentExpression</emu-grammar>
```
bakkot, line 15282: Instead of adding a new section here, you can just modify the preceding early error so it lists this production as well. That's the [existing pattern](https://tc39.es/ecma262/#sec-continue-statement-static-semantics-early-errors).

bakkot:
> This is likely from defining the grammar as `LeftHandSideExpression ShortCircuitAssignmentOperator AssignmentExpression`, but defining individual algorithms as `AssignmentExpression : LeftHandSideExpression `||=` AssignmentExpression`.

That is indeed the reason! That's the linter doing its job. I'm glad the error message was clear enough for you to figure it out. (We're all going to be beta-testing it for a bit.) Apologies for not catching this during the review.

> We could define each grammar individually, or do a bit of metaprogramming for the algorithms, like `AssignmentOperator` did. I just don't know how to metaprogram in spec-ese.

In this instance I think it will be clearest to just do away with `ShortCircuitAssignmentOperator` and instead add each of the three cases individually to `AssignmentExpression`.

jmdyck:
> In this instance I think it will be clearest to just do away with `ShortCircuitAssignmentOperator` and instead add each of the three cases individually to `AssignmentExpression`.

With the existing production, you could do something like:
```
<emu-grammar>AssignmentExpression : LeftHandSideExpression ShortCircuitAssignmentOperator AssignmentExpression</emu-grammar>
<emu-alg>
  1. Let _lref_ be the result of evaluating |LeftHandSideExpression|.
  1. Let _lval_ be ? GetValue(_lref_).
  1. Let _opText_ be the source text matched by |ShortCircuitAssignmentOperator|.
  1. If _opText_ is `??=`, then
    1. If _lval_ is neither *undefined* nor *null*, return _lval_.
  1. Else,
    1. Let _lbool_ be ! ToBoolean(_lval_).
    1. If _opText_ is `||=`, then
      1. If _lbool_ is *true*, return _lval_.
    1. Else,
      1. Assert: _opText_ is `&amp;&amp=`.
      1. If _lbool_ is *false*, return _lval_.
  1. If IsAnonymousFunctionDefinition(|AssignmentExpression|) and IsIdentifierRef of |LeftHandSideExpression| are both *true*, then
    1. Let _rval_ be NamedEvaluation of |AssignmentExpression| with argument GetReferencedName(_lref_).
  1. Else,
    1. Let _rref_ be the result of evaluating |AssignmentExpression|.
  1. Perform ? PutValue(_lref_, _rval_).
  1. Return _rval_.
</emu-alg>
```
(I started it before @bakkot's comment and didn't want to just throw it away.)

jmdyck:
If you do keep ShortCircuitAssignmentOperator, you'll need to reference it from Annex A.

And in any event, HasCallInTailPosition will need to be defined for the new right-hand-side(s).

spec.html
jridgewell: Done.

jridgewell:
> If you do keep ShortCircuitAssignmentOperator, you'll need to reference it from Annex A.

It's removed now, so I think this is done.

> And in any event, HasCallInTailPosition will need to be defined for the new right-hand-side(s).

Whoops, not sure how I missed that. Updated.

spec.html
```diff
@@ -15268,7 +15272,13 @@ <h1>Static Semantics: Early Errors</h1>
           It is a Syntax Error if AssignmentTargetType of |LeftHandSideExpression| is not ~simple~.
         </li>
       </ul>
-      <emu-grammar>AssignmentExpression : LeftHandSideExpression AssignmentOperator AssignmentExpression</emu-grammar>
+      <emu-grammar>
+        AssignmentExpression :
+          LeftHandSideExpression AssignmentOperator AssignmentExpression
+          LeftHandSideExpression `&amp;&amp;=` AssignmentExpression
+          LeftHandSideExpression `||=`  AssignmentExpression
+          LeftHandSideExpression `??=`   AssignmentExpression
```
bakkot, line 15280: There's some extra whitespace on this line and the previous. (Someday the linter will warn for this, but it does not yet know how.)

jridgewell: Fixed.

jridgewell:
This is going for Stage 4 at the July meeting, and still requires Spec Editor approvals.

/cc @syg @michaelficarra @ljharb

ljharb:
We plan to review it during next week's editor call, if not before.

spec.html
```diff
@@ -15377,11 +15393,52 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+        <p>When this expression occurs within strict mode code and the respective GetValue of _lref_ step is reached, it is a runtime error if _lref_ is an unresolvable reference. If it is, a *ReferenceError* exception is thrown. Additionally if the PutValue step is reached, the _lref_ may not be a reference to a data property with the attribute value { [[Writable]]: *false* }, to an accessor property with the attribute value { [[Set]]: *undefined* }, nor to a non-existent property of an object for which the IsExtensible predicate returns the value *false*. In these cases a *TypeError* exception is thrown.</p>
```
syg, line 15441: Presumably whatever made the two PutValue steps worth linking to before holds now, perhaps even more so, given the addition of new assignment forms. Mind adding `[id=]`s and enumerating the places where the ReferenceError is thrown in strict mode?

spec.html
```diff
@@ -15377,11 +15393,52 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+        1. Perform ? PutValue(_lref_, _r_).
         1. Return _r_.
       </emu-alg>
+      <emu-grammar>AssignmentExpression : LeftHandSideExpression `&amp;&amp;=` AssignmentExpression</emu-grammar>
+      <emu-alg>
+        1. Let _lref_ be the result of evaluating |LeftHandSideExpression|.
+        1. Let _lval_ be ? GetValue(_lref_).
+        1. Let _lbool_ be ! ToBoolean(_lval_).
+        1. If _lbool_ is *false*, return _lval_.
+        1. If IsAnonymousFunctionDefinition(|AssignmentExpression|) and IsIdentifierRef of |LeftHandSideExpression| are both *true*, then
```
michaelficarra, line 15405: This reads a bit like a garden path sentence to me. I have a slight preference for

```suggestion
        1. If IsAnonymousFunctionDefinition(|AssignmentExpression|) is *true* and IsIdentifierRef of |LeftHandSideExpression| is *true*, then
```

if for no reason other than more consistency in phrasing.

jridgewell: Done.

spec.html
jridgewell: Done.

PR #2043, opened 2020-06-12
Normative: add NumericLiteralSeparator

rwaldron:
There is nothing surprising here, it's the same spec that's been in the proposal for almost two years and is now in many engines. https://github.com/tc39/proposal-numeric-separator

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4912,7 +4912,6 @@
             `+` StrDecimalDigits
             `-` StrDecimalDigits
 
-
           StrBinaryLiteral :::
             `0b` StrBinaryDigits
             `0B` StrBinaryDigits
@@ -4920,7 +4919,6 @@
           StrBinaryDigits :::
             BinaryDigit
             StrBinaryDigits BinaryDigit
-
 
           StrOctalLiteral :::
             `0o` StrOctalDigits
@@ -5022,8 +5020,6 @@
             <li>
               The MV of <emu-grammar>StrSignedInteger ::: `-` StrDecimalDigits</emu-grammar> is the negative of the MV of |StrDecimalDigits|.
             </li>
-
-
             <li>
               The MV of <emu-grammar>StrBinaryLiteral ::: `0b` StrBinaryDigits</emu-grammar> is the MV of |StrBinaryDigits|.
             </li>
@@ -11604,7 +11600,6 @@
           NonZeroDigit
           NonZeroDigit NumericLiteralSeparator? DecimalDigits
 
-
         DecimalDigits ::
           DecimalDigit
           DecimalDigits NumericLiteralSeparator? DecimalDigit
@@ -11633,7 +11628,6 @@
         BinaryDigits ::
           BinaryDigit
           BinaryDigits NumericLiteralSeparator? BinaryDigit
-
 
         BinaryDigit :: one of
           `0` `1`
--- a/spec.html
+++ b/spec.html
@@ -11579,8 +11579,8 @@
 
         DecimalBigIntegerLiteral ::
           `0` BigIntLiteralSuffix
-          NonZeroDigit
-          NonZeroDigit NumericLiteralSeparator? DecimalDigits BigIntLiteralSuffix
+          NonZeroDigit DecimalDigits? BigIntLiteralSuffix
+          NonZeroDigit NumericLiteralSeparator DecimalDigits BigIntLiteralSuffix
 
         NonDecimalIntegerLiteral ::
           BinaryIntegerLiteral
@@ -11861,7 +11861,11 @@
         <emu-alg>
           1. Return the BigInt value that represents the MV of |NonZeroDigit|.
         </emu-alg>
-        <emu-grammar>DecimalBigIntegerLiteral :: NonZeroDigit DecimalDigits BigIntLiteralSuffix</emu-grammar>
+        <emu-grammar>
+          DecimalBigIntegerLiteral ::
+            NonZeroDigit DecimalDigits BigIntLiteralSuffix
+            NonZeroDigit NumericLiteralSeparator DecimalDigits BigIntLiteralSuffix
+        </emu-grammar>
         <emu-alg>
           1. Let _n_ be the mathematical integer number of code points in |DecimalDigits|.
           1. Let _mv_ be (the MV of |NonZeroDigit| &times; 10<sub>ℝ</sub><sup>_n_</sup>) plus the MV of |DecimalDigits|.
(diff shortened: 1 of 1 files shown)


rwaldron:
I'll fix that failure this weekend, or sometime soon

ljharb:
@rwaldron i'm pretty confident i've seen this proposal's tests landed; but if not please update the label :-)

leobalter:
> @rwaldron i'm pretty confident i've seen this proposal's tests landed; but if not please update the label :-)

https://test262.report/features/numeric-separator-literal 

Landed and shipping in several mainstream implementations.

spec.html
```diff
@@ -11097,7 +11196,8 @@ <h2>Syntax</h2>
 
         DecimalBigIntegerLiteral ::
           `0` BigIntLiteralSuffix
-          NonZeroDigit DecimalDigits? BigIntLiteralSuffix
+          NonZeroDigit
+          NonZeroDigit NumericLiteralSeparator? DecimalDigits BigIntLiteralSuffix
```
leobalter, line 11200:
```suggestion
          NonZeroDigit DecimalDigits? BigIntLiteralSuffix
          NonZeroDigit NumericLiteralSeparator? DecimalDigits BigIntLiteralSuffix
```

leobalter: I updated my suggestion to consider that if NLS is present, it needs to be followed up by mandatory DecimalDigits and BigIntLiteralSuffix. (I removed the `?` from DecimalDigits).

Without the NLS, the first line already covers everything well.

leobalter:
This will also need another change in the spec.html file.

From:

```
        <emu-grammar>DecimalBigIntegerLiteral :: NonZeroDigit DecimalDigits BigIntLiteralSuffix</emu-grammar>
        <emu-alg>
          1. Let _n_ be the mathematical integer number of code points in |DecimalDigits|.
          1. Let _mv_ be (the MV of |NonZeroDigit| &times; 10<sub>ℝ</sub><sup>_n_</sup>) plus the MV of |DecimalDigits|.
          1. Return the BigInt value that represents _mv_.
        </emu-alg>
```

```
        <emu-grammar>
          DecimalBigIntegerLiteral ::
            NonZeroDigit DecimalDigits BigIntLiteralSuffix
            NonZeroDigit NumericLiteralSeparator DecimalDigits BigIntLiteralSuffix
        </emu-grammar>
        <emu-alg>
          1. Let _n_ be the mathematical integer number of code points in |DecimalDigits|.
          1. Let _mv_ be (the MV of |NonZeroDigit| &times; 10<sub>ℝ</sub><sup>_n_</sup>) plus the MV of |DecimalDigits|.
          1. Return the BigInt value that represents _mv_.
        </emu-alg>
```

spec.html
```diff
@@ -42066,7 +42183,7 @@ <h2>Syntax</h2>
 
         DecimalIntegerLiteral ::
           `0`
-          NonZeroDigit DecimalDigits?
+          NonZeroDigit NumericLiteralSeparator? DecimalDigits
```
leobalter, line 42186:
```suggestion
          NonZeroDigit
          NonZeroDigit NumericLiteralSeparator? DecimalDigits
```

leobalter:
I applied the changes directly at @rwaldron's branch as he gave me access to do it.

I also rebased the branch and forced pushed the contents here.

spec.html
```diff
@@ -11201,6 +11302,9 @@ <h2>Syntax</h2>
         <h1>Static Semantics: MV</h1>
         <p>A numeric literal stands for a value of the Number type or the BigInt type.</p>
         <ul>
+          <li>
+            <emu-grammar>NumericLiteralSeparator :: `_`</emu-grammar> has no MV.
```
syg, line 11306: Should this be an editorial note instead of a normative clause for the MV semantics? Is MV ever "invoked" on NumericLiteralSeparator?

bakkot:
I would be in favor of reducing the duplication by introducing a new grammar flag to control whether the separator can be used.

leobalter:
> I would be in favor of reducing the duplication by introducing a new grammar flag to control whether the separator can be used.

I wish this could have been solved earlier or if it were simple as it sounds. I might give it a try but I'm open for suggestions how to better do it.

rwaldron:
> I would be in favor of reducing the duplication by introducing a new grammar flag to control whether the separator can be used.

I actually tried to figure out how to do this with some kind of specialized grammar about a year or two ago, and it just got messy. Since this wasn't brought up during the discussion for advancement, I'm going to assume that you're not hard blocking on it?

bakkot:
I'm not blocking stage 4, but editorially I would prefer to resolve this before landing rather than after. I'm surprised it got messy, though; certainly I wouldn't want to go this route if it's not clean.

I was expecting this to just be a matter of adding a new `[Sep]` (or whatever) grammar flag which controlled whether `_` could be used, and then passing it through as positive from `DecimalIntegerLiteral` and negative from `StrNumericLiteral` (or whatever other appropriate entry points). I can write it up more formally when I get time, if that's not clear. Was that the strategy you found to be messy?

rwaldron:
> Was that the strategy you found to be messy?

No, I'd like to see what you come up with. 

I'm curious, what duplication are you referring to?

bakkot:
> I'm curious, what duplication are you referring to?

The thing where there is now a `StrDecimalDigits` as well as a `DecimalDigits` (etc). These aren't identical, of course, but they're very similar, and I'd prefer to introduce one new grammar flag rather than nine new productions which are so similar to existing ones.

rwaldron:
@bakkot I haven't had any time to dedicate to making an experiment of changing this design. I'm also starting to feel like this obstacle you've created is unfair, considering how long this proposal has existed and that's it's been accepted through several reviews in its current design. I'm also worried that because it's essentially a redesign of the feature, that such an undertaking could put it at risk of losing Stage 4 acceptance. 

If we stick to the process, then this should be allowed to merge as-is and if there are refinements to be made, they can be made in follow up PRs.

rwaldron:
Rebased!

syg:
@rwaldron We chatted about this on the editor call.

I want to first make sure there's no misunderstanding. We think the refactoring-with-flags proposal is a mechanical refactoring of the current grammar, not a redesign. It won't have normative changes, and definitely won't require going back to an earlier stage. Further, there won't be future hurdles. Editorially we don't like the blowup in the grammar currently.

Out of respect for your time, we're also happy to take over the branch and do the refactoring if you don't have cycles for it.

PR #2050, opened 2020-06-15
Editorial: add PromiseCapability Record assertions

ljharb:
I didn't bother to add an assertion after every callsite of `NewPromiseCapability`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -24102,7 +24102,7 @@
 
       <emu-clause id="sec-finishdynamicimport" aoid="FinishDynamicImport">
         <h1>Runtime Semantics: FinishDynamicImport ( _referencingScriptOrModule_, _specifier_, _promiseCapability_, _completion_ )</h1>
-        <p>The abstract operation FinishDynamicImport takes arguments _referencingScriptOrModule_, _specifier_, _promiseCapability_, and _completion_. FinishDynamicImport completes the process of a dynamic import originally started by an <emu-xref href="#sec-import-calls">`import()`</emu-xref> call, resolving or rejecting the promise returned by that call as appropriate according to _completion_. It is performed by host environments as part of HostImportModuleDynamically. It performs the following steps when called:</p>
+        <p>The abstract operation FinishDynamicImport takes arguments _referencingScriptOrModule_, _specifier_, _promiseCapability_ (a PromiseCapability Record), and _completion_. FinishDynamicImport completes the process of a dynamic import originally started by an <emu-xref href="#sec-import-calls">`import()`</emu-xref> call, resolving or rejecting the promise returned by that call as appropriate according to _completion_. It is performed by host environments as part of HostImportModuleDynamically. It performs the following steps when called:</p>
 
         <emu-alg>
           1. If _completion_ is an abrupt completion, then perform ! Call(_promiseCapability_.[[Reject]], *undefined*, « _completion_.[[Value]] »).
@@ -39019,7 +39019,7 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-asyncfromsynciteratorcontinuation" aoid="AsyncFromSyncIteratorContinuation">
         <h1>AsyncFromSyncIteratorContinuation ( _result_, _promiseCapability_ )</h1>
-        <p>The abstract operation AsyncFromSyncIteratorContinuation takes arguments _result_ and _promiseCapability_. It performs the following steps when called:</p>
+        <p>The abstract operation AsyncFromSyncIteratorContinuation takes arguments _result_ and _promiseCapability_ (a PromiseCapability Record). It performs the following steps when called:</p>
 
         <emu-alg>
           1. Let _done_ be IteratorComplete(_result_).
@@ -39572,7 +39572,7 @@ THH:mm:ss.sss
             </tr>
             <tr>
               <td>[[Capability]]</td>
-              <td>A PromiseCapability record</td>
+              <td>A PromiseCapability Record</td>
               <td>The promise capabilities associated with this request.</td>
             </tr>
             </tbody>
@@ -39780,7 +39780,7 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-promisecapability-records">
         <h1>PromiseCapability Records</h1>
-        <p>A PromiseCapability is a Record value used to encapsulate a promise object along with the functions that are capable of resolving or rejecting that promise object. PromiseCapability Records are produced by the NewPromiseCapability abstract operation.</p>
+        <p>A <dfn>PromiseCapability Record</dfn> is a Record value used to encapsulate a promise object along with the functions that are capable of resolving or rejecting that promise object. PromiseCapability Records are produced by the NewPromiseCapability abstract operation.</p>
         <p>PromiseCapability Records have the fields listed in <emu-xref href="#table-57"></emu-xref>.</p>
         <emu-table id="table-57" caption="PromiseCapability Record Fields">
           <table>
@@ -39987,7 +39987,7 @@ THH:mm:ss.sss
         <emu-alg>
           1. If IsConstructor(_C_) is *false*, throw a *TypeError* exception.
           1. NOTE: _C_ is assumed to be a constructor function that supports the parameter conventions of the Promise constructor (see <emu-xref href="#sec-promise-executor"></emu-xref>).
-          1. Let _promiseCapability_ be the PromiseCapability { [[Promise]]: *undefined*, [[Resolve]]: *undefined*, [[Reject]]: *undefined* }.
+          1. Let _promiseCapability_ be the PromiseCapability Record { [[Promise]]: *undefined*, [[Resolve]]: *undefined*, [[Reject]]: *undefined* }.
           1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-getcapabilitiesexecutor-functions" title></emu-xref>.
           1. Let _executor_ be ! CreateBuiltinFunction(_steps_, &laquo; [[Capability]] &raquo;).
           1. Set _executor_.[[Capability]] to _promiseCapability_.
@@ -40099,6 +40099,7 @@ THH:mm:ss.sss
             1. If _promiseCapability_ is *undefined*, then
               1. Assert: _handlerResult_ is not an abrupt completion.
               1. Return NormalCompletion(~empty~).
+            1. Assert: _promiseCapability_ is a PromiseCapability Record.
             1. If _handlerResult_ is an abrupt completion, then
               1. Let _status_ be Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _handlerResult_.[[Value]] &raquo;).
             1. Else,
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -24084,6 +24084,7 @@ <h1>Runtime Semantics: FinishDynamicImport ( _referencingScriptOrModule_, _speci
         <p>The abstract operation FinishDynamicImport takes arguments _referencingScriptOrModule_, _specifier_, _promiseCapability_, and _completion_. FinishDynamicImport completes the process of a dynamic import originally started by an <emu-xref href="#sec-import-calls">`import()`</emu-xref> call, resolving or rejecting the promise returned by that call as appropriate according to _completion_. It is performed by host environments as part of HostImportModuleDynamically. It performs the following steps when called:</p>
 
         <emu-alg>
+          1. Assert: _promiseCapability_ is a PromiseCapability Record.
```
michaelficarra, line 24087: A few of these are making assertions on inputs to AOs as the first step. We usually just put these in parentheticals following the alias name in the summary.

ljharb: i don't think there's a clear winner right now; both explicit and implicit assertions are possible consistent outcomes. i can make the change tho if you think it's important?

spec.html
```diff
@@ -40179,6 +40184,7 @@ <h1>`Promise.all` Resolve Element Functions</h1>
             1. Let _index_ be _F_.[[Index]].
             1. Let _values_ be _F_.[[Values]].
             1. Let _promiseCapability_ be _F_.[[Capability]].
+            1. Assert: _promiseCapability_ is a PromiseCapability Record.
```
anba, line 40187: Adding an assertion for only this internal slot seems a bit inconsistent. What makes `F.[[Capability]]` so special that we need to assert its type is a PromiseCapability Record, whereas the types for `F.[[AlreadyCalled]]`, `F.[[Index]]`, `F.[[Values]]`, and `F.[[RemainingElements]]` still need to inferred from the point where `F` was created? If more explicit type descriptions are necessary, I'd suggest to document them as part of the preamble in [25.6.4.1.3 Promise.all Resolve Element Functions](https://tc39.es/ecma262/#sec-promise.all-resolve-element-functions).

ljharb: That's a fair point - I think they *all* need assertions, this PR was only focusing on PromiseCapability Records.

I wouldn't expect, though, the type of an internal slot to be assertable in prose, since the value could change at any point - i'd expect all of them to have assertions at the point they're first extracted.

anba: There could be a table with (Name, Type, Description) to describe the internal slots and their usage, if we don't want to put it all in prose. 

And I don't understand "the value could change at any point", why should an internal slot suddenly contain a different type of value? For example for [ordinary objects](https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots), the spec describes that [[Prototype]] contains either an object or null, and [[Extensible]] contains a boolean value. And that description is deemed to be enough, we don't assert the types whenever we read [[Prototype]] or [[Extensible]].

ljharb: That’s also a fair point. I think we should somehow - ideally without boilerplate everywhere - but given that, for this case at least, I’ll try to move it to prose.

spec.html
bakkot: I prefer the implicit ones, but since both styles exist currently I'm OK with having explicit ones for now.

(I think it's easier to read simple types out of the argument lists, and I don't think explicit assertions add any value once they're in the argument lists.)

ljharb: For things that are simple types, I tend to agree; for things that are, for example, non-negative integers, or something more subtle, it seems like explicit ones might make more sense.

bakkot: Yeah. Here it's a named record type, which seems simple enough to me - e.g. we already use [Realm Record](https://tc39.es/ecma262/#sec-addrestrictedfunctionproperties). and [Cyclic Module Record](https://tc39.es/ecma262/#sec-InnerModuleLinking) as a types in an argument list.

ljharb: given michael's +1 for implicit and your +0 for it, i'll go ahead and change it in this PR.

bakkot:
We should also add a `dfn` for "[PromiseCapability Record](https://tc39.es/ecma262/#sec-promisecapability-records)" (and maybe also capitalize the word "record" in [Table 75](https://tc39.es/ecma262/#sec-asyncgeneratorrequest-records), and also say "the PromiseCapability Record {" instead of "the PromiseCapability {" in step 3 of [NewPromiseCapability](https://tc39.es/ecma262/#sec-newpromisecapability)).

Doesn't need to be done in this PR if you're not inclined to, though.

ljharb:
all updated

spec.html
```diff
@@ -39695,7 +39697,7 @@ <h1>Promise Abstract Operations</h1>
         <h1>PromiseCapability Records</h1>
         <p>A PromiseCapability is a Record value used to encapsulate a promise object along with the functions that are capable of resolving or rejecting that promise object. PromiseCapability Records are produced by the NewPromiseCapability abstract operation.</p>
```
bakkot, line 39698: Rather than the putting the definition in the table below, I'd suggest

```suggestion
        <p>A <dfn>PromiseCapability Record</dfn> is a Record value used to encapsulate a promise object along with the functions that are capable of resolving or rejecting that promise object. PromiseCapability Records are produced by the NewPromiseCapability abstract operation.</p>
```

The usual style is to put the definition in the paragraph which introduces the term. Compare e.g. [ImportEntry Record](https://tc39.es/ecma262/#importentry-record), [Agent Events Record](https://tc39.es/ecma262/#sec-agent-event-records), etc.

ljharb: thanks, that's much cleaner

spec.html
```diff
@@ -39528,6 +39528,7 @@ <h1>AsyncGeneratorResolve ( _generator_, _value_, _done_ )</h1>
           1. Assert: _queue_ is not an empty List.
           1. Remove the first element from _queue_ and let _next_ be the value of that element.
           1. Let _promiseCapability_ be _next_.[[Capability]].
+          1. Assert: _promiseCapability_ is a PromiseCapability Record.
```
bakkot, line 39531: I am not convinced it's good to assert on the type of an internal slot which is described as holding a value of the type in question. From reading this assertion I would assume that it's possible for the slot to hold some other type, so that it's meaningful to assert that it does not happen to in this particular case - but that's not so; the slot only ever holds a PromiseCapability.

ljharb: Without these assertions tho, that's only the case "accidentally" - iow, there's nothing anywhere else that ensures it's always one in the event of a spec authoring error.

If that's not a concern, I can remove these, but I'm not sure what the downside would be of keeping them?

bakkot: An assertion also would not ensure the type is correct. It just asserts it, exactly the same way the description of the field does. I don't see what the value of the assertion is given that the field is described as containing a PromiseCapability Record.

> I'm not sure what the downside would be of keeping them?

The downside is as I said in the previous comment: from reading this assertion I would assume that it's possible for the slot to hold some other type, so that it's meaningful to assert that it does not happen to in this particular case. Which would be incorrect.

ljharb: an implementation can alter internal slots at any time; without the assertion *somewhere* then it *already* can be any possible type.

bakkot: > an implementation can alter internal slots at any time

... Why do you believe this to be true? I don't think this is true.

> without the assertion _somewhere_ then it _already_ can be any possible type.

Again, assertions have no normative force. They are just comments for clarity to the reader.

ljharb: They're normative in the sense that an implementation doesn't have any observable action to take, and can thus ignore them, but if the assertion is violated, it's either a spec bug or an implementation bug.

bakkot: > if the assertion is violated, it's either a spec bug or an implementation bug.

If an assertion is violated, it's a spec bug, yes. (I'm not really sure what it would mean for an implementation to violate an assertion.) But it would be exactly as much of a bug to put a non-PromiseCapability Record value in a slot which is described as holding a PromiseCapability Record. And it would similarly be a bug to treat a non-PromiseCapability Record value as a PromiseCapability Record. It does not somehow become _more_ of a spec bug just because there's an assertion.

Assertions in the context of the spec are just comments. I think it is a mistake to add a comment for any reason other than because we think it will aid clarity for readers.

(I don't really know what you mean by "normative" in this context.)

ljharb: Fair - I do think this change adds clarity for readers, because I don't have to go elsewhere to figure out what kind of thing is held in the slot.

jmdyck: > Fair - I do think this change adds clarity for readers, because I don't have to go elsewhere to figure out what kind of thing is held in the slot.

We could achieve that by naming the slot more precisely, e.g. `[[PromiseCapability]]` (or, if that's still not enough, `[[PromiseCapabilityRecord]]`).

ljharb: That still wouldn't make it auto-link to "PromiseCapability Record", though.

ljharb:
Removed the assertions immediately following a slot extraction.

spec.html
```diff
@@ -40513,13 +40511,9 @@ <h1>Promise.prototype.then ( _onFulfilled_, _onRejected_ )</h1>
...
+          <p>The abstract operation PerformPromiseThen takes arguments _promise_, _onFulfilled_, and _onRejected_ and optional argument _resultCapability_ (a PromiseCapability Record). It performs the &ldquo;then&rdquo; operation on _promise_ using _onFulfilled_ and _onRejected_ as its settlement actions. If _resultCapability_ is passed, the result is stored by updating _resultCapability_'s promise. If it is not passed, then PerformPromiseThen is being called by a specification-internal operation where the result does not matter. It performs the following steps when called:</p>
           <emu-alg>
             1. Assert: IsPromise(_promise_) is *true*.
-            1. If _resultCapability_ is present, then
-              1. Assert: _resultCapability_ is a PromiseCapability Record.
-            1. Else,
-              1. Set _resultCapability_ to *undefined*.
```
jmdyck, line 40522: I'm guessing you removed this thinking that a not-present parameter would be `*undefined*` anyway. However, that's not a thing for abstract operations.

Rather than reinstate these lines, you could convert `_resultCapability_` to a required parameter, and pass `*undefined*` at the 2 calls where it's currently omitted. (I seem to recall some editor (@syg?) expressing approval for changing optional AO parameters to required.)

ljharb: hmm, i'm not a fan of that idea, so i'll partially revert this back.

spec.html
```diff
@@ -40513,17 +40511,15 @@ <h1>Promise.prototype.then ( _onFulfilled_, _onRejected_ )</h1>
...
             1. If IsCallable(_onFulfilled_) is *false*, then
               1. Set _onFulfilled_ to *undefined*.
             1. If IsCallable(_onRejected_) is *false*, then
               1. Set _onRejected_ to *undefined*.
+            1. If _resultCapability_ is not present, then
+              1. Set _resultCapability_ to *undefined*.
```
syg, line 40522: Nit: I'd prefer this moved back to the original place.

PR #2052, opened 2020-06-16
Editorial: use step labels rather than numbers to refer to steps

bakkot:
This fixes a persistent problem where step numbers would get stale and we'd forget to update them. Now, by given the steps (somewhat arbitrarily chosen) labels, we can refer and link to them explicitly. See https://github.com/tc39/ecmarkup/pull/217 for a description of the syntax.

~Marked as draft until a version of ecmarkup with support for the new syntax is published.~ Edit: done, this should be good to go. Deploy preview looks good to me.

Corresponding ecmarkdown PR at https://github.com/tc39/ecmarkdown/pull/74. Corresponding ecmarkup PR at https://github.com/tc39/ecmarkup/pull/217.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -760,12 +760,12 @@
       <h1>Syntax-Directed Operations</h1>
       <p>A <dfn>syntax-directed operation</dfn> is a named operation whose definition consists of algorithms, each of which is associated with one or more productions from one of the ECMAScript grammars. A production that has multiple alternative definitions will typically have a distinct algorithm for each alternative. When an algorithm is associated with a grammar production, it may reference the terminal and nonterminal symbols of the production alternative as if they were parameters of the algorithm. When used in this manner, nonterminal symbols refer to the actual alternative definition that is matched when parsing the source text. The <dfn>source text matched by</dfn> a grammar production is the portion of the source text that starts at the beginning of the first terminal that participated in the match and ends at the end of the last terminal that participated in the match.</p>
       <p>When an algorithm is associated with a production alternative, the alternative is typically shown without any &ldquo;[ ]&rdquo; grammar annotations. Such annotations should only affect the syntactic recognition of the alternative and have no effect on the associated semantics for the alternative.</p>
-      <p>Syntax-directed operations are invoked with a parse node and, optionally, other parameters by using the conventions on steps 1, 3, and 4 in the following algorithm:</p>
+      <p>Syntax-directed operations are invoked with a parse node and, optionally, other parameters by using the conventions on steps <emu-xref href="#step-sdo-invocation-example-1"></emu-xref>, <emu-xref href="#step-sdo-invocation-example-2"></emu-xref>, and <emu-xref href="#step-sdo-invocation-example-3"></emu-xref> in the following algorithm:</p>
       <emu-alg>
-        1. Let _status_ be SyntaxDirectedOperation of |SomeNonTerminal|.
+        1. [id="step-sdo-invocation-example-1"] Let _status_ be SyntaxDirectedOperation of |SomeNonTerminal|.
         1. Let _someParseNode_ be the parse of some source text.
-        1. Perform SyntaxDirectedOperation of _someParseNode_.
-        1. Perform SyntaxDirectedOperation of _someParseNode_ passing *"value"* as the argument.
+        1. [id="step-sdo-invocation-example-2"] Perform SyntaxDirectedOperation of _someParseNode_.
+        1. [id="step-sdo-invocation-example-3"] Perform SyntaxDirectedOperation of _someParseNode_ passing *"value"* as the argument.
       </emu-alg>
       <p>Unless explicitly specified otherwise, all chain productions have an implicit definition for every operation that might be applied to that production's left-hand side nonterminal. The implicit definition simply reapplies the same operation with the same parameters, if any, to the chain production's sole right-hand side nonterminal and then returns the result. For example, assume that some algorithm has a step of the form: &ldquo;Return the result of evaluating |Block|&rdquo; and that there is a production:</p>
       <emu-grammar type="example">
@@ -1808,7 +1808,7 @@
             1. If _x_ is *+0* or *-0*, return the String *"0"*.
             1. If _x_ is less than zero, return the string-concatenation of *"-"* and ! Number::toString(-_x_).
             1. If _x_ is *+&infin;*, return the String *"Infinity"*.
-            1. Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, 10<sup>_k_ - 1</sup> &le; _s_ &lt; 10<sup>_k_</sup>, the Number value for ℝ(_s_) &times; 10<sub>ℝ</sub><sup>ℝ(_n_) - ℝ(_k_)</sup> is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the decimal representation of _s_, that _s_ is not divisible by 10<sub>ℝ</sub>, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
+            1. [id="step-number-tostring-intermediate-values"] Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, 10<sup>_k_ - 1</sup> &le; _s_ &lt; 10<sup>_k_</sup>, the Number value for ℝ(_s_) &times; 10<sub>ℝ</sub><sup>ℝ(_n_) - ℝ(_k_)</sup> is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the decimal representation of _s_, that _s_ is not divisible by 10<sub>ℝ</sub>, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
             1. If _k_ &le; _n_ &le; 21, return the string-concatenation of:
               * the code units of the _k_ digits of the decimal representation of _s_ (in order, with no leading zeroes)
               * _n_ - _k_ occurrences of the code unit 0x0030 (DIGIT ZERO)
@@ -1841,14 +1841,14 @@
                 If x is any Number value other than *-0*, then ToNumber(ToString(x)) is exactly the same Number value as x.
               </li>
               <li>
-                The least significant digit of s is not always uniquely determined by the requirements listed in step 5.
+                The least significant digit of s is not always uniquely determined by the requirements listed in step <emu-xref href="#step-number-tostring-intermediate-values"></emu-xref>.
               </li>
             </ul>
           </emu-note>
           <emu-note>
-            <p>For implementations that provide more accurate conversions than required by the rules above, it is recommended that the following alternative version of step 5 be used as a guideline:</p>
-            <emu-alg>
-              5. Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, 10<sup>_k_ - 1</sup> &le; _s_ &lt; 10<sup>_k_</sup>, the Number value for ℝ(_s_) &times; 10<sub>ℝ</sub><sup>ℝ(_n_) - ℝ(_k_)</sup> is _x_, and _k_ is as small as possible. If there are multiple possibilities for _s_, choose the value of _s_ for which ℝ(_s_) &times; 10<sub>ℝ</sub><sup>ℝ(_n_) - ℝ(_k_)</sup> is closest in value to ℝ(_x_). If there are two such possible values of _s_, choose the one that is even. Note that _k_ is the number of digits in the decimal representation of _s_ and that _s_ is not divisible by 10<sub>ℝ</sub>.
+            <p>For implementations that provide more accurate conversions than required by the rules above, it is recommended that the following alternative version of step <emu-xref href="#step-number-tostring-intermediate-values"></emu-xref> be used as a guideline:</p>
+            <emu-alg replaces-step="step-number-tostring-intermediate-values">
+              1. Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, 10<sup>_k_ - 1</sup> &le; _s_ &lt; 10<sup>_k_</sup>, the Number value for ℝ(_s_) &times; 10<sub>ℝ</sub><sup>ℝ(_n_) - ℝ(_k_)</sup> is _x_, and _k_ is as small as possible. If there are multiple possibilities for _s_, choose the value of _s_ for which ℝ(_s_) &times; 10<sub>ℝ</sub><sup>ℝ(_n_) - ℝ(_k_)</sup> is closest in value to ℝ(_x_). If there are two such possible values of _s_, choose the one that is even. Note that _k_ is the number of digits in the decimal representation of _s_ and that _s_ is not divisible by 10<sub>ℝ</sub>.
             </emu-alg>
           </emu-note>
           <emu-note>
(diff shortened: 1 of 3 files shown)


spec.html
```diff
@@ -760,12 +760,12 @@ <h1>Abstract Operations</h1>
...
       <p>When an algorithm is associated with a production alternative, the alternative is typically shown without any &ldquo;[ ]&rdquo; grammar annotations. Such annotations should only affect the syntactic recognition of the alternative and have no effect on the associated semantics for the alternative.</p>
-      <p>Syntax-directed operations are invoked with a parse node and, optionally, other parameters by using the conventions on steps 1, 3, and 4 in the following algorithm:</p>
+      <p>Syntax-directed operations are invoked with a parse node and, optionally, other parameters by using the conventions on steps <emu-xref href="#step-sdo-invocation-example-1"></emu-xref>, <emu-xref href="#step-sdo-invocation-example-2"></emu-xref>, and <emu-xref href="#step-sdo-invocation-example-3"></emu-xref> in the following algorithm:</p>
```
ljharb, line 763:
```suggestion
      <p>Syntax-directed operations are invoked with a parse node and, optionally, other parameters by using the conventions on steps <emu-xref title href="#step-sdo-invocation-example-1"></emu-xref>, <emu-xref title href="#step-sdo-invocation-example-2"></emu-xref>, and <emu-xref title href="#step-sdo-invocation-example-3"></emu-xref> in the following algorithm:</p>
```

spec.html
```diff
@@ -1841,14 +1841,14 @@ <h1>Number::toString ( _x_ )</h1>
                 If x is any Number value other than *-0*, then ToNumber(ToString(x)) is exactly the same Number value as x.
               </li>
               <li>
-                The least significant digit of s is not always uniquely determined by the requirements listed in step 5.
+                The least significant digit of s is not always uniquely determined by the requirements listed in step <emu-xref href="#step-number-tostring-intermediate-values"></emu-xref>.
```
ljharb, line 1844:
```suggestion
                The least significant digit of s is not always uniquely determined by the requirements listed in step <emu-xref title href="#step-number-tostring-intermediate-values"></emu-xref>.
```

spec.html
```diff
@@ -1841,14 +1841,14 @@ <h1>Number::toString ( _x_ )</h1>
...
             </ul>
           </emu-note>
           <emu-note>
-            <p>For implementations that provide more accurate conversions than required by the rules above, it is recommended that the following alternative version of step 5 be used as a guideline:</p>
-            <emu-alg>
-              5. Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, 10<sup>_k_ - 1</sup> &le; _s_ &lt; 10<sup>_k_</sup>, the Number value for ℝ(_s_) &times; 10<sub>ℝ</sub><sup>ℝ(_n_) - ℝ(_k_)</sup> is _x_, and _k_ is as small as possible. If there are multiple possibilities for _s_, choose the value of _s_ for which ℝ(_s_) &times; 10<sub>ℝ</sub><sup>ℝ(_n_) - ℝ(_k_)</sup> is closest in value to ℝ(_x_). If there are two such possible values of _s_, choose the one that is even. Note that _k_ is the number of digits in the decimal representation of _s_ and that _s_ is not divisible by 10<sub>ℝ</sub>.
+            <p>For implementations that provide more accurate conversions than required by the rules above, it is recommended that the following alternative version of step <emu-xref href="#step-number-tostring-intermediate-values"></emu-xref> be used as a guideline:</p>
```
ljharb, line 1849:
```suggestion
            <p>For implementations that provide more accurate conversions than required by the rules above, it is recommended that the following alternative version of step <emu-xref title href="#step-number-tostring-intermediate-values"></emu-xref> be used as a guideline:</p>
```

spec.html
```diff
@@ -4256,14 +4256,14 @@ <h1>GetValue ( _V_ )</h1>
...
-              1. Set _base_ to ! ToObject(_base_).
+              1. [label="getvalue-toobject"] Set _base_ to ! ToObject(_base_).
             1. Return ? _base_.[[Get]](GetReferencedName(_V_), GetThisValue(_V_)).
           1. Else,
             1. Assert: _base_ is an Environment Record.
             1. Return ? _base_.GetBindingValue(GetReferencedName(_V_), IsStrictReference(_V_)) (see <emu-xref href="#sec-environment-records"></emu-xref>).
         </emu-alg>
         <emu-note>
-          <p>The object that may be created in step 5.a.ii is not accessible outside of the above abstract operation and the ordinary object [[Get]] internal method. An implementation might choose to avoid the actual creation of the object.</p>
+          <p>The object that may be created in step <emu-xref href="#step-getvalue-toobject"></emu-xref> is not accessible outside of the above abstract operation and the ordinary object [[Get]] internal method. An implementation might choose to avoid the actual creation of the object.</p>
```
ljharb, line 4266:
```suggestion
          <p>The object that may be created in step <emu-xref title href="#step-getvalue-toobject"></emu-xref> is not accessible outside of the above abstract operation and the ordinary object [[Get]] internal method. An implementation might choose to avoid the actual creation of the object.</p>
```

spec.html
```diff
@@ -4292,7 +4292,7 @@ <h1>PutValue ( _V_, _W_ )</h1>
             1. Return ? _base_.SetMutableBinding(GetReferencedName(_V_), _W_, IsStrictReference(_V_)) (see <emu-xref href="#sec-environment-records"></emu-xref>).
         </emu-alg>
         <emu-note>
-          <p>The object that may be created in step 6.a.ii is not accessible outside of the above abstract operation and the ordinary object [[Set]] internal method. An implementation might choose to avoid the actual creation of that object.</p>
+          <p>The object that may be created in step <emu-xref href="#step-putvalue-toobject"></emu-xref> is not accessible outside of the above abstract operation and the ordinary object [[Set]] internal method. An implementation might choose to avoid the actual creation of that object.</p>
```
ljharb, line 4295:
```suggestion
          <p>The object that may be created in step <emu-xref title href="#step-putvalue-toobject"></emu-xref> is not accessible outside of the above abstract operation and the ordinary object [[Set]] internal method. An implementation might choose to avoid the actual creation of that object.</p>
```

spec.html
```diff
@@ -4918,13 +4918,13 @@ <h1>ToUint32 ( _argument_ )</h1>
...
         1. Let _int32bit_ be _int_ modulo 2<sup>32</sup>.
-        1. Return _int32bit_.
+        1. [label="touint32-return"] Return _int32bit_.
       </emu-alg>
       <emu-note>
         <p>Given the above definition of ToUint32:</p>
         <ul>
           <li>
-            Step 5 is the only difference between ToUint32 and ToInt32.
+            Step <emu-xref href="#step-touint32-return"></emu-xref> is the only difference between ToUint32 and ToInt32.
```
ljharb, line 4927:
```suggestion
            Step <emu-xref title href="#step-touint32-return"></emu-xref> is the only difference between ToUint32 and ToInt32.
```

spec.html
```diff
@@ -4958,14 +4958,14 @@ <h1>ToUint16 ( _argument_ )</h1>
...
-        1. Let _int16bit_ be _int_ modulo 2<sup>16</sup>.
+        1. [label="touint16-mod"] Let _int16bit_ be _int_ modulo 2<sup>16</sup>.
         1. Return _int16bit_.
       </emu-alg>
       <emu-note>
         <p>Given the above definition of ToUint16:</p>
         <ul>
           <li>
-            The substitution of 2<sup>16</sup> for 2<sup>32</sup> in step 4 is the only difference between ToUint32 and ToUint16.
+            The substitution of 2<sup>16</sup> for 2<sup>32</sup> in step <emu-xref href="#step-touint16-mod"></emu-xref> is the only difference between ToUint32 and ToUint16.
```
ljharb, line 4968:
```suggestion
            The substitution of 2<sup>16</sup> for 2<sup>32</sup> in step <emu-xref title href="#step-touint16-mod"></emu-xref> is the only difference between ToUint32 and ToUint16.
```

spec.html
```diff
@@ -5618,7 +5618,7 @@ <h1>Abstract Relational Comparison</h1>
           1. If the mathematical value of _nx_ is less than the mathematical value of _ny_, return *true*; otherwise return *false*.
       </emu-alg>
       <emu-note>
-        <p>Step 3 differs from step 1.c in the algorithm that handles the addition operator `+` (<emu-xref href="#sec-applystringornumericbinaryoperator"></emu-xref>) by using the logical-and operation instead of the logical-or operation.</p>
+        <p>Step <emu-xref href="#step-arc-string-check"></emu-xref> differs from step <emu-xref href="#step-binary-op-string-check"></emu-xref> in the algorithm that handles the addition operator `+` (<emu-xref href="#sec-applystringornumericbinaryoperator"></emu-xref>) by using the logical-and operation instead of the logical-or operation.</p>
```
ljharb, line 5621:
```suggestion
        <p>Step <emu-xref title href="#step-arc-string-check"></emu-xref> differs from step <emu-xref title href="#step-binary-op-string-check"></emu-xref> in the algorithm that handles the addition operator `+` (<emu-xref href="#sec-applystringornumericbinaryoperator"></emu-xref>) by using the logical-and operation instead of the logical-or operation.</p>
```

spec.html
```diff
@@ -6024,10 +6024,10 @@ <h1>GetFunctionRealm ( _obj_ )</h1>
           1. If _obj_.[[ProxyHandler]] is *null*, throw a *TypeError* exception.
           1. Let _proxyTarget_ be _obj_.[[ProxyTarget]].
           1. Return ? GetFunctionRealm(_proxyTarget_).
-        1. Return the current Realm Record.
+        1. [label="getfunctionrealm-default-return"] Return the current Realm Record.
       </emu-alg>
       <emu-note>
-        <p>Step 5 will only be reached if _obj_ is a non-standard function exotic object that does not have a [[Realm]] internal slot.</p>
+        <p>Step <emu-xref href="#step-getfunctionrealm-default-return"></emu-xref> will only be reached if _obj_ is a non-standard function exotic object that does not have a [[Realm]] internal slot.</p>
```
ljharb, line 6030:
```suggestion
        <p>Step <emu-xref title href="#step-getfunctionrealm-default-return"></emu-xref> will only be reached if _obj_ is a non-standard function exotic object that does not have a [[Realm]] internal slot.</p>
```

spec.html
```diff
@@ -6414,7 +6414,7 @@ <h1>SetMutableBinding ( _N_, _V_, _S_ )</h1>
             1. Return NormalCompletion(~empty~).
           </emu-alg>
           <emu-note>
-            <p>An example of ECMAScript code that results in a missing binding at step 2 is:</p>
+            <p>An example of ECMAScript code that results in a missing binding at step <emu-xref href="#step-setmutablebinding-missing-binding"></emu-xref> is:</p>
```
ljharb, line 6417:
```suggestion
            <p>An example of ECMAScript code that results in a missing binding at step <emu-xref title href="#step-setmutablebinding-missing-binding"></emu-xref> is:</p>
```

spec.html
```diff
@@ -7590,15 +7590,15 @@ <h1>GetThisEnvironment ( )</h1>
...
+        1. [label="getthisenvironment-loop"] Repeat,
           1. Let _exists_ be _env_.HasThisBinding().
           1. If _exists_ is *true*, return _env_.
           1. Let _outer_ be _env_.[[OuterEnv]].
           1. Assert: _outer_ is not *null*.
           1. Set _env_ to _outer_.
       </emu-alg>
       <emu-note>
-        <p>The loop in step 2 will always terminate because the list of environments always ends with the global environment which has a `this` binding.</p>
+        <p>The loop in step <emu-xref href="#step-getthisenvironment-loop"></emu-xref> will always terminate because the list of environments always ends with the global environment which has a `this` binding.</p>
```
ljharb, line 7601:
```suggestion
        <p>The loop in step <emu-xref title href="#step-getthisenvironment-loop"></emu-xref> will always terminate because the list of environments always ends with the global environment which has a `this` binding.</p>
```

spec.html
```diff
@@ -7906,7 +7906,7 @@ <h1>OrdinarySetPrototypeOf ( _O_, _V_ )</h1>
           1. Return *true*.
         </emu-alg>
         <emu-note>
-          <p>The loop in step 8 guarantees that there will be no circularities in any prototype chain that only includes objects that use the ordinary object definitions for [[GetPrototypeOf]] and [[SetPrototypeOf]].</p>
+          <p>The loop in step <emu-xref href="#step-ordinarysetprototypeof-loop"></emu-xref> guarantees that there will be no circularities in any prototype chain that only includes objects that use the ordinary object definitions for [[GetPrototypeOf]] and [[SetPrototypeOf]].</p>
```
ljharb, line 7909:
```suggestion
          <p>The loop in step <emu-xref title href="#step-ordinarysetprototypeof-loop"></emu-xref> guarantees that there will be no circularities in any prototype chain that only includes objects that use the ordinary object definitions for [[GetPrototypeOf]] and [[SetPrototypeOf]].</p>
```

bakkot:
The presence of the `title` attribute [indicates](https://tc39.es/ecmarkup/#emu-xref) that the title of the referent rather than its number should be used to provide the text of the reference. As such, `title` doesn't seem appropriate - steps do not have titles; they can only be referred to by number. Is there a different reason to add it?

> and also hopefully allowing it to be self-closing

Ecma262 doesn't currently use any self-closing tags. I'm fine with discussing the possibility of adopting their use, but I would not want to do that in this PR.

ljharb:
Sure, set that aside; maybe `title` is the wrong thing, but the contents of the emu-xref should be inferred somehow. Maybe a `step` attribute?

bakkot:
> the contents of the emu-xref should be inferred somehow

They are; see the screenshot in https://github.com/tc39/ecmarkup/pull/217, which corresponds to https://github.com/tc39/ecmarkup/blob/e819295dd0f0df49c23c44696766595ec580963d/test/algorithm-replacements.html. That's [the default behavior](https://tc39.es/ecmarkup/#emu-xref) when you leave the tag empty.

ljharb:
That's not what I see in this PR's build preview - maybe I just got confused because this doesn't include the ecmarkup PR yet?

bakkot:
Yeah, the build preview won't look right until the syntax it relies on is released.

ljharb:
Would it be better to have an emu-step tag rather than an xref one then?

brabalan:
Thank you, this will be most useful for JSExplain. It might also solve our need to reference individual algorithms (https://github.com/tc39/ecma262/pull/1800 and https://github.com/tc39/ecmarkup/issues/172).

bakkot:
> Would it be better to have an emu-step tag rather than an xref one then?

I don't see step references as fundamentally different from any other kind of thing you can emu-xref. What would be the advantage of having a separate tag?

ljharb:
Other kinds of xrefs can have custom link text which prevents automatically showing the reference's title; i don't see any use case for a link to a step without including a step number.

bakkot:
That's true; I just made it a warning ([to be promoted](https://github.com/tc39/ecmarkup/issues/173#issuecomment-643794350) to a linting error) to include text. It doesn't seem like enough of a difference to warrant a different kind of element, to me.

ljharb:
With the linting error i guess it's fine, but it seems conceptually like a different element to me. (we probably also want a warning on a non-step xref without `title` and without text?)

bakkot:
> but it seems conceptually like a different element to me

How so?

> we probably also want a warning on a non-step xref without title and without text?

That populates the link text with the clause number / figure number / etc, just as here it is populated with the step number; I don't see any particular problem with it. (We do that in ecma262.)

ljharb:
I think that "step" is a more specialized subtype of "a link", and i think that merits a different type.

It's not a blocking point, though - it was just an idea :-)

bakkot:
@ljharb Now that ecmarkup has been bumped the deploy preview renders correctly (see below). If that addresses your concern about `title`, would you resolve your comments above?

Screen Shot 2020-06-16 at 11 59 46 PM

jmdyck:
Currently, if you're looking at an `<emu-xref href="#foo">` in the spec source, and want to go to the referenced point, you can do a search on `foo` (or, for less noise, `id="foo"`). This wouldn't work for the `<emu-xref>` elements introduced by this PR, because the full step-identifiers don't exist in the source. So consider changing the defining syntax from
```[label="something"]```
to
```[id="step-something"]```

Similarly, with regard to`<emu-alg replaces-step="something">`, consider that you might want to allow an emu document (e.g., a proposal) to replace a step in an algorithm in a different spec. For that, a reasonable syntax might be, for example:
```<emu-alg replaces-step="https://tc39.es/ecma262/#step-something">```
which suggests that the syntax for an in-spec reference should be:
```<emu-alg replaces-step="#step-something">```

In general, I think you should avoid using not-quite-the-real-identifier in source.

bakkot:
@jmdyck @syg Updated.

PR #2054, opened 2020-06-16
Normative: Specify \8 and \9 in sloppy non-template strings.

### LINKED ISSUE #2039: \8 in sloppy non-template strings remains unspecified
Currently, JSC bans `\8` (and `\9`) in templates and strict strings but SM/V8/XS/Ch all allow it.

At first I thought this was simply due to a lack of tests, so I created https://github.com/tc39/test262/pull/2654, but it turns out this has never been specified at all (though it was [planned for ES7](https://tc39.es/archives/bugzilla/3477/)).

I still view the tests above as the goal, because:
- We definitely should not leave this unspecified.
- [11.8.4](https://tc39.es/ecma262/#prod-EscapeSequence) and [11.8.6](https://tc39.es/ecma262/#prod-TemplateCharacter) are clear about disallowing B.1.2 for templates and strict strings.
- [B.1.1](https://tc39.es/ecma262/#sec-additional-syntax-numeric-literals) has a non-octal production, so it makes sense for [B.1.2](https://tc39.es/ecma262/#sec-additional-syntax-string-literals) to follow suit.

Assuming no one is deeply opposed to this, I'll make a PR and present it at the next meeting.

rkirsling:
Resolves #2039.

Specifically:
- (Untagged) templates already have early errors for <code>\`\8\`</code> and <code>\`\9\`</code> so no action is needed there.
- All engines agree that `'\8'` and `'\9'` are just `'8'` and `'9'` in sloppy mode, so specify that in B.1.2.
- JSC has never(?) supported `'\8'` and `'\9'` in strict mode, so there should be no web compatibility risk.

Tests: https://github.com/tc39/test262/pull/2654

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -11508,7 +11508,7 @@
           HexEscapeSequence
           UnicodeEscapeSequence
       </emu-grammar>
-      <p>A conforming implementation, when processing strict mode code, must not extend the syntax of |EscapeSequence| to include <emu-xref href="#prod-annexB-LegacyOctalEscapeSequence"></emu-xref> as described in <emu-xref href="#sec-additional-syntax-string-literals"></emu-xref>.</p>
+      <p>A conforming implementation, when processing strict mode code, must not extend the syntax of |EscapeSequence| to include <emu-xref href="#prod-annexB-LegacyOctalEscapeSequence"></emu-xref> or <emu-xref href="#prod-annexB-NonOctalDecimalEscapeSequence"></emu-xref> as described in <emu-xref href="#sec-additional-syntax-string-literals"></emu-xref>.</p>
       <emu-grammar type="definition">
         CharacterEscapeSequence ::
           SingleEscapeCharacter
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -42121,15 +42122,15 @@ <h2>Syntax</h2>
 
         FourToSeven :: one of
           `4` `5` `6` `7`
+
+        NonOctalDecimalEscapeSequence :: one of
```
rkirsling, line 42126: This could just be an alias for `NonOctalDigit`, but I think SV would be less clean to specify that way? Feel free to correct me though.

spec.html
```diff
@@ -42121,15 +42122,15 @@ <h2>Syntax</h2>
...
+          `8` `9`
```
michaelficarra, line 42127: When there's only 2 alternatives, it feels a bit unnecessary to use "one of".

rkirsling: Hmm, that's a copy-paste from [B.1.1](https://tc39.es/ecma262/#sec-additional-syntax-numeric-literals), which in turns follows the suit of [11.8.3](https://tc39.es/ecma262/#sec-literals-numeric-literals) (e.g., ExponentIndicator and BinaryDigit).

We could change them all if we wanted to enforce a minimum of three disjuncts though.

michaelficarra: No it's probably not worth it.

PR #2057, opened 2020-06-19
Normative: add `Reflect[Symbol.toStringTag]`

### LINKED ISSUE #1970: Reconsidering Reflect[Symbol.toStringTag]
https://github.com/tc39/ecma262/issues/495 indicates that it is intentional to not give `Reflect` a `Symbol.toStringTag` value, making it intentionally inconsistent with the older namespaces `Math` and `JSON`. The implication was then that all future namespace objects should not have such toStringTags.

However, since that time `Atomics` was added, and given a toStringTag. At this point it seems like `Reflect` is the odd one out.

It'd be good to get a conclusion here especially to inform the web platform. Currently the specification for web platform namespaces does not indicate they should have any toStringTag. But, browsers are inconsistent: of the three namespaces `WebAssembly`, `console`, and `CSS`, there seems to be some agreement that `WebAssembly` gets a toStringTag, but the others do not. We're trying to straighten this out, but can't figure out which direction to go in, because the ES spec itself is inconsistent.

My recommendation is that everything gain a `Symbol.toStringTag`, including `Reflect`, and that we align the web platform namespaces in the same way. I.e. favor consistency over the simplicity arguments from #495. However, I'm fine with the opposite conclusion, where `Math`, `JSON`, and `Atomics` become legacy exceptions; we should just state it clearly (perhaps with non-normative notes in the toStringTag subsections for each of these objects).

ljharb:
Fixes #1970.

Essentially, this PR makes Reflect no longer be the "odd one out" - ie, the only namespace-like builtin object that lacks a `Symbol.toStringTag`.

See also: https://github.com/tc39/ecma402/pull/430

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -41010,6 +41010,12 @@ THH:mm:ss.sss
         1. Return ? _target_.[[SetPrototypeOf]](_proto_).
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-reflect-@@tostringtag">
+      <h1>Reflect [ @@toStringTag ]</h1>
+      <p>The initial value of the @@toStringTag property is the String value *"Reflect"*.</p>
+      <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
+    </emu-clause>
   </emu-clause>
 
   <emu-clause id="sec-proxy-objects">


domenic:
Just putting a note in here to also discuss the `Intl` namespace object, when this comes to committee.

ljharb:
Consensus for this change, as well as the future principle of "consistency" for `Intl`, and web platform objects, was achieved today at TC39.

ljharb:
Tests: https://github.com/tc39/test262/pull/2710

PR #2059, opened 2020-06-20
Editorial: Fix %Generator% and %AsyncGenerator% <dfn>s

ExE-Boss:
Currently, `%AsyncGenerator%` auto‑links to <https://tc39.es/ecma262/#sec-asyncgeneratorfunction-prototype> instead of <https://tc39.es/ecma262/#sec-properties-of-asyncgeneratorfunction-prototype>.

I also sorted the `<ul>` to match other definitions and made the “see Figure” text an `<emu‑xref>`.

---

[Preview](https://ci.tc39.es/preview/tc39/ecma262/pull/2059) ([#sec-well-known-intrinsic-objects](https://ci.tc39.es/preview/tc39/ecma262/pull/2059#sec-well-known-intrinsic-objects), [#sec-properties-of-asyncgeneratorfunction-prototype](https://ci.tc39.es/preview/tc39/ecma262/pull/2059#sec-properties-of-asyncgeneratorfunction-prototype))

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -18144,7 +18144,7 @@
           <ul>
             <li>has properties that are inherited by all For-In Iterator Objects.</li>
             <li>is an ordinary object.</li>
-            <li>has a [[Prototype]] internal slot whose value is the intrinsic object %IteratorPrototype%.</li>
+            <li>has a [[Prototype]] internal slot whose value is %IteratorPrototype%.</li>
             <li>is never directly accessible to ECMAScript code.</li>
             <li>has the following properties:</li>
           </ul>
@@ -25710,7 +25710,7 @@
       <h1>The Object Constructor</h1>
       <p>The Object constructor:</p>
       <ul>
-        <li>is the intrinsic object <dfn>%Object%</dfn>.</li>
+        <li>is <dfn>%Object%</dfn>.</li>
         <li>is the initial value of the *"Object"* property of the global object.</li>
         <li>creates a new ordinary object when called as a constructor.</li>
         <li>performs a type conversion when called as a function rather than as a constructor.</li>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -40671,10 +40671,10 @@ <h1>AsyncFunction.prototype</h1>
       <h1>Properties of the AsyncFunction Prototype Object</h1>
       <p>The AsyncFunction prototype object:</p>
       <ul>
+        <li>is the intrinsic object <dfn>%AsyncFunctionPrototype%</dfn>.</li>
+        <li>is the value of the *"prototype"* property of %AsyncFunction%.</li>
```
bakkot, line 40675:
```suggestion
        <li>is the initial value of the *"prototype"* property of %AsyncFunction%.</li>
```

ljharb: another alternative:
```suggestion
        <li>is %AsyncFunction.prototype%.</li>
```
which implicitly conveys "initial" as well.

ExE-Boss:
Well, non‑dotted intrinsic `<dfn>`s all use “is the intrinsic object &lt;dfn&gt;%Foo%&lt;/dfn&gt;”, and all dotted intrinsic `<dfn>`s currently use “is &lt;dfn&gt;%Foo.bar&lt;/dfn&gt;”

syg:
> Well, all non‑dotted intrinsic <dfn>s all use “is the intrinsic object <dfn>%Foo%</dfn>”, and all (1) dotted intrinsic <dfn>s currently use “is <dfn>%Foo.bar%</dfn>”.

Ah I see, thanks for pointing this out. I retract my inconsistency statement. Still, I'm not sure the "intrinsic object %Foo%" is any clearer than "%Foo%". While it is true all %Foo% without dotted access are objects, is that property that salient?

jmdyck:
> We're not consistent with whether we're saying "is the intrinsic object %Foo%" or "is %Foo%". My preference is we standardize on the latter.

PR #1376 deleted "the intrinsic object" from pretty much every occurrence of "the intrinsic object %Foo%", other than the ones with `<dfn>...</dfn>` around "%Foo%". I think @ljharb gave a reason for excluding the latter cases, but I can't find it.

ljharb:
At this point, regardless what I may have thought previously, I think "is %Foo%" or "is %Foo.bar%"` is clearer. The very use of the `%` syntax implies "intrinsic", and as such, is redundant.

ljharb:
The editors have decided to remove the redundant "the intrinsic object" phrasing, which I've included in this PR as a separate editorial commit, and rebased the PR's original commit on top of it.

PR #2063, opened 2020-06-24
Normative: Make eval?.() direct eval.

### LINKED ISSUE #2062: Should eval?.() be direct eval?
Issue-ization of https://github.com/tc39/test262/pull/2667#issuecomment-648442908.

`eval?.()` is specified as indirect eval, which is [evidently](https://github.com/tc39/proposal-optional-chaining/issues/21) intentional, but speaking as a reviewer and implementer, I'm not sure how many of us understood this to be the case.

In particular, JSC / SM / V8 all implemented and shipped what we expected the spec to say—i.e., that "code with `?.` behaves the same as code without it when the LHS is non-nullish"—so we have a divergent reality here.

Whichever way we resolve this, I think it's necessary to have a proper discussion about it.
I'll also create a PR for moving to direct eval, at least for demonstration's sake.

rkirsling:
Resolves #2062.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13955,6 +13955,14 @@
         <p>With parameters _baseValue_ and _baseReference_.</p>
         <emu-grammar>OptionalChain : `?.` Arguments</emu-grammar>
         <emu-alg>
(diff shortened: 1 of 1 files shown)


rkirsling:
Closing; we decided to leave the spec as-is in today's meeting.

PR #2067, opened 2020-06-25
fixes #2066: add a language tag to the document body

### LINKED ISSUE #2066: Denote language
The published specification lacks a `lang` attribute.

michaelficarra:
This fixes #2066. There wasn't already a `<body>`, so I added one.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,24 +11,24 @@
       "dev": true
     },
     "@babel/code-frame": {
-      "version": "7.10.1",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.1.tgz",
-      "integrity": "sha512-IGhtTmpjGbYzcEDOw7DcQtbQSXcG9ftmAXtWTu9V936vDye4xjjekktFAtgZsWpzTj/X01jocB46mTywm/4SZw==",
+      "version": "7.10.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.3.tgz",
+      "integrity": "sha512-fDx9eNW0qz0WkUeqL6tXEXzVlPh6Y5aCDEZesl0xBGA8ndRukX91Uk44ZqnkECp01NAZUdCAl+aiQNGi0k88Eg==",
       "requires": {
-        "@babel/highlight": "^7.10.1"
+        "@babel/highlight": "^7.10.3"
       }
     },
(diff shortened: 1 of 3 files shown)


rkirsling:
Isn't `lang` more commonly added to the `<html>` element?

michaelficarra:
@rkirsling It can be added anywhere. I figured `<body>` was appropriate since I wasn't trying to describe the language of anything outside of there. None of those elements are displayed on the page anyway.

michaelficarra:
@ljharb We don't have an `<html>` tag either. Should I add that and move the `lang` attr? If so, should I leave the `<body>` in?

ljharb:
(I'm happy to land this as-is as well) i'd lightly prefer to create both the html and body tags (and head, if we're missing it), and stick `lang` on html.

michaelficarra:
@ljharb @rkirsling I tried moving it to `<html>` but it seems that ecmarkup just never outputs an `<html>` tag, so the attr was being dropped. So I'm leaving it on the body for now. I've added the missing `<html>` and `<head>` tags to make @ljharb happy.

annevk:
> None of those elements are displayed on the page anyway.

You can probably do it on `body` if you also add it explicitly on `title`. `html` is still somewhat preferable in case you ever add any elements to `head` with data intended to surface to humans somewhere (e.g., titles of alternate style sheets as a somewhat esoteric example).

mathiasbynens:
> it seems that ecmarkup just never outputs an `<html>` tag, so the attr was being dropped

Sounds like a bug worth fixing in ecmarkup? It's fine (and good!) to drop `<html>`, except when it has attributes.

bakkot:
Opened https://github.com/tc39/ecmarkup/pull/224 to fix ecmarkup so we can just put it on the html element, where it normally goes.

bakkot:
ecmarkup@3.25.3 has been published with a fix for the above issue. @michaelficarra want to bump it in this PR and add the attribute to `html` instead of `body`?

michaelficarra:
@bakkot done

ljharb:
Explicit > implicit :-)

bakkot:
> (You could still leave out the head and body tags if you wanted, as well as the html closing tag.)

Ecmarkup uses jsdom's `.innerHTML` for rendering its HTML, which will include the closing tags regardless (at least in the version we are [constrained to use](https://github.com/jsdom/jsdom/issues/2968)).

PR #2076, opened 2020-07-04
Editorial: Add missing part of note in Array `find`/`findIndex`/`forEach`

### LINKED ISSUE #2075: Editorial - Array.forEach Note 1 has some missing part
In [array.map](https://tc39.es/ecma262/#sec-array.prototype.map), the note says that new elements in array are not considered after map starts (last para in note 1). 

But a similar note in [array.foreach](https://tc39.es/ecma262/#sec-array.prototype.foreach) doesn't have that part of the note - even though it works in the same way.

bendtherules:
Fixes #2075

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33671,7 +33671,7 @@ THH:mm:ss.sss
           <p>If a _thisArg_ parameter is provided, it will be used as the *this* value for each invocation of _predicate_. If it is not provided, *undefined* is used instead.</p>
           <p>_predicate_ is called with three arguments: the value of the element, the index of the element, and the object being traversed.</p>
           <p>`find` does not directly mutate the object on which it is called but the object may be mutated by the calls to _predicate_.</p>
-          <p>The range of elements processed by `find` is set before the first call to _predicate_. Elements that are appended to the array after the call to `find` begins will not be visited by _predicate_. If existing elements of the array are changed, their value as passed to _predicate_ will be the value at the time that `find` visits them.</p>
+          <p>The range of elements processed by `find` is set before the first call to _predicate_. Elements that are appended to the array after the call to `find` begins will not be visited by _predicate_. If existing elements of the array are changed, their value as passed to _predicate_ will be the value at the time that `find` visits them; elements that are deleted after the call to `find` begins and before being visited are not visited.</p>
         </emu-note>
         <p>When the `find` method is called, the following steps are taken:</p>
         <emu-alg>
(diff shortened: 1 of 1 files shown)


rkirsling:
This is the only Array method with a `callbackfn` / `predicate` first argument that is missing this paragraph, so I think this PR is complete as it is.

That said, it does seem rather odd that `flatMap` has no notes whatsoever; we may wish to address that in a separate PR.

michaelficarra:
@rkirsling clearly, `flatMap` is self-explanatory 🙃

spec.html
```diff
@@ -33789,6 +33789,7 @@ <h1>Array.prototype.forEach ( _callbackfn_ [ , _thisArg_ ] )</h1>
           <p>If a _thisArg_ parameter is provided, it will be used as the *this* value for each invocation of _callbackfn_. If it is not provided, *undefined* is used instead.</p>
           <p>_callbackfn_ is called with three arguments: the value of the element, the index of the element, and the object being traversed.</p>
           <p>`forEach` does not directly mutate the object on which it is called but the object may be mutated by the calls to _callbackfn_.</p>
+          <p>The range of elements processed by `forEach` is set before the first call to _callbackfn_. Elements which are appended to the array after the call to `forEach` begins will not be visited by _callbackfn_. If existing elements of the array are changed, their value as passed to _callbackfn_ will be the value at the time `forEach` visits them; elements that are deleted after the call to `forEach` begins and before being visited are not visited.</p>
```
michaelficarra, line 33792: This "elements that are deleted after the call" portion is missing from `find`/`findIndex`. Should we also fix that?

leobalter:
I'd also consider rephrasing this to something that sounds not much in the passive voice, maybe:

```
The `forEach` method must only store the original range's length of items. It captures the values for each item before their respective _callbackfn_ calls, so the elements may be added, modified or removed within, affecting consecutive calls.

The `forEach` method will limit the loop up to the initial length, and any element added beyond this boundary will be ignored.
```

Maybe this is too much of a bikeshed.

ljharb:
@leobalter i think the intention is to match the existing notes on many other array methods

leobalter:
No problem, thanks!

spec.html
bendtherules: 👍 I'll add that.

bendtherules: Done now

PR #2085, opened 2020-07-08
Editorial: Make Reference type a Record value

### LINKED ISSUE #333: Rewrite References as Records
Hi,

Do you think it is a good idea to rewrite [References](https://tc39.github.io/ecma262/#sec-reference-specification-type) as [Records](https://tc39.github.io/ecma262/#sec-list-and-record-specification-type)?

A Reference would be a Record with the following fields:

| Field | Value | Remarks |
| --- | --- | --- |
| [[base]] | <div>One of: <ul><li>an ECMAScript language value different from **undefined** and **null**; <li>an Environment Record;<li>empty.</ul></div> | A value of empty indicates that the Reference could not be resolved to a binding. _[Note: I have switched from undefined to empty]_ |
| [[referencedName]] | A String or a Symbol |  |
| [[strict]] | A Boolean |  |
| [[thisValue]] | empty or an ECMAScript language value | A Reference whose [[thisValue]] component is not empty is called a _Super Reference_ |

And the following prose:

> Return a value of type Reference whose base value is _undefined_, whose referenced name is _name_, and whose strict reference flag is _strict_.

would become:

> Return Reference{[[base]]: empty, [[referencedName]]: _name_, [[strict]]: **true**, [[thisValue]]: empty}.

or even:

> Return Reference(empty, _name_, **true**).

shvaikalesh:
Resolves #333.

This PR:
* [makes](https://github.com/tc39/ecma262/commit/f3d2ad9180fb2eb5ea00eeb8df314b5c2e99f7c7) Reference type a Record, [moving](https://github.com/tc39/ecma262/commit/5497a598561bbb3a886e99fe59d959da7b2a685d) field description to a Reference Fields table;
* makes `[[Base]]` of an unresolvable reference _absent_ instead of `undefined` (nice improvement suggested by @claudepache);
* [replaces](https://github.com/tc39/ecma262/commit/ac0d4cf47fd919dda9b0f2db09cee7292c058014) (currently inconsistent) prose like "base value component of V" with field access, removing a few now unnecessary helpers ([`GetBase`](https://github.com/tc39/ecma262/commit/2699f7d7a8baa8095cc19167e1a8524e7beca26b), [`GetReferencedName`](https://github.com/tc39/ecma262/commit/46deceff02cab03d93e0be9c9c7a5685a245376a), and [`IsStrictReference`](https://github.com/tc39/ecma262/commit/c21ca861a541ebbe21fb8327ceaf89d715576c72));
* [replaces](https://github.com/tc39/ecma262/commit/7ec703413abe39cbb3618c369359c0d31a3c5413) Reference definitions ("a value of type Reference whose ... component ...") with record literals.

A few drive-by tweaks: [earlier `IsUnresolvableReference` check](https://github.com/tc39/ecma262/commit/2b37baf083509915ef18202e619cddd397872014), [inline error](https://github.com/tc39/ecma262/commit/d0d5346c8325da029c0bbdff6f5122a1a4c1720c), and [typo fix](https://github.com/tc39/ecma262/commit/dacedc4eceecf1596ff139248e637f72f5878c45).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3547,7 +3547,7 @@
     <emu-clause id="sec-completion-record-specification-type" aoid="Completion">
       <h1>The Completion Record Specification Type</h1>
       <p>The Completion type is a Record used to explain the runtime propagation of values and control flow such as the behaviour of statements (`break`, `continue`, `return` and `throw`) that perform nonlocal transfers of control.</p>
-      <p>Values of the Completion type are Record values whose fields are defined as by <emu-xref href="#table-completion-record-fields"></emu-xref>. Such values are referred to as <dfn>Completion Record</dfn>s.</p>
+      <p>Values of the Completion type are Record values whose fields are defined by <emu-xref href="#table-completion-record-fields"></emu-xref>. Such values are referred to as <dfn>Completion Record</dfn>s.</p>
       <emu-table id="table-completion-record-fields" caption="Completion Record Fields" oldids="table-8">
         <table>
           <tbody>
@@ -3718,57 +3718,65 @@
       </emu-clause>
     </emu-clause>
 
-    <emu-clause id="sec-reference-specification-type">
-      <h1>The Reference Specification Type</h1>
-      <emu-note>
-        <p>The Reference type is used to explain the behaviour of such operators as `delete`, `typeof`, the assignment operators, the `super` keyword and other language features. For example, the left-hand operand of an assignment is expected to produce a reference.</p>
-      </emu-note>
-      <p>A <dfn>Reference</dfn> is a resolved name or property binding. A Reference consists of three components, the base value component, the referenced name component, and the Boolean-valued strict reference flag. The base value component is either *undefined*, an Object, a Boolean, a String, a Symbol, a Number, a BigInt, or an Environment Record. A base value component of *undefined* indicates that the Reference could not be resolved to a binding. The referenced name component is a String or Symbol value.</p>
-      <p>A <dfn id="super-reference">Super Reference</dfn> is a Reference that is used to represent a name binding that was expressed using the super keyword. A Super Reference has an additional thisValue component, and its base value component will never be an Environment Record.</p>
-      <p>The following abstract operations are used in this specification to operate on references:</p>
+    <emu-clause id="sec-reference-record-specification-type" oldids="sec-reference-specification-type">
+      <h1>The Reference Record Specification Type</h1>
+      <p>The <dfn>Reference Record</dfn> type is used to explain the behaviour of such operators as `delete`, `typeof`, the assignment operators, the `super` keyword and other language features. For example, the left-hand operand of an assignment is expected to produce a Reference Record.</p>
+      <p>A Reference Record is a resolved name or property binding; its fields are defined by <emu-xref href="#table-reference-record-fields"></emu-xref>.</p>
 
-      <emu-clause id="sec-getbase" aoid="GetBase" oldids="ao-getbase">
-        <h1>GetBase ( _V_ )</h1>
-        <p>The abstract operation GetBase takes argument _V_. It performs the following steps when called:</p>
-        <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. Return the base value component of _V_.
-        </emu-alg>
-      </emu-clause>
+      <emu-table id="table-reference-record-fields" caption="Reference Record Fields">
+        <table>
+          <tbody>
+            <tr>
+              <th>Field Name</th>
+              <th>Value</th>
+              <th>Meaning</th>
+            </tr>
+            <tr>
+              <td oldids="sec-getbase,ao-getbase">[[Base]]</td>
+              <td>
+                One of:
+                <ul>
+                  <li>
+                    any ECMAScript language value except *undefined* or *null*,
+                  </li>
+                  <li>
+                    an Environment Record, or
+                  </li>
+                  <li>
+                    ~unresolvable~.
+                  </li>
+                </ul>
+              </td>
+              <td>The value or Environment Record which holds the binding. A [[Base]] of ~unresolvable~ indicates that the binding could not be resolved.</td>
+            </tr>
+            <tr>
+              <td oldids="sec-getreferencedname,ao-getreferencedname">[[ReferencedName]]</td>
+              <td>String or Symbol</td>
+              <td>The name of the binding. Always a String if [[Base]] value is an Environment Record.</td>
+            </tr>
+            <tr>
+              <td oldids="sec-isstrictreference,ao-isstrictreference">[[Strict]]</td>
+              <td>Boolean</td>
+              <td>*true* if the Reference Record originated in strict mode code, *false* otherwise.</td>
+            </tr>
+            <tr>
+              <td>[[ThisValue]]</td>
+              <td>any ECMAScript language value or ~empty~</td>
+              <td>If not ~empty~, the Reference Record represents a property binding that was expressed using the `super` keyword; it is called a <dfn id="super-reference-record">Super Reference Record</dfn> and its [[Base]] value will never be an Environment Record. In that case, the [[ThisValue]] field holds the *this* value at the time the Reference Record was created.</td>
+            </tr>
+          </tbody>
+        </table>
+      </emu-table>
 
-      <emu-clause id="sec-getreferencedname" aoid="GetReferencedName" oldids="ao-getreferencedname">
-        <h1>GetReferencedName ( _V_ )</h1>
-        <p>The abstract operation GetReferencedName takes argument _V_. It performs the following steps when called:</p>
-        <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. Return the referenced name component of _V_.
-        </emu-alg>
-      </emu-clause>
-
-      <emu-clause id="sec-isstrictreference" aoid="IsStrictReference" oldids="ao-isstrictreference">
-        <h1>IsStrictReference ( _V_ )</h1>
-        <p>The abstract operation IsStrictReference takes argument _V_. It performs the following steps when called:</p>
-        <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. Return the strict reference flag of _V_.
-        </emu-alg>
-      </emu-clause>
-
-      <emu-clause id="sec-hasprimitivebase" aoid="HasPrimitiveBase" oldids="ao-hasprimitivebase">
-        <h1>HasPrimitiveBase ( _V_ )</h1>
-        <p>The abstract operation HasPrimitiveBase takes argument _V_. It performs the following steps when called:</p>
-        <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. If Type(_V_'s base value component) is Boolean, String, Symbol, BigInt, or Number, return *true*; otherwise return *false*.
-        </emu-alg>
-      </emu-clause>
+      <p>The following abstract operations are used in this specification to operate upon References:</p>
 
       <emu-clause id="sec-ispropertyreference" aoid="IsPropertyReference" oldids="ao-ispropertyreference">
         <h1>IsPropertyReference ( _V_ )</h1>
         <p>The abstract operation IsPropertyReference takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. If either the base value component of _V_ is an Object or HasPrimitiveBase(_V_) is *true*, return *true*; otherwise return *false*.
+          1. Assert: _V_ is a Reference Record.
+          1. If _V_.[[Base]] is ~unresolvable~, return *false*.
+          1. If Type(_V_.[[Base]]) is Boolean, String, Symbol, BigInt, Number, or Object, return *true*; otherwise return *false*.
         </emu-alg>
       </emu-clause>
 
@@ -3776,8 +3784,8 @@
         <h1>IsUnresolvableReference ( _V_ )</h1>
         <p>The abstract operation IsUnresolvableReference takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. If the base value component of _V_ is *undefined*, return *true*; otherwise return *false*.
+          1. Assert: _V_ is a Reference Record.
+          1. If _V_.[[Base]] is ~unresolvable~, return *true*; otherwise return *false*.
         </emu-alg>
       </emu-clause>
 
@@ -3785,8 +3793,8 @@
         <h1>IsSuperReference ( _V_ )</h1>
         <p>The abstract operation IsSuperReference takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. If _V_ has a thisValue component, return *true*; otherwise return *false*.
+          1. Assert: _V_ is a Reference Record.
+          1. If _V_.[[ThisValue]] is not ~empty~, return *true*; otherwise return *false*.
         </emu-alg>
       </emu-clause>
 
@@ -3795,17 +3803,15 @@
         <p>The abstract operation GetValue takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. ReturnIfAbrupt(_V_).
-          1. If Type(_V_) is not Reference, return _V_.
-          1. Let _base_ be GetBase(_V_).
+          1. If _V_ is not a Reference Record, return _V_.
           1. If IsUnresolvableReference(_V_) is *true*, throw a *ReferenceError* exception.
           1. If IsPropertyReference(_V_) is *true*, then
-            1. If HasPrimitiveBase(_V_) is *true*, then
-              1. Assert: In this case, _base_ will never be *undefined* or *null*.
-              1. [id="step-getvalue-toobject"] Set _base_ to ! ToObject(_base_).
-            1. Return ? _base_.[[Get]](GetReferencedName(_V_), GetThisValue(_V_)).
+            1. [id="step-getvalue-toobject"] Let _baseObj_ be ! ToObject(_V_.[[Base]]).
+            1. Return ? _baseObj_.[[Get]](_V_.[[ReferencedName]], GetThisValue(_V_)).
           1. Else,
+            1. Let _base_ be _V_.[[Base]].
             1. Assert: _base_ is an Environment Record.
-            1. Return ? _base_.GetBindingValue(GetReferencedName(_V_), IsStrictReference(_V_)) (see <emu-xref href="#sec-environment-records"></emu-xref>).
+            1. Return ? _base_.GetBindingValue(_V_.[[ReferencedName]], _V_.[[Strict]]) (see <emu-xref href="#sec-environment-records"></emu-xref>).
         </emu-alg>
         <emu-note>
           <p>The object that may be created in step <emu-xref href="#step-getvalue-toobject"></emu-xref> is not accessible outside of the above abstract operation and the ordinary object [[Get]] internal method. An implementation might choose to avoid the actual creation of the object.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -4200,46 +4200,62 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
 
-      <emu-clause id="sec-isstrictreference" aoid="IsStrictReference" oldids="ao-isstrictreference">
-        <h1>IsStrictReference ( _V_ )</h1>
-        <p>The abstract operation IsStrictReference takes argument _V_. It performs the following steps when called:</p>
-        <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. Return the strict reference flag of _V_.
-        </emu-alg>
-      </emu-clause>
+      <p>The following abstract operations are used in this specification to operate on references:</p>
```
michaelficarra, line 4251:
```suggestion
      <p>The following abstract operations are used in this specification to operate upon References:</p>
```

spec.html
```diff
@@ -4200,46 +4200,62 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
+                  </li>
+                  <li>
+                    ~empty~.
+                  </li>
+                </ul>
+              </td>
+              <td>~empty~ value indicates that the Reference could not be resolved to a binding.</td>
+            </tr>
+            <tr>
+              <td>[[ReferencedName]]</td>
```
michaelficarra, line 4233: Can we give these IDs or something so that links to `GetReferencedName`, for example, don't break?

spec.html
```diff
@@ -4200,46 +4200,62 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
+                  </li>
+                  <li>
+                    Environment Record;
+                  </li>
+                  <li>
+                    ~empty~.
+                  </li>
+                </ul>
+              </td>
+              <td>~empty~ value indicates that the Reference could not be resolved to a binding.</td>
```
michaelficarra, line 4230: This was previously true of `*undefined*`. Does replacing it with `empty` here have any normative consequences?

spec.html
```diff
@@ -7264,10 +7277,10 @@ <h1>GetIdentifierReference ( _env_, _name_, _strict_ )</h1>
         <p>The abstract operation GetIdentifierReference takes arguments _env_ (an Environment Record or *null*), _name_ (a String), and _strict_ (a Boolean). It performs the following steps when called:</p>
         <emu-alg>
           1. If _env_ is the value *null*, then
-            1. Return a value of type Reference whose base value component is *undefined*, whose referenced name component is _name_, and whose strict reference flag is _strict_.
+            1. Return Reference { [[ReferencedName]]: _name_, [[Strict]]: _strict_ }.
```
michaelficarra, line 7280: All fields of the Reference should be present whenever one is constructed. It's not clear here whether `[[Base]]` should have the value `*undefined*` (as it does in the current text), `empty`, or something else.

spec.html
```diff
@@ -4276,16 +4292,16 @@ <h1>GetValue ( _V_ )</h1>
         <emu-alg>
           1. ReturnIfAbrupt(_V_).
           1. If Type(_V_) is not Reference, return _V_.
-          1. Let _base_ be GetBase(_V_).
           1. If IsUnresolvableReference(_V_) is *true*, throw a *ReferenceError* exception.
+          1. Let _base_ be _V_.[[Base]].
```
syg, line 4296: Any reason this step was moved? It's not observable, but curious.

spec.html
syg: +1, here and below, all fields should be explicitly filled in.

Could make convenience AOs like MakeReference and MakeSuperReference, etc, but it doesn't look like there are that many uses, so spelling out the full Record each time also seems fine to me.

spec.html
```diff
@@ -4276,16 +4292,16 @@ <h1>GetValue ( _V_ )</h1>
...
           1. If IsPropertyReference(_V_) is *true*, then
             1. If HasPrimitiveBase(_V_) is *true*, then
               1. Assert: In this case, _base_ will never be *undefined* or *null*.
```
syg, line 4299: This assert needs to be changed to include `~empty~`, and probably no longer include *undefined* or *null*, since the latter two are never legal values anymore?

spec.html
syg: Good question. We need to audit uses of GetBase. There was one assert that was out of date. There's one case where the result of GetBase is directly passed to ToObject in `delete`. It's not defined what happens if `~empty~` is passed to that.

spec.html
```diff
@@ -14346,18 +14360,18 @@ <h1>Runtime Semantics: Evaluation</h1>
...
           1. If IsUnresolvableReference(_ref_) is *true*, then
-            1. Assert: IsStrictReference(_ref_) is *false*.
+            1. Assert: _ref_.[[Strict]] is *false*.
             1. Return *true*.
           1. If IsPropertyReference(_ref_) is *true*, then
             1. If IsSuperReference(_ref_) is *true*, throw a *ReferenceError* exception.
-            1. Let _baseObj_ be ! ToObject(GetBase(_ref_)).
-            1. Let _deleteStatus_ be ? _baseObj_.[[Delete]](GetReferencedName(_ref_)).
-            1. If _deleteStatus_ is *false* and IsStrictReference(_ref_) is *true*, throw a *TypeError* exception.
+            1. Let _baseObj_ be ! ToObject(_ref_.[[Base]]).
```
syg, line 14367: It's not well-defined what happens if `~empt~` gets passed to ToObject.

claudepache: @syg Since [IsPropertyReference](https://tc39.es/ecma262/#sec-ispropertyreference)(_ref_) is **true**, _ref_.[[Base]] is not `~empty~`.

spec.html
claudepache: @syg. No, this assert concerns specifically what type of primitive value can appear as base when HasPrimitiveBase() is **true**. Even, the PR clears the confusion between “a Reference whose base is a primitive value” (as produced by `true.foo`, and as detected by the HasPrimitiveBase abstract op just above), and “a Reference whose base is absent” (which was confusingly (before this PR) encoded by a specific primitive base value, namely `undefined`).

The point of this assert is that `null.foo` and `undefined.foo` do not produce a Reference whose base is null/undefined, but a TypeError.

spec.html
claudepache: It should not have any normative consequence, unless there were bugs in the previous text. That is to say, before this PR, `undefined` was not used to encode that specific ECMAScript value, but the absence of value.

Specifically, the base of a Reference can be:
* an ECMAScript value different from null/undefined (detected by [IsPropertyReference]):
    * an object,
    * a primitive value different from null/undefined (detected by [HasPrimitiveBase]);
* an environment record;
* nothing (detected by [IsUnresolvableReference]). Before this PR, “nothing” was conventionally encoded as `*undefined*`; after this PR, it is encoded as `~empty~`.

[HasPrimitiveBase]: https://tc39.es/ecma262/#sec-hasprimitivebase
[IsPropertyReference]: https://tc39.es/ecma262/#sec-ispropertyreference
[IsUnresolvableReference]: https://tc39.es/ecma262/#sec-isunresolvablereference

claudepache: @shvaikalesh Other parts of this PR uses “absent” rather than `~empty~` in order to indicate an unresolvable reference. You should pick one of the two options.

spec.html
```diff
@@ -4200,46 +4200,62 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
-      </emu-clause>
+      <p>The following abstract operations are used in this specification to operate on references:</p>
 
       <emu-clause id="sec-hasprimitivebase" aoid="HasPrimitiveBase" oldids="ao-hasprimitivebase">
         <h1>HasPrimitiveBase ( _V_ )</h1>
         <p>The abstract operation HasPrimitiveBase takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: Type(_V_) is Reference.
-          1. If Type(_V_'s base value component) is Boolean, String, Symbol, BigInt, or Number, return *true*; otherwise return *false*.
+          1. If Type(_V_.[[Base]]) is either Boolean, String, Symbol, BigInt, or Number, return *true*; otherwise return *false*.
```
claudepache, line 4258: Although the intent is clear, `Type(_V_.[[Base]])` is not well-defined when `_V_.[[Base]]` is absent/\~empty\~. You should write: “If V.[[Base]] is present/is not \~empty\~ and `Type(_V_.[[Base]]) is etc.”

The same remark applies to IsPropertyReference() below

spec.html
```diff
@@ -4299,22 +4315,21 @@ <h1>PutValue ( _V_, _W_ )</h1>
           1. ReturnIfAbrupt(_V_).
           1. ReturnIfAbrupt(_W_).
           1. If Type(_V_) is not Reference, throw a *ReferenceError* exception.
-          1. Let _base_ be GetBase(_V_).
+          1. Let _base_ be _V_.[[Base]].
```
claudepache, line 4318: If _V_.[[Base]] is absent, then the meaning of this step is not well-defined.

spec.html
```diff
@@ -4200,46 +4200,62 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
+              <td>Binding name. Always a String if [[Base]] value is an Environment Record.</td>
+            </tr>
+            <tr>
+              <td>[[Strict]]</td>
+              <td>Boolean</td>
+              <td>*true* if the Reference originated in strict mode code, *false* otherwise.</td>
+            </tr>
+            <tr>
+              <td>[[ThisValue]]</td>
+              <td>any ECMAScript language value or ~empty~</td>
```
claudepache, line 4244: Same remark as for [[Base]]: You should choose between “\~empty\~” and “absent” and be consistent. (In this table, you use “\~empty\~`; in IsSuperReference() and everywhere else, you use “absent”.)

spec.html
claudepache: > All fields of the Reference should be present

In general, all fields of a Record following a given schema are not necessarily present. For example, a complete [Property Descriptor](https://tc39.es/ecma262/#sec-property-descriptor-specification-type) will either omit its [[Get]] and [[Set]] fields (in case of value property) or omit its [[Value]] field (in case of accessor property).

In the present case, it depends whether the PR author wants to use “\~empty\~” or “absent” in order to encode the relevant information.

ljharb: I suspect, though, that that's the only Record for which all fields are not always present.

spec.html
shvaikalesh: Resolved in 06977c4.

spec.html
shvaikalesh: Resolved in 8145f01.

spec.html
shvaikalesh: @claudepache I've picked `~empty~` in b488801.

spec.html
shvaikalesh: I chose absent/present for `[[ThisValue]]` in 29905b0.

spec.html
shvaikalesh: Resolved in b488801.

spec.html
shvaikalesh: With b488801, `V.[[Base]]` is always present, and 3a705d9 moves `base` declaration after `IsUnresolvableReference` branch, ensuring that `base` is never `~empty~`.

spec.html
shvaikalesh: In order to avoid `base` being `~empty~`; 3a705d9 moves other occurrences.

spec.html
shvaikalesh: With ee73d98 and 5a03a3a, Reference records are always defined with all fields present.

spec.html
```diff
@@ -4278,73 +4278,91 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
 
       <emu-clause id="sec-hasprimitivebase" aoid="HasPrimitiveBase" oldids="ao-hasprimitivebase">
         <h1>HasPrimitiveBase ( _V_ )</h1>
         <p>The abstract operation HasPrimitiveBase takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. If Type(_V_'s base value component) is Boolean, String, Symbol, BigInt, or Number, return *true*; otherwise return *false*.
+          1. Assert: _V_ is a Reference Record.
+          1. If _V_.[[Base]] is ~empty~, then return *false*.
+          1. If Type(_V_.[[Base]]) is either Boolean, String, Symbol, BigInt, or Number, return *true*; otherwise return *false*.
```
ljharb, line 4337: hm, why aren't null/undefined here?

bakkot: There is (as of this PR) no way to produce a Reference whose `[[Base]]` is either `null` or `undefined`, so it wouldn't make sense to include them.

(Prior to this PR, a base value component of `undefined` was used to indicate an unresolvable reference, that is, a reference to an undeclared global name. But there still wasn't a way to have a _property_ reference whose base was `undefined` or `null`, because all of the paths which might have produced such a reference are guarded by `RequireObjectCoercible`.)

spec.html
```diff
@@ -4278,73 +4278,91 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
+      <p>The following abstract operations are used in this specification to operate upon References:</p>
 
       <emu-clause id="sec-hasprimitivebase" aoid="HasPrimitiveBase" oldids="ao-hasprimitivebase">
         <h1>HasPrimitiveBase ( _V_ )</h1>
         <p>The abstract operation HasPrimitiveBase takes argument _V_. It performs the following steps when called:</p>
         <emu-alg>
-          1. Assert: Type(_V_) is Reference.
-          1. If Type(_V_'s base value component) is Boolean, String, Symbol, BigInt, or Number, return *true*; otherwise return *false*.
+          1. Assert: _V_ is a Reference Record.
+          1. If _V_.[[Base]] is ~empty~, then return *false*.
```
bakkot, line 4336: Nit: we generally say `If foo, return *false*.` rather than `If foo, then return *false*.` I'll add a lint rule for this at some point.

spec.html
```diff
@@ -4353,17 +4371,17 @@ <h1>GetValue ( _V_ )</h1>
...
         <emu-alg>
           1. ReturnIfAbrupt(_V_).
-          1. If Type(_V_) is not Reference, return _V_.
-          1. Let _base_ be GetBase(_V_).
+          1. If _V_ is not a Reference Record, return _V_.
           1. If IsUnresolvableReference(_V_) is *true*, throw a *ReferenceError* exception.
+          1. Let _base_ be _V_.[[Base]].
           1. If IsPropertyReference(_V_) is *true*, then
             1. If HasPrimitiveBase(_V_) is *true*, then
               1. Assert: In this case, _base_ will never be *undefined* or *null*.
```
bakkot, line 4379: Since the `[[Base]]` field never holds `*undefined*` or `*null*`, I'm not sure this is very useful as worded. Also, it fails to exclude environment records and `~empty~`. I suggest changing this to something like

```suggestion
              1. Assert: _base_ is an ECMAScript language value other than *undefined* or *null*.
```

spec.html
```diff
@@ -4376,23 +4394,22 @@ <h1>PutValue ( _V_, _W_ )</h1>
...
-              1. Throw a *ReferenceError* exception.
+            1. If _V_.[[Strict]] is *true*, throw a *ReferenceError* exception.
             1. Let _globalObj_ be GetGlobalObject().
-            1. Return ? Set(_globalObj_, GetReferencedName(_V_), _W_, *false*).
-          1. Else if IsPropertyReference(_V_) is *true*, then
+            1. Return ? Set(_globalObj_, _V_.[[ReferencedName]], _W_, *false*).
+          1. Let _base_ be _V_.[[Base]].
+          1. If IsPropertyReference(_V_) is *true*, then
             1. If HasPrimitiveBase(_V_) is *true*, then
               1. Assert: In this case, _base_ will never be *undefined* or *null*.
```
bakkot, line 4405: Ditto above comment about the assertion.

spec.html
shvaikalesh: Fixed in ae8fb93. A lint rule would be great: there are a few `If foo, then return *false*.` occurrences remaining.

spec.html
shvaikalesh: Nice, changed in d8f4575.

spec.html
```diff
@@ -4278,73 +4278,91 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
-      <p>A <dfn id="super-reference">Super Reference</dfn> is a Reference that is used to represent a name binding that was expressed using the super keyword. A Super Reference has an additional thisValue component, and its base value component will never be an Environment Record.</p>
-      <p>The following abstract operations are used in this specification to operate on references:</p>
+      <p>The <dfn>Reference</dfn> type is a Record used to explain the behaviour of such operators as `delete`, `typeof`, the assignment operators, the `super` keyword and other language features. For example, the left-hand operand of an assignment is expected to produce a Reference.</p>
+      <p>A Reference is a resolved name or property binding. Values of the Reference type are Record values whose fields are defined by <emu-xref href="#table-reference-fields"></emu-xref>. Such values are referred to as <dfn>Reference Record</dfn>s.</p>
```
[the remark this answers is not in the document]
shvaikalesh: 1eda3be changes all "Reference" and "Reference value" prose to "Reference Record", including "Super Reference Record", while preserving `oldids`.

We may consider doing the same for [`Completion Record`](https://tc39.es/ecma262/#sec-completion-record-specification-type).

bakkot:
I rebased and pushed a couple of wording tweaks.

bakkot (approved):

spec.html
```diff
@@ -3712,75 +3712,93 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
+                One of:
+                <ul>
+                  <li>
+                    any ECMAScript language value except *undefined* or *null*,
+                  </li>
+                  <li>
+                    an Environment Record, or
+                  </li>
+                  <li>
+                    ~empty~.
```
syg, line 3740: I understand the desire to make `~empty~` a distinguished thing for many use cases, but in this case I feel like renaming this to `~unresolvable~` would be clearer.

spec.html
```diff
@@ -3712,75 +3712,93 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
+                  </li>
+                </ul>
+              </td>
+              <td>The value or environment record which holds the binding. A [[Base]] of ~empty~ indicates that the binding could not be resolved.</td>
```
syg, line 3744:
```suggestion
              <td>The value or Environment Record which holds the binding. A [[Base]] of ~empty~ indicates that the binding could not be resolved.</td>
```

spec.html
```diff
@@ -3789,17 +3807,17 @@ <h1>GetValue ( _V_ )</h1>
...
           1. If IsUnresolvableReference(_V_) is *true*, throw a *ReferenceError* exception.
+          1. Let _base_ be _V_.[[Base]].
           1. If IsPropertyReference(_V_) is *true*, then
             1. If HasPrimitiveBase(_V_) is *true*, then
-              1. Assert: In this case, _base_ will never be *undefined* or *null*.
-              1. [id="step-getvalue-toobject"] Set _base_ to ! ToObject(_base_).
-            1. Return ? _base_.[[Get]](GetReferencedName(_V_), GetThisValue(_V_)).
+              1. Assert: _base_ is an ECMAScript language value other than *undefined* or *null*.
+              1. [id="step-getvalue-toobject"] Let _baseObj_ be ! ToObject(_base_).
+            1. Return ? _baseObj_.[[Get]](_V_.[[ReferencedName]], GetThisValue(_V_)).
```
syg, line 3817: Needs to assign something to _baseObj_ when HasPrimitiveBase is false, or unconditionally call ToObject(_base_).

spec.html
```diff
@@ -6796,10 +6811,10 @@ <h1>GetIdentifierReference ( _env_, _name_, _strict_ )</h1>
         <p>The abstract operation GetIdentifierReference takes arguments _env_ (an Environment Record or *null*), _name_ (a String), and _strict_ (a Boolean). It performs the following steps when called:</p>
         <emu-alg>
           1. If _env_ is the value *null*, then
-            1. Return a value of type Reference whose base value component is *undefined*, whose referenced name component is _name_, and whose strict reference flag is _strict_.
+            1. Return Reference Record { [[Base]]: ~empty~, [[ReferencedName]]: _name_, [[Strict]]: _strict_, [[ThisValue]]: ~empty~ }.
```
syg, line 6814: Here and elsewhere, "Return Reference Record { }" reads odd. I'd prefer "Return the Reference Record { }", since we're referring Reference Records with Record in the type name. (As contrasted with something like Completion Records, which we just call Completions.)

syg (commented):
This is a great improvement and generally looks very good.

I'd like to see the nits below addressed as well as hear from the other editors on their thoughts on naming the sentinel value for [[Base]] `~unresolvable~` instead of `~empty~`.

spec.html
ljharb: Sounds good to me; my only request was that it be a spec value and not `undefined`.

spec.html
shvaikalesh: Fixed in 2e92d58.

spec.html
shvaikalesh: Changed to `~unresolvable~` in 4ca96ec.

spec.html
shvaikalesh: 7a7a585 calls `ToObject` unconditionally, like in step **5.b** of [`delete` operator](https://tc39.es/ecma262/#sec-delete-operator-runtime-semantics-evaluation), and removes `HasPrimitiveBase` because:
* it doesn't seem justified to be in Reference's _public API_, unlike other `Is*Reference` methods;
* it doesn't feel right to query `base` type indirectly via an op.

`!` prefix asserts that `[[Base]]` value is not `undefined` or `null`.

spec.html
```diff
@@ -3712,75 +3712,83 @@ <h1>UpdateEmpty ( _completionRecord_, _value_ )</h1>
...
+            </tr>
+            <tr>
+              <td oldids="sec-isstrictreference,ao-isstrictreference">[[Strict]]</td>
+              <td>Boolean</td>
+              <td>*true* if the Reference Record originated in strict mode code, *false* otherwise.</td>
+            </tr>
+            <tr>
+              <td>[[ThisValue]]</td>
+              <td>any ECMAScript language value or ~empty~</td>
+              <td>If not ~empty~, the Reference Record represents a property binding that was expressed using the `super` keyword; it is called a <dfn id="super-reference-record">Super Reference Record</dfn> and its [[Base]] value will never be an Environment Record. In that case, the [[ThisValue]] field holds the *this* value at time the Reference Record was created.</td>
```
syg, line 3759:
```suggestion
              <td>If not ~empty~, the Reference Record represents a property binding that was expressed using the `super` keyword; it is called a <dfn id="super-reference-record">Super Reference Record</dfn> and its [[Base]] value will never be an Environment Record. In that case, the [[ThisValue]] field holds the *this* value at the time the Reference Record was created.</td>
```

syg (approved):
lgtm, this is a great PR.

dstiers11:
Thanks

spec.html
bakkot: PR to add lint rule: https://github.com/tc39/ecmarkup/pull/264

PR #2089, opened 2020-07-10
Normative: Add Weakrefs and FinalizationRegistry

codehag:
Specification text for https://github.com/tc39/proposal-weakrefs

This version of the specification is _without_ cleanupSome (see https://github.com/codehag/proposal-cleanup-some), in anticipation of the discussion of the upcoming meeting. Adding this back in would be trivial if we decide against it. 

Note: we're adding inline normative optional spec text.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7839,6 +7839,11 @@
             <td>A candidate execution Record</td>
             <td>See the memory model.</td>
           </tr>
+          <tr>
+            <td>[[KeptAlive]]</td>
+            <td>List of objects</td>
+            <td>Initially a new empty List, representing the list of objects to be kept alive until the end of the current Job</td>
+          </tr>
         </tbody>
       </table>
     </emu-table>
@@ -7948,6 +7953,150 @@
     <emu-note>
       <p>This, along with the liveness guarantee in the memory model, ensures that all ~SeqCst~ writes eventually become observable to all agents.</p>
     </emu-note>
+  </emu-clause>
+
+  <emu-clause id="sec-weakref-processing-model">
+    <h1>Processing model of WeakRef and FinalizationRegistry objects</h1>
+    <emu-clause id="sec-weakref-invariants">
+      <h1>Objectives</h1>
+
+      <p> This specification does not make any guarantees that any object will be garbage collected. Objects which are not live may be released after long periods of time, or never at all. For this reason, this specification uses the term "may" when describing behaviour triggered by garbage collection.</p>
+
+      <p>The semantics of WeakRef and FinalizationRegistry objects is based on two operations which happen at particular points in time: </p>
+
+      <ul>
+        <li>
+          When `WeakRef.prototype.deref` is called, the referent (if *undefined* is not returned) is kept alive so that subsequent, synchronous accesses also return the object. This list is reset when synchronous work is done using the ClearKeptObjects abstract operation.
+        </li>
+
+        <li>
+          When an object which is registered with a FinalizationRegistry becomes unreachable, a call of the FinalizationRegistry's cleanup callback may eventually be made, after synchronous ECMAScript execution completes.  The FinalizationRegistry cleanup is performed with the CleanupFinalizationRegistry abstract operation.
+        </li>
+      </ul>
+
+      <p>Neither of these actions (ClearKeptObjects or CleanupFinalizationRegistry) may interrupt synchronous ECMAScript execution. Because hosts may assemble longer, synchronous ECMAScript execution runs, this specification defers the scheduling of ClearKeptObjects and CleanupFinalizationRegistry to the host environment.</p>
+
+      <p>Some ECMAScript implementations include garbage collector implementations which run in the background, including when ECMAScript is idle. Letting the host environment schedule CleanupFinalizationRegistry allows it to resume ECMAScript execution in order to run finalizer work, which may free up held values, reducing overall memory usage.</p>
+    </emu-clause>
+
+    <emu-clause id="sec-liveness">
+      <h1>Liveness</h1>
+
+      <p>For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref of a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
+
+      <emu-note>
+        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep an object alive. Two, that cycles in liveness does not imply that an object is live. To be concrete, if determining _obj_'s liveness depends on determining the liveness of another WeakRef referent, _obj2_, _obj2_'s liveness cannot assume _obj_'s liveness, which would be circular reasoning.
+      </emu-note>
+      <emu-note>
+        WeakRef-obliviousness is defined on sets of objects instead of individual objects to account for cycles. If it were defined on individual objects, then an object in a cycle will be considered live even though its Object value is only observed via WeakRefs of other objects in the cycle.
+      </emu-note>
+      <emu-note>
+        Colloquially, we say that an individual object is live if every set of objects containing it is live.
+      </emu-note>
+
+      <p>At any point during evaluation, a set of objects _S_ is considered <dfn>live</dfn> if either of the following conditions is met: </p>
+
+      <ul>
+        <li>
+          Any element in _S_ is included in any agent's [[KeptAlive]] List.
+        </li>
+        <li>
+          There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes the Object value of any object in _S_.
+        </li>
+      </ul>
+      <emu-note>
+        The intuition the second condition above intends to capture is that an object is live if its identity is observable via non-WeakRef means. An object's identity may be observed by observing a strict equality comparison between objects or observing the object being used as key in a Map.
+      </emu-note>
+      <emu-note>
+        <p>Presence of an object in a field, an internal slot, or a property does not imply that the object is live. For example if the object in question is never passed back to the program, then it cannot be observed.</p>
+
+        <p>This is the case for keys in a WeakMap, members of a WeakSet, as well as the [[WeakRefTarget]] and [[UnregisterToken]] fields of a FinalizationRegistry Cell record.</p>
+
+        <p>The above definition implies that, if a key in a WeakMap is not live, then its corresponding value is not necessarily live either.</p>
+      </emu-note>
+      <emu-note>
+        Liveness is the lower bound for guaranteeing which WeakRefs engines must not empty. Liveness as defined here is undecidable. In practice, engines use conservative approximations such as reachability. There is expected to be significant implementation leeway.
+      </emu-note>
+    </emu-clause>
+
+    <emu-clause id="sec-weakref-execution">
+      <h1>Execution</h1>
+
+      <p>At any time, if a set of objects _S_ is not live, an ECMAScript implementation may perform the following steps atomically:</p>
+
+      <emu-alg>
+        1. For each _obj_ of _S_, do
+          1. For each WeakRef _ref_ such that _ref_.[[WeakRefTarget]] is _obj_, do
+            1. Set _ref_.[[WeakRefTarget]] to ~empty~.
+          1. For each FinalizationRegistry _fg_ such that _fg_.[[Cells]] contains _cell_, and _cell_.[[WeakRefTarget]] is _obj_, do
+            1. Set _cell_.[[WeakRefTarget]] to ~empty~.
+            1. Optionally, perform ! HostEnqueueFinalizationRegistryCleanupJob(_fg_).
+          1. For each WeakMap _map_ such that _map_.[[WeakMapData]] contains a Record value _r_ such that _r_.[[Key]] is _obj_, do
+            1. Set _r_.[[Key]] to ~empty~.
+            1. Set _r_.[[Value]] to ~empty~.
+          1. For each WeakSet _set_ such that _set_.[[WeakSetData]] contains _obj_, do
+            1. Replace the element of _set_ whose value is _obj_ with an element whose value is ~empty~.
+      </emu-alg>
+
+      <emu-note>
+        <p>Together with the definition of liveness, this clause prescribes legal optimizations that an implementation may apply regarding WeakRefs.</p>
+
+        <p>It is possible to access an object without observing its identity. Optimizations such as dead variable elimination and scalar replacement on properties of non-escaping objects whose identity is not observed are allowed. These optimizations are thus allowed to observably empty WeakRefs that point to such objects.</p>
+
+        <p>On the other hand, if an object's identity is observable, and that object is in the [[WeakRefTarget]] internal slot of a WeakRef, optimizations such as rematerialization that observably empty the WeakRef are prohibited.</p>
+
+        <p>Because calling HostEnqueueFinalizationRegistryCleanupJob is optional, registered objects in a FinalizationRegistry do not necessarily hold that FinalizationRegistry live. Implementations may omit FinalizationRegistry callbacks for any reason, e.g., if the FinalizationRegistry itself becomes dead, or if the application is shutting down.</p>
+      </emu-note>
+      <emu-note>
+        <p>Implementations are not obligated to empty WeakRefs for maximal sets of non-live objects.</p>
+        <p>If an implementation chooses a non-live set _S_ in which to empty WeakRefs, it must empty WeakRefs for all objects in _S_ simultaneously. In other words, an implementation must not empty a WeakRef pointing to an object _obj_ without emptying out other WeakRefs that, if not emptied, could result in an execution that observes the Object value of _obj_.</p>
+      </emu-note>
+    </emu-clause>
+
+    <emu-clause id="sec-weakref-host-hooks">
+      <h1>Host Hooks</h1>
+
+      <emu-clause id="sec-host-cleanup-finalization-registry" aoid=HostEnqueueFinalizationRegistryCleanupJob>
+        <h1>HostEnqueueFinalizationRegistryCleanupJob ( _finalizationRegistry_ )</h1>
+
+        <p>The abstract operation HostEnqueueFinalizationRegistryCleanupJob takes argument _finalizationRegistry_ (a FinalizationRegistry). HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that is expected to call CleanupFinalizationRegistry(_finalizationRegistry_) at some point in the future, if possible. The host's responsibility is to make this call at a time which does not interrupt synchronous ECMAScript code execution.</p>
+      </emu-clause>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-clear-kept-objects" aoid=ClearKeptObjects>
+    <h1>ClearKeptObjects ( )</h1>
+    <p>The abstract operation ClearKeptObjects takes no arguments. ECMAScript implementations are expected to call ClearKeptObjects when a synchronous sequence of ECMAScript executions completes. It performs the following steps when called:</p>
+    <emu-alg>
+      1. Let _agentRecord_ be the surrounding agent's Agent Record.
+      1. Set _agentRecord_.[[KeptAlive]] to a new empty List.
+    </emu-alg>
+  </emu-clause>
+
+  <emu-clause id="sec-addtokeptobjects" aoid="AddToKeptObjects">
+    <h1>AddToKeptObjects ( _object_ )</h1>
+    <p>The abstract operation AddToKeptObjects takes argument _object_ (an Object). It performs the following steps when called:</p>
+    <emu-alg>
+      1. Let _agentRecord_ be the surrounding agent's Agent Record.
+      1. Append _object_ to _agentRecord_.[[KeptAlive]].
+    </emu-alg>
+    <emu-note>
+      When the abstract operation AddToKeptObjects is called with a target object reference, it adds the target to a list that will point strongly at the target until ClearKeptObjects is called.
+    </emu-note>
+  </emu-clause>
+
+  <emu-clause id="sec-cleanup-finalization-registry" aoid="CleanupFinalizationRegistry">
+    <h1>CleanupFinalizationRegistry ( _finalizationRegistry_ )</h1>
+    <p>The abstract operation CleanupFinalizationRegistry takes argument _finalizationRegistry_ (a finalization registry object). It performs the following steps when called:</p>
+    <emu-alg>
+      1. Assert: _finalizationRegistry_ has [[Cells]] and [[CleanupCallback]] internal slots.
+      1. Set _callback_ to _finalizationRegistry_.[[CleanupCallback]].
+      1. While _finalizationRegistry_.[[Cells]] contains a Record _cell_ such that _cell_.[[WeakRefTarget]] is ~empty~, an implementation may perform the following steps:
+        1. Choose any such _cell_.
+        1. Remove _cell_ from _finalizationRegistry_.[[Cells]].
+        1. Perform ? Call(_callback_, *undefined*, &laquo; _cell_.[[HeldValue]] &raquo;).
+      1. Return NormalCompletion(*undefined*).
+    </emu-alg>
   </emu-clause>
 </emu-clause>
 
@@ -25676,6 +25825,11 @@
       <p>See <emu-xref href="#sec-native-error-types-used-in-this-standard-evalerror"></emu-xref>.</p>
     </emu-clause>
 
+    <emu-clause id="sec-constructor-properties-of-the-global-object-finnalization-registry">
+      <h1>FinalizationRegistry ( . . . )</h1>
+      <p>See <emu-xref href="#sec-finalization-registry-constructor"></emu-xref>.</p>
+    </emu-clause>
+
     <emu-clause id="sec-float32array">
       <h1>Float32Array ( . . . )</h1>
       <p>See <emu-xref href="#sec-typedarray-constructors"></emu-xref>.</p>
@@ -25804,6 +25958,11 @@
     <emu-clause id="sec-constructor-properties-of-the-global-object-weakmap">
       <h1>WeakMap ( . . . )</h1>
       <p>See <emu-xref href="#sec-weakmap-constructor"></emu-xref>.</p>
+    </emu-clause>
+
+    <emu-clause id="sec-constructor-properties-of-the-global-object-weakref">
+      <h1>WeakRef ( . . . )</h1>
+      <p>See <emu-xref href="#sec-weak-ref-constructor"></emu-xref>.</p>
     </emu-clause>
 
     <emu-clause id="sec-constructor-properties-of-the-global-object-weakset">
@@ -36779,7 +36938,7 @@ THH:mm:ss.sss
 
   <emu-clause id="sec-weakmap-objects">
     <h1>WeakMap Objects</h1>
-    <p>WeakMap objects are collections of key/value pairs where the keys are objects and values may be arbitrary ECMAScript language values. A WeakMap may be queried to see if it contains a key/value pair with a specific key, but no mechanism is provided for enumerating the objects it holds as keys. If an object that is being used as the key of a WeakMap key/value pair is only reachable by following a chain of references that start within that WeakMap, then that key/value pair is inaccessible and is automatically removed from the WeakMap. WeakMap implementations must detect and remove such key/value pairs and any associated resources.</p>
+    <p>WeakMap objects are collections of key/value pairs where the keys are objects and values may be arbitrary ECMAScript language values. A WeakMap may be queried to see if it contains a key/value pair with a specific key, but no mechanism is provided for enumerating the objects it holds as keys. In certain conditions, objects which are not live are removed as WeakMap keys, as described in <emu-xref href="#sec-weakref-execution"></emu-xref>.</p>
     <p>An implementation may impose an arbitrarily determined latency between the time a key/value pair of a WeakMap becomes inaccessible and the time when the key/value pair is removed from the WeakMap. If this latency was observable to ECMAScript program, it would be a source of indeterminacy that could impact program execution. For that reason, an ECMAScript implementation must not provide any means to observe a key of a WeakMap that does not require the observer to present the observed key.</p>
     <p>WeakMap objects must be implemented using either hash tables or other mechanisms that, on average, provide access times that are sublinear on the number of key/value pairs in the collection. The data structure used in this WeakMap objects specification are only intended to describe the required observable semantics of WeakMap objects. It is not intended to be a viable implementation model.</p>
     <emu-note>
@@ -36927,7 +37086,7 @@ THH:mm:ss.sss
 
   <emu-clause id="sec-weakset-objects">
     <h1>WeakSet Objects</h1>
-    <p>WeakSet objects are collections of objects. A distinct object may only occur once as an element of a WeakSet's collection. A WeakSet may be queried to see if it contains a specific object, but no mechanism is provided for enumerating the objects it holds. If an object that is contained by a WeakSet is only reachable by following a chain of references that start within that WeakSet, then that object is inaccessible and is automatically removed from the WeakSet. WeakSet implementations must detect and remove such objects and any associated resources.</p>
+    <p>WeakSet objects are collections of objects. A distinct object may only occur once as an element of a WeakSet's collection. A WeakSet may be queried to see if it contains a specific object, but no mechanism is provided for enumerating the objects it holds. In certain conditions, objects which are not live are removed as WeakSet elements, as described in <emu-xref href="#sec-weakref-execution"></emu-xref>.</p>
     <p>An implementation may impose an arbitrarily determined latency between the time an object contained in a WeakSet becomes inaccessible and the time when the object is removed from the WeakSet. If this latency was observable to ECMAScript program, it would be a source of indeterminacy that could impact program execution. For that reason, an ECMAScript implementation must not provide any means to determine if a WeakSet contains a particular object that does not require the observer to present the observed object.</p>
     <p>WeakSet objects must be implemented using either hash tables or other mechanisms that, on average, provide access times that are sublinear on the number of elements in the collection. The data structure used in this WeakSet objects specification is only intended to describe the required observable semantics of WeakSet objects. It is not intended to be a viable implementation model.</p>
     <emu-note>
@@ -38786,6 +38945,267 @@ THH:mm:ss.sss
       <h1>JSON [ @@toStringTag ]</h1>
       <p>The initial value of the @@toStringTag property is the String value *"JSON"*.</p>
       <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
+    </emu-clause>
+  </emu-clause>
+</emu-clause>
+
+<emu-clause id="sec-managing-memory">
+  <h1>Managing Memory</h1>
+
+  <emu-clause id="sec-weak-ref-objects">
+    <h1>WeakRef Objects</h1>
+    <p>A WeakRef is an object that is used to refer to a target object without preserving it from garbage collection. WeakRefs can be dereferenced to allow access to the target object, if the target object hasn't been reclaimed by garbage collection.</p>
+
+    <emu-clause id="sec-weak-ref-constructor">
+      <h1>The WeakRef Constructor</h1>
+      <p>The WeakRef constructor:</p>
+      <ul>
+        <li>is the intrinsic object %WeakRef%.</li>
+        <li>
+          is the initial value of the *"WeakRef"* property of the global object.
+        </li>
+        <li>
+          creates and initializes a new WeakRef object when called as a constructor.
+        </li>
+        <li>
+          is not intended to be called as a function and will throw an exception when called in that manner.
+        </li>
+        <li>
+          is designed to be subclassable. It may be used as the value in an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified `WeakRef` behaviour must include a `super` call to the `WeakRef` constructor to create and initialize the subclass instance with the internal state necessary to support the `WeakRef.prototype` built-in methods.
+        </li>
+      </ul>
+
+      <emu-clause id="sec-weak-ref-target">
+        <h1>WeakRef ( _target_ )</h1>
+        <p>When the `WeakRef` function is called with argument _target_, the following steps are taken:</p>
+        <emu-alg>
+          1. If NewTarget is *undefined*, throw a *TypeError* exception.
+          1. If Type(_target_) is not Object, throw a *TypeError* exception.
+          1. Let _weakRef_ be ? OrdinaryCreateFromConstructor(NewTarget, `"%WeakRef.prototype%"`, &laquo; [[WeakRefTarget]] &raquo;).
+          1. Perfom ! AddToKeptObjects(_target_).
+          1. Set _weakRef_.[[WeakRefTarget]] to _target_.
+          1. Return _weakRef_.
+        </emu-alg>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-properties-of-the-weak-ref-constructor">
+      <h1>Properties of the WeakRef Constructor</h1>
+      <p>The WeakRef constructor:</p>
+      <ul>
+        <li>
+          has a [[Prototype]] internal slot whose value is the intrinsic object %Function.prototype%.
+        </li>
+        <li>has the following properties:</li>
+      </ul>
+
+      <emu-clause id="sec-weak-ref.prototype">
+        <h1>WeakRef.prototype</h1>
+        <p>The initial value of `WeakRef.prototype` is the intrinsic %WeakRef.prototype% object. </p>
+        <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-properties-of-the-weak-ref-prototype-object">
+      <h1>Properties of the WeakRef Prototype Object</h1>
+      <p>The WeakRef prototype object:</p>
+      <ul>
+        <li>is the intrinsic object <dfn>%WeakRef.prototype%</dfn>.</li>
+        <li>
+          has a [[Prototype]] internal slot whose value is %Object.prototype%.
+        </li>
+        <li>is an ordinary object.</li>
+        <li>does not have a [[WeakRefTarget]] internal slot.</li>
+      </ul>
+
+      <emu-clause id="sec-weak-ref.prototype.constructor" normative-optional>
+        <h1>WeakRef.prototype.constructor</h1>
+
+        <p>The initial value of `WeakRef.prototype.constructor` is the intrinsic object %WeakRef%.</p>
+
+        <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
+      </emu-clause>
+
+      <emu-clause id="sec-weak-ref.prototype.deref">
+        <h1>WeakRef.prototype.deref ( )</h1>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _weakRef_ be the *this* value.
+          1. Perform ? RequireInternalSlot(_weakRef_, [[WeakRefTarget]]).
+          1. Return ! WeakRefDeref(_weakRef_).
+        </emu-alg>
+
+        <emu-note>
+          <p>If the WeakRef returns a _target_ Object that is not *undefined*, then this _target_ object should not be garbage collected until the current execution of ECMAScript code has completed. The AddToKeptObjects operation makes sure read consistency is maintained.</p>
+
+          <pre><code class="javascript">
+            target = { foo: function() {} };
+            let weakRef = new WeakRef(target);
+
+            ... later ...
+
+            if (weakRef.deref()) {
+              weakRef.deref().foo();
+            }
+          </code></pre>
+
+          <p>In the above example, if the first deref does not evaluate to *undefined* then the second deref cannot either.</p>
+        </emu-note>
+      </emu-clause>
+
+      <emu-clause id="sec-weak-ref.prototype-@@tostringtag">
+        <h1>WeakRef.prototype [ @@toStringTag ]</h1>
+        <p>The initial value of the @@toStringTag property is the String value *"WeakRef"*.</p>
+        <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-weakref-abstract-operations">
+      <h1>WeakRef Abstract Operations</h1>
+      <emu-clause id="sec-weakrefderef" aoid="WeakRefDeref">
+        <h1>WeakRefDeref ( _weakRef_ )</h1>
+        <p>The abstract operation WeakRefDeref takes argument _weakRef_ (a WeakRef). It performs the following steps when called:</p>
+        <emu-alg>
+          1. Let _target_ be _weakRef_.[[WeakRefTarget]].
+          1. If _target_ is not ~empty~, then
+            1. Perform ! AddToKeptObjects(_target_).
+            1. Return _target_.
+          1. Return *undefined*.
+        </emu-alg>
+        <emu-note>
+          <p>This abstract operation is defined separately from WeakRef.prototype.deref strictly to make it possible to succinctly define liveness.</p>
+        </emu-note>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-properties-of-weak-ref-instances">
+      <h1>Properties of WeakRef Instances</h1>
+      <p>WeakRef instances are ordinary objects that inherit properties from the WeakRef prototype. WeakRef instances also have a [[WeakRefTarget]] internal slot.</p>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-finalization-registry-objects">
+    <h1>FinalizationRegistry Objects</h1>
+    <p>A FinalizationRegistry is an object that manages registration and unregistration of cleanup operations that are performed when target objects are garbage collected.</p>
+
+    <emu-clause id="sec-finalization-registry-constructor">
+      <h1>The FinalizationRegistry Constructor</h1>
+      <p>The FinalizationRegistry constructor:</p>
+      <ul>
+        <li>is the intrinsic object <dfn>%FinalizationRegistry%</dfn>.</li>
+        <li>
+          is the initial value of the *"FinalizationRegistry"* property of the global object.
+        </li>
+        <li>
+          creates and initializes a new FinalizationRegistry object when called as a constructor.
+        </li>
+        <li>
+          is not intended to be called as a function and will throw an exception when called in that manner.
+        </li>
+        <li>
+          is designed to be subclassable. It may be used as the value in an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified `FinalizationRegistry` behaviour must include a `super` call to the `FinalizationRegistry` constructor to create and initialize the subclass instance with the internal state necessary to support the `FinalizationRegistry.prototype` built-in methods.
+        </li>
+      </ul>
+
+      <emu-clause id="sec-finalization-registry-cleanup-callback">
+        <h1>FinalizationRegistry ( _cleanupCallback_ )</h1>
+        <p>When the `FinalizationRegistry` function is called with argument _cleanupCallback_, the following steps are taken:</p>
+        <emu-alg>
+          1. If NewTarget is *undefined*, throw a *TypeError* exception.
+          1. If IsCallable(_cleanupCallback_) is *false*, throw a *TypeError* exception.
+          1. Let _finalizationRegistry_ be ? OrdinaryCreateFromConstructor(NewTarget, *"%FinalizationRegistry.prototype%"*, &laquo; [[Realm]], [[CleanupCallback]], [[Cells]] &raquo;).
+          1. Let _fn_ be the active function object.
+          1. Set _finalizationRegistry_.[[Realm]] to _fn_.[[Realm]].
+          1. Set _finalizationRegistry_.[[CleanupCallback]] to _cleanupCallback_.
+          1. Set _finalizationRegistry_.[[Cells]] to a new empty List.
+          1. Return _finalizationRegistry_.
+        </emu-alg>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-properties-of-the-finalization-registry-constructor">
+      <h1>Properties of the FinalizationRegistry Constructor</h1>
+      <p>The FinalizationRegistry constructor:</p>
+      <ul>
+        <li>
+          has a [[Prototype]] internal slot whose value is the intrinsic object %Function.prototype%.
+        </li>
+        <li>has the following properties:</li>
+      </ul>
+
+      <emu-clause id="sec-finalization-registry.prototype">
+        <h1>FinalizationRegistry.prototype</h1>
+        <p>The initial value of `FinalizationRegistry.prototype` is the intrinsic %FinalizationRegistry.prototype% object.</p>
+        <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-properties-of-the-finalization-registry-prototype-object">
+      <h1>Properties of the FinalizationRegistry Prototype Object</h1>
+      <p>The FinalizationRegistry prototype object:</p>
+      <ul>
+        <li>is the intrinsic object <dfn>%FinalizationRegistry.prototype%</dfn>.</li>
+        <li>
+          has a [[Prototype]] internal slot whose value is %Object.prototype%.
+        </li>
+        <li>is an ordinary object.</li>
+        <li>
+          does not have [[Cells]] and [[CleanupCallback]] internal slots.
+        </li>
+      </ul>
+
+      <emu-clause id="sec-finalization-registry.prototype.constructor">
+        <h1>FinalizationRegistry.prototype.constructor</h1>
+        <p>The initial value of `FinalizationRegistry.prototype.constructor` is the intrinsic object %FinalizationRegistry%.</p>
+      </emu-clause>
+
+      <emu-clause id="sec-finalization-registry.prototype.register">
+        <h1>FinalizationRegistry.prototype.register ( _target_, _heldValue_ [ , _unregisterToken_ ] )</h1>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _finalizationRegistry_ be the *this* value.
+          1. Perform ? RequireInternalSlot(_finalizationRegistry_, [[Cells]]).
+          1. If Type(_target_) is not Object, throw a *TypeError* exception.
+          1. If SameValue(_target_, _heldValue_) is *true*, throw a *TypeError* exception.
+          1. If Type(_unregisterToken_) is not Object, then
+            1. If _unregisterToken_ is not *undefined*, throw a *TypeError* exception.
+            1. Set _unregisterToken_ to ~empty~.
+          1. Let _cell_ be the Record { [[WeakRefTarget]]: _target_, [[HeldValue]]: _heldValue_, [[UnregisterToken]]: _unregisterToken_ }.
+          1. Append _cell_ to _finalizationRegistry_.[[Cells]].
+          1. Return *undefined*.
+        </emu-alg>
+
+        <emu-note>
+          <p>Based on the algorithms and definitions in this specification, _cell_.[[HeldValue]] is live when _cell_ is in _finalizationRegistry_.[[Cells]]; however, this does not necessarily mean that _cell_.[[UnregisterToken]] or _cell_.[[Target]] are live.  For example, registering an object with itself as its unregister token would not keep the object alive forever.</p>
+        </emu-note>
+      </emu-clause>
+
+      <emu-clause id="sec-finalization-registry.prototype.unregister">
+        <h1>FinalizationRegistry.prototype.unregister ( _unregisterToken_ )</h1>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _finalizationRegistry_ be the *this* value.
+          1. Perform ? RequireInternalSlot(_finalizationRegistry_, [[Cells]]).
+          1. If Type(_unregisterToken_) is not Object, throw a *TypeError* exception.
+          1. Let _removed_ be *false*.
+          1. For each Record { [[WeakRefTarget]], [[HeldValue]], [[UnregisterToken]] } _cell_ that is an element of _finalizationRegistry_.[[Cells]], do
+            1. If _cell_.[[UnregisterToken]] is not ~empty~ and SameValue(_cell_.[[UnregisterToken]], _unregisterToken_) is *true*, then
+              1. Remove _cell_ from _finalizationRegistry_.[[Cells]].
+              1. Set _removed_ to *true*.
+          1. Return _removed_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-finalization-registry.prototype-@@tostringtag">
+        <h1>FinalizationRegistry.prototype [ @@toStringTag ]</h1>
+        <p>The initial value of the @@toStringTag property is the String value *"FinalizationRegistry"*.</p>
+        <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
+      </emu-clause>
+    </emu-clause>
+
+    <emu-clause id="sec-properties-of-finalization-registry-instances">
+      <h1>Properties of FinalizationRegistry Instances</h1>
+      <p>FinalizationRegistry instances are ordinary objects that inherit properties from the FinalizationRegistry prototype. FinalizationRegistry instances also have [[Cells]] and [[CleanupCallback]] internal slots.</p>
     </emu-clause>
   </emu-clause>
 </emu-clause>
@@ -43767,6 +44187,7 @@ THH:mm:ss.sss
   <p>See <emu-xref href="#sec-hosts-and-implementations"></emu-xref> for the definition of host.</p>
   <emu-annex id="sec-host-hooks-summary">
     <h1>Host Hooks</h1>
+    <p><b>HostEnqueueFinalizationRegistryCleanupJob(...)</b></p>
     <p><b>HostEnqueuePromiseJob(...)</b></p>
     <p><b>HostEnsureCanCompileStrings(...)</b></p>
     <p><b>HostFinalizeImportMeta(...)</b></p>


spec.html
```diff
@@ -7875,6 +7880,245 @@ <h1>Forward Progress</h1>
...
+    </emu-clause>
+
+    <emu-clause id="sec-weakref-host-hooks">
+      <h1>Host Hooks</h1>
+
+      <emu-clause id="sec-host-cleanup-finalization-registry" aoid=HostCleanupFinalizationRegistry>
+        <h1>HostCleanupFinalizationRegistry(_finalizationRegistry_)</h1>
+
+        <p>
+          HostCleanupFinalizationRegistry is an implementation-defined abstract
```
codehag, line 8072: @syg This might be important for the incumbent realm tracking.

codehag:
I am not sure what the linter wants from me, any recommendations?

jmdyck:
> I am not sure what the linter wants from me, any recommendations?

There should be a space on either side of each square bracket in that heading, I believe.

spec.html
syg: For incumbents, see https://github.com/tc39/ecma262/pull/2086 and https://github.com/whatwg/html/pull/5722

As both an editor and a delegate, given the reality with Promises, I think it'd be fine to fix up the callbacks in this PR to use JobCallback after #2086 is merged, even for Stage 4. There is no need to gate on that.

bakkot:
Ugh, my apologies for the terrible error messages from ecmarkup. I'll [work on that](https://github.com/tc39/ecmarkup/issues/226).

I think the error you're seeing in the most recent commit is because of the inclusion of blank lines within `<p>`s and an `<emu-note>`. That shouldn't be a hard error (just a lint failure), but the parser is pretty rigid right now.

codehag:
I think I fixed it, should be good now

ljharb:
Tests: https://github.com/tc39/test262/pull/2192

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+  <emu-clause id="sec-weakref-processing-model">
+    <h1>Processing model of WeakRef and FinalizationRegistry objects</h1>
+    <emu-clause id="sec-weakref-invariants">
+      <h1>Objectives</h1>
+
+      <p>
+        This specification does not make any guarantees that any object will be garbage collected.
+        Objects which are unreachable from ECMAScript may be released after long periods
+        of time, or never at all. For this reason, this specification uses the term
+        "may" when describing behaviour triggered by garbage collection.
```
ljharb, line 7893:
```suggestion
        This specification does not make any guarantees that any object will be garbage collected. Objects which are unreachable from ECMAScript may be released after long periods of time, or never at all. For this reason, this specification uses the term "may" when describing behaviour triggered by garbage collection.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+      </p>
+
+      <p>
+        The semantics of WeakRef and FinalizationRegistry objects is based on two
+        operations which happen at particular points in time:
```
ljharb, line 7898:
```suggestion
        The semantics of WeakRef and FinalizationRegistry objects is based on two operations which happen at particular points in time:
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+      </p>
+
+      <ul>
+        <li>
+          When WeakRef.prototype.deref is called, the referent (if *undefined* is not returned)
+          is kept alive so that subsequent, synchronous accesses also
+          return the object. This list is reset when synchronous work is done using
+          the ClearKeptObjects abstract operation.
```
ljharb, line 7906:
```suggestion
          When WeakRef.prototype.deref is called, the referent (if *undefined* is not returned) is kept alive so that subsequent, synchronous accesses also return the object. This list is reset when synchronous work is done using the ClearKeptObjects abstract operation.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+        </li>
+
+        <li>
+          When an object which is registered with a FinalizationRegistry becomes
+          unreachable, a call of the FinalizationRegistry's cleanup callback may
+          eventually be made, after synchronous ECMAScript execution completes.
+          The FinalizationRegistry cleanup is performed with the
+          CleanupFFinalizationRegistry abstract operation.
```
ljharb, line 7914:
```suggestion
          When an object which is registered with a FinalizationRegistry becomes unreachable, a call of the FinalizationRegistry's cleanup callback may eventually be made, after synchronous ECMAScript execution completes. The FinalizationRegistry cleanup is performed with the CleanupFinalizationRegistry abstract operation.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+        </li>
+      </ul>
+
+      <p>
+        Neither of these actions (ClearKeptObjects or CleanupFinalizationRegistry)
+        may interrupt synchronous ECMAScript execution. Because embedding
+        environments may assemble longer, synchronous ECMAScript execution runs,
+        this specification defers the scheduling of ClearKeptObjects and
+        CleanupFinalizationRegistry to the embedding environment.
```
ljharb, line 7923:
```suggestion
        Neither of these actions (ClearKeptObjects or CleanupFinalizationRegistry) may interrupt synchronous ECMAScript execution. Because embedding environments may assemble longer, synchronous ECMAScript execution runs, this specification defers the scheduling of ClearKeptObjects and CleanupFinalizationRegistry to the embedding environment.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+      </p>
+
+      <p>
+        Some ECMAScript implementations include garbage collector implementations
+        which run in the background, including when ECMAScript is idle. Letting the
+        embedding environment schedule CleanupFinalizationRegistry allows it to resume
+        ECMAScript execution in order to run finalizer work, which may free up held values,
+        reducing overall memory usage.
```
ljharb, line 7931:
```suggestion
        Some ECMAScript implementations include garbage collector implementations which run in the background, including when ECMAScript is idle. Letting the embedding environment schedule CleanupFinalizationRegistry allows it to resume ECMAScript execution in order to run finalizer work, which may free up held values, reducing overall memory usage.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+    </emu-clause>
+
+    <emu-clause id="sec-liveness">
+      <h1>Liveness</h1>
+
+      <p>
+        For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn>
+        execution with respect to _S_ is an execution whereby
+        WeakRef.prototype.deref being called on a WeakRef whose referent is an
+        element of _S_ always returns *undefined*.
```
ljharb, line 7942:
```suggestion
        For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby `WeakRef.prototype.deref` being called on a WeakRef whose referent is an element of _S_ always returns *undefined*.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+      </p>
+      <emu-note>
+        WeakRef-obliviousness, together with liveness, capture two notions. One,
+        that a WeakRef itself does not keep an object alive. Two, that cycles in
+        liveness does not imply that an object is live. To be concrete, if
+        determining _obj_'s liveness depends on determining the liveness of
+        another WeakRef referent, _obj2_, _obj2_'s liveness cannot assume
+        _obj_'s liveness, which would beg the question.
```
ljharb, line 7950:
```suggestion
        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep an object alive. Two, that cycles in liveness does not imply that an object is live. To be concrete, if determining _obj_'s liveness depends on determining the liveness of another WeakRef referent, _obj2_, _obj2_'s liveness cannot assume _obj_'s liveness, which would beg the question.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+      </emu-note>
+      <emu-note>
+        WeakRef-obliviousness is defined on sets of objects instead of
+        individual objects to account for cycles. If it were defined on
+        individual objects, then an object in a cycle will be considered live
+        even though its Object value is only observed via WeakRefs of other
+        objects in the cycle.
```
ljharb, line 7957:
```suggestion
        WeakRef-obliviousness is defined on sets of objects instead of individual objects to account for cycles. If it were defined on individual objects, then an object in a cycle will be considered live even though its Object value is only observed via WeakRefs of other objects in the cycle.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+      </emu-note>
+
+      <p>
+        At any point during evaluation, a set of objects _S_ is
+        considered <dfn>live</dfn> if either of the following conditions is
+        met:
```
ljharb, line 7963:
```suggestion
        At any point during evaluation, a set of objects _S_ is considered <dfn>live</dfn> if either of the following conditions is met:
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+      </p>
+
+      <ul>
+        <li>
+          Any element in _S_ is included in any agent's [[KeptAlive]] List.
+        </li>
+        <li>
+          There exists a valid future hypothetical WeakRef-oblivious execution
+          with respect to _S_ that observes the Object value of any object in _S_.
```
ljharb, line 7972:
```suggestion
          There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes the Object value of any object in _S_.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+        </li>
+      </ul>
+      <emu-note>
+        The intuition the second condition above intends to capture is that an
+        object is live if its identity is observable via non-WeakRef means. An
+        object's identity may be observed by observing a strict equality
+        comparison between objects or observing the object being used as key in
+        a Map.
```
ljharb, line 7980:
```suggestion
        The intuition the second condition above intends to capture is that an object is live if its identity is observable via non-WeakRef means. An object's identity may be observed by observing a strict equality comparison between objects or observing the object being used as key in a Map.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+      </emu-note>
+      <emu-note>
+        <p>
+        Presence of an object in a field, an internal slot, or a property does
+        not imply that the object is live. For example if the object in
+        question is never passed back to the program, then it cannot be
+        observed.
```
ljharb, line 7987:
```suggestion
        Presence of an object in a field, an internal slot, or a property does not imply that the object is live. For example if the object in question is never passed back to the program, then it cannot be observed.
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+
+      <p>At any time, if a set of objects _S_ is not live, an ECMAScript implementation may perform the following steps atomically:</p>
+
+      <emu-alg>
+        1. For each _obj_ of _S_, do
+          1. For each WeakRef _ref_ such that _ref_.[[WeakRefTarget]] is _obj_, do
+            1. Set _ref_.[[WeakRefTarget]] to ~empty~.
+          1. For each FinalizationRegistry _fg_ such that _fg_.[[Cells]] contains _cell_, such that _cell_.[[WeakRefTarget]] is _obj_, do
+            1. Set _cell_.[[WeakRefTarget]] to ~empty~.
+            1. Optionally, perform ! HostCleanupFinalizationRegistry(_fg_).
```
devsnek, line 8020: The implementation of `HostCleanupFinalizationRegistry` is already free to not clean things up so I don't think this needs to be optional. 

```suggestion
            1. Perform ! HostCleanupFinalizationRegistry(_fg_).
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+
+        <p>
+          On the other hand, if an object's identity is observable, and that
+          object is in the [[WeakRefTarget]] internal slot of a WeakRef, optimizations
+          such as rematerialization that observably empty the WeakRef are
+          prohibited.
+        </p>
+
+        <p>
+          Because calling HostCleanupFinalizationRegistry is optional, registered objects
```
devsnek, line 8049:
```suggestion
          Because calling CleanupFinalizationRegistry is optional, registered objects
```

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+
+        <p>
+        The above definition implies that, if a key in a WeakMap is not live,
+        then its corresponding value is not necessarily live either.
+        </p>
+      </emu-note>
+      <emu-note>
+        Liveness is the lower bound for guaranteeing which WeakRefs that
+        engines must not empty. In practice, liveness as defined here is
+        undecidable and engines use conservative approximations. There is
```
syg, line 8004: Hm, perhaps "conservative approximations such as reachability"

spec.html
syg: I disagree with this suggestion. While it is true that the host hook reserves the right to not do anything, calling out the optionality means hosts don't need to remember to explicitly allow optionality.

spec.html
syg: I also disagree with this suggestion for the same reason as above.

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+          some point in the future, if possible. The host's responsibility is to
+          make this call at a time which does not interrupt synchronous
+          ECMAScript code execution.
+        </p>
+      </emu-clause>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-clear-kept-objects" aoid=ClearKeptObjects>
+    <h1>ClearKeptObjects ( )</h1>
```
syg, line 8089: This and other AOs here should have the "The abstract operation Foo takes arguments blah [...]" frontmatter.

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
+          in an execution that observes the Object value of _obj_.
+        </p>
+      </emu-note>
+    </emu-clause>
+
+    <emu-clause id="sec-weakref-host-hooks">
+      <h1>Host Hooks</h1>
+
+      <emu-clause id="sec-host-cleanup-finalization-registry" aoid=HostCleanupFinalizationRegistry>
+        <h1>HostCleanupFinalizationRegistry ( _finalizationRegistry_ )</h1>
```
syg, line 8075: To align with HostEnqueuePromiseJob, WDYT about renaming this to HostEnqueueFinalizationRegistryCleanupJob?

codehag: Makes sense to me

spec.html
codehag: Looking at the rest of the spec, does this also apply in cases where the AO has no arguments?

codehag: Also following this i should update CleanupFinalizationRegistry, HostEnqueueFinalizationRegistryCleanupJob etc?

codehag:
I did a first round of fixes in a separate commit so that we can reference the comments and make sure what I did is correct. I will squash this once we finish the review.

spec.html
devsnek: Why don't we just explicitly call out the optionality in the hook then? Isn't the entire point of host hooks to abstract this kind of thing?

spec.html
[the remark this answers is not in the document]
littledan: I was heartened by this wording, personally. I don't know if everyone will understand it, but it's also in a note and the phrase is not core to understanding the note. I think authors should maintain a bit of poetic license with non-normative content.

spec.html
littledan: It's important to call out optionality here to avoid a misunderstanding that it would be mandatory to trigger FinalizationRegistry. The word "optionally" makes the intended interpretation unambiguous, and it does not sacrifice precision/correctness. Remember that specifications are communication devices, not just mathematical artifacts. Because @devsnek's suggestion is an editorial one, we can continue considering it after this PR lands and possibly make the change later.

bakkot: As an editor I also prefer the "Optionally" phrasing.

spec.html
```diff
@@ -7875,6 +7880,251 @@ <h1>Forward Progress</h1>
...
   </emu-clause>
+
+  <emu-clause id="sec-weakref-processing-model">
+    <h1>Processing model of WeakRef and FinalizationRegistry objects</h1>
+    <emu-clause id="sec-weakref-invariants">
+      <h1>Objectives</h1>
+
+      <p>
+        This specification does not make any guarantees that any object will be garbage collected.
+        Objects which are unreachable from ECMAScript may be released after long periods
```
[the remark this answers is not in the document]
bakkot: I'm not sure if it will be quite that simple, since we define liveness for _sets_ of objects. A given object may be a member of both a live set and a non-live set.

spec.html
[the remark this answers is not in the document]
bakkot: Works for me.

spec.html
bakkot: > Looking at the rest of the spec, does this also apply in cases where the AO has no arguments?

Yes, see e.g. [CreateRealm](https://tc39.es/ecma262/#sec-createrealm).

> Also following this i should update CleanupFinalizationRegistry, HostEnqueueFinalizationRegistryCleanupJob etc?

Yup, the intent is to use this phrasing for every abstract operation.

spec.html
```diff
@@ -38592,6 +38771,290 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+    <emu-clause id="sec-properties-of-weak-ref-instances">
+      <h1>Properties of WeakRef Instances</h1>
+      <p>
+        WeakRef instances are ordinary objects that inherit properties from the WeakRef prototype. WeakRef instances also have a [[WeakRefTarget]] internal slot.
+      </p>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-finalization-registry-objects">
+    <h1> FinalizationRegistry Objects </h1>
```
ljharb, line 38912:
```suggestion
    <h1>FinalizationRegistry Objects</h1>
```

spec.html
```diff
@@ -38592,6 +38771,290 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+        <li>
+          is not intended to be called as a function and will throw an exception when called in that manner.
+        </li>
+        <li>
+          is designed to be subclassable. It may be used as the value in an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified `FinalizationRegistry` behaviour must include a `super` call to the `FinalizationRegistry` constructor to create and initialize the subclass instance with the internal state necessary to support the `FinalizationRegistry.prototype` built-in methods.
+        </li>
+      </ul>
+
+      <emu-clause id="sec-finalization-registry-cleanup-callback">
+        <h1> FinalizationRegistry ( _cleanupCallback_ ) </h1>
```
ljharb, line 38937:
```suggestion
        <h1>FinalizationRegistry ( _cleanupCallback_ )</h1>
```

spec.html
```diff
@@ -38592,6 +38771,290 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+    </emu-clause>
+
+    <emu-clause id="sec-properties-of-the-finalization-registry-prototype-object">
+      <h1>Properties of the FinalizationRegistry Prototype Object</h1>
+      <p>The FinalizationRegistry prototype object:</p>
+      <ul>
+        <li>is the intrinsic object <dfn>%FinalizationRegistry.prototype%</dfn>.</li>
+        <li>
+          has a [[Prototype]] internal slot whose value is the intrinsic
+          object %Object.prototype%.
```
ljharb, line 38984:
```suggestion
          has a [[Prototype]] internal slot whose value is the intrinsic object %Object.prototype%.
```

spec.html
```diff
@@ -38592,6 +38771,290 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+        <li>
+          does not have [[Cells]] and [[CleanupCallback]] internal slots.
+        </li>
+      </ul>
+
+      <emu-clause id="sec-finalization-registry.prototype.constructor">
+        <h1>FinalizationRegistry.prototype.constructor</h1>
+        <p>The initial value of
+        `FinalizationRegistry.prototype.constructor` is the intrinsic
+        object %FinalizationRegistry%.</p>
```
ljharb, line 38996:
```suggestion
        <p>The initial value of `FinalizationRegistry.prototype.constructor` is the intrinsic object %FinalizationRegistry%.</p>
```

spec.html
```diff
@@ -38592,6 +38771,290 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+        <emu-alg>
+          1. Let _finalizationRegistry_ be the *this* value.
+          1. Perform ? RequireInternalSlot(_finalizationRegistry_, [[Cells]]).
+          1. If Type(_unregisterToken_) is not Object, throw a *TypeError* exception.
+          1. Let _removed_ be *false*.
+          1. For each Record { [[WeakRefTarget]], [[HeldValue]],
+          [[UnregisterToken]] } _cell_ that is an element of
+          _finalizationRegistry_.[[Cells]], do
+            1. If _cell_.[[UnregisterToken]] is not ~empty~ and
+            SameValue(_cell_.[[UnregisterToken]], _unregisterToken_) is *true*, then
```
ljharb, line 39035:
```suggestion
          1. For each Record { [[WeakRefTarget]], [[HeldValue]], [[UnregisterToken]] } _cell_ that is an element of _finalizationRegistry_.[[Cells]], do
            1. If _cell_.[[UnregisterToken]] is not ~empty~ and SameValue(_cell_.[[UnregisterToken]], _unregisterToken_) is *true*, then
```

spec.html
```diff
@@ -7875,6 +7880,180 @@ <h1>Forward Progress</h1>
...
+      </emu-note>
+      <emu-note>
+        Liveness is the lower bound for guaranteeing which WeakRefs that engines must not empty. In practice, liveness as defined here is undecidable and engines use conservative approximations such as reachability. There is expected to be significant implementation leeway.
+      </emu-note>
+    </emu-clause>
+
+    <emu-clause id="sec-weakref-execution">
+      <h1>Execution</h1>
+
+      <p>At any time, if a set of objects _S_ is not live, an ECMAScript implementation may perform the following steps atomically:</p>
```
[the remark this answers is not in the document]
littledan: My general thought was to use "Object or Symbol" all over the place. If we use the HasIdentity operation as a chosen name, we could say, "values with identity" all over the place instead. In any case, I think this should be a pretty mechanical find-and-replace with whatever wording we decide on.

spec.html
```diff
@@ -7875,6 +7880,180 @@ <h1>Forward Progress</h1>
...
+  </emu-clause>
+
+  <emu-clause id="sec-addtokeptobjects" aoid="AddToKeptObjects">
+    <h1>AddToKeptObjects ( _object_ )</h1>
+    <emu-alg>
+      1. Let _agent_ be the surrounding agent.
+      1. Append _object_ to _agent_.[[KeptAlive]].
+    </emu-alg>
+    <emu-note>
+      When the abstract operation AddToKeptObjects is called with a target object reference, it adds the target to an identity Set that will point strongly at the target until ClearKeptObjects is called.
```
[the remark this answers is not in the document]
littledan: We have this already, as part of the memory model https://tc39.es/ecma262/#sec-set-and-relation-specification-type

spec.html
```diff
@@ -7875,6 +7880,180 @@ <h1>Forward Progress</h1>
...
+      1. Let _agent_ be the surrounding agent.
+      1. Set _agent_.[[KeptAlive]] to a new empty List.
+    </emu-alg>
+  </emu-clause>
+
+  <emu-clause id="sec-addtokeptobjects" aoid="AddToKeptObjects">
+    <h1>AddToKeptObjects ( _object_ )</h1>
+    <emu-alg>
+      1. Let _agent_ be the surrounding agent.
+      1. Append _object_ to _agent_.[[KeptAlive]].
```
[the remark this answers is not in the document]
codehag: Do we do that elsewhere?

spec.html
```diff
@@ -38592,6 +38771,290 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+    <h1> WeakRef Objects </h1>
+    <p>
+      A WeakRef is an object that is used to refer to a target object without preserving it from garbage collection. WeakRefs can dereference to allow access to the target object, if the target object hasn't been reclaimed by garbage collection.
+    </p>
+
+    <emu-clause id="sec-weak-ref-constructor">
+      <h1>The WeakRef Constructor</h1>
+      <p>The WeakRef constructor:</p>
+      <ul>
+        <li>is the intrinsic object %WeakRef%.</li>
```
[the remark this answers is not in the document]
codehag: Ok, waiting for that

spec.html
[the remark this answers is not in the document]
codehag: I will wait for a resolution here before changing this.

spec.html
```diff
@@ -7875,6 +7880,180 @@ <h1>Forward Progress</h1>
...
+    </emu-clause>
+
+    <emu-clause id="sec-liveness">
+      <h1>Liveness</h1>
+
+      <p>
+        For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby WeakRef.prototype.deref being called on a WeakRef whose referent is an element of _S_ always returns *undefined*.
+      </p>
+      <emu-note>
+        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep an object alive. Two, that cycles in liveness does not imply that an object is live. To be concrete, if determining _obj_'s liveness depends on determining the liveness of another WeakRef referent, _obj2_, _obj2_'s liveness cannot assume _obj_'s liveness, which would beg the question.
```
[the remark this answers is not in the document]
syg: How do you find it confusing?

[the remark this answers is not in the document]
ljharb: I'm not particularly familiar with it. I've heard it before, and I have a vague notion of what it means, but I had to google it to be sure. It seems like an idiom rather than a clear technical term, and it seems like an alternative explanation would be more accessible, both globally and to potential readers of the spec.

spec.html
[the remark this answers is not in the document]
ljharb: I don't think we have a clear precedent; but in general I'd prefer, before assuming a slot that has more than one potential type *is* one of those types, to employ those assertions.

however, in this case, it seems like the slot is always a List (it's just replaced with a new List, at times), so the assertion wouldn't add value.

spec.html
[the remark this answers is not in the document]
syg: In that case I'm also happy with "which would be circular."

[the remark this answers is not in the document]
syg: To clarify, that would be an alternative phrasing but not an alternative explanation. The reasoning in the explanation has no alternatives.

spec.html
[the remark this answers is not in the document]
ljharb: Ah, thanks. Probably a poor editorial choice to use the same name as a JS value :-) i'll look into updating that separately.

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
   </emu-clause>
+
+  <emu-clause id="sec-weakref-processing-model">
+    <h1>Processing model of WeakRef and FinalizationRegistry objects</h1>
+    <emu-clause id="sec-weakref-invariants">
+      <h1>Objectives</h1>
+
+      <p>
+        This specification does not make any guarantees that any object will be garbage collected.
+        Objects which are not live may be released after long periods of time, or never at all. For this reason, this specification uses the term "may" when describing behaviour triggered by garbage collection.
```
ljharb, line 7891:
```suggestion
        This specification does not make any guarantees that any object will be garbage collected. Objects which are not live may be released after long periods of time, or never at all. For this reason, this specification uses the term "may" when describing behaviour triggered by garbage collection.
```
(unless it's intended to be separate paragraphs, in which case it'd need separate `<p>` tags to render as such)

spec.html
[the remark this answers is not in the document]
ljharb: As is, my reaction is "would beg *what* question?". With your suggestion, it's tautological but also makes sense to me, so I'm happy with that change.

spec.html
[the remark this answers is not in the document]
syg: Continuing from @ljharb's comment above since this was where the actually line was.

To get around the confusion, "which would beg the question" -> "which would be circular reasoning" should be fine. Slightly less precise, but gets the point across.

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
+        <p>
+        This is the case for keys in a WeakMap, members of a WeakSet, as well as the [[WeakRefTarget]] and [[UnregisterToken]] fields of a FinalizationRegistry Cell record.
+        </p>
+
+        <p>
+        The above definition implies that, if a key in a WeakMap is not live, then its corresponding value is not necessarily live either.
+        </p>
+      </emu-note>
+      <emu-note>
+        Liveness is the lower bound for guaranteeing which WeakRefs that engines must not empty. In practice, liveness as defined here is undecidable and engines use conservative approximations such as reachability. There is expected to be significant implementation leeway.
```
syg, line 7959: Sorry, one more round of rewording nit for this sentence:

"In practice, liveness as defined here is undecidable and engines use conservative approximations such as reachability."

->

"Liveness as defined here is undecidable. In practice, engines use conservative approximations such as reachability."

spec.html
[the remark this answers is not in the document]
syg: I don't expect adding new kinds of values with identity in the near future, and since there are only two, I'm leaning towards "Object or Symbol" over "values with identity", since "values with identity" is also 3 words and would need to be enumerated in another section that the reader clicks through to find what are the values with identity anyway.

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
+    <emu-clause id="sec-weakref-execution">
+      <h1>Execution</h1>
+
+      <p>At any time, if a set of objects _S_ is not live, an ECMAScript implementation may perform the following steps atomically:</p>
+
+      <emu-alg>
+        1. For each _obj_ of _S_, do
+          1. For each WeakRef _ref_ such that _ref_.[[WeakRefTarget]] is _obj_, do
+            1. Set _ref_.[[WeakRefTarget]] to ~empty~.
+          1. For each FinalizationRegistry _fg_ such that _fg_.[[Cells]] contains _cell_, such that _cell_.[[WeakRefTarget]] is _obj_, do
```
syg, line 7972: Two consecutive "such that" sounds weird, should be "and". How about "such that _fg_.[[Cells]] contains _cell_ and _cell_.[[WeakRefTarget]] is _obj_,"?

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
+            1. Set _cell_.[[WeakRefTarget]] to ~empty~.
+            1. Optionally, perform ! HostEnqueueFinalizationRegistryCleanupJob(_fg_).
+          1. For each WeakMap _map_ such that _map_.[[WeakMapData]] contains a record _r_ such that _r_.[[Key]] is _obj_, do
```
syg, line 7975: "record" -> "a Record value"

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
+          The abstract operation HostEnqueueFinalizationRegistryCleanupJob takes argument _finalizationRegistry_ (a FinalizationRegistry). HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that is expected to call CleanupFinalizationRegistry(_finalizationRegistry_) at some point in the future, if possible. The host's responsibility is to make this call at a time which does not interrupt synchronous ECMAScript code execution.
+        </p>
+      </emu-clause>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-clear-kept-objects" aoid=ClearKeptObjects>
+    <h1>ClearKeptObjects ( )</h1>
+    <p>
+      The abstract operation CreateRealm takes no arguments. ECMAScript implementations are expected to call ClearKeptObjects when a synchronous sequence of ECMAScript execution completes.
```
syg, line 8024: Copypasta'd CreateRealm.

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
+    </p>
+    <p>The following steps are performed:</p>
+    <emu-alg>
+      1. Let _agent_ be the surrounding agent.
```
syg, line 8028:
```suggestion
      1. Let _agentRecord_ be the surrounding agent's Agent Record.
```

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
+      1. Set _agent_.[[KeptAlive]] to a new empty List.
+    </emu-alg>
+  </emu-clause>
+
+  <emu-clause id="sec-addtokeptobjects" aoid="AddToKeptObjects">
+    <h1>AddToKeptObjects ( _object_ )</h1>
+    <p> The abstract operation AddToKeptObjects takes argument _object_ (an object). The following steps are performed: </p>
+    <emu-alg>
+      1. Let _agent_ be the surrounding agent.
```
syg, line 8037: Ditto on getting the Agent Record.

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
+    <p>The following steps are performed:</p>
+    <emu-alg>
+      1. Let _agent_ be the surrounding agent.
+      1. Set _agent_.[[KeptAlive]] to a new empty List.
+    </emu-alg>
+  </emu-clause>
+
+  <emu-clause id="sec-addtokeptobjects" aoid="AddToKeptObjects">
+    <h1>AddToKeptObjects ( _object_ )</h1>
+    <p> The abstract operation AddToKeptObjects takes argument _object_ (an object). The following steps are performed: </p>
```
syg, line 8035:
```suggestion
    <p> The abstract operation AddToKeptObjects takes argument _object_ (an Object). The following steps are performed: </p>
```

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
+    <emu-alg>
+      1. Let _agent_ be the surrounding agent.
+      1. Append _object_ to _agent_.[[KeptAlive]].
+    </emu-alg>
+    <emu-note>
+      When the abstract operation AddToKeptObjects is called with a target object reference, it adds the target to an identity Set that will point strongly at the target until ClearKeptObjects is called.
```
syg, line 8041: This comment suggests that you want to use a Set for [[KeptAlive]] instead of a List. That sounds fine to me, but isn't what this PR currently does.

spec.html
```diff
@@ -38592,6 +38772,282 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+
+            ... later ...
+
+            if (weakRef.deref()) {
+              weakRef.deref().foo();
+            }
+          </pre>
+
+          <p>
+          In the above example, if the first deref evaluates to true then the second deref can not fail.
```
syg, line 38890:
```suggestion
          In the above example, if the first deref evaluates to true then the second deref cannot return *undefined*.
```

spec.html
```diff
@@ -38592,6 +38772,282 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+        <emu-alg>
+          1. Let _finalizationRegistry_ be the *this* value.
+          1. Perform ? RequireInternalSlot(_finalizationRegistry_, [[Cells]]).
+          1. If Type(_target_) is not Object, throw a *TypeError* exception.
+          1. If SameValue(_target_, _heldValue_), throw a *TypeError* exception.
+          1. If Type(_unregisterToken_) is not Object, then
+            1. If _unregisterToken_ is not *undefined*, throw a *TypeError* exception.
+            1. Set _unregisterToken_ to ~empty~.
+          1. Let _cell_ be the Record { [[WeakRefTarget]] : _target_, [[HeldValue]]: _heldValue_,
+          [[UnregisterToken]]: _unregisterToken_ }.
```
syg, line 39007: Stray linebreak here, join into above line.

spec.html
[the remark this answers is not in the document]
ljharb: the primary advantage i see to coming up with a single term for it is that there'd be a linkable dfn which serves as the single source of truth for that that means. perhaps since it'll only ever be two items, it's not that important.

spec.html
```diff
@@ -7875,6 +7880,181 @@ <h1>Forward Progress</h1>
...
+          When an object which is registered with a FinalizationRegistry becomes unreachable, a call of the FinalizationRegistry's cleanup callback may eventually be made, after synchronous ECMAScript execution completes.  The FinalizationRegistry cleanup is performed with the CleanupFinalizationRegistry abstract operation.
+        </li>
+      </ul>
+
+      <p>
+        Neither of these actions (ClearKeptObjects or CleanupFinalizationRegistry) may interrupt synchronous ECMAScript execution. Because embedding environments may assemble longer, synchronous ECMAScript execution runs, this specification defers the scheduling of ClearKeptObjects and CleanupFinalizationRegistry to the host environment.
+      </p>
+
+      <p>
+        Some ECMAScript implementations include garbage collector implementations which run in the background, including when ECMAScript is idle. Letting the embedding environment schedule CleanupFinalizationRegistry allows it to resume ECMAScript execution in order to run finalizer work, which may free up held values, reducing overall memory usage.
```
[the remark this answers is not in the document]
codehag: I change the above instance as well.

spec.html
codehag: I have changed it to "a list"

codehag:
Branch has also been rebased

spec.html
```diff
@@ -7872,6 +7877,180 @@ <h1>Forward Progress</h1>
...
+
+  <emu-clause id="sec-clear-kept-objects" aoid=ClearKeptObjects>
+    <h1>ClearKeptObjects ( )</h1>
+    <p>
+      The abstract operation ClearKeptObjects takes no arguments. ECMAScript implementations are expected to call ClearKeptObjects when a synchronous sequence of ECMAScript execution completes.
+    </p>
+    <p>The following steps are performed:</p>
+    <emu-alg>
+      1. Let _agentRecord_ be the surrounding agent's Agent Record.
+      1. Set _agent_.[[KeptAlive]] to a new empty List.
```
syg, line 8025:
```suggestion
      1. Set _agentRecord_.[[KeptAlive]] to a new empty List.
```

spec.html
```diff
@@ -7872,6 +7877,180 @@ <h1>Forward Progress</h1>
...
+    </emu-alg>
+  </emu-clause>
+
+  <emu-clause id="sec-addtokeptobjects" aoid="AddToKeptObjects">
+    <h1>AddToKeptObjects ( _object_ )</h1>
+    <p> The abstract operation AddToKeptObjects takes argument _object_ (an Object). The following steps are performed: </p>
+    <emu-alg>
+      1. Let _agentRecord_ be the surrounding agent's Agent Record.
+      1. Append _object_ to _agent_.[[KeptAlive]].
```
syg, line 8034:
```suggestion
      1. Append _object_ to _agentRecord_.[[KeptAlive]].
```

spec.html
```diff
@@ -38589,6 +38768,281 @@ <h1>JSON [ @@toStringTag ]</h1>
   </emu-clause>
 </emu-clause>
 
+<emu-clause id="sec-managing-memory">
+  <h1>Managing Memory</h1>
```
syg, line 38772: (No action needed for this PR)

To the other editors @bakkot @ljharb @michaelficarra: we should have links to WeakMap and WeakSet in this new section.

spec.html
```diff
@@ -7872,6 +7877,180 @@ <h1>Forward Progress</h1>
...
+      </p>
+      <emu-note>
+        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep an object alive. Two, that cycles in liveness does not imply that an object is live. To be concrete, if determining _obj_'s liveness depends on determining the liveness of another WeakRef referent, _obj2_, _obj2_'s liveness cannot assume _obj_'s liveness, which would be circular reasoning.
+      </emu-note>
+      <emu-note>
+        WeakRef-obliviousness is defined on sets of objects instead of individual objects to account for cycles. If it were defined on individual objects, then an object in a cycle will be considered live even though its Object value is only observed via WeakRefs of other objects in the cycle.
+      </emu-note>
+
+      <p>
+        At any point during evaluation, a set of objects _S_ is considered <dfn>live</dfn> if either of the following conditions is met:
```
bakkot, line 7927: This definition is for sets of objects, but elsewhere (in notes) we speak of individual objects being live. Thoughts on adding a NOTE which says something like "Colloquially, we say that an individual object is live if every set of objects containing it is live."?

spec.html
```diff
@@ -38589,6 +38768,281 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+            1. If _unregisterToken_ is not *undefined*, throw a *TypeError* exception.
+            1. Set _unregisterToken_ to ~empty~.
+          1. Let _cell_ be the Record { [[WeakRefTarget]] : _target_, [[HeldValue]]: _heldValue_, [[UnregisterToken]]: _unregisterToken_ }.
+          1. Append _cell_ to _finalizationRegistry_.[[Cells]].
+          1. Return *undefined*.
+        </emu-alg>
+
+        <emu-note>
+          <p>
+          Based on the algorithms and definitions in this specification, across the time when _cell_ is in _finalizationRegistry_.[[Cells]], then _cell_.[[Holdings]] is live; however, _cell_.[[UnregisterToken]] and _cell_.[[Target]] are not necessarily live.  For example, registering an object with itself as its unregister token would not keep the object alive forever.
```
bakkot, line 39009: `[[Holdings]]` does not appear to be a thing. I assume this should be `[[HeldValue]]`?

Also, nit: `across the time when _cell_ is in _finalizationRegistry_.[[Cells]], then _cell_.[[Holdings]] is live` does not read very clearly to me. Can I suggest

```suggestion
          Based on the algorithms and definitions in this specification, _cell_.[[HeldValue]] is live while _cell_ is in _finalizationRegistry_.[[Cells]]; however, this does not necessarily mean that _cell_.[[UnregisterToken]] or _cell_.[[Target]] are live.  For example, registering an object with itself as its unregister token would not keep the object alive forever.
```

spec.html
```diff
@@ -38589,6 +38768,281 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+            1. If _cell_.[[UnregisterToken]] is not ~empty~ and SameValue(_cell_.[[UnregisterToken]], _unregisterToken_) is *true*, then
+              1. Remove _cell_ from _finalizationRegistry_.[[Cells]].
+              1. Set _removed_ to *true*.
+          1. Return _removed_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-finalization-registry.prototype-@@tostringtag">
+        <h1>FinalizationRegistry.prototype [ @@toStringTag ]</h1>
+        <p>The initial value of the @@toStringTag property is the String value `"FinalizationRegistry"`.</p>
```
bakkot, line 39032:
```suggestion
        <p>The initial value of the @@toStringTag property is the String value *"FinalizationRegistry"*.</p>
```

spec.html
```diff
@@ -7872,6 +7877,180 @@ <h1>Forward Progress</h1>
...
+          The abstract operation HostEnqueueFinalizationRegistryCleanupJob takes argument _finalizationRegistry_ (a FinalizationRegistry). HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that is expected to call CleanupFinalizationRegistry(_finalizationRegistry_) at some point in the future, if possible. The host's responsibility is to make this call at a time which does not interrupt synchronous ECMAScript code execution.
+        </p>
+      </emu-clause>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-clear-kept-objects" aoid=ClearKeptObjects>
+    <h1>ClearKeptObjects ( )</h1>
+    <p>
+      The abstract operation ClearKeptObjects takes no arguments. ECMAScript implementations are expected to call ClearKeptObjects when a synchronous sequence of ECMAScript execution completes.
```
bakkot, line 8020: `a synchronous sequence of ECMAScript execution` reads strangely to me - should that be "executions"? Or we could find another way to phrase this.

spec.html
```diff
@@ -38589,6 +38768,281 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+          1. Let _target_ be the value of _weakRef_.[[WeakRefTarget]].
+          1. If _target_ is not ~empty~, then
+            1. Perform ! AddToKeptObjects(_target_).
+            1. Return _target_.
+          1. Return *undefined*.
+        </emu-alg>
+
+        <emu-note>
+          <p>
+          If the WeakRef returns a _target_ Object that is not *undefined*, then this _target_ object should not be garbage collected in that turn. The AddToKeptObjects operation makes sure read consistency is maintained.
```
bakkot, line 38871: "turn" is not a term that ECMA-262 uses elsewhere and I would be hesitant to introduce it here without providing a definition.

spec.html
```diff
@@ -38589,6 +38768,281 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+        <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
+      </emu-clause>
+
+      <emu-clause id="sec-weak-ref.prototype.deref">
+        <h1>WeakRef.prototype.deref ( )</h1>
+        <p>The following steps are taken:</p>
+        <emu-alg>
+          1. Let _weakRef_ be the *this* value.
+          1. Perform ? RequireInternalSlot(_weakRef_, [[WeakRefTarget]]).
+          1. Let _target_ be the value of _weakRef_.[[WeakRefTarget]].
```
bakkot, line 38862:
```suggestion
          1. Let _target_ be _weakRef_.[[WeakRefTarget]].
```

spec.html
```diff
@@ -38589,6 +38768,281 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+
+            ... later ...
+
+            if (weakRef.deref()) {
+              weakRef.deref().foo();
+            }
+          </pre>
+
+          <p>
+          In the above example, if the first deref evaluates to true then the second deref cannot return *undefined*.
```
bakkot, line 38886: "evaluates to true" is an odd thing to say, since the deref itself can only return an object ro `*undefined*`. Perhaps

```suggestion
          In the above example, if the first deref does not evaluate to *undefined* then the second deref cannot either.
```

spec.html
```diff
@@ -38589,6 +38768,281 @@ <h1>JSON [ @@toStringTag ]</h1>
...
+          </p>
+        </emu-note>
+      </emu-clause>
+
+      <emu-clause id="sec-weak-ref.prototype-@@tostringtag">
+        <h1>WeakRef.prototype [ @@toStringTag ]</h1>
+        <p>The initial value of the @@toStringTag property is the String value `"WeakRef"`.</p>
```
bakkot, line 38893:
```suggestion
        <p>The initial value of the @@toStringTag property is the String value *"WeakRef"*.</p>
```

spec.html
```diff
@@ -7872,6 +7877,180 @@ <h1>Forward Progress</h1>
...
+      When the abstract operation AddToKeptObjects is called with a target object reference, it adds the target to a list that will point strongly at the target until ClearKeptObjects is called.
+    </emu-note>
+  </emu-clause>
+
+  <emu-clause id="sec-cleanup-finalization-registry" aoid="CleanupFinalizationRegistry">
+    <h1> CleanupFinalizationRegistry ( _finalizationRegistry_ [ , _callback_ ] ) </h1>
+    <p> The abstract operation CleanupFinalizationRegistry takes argument _finalizationRegistry_ (a finalization registry object) and optional argument _callback_ (a function object). The following steps are performed: </p>
+    <emu-alg>
+      1. Assert: _finalizationRegistry_ has [[Cells]] and [[CleanupCallback]] internal slots.
+      1. If _callback_ is not present or *undefined*, set _callback_ to _finalizationRegistry_.[[CleanupCallback]].
```
bakkot, line 8046: "is not present or undefined" could be read as either "is not (present or undefined)" or "is (not present) or undefined". I assume it should be

```suggestion
      1. If _callback_ is not present or is *undefined*, set _callback_ to _finalizationRegistry_.[[CleanupCallback]].
```

spec.html
```diff
@@ -7872,6 +7877,180 @@ <h1>Forward Progress</h1>
...
+      1. While _finalizationRegistry_.[[Cells]] contains a Record _cell_ such that _cell_.[[WeakRefTarget]] is ~empty~, then an implementation may perform the following steps:
```
bakkot, line 8047: English does not normally use "while-then", in my experience.

```suggestion
      1. While _finalizationRegistry_.[[Cells]] contains a Record _cell_ such that _cell_.[[WeakRefTarget]] is ~empty~, an implementation may perform the following steps:
```

spec.html
```diff
@@ -38589,6 +38768,281 @@ <h1>JSON [ @@toStringTag ]</h1>
   </emu-clause>
 </emu-clause>
 
+<emu-clause id="sec-managing-memory">
+  <h1>Managing Memory</h1>
+
+  <emu-clause id="sec-weak-ref-objects">
+    <h1>WeakRef Objects</h1>
+    <p>
+      A WeakRef is an object that is used to refer to a target object without preserving it from garbage collection. WeakRefs can dereference to allow access to the target object, if the target object hasn't been reclaimed by garbage collection.
```
bakkot, line 38777:
```suggestion
      A WeakRef is an object that is used to refer to a target object without preserving it from garbage collection. WeakRefs can be dereferenced to allow access to the target object if the target object hasn't been reclaimed by garbage collection.
```

spec.html
```diff
@@ -7872,6 +7877,180 @@ <h1>Forward Progress</h1>
...
+        1. For each _obj_ of _S_, do
+          1. For each WeakRef _ref_ such that _ref_.[[WeakRefTarget]] is _obj_, do
+            1. Set _ref_.[[WeakRefTarget]] to ~empty~.
+          1. For each FinalizationRegistry _fg_ such that _fg_.[[Cells]] contains _cell_, and _cell_.[[WeakRefTarget]] is _obj_, do
+            1. Set _cell_.[[WeakRefTarget]] to ~empty~.
+            1. Optionally, perform ! HostEnqueueFinalizationRegistryCleanupJob(_fg_).
+          1. For each WeakMap _map_ such that _map_.[[WeakMapData]] contains a Record value _r_ such that _r_.[[Key]] is _obj_, do
+            1. Remove _r_ from _map_.[[WeakMapData]].
+          1. For each WeakSet _set_ such that _set_.[[WeakSetData]] contains _obj_, do
+            1. Remove _obj_ from _set_.[[WeakSetData]].
```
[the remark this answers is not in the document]
codehag: Done, hopefully as you intended? Please double check.

spec.html
codehag: Changed this to executions

spec.html
codehag: do you have any alternatives you would suggest here?

spec.html
codehag: I have added this note, pending comments from the other editors.

spec.html
```diff
@@ -7872,6 +7877,180 @@ <h1>Forward Progress</h1>
...
+
+      <emu-alg>
+        1. For each _obj_ of _S_, do
+          1. For each WeakRef _ref_ such that _ref_.[[WeakRefTarget]] is _obj_, do
+            1. Set _ref_.[[WeakRefTarget]] to ~empty~.
+          1. For each FinalizationRegistry _fg_ such that _fg_.[[Cells]] contains _cell_, and _cell_.[[WeakRefTarget]] is _obj_, do
+            1. Set _cell_.[[WeakRefTarget]] to ~empty~.
+            1. Optionally, perform ! HostEnqueueFinalizationRegistryCleanupJob(_fg_).
+          1. For each WeakMap _map_ such that _map_.[[WeakMapData]] contains a Record value _r_ such that _r_.[[Key]] is _obj_, do
+            1. Remove _r_ from _map_.[[WeakMapData]].
```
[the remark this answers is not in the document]
codehag: Done, hopefully as you intended? Please double check.

codehag:
@bakkot please check in your second pass if my changes adequately cover your concerns.

spec.html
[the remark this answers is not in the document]
littledan: Can we move this discussion to the [Symbols as WeakMap keys repo](https://github.com/tc39/proposal-symbols-as-weakmap-keys)? This PR will not be one where we land any wording on this question. I encourage people who have opinions here to file an issue. I hope that this PR will land before that proposal goes to Stage 3, and so we have plenty of time to work out the wording.

littledan:
> lgtm, pending wordsmithing "Objects or Symbols" vs "values with identity"

For this PR, let's stick with "Objects". Symbols as WeakMap keys is at Stage 1, not landing with this PR.

spec.html
```diff
@@ -8007,44 +7981,40 @@ <h1>Host Hooks</h1>
...
     <emu-note>
       When the abstract operation AddToKeptObjects is called with a target object reference, it adds the target to a list that will point strongly at the target until ClearKeptObjects is called.
     </emu-note>
   </emu-clause>
 
   <emu-clause id="sec-cleanup-finalization-registry" aoid="CleanupFinalizationRegistry">
-    <h1> CleanupFinalizationRegistry ( _finalizationRegistry_ [ , _callback_ ] ) </h1>
-    <p> The abstract operation CleanupFinalizationRegistry takes argument _finalizationRegistry_ (a finalization registry object) and optional argument _callback_ (a function object). The following steps are performed: </p>
+    <h1>CleanupFinalizationRegistry ( _finalizationRegistry_ [ , _callback_ ] )</h1>
+    <p>The abstract operation CleanupFinalizationRegistry takes argument _finalizationRegistry_ (a finalization registry object) and optional argument _callback_ (a function object). The following steps are performed:</p>
```
bakkot, line 8013:
```suggestion
    <p>The abstract operation CleanupFinalizationRegistry takes argument _finalizationRegistry_ (a finalization registry object) and optional argument _callback_ (a function object). It performs the following steps when called:</p>
```

spec.html
```diff
@@ -38773,17 +38743,15 @@ <h1>Managing Memory</h1>
...
+    <p>A WeakRef is an object that is used to refer to a target object without preserving it from garbage collection. WeakRefs can be dereferenced to allow access to the target object, if the target object hasn't been reclaimed by garbage collection.</p>
 
     <emu-clause id="sec-weak-ref-constructor">
       <h1>The WeakRef Constructor</h1>
       <p>The WeakRef constructor:</p>
       <ul>
         <li>is the intrinsic object %WeakRef%.</li>
         <li>
-          is the initial value of the `WeakRef` property of the global object.
+          is the initial value of the *WeakRef* property of the global object.
```
bakkot, line 38754:
```suggestion
          is the initial value of the *"WeakRef"* property of the global object.
```

spec.html
```diff
@@ -38882,40 +38842,34 @@ <h1>WeakRef.prototype.deref ( )</h1>
...
+    <p>A FinalizationRegistry is an object that manages registration and unregistration of cleanup operations that are performed when target objects are garbage collected.</p>
 
     <emu-clause id="sec-finalization-registry-constructor">
       <h1>The FinalizationRegistry Constructor</h1>
       <p>The FinalizationRegistry constructor:</p>
       <ul>
         <li>is the intrinsic object <dfn>%FinalizationRegistry%</dfn>.</li>
         <li>
-          is the initial value of the `FinalizationRegistry` property of the global object.
+          is the initial value of the *FinalizationRegistry* property of the global object.
```
bakkot, line 38872:
```suggestion
          is the initial value of the *"FinalizationRegistry"* property of the global object.
```

spec.html
```diff
@@ -8007,44 +7981,40 @@ <h1>Host Hooks</h1>
...
       </emu-clause>
     </emu-clause>
   </emu-clause>
 
   <emu-clause id="sec-clear-kept-objects" aoid=ClearKeptObjects>
     <h1>ClearKeptObjects ( )</h1>
-    <p>
-      The abstract operation ClearKeptObjects takes no arguments. ECMAScript implementations are expected to call ClearKeptObjects when a synchronous sequence of ECMAScript execution completes.
-    </p>
+    <p>The abstract operation ClearKeptObjects takes no arguments. ECMAScript implementations are expected to call ClearKeptObjects when a synchronous sequence of executions.</p>
```
bakkot, line 7991: > when a synchronous sequence of executions.

I think you may have have edited this line incorrectly.

spec.html
```diff
@@ -8007,44 +7981,40 @@ <h1>Host Hooks</h1>
...
       1. Let _agentRecord_ be the surrounding agent's Agent Record.
-      1. Set _agent_.[[KeptAlive]] to a new empty List.
+      1. Set _agentRecord_.[[KeptAlive]] to a new empty List.
     </emu-alg>
   </emu-clause>
 
   <emu-clause id="sec-addtokeptobjects" aoid="AddToKeptObjects">
     <h1>AddToKeptObjects ( _object_ )</h1>
-    <p> The abstract operation AddToKeptObjects takes argument _object_ (an Object). The following steps are performed: </p>
+    <p>The abstract operation AddToKeptObjects takes argument _object_ (an Object). The following steps are performed:</p>
```
bakkot, line 8001:
```suggestion
    <p>The abstract operation AddToKeptObjects takes argument _object_ (an Object). It performs the following steps when called:</p>
```

spec.html
```diff
@@ -7872,6 +7877,150 @@ <h1>Forward Progress</h1>
...
+
+        <p>The abstract operation HostEnqueueFinalizationRegistryCleanupJob takes argument _finalizationRegistry_ (a FinalizationRegistry). HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that is expected to call CleanupFinalizationRegistry(_finalizationRegistry_) at some point in the future, if possible. The host's responsibility is to make this call at a time which does not interrupt synchronous ECMAScript code execution.</p>
+      </emu-clause>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-clear-kept-objects" aoid=ClearKeptObjects>
+    <h1>ClearKeptObjects ( )</h1>
+    <p>The abstract operation ClearKeptObjects takes no arguments. ECMAScript implementations are expected to call ClearKeptObjects when a synchronous sequence of executions.</p>
+    <p>The following steps are performed:</p>
```
bakkot, line 7992: This should be joined with the preceding paragraph and replaced with "It performs the following steps when called:"

spec.html
codehag: Its not entirely clear to me if I did what you intended. If it is right now, can you mark it as resolved?

spec.html
syg: @bakkot's suggestion sounds good to me.

spec.html
codehag: Good, its fixed.

spec.html
```diff
@@ -7872,6 +7877,149 @@ <h1>Forward Progress</h1>
...
+      1. Let _agentRecord_ be the surrounding agent's Agent Record.
+      1. Append _object_ to _agentRecord_.[[KeptAlive]].
+    </emu-alg>
+    <emu-note>
+      When the abstract operation AddToKeptObjects is called with a target object reference, it adds the target to a list that will point strongly at the target until ClearKeptObjects is called.
+    </emu-note>
+  </emu-clause>
+
+  <emu-clause id="sec-cleanup-finalization-registry" aoid="CleanupFinalizationRegistry">
+    <h1>CleanupFinalizationRegistry ( _finalizationRegistry_ [ , _callback_ ] )</h1>
```
bakkot, line 8011: As far as I can tell, nothing in this specification ever passes the `_callback_` argument to this abstract operation - the only caller is in the prose

> `HostEnqueueFinalizationRegistryCleanupJob` is an implementation-defined abstract operation that is expected to call `CleanupFinalizationRegistry(finalizationRegistry)` at some point in the future

which omits the second argument. Looking at the proposal history, it seems that the second argument is for use by `cleanupSome`, which has been split out of this proposal. As such, I think it makes sense to drop it from this PR and add it back once necessary, since that change is straightforward.

spec.html
```diff
@@ -7872,6 +7877,149 @@ <h1>Forward Progress</h1>
...
+    <p>The abstract operation CleanupFinalizationRegistry takes argument _finalizationRegistry_ (a finalization registry object) and optional argument _callback_ (a function object). The following steps are performed:</p>
```
bakkot, line 8012:
```suggestion
    <p>The abstract operation CleanupFinalizationRegistry takes argument _finalizationRegistry_ (a finalization registry object) and optional argument _callback_ (a function object). It performs the following steps when called:</p>
```

spec.html
```diff
@@ -7872,6 +7877,149 @@ <h1>Forward Progress</h1>
...
+        <li>
+          When `WeakRef.prototype.deref` is called, the referent (if *undefined* is not returned) is kept alive so that subsequent, synchronous accesses also return the object. This list is reset when synchronous work is done using the ClearKeptObjects abstract operation.
+        </li>
+
+        <li>
+          When an object which is registered with a FinalizationRegistry becomes unreachable, a call of the FinalizationRegistry's cleanup callback may eventually be made, after synchronous ECMAScript execution completes.  The FinalizationRegistry cleanup is performed with the CleanupFinalizationRegistry abstract operation.
+        </li>
+      </ul>
+
+      <p>Neither of these actions (ClearKeptObjects or CleanupFinalizationRegistry) may interrupt synchronous ECMAScript execution. Because host environments may assemble longer, synchronous ECMAScript execution runs, this specification defers the scheduling of ClearKeptObjects and CleanupFinalizationRegistry to the host environment.</p>
```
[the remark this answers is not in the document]
syg: More accurately, probably "to the host or implementation", given the GC variance.

spec.html
```diff
@@ -7872,6 +7877,149 @@ <h1>Forward Progress</h1>
...
+      </emu-note>
+    </emu-clause>
+
+    <emu-clause id="sec-weakref-host-hooks">
+      <h1>Host Hooks</h1>
+
+      <emu-clause id="sec-host-cleanup-finalization-registry" aoid=HostEnqueueFinalizationRegistryCleanupJob>
+        <h1>HostEnqueueFinalizationRegistryCleanupJob ( _finalizationRegistry_ )</h1>
+
+        <p>The abstract operation HostEnqueueFinalizationRegistryCleanupJob takes argument _finalizationRegistry_ (a FinalizationRegistry). HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that is expected to call CleanupFinalizationRegistry(_finalizationRegistry_) at some point in the future, if possible. The host's responsibility is to make this call at a time which does not interrupt synchronous ECMAScript code execution.</p>
```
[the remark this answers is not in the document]
syg: +1 Please add this new hook to Annex D.

spec.html
```diff
@@ -7872,6 +7877,149 @@ <h1>Forward Progress</h1>
...
+
+      <p>Neither of these actions (ClearKeptObjects or CleanupFinalizationRegistry) may interrupt synchronous ECMAScript execution. Because host environments may assemble longer, synchronous ECMAScript execution runs, this specification defers the scheduling of ClearKeptObjects and CleanupFinalizationRegistry to the host environment.</p>
+
+      <p>Some ECMAScript implementations include garbage collector implementations which run in the background, including when ECMAScript is idle. Letting the host environment schedule CleanupFinalizationRegistry allows it to resume ECMAScript execution in order to run finalizer work, which may free up held values, reducing overall memory usage.</p>
+    </emu-clause>
+
+    <emu-clause id="sec-liveness">
+      <h1>Liveness</h1>
+
+      <p>For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby `WeakRef.prototype.deref` being called on a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
```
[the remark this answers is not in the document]
syg: The intention is that it's cross-Realm but WeakRefs are not cross-agent. Agents, outside of SABs, have disjoint object closures in practice (though as you know this isn't exactly formally spelled out).

You are correct though that the intention is that no matter which particular Realm's `WeakRef.prototype.deref` is called within an agent, this constraint applies. It's kind of unfortunate if that's the only reason we need to lift out the steps of `WeakRef.prototype.deref` into an AO, but the alternative is to say something like "the algorithm steps of `WeakRef.prototype.deref` being evaluated", which is worse. If we had an AO it'd also link straightforwardly.

All in all I agree refactoring into an AO is a good fix here.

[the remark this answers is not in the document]
bakkot: It's possible we could find a way of phrasing this so that it was talking about any realm's % WeakRef.prototype.deref%, without needing the new AO, but I don't have a phrasing I'm happy with. @codehag was going to try to think of something.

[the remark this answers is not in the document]
ljharb: I suspect there's other places in the spec where an intrinsic is discussed, and is intended to refer to *every* realm's copy of that intrinsic.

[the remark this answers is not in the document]
ljharb: Just audited the spec; you're correct that there are no others.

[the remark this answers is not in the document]
codehag: Proposed wording (with help from @littledan): 

For some set of objects S, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to S is an execution whereby, for every Realm r, r.[[Intrinsics]].[[%WeakRef.prototype.deref%]] being called on a WeakRef whose referent is an element of S always returns undefined.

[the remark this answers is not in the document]
codehag: Also, it may be easier to do this as an AO in the end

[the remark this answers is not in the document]
littledan: AO as @bakkot and @syg suggested seems simplest, as it removes the need to think about Realms at all.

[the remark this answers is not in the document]
codehag: I have drafted a version of the AO. It's the first time I do it so likely needs an iteration.

spec.html
```diff
@@ -7939,7 +7939,8 @@ <h1>Objectives</h1>
     <emu-clause id="sec-liveness">
       <h1>Liveness</h1>
 
-      <p>For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby %WeakRef.prototype.deref% being called on a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
+      <p>For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref being called on a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
```
littledan, line 7942: Nit: I'm not sure if we say "being called on" for abstract operations; maybe this would be better expressed as "of"

spec.html
```diff
@@ -38886,6 +38882,22 @@ <h1>WeakRef.prototype [ @@toStringTag ]</h1>
       </emu-clause>
     </emu-clause>
 
+    <emu-clause id="sec-weakref-abstract-operations">
+      <h1>WeakRef Abstract Operations</h1>
+      <emu-clause id="sec-weakrefderef" aoid="WeakRefDeref">
+        <h1>WeakRefDeref ( _weakRef_ )</h1>
+        <p>The abstract operation WeakRefDeref takes argument _weakRef_ (a WeakRef). It performs the following steps when called:</p>
```
bakkot, line 38889: The types in the headers for AOs are intended to be effectively assertions (i.e., they should _always_ hold), rather than descriptions of the happy-path use. Here the `a WeakRef` part isn't necessarily true, because `WeakRef.prototype.deref` can be `.call`'d on anything.

The simplest fix is to just drop the type here, but you could alternatively say "`(an ECMAScript language value)`" in place of "`(a WeakRef)`", or move the `RequireInternalSlot` into `WeakRef.prototype.deref` so that this AO really would be called exclusively with real `WeakRef`s. I'd lean towards moving the internal slot check, personally.

spec.html
```diff
@@ -38886,6 +38882,22 @@ <h1>WeakRef.prototype [ @@toStringTag ]</h1>
...
+        <emu-alg>
+          1. Perform ? RequireInternalSlot(_weakRef_, [[WeakRefTarget]]).
+          1. Let _target_ be _weakRef_.[[WeakRefTarget]].
+          1. If _target_ is not ~empty~, then
+            1. Perform ! AddToKeptObjects(_target_).
+            1. Return _target_.
+          1. Return *undefined*.
+        </emu-alg>
+      </emu-clause>
```
bakkot, line 38898: It's probably worth adding a `NOTE` to this clause explaining that this is a distinct operation strictly for making it easier to define liveness.

spec.html
```diff
@@ -7906,6 +7911,150 @@ <h1>Forward Progress</h1>
...
+        <h1>HostEnqueueFinalizationRegistryCleanupJob ( _finalizationRegistry_ )</h1>
+
+        <p>The abstract operation HostEnqueueFinalizationRegistryCleanupJob takes argument _finalizationRegistry_ (a FinalizationRegistry). HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that is expected to call CleanupFinalizationRegistry(_finalizationRegistry_) at some point in the future, if possible. The host's responsibility is to make this call at a time which does not interrupt synchronous ECMAScript code execution.</p>
+      </emu-clause>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-clear-kept-objects" aoid=ClearKeptObjects>
+    <h1>ClearKeptObjects ( )</h1>
+    <p>The abstract operation ClearKeptObjects takes no arguments. ECMAScript implementations are expected to call ClearKeptObjects when a synchronous sequence of ECMAScript executions completes. It performs the following steps when called:</p>
```
gibson042, line 8026: Given the plausible future addition of Symbols, it would be nice to give these operations initial names that do not directly reference the Object type. "KeptObject" could be replaced with something like "KeptAlive" or "KeptList".

codehag: A similar comment was brought up above, and it was agreed that we can change this after the Symbols in WeakMap proposal moved forward. I would rather do this is a group change, as there will be other edits,  rather than anticipating an unrelated proposal in this one.

spec.html
```diff
@@ -7906,6 +7911,150 @@ <h1>Forward Progress</h1>
...
+        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep an object alive. Two, that cycles in liveness does not imply that an object is live. To be concrete, if determining _obj_'s liveness depends on determining the liveness of another WeakRef referent, _obj2_, _obj2_'s liveness cannot assume _obj_'s liveness, which would be circular reasoning.
+      </emu-note>
+      <emu-note>
+        WeakRef-obliviousness is defined on sets of objects instead of individual objects to account for cycles. If it were defined on individual objects, then an object in a cycle will be considered live even though its Object value is only observed via WeakRefs of other objects in the cycle.
+      </emu-note>
+      <emu-note>
+        Colloquially, we say that an individual object is live if every set of objects containing it is live.
+      </emu-note>
+
+      <p>At any point during evaluation, a set of objects _S_ is considered <dfn>live</dfn> if either of the following conditions is met: </p>
```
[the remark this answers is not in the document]
littledan: I think it's reasonable to define "live" this way within the JavaScript specification. I don't think "liveness" is really any more specific. This proposal successfully links with the term "live" where relevant in normative text (liveness is just used in notes) and the rest of ecma262 does not seem to use "live" in a conflicting way.

spec.html
gibson042: Renaming operations requires introducing `oldids`, which is a bit of an ugly hack that shouldn't be necessary if the initial names do not include "Object". I'm not suggesting that you make the definitions generic right now, just that the names theirselves are not overly specific.

spec.html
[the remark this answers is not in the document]
gibson042: "live" is also the spelling of a common English word (ranked somewhere between 100 and 300 AFAICT) that could easily find a place in the spec, as opposed to "liveness" which will only ever have this meaning.

spec.html
littledan: As the champion of the Symbols as WeakMap keys proposal, I would prefer to not generalize this prematurely. oldids are fine.

spec.html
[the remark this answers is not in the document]
littledan: I think it would be fine to iterate on this in an editorial PR after landing this one.

devsnek:
I think these need to be added to the intrinsics table and to the "constructor properties of the global object" section.

codehag:
I think I still need one more editor to approve this, so pinging @bakkot  and @michaelficarra for one last look before I rebase

spec.html
```diff
@@ -7906,6 +7911,150 @@ <h1>Forward Progress</h1>
       <p>This, along with the liveness guarantee in the memory model, ensures that all ~SeqCst~ writes eventually become observable to all agents.</p>
     </emu-note>
   </emu-clause>
+
+  <emu-clause id="sec-weakref-processing-model">
+    <h1>Processing model of WeakRef and FinalizationRegistry objects</h1>
+    <emu-clause id="sec-weakref-invariants">
+      <h1>Objectives</h1>
+
+      <p> This specification does not make any guarantees that any object will be garbage collected. Objects which are not live may be released after long periods of time, or never at all. For this reason, this specification uses the term "may" when describing behaviour triggered by garbage collection.</p>
```
bakkot, line 7920:
```suggestion
      <p>This specification does not make any guarantees that any object will be garbage collected. Objects which are not live may be released after long periods of time, or never at all. For this reason, this specification uses the term "may" when describing behaviour triggered by garbage collection.</p>
```

bakkot:
@codehag We've been letting @ljharb handle the merging process because he has idiosyncratic preferences for how that should happen.

PR #2092, opened 2020-07-13
Editorial: Module Record in InnerModuleEvaluation need not be a Source Text Module Record

codehag:
In InnerModuleEvaluation, the argument _module_ is defined as `(a Source Text Module Record)`, but this could be a non-source text module record as well it looks like...


codehag:
Do we need more reviews here @ljharb ?

ljharb:
@codehag yes; 3 out of the 4 editors need to approve before it can land.

Issue #467, opened 2016-03-09
Reference type and implementation reality


littledan:
Does anyone have an idea of a resolution on this bug in the old repository?

https://bugs.ecmascript.org/show_bug.cgi?id=4379

Quoting @anba 

> Tests added to test262 [1] have revealed long-standing deviations [2] between the specification for the Reference type and actual implementation reality. Whether the specification should be changed to reflect implementation reality [3] or alternatively implementers should try to comply with specified behaviour should be discussed. 
>
> [1] Added/Proposed PRs for test262 coverage:
>  https://github.com/tc39/test262/pull/91
>  https://github.com/tc39/test262/pull/273
>  https://github.com/tc39/test262/pull/275
>
> [2] Mozilla source code from 1998 already shows the spec violation which is still present in modern engines. Details can be found in https://github.com/tc39/test262/pull/273.
>
> [3] The exact "implementation reality" still needs to be determined.

@ajklein

bterlson:
There is no resolution, and further the deviation continues to exist. I think if we want to change this it should be a full proposal because it is a core part of the semantics and changing this has been very controversial in the past. The difficult part in my mind is that the current spec seems correct and the biggest argment in favor of fixing this problem is that it's hard for implementations to do so. Maybe that's enough?

bakkot:
For reference, here's a couple of simple tests, with results on current (public) builds of various engines. The current spec dictates that both should have `hits` = 1 immediately after execution.

`var hits = 0, obj = {p: 0}, prop = {toString: function(){++hits; return 'p';}}; obj[prop]++;`
- SpiderMonkey: 1
- v8: 2
- WebkitCore: 2
- ChakraCore: 2
- Nashorn: 2

`var hits = 0, obj = {p: 0}, prop = {toString: function(){++hits; return 'p';}}; obj[prop]+=1;`
- SpiderMonkey: 2
- v8: 2
- WebkitCore: 2
- ChakraCore: 2
- Nashorn: 2

Prefix and postfix increment and decrement all behave identically on a given platform, as do all of the compound assignment operators. Per the above, the only variation is that SpiderMonkey only performs `ToPropertyKey` once for increment/decrement. In all other cases, across this set of engines, `ToPropertyKey` is performed exactly twice, in defiance of spec.

littledan:
@bakkot Would you be interested in writing up a proposal for a spec change here? Or do you prefer the semantics of the spec?

bakkot:
[...]

PR #2094, opened 2020-07-14
Normative: Verify if Binding is preserved for obj's SetMutableBinding

### LINKED ISSUE #2093: Assignment of global variables deleted by their RHS in strict mode is permitted
(Copying from [bugzilla](https://tc39.es/archives/bugzilla/4532/); see [discussion in esdiscuss](https://esdiscuss.org/topic/assignment-of-global-variables-deleted-by-their-rhs-in-strict-mode).)

In short, it is possible to have a Reference to a global variable or object-environment-record property which has been deleted. Normally, bare assignments to undeclared variables in strict mode cause ReferenceErrors. However, calling PutValue on a reference to a global variable which has been deleted since the reference was created does not throw a ReferenceError in strict mode, even though, at the time of writing, that variable does not exist.

This circumvents a protection which strict mode is intended to provide.

For example, in

```js
var global = this;
Object.defineProperty(this, "x", {
  configurable: true,
  value: 1
});

(function() {
  "use strict";
  x = (delete global.x, 2);
})();
```
the `x =` goes through [PutValue](https://tc39.es/ecma262/#sec-putvalue), which in step 5 performs [`IsUnresolvableReference`](https://tc39.es/ecma262/#sec-isunresolvablereference), which returns `false` because the Reference _was_ resolvable when it was created. Hence no error is thrown.

However, Firefox, Chrome, and Safari all throw for the above code, in violation of the spec.

I would regard this as a (normative) bug in the spec.

leobalter:
Verify binding existence in the object's SetMutableBinding on Strict Mode.

Closes #2093
Ref #467
Ref tc39/test262#427

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6626,6 +6626,8 @@
           <emu-alg>
             1. Let _envRec_ be the object Environment Record for which the method was invoked.
             1. Let _bindings_ be the binding object for _envRec_.
+            1. Let _stillExists_ be ? HasProperty(_bindings_, _N_).
+            1. If _stillExists_ is *false* and _S_ is *true*, throw a *ReferenceError* exception.
             1. Return ? Set(_bindings_, _N_, _V_, _S_).
           </emu-alg>
         </emu-clause>


bakkot:
Note that the `HasBinding` call is observable. For example, consider

```js
let scope = { x: 1 };
let handler = {
  has(o, p) {
    if (p === 'x') {
      console.log('hit', Reflect.has(o, p));
    }
    return Reflect.has(o, p);
  },
};
with (new Proxy(scope, handler)) {
  (function () {
    'use strict';
    x = (delete scope.x, 2);
  })();
}
```

What should this print / do? In Chrome it prints `hit, false` and throws; in Firefox it prints `hit, true` and does not throw; in Safari it prints `hit, true; hit, false` and throws.

For a trickier example, consider

```js
let scope = { x: 1 };
let first = true;
let handler = {
  has(o, p) {
    if (p === 'scope') {
      return false;
    }
    console.log('hit');
    if (first) {
      first = false;
      return true;
    }
    return false;
  },
};
with (new Proxy({ x: 1 }, handler)) {
  (function () {
    'use strict';
    x = 2;
  })();
}
```
where here the call to `HasBinding` is the thing which causes the property to not exist!

anba:
Can't we perform this check as part of [8.1.1.2.5 SetMutableBinding](https://tc39.es/ecma262/#sec-object-environment-records-setmutablebinding-n-v-s) as proposed in the es-discuss thread? That way it nicely lines up with other strict mode checks in environment records, like for example [8.1.1.2.6 GetBindingValue, step 4](https://tc39.es/ecma262/#sec-object-environment-records-getbindingvalue-n-s) or [8.1.1.1.5 SetMutableBinding, step 2](https://tc39.es/ecma262/#sec-declarative-environment-records-setmutablebinding-n-v-s).

leobalter:
@anba @bakkot how does it look now?

We are still caught by Proxies `[[HasProperty]]` that can falsely report as an existing property. IMO, it sounds like not a problem for the specs but user land.

anba:
LGTM! 

FWIW the proxy issue actually also applies to [8.1.1.2.6 GetBindingValue](https://tc39.es/ecma262/#sec-object-environment-records-getbindingvalue-n-s).

leobalter:
This has reached consensus today. We need to update the current tests - starting with those pointed out at tc39/test262#427 - and also verify observable implications with Proxies.

spec.html
```diff
@@ -6551,6 +6551,9 @@ <h1>SetMutableBinding ( _N_, _V_, _S_ )</h1>
           <emu-alg>
             1. Let _envRec_ be the object Environment Record for which the method was invoked.
             1. Let _bindings_ be the binding object for _envRec_.
+            1. Let _value_ be ? HasProperty(_bindings_, _N_).
+            1. If _value_ is *false*, then
+              1. If _S_ is *true*, throw a *ReferenceError* exception.
```
michaelficarra, line 6556: I would prefer a better name than `_value_` and I think we can combine the nested ifs.

```suggestion
            1. Let _stillExists_ be ? HasProperty(_bindings_, _N_).
            1. If _stillExists_ is *false* and _S_ is *true*, throw a *ReferenceError* exception.
```

On the topic of names, the use of `S` instead of `strict` (and likewise for `N` and `V`) is unfortunate.

anba: The current PR matches the structure used in <https://tc39.es/ecma262/#sec-object-environment-records-getbindingvalue-n-s>, so any changes in that regard should probably happen throughout the whole clause 8.1.1. For example changing `value` to `foundBinding` to match 8.1.1.2.1.

michaelficarra: I don't really care about consistency in aliases, especially when we have the opportunity to make an incremental improvement. And the structure in the section you linked can't be simplified in the way I suggested here.

jmdyck: > I don't really care about consistency in aliases

To be clear, when you say "aliases", are you excluding parameter names?
- I agree that `_value_` is a bad name here, and `_stillExists_` or `_foundBinding_` would be better.
- But `S`, `N`, and `V` are parameter names that appear multiple times across 8.1.1, and should probably be changed all-or-nothing, rather than incrementally.

michaelficarra: I would love the parameter names changed also, but since they're not being touched by this PR, I'm not going to ask for them to be changed. But I am going to ask that lines that *are* touched use better names.

leobalter: well, picking better names is a subjective game. I'm accepting the changes here but I'd still try to follow perceived convention and/or consistency from the existing parts.

Is there anything else beyond this commit suggestion I could do to unblock this PR?

leobalter:
I rebased the branch but kept the original commits for the history of changes.

PR #2106, opened 2020-07-21
Normative: Add `import.meta[Symbol.toStringTag]`

ExE-Boss:
See <https://github.com/tc39/proposal-import-meta/issues/18> and <https://github.com/tc39/ecma262/issues/1970#issuecomment-622248793>.

---

[Preview](https://ci.tc39.es/preview/tc39/ecma262/pull/2106) ([#sec-meta-properties-runtime-semantics-evaluation](https://ci.tc39.es/preview/tc39/ecma262/pull/2106#sec-meta-properties-runtime-semantics-evaluation))

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -14158,6 +14158,7 @@
           1. Let _importMeta_ be _module_.[[ImportMeta]].
           1. If _importMeta_ is ~empty~, then
             1. Set _importMeta_ to ! OrdinaryObjectCreate(*null*).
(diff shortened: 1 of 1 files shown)


ljharb:
I failed to mention this in this week's meeting, so it will probably have to wait until the next meeting to be discussed.

bmeck:
I'm not convinced we should do this.

* Hosts can supply toStringTag already
* Hosts can replace the prototype of import.meta objects (none do ?)
* Unlike other objects created by the spec this is generated by host data and not something we have looked at adding properties to at TC39. I'd even go so far as to state TC39 should never populate these objects as even things designed to have some host interoperability like `import.meta.url` are not consistent across platforms for various known reasons.
* There are *many* of these objects unlike the other things we add `@@toStringTag`, it seems odd to add many `own` properties where other namespace objects only add it to a central location (because they are a centralized location).

PR #2045, opened 2020-06-14
Editorial: Add Yield, CreateIteratorFromClosure, CreateAsyncIteratorFromClosure abs op.

Jack-Works:
Introduce `Yield` and `CreateIteratorFromClosure`, `CreateAsyncIteratorFromClosure` in the spec to make it easier to define builtin generators.

Diff: https://arai-a.github.io/ecma262-compare/?pr=2045

Issue #1894, opened 2020-03-10
use abstract closures for CreateBuiltInFunction


bakkot:
https://github.com/tc39/ecma262/pull/1889 introduced a notion of an abstract closure, but we still pass algorithm steps around directly as an argument to CreateBuiltInFunction. It would be nice to replace those with abstract closures too. Plus, the inline style for abstract closures is probably more readable - this would resolve https://github.com/tc39/ecma262/issues/933.

michaelficarra:
@bakkot Is this fixed by #2109?

bakkot:
No, there's still a bunch of places we do this. I'm preparing a PR.

PR #2109, opened 2020-07-21
Editorial: Add support for Abstract Closures to `CreateBuiltinFunction`

ExE-Boss:
Part of <https://github.com/tc39/ecma262/issues/1894>.

I’ve also converted [`Object.fromEntries`](https://tc39.es/ecma262/#sec-object.fromentries) to use it, as it’s the only built‑in anonymous function that has no internal slots apart from [`CreateListIteratorRecord`](https://tc39.es/ecma262/#sec-createlistiteratorRecord), but built‑in iterators are being converted to use [Abstract Closures](https://tc39.es/ecma262/#sec-abstract-closure) in <https://github.com/tc39/ecma262/pull/2045>.

## Depends on:
- [x] <https://github.com/tc39/ecma262/pull/2116> (merged)

---

[Preview](https://ci.tc39.es/preview/tc39/ecma262/pull/2109) ([#sec-createbuiltinfunction](https://ci.tc39.es/preview/tc39/ecma262/pull/2109#sec-createbuiltinfunction), [#sec-object.fromentries](https://ci.tc39.es/preview/tc39/ecma262/pull/2109#sec-object.fromentries))

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -11632,14 +11632,14 @@
     </emu-clause>
 
     <emu-clause id="sec-createbuiltinfunction" aoid="CreateBuiltinFunction">
-      <h1>CreateBuiltinFunction ( _steps_, _length_, _name_, _internalSlotsList_ [ , _realm_ [ , _prototype_ [ , _prefix_ ] ] ] )</h1>
-      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_, _length_, _name_, and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_, _prototype_, and _prefix_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
+      <h1>CreateBuiltinFunction ( _behaviour_, _length_, _name_, _internalSlotsList_ [ , _realm_ [ , _prototype_ [ , _prefix_ ] ] ] )</h1>
+      <p>The abstract operation CreateBuiltinFunction takes arguments _behaviour_, _length_ (a non-negative integer or +&infin;), _name_ (a property key), and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_ (a Realm Record), _prototype_ (an Object or *null*), and _prefix_ (a String). _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
       <emu-alg>
-        1. Assert: _steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification.
-        1. If _realm_ is not present or _realm_ is ~empty~, set _realm_ to the current Realm Record.
+        1. Assert: _behaviour_ is either an Abstract Closure, a set of algorithm steps, or some other definition of a function's behaviour provided in this specification.
+        1. If _realm_ is not present, set _realm_ to the current Realm Record.
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
+        1. Let _func_ be a new built-in function object that, when called, performs the action described by _behaviour_ using the provided arguments as the values of the corresponding parameters specified by _behaviour_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
         1. Set _func_.[[Realm]] to _realm_.
         1. Set _func_.[[Prototype]] to _prototype_.
         1. Set _func_.[[Extensible]] to *true*.
@@ -21515,7 +21515,7 @@
         1. Set the running execution context's PrivateEnvironment to _classPrivateEnvironment_.
         1. If _constructor_ is ~empty~, then
           1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-default-constructor-functions" title></emu-xref>.
-          1. Let _F_ be ! CreateBuiltinFunction(_steps_, 0, _className_, &laquo; [[ConstructorKind]], [[SourceText]] &raquo;, ~empty~, _constructorParent_).
+          1. Let _F_ be ! CreateBuiltinFunction(_steps_, 0, _className_, &laquo; [[ConstructorKind]], [[SourceText]] &raquo;, the current Realm Record, _constructorParent_).
         1. Else,
           1. Let _constructorInfo_ be ! DefineMethod of _constructor_ with arguments _proto_ and _constructorParent_.
           1. Let _F_ be _constructorInfo_.[[Closure]].
@@ -25507,34 +25507,23 @@
         </emu-alg>
       </emu-clause>
 
-      <emu-clause id="sec-object.fromentries">
+      <emu-clause id="sec-object.fromentries" oldids="sec-create-data-property-on-object-functions">
         <h1>Object.fromEntries ( _iterable_ )</h1>
         <p>When the `fromEntries` method is called with argument _iterable_, the following steps are taken:</p>
         <emu-alg>
           1. Perform ? RequireObjectCoercible(_iterable_).
           1. Let _obj_ be ! OrdinaryObjectCreate(%Object.prototype%).
           1. Assert: _obj_ is an extensible ordinary object with no own properties.
-          1. Let _stepsDefine_ be the algorithm steps defined in <emu-xref href="#sec-create-data-property-on-object-functions" title></emu-xref>.
-          1. Let _lengthDefine_ be the number of non-optional parameters of the function definition in <emu-xref href="#sec-create-data-property-on-object-functions" title></emu-xref>.
-          1. Let _adder_ be ! CreateBuiltinFunction(_stepsDefine_, _lengthDefine_, *""*, &laquo; &raquo;).
+          1. Let _closure_ be a new Abstract Closure with parameters (_key_, _value_) that captures _obj_ and performs the following steps when called:
+            1. Let _propertyKey_ be ? ToPropertyKey(_key_).
+            1. Perform ! CreateDataPropertyOrThrow(_obj_, _propertyKey_, _value_).
+            1. Return *undefined*.
+          1. Let _adder_ be ! CreateBuiltinFunction(_closure_, 2, *""*, &laquo; &raquo;).
           1. Return ? AddEntriesFromIterable(_obj_, _iterable_, _adder_).
         </emu-alg>
         <emu-note>
           The function created for _adder_ is never directly accessible to ECMAScript code.
         </emu-note>
-
-        <emu-clause id="sec-create-data-property-on-object-functions">
-          <h1>CreateDataPropertyOnObject Functions</h1>
-          <p>A CreateDataPropertyOnObject function is an anonymous built-in function. When a CreateDataPropertyOnObject function is called with arguments _key_ and _value_, the following steps are taken:</p>
-          <emu-alg>
-            1. Let _O_ be the *this* value.
-            1. Assert: Type(_O_) is Object.
-            1. Assert: _O_ is an extensible ordinary object.
-            1. Let _propertyKey_ be ? ToPropertyKey(_key_).
-            1. Perform ! CreateDataPropertyOrThrow(_O_, _propertyKey_, _value_).
-            1. Return *undefined*.
-          </emu-alg>
-        </emu-clause>
       </emu-clause>
 
       <emu-clause id="sec-object.getownpropertydescriptor">


spec.html
```diff
@@ -8871,18 +8871,22 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_.
+        1. Let _func_ be a new built-in function object that when called performs the action described by _stepsOrClosure_. The new function object has internal slots whose names are the elements of _internalSlotsList_.
         1. Set _func_.[[Realm]] to _realm_.
         1. Set _func_.[[Prototype]] to _prototype_.
         1. Set _func_.[[Extensible]] to *true*.
         1. Set _func_.[[ScriptOrModule]] to *null*.
+        1. If _stepsOrClosure_ is an Abstract Closure, then
+          1. Let _len_ be the largest number of named parameters in the definition of the Abstract Closure _stepsOrClosure_. Optional parameters (which are indicated with brackets: `[` `]`) or rest parameters (which are shown using the form &laquo;...name&raquo;) are not included in the parameter count.
+          1. Perform ! SetFunctionLength(_func_, _len_).
+          1. Perform ! SetFunctionName(_func_, *""*).
```
bakkot, line 8889: I appreciate that this makes the current in-prose requirement

> Functions that are identified as anonymous functions use the empty String as the value of the "name" property.

into an explicit algorithm step.

However, this does have the effect of defining a creation order for these two properties, where currently I believe it is undefined. That means that using this mechanism for any user-observable function is technically normative.

(On the plus side, most engines seem to agree with this ordering:
```js
(async function () {
  return await {
    then: v => v(v),
  };
})().then(v => {
  console.log(Reflect.ownKeys(v));
});
```
returns `["length", "name"]` on at least JSC, SM, and V8.)

spec.html
```diff
@@ -8871,18 +8871,22 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_.
+        1. Let _func_ be a new built-in function object that when called performs the action described by _stepsOrClosure_. The new function object has internal slots whose names are the elements of _internalSlotsList_.
         1. Set _func_.[[Realm]] to _realm_.
         1. Set _func_.[[Prototype]] to _prototype_.
         1. Set _func_.[[Extensible]] to *true*.
         1. Set _func_.[[ScriptOrModule]] to *null*.
+        1. If _stepsOrClosure_ is an Abstract Closure, then
+          1. Let _len_ be the largest number of named parameters in the definition of the Abstract Closure _stepsOrClosure_. Optional parameters (which are indicated with brackets: `[` `]`) or rest parameters (which are shown using the form &laquo;...name&raquo;) are not included in the parameter count.
```
bakkot, line 8887: Abstract closures do not have optional parameters, so this could just be
```suggestion
          1. Let _len_ be number of parameters taken by _stepsOrClosure_.
```

spec.html
```diff
@@ -25970,33 +25974,26 @@ <h1>Object.freeze ( _O_ )</h1>
...
-        </emu-alg>
-        <emu-note>
-          The function created for _adder_ is never directly accessible to ECMAScript code.
-        </emu-note>
-
-        <emu-clause id="sec-create-data-property-on-object-functions">
-          <h1>CreateDataPropertyOnObject Functions</h1>
-          <p>A CreateDataPropertyOnObject function is an anonymous built-in function. When a CreateDataPropertyOnObject function is called with arguments _key_ and _value_, the following steps are taken:</p>
-          <emu-alg>
+          1. Let _closure_ be a new Abstract Closure with no parameters that performs the following steps when called:
```
bakkot, line 25984:
```suggestion
          1. Let _closure_ be a new Abstract Closure with parameters (_key_, _value_) that performs the following steps when called:
```

spec.html
```diff
@@ -8871,18 +8871,22 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
+      <h1>CreateBuiltinFunction ( _stepsOrClosure_, _internalSlotsList_ [ , _realm_ [ , _prototype_ ] ] )</h1>
+      <p>The abstract operation CreateBuiltinFunction takes arguments _stepsOrClosure_ and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_ and _prototype_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
       <emu-alg>
-        1. Assert: _steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification.
+        1. Assert: _stepsOrClosure_ is either an Abstract Closure, a set of algorithm steps or other definition of a function's behaviour provided in this specification.
```
bakkot, line 8877:
```suggestion
        1. Assert: _stepsOrClosure_ is either an Abstract Closure, a set of algorithm steps, or some other definition of a function's behaviour provided in this specification.
```

spec.html
ExE-Boss: This order was made normative in <https://github.com/tc39/ecma262/pull/1490>.

bakkot: #1490 does not imply an order for these properties for built-in functions, as far as I can tell. Am I misreading it?

ljharb: I very much doubt if the *intention* of #1490 was to leave the ordering of those properties different between builtin and user functions, but it certainly may not apply to builtin functions as written.

jmdyck: For non-builtin functions, the order has been "length" then "name" since ES6. ("length" is defined as part of the operation that creates the function, and "name" is defined shortly thereafter, if it all.) (PR #1490 was originally going to define "name" as part of the function-creating operation, but pulled back from that.)

For built-in functions, the creation of those properties isn't specified via algorithm, but FWIW, [ECMAScript Standard Built-in Objects](https://tc39.es/ecma262/#sec-ecmascript-standard-built-in-objects) has discussed them in the order "length" then "name" since ES6.

spec.html
```diff
@@ -8871,18 +8871,22 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_.
+        1. Let _func_ be a new built-in function object that when called performs the action described by _stepsOrClosure_. The new function object has internal slots whose names are the elements of _internalSlotsList_.
         1. Set _func_.[[Realm]] to _realm_.
         1. Set _func_.[[Prototype]] to _prototype_.
         1. Set _func_.[[Extensible]] to *true*.
         1. Set _func_.[[ScriptOrModule]] to *null*.
+        1. If _stepsOrClosure_ is an Abstract Closure, then
+          1. Let _len_ be number of parameters taken by _stepsOrClosure_.
+          1. Perform ! SetFunctionLength(_func_, _len_).
+          1. Perform ! SetFunctionName(_func_, *""*).
```
jmdyck, line 8889: This is odd, because it implies that, of all the built-in functions, only the ones created via a closure get `length` and `name` properties.

I'd be inclined to leave this out until/unless we have a mechanism that works for all built-in functions.

spec.html
```diff
@@ -25970,33 +25974,26 @@ <h1>Object.freeze ( _O_ )</h1>
...
-        </emu-alg>
-        <emu-note>
-          The function created for _adder_ is never directly accessible to ECMAScript code.
-        </emu-note>
-
-        <emu-clause id="sec-create-data-property-on-object-functions">
-          <h1>CreateDataPropertyOnObject Functions</h1>
-          <p>A CreateDataPropertyOnObject function is an anonymous built-in function. When a CreateDataPropertyOnObject function is called with arguments _key_ and _value_, the following steps are taken:</p>
-          <emu-alg>
+          1. Let _closure_ be a new Abstract Closure with parameters (_key_, _value_) that performs the following steps when called:
```
jmdyck, line 25984:
```suggestion
          1. Let _closure_ be a new Abstract Closure with parameters (_key_, _value_) that captures nothing and performs the following steps when called:
```

spec.html
```diff
@@ -25970,33 +25974,26 @@ <h1>Object.freeze ( _O_ )</h1>
...
             1. Let _O_ be the *this* value.
```
jmdyck, line 25985: I'm not sure it makes sense for an abstract closure to access `the *this* value`. Anyway, I think it would be clearer for the closure to capture `_obj_` and use that instead of `_O_`.

spec.html
ExE-Boss: Well, all the current built‑in functions are created by passing <var>steps</var> to `CreateBuiltinFunction` define `name` and `length` in the description of the steps.

jmdyck: > Well, all the current built‑in functions are created by passing steps to `CreateBuiltinFunction`

Pretty much, yes.

> define `name` and `length` in the description of the steps.

"_steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification."
I don't think either possibility includes the values of the function's `name` and `length` properties.

bakkot: I am inclined to make `CreateBuiltInFunction` explicitly take `_name_` and `_length_` parameters. The function only has 20 callsites in the current spec.

ExE-Boss: I’ll do that in its own PR and then rebase this on top of it.

spec.html
```diff
@@ -8871,18 +8871,22 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
       <emu-alg>
-        1. Assert: _steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification.
+        1. Assert: _algorithm_ is either an Abstract Closure, a set of algorithm steps, or some other definition of a function's behaviour provided in this specification.
         1. If _realm_ is not present, set _realm_ to the current Realm Record.
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_.
+        1. Let _func_ be a new built-in function object that when called performs the action described by _algorithm_. The new function object has internal slots whose names are the elements of _internalSlotsList_.
```
syg, line 8881: I feel like this lacking clarity a bit for the case with `_algorithm_` is an Abstract Closure. Since the definition of ACs explicitly talk about invoking them, perhaps we should have a sentence here that in the case _algorithm_ is an AC, it is invoked with the same arguments in the same order that the built-in function objects was called with, and its return value returned.

syg (approved):
lgtm with one comment

spec.html
jmdyck: Theoretically, _algorithm_ (whether or not it's an AC) is never invoked, it merely specifies the behavior that the new function must exhibit.

spec.html
ExE-Boss: This has now been normatively specified in <https://github.com/tc39/ecma262/pull/2116>.

spec.html
```diff
@@ -11151,14 +11151,14 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
-      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_, _length_, _name_, and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_, _prototype_, and _prefix_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
+      <h1>CreateBuiltinFunction ( _algorithm_, _length_, _name_, _internalSlotsList_ [ , _realm_ [ , _prototype_ [ , _prefix_ ] ] ] )</h1>
+      <p>The abstract operation CreateBuiltinFunction takes arguments _algorithm_, _length_ (a non-negative integer or +&infin;), _name_ (a property key), and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_ (a Realm Record), _prototype_ (an Object or *null*), and _prefix_ (a String). _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
```
ljharb, line 11155:
```suggestion
      <p>The abstract operation CreateBuiltinFunction takes arguments _algorithm_, _length_ (a non-negative integer or +&infin;), _name_ (a property key), _internalSlotsList_ (a List of names of internal slots), and optional arguments _realm_ (a Realm Record), _prototype_ (an Object or *null*), and _prefix_ (a String). _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
```

jmdyck: Actually, the precendent is to have that "and" there. See the preambles of:
- Call ( _F_, _V_ [ , _argumentsList_ ] )
- Invoke ( _V_, _P_ [ , _argumentsList_ ] )
- OrdinaryCreateFromConstructor ( _constructor_, _intrinsicDefaultProto_ [ , _internalSlotsList_ ] )
- SetFunctionName ( _F_, _name_ [ , _prefix_ ] )
- CreateBuiltinFunction ( _algorithm_, _length_, _name_, _internalSlotsList_ [ , _realm_ [ , _prototype_ [ , _prefix_ ] ] ] )
- ForIn/OfBodyEvaluation ( _lhs_, _stmt_, _iteratorRecord_, _iterationKind_, _lhsKind_, _labelSet_ [ , _iteratorKind_ ] )
- FlattenIntoArray ( _target_, _source_, _sourceLen_, _start_, _depth_ [ , _mapperFunction_, _thisArg_ ] )
- AllocateTypedArray ( _constructorName_, _newTarget_, _defaultProto_ [ , _length_ ] )
- GetValueFromBuffer ( _arrayBuffer_, _byteIndex_, _type_, _isTypedArray_, _order_ [ , _isLittleEndian_ ] )
- SetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _isTypedArray_, _order_ [ , _isLittleEndian_ ] )
- GetModifySetValueInBuffer ( _arrayBuffer_, _byteIndex_, _type_, _value_, _op_ [ , _isLittleEndian_ ] )
- DataView.prototype.setFoo ( _byteOffset_, _value_ [ , _littleEndian_ ] )
- PerformPromiseThen ( _promise_, _onFulfilled_, _onRejected_ [ , _resultCapability_ ] )

(or scan spec.html for " and .* and optional arg")

ljharb: Gross, ok. Will look into removing it from everywhere in a separate PR.

bakkot: I think the current form is grammatically correct. Compare:

> The state is served by representatives Alice, Bob, and Carol and senators Dan and Erin.

Here dropping the initial "and" would be clearly incorrect, because "representatives" does not describe Dan or Erin. And there's no comma after "Carol" because the following "and" is joining a list of two: "representatives and senators", not "representatives, and senators".

Anyway we should talk about it before you go to the trouble of making a new PR.

ljharb: In this case it'd be like "politicians" and "senators", because the second is a subset of the first, which is where i think the difference comes in.

spec.html
syg: Fair enough, it is not invoked in the sense of an actual JS function invocation. There's no term we use for the "meta invocation" that's spec-internal calling of an AC. The high order bit for me here was to say something about forwarding of arguments. Perhaps nobody else finds this under-specified and considers "performs the action described" to include the argument forwarding?

jmdyck: Actually, when I said "invoked", I did mean it in your sense of "meta invocation": _algorithm_ is never meta-invoked. When the new function object is called, it doesn't forward arguments to _algorithm_, because _algorithm_ isn't in the picture.

E.g., you can imagine that the implementation uses _algorithm_ as a template/blueprint for building the function, and then discards it once the function is created (i.e., by the time this step finishes). _algorithm_'s parameters are a template for the function's parameters, so when the function is called, it will handle its arguments in a way that is consistent with how _algorithm_ handles its parameters.

That's how I read it, anyway.

syg: That's a fine mental model, I'd still like a sentence about it, like "the function object handles parameters in a way consistent with how _algorithm_ handles parameters"

spec.html
```diff
@@ -11159,14 +11159,14 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
     </emu-clause>
 
     <emu-clause id="sec-createbuiltinfunction" aoid="CreateBuiltinFunction">
-      <h1>CreateBuiltinFunction ( _steps_, _length_, _name_, _internalSlotsList_ [ , _realm_ [ , _prototype_ [ , _prefix_ ] ] ] )</h1>
-      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_, _length_, _name_, and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_, _prototype_, and _prefix_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
+      <h1>CreateBuiltinFunction ( _algorithm_, _length_, _name_, _internalSlotsList_ [ , _realm_ [ , _prototype_ [ , _prefix_ ] ] ] )</h1>
```
michaelficarra, line 11162:
```suggestion
      <h1>CreateBuiltinFunction ( _behaviour_, _length_, _name_, _internalSlotsList_ [ , _realm_ [ , _prototype_ [ , _prefix_ ] ] ] )</h1>
```

I would call this parameter `behaviour`.

spec.html
bakkot: Perhaps "performs the action described by _algorithm_ using the provided arguments as the values of the corresponding parameters specified by _algorithm_"?

That doesn't describe how to deal with the case that the function is called with the wrong number of arguments, but that's covered by these paragraphs in ECMAScript Standard Built-in Objects:

> Unless otherwise specified in the description of a particular function, if a built-in function or constructor is given fewer arguments than the function is specified to require, the function or constructor shall behave exactly as if it had been given sufficient additional arguments, each such argument being the undefined value. Such missing arguments are considered to be “not present” and may be identified in that manner by specification algorithms. In the description of a particular function, the terms “this value” and “NewTarget” have the meanings given in 10.3.
> 
> Unless otherwise specified in the description of a particular function, if a built-in function or constructor described is given more arguments than the function is specified to allow, the extra arguments are evaluated by the call and then ignored by the function. However, an implementation may define implementation specific behaviour relating to such arguments as long as the behaviour is not the throwing of a TypeError exception that is predicated simply on the presence of an extra argument.

syg: This wording sgtm.

michaelficarra (approved):
LGTM otherwise.

spec.html
```diff
@@ -11632,14 +11632,14 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
       <emu-alg>
-        1. Assert: _steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification.
-        1. If _realm_ is not present or _realm_ is ~empty~, set _realm_ to the current Realm Record.
+        1. Assert: _behaviour_ is either an Abstract Closure, a set of algorithm steps, or some other definition of a function's behaviour provided in this specification.
+        1. If _realm_ is not present, set _realm_ to the current Realm Record.
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
+        1. Let _func_ be a new built-in function object that, when called, performs the action described by _behaviour_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
```
bakkot, line 11642:
```suggestion
        1. Let _func_ be a new built-in function object that, when called, performs the action described by _behaviour_ using the provided arguments as the values of the corresponding parameters specified by _behaviour_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
```

per [discussion](https://github.com/tc39/ecma262/pull/2109#discussion_r632884824).

PR #2112, opened 2020-07-24
Editorial: Improve internal and external consistency of RegExp pattern semantics

gibson042:


spec.html
```diff
@@ -31878,23 +31882,21 @@ <h1>Assertion</h1>
...
             1. Assert: _c_ is a Continuation.
             1. Let _e_ be _x_'s _endIndex_.
-            1. Call IsWordChar(_e_ - 1) and let _a_ be the Boolean result.
-            1. Call IsWordChar(_e_) and let _b_ be the Boolean result.
-            1. If _a_ is *true* and _b_ is *false*, or if _a_ is *false* and _b_ is *true*, then
-              1. Call _c_(_x_) and return its result.
-            1. Return ~failure~.
+            1. Let _a_ be ! IsWordChar(_e_ - 1).
+            1. Let _b_ be ! IsWordChar(_e_).
+            1. If _a_ = _b_, return ~failure~.
```
bakkot, line 31887: > `_a_ = _b_`

Since equality is a confusing concept in JavaScript, I would prefer never to use `=` to talk about anything except mathematical values. I prefer the previous, more explicit text.

I also prefer keeping this as `1. If foo, return _c_(_x_). 1. Return ~failure~.`, rather than the other way around, for consistency with the rest of this section.

spec.html
```diff
@@ -31878,23 +31882,21 @@ <h1>Assertion</h1>
...
             1. Assert: _c_ is a Continuation.
             1. Let _e_ be _x_'s _endIndex_.
-            1. Call IsWordChar(_e_ - 1) and let _a_ be the Boolean result.
-            1. Call IsWordChar(_e_) and let _b_ be the Boolean result.
-            1. If _a_ is *true* and _b_ is *true*, or if _a_ is *false* and _b_ is *false*, then
-              1. Call _c_(_x_) and return its result.
-            1. Return ~failure~.
+            1. Let _a_ be ! IsWordChar(_e_ - 1).
+            1. Let _b_ be ! IsWordChar(_e_).
+            1. If _a_ &ne; _b_, return ~failure~.
```
bakkot, line 31898: Ditto above about `=` and order.

spec.html
```diff
@@ -32422,38 +32179,40 @@ <h1>AtomEscape</h1>
...
           <h1>Runtime Semantics: BackreferenceMatcher ( _n_, _direction_ )</h1>
           <p>The abstract operation BackreferenceMatcher takes arguments _n_ (an integer) and _direction_ (an integer). It performs the following steps when called:</p>
           <emu-alg>
+            1. Assert: _n_ &gt; 0.
             1. Return a new Matcher with parameters (_x_, _c_) that captures _n_ and _direction_ and performs the following steps when called:
               1. Assert: _x_ is a State.
               1. Assert: _c_ is a Continuation.
               1. Let _cap_ be _x_'s _captures_ List.
+              1. NOTE: _cap_ is indexed from 1 (<emu-xref href="#sec-notation"></emu-xref>).
               1. Let _s_ be _cap_[_n_].
```
bakkot, line 32216: As long as we're here, I'd be happier if we added an explicit bound check, as in

```
1. If the number of elements in _cap_ is less than _n_, let _s_ be *undefined*.
1. Let _s_ be _cap_[_n_].
```

spec.html
```diff
@@ -32502,27 +32261,27 @@ <h1>DecimalEscape</h1>
         <h1>CharacterClassEscape</h1>
         <p>The production <emu-grammar>CharacterClassEscape :: `d`</emu-grammar> evaluates as follows:</p>
         <emu-alg>
-          1. Return the ten-element set of characters containing the characters `0` through `9` inclusive.
+          1. Return the ten-element CharSet containing the characters `0` through `9` inclusive.
```
bakkot, line 32264:
```suggestion
          1. Return the ten-element CharSet whose elements are the characters `0` through `9` inclusive.
```

spec.html
```diff
@@ -32609,7 +32368,7 @@ <h1>Runtime Semantics: CharacterRange ( _A_, _B_ )</h1>
             1. Let _i_ be the character value of character _a_.
             1. Let _j_ be the character value of character _b_.
             1. Assert: _i_ &le; _j_.
-            1. Return the set containing all characters numbered _i_ through _j_, inclusive.
+            1. Return a CharSet containing all characters with a character value greater than or equal to _i_ and less than or equal to _j_.
```
bakkot, line 32371:
```suggestion
            1. Return the CharSet containing all characters with a character value greater than or equal to _i_ and less than or equal to _j_.
```

spec.html
```diff
@@ -32935,7 +32696,7 @@ <h1>Runtime Semantics: RegExpBuiltinExec ( _R_, _S_ )</h1>
               1. Let _groups_ be *undefined*.
             1. Perform ! CreateDataPropertyOrThrow(_A_, *"groups"*, _groups_).
             1. For each integer _i_ such that _i_ &gt; 0 and _i_ &le; _n_, do
```
bakkot, line 32698:
```suggestion
            1. For each integer _i_ such that _i_ &ge; 1 and _i_ &le; _n_, do
```

spec.html
```diff
@@ -32422,38 +32179,40 @@ <h1>AtomEscape</h1>
...
-          1. Call BackreferenceMatcher(_parenIndex_, _direction_) and return its Matcher result.
+          1. Let _parenIndex_ be the number of left-capturing parentheses in the entire regular expression that occur to the left of the located |GroupSpecifier|. This is the total number of <emu-grammar>Atom :: `(` GroupSpecifier Disjunction `)`</emu-grammar> Parse Nodes prior to or enclosing the located |GroupSpecifier|, including its immediately enclosing |Atom|.
+          1. Return ! BackreferenceMatcher(_parenIndex_, _direction_).
         </emu-alg>
 
         <emu-clause id="sec-backreference-matcher" aoid="BackreferenceMatcher">
           <h1>Runtime Semantics: BackreferenceMatcher ( _n_, _direction_ )</h1>
           <p>The abstract operation BackreferenceMatcher takes arguments _n_ (an integer) and _direction_ (an integer). It performs the following steps when called:</p>
           <emu-alg>
+            1. Assert: _n_ &gt; 0.
```
bakkot, line 32210:
```suggestion
            1. Assert: _n_ &ge; 1.
```

spec.html
```diff
@@ -31640,14 +31640,17 @@ <h1>Notation</h1>
           <li>
             _Unicode_ is *true* if the RegExp object's [[OriginalFlags]] internal slot contains *"u"* and otherwise is *false*.
           </li>
+          <li oldids="sec-runtime-semantics-wordcharacters-abstract-operation">
+            _WordCharacters_ is the mathematical set that is the union of all sixty-three characters in *"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_"* (letters, numbers, and U+005F (LOW LINE) in the Unicode Basic Latin block) and all characters _c_ for which _c_ is not in that set but Canonicalize(_c_) is. _WordCharacters_ cannot contain more than sixty-three characters unless _Unicode_ and _IgnoreCase_ are both *true*.
```
bakkot, line 31644: > `_WordCharacters_` cannot contain more than sixty-three characters unless `_Unicode_` and `_IgnoreCase_` are both `*true*`.

This scares me a little because it could become false in with a new version of Unicode, and we incorporate the latest version by reference. (Unless they've committed to not doing that, which would be great.)

But I guess it's fine.

spec.html
```diff
@@ -31640,14 +31640,17 @@ <h1>Notation</h1>
...
+          </li>
         </ul>
         <p>Furthermore, the descriptions below use the following internal data structures:</p>
         <ul>
           <li>
-            A <em>CharSet</em> is a mathematical set of characters, either code units or code points depending up the state of the _Unicode_ flag. &ldquo;All characters&rdquo; means either all code unit values or all code point values also depending upon the state of _Unicode_.
+            A <em>CharSet</em> is a mathematical set of characters. &ldquo;All characters&rdquo; means the CharSet containing all code point values when the _Unicode_ flag is *true*, and otherwise means the CharSet containing all code unit values.
```
bakkot, line 31650: Wordsmithing: I'd write this as

```suggestion
            A <em>CharSet</em> is a mathematical set of characters. When the _Unicode_ flag is *true* &ldquo;all characters&rdquo; means the CharSet containing all code point values; when it is *false* &ldquo;all characters&rdquo; means the CharSet containing all code unit values.
```

spec.html
gibson042: It's dependent only upon the [Canonicalize](https://tc39.es/ecma262/#sec-runtime-semantics-canonicalize-ch) operation continuing to use those same conditions as a guard for projecting its input into a single ASCII character. We're safe unless Unicode toUppercase starts mapping ASCII punctuation into letters, but I'll still try to clarify a bit.

spec.html

spec.html
gibson042: Your point about the ordering is fair, but I want to make the "do these tests return the same result?" logic underlying this algorithm more obvious... do you really think that `=` is hard to understand for Boolean values? And if so, how would you feel about using SameValue instead?

spec.html
gibson042: The length of _cap_ is always _NcapturingParens_ by both definition and assignment, and there's a fair amount of other code that reads directly from it. I'd like to leave this unless you insist upon adding an assertion.

spec.html
gibson042: Done.

spec.html
gibson042: Done.

spec.html
bakkot: > do you really think that `=` is hard to understand for Boolean values?

I think it requires a lot more thinking about it than the current wording, yes. Ditto for SameValue.

spec.html
bakkot: Ah, that's fine then. Somehow I convinced myself that wasn't necessarily so.

spec.html
bakkot: > We're safe unless Unicode toUppercase starts mapping ASCII punctuation into letters

Yeah, fair enough, that's not worth worrying about.

spec.html
gibson042: I'm not going to die on this hill, but I would like some other opinions.

bakkot: cc @ljharb @michaelficarra @syg thoughts?

ljharb: In JavaScript, `=` means assignment, and it causes a lot of bugs when people assume it works for comparison. I think using `=` for comparison in the spec for JavaScript would be very confusing.

I'm totally fine with SameValue - I love using AOs, and I wish [`BinaryXor`](https://tc39.es/ecma262/#sec-binaryxor) was trivially usable here - but the original wording is also fine to me.

spec.html
```diff
@@ -31957,7 +31957,7 @@ <h1>Notation</h1>
...
+            A <em>State</em> is an ordered pair (_endIndex_, _captures_) where _endIndex_ is an integer and _captures_ is a List of length _NcapturingParens_, indexed 1 through _NcapturingParens_. States are used to represent partial match states in the regular expression matching algorithms. The _endIndex_ is one plus the index of the last input character matched so far by the pattern, while _captures_ holds the results of capturing parentheses. The _n_<sup>th</sup> element of _captures_ is either a List of characters that represents the value obtained by the _n_<sup>th</sup> set of capturing parentheses or *undefined* if the _n_<sup>th</sup> set of capturing parentheses hasn't been reached yet. Due to backtracking, many States may be in use at any time during the matching process.
```
michaelficarra, line 31960: I don't like this 1-indexed list weirdness. I prefer the approach I took in https://github.com/tc39/ecma262/pull/2142#discussion_r472550811.

spec.html
bakkot: Editor group as a whole is in favor of keeping the existing wording, though we all had slightly different reasons for it (and @michaelficarra was fine with your proposed wording).

gibson042: Ok, reverted.

bakkot:
@gibson042: are you OK with dropping https://github.com/tc39/ecma262/pull/2112/commits/e82e75b2418bf6055bce181b2bc03988ad624043 and https://github.com/tc39/ecma262/pull/2112/commits/1f968cd0f2d5f0d1074ad64dd24411442b9efa36 from this PR and continuing discussion of those bits elsewhere? I know your preference is to land those as well before continuing the discussion, but @michaelficarra disagrees; the rest of this is uncontroversial and I don't want to block it on that point.

The other commits will also need rebasing into a collection of commits which can individually land; I'm happy to do that myself if you'd like.

Kurniawan1983 (approved):

gibson042:
@bakkot Done; I extracted the controversial commits into #2150 so they aren't lost.

spec.html
bakkot: Conversation moved to https://github.com/tc39/ecma262/pull/2150.

bakkot:
Thanks! I rebased the review commits into the applicable commit, so that these commits can all land rather than being squashed.

spec.html
```diff
@@ -31953,14 +31953,17 @@ <h1>Notation</h1>
           <li>
             _Unicode_ is *true* if the RegExp object's [[OriginalFlags]] internal slot contains *"u"* and otherwise is *false*.
           </li>
+          <li oldids="sec-runtime-semantics-wordcharacters-abstract-operation">
+            _WordCharacters_ is the mathematical set that is the union of all sixty-three characters in *"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_"* (letters, numbers, and U+005F (LOW LINE) in the Unicode Basic Latin block) and all characters _c_ for which _c_ is not in that set but Canonicalize(_c_) is. _WordCharacters_ cannot contain more than sixty-three characters unless _Unicode_ and _IgnoreCase_ are both *true*.
```
syg, line 31957: For _WordCharacters_ and _CharSet_, we could use the specification Set type, though the prose would need to be updated to say it is used in more places than just the memory model.

spec.html
```diff
@@ -32634,34 +32391,31 @@ <h1>Runtime Semantics: CharacterSetMatcher ( _A_, _invert_, _direction_ )</h1>
...
-            1. Else,
-              1. Assert: _ch_ is a UTF-16 code unit.
-              1. Let _s_ be the String value consisting of the single code unit _ch_.
-              1. Let _u_ be the same result produced as if by performing the algorithm for `String.prototype.toUpperCase` using _s_ as the *this* value.
-              1. Assert: Type(_u_) is String.
-              1. If _u_ does not consist of a single code unit, return _ch_.
-              1. Let _cu_ be _u_'s single code unit element.
-              1. If the numeric value of _ch_ &ge; 128 and the numeric value of _cu_ &lt; 128, return _ch_.
-              1. Return _cu_.
+            1. If _IgnoreCase_ is *false*, return _ch_.
```
syg, line 32410: Why was this branch moved down?

syg: We suspect that this is to call out that the first branch only applies when both _Unicode_ and _IgnoreCase_ is *true*, which would be a net readability improvement, so I'm fine with this.

PR #2113, opened 2020-07-26
Editorial: Add `IfAbruptCloseIterator` shorthand

ExE-Boss:
This adds the `IfAbruptCloseIterator` shorthand that matches the signature of [`IfAbruptRejectPromise`][IfAbruptRejectPromise].

Also, `?` is not necessary when doing <code>Return [IteratorClose][]\(…\)</code>, since the step will return regardless of whether the result is an abrupt completion.

[IfAbruptRejectPromise]: https://tc39.es/ecma262/#sec-ifabruptrejectpromise
[IteratorClose]: https://tc39.es/ecma262/#sec-iteratorclose


spec.html
```diff
@@ -6226,6 +6226,19 @@ <h1>IteratorClose ( _iteratorRecord_, _completion_ )</h1>
...
 
+    <emu-clause id="sec-ifabruptcloseiterator" aoid="IfAbruptCloseIterator">
+      <h1>IfAbruptCloseIterator ( _value_, _iteratorRecord_ )</h1>
+      <p>IfAbruptCloseIterator is a shorthand for a sequence of algorithm steps that use an Iterator Record. An algorithm step of the form:</p>
+      <emu-alg>
+        1. IfAbruptCloseIterator(_value_, _iteratorRecord_).
+      </emu-alg>
+      <p>means the same thing as:</p>
+      <emu-alg>
+        1. If _value_ is an abrupt completion, return IteratorClose(_iteratorRecord_, _value_).
```
ljharb, line 6237:
```suggestion
        1. If _value_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _value_).
```
Even though it’s being returned, I find the ? adds a lot of clarity, since it informs that this operation might return abruptly.

spec.html
```diff
@@ -6226,6 +6226,19 @@ <h1>IteratorClose ( _iteratorRecord_, _completion_ )</h1>
...
+    <emu-clause id="sec-ifabruptcloseiterator" aoid="IfAbruptCloseIterator">
+      <h1>IfAbruptCloseIterator ( _value_, _iteratorRecord_ )</h1>
+      <p>IfAbruptCloseIterator is a shorthand for a sequence of algorithm steps that use an Iterator Record. An algorithm step of the form:</p>
+      <emu-alg>
+        1. IfAbruptCloseIterator(_value_, _iteratorRecord_).
+      </emu-alg>
+      <p>means the same thing as:</p>
+      <emu-alg>
+        1. If _value_ is an abrupt completion, return ? IteratorClose(_iteratorRecord_, _value_).
+        1. Else if _value_ is a Completion Record, set _value_ to _value_.[[Value]].
```
michaelficarra, line 6238: FYI @bakkot, this should be assumed to be a completion record after completion value reform.

ExE-Boss: Well, this matches all the other `IfAbrupt` shorthands.

bakkot: Yup, I'll fix all of these when I get to this.

ljharb (approved):

ExE-Boss:
🔔 @bakkot @ljharb @michaelficarra @syg

PR #2115, opened 2020-07-28
Editorial: Define required internal slots in `CreateBuiltinFunction` steps

ExE-Boss:
~~The way <code>[CreateBuiltinFunction]</code> is currently specified would’ve resulted in it not having the `[[Prototype]]`, `[[Extensible]]`, `[[Realm]]` and `[[ScriptOrModule]]` internal slots unless they were explicitly passed using <code>[CreateBuiltinFunction][]\(<var>steps</var>, « [[Prototype]], [[Extensible]], [[Realm]], [[ScriptOrModule]] »\)</code>.~~

~~I’ve also made it optional in preparation for converting many of the built‑in functions to use abstract closures after <https://github.com/tc39/ecma262/pull/2109>.~~

[CreateBuiltinFunction]: https://tc39.es/ecma262/#sec-createbuiltinfunction

## See also:
- <https://github.com/tc39/ecma262/pull/2109>
- <https://github.com/tc39/ecma262/pull/2116>

---

[Preview](https://ci.tc39.es/preview/tc39/ecma262/pull/2115) ([#sec-createbuiltinfunction](https://ci.tc39.es/preview/tc39/ecma262/pull/2115#sec-createbuiltinfunction))

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13416,12 +13416,12 @@
   <emu-clause id="sec-built-in-function-objects">
     <h1>Built-in Function Objects</h1>
     <p>The built-in function objects defined in this specification may be implemented as either ECMAScript function objects (<emu-xref href="#sec-ecmascript-function-objects"></emu-xref>) whose behaviour is provided using ECMAScript code or as implementation provided function exotic objects whose behaviour is provided in some other manner. In either case, the effect of calling such functions must conform to their specifications. An implementation may also provide additional built-in function objects that are not defined in this specification.</p>
-    <p>If a built-in function object is implemented as an exotic object it must have the ordinary object behaviour specified in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>. All such function exotic objects also have [[Prototype]], [[Extensible]], and [[Realm]] internal slots.</p>
+    <p>If a built-in function object is implemented as an ECMAScript function object, it must have all the internal slots described in <emu-xref href="#sec-ecmascript-function-objects"></emu-xref> ([[Prototype]], [[Extensible]], and the slots listed in <emu-xref href="#table-internal-slots-of-ecmascript-function-objects"></emu-xref>), plus [[InitialName]]. The value of the [[InitialName]] internal slot is a String value that is the initial name of the function. It is used by <emu-xref href="#sec-function.prototype.tostring"></emu-xref>.</p>
+    <p>If a built-in function object is implemented as an exotic object, it must have the ordinary object behaviour specified in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>. All such function exotic objects have [[Prototype]], [[Extensible]], [[Realm]], and [[InitialName]] internal slots, with the same meanings as above.</p>
     <p>Unless otherwise specified every built-in function object has the %Function.prototype% object as the initial value of its [[Prototype]] internal slot.</p>
     <p>The behaviour specified for each built-in function via algorithm steps or other means is the specification of the function body behaviour for both [[Call]] and [[Construct]] invocations of the function. However, [[Construct]] invocation is not supported by all built-in functions. For each built-in function, when invoked with [[Call]], the [[Call]] _thisArgument_ provides the *this* value, the [[Call]] _argumentsList_ provides the named parameters, and the NewTarget value is *undefined*. When invoked with [[Construct]], the *this* value is uninitialized, the [[Construct]] _argumentsList_ provides the named parameters, and the [[Construct]] _newTarget_ parameter provides the NewTarget value. If the built-in function is implemented as an ECMAScript function object then this specified behaviour must be implemented by the ECMAScript code that is the body of the function. Built-in functions that are ECMAScript function objects must be strict functions. If a built-in constructor has any [[Call]] behaviour other than throwing a *TypeError* exception, an ECMAScript implementation of the function must be done in a manner that does not cause the function's [[IsClassConstructor]] internal slot to have the value *true*.</p>
     <p>Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function. When a built-in constructor is called as part of a `new` expression the _argumentsList_ parameter of the invoked [[Construct]] internal method provides the values for the built-in constructor's named parameters.</p>
     <p>Built-in functions that are not constructors do not have a *"prototype"* property unless otherwise specified in the description of a particular function.</p>
-    <p>Built-in functions have an [[InitialName]] internal slot.</p>
     <p>If a built-in function object is not implemented as an ECMAScript function it must provide [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
 
     <emu-clause id="sec-built-in-function-objects-call-thisargument-argumentslist" type="internal method">
@@ -13479,7 +13479,7 @@
           _behaviour_: unknown,
           _length_: a non-negative integer or +&infin;,
           _name_: a property key,
-          _internalSlotsList_: a List of names of internal slots,
+          _additionalInternalSlotsList_: a List of names of internal slots,
           optional _realm_: a Realm Record,
           optional _prototype_: an Object or *null*,
           optional _prefix_: a String,
@@ -13487,17 +13487,19 @@
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>_internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object.</dd>
+        <dd>_additionalInternalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object.</dd>
       </dl>
       <emu-alg>
         1. Assert: _behaviour_ is either an Abstract Closure, a set of algorithm steps, or some other definition of a function's behaviour provided in this specification.
         1. If _realm_ is not present, set _realm_ to the current Realm Record.
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
+        1. Let _internalSlotsList_ be a List containing the names of all the internal slots that <emu-xref href="#sec-built-in-function-objects"></emu-xref> requires for the built-in function object that is about to be created.
+        1. Append to _internalSlotsList_ the elements of _additionalInternalSlotsList_.
         1. Let _func_ be a new built-in function object that, when called, performs the action described by _behaviour_ using the provided arguments as the values of the corresponding parameters specified by _behaviour_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
-        1. Set _func_.[[Realm]] to _realm_.
         1. Set _func_.[[Prototype]] to _prototype_.
         1. Set _func_.[[Extensible]] to *true*.
+        1. Set _func_.[[Realm]] to _realm_.
         1. Set _func_.[[InitialName]] to *null*.
         1. Perform ! SetFunctionLength(_func_, _length_).
         1. If _prefix_ is not present, then


anba:
Built-in functions can be implemented as ECMAScript function objects, in which case they have even more internal slots. If they are not implemented as ECMAScript function objects, but instead as implementation provided function exotic objects, they are already described as [required to have these slots](https://tc39.es/ecma262/#sec-built-in-function-objects):
> If a built-in function object is implemented as an exotic object it must have the ordinary object behaviour specified in 9.1. All such function exotic objects also have [[Prototype]], [[Extensible]], [[Realm]], and [[ScriptOrModule]] internal slots.

ExE-Boss:
> they are already described as [required to have these slots](https://tc39.es/ecma262/#sec-built-in-function-objects):

This makes that requirement explicit, rather than implicit.

anba:
But it gives the impression that only [[Prototype]], [[Extensible]], [[Realm]], and [[ScriptOrModule]] are needed, which isn't true for  ECMAScript function objects. If this needs to be explicit, it should have two branches to cover the case if the built-in function is an ECMAScript function object or an exotic object. And in each branch the appropriate slots should be initialised. (For example right now the spec also doesn't cover how the other internal slots for a built-in function which is an ECMAScript function object are initialised. This is all left to be defined in an actual implementation.)

spec.html
```diff
@@ -8871,17 +8871,19 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
+      <h1>CreateBuiltinFunction ( _steps_ [ , _additionalInternalSlotsList_ [ , _realm_ [ , _prototype_ ] ] ] )</h1>
+      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_ and optional arguments _additionalInternalSlotsList_ (a List of internal slot names), _realm_, and _prototype_. _additionalInternalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
       <emu-alg>
         1. Assert: _steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification.
         1. If _realm_ is not present, set _realm_ to the current Realm Record.
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
+        1. Let _internalSlotsList_ be &laquo; [[Prototype]], [[Extensible]], [[Realm]], [[ScriptOrModule]] &raquo;.
```
ljharb, line 8881: I'm not sure this is needed.

https://tc39.es/ecma262/#sec-ecmascript-function-objects already dictates that `[[Realm]]` and `[[ScriptOrModule]]` implicitly exist on a built-in function object, and https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots dictates `[[Prototype]]` and `[[Extensible]]`, so they don't need to be explicitly specified.

ljharb (dismissed):

ljharb:
The requirement is already explicit - it's just not in this location.

spec.html
```diff
@@ -9049,17 +9049,22 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
         1. Assert: _steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification.
         1. If _realm_ is not present, set _realm_ to the current Realm Record.
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
-        1. Set _func_.[[Realm]] to _realm_.
-        1. Set _func_.[[Prototype]] to _prototype_.
+        1. If the built-in function object is implemented using ECMAScript code, then
+          1. Let _internalSlotsList_ be the internal slots listed in <emu-xref href="#table-27"></emu-xref>.
+        1. Otherwise, let _internalSlotsList_ be &laquo; [[Prototype]], [[Extensible]], [[Realm]], [[ScriptOrModule]] &raquo;.
```
ljharb, line 9061: This also seems unnecessary - `func` is "a new built-in function object", so it *automatically* has these slots with no further spec text.

ExE-Boss: I’d like to convert `CreateBuiltinFunction` to use `MakeBasicObject` if possible.

jmdyck: > I’d like to convert `CreateBuiltinFunction` to use `MakeBasicObject` if possible.

I imagine it'd be possible, but the algorithm would have to convey that the implementation gets to choose whether the resulting function is ordinary or exotic. I'm not sure it'd be an improvement.

spec.html
```diff
@@ -9049,17 +9049,22 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
       <emu-alg>
         1. Assert: _steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification.
         1. If _realm_ is not present, set _realm_ to the current Realm Record.
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
-        1. Set _func_.[[Realm]] to _realm_.
-        1. Set _func_.[[Prototype]] to _prototype_.
+        1. If the built-in function object is implemented using ECMAScript code, then
```
bakkot, line 9059: I am against this change. Switching on implementation details in normative prose seems bad to me.

jmdyck: In the current spec, I think the only obvious example of switching on an implementation detail is:
- `If an implementation-defined debugging facility is available and enabled, ...`.

For "host", there's:
- `If the host requires use of an exotic object to serve as realm's global object, ...`
- `If the host requires that the *this* binding in realm's global scope return an object other than the global object, ...`
- `If <various conditions> and ! HostHasSourceTextAvailable(_func_) is *true*, ...`

That's all I could find.

spec.html
```diff
@@ -9117,17 +9117,20 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
         1. If _realm_ is not present, set _realm_ to the current Realm Record.
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
-        1. Set _func_.[[Realm]] to _realm_.
-        1. Set _func_.[[Prototype]] to _prototype_.
+        1. Let _internalSlotsList_ be the internal slots specified in <emu-xref href="#sec-built-in-function-objects"></emu-xref>.
+        1. Append [[InitialName]] to _internalSlotsList_.
+        1. Append each element of _additionalInternalSlotsList_ to _internalSlotsList_.
+        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_.
```
jmdyck, line 9130: As @anba pointed out: if `_func_` is an ECMAScript (i.e. ordinary) function object, it must have all the internal slots listed [here](https://tc39.es/ecma262/#table-internal-slots-of-ecmascript-function-objects), but this operation doesn't mention most of them or say how they're initialized. Of course, that's been true since `CreateBuiltinFunction` was introduced, but this PR appears to be trying for completeness. I think it's somewhat misleading in its current state.

jmdyck (changes requested):

spec.html
ExE-Boss: I’m not sure how I should specify that given @bakkot’s review in <https://github.com/tc39/ecma262/pull/2115#discussion_r467621952>.

spec.html
jmdyck: @bakkot:
> Switching on implementation details in normative prose seems bad to me.

[9.3 Built-in Function Objects](https://www.ecma-international.org/ecma-262/#sec-built-in-function-objects) says that *if* you implement a built-in function as an exotic function, then it must satisfy some extra requirements (that run-of-the-mill exotic functions don't have to). So that is normative prose that is switching on an implementation 'detail'.  The proposed text that you were objecting to seems comparable, just in algorithmic form.

spec.html
jmdyck: If @bakkot continues to object, then I think the solution is for `CreateBuiltinFunction` to not be as specific about slots as you'd like. E.g., you could ditch the construction of `_internalSlotsList_` and then say something like: `The new function object has the internal slots required by either (9.2) or (9.3), as appropriate, and also has internal slots whose names are the elements of _additionalInternalSlotsList_.`

(It *should* be possible to replace "either (9.2) or (9.3), as appropriate" with just "(9.3)", since 9.3 deals with all built-in functions, and covers the ordinary/exotic implementation choice. But I think, as 9.3 currently stands, that would be a little misleading, as readers would only look at the slots explicitly mentioned in 9.3 and not consider those included by reference to 9.2.)

spec.html
```diff
@@ -8592,17 +8645,19 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
...
       <emu-alg>
         1. Assert: _steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification.
         1. If _realm_ is not present, set _realm_ to the current Realm Record.
         1. Assert: _realm_ is a Realm Record.
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
-        1. Let _func_ be a new built-in function object that when called performs the action described by _steps_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
-        1. Set _func_.[[Realm]] to _realm_.
-        1. Set _func_.[[Prototype]] to _prototype_.
+        1. Let _internalSlotsList_ be &laquo; [[Prototype]], [[Extensible]] &raquo; and the internal slots specified in <emu-xref href="#table-internal-slots-of-built-in-function-objects"></emu-xref> or <emu-xref href="#table-internal-slots-of-ecmascript-function-objects"></emu-xref> and &laquo; [[InitialName]] &raquo;, as appropriate.
```
jmdyck, line 8655:
```suggestion
        1. Let _internalSlotsList_ be a List containing [[Prototype]], [[Extensible]], [[InitialName]], and the internal slots specified in <emu-xref href="#table-internal-slots-of-ecmascript-function-objects"></emu-xref> or <emu-xref href="#table-internal-slots-of-built-in-function-objects"></emu-xref>, as appropriate.

jmdyck, line 8655: For exotic built-ins, this includes [[InitialName]] twice.

jmdyck (changes requested):

spec.html
ExE-Boss: So does [your version](https://github.com/tc39/ecma262/pull/2115#discussion_r493967623).

jmdyck: Yes, because that suggestion wasn't attempting to resolve this problem. That's why I raised it separately.

jmdyck:
I've been requested to review this, and I'm willing to do so, but I think the merge conflicts should be resolved first. In particular, note that #2190 dropped the `[[ScriptOrModule]]` slot for exotic built-ins.

ljharb:
@jmdyck i've rebased it; feel free to take a look.

jmdyck (changes requested):
It turned out my comments wouldn't have fit very well into the "suggested change" boxes, so instead I filed https://github.com/ExE-Boss/ecma262/pull/1 against this PR's branch.

jmdyck:
It's still the case (as it has been since CreateBuiltinFunction was introduced) that if the function being created is an ordinary function object, CreateBuiltinFunction doesn't say how most of its internal slots are initialized. Without trying to 'fix' that shortcoming, we could maybe at least acknowledge it in a Note. E.g.:
> NOTE: If _func_ is an ECMAScript function object, it has many other internal slots that must be initialized, but this operation does not address how that happens.

With or without that, I think this PR is somewhat clearer than the status quo about internal slots of built-ins.

michaelficarra (approved):
This looks good, but it will need to be rebased (due to #545) to have the proper header format for `CreateBuiltinFunction`.

bakkot:
Done.

bakkot (approved):

PR #2116, opened 2020-07-28
Normative: Set function `name` and `length` in `CreateBuiltinFunction`

ExE-Boss:
This PR is normative as it defines the order in which the `length` and `name` properties are added to built-in functions.

I’ve added `name` as a parameter, since abstract closures don’t have names, which would make using them to create non‑anonymous functions impossible.

It also preserves compatibility [for WebIDL](https://heycam.github.io/webidl/#create-an-interface-object), which uses inline definitions for steps.

**Tests:** <https://github.com/tc39/test262/pull/2921>

## Unsolved issues:
- [x] Passing the `prefix` parameter to `SetFunctionName` (<https://github.com/tc39/ecma262/pull/2116#issuecomment-744127225>)

## Blocks:
- <https://github.com/tc39/ecma262/pull/2109>
- <https://github.com/tc39/ecma262/pull/2216>

## See also:
- <https://github.com/tc39/ecma262/pull/2115>
- <https://github.com/tc39/ecma262/pull/2118>
- <https://github.com/tc39/ecma262/pull/2219>

---

[Preview](https://ci.tc39.es/preview/tc39/ecma262/pull/2116) ([#sec-createbuiltinfunction](https://ci.tc39.es/preview/tc39/ecma262/pull/2116#sec-createbuiltinfunction))

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3592,10 +3592,12 @@
           1. Let _asyncContext_ be the running execution context.
           1. Let _promise_ be ? PromiseResolve(%Promise%, _value_).
           1. Let _stepsFulfilled_ be the algorithm steps defined in <emu-xref href="#await-fulfilled" title></emu-xref>.
-          1. Let _onFulfilled_ be ! CreateBuiltinFunction(_stepsFulfilled_, &laquo; [[AsyncContext]] &raquo;).
+          1. Let _lengthFulfilled_ be the number of non-optional parameters of the function definition in <emu-xref href="#await-fulfilled" title></emu-xref>.
+          1. Let _onFulfilled_ be ! CreateBuiltinFunction(_stepsFulfilled_, _lengthFulfilled_, *""*, &laquo; [[AsyncContext]] &raquo;).
           1. Set _onFulfilled_.[[AsyncContext]] to _asyncContext_.
           1. Let _stepsRejected_ be the algorithm steps defined in <emu-xref href="#await-rejected" title></emu-xref>.
-          1. Let _onRejected_ be ! CreateBuiltinFunction(_stepsRejected_, &laquo; [[AsyncContext]] &raquo;).
+          1. Let _lengthRejected_ be the number of non-optional parameters of the function definition in <emu-xref href="#await-rejected" title></emu-xref>.
+          1. Let _onRejected_ be ! CreateBuiltinFunction(_stepsRejected_, _lengthRejected_, *""*, &laquo; [[AsyncContext]] &raquo;).
           1. Set _onRejected_.[[AsyncContext]] to _asyncContext_.
           1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
@@ -9511,7 +9513,7 @@
       <emu-alg>
         1. Let _intrinsics_ be a new Record.
         1. Set _realmRec_.[[Intrinsics]] to _intrinsics_.
-        1. Set fields of _intrinsics_ with the values listed in <emu-xref href="#table-well-known-intrinsic-objects"></emu-xref>. The field names are the names listed in column one of the table. The value of each field is a new object value fully and recursively populated with property values as defined by the specification of each object in clauses <emu-xref href="#sec-global-object"></emu-xref> through <emu-xref href="#sec-reflection"></emu-xref>. All object property values are newly created object values. All values that are built-in function objects are created by performing CreateBuiltinFunction(_steps_, _slots_, _realmRec_, _prototype_) where _steps_ is the definition of that function provided by this specification, _slots_ is a list of the names, if any, of the function's specified internal slots, and _prototype_ is the specified value of the function's [[Prototype]] internal slot. The creation of the intrinsics and their properties must be ordered to avoid any dependencies upon objects that have not yet been created.
+        1. Set fields of _intrinsics_ with the values listed in <emu-xref href="#table-well-known-intrinsic-objects"></emu-xref>. The field names are the names listed in column one of the table. The value of each field is a new object value fully and recursively populated with property values as defined by the specification of each object in clauses <emu-xref href="#sec-global-object"></emu-xref> through <emu-xref href="#sec-reflection"></emu-xref>. All object property values are newly created object values. All values that are built-in function objects are created by performing CreateBuiltinFunction(_steps_, _length_, _name_, _slots_, _realmRec_, _prototype_) where _steps_ is the definition of that function provided by this specification, _name_ is the initial value of the function's `name` property, _length_ is the initial value of the function's `length` property, _slots_ is a list of the names, if any, of the function's specified internal slots, and _prototype_ is the specified value of the function's [[Prototype]] internal slot. The creation of the intrinsics and their properties must be ordered to avoid any dependencies upon objects that have not yet been created.
         1. Perform AddRestrictedFunctionProperties(_intrinsics_.[[%Function.prototype%]], _realmRec_).
         1. Return _intrinsics_.
       </emu-alg>
@@ -11166,8 +11168,8 @@
     </emu-clause>
 
     <emu-clause id="sec-createbuiltinfunction" aoid="CreateBuiltinFunction">
-      <h1>CreateBuiltinFunction ( _steps_, _internalSlotsList_ [ , _realm_ [ , _prototype_ ] ] )</h1>
-      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_ and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_ and _prototype_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
+      <h1>CreateBuiltinFunction ( _steps_, _length_, _name_, _internalSlotsList_ [ , _realm_ [ , _prototype_ [ , _prefix_ ] ] ] )</h1>
+      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_, _length_, _name_, and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_, _prototype_, and _prefix_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
       <emu-alg>
         1. Assert: _steps_ is either a set of algorithm steps or other definition of a function's behaviour provided in this specification.
         1. If _realm_ is not present, set _realm_ to the current Realm Record.
@@ -11178,6 +11180,11 @@
         1. Set _func_.[[Prototype]] to _prototype_.
         1. Set _func_.[[Extensible]] to *true*.
         1. Set _func_.[[InitialName]] to *null*.
+        1. Perform ! SetFunctionLength(_func_, _length_).
+        1. If _prefix_ is not present, then
+          1. Perform ! SetFunctionName(_func_, _name_).
+        1. Else,
+          1. Perform ! SetFunctionName(_func_, _name_, _prefix_).
         1. Return _func_.
       </emu-alg>
       <p>Each built-in function defined in this specification is created by calling the CreateBuiltinFunction abstract operation.</p>
@@ -11668,7 +11675,8 @@
           <p>The abstract operation MakeArgGetter takes arguments _name_ (a String) and _env_ (an Environment Record). It creates a built-in function object that when executed returns the value bound for _name_ in _env_. It performs the following steps when called:</p>
           <emu-alg>
             1. Let _steps_ be the steps of an ArgGetter function as specified below.
-            1. Let _getter_ be ! CreateBuiltinFunction(_steps_, &laquo; [[Name]], [[Env]] &raquo;).
+            1. Let _length_ be the number of non-optional parameters of an ArgGetter function as specified below.
+            1. Let _getter_ be ! CreateBuiltinFunction(_steps_, _length_, *""*, &laquo; [[Name]], [[Env]] &raquo;).
             1. Set _getter_.[[Name]] to _name_.
             1. Set _getter_.[[Env]] to _env_.
             1. Return _getter_.
@@ -11690,7 +11698,8 @@
           <p>The abstract operation MakeArgSetter takes arguments _name_ (a String) and _env_ (an Environment Record). It creates a built-in function object that when executed sets the value bound for _name_ in _env_. It performs the following steps when called:</p>
           <emu-alg>
             1. Let _steps_ be the steps of an ArgSetter function as specified below.
-            1. Let _setter_ be ! CreateBuiltinFunction(_steps_, &laquo; [[Name]], [[Env]] &raquo;).
+            1. Let _length_ be the number of non-optional parameters of an ArgSetter function as specified below.
+            1. Let _setter_ be ! CreateBuiltinFunction(_steps_, _length_, *""*, &laquo; [[Name]], [[Env]] &raquo;).
             1. Set _setter_.[[Name]] to _name_.
             1. Set _setter_.[[Env]] to _env_.
             1. Return _setter_.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -8871,18 +8871,24 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
     </emu-clause>
 
     <emu-clause id="sec-createbuiltinfunction" aoid="CreateBuiltinFunction">
-      <h1>CreateBuiltinFunction ( _steps_, _internalSlotsList_ [ , _realm_ [ , _prototype_ ] ] )</h1>
-      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_ and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_ and _prototype_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
+      <h1>CreateBuiltinFunction ( _steps_ [ , _name_ [ , _length_ [ , _additionalInternalSlotsList_ [ , _realm_ [ , _prototype_ ] ] ] ] ] )</h1>
```
bakkot, line 8874: I'd prefer to always be explicit about the name and length, rather than introducing new optional parameters.

ljharb:
I'm confused; the steps themselves should already specify the name and length. Why are they needed here?

bakkot:
> I'm confused; the steps themselves should already specify the name and length.

They do not.

ljharb:
For example, https://tc39.es/ecma262/#await-fulfilled says "The "length" property of an Await fulfilled function is 1.", and since it says they're anonymous, that means the name is the empty string.

Any "steps" that don't specify these should be fixed to do so.

bakkot:
There is _prose_ which specifies these, yes, but not steps. It is odd to explicitly create an object and have steps to set up some of its properties, but have others be described only in prose, which is the current state of affairs. 

(The concrete reason to add these as parameters is so that they can be used when passing an abstract closure instead of a sequence of steps, in which case there is no obvious place to put normative prose.)

ljharb:
The length and name of a function defined from steps should imo be determined at the steps' location, and not where the steps happen to be used.

bakkot:
I am fine with _mentioning_ it in those locations, but it seems pretty odd to say that's where it should be _determined_, given that the place the steps are used is the place that sets up internal slots, etc.

(Also, I would like to eventually inline most such steps into the callsites of CreateBuiltInFunction, as in https://github.com/tc39/ecma262/pull/2109, and this is effectively a prerequisite for that, since after such an inlining there is no longer any other place to put the name and length.)

ljharb:
To me, the steps are what defines the function. Just like how methods define the steps, name, and length in one place, I think these ideally should too.

In other words, I'd prefer to see all these things moved to the steps definition location, rather than moving more of them out to the callsite.

spec.html
```diff
@@ -8883,6 +8883,10 @@ <h1>CreateBuiltinFunction ( _steps_, _internalSlotsList_ [ , _realm_ [ , _protot
         1. Set _func_.[[Prototype]] to _prototype_.
         1. Set _func_.[[Extensible]] to *true*.
         1. Set _func_.[[ScriptOrModule]] to *null*.
+        1. Let _length_ be the number of parameters taken by _steps_.
```
ljharb, line 8886:
```suggestion
        1. Let _length_ be the number of non-optional parameters taken by _steps_.
```

spec.html
ljharb:
```suggestion
      <h1>CreateBuiltinFunction ( _steps_, _name_, _length_, _additionalInternalSlotsList_ [ , _realm_ [ , _prototype_ ] ] ] ] ] )</h1>
```

ExE-Boss:
Arguably, the <var>length</var> parameter should be optional, as it can be determined inside [CreateBuiltinFunction].

[CreateBuiltinFunction]: https://tc39.es/ecma262/#sec-createbuiltinfunction

spec.html
```diff
@@ -8871,8 +8874,8 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
     </emu-clause>
 
     <emu-clause id="sec-createbuiltinfunction" aoid="CreateBuiltinFunction">
-      <h1>CreateBuiltinFunction ( _steps_, _internalSlotsList_ [ , _realm_ [ , _prototype_ ] ] )</h1>
-      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_ and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_ and _prototype_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
+      <h1>CreateBuiltinFunction ( _steps_, _name_, _length_, _internalSlotsList_ [ , _realm_ [ , _prototype_ ] ] )</h1>
+      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_ and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_, _prototype_, and _name_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
```
ljharb, line 8878:
```suggestion
      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_ and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_ and _prototype_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
```

spec.html
```diff
@@ -8871,8 +8874,8 @@ <h1>[[Construct]] ( _argumentsList_, _newTarget_ )</h1>
     </emu-clause>
 
     <emu-clause id="sec-createbuiltinfunction" aoid="CreateBuiltinFunction">
-      <h1>CreateBuiltinFunction ( _steps_, _internalSlotsList_ [ , _realm_ [ , _prototype_ ] ] )</h1>
-      <p>The abstract operation CreateBuiltinFunction takes arguments _steps_ and _internalSlotsList_ (a List of names of internal slots) and optional arguments _realm_ and _prototype_. _internalSlotsList_ contains the names of additional internal slots that must be defined as part of the object. This operation creates a built-in function object. It performs the following steps when called:</p>
+      <h1>CreateBuiltinFunction ( _steps_, _name_, _length_, _internalSlotsList_ [ , _realm_ [ , _prototype_ ] ] )</h1>
```
bakkot, line 8877: Nit: since `length` is used before `name`, I'd prefer the parameters occur in that order as well.

spec.html
```diff
@@ -24972,7 +24979,7 @@ <h1>ECMAScript Standard Built-in Objects</h1>
...
   <p>Unless otherwise specified every built-in prototype object has the Object prototype object, which is the initial value of the expression `Object.prototype` (<emu-xref href="#sec-properties-of-the-object-prototype-object"></emu-xref>), as the value of its [[Prototype]] internal slot, except the Object prototype object itself.</p>
   <p>Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function.</p>
-  <p>Each built-in function defined in this specification is created by calling the CreateBuiltinFunction abstract operation (<emu-xref href="#sec-createbuiltinfunction"></emu-xref>).</p>
+  <p>Each built-in function defined in this specification is created by calling the CreateBuiltinFunction abstract operation (<emu-xref href="#sec-createbuiltinfunction"></emu-xref>). The values of the _name_ and _length_ parameters are the initial values of the _length_ and _name_ properties as discussed below.</p>
```
bakkot, line 24982:
```suggestion
  <p>Each built-in function defined in this specification is created by calling the CreateBuiltinFunction abstract operation (<emu-xref href="#sec-createbuiltinfunction"></emu-xref>). The values of the _length_ and _name_ parameters are the initial values of the _length_ and _name_ properties as discussed below.</p>
```

bakkot:
@ExE-Boss I believe you're not a TC39 delegate and so can't present this in plenary (which needs to happen, since it's normative). If that's so, I'll present it next meeting if you're alright with that.

ExE-Boss:
@bakkot Feel free to do that.

kmiller68:
JSC should match the new behavior now, although, it's possible I missed some cases... See: https://bugs.webkit.org/show_bug.cgi?id=216789

rwaldron:
It's unclear to me what tests need to be written for this. The existing tests for all built-in function object's name and length properties should be sufficient, right?

ljharb:
@rwaldron if those existing tests also assert the relative order of those properties, then they're sufficient.

rwaldron:
> @rwaldron if those existing tests also assert the relative order of those properties, then they're sufficient.

Sorry, not sure I understand: do you mean the relative order in which those properties are set?

michaelficarra:
@rwaldron Property enumeration order.

rwaldron:
> @rwaldron Property enumeration order.

Ah, I see. Thanks.

syg (approved):

spec.html
```diff
@@ -39974,16 +39989,18 @@ <h1>Runtime Semantics: PerformPromiseAllSettled ( _iteratorRecord_, _constructor
               1. ReturnIfAbrupt(_nextValue_).
               1. Append *undefined* to _values_.
               1. Let _nextPromise_ be ? Call(_promiseResolve_, _constructor_, &laquo; _nextValue_ &raquo;).
-              1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-promise.allsettled-resolve-element-functions" title></emu-xref>.
-              1. Let _resolveElement_ be ! CreateBuiltinFunction(_steps_, &laquo; [[AlreadyCalled]], [[Index]], [[Values]], [[Capability]], [[RemainingElements]] &raquo;).
+              1. Let _resolveSteps_ be the algorithm steps defined in <emu-xref href="#sec-promise.allsettled-resolve-element-functions" title></emu-xref>.
+              1. Let _resolveLength_ be the number of non-optional parameters of the function definition in <emu-xref href="#sec-promise.allsettled-resolve-element-functions" title></emu-xref>.
```
bakkot, line 39993: Nit: the convention in the rest of this PR would suggest that the name of this should be `lengthResolve`, not `resolveLength`. For consistency, I think `resolveSteps` and `resolveLength` as well as both instances of each of `rejectSteps` and `rejectLength` should all be switched around.

(I assume you chose these names to match the existing `rejectSteps` variable, but I note that the `Await` macro uses `stepsFulfilled`, so there's not a clear precedent. I think it's worth renaming that variable so there's consistency.)

bakkot (approved):
LGTM other than extremely minor nit.

spec.html
ExE-Boss: That’s probably best left for an editorial PR that unifies this.

bakkot: You're introducing new variable names here, and those names should be consistent - so, `stepsResolve` and `lengthResolve`, not (as you have here) `resolveSteps` and `resolveLength`. And you might as well unify the existing inconsistent one while you're at it. I'm happy to do that for you if you'd prefer.

ExE-Boss: Actually, `stepsFulfilled` and `onFulfilled` is used more.

---

`stepsResolve` and `stepsReject` are only used in <code>[CreateResolvingFunctions](https://tc39.es/ecma262/#sec-createresolvingfunctions)</code>.

bakkot: Works for me. I don't really care which convention we use as long as we're consistent throughout the spec. (The single `rejectSteps`, in `PerformPromiseAllSettled`, is still inconsistent, but we can fix it in a followup if you'd prefer to leave it alone here.)

ExE-Boss: I thought I fixed that.

ljharb:
Are there test262 tests yet for this change?

ExE-Boss:
@ljharb There are now: <https://github.com/tc39/ecma262/pull/2116>

ExE-Boss:
**Question:** How should this pass a prefix to `SetFunctionName`?

bakkot:
@ExE-Boss:

> How should this pass a prefix to `SetFunctionName`?

I'm not sure what you mean by this. Why would you want to pass a prefix?

ExE-Boss:
Because built‑in accessor properties (e.g.: `get %TypedArray%.prototype[Symbol.toStringTag]`) have a `name` of `Symbol.toStringTag` and `prefix` of `"get"`, so that the resulting `%TypedArray%.prototype[Symbol.toStringTag].name` is `"get [Symbol.toStringTag]"` and so that `[[InitialName]]` is set correctly.

bakkot:
Ah, got it. There's not a place where that would come up currently, right? It's just a matter of having

> All values that are built-in function objects are created by performing CreateBuiltinFunction

make sense?

ExE-Boss:
> Ah, got it. There's not a place where that would come up currently, right? It's just a matter of having
> 
> > All values that are built-in function objects are created by performing CreateBuiltinFunction
> 
> make sense?

Pretty much, also it does come up in **WebIDL**.

jmdyck:
> Ah, got it. There's not a place where that would come up currently, right?

I think you're saying that, in all the places where CreateBuiltinFunction is explicitly invoked, the resulting function doesn't need a name with a prefix, so having CreateBuiltinFunction omit the `_prefix_` argument when invoking SetFunctionName works for them.

> It's just a matter of having
> 
> > All values that are built-in function objects are created by performing CreateBuiltinFunction
> 
> make sense?

Here you're talking about the implicit invocations of CreateBuiltinFunction, but I'm not sure what you're saying. What do you mean by "have [that sentence] make sense"?

bakkot:
> I think you're saying that, in all the places where CreateBuiltinFunction is explicitly invoked, the resulting function doesn't need a name with a prefix, so having CreateBuiltinFunction omit the `_prefix_` argument when invoking SetFunctionName works for them.

Right.

> Here you're talking about the implicit invocations of CreateBuiltinFunction, but I'm not sure what you're saying. What do you mean by "have [that sentence] make sense"?

If CreateBuiltinFunction does not pass a `_prefix_` to SetFunctionName, then there is no way that an invocation of CreateBuiltinFunction could result in a function whose name has a prefix. But some built-in functions _do_ have names which have prefixes, as in `Object.getOwnPropertyDescriptor(Uint8Array.prototype.__proto__, Symbol.toStringTag).get.name`. So how could such a function have been created by CreateBuiltinFunction, as this sentence claims?

jmdyck:
> If CreateBuiltinFunction does not pass a `_prefix_` to SetFunctionName, then there is no way that an invocation of CreateBuiltinFunction could result in a function whose name has a prefix. But some built-in functions _do_ have names which have prefixes, as in `Object.getOwnPropertyDescriptor(Uint8Array.prototype.__proto__, Symbol.toStringTag).get.name`. So how could such a function have been created by CreateBuiltinFunction, as this sentence claims?

Okay, good. (That's the problem I saw too.)

spec.html
```diff
@@ -6968,7 +6971,7 @@ <h1>CreateIntrinsics ( _realmRec_ )</h1>
...
+        1. Set fields of _intrinsics_ with the values listed in <emu-xref href="#table-well-known-intrinsic-objects"></emu-xref>. The field names are the names listed in column one of the table. The value of each field is a new object value fully and recursively populated with property values as defined by the specification of each object in clauses <emu-xref href="#sec-global-object"></emu-xref> through <emu-xref href="#sec-reflection"></emu-xref>. All object property values are newly created object values. All values that are built-in function objects are created by performing CreateBuiltinFunction(_steps_, _length_, _name_, _slots_, _realmRec_, _prototype_) where _steps_ is the definition of that function provided by this specification, _name_ is the initial value of the function's `name` property, _length_ is the initial value of the function's `length` property, , _slots_ is a list of the names, if any, of the function's specified internal slots, and _prototype_ is the specified value of the function's [[Prototype]] internal slot. The creation of the intrinsics and their properties must be ordered to avoid any dependencies upon objects that have not yet been created.
```
bakkot, line 6974:
```suggestion
        1. Set fields of _intrinsics_ with the values listed in <emu-xref href="#table-well-known-intrinsic-objects"></emu-xref>. The field names are the names listed in column one of the table. The value of each field is a new object value fully and recursively populated with property values as defined by the specification of each object in clauses <emu-xref href="#sec-global-object"></emu-xref> through <emu-xref href="#sec-reflection"></emu-xref>. All object property values are newly created object values. All values that are built-in function objects are created by performing CreateBuiltinFunction(_steps_, _length_, _name_, _slots_, _realmRec_, _prototype_) where _steps_ is the definition of that function provided by this specification, _name_ is the initial value of the function's `name` property, _length_ is the initial value of the function's `length` property, _slots_ is a list of the names, if any, of the function's specified internal slots, and _prototype_ is the specified value of the function's [[Prototype]] internal slot. The creation of the intrinsics and their properties must be ordered to avoid any dependencies upon objects that have not yet been created.
```

ljharb: had to diff this locally to figure out this was removing an excess `, ` between `property` and `_slots`.

syg:
Note to self: figure out if latitude is needed for the web reality of builtin getters and setters and the presence of the `get` and `set` prefixes in the function names.

spec.html
```diff
@@ -24705,13 +24712,13 @@ <h1>ECMAScript Standard Built-in Objects</h1>
...
+  <p>Every built-in function object, including constructors, has a *"name"* property whose value is a String. Unless otherwise specified, this value is the name that is given to the function in this specification. Functions that are identified as anonymous functions use the empty String as the value of the *"name"* property. For functions that are specified as properties of objects, the name value is the property name string used to access the function. Functions that are specified as get or set accessor functions of built-in properties have *"get "* or *"set "* prepended to the property name string. The value of the *"name"* property is explicitly specified for each built-in functions whose property key is a Symbol value.</p>
```
syg, line 24721: All engines agree that the `name` property for built-in accessors have `"get "` or `"set "` prepended, depending on the type of accessor.

For Symbol-named getters like `@@toStringTag`, their `name` properties are indeed explicitly defined.

syg (approved):
Implementation reality doesn't need extra latitude for accessors, so by reading this PR is good to go.

spec.html
bakkot: So the idea here is that rather than using the `prefix` argument to `SetFunctionName`, the `name` parameter for such functions should be understood to include the `get` part already. (Since Symbol-named builtins have their name fully specified, they don't need to hit the `Type(_name_) is Symbol` path in `SetFunctionName`.)

Works for me.

bakkot:
@ExE-Boss This needs a rebase again, after which I believe it will be ready to land. Do you want to take care of it? Otherwise the editors will get to it in a week or two.

spec.html
ExE-Boss: The issue there is that engines are allowed to omit the `prefix` from the <code><var>func</var>.[[InitialName]]</code> internal slot.

ljharb: Whether they omit or include it, i think this suggestion is still valid.

bakkot: I am inclined to say that that's not worth worrying about.

spec.html
```diff
@@ -24705,13 +24715,14 @@ <h1>ECMAScript Standard Built-in Objects</h1>
...
+  <p>Every built-in function object, including constructors, has a *"name"* property whose value is a String. Unless otherwise specified, this value is the name that is given to the function in this specification. Functions that are identified as anonymous functions use the empty String as the value of the *"name"* property. For functions that are specified as properties of objects, the name value is the property name string used to access the function. Functions that are specified as get or set accessor functions of built-in properties and whose property keys are not Symbol values have *"get"* or *"set"* passed to the _prefix_ parameter when calling CreateBuiltinFunction.
```
bakkot, line 24724:
```suggestion
  <p>Every built-in function object, including constructors, has a *"name"* property whose value is a String. Unless otherwise specified, this value is the name that is given to the function in this specification. Functions that are identified as anonymous functions use the empty String as the value of the *"name"* property. For functions that are specified as properties of objects, the name value is the property name string used to access the function. Functions that are specified as get or set accessor functions of built-in properties and whose property keys are not Symbol values have *"get"* or *"set"* (respectively) passed to the _prefix_ parameter when calling CreateBuiltinFunction.
```

spec.html
```diff
@@ -24705,13 +24715,14 @@ <h1>ECMAScript Standard Built-in Objects</h1>
...
+  <p>The value of the *"name"* property is explicitly specified for each built-in functions whose property key is a Symbol value. If such an explicitly specified value starts with the prefix *"get "* or *"set "* and the function for which it is specified is a get or set accessor function of a built-in property, the value without the prefix is passed to the _name_ parameter, and the value *"get"* or *"set"* is passed to the _prefix_ parameter when calling CreateBuiltinFunction.</p>
```
bakkot, line 24725:
```suggestion
  <p>The value of the *"name"* property is explicitly specified for each built-in functions whose property key is a Symbol value. If such an explicitly specified value starts with the prefix *"get "* or *"set "* and the function for which it is specified is a get or set accessor function of a built-in property, the value without the prefix is passed to the _name_ parameter, and the value *"get"* or *"set"* (respectively) is passed to the _prefix_ parameter when calling CreateBuiltinFunction.</p>
```

spec.html
```diff
@@ -24705,13 +24715,14 @@ <h1>ECMAScript Standard Built-in Objects</h1>
...
+  <p>Every built-in function object, including constructors, has a *"name"* property whose value is a String. Unless otherwise specified, this value is the name that is given to the function in this specification. Functions that are identified as anonymous functions use the empty String as the value of the *"name"* property. For functions that are specified as properties of objects, the name value is the property name string used to access the function. Functions that are specified as get or set accessor functions of built-in properties and whose property keys are not Symbol values have *"get"* or *"set"* (respectively) passed to the _prefix_ parameter when calling CreateBuiltinFunction.
```
ExE-Boss, line 24724: The name all existing symbol accessor properties include the accessor prefixes in all engines that I’ve tested (**V8** and **SpiderMonkey**):

```suggestion
  <p>Every built-in function object, including constructors, has a *"name"* property whose value is a String. Unless otherwise specified, this value is the name that is given to the function in this specification. Functions that are identified as anonymous functions use the empty String as the value of the *"name"* property. For functions that are specified as properties of objects, the name value is the property name string used to access the function. Functions that are specified as get or set accessor functions of built-in properties have *"get"* or *"set"* (respectively) passed to the _prefix_ parameter when calling CreateBuiltinFunction.
```

ljharb: It’d be good to test (with eshost, perhaps) to confirm what jsc, xs, and engine262 do, but I’d certainly hope that the property key type wouldn’t affect the behavior.

ExE-Boss: @engine262 also [includes the prefixes for accessor properties](https://github.com/engine262/engine262/blob/7c5d6f77dcf54839b446081c09feba3740e42857/src/intrinsics/bootstrap.mjs#L24-L50), regardless of the property key type.

ljharb (approved):

michaelficarra (approved):

PR #2121, opened 2020-08-01
Editorial: reference correct nonterminals in IteratorDestructuringAssignmentEvaluation

bakkot:
The referenced nonterminals don't exist. This was changed in #2030, presumably by accident (cc @jridgewell), and this PR puts it back (while keeping the tweaked wording @michaelficarra preferred).

Thanks to @devsnek for spotting it.


jridgewell:
Ha, I think I replaced all instances of `NamedEvaluation`. This is indeed unintentional.

PR #2122, opened 2020-08-03
Editorial: define Math functions using algorithm steps

### LINKED ISSUE #2119: Define Math functions using algorithm steps
[Function Properties of the Math Object](https://tc39.es/ecma262/#sec-function-properties-of-the-math-object) has a header which says

> Each of the following Math object functions applies the ToNumber abstract operation to each of its arguments (in left-to-right order if there is more than one). If ToNumber returns an abrupt completion, that Completion Record is immediately returned. Otherwise, the function performs a computation on the resulting Number value(s). The value returned by each function is a Number.

And then the actual functions are typically defined not using algorithm steps, but rather with a [prose description](https://tc39.es/ecma262/#sec-math.trunc) such as 

> Returns the integral part of the number x, removing any fractional digits. If x is already an integer, the result is x.

followed by some descriptions of edge-case behavior.

I think it would be clearer if these were defined with prose steps which explicitly called `ToNumber` on each argument, in order, and then handled each edge case in turn, and then had a prose description (or, when possible, precise algorithm) for the usual case.

ryanjduffy:
Starting a few methods to get feedback.

Fixes #2119

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27579,8 +27579,6 @@
 
     <emu-clause id="sec-function-properties-of-the-math-object">
       <h1>Function Properties of the Math Object</h1>
-      <p>Each of the following `Math` object functions applies the ToNumber abstract operation to each of its arguments (in left-to-right order if there is more than one). If ToNumber returns an abrupt completion, that Completion Record is immediately returned. Otherwise, the function performs a computation on the resulting Number value(s). The value returned by each function is a Number.</p>
-      <p>In the function descriptions below, the symbols *NaN*, *-0*, *+0*, *-&infin;* and *+&infin;* refer to the Number values described in <emu-xref href="#sec-ecmascript-language-types-number-type"></emu-xref>.</p>
       <emu-note>
         <p>The behaviour of the functions `acos`, `acosh`, `asin`, `asinh`, `atan`, `atanh`, `atan2`, `cbrt`, `cos`, `cosh`, `exp`, `expm1`, `hypot`, `log`,`log1p`, `log2`, `log10`, `pow`, `random`, `sin`, `sinh`, `sqrt`, `tan`, and `tanh` is not precisely specified here except to require specific results for certain argument values that represent boundary cases of interest. For other argument values, these functions are intended to compute approximations to the results of familiar mathematical functions, but some latitude is allowed in the choice of approximation algorithms. The general intent is that an implementer should be able to use the same mathematical library for ECMAScript on a given hardware platform that is available to C programmers on that platform.</p>
         <p>Although the choice of algorithms is left to the implementation, it is recommended (but not specified by this standard) that implementations use the approximation algorithms for IEEE 754-2019 arithmetic contained in `fdlibm`, the freely distributable mathematical library from Sun Microsystems (<a href="http://www.netlib.org/fdlibm">http://www.netlib.org/fdlibm</a>).</p>
@@ -27589,278 +27587,158 @@
       <emu-clause id="sec-math.abs">
         <h1>Math.abs ( _x_ )</h1>
         <p>Returns the absolute value of _x_; the result has the same magnitude as _x_ but has positive sign.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is *+&infin;*.
-          </li>
-        </ul>
+        <p>When the `Math.abs` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, return *NaN*.
+          1. If _n_ is *-0*, return *+0*.
+          1. If _n_ is *-&infin;*, return *+&infin;*.
+          1. If _n_ &lt; 0, return -_n_.
+          1. Return _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.acos">
         <h1>Math.acos ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the arc cosine of _x_. The result is expressed in radians and ranges from *+0* to +&pi;.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is greater than 1, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is less than -1, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is exactly 1, the result is *+0*.
-          </li>
-        </ul>
+        <p>Returns the inverse cosine of _x_. The result is expressed in radians and ranges from *+0* to +&pi;, inclusive.</p>
+        <p>When the `Math.acos` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ &gt; 1, or _n_ &lt; -1, return *NaN*.
+          1. If _n_ is 1, return *+0*.
+          1. Return an implementation-approximated value representing the result of the inverse cosine of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.acosh">
         <h1>Math.acosh ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the inverse hyperbolic cosine of _x_.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If x is less than 1, the result is *NaN*.
-          </li>
-          <li>
-            If x is 1, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *+&infin;*, the result is *+&infin;*.
-          </li>
-        </ul>
+        <p>Returns the inverse hyperbolic cosine of _x_.</p>
+        <p>When the `Math.acosh` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN* or _n_ is *+&infin;*, return _n_.
+          1. If _n_ is 1, return *+0*.
+          1. If _n_ &lt; 1, return *NaN*.
+          1. Return an implementation-approximated value representing the result of the inverse hyperbolic cosine of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.asin">
         <h1>Math.asin ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the arc sine of _x_. The result is expressed in radians and ranges from -&pi; / 2 to +&pi; / 2.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is greater than 1, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is less than -1, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is *+0*, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is *-0*.
-          </li>
-        </ul>
+        <p>Returns the inverse sine of _x_. The result is expressed in radians and ranges from -&pi; / 2 to +&pi; / 2, inclusive.</p>
+        <p>When the `Math.asin` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
+          1. If _n_ &gt; 1 or _n_ &lt; -1, return *NaN*.
+          1. Return an implementation-approximated value representing the result of the inverse sine of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.asinh">
         <h1>Math.asinh ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the inverse hyperbolic sine of _x_.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is *+0*, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is *-0*.
-          </li>
-          <li>
-            If _x_ is *+&infin;*, the result is *+&infin;*.
-          </li>
-          <li>
-            If x is *-&infin;*, the result is *-&infin;*.
-          </li>
-        </ul>
+        <p>Returns the inverse hyperbolic sine of _x_.</p>
+        <p>When the `Math.asinh` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+          1. Return an implementation-approximated value representing the result of the inverse hyperbolic sine of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.atan">
         <h1>Math.atan ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the arc tangent of _x_. The result is expressed in radians and ranges from -&pi; / 2 to +&pi; / 2.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is *+0*, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is *-0*.
-          </li>
-          <li>
-            If _x_ is *+&infin;*, the result is an implementation-approximated value representing +&pi; / 2.
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is an implementation-approximated value representing to -&pi; / 2.
-          </li>
-        </ul>
+        <p>Returns the inverse tangent of _x_. The result is expressed in radians and ranges from -&pi; / 2 to +&pi; / 2, inclusive.</p>
+        <p>When the `Math.atan` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
+          1. If _n_ is *+&infin;*, return an implementation-approximated value representing +&pi; / 2.
+          1. If _n_ is *-&infin;*, return an implementation-approximated value representing -&pi; / 2.
+          1. Return an implementation-approximated value representing the result of the inverse tangent of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.atanh">
         <h1>Math.atanh ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the inverse hyperbolic tangent of _x_.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is less than -1, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is greater than 1, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is -1, the result is *-&infin;*.
-          </li>
-          <li>
-            If _x_ is +1, the result is *+&infin;*.
-          </li>
-          <li>
-            If _x_ is *+0*, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is *-0*.
-          </li>
-        </ul>
+        <p>Returns the inverse hyperbolic tangent of _x_.</p>
+        <p>When the `Math.atanh` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
+          1. If _n_ &gt; 1 or _n_ &lt; -1, return *NaN*.
+          1. If _n_ is +1, return *+&infin;*.
+          1. If _n_ is -1, return *-&infin;*.
+          1. Return an implementation-approximated value representing the result of the inverse hyperbolic tangent of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.atan2">
         <h1>Math.atan2 ( _y_, _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the arc tangent of the quotient <emu-eqn>_y_ / _x_</emu-eqn> of the arguments _y_ and _x_, where the signs of _y_ and _x_ are used to determine the quadrant of the result. Note that it is intentional and traditional for the two-argument arc tangent function that the argument named _y_ be first and the argument named _x_ be second. The result is expressed in radians and ranges from -&pi; to +&pi;.</p>
-        <ul>
-          <li>
-            If either _x_ or _y_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _y_ &gt; 0 and _x_ is *+0*, the result is an implementation-approximated value representing +&pi; / 2.
-          </li>
-          <li>
-            If _y_ &gt; 0 and _x_ is *-0*, the result is an implementation-approximated value representing +&pi; / 2.
-          </li>
-          <li>
-            If _y_ is *+0* and _x_ &gt; 0, the result is *+0*.
-          </li>
-          <li>
-            If _y_ is *+0* and _x_ is *+0*, the result is *+0*.
-          </li>
-          <li>
-            If _y_ is *+0* and _x_ is *-0*, the result is an implementation-approximated value representing +&pi;.
-          </li>
-          <li>
-            If _y_ is *+0* and _x_ &lt; 0, the result is an implementation-approximated value representing +&pi;.
-          </li>
-          <li>
-            If _y_ is *-0* and _x_ &gt; 0, the result is *-0*.
-          </li>
-          <li>
-            If _y_ is *-0* and _x_ is *+0*, the result is *-0*.
-          </li>
-          <li>
-            If _y_ is *-0* and _x_ is *-0*, the result is an implementation-approximated value representing -&pi;.
-          </li>
-          <li>
-            If _y_ is *-0* and _x_ &lt; 0, the result is an implementation-approximated value representing -&pi;.
-          </li>
-          <li>
-            If _y_ &lt; 0 and _x_ is *+0*, the result is an implementation-approximated value representing -&pi; / 2.
-          </li>
-          <li>
-            If _y_ &lt; 0 and _x_ is *-0*, the result is an implementation-approximated value representing -&pi; / 2.
-          </li>
-          <li>
-            If _y_ &gt; 0 and _y_ is finite and _x_ is *+&infin;*, the result is *+0*.
-          </li>
-          <li>
-            If _y_ &gt; 0 and _y_ is finite and _x_ is *-&infin;*, the result is an implementation-approximated value representing +&pi;.
-          </li>
-          <li>
-            If _y_ &lt; 0 and _y_ is finite and _x_ is *+&infin;*, the result is *-0*.
-          </li>
-          <li>
-            If _y_ &lt; 0 and _y_ is finite and _x_ is *-&infin;*, the result is an implementation-approximated value representing -&pi;.
-          </li>
-          <li>
-            If _y_ is *+&infin;* and _x_ is finite, the result is an implementation-approximated value representing +&pi; / 2.
-          </li>
-          <li>
-            If _y_ is *-&infin;* and _x_ is finite, the result is an implementation-approximated value representing -&pi; / 2.
-          </li>
-          <li>
-            If _y_ is *+&infin;* and _x_ is *+&infin;*, the result is an implementation-approximated value representing +&pi; / 4.
-          </li>
-          <li>
-            If _y_ is *+&infin;* and _x_ is *-&infin;*, the result is an implementation-approximated value representing +3&pi; / 4.
-          </li>
-          <li>
-            If _y_ is *-&infin;* and _x_ is *+&infin;*, the result is an implementation-approximated value representing -&pi; / 4.
-          </li>
-          <li>
-            If _y_ is *-&infin;* and _x_ is *-&infin;*, the result is an implementation-approximated value representing -3&pi; / 4.
-          </li>
-        </ul>
+        <p>Returns the inverse tangent of the quotient <emu-eqn>_y_ / _x_</emu-eqn> of the arguments _y_ and _x_, where the signs of _y_ and _x_ are used to determine the quadrant of the result. Note that it is intentional and traditional for the two-argument inverse tangent function that the argument named _y_ be first and the argument named _x_ be second. The result is expressed in radians and ranges from -&pi; to +&pi;, inclusive.</p>
+        <p>When the `Math.atan2` method is called with arguments _x_ and _y_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _ny_ be ? ToNumber(_y_).
+          1. Let _nx_ be ? ToNumber(_x_).
+          1. If _ny_ is *NaN* or _nx_ is *NaN*, return *NaN*.
+          1. If _ny_ is *+&infin;*, then
+            1. If _nx_ is *+&infin;*, return an implementation-approximated value representing +&pi; / 4.
+            1. If _nx_ is *-&infin;*, return an implementation-approximated value representing +3&pi; / 4.
+            1. Return an implementation-approximated value representing +&pi; / 2.
+          1. If _ny_ is *-&infin;*, then
+            1. If _nx_ is *+&infin;*, return an implementation-approximated value representing -&pi; / 4.
+            1. If _nx_ is *-&infin;*, return an implementation-approximated value representing -3&pi; / 4.
+            1. Return an implementation-approximated value representing -&pi; / 2.
+          1. If _ny_ is *+0*, then
+            1. If _nx_ &gt; 0 or _nx_ is *+0*, return *+0*.
+            1. Return an implementation-approximated value representing +&pi;.
+          1. If _ny_ is *-0*, then
+            1. If _nx_ &gt; 0 or _nx_ is *+0*, return *-0*.
+            1. Return an implementation-approximated value representing -&pi;.
+          1. Assert: _ny_ is finite and is neither *+0* nor *-0*.
+          1. If _ny_ &gt; 0, then
+            1. If _nx_ is *+&infin;*, return *+0*.
+            1. If _nx_ is *-&infin;*, return an implementation-approximated value representing +&pi;.
+            1. If _nx_ is *+0* or _nx_ is *-0*, return an implementation-approximated value representing +&pi; / 2.
+          1. If _ny_ &lt; 0, then
+            1. If _nx_ is *+&infin;*, return *-0*.
+            1. If _nx_ is *-&infin;*, return an implementation-approximated value representing -&pi;.
+            1. If _nx_ is *+0* or _nx_ is *-0*, return an implementation-approximated value representing -&pi; / 2.
+          1. Assert: _nx_ is finite and is neither *+0* nor *-0*.
+          1. Return an implementation-approximated value representing the result of the inverse tangent of the quotient _ny_ / _nx_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.cbrt">
         <h1>Math.cbrt ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the cube root of _x_.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is *+0*, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is *-0*.
-          </li>
-          <li>
-            If _x_ is *+&infin;*, the result is *+&infin;*.
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is *-&infin;*.
-          </li>
-        </ul>
+        <p>Returns the cube root of _x_.</p>
+        <p>When the `Math.cbrt` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+          1. Return an implementation-approximated value representing the result of the cube root of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.ceil">
         <h1>Math.ceil ( _x_ )</h1>
-        <p>Returns the smallest (closest to *-&infin;*) Number value that is not less than _x_ and is an integer. If _x_ is already an integer, the result is _x_.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is *+0*, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is *-0*.
-          </li>
-          <li>
-            If _x_ is *+&infin;*, the result is *+&infin;*.
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is *-&infin;*.
-          </li>
-          <li>
-            If _x_ is less than 0 but greater than -1, the result is *-0*.
-          </li>
-        </ul>
-        <p>The value of `Math.ceil(x)` is the same as the value of `-Math.floor(-x)`.</p>
+        <p>Returns the smallest (closest to *-&infin;*) integral Number value that is not less than _x_. If _x_ is already an integral Number, the result is _x_.</p>
+        <p>When the `Math.ceil` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+          1. If _n_ &lt; 0 and _n_ &gt; -1, return *-0*.
+          1. If _n_ is an integral Number, return _n_.
+          1. Return the smallest (closest to *-&infin;*) integral Number value that is not less than _n_.
+        </emu-alg>
+        <emu-note>
+          <p>The value of `Math.ceil(x)` is the same as the value of `-Math.floor(-x)`.</p>
+        </emu-note>
       </emu-clause>
 
       <emu-clause id="sec-math.clz32">
         <h1>Math.clz32 ( _x_ )</h1>
-        <p>When `Math.clz32` is called with one argument _x_, the following steps are taken:</p>
+        <p>When the `Math.clz32` method is called with argument _x_, the following steps are taken:</p>
         <emu-alg>
           1. Let _n_ be ? ToUint32(_x_).
           1. Let _p_ be the number of leading zero bits in the 32-bit binary representation of _n_.
@@ -27873,46 +27751,26 @@
 
       <emu-clause id="sec-math.cos">
         <h1>Math.cos ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the cosine of _x_. The argument is expressed in radians.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is *+0*, the result is 1.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is 1.
-          </li>
-          <li>
-            If _x_ is *+&infin;*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is *NaN*.
-          </li>
-        </ul>
+        <p>Returns the cosine of _x_. The argument is expressed in radians.</p>
+        <p>When the `Math.cos` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
+          1. If _n_ is *+&infin;* or _n_ is *-&infin;*, return *NaN*.
+          1. Return an implementation-approximated value representing the result of the cosine of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.cosh">
         <h1>Math.cosh ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the hyperbolic cosine of _x_.</p>
-        <ul>
-          <li>
-            If _x_ is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is *+0*, the result is 1.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is 1.
-          </li>
-          <li>
-            If _x_ is *+&infin;*, the result is *+&infin;*.
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is *+&infin;*.
-          </li>
-        </ul>
+        <p>Returns the hyperbolic cosine of _x_.</p>
+        <p>When the `Math.cosh` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+          1. If _n_ is *+0* or _n_ is *-0*, return 1.
+          1. Return an implementation-approximated value representing the result of the hyperbolic cosine of _n_.
+        </emu-alg>
         <emu-note>
           <p>The value of `Math.cosh(x)` is the same as the value of `(Math.exp(x) + Math.exp(-x)) / 2`.</p>
         </emu-note>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -27824,225 +27824,121 @@ <h1>Function Properties of the Math Object</h1>
...
-          <li>
-            If _x_ is *-0*, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is *+&infin;*.
-          </li>
-        </ul>
+        <emu-alg>
+            1. Let _n_ be ? ToNumber(_x_).
+            1. If Number.isNaN(_n_), return *NaN*.
```
ljharb, line 27829: this can't observably call into `Number.isNaN`; I think you may mean:
```suggestion
            1. If _n_ is *NaN*, return *NaN*.
```

spec.html
```diff
@@ -27824,225 +27824,121 @@ <h1>Function Properties of the Math Object</h1>
...
+            1. If _n_ is *-0*, return *+0*.
+            1. If _n_ is *-&infin;*, return *+&infin;*.
+            1. If _n_ < 0, return ((_n_) * (-1)).
```
ljharb, line 27832:
```suggestion
            1. If _n_ < 0, return -_n_.
```

spec.html
```diff
@@ -27824,225 +27824,121 @@ <h1>Function Properties of the Math Object</h1>
...
+            1. Else return _n_.
```
ljharb, line 27833:
```suggestion
            1. Return _n_.
```

spec.html
```diff
@@ -27824,225 +27824,121 @@ <h1>Function Properties of the Math Object</h1>
...
-          <li>
-            If _x_ is less than -1, the result is *NaN*.
-          </li>
-          <li>
-            If _x_ is exactly 1, the result is *+0*.
-          </li>
-        </ul>
+        <emu-alg>
+            1. Let _n_ be ? ToNumber(_x_).
+            1. If Number.isNaN(_n_) or _n_ &gt;1 or _n_ &lt; -1, return *NaN*.
```
ljharb, line 27842:
```suggestion
            1. If _n_ is *NaN*, _n_ &gt; 1, or _n_ &lt; -1, return *NaN*.
```

ryanjduffy:
Pushed a couple commits to apply @ljharb's feedback and generally:

* Separate conditions with commas (If _condition_, or _condition_, or _condition_, then ...`)
* Use `is *NaN*` over `Number.isNaN` and `is finite` over `Number.isFinite`

spec.html
ljharb: (this still has two "or"s)

spec.html
```diff
@@ -27824,225 +27824,121 @@ <h1>Function Properties of the Math Object</h1>
...
-          <li>
-            If _x_ is *+0*, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is *-0*, the result is *-0*.
-          </li>
-        </ul>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, or _n_ &gt;1, or _n_ &lt; -1, return *NaN*.
```
ljharb, line 27865: this also has two "or"s, as do a few others

ljharb: Neither; i'm saying that in any comma-separated list, only the final item should have a conjunction.

gibson042: These could also be collapsed together, although that may not be true in every case.
```suggestion
          1. If _n_ is *NaN* or abs(_n_) &gt; 1, return *NaN*.
```

spec.html
```diff
@@ -28586,29 +28371,14 @@ <h1>Math.tanh ( _x_ )</h1>
...
-          </li>
-          <li>
-            If _x_ is less than 0 but greater than -1, the result is *-0*.
-          </li>
-        </ul>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *-0*, _n_ is *+0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+          1. If _n_ &gt; 0 and _n_ &lt; 1, return *+0*.
+          1. If _n_ &lt; 0 and _n_ &lt;= -1, return *-0*.
```
michaelficarra, line 28378:
```suggestion
          1. If _n_ &lt; 0 and _n_ &gt; -1, return *-0*.
```

spec.html
```diff
@@ -28586,29 +28371,14 @@ <h1>Math.tanh ( _x_ )</h1>
...
+          1. Let _r_ be _n_ modulo 1.
+          1. Return _n_ - _r_.
```
michaelficarra, line 28380: This doesn't work for negative `x`, right? I think you'll have to say something like

```suggestion
          1. If n &gt; 0, return _n_ - _r_; otherwise, return _n_ + (1 - _r_).
```

ryanjduffy: Yep!

michaelficarra, line 28380: Something like this might be even better (due to float maths weirdness):

```suggestion
          1. Return the integral Number nearest _n_ in the direction of *+0*.
```

@bakkot Thoughts?

spec.html
```diff
@@ -28420,29 +28274,19 @@ <h1>Math.random ( )</h1>
...
+          1. If _n_ is *NaN*, _n_ is *-0*, _n_ is *+0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+          1. If _n_ &gt; 0 and _n_ &lt; 0.5, return *+0*.
+          1. If _n_ &lt; 0 and _n_ &lt;= -0.5, return *-0*.
+          1. Let _r_ be _n_ modulo 1.
+          1. Let _v_ be _n_ - _r_.
+          1. If _r_ is 0, return _v_.
+          1. If _r_ &gt;= -0.5 and _r_ &lt; 0, return _v_.
+          1. If _r_ &lt; 0.5 and _r_ &gt; 0, return _v_.
+          1. If _r_ &lt; -0.5, return _v_ - 1.
+          1. Return _v_ + 1.
```
michaelficarra, line 28288: I think we can say this more simply and equally clearly with prose:

```suggestion
          1. If _n_ is an integral Number, return _n_.
          1. Return the integral Number closest to _n_, preferring the Number closer to *+&infin;* in the case of a tie.
```

spec.html
```diff
@@ -28276,130 +28172,88 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          <li>
-            If any value is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            The comparison of values to determine the largest value is done using the Abstract Relational Comparison algorithm except that *+0* is considered to be larger than *-0*.
-          </li>
-        </ul>
+        <emu-alg>
+          1. Let _values_ be a List whose elements are, in left to right order, the arguments that were passed to this function invocation.
+          1. Let _len_ be the number of elements in _values_.
```
michaelficarra, line 28228: This isn't used.

```suggestion
```

spec.html
```diff
@@ -28276,130 +28172,88 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          </li>
-          <li>
-            If any value is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            The comparison of values to determine the largest value is done using the Abstract Relational Comparison algorithm except that *+0* is considered to be larger than *-0*.
-          </li>
-        </ul>
+        <emu-alg>
+          1. Let _values_ be a List whose elements are, in left to right order, the arguments that were passed to this function invocation.
```
michaelficarra, line 28227: I don't know if you need this line given that `values` is declared in the parameter list. @tc39/ecma262-editors?

spec.html
```diff
@@ -28276,130 +28172,88 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
+          1. Let _len_ be the number of elements in _values_.
+          1. Let _v_ be *-&infin;*.
+          1. Repeat, while _values_ is not empty,
+            1. Remove the first element from _values_ and let _value_ be the value of the element.
```
michaelficarra, line 28231: I would just make this two steps.

```suggestion
            1. Let _value_ be the first element of _values_.
            1. Remove the first element from _values_.
```

spec.html
```diff
@@ -28276,130 +28172,88 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-            If any value is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            The comparison of values to determine the largest value is done using the Abstract Relational Comparison algorithm except that *+0* is considered to be larger than *-0*.
-          </li>
-        </ul>
+        <emu-alg>
+          1. Let _values_ be a List whose elements are, in left to right order, the arguments that were passed to this function invocation.
+          1. Let _len_ be the number of elements in _values_.
+          1. Let _v_ be *-&infin;*.
```
michaelficarra, line 28229: I'd probably use a better alias name here. Something like `lowest`.

spec.html
```diff
@@ -28276,130 +28172,88 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
+          1. If _n_ is *+0* or _n_ is *-0*, return *-&infin;*.
+          1. If _n_ is 1, return *+0*.
+          1. If _n_ is *+&infin;*, return *+&infin;*.
+          1. Return an implementation-approximated value representing the result of the base 2 logarithm of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.max">
-        <h1>Math.max ( _value1_, _value2_, ..._values_ )</h1>
+        <h1>Math.max ( ..._values_ )</h1>
```
michaelficarra, line 28224: It may be surprising, but this change is actually normative. `Math.max.length` is defined to be `2` because of the number of mandatory parameters listed here. We can resolve this by explicitly stating that the length of this function is `2` in the description. Or you could change the signature back. Either way works for me.

spec.html
michaelficarra, line 28229: If you end up not changing the signature, you can do something like this:

```suggestion
          1. Let _v_ be *-&infin;*.
          1. If _value1_ &gt; _v_, set _v_ to _value1_.
          1. If _value2_ &gt; _v_, set _v_ to _value2_.
```

spec.html
```diff
@@ -28276,130 +28172,88 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          <li>
-            If _x_ is *+&infin;*, the result is *+&infin;*.
-          </li>
-        </ul>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN* or _n_ &lt; 0, return *NaN*.
+          1. If _n_ is *+0* or _n_ is *-0*, return *-&infin;*.
+          1. If _n_ is 1, return *+0*.
+          1. If _n_ is *+&infin;*, return *+&infin;*.
```
michaelficarra, line 28218: I slightly prefer grouping together the pass-throughs like the first line here:

```suggestion
          1. If _n_ is *NaN* or _n_ is *+&infin;*, return _n_.
          1. If _n_ &lt; 0, return *NaN*.
          1. If _n_ is *+0* or _n_ is *-0*, return *-&infin;*.
          1. If _n_ is 1, return *+0*.
```

spec.html
```diff
@@ -27824,225 +27824,121 @@ <h1>Function Properties of the Math Object</h1>
...
-            If _x_ is *+&infin;*, the result is *+&infin;*.
-          </li>
-          <li>
-            If x is *-&infin;*, the result is *-&infin;*.
-          </li>
-        </ul>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, return *NaN*.
+          1. If _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
```
michaelficarra, line 27877:
```suggestion
          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
```

spec.html
ljharb: agreed; it's redundant with `...values` in the signature.

spec.html
ljharb: I have a strong preference to change the signature back, and to minimize the places in the spec where the length is explicitly overriding the signature.

ryanjduffy:
@michaelficarra Thanks for the feedback. Hoping to get back to this later tonight.

spec.html
jmdyck: > Neither; i'm saying that in any comma-separated list, only the final item should have a conjunction.

It sounds like you're saying to use (e.g.):
`If _n_ is 1, 2, or 3, ...`
rather than:
`If _n_ is 1, or 2, or 3, ...`
I'm fine with that rule (and it pretty much agrees with usage in the spec).

However, it doesn't apply to the current case, which has the structure
`If A, or B, or C, ...`
where each of A/B/C is a separate 'condition'. Currently, for such cases, the spec is somewhat split on whether to use any commas, but it's quite consistent that all the `or`s should appear.

(It's less clear when the conjunction is `and`:
`If A and B and C, ...`
is about as common as
`If A, B, and C, ...`
)

ljharb: I wouldn't expect any differentiation between values or conditions; I'd expect what I'm used to in English sentences - that a comma-separated clause has only a final conjunction (and an Oxford comma, ofc).

jmdyck: Well, then your expectation is at odds with current usage in the spec.

One downside of that style (in English and in the spec) is that you have to read most of the condition before you know whether the elided conjunction is "and" or "or". Which is not a big problem if the condition is short, but the spec has some long ones.

spec.html
ryanjduffy: I borrowed this from `Array.prototype.push` and `Array.prototype.splice` which I only point out so they can also be updated separately if that's the desire. I'll drop this here.

spec.html
ryanjduffy: This format is also used in `Array` methods (e.g. `Array.prototype.concat`) which is from where I borrowed it.

spec.html
ryanjduffy: That's fine. It was unclear to me to how close these should be to implementation algorithms so I probably went too far on this one.

spec.html
ryanjduffy: I went with this since it aligns with the changes to `Math.round()`.

spec.html
michaelficarra: Thanks, I've sent PR https://github.com/tc39/ecma262/pull/2137 to address those cases.

michaelficarra:
@ryanjduffy Is this ready for another review?

ryanjduffy:
Finished converting the methods but haven't reviewed the code yet. Reviews are still welcome but I won't move this out of draft until I've had a chance to review in detail.

cc: @ljharb, @michaelficarra

ryanjduffy:
@michaelficarra - just saw your note. Yes, I can do that.

spec.html
ryanjduffy: Agreed!

spec.html
ryanjduffy: In this PR, I've used the following format but I can adjust as needed:

```
1. If condition1, then ...
1. If condition1 or condition2, then ...
1. If condition1, condition2, or condition3, then ...
```

spec.html
michaelficarra: We'll discuss this in https://github.com/tc39/ecma262/pull/2139.

michaelficarra:
@ryanjduffy We should remove these paragraphs at the top of 20.3.2 as part of this PR:

image

spec.html
```diff
@@ -28130,273 +28130,144 @@ <h1>Function Properties of the Math Object</h1>
...
-          <li>
-            If _x_ is less than 0 but greater than -1, the result is *-0*.
-          </li>
-        </ul>
-        <p>The value of `Math.ceil(x)` is the same as the value of `-Math.floor(-x)`.</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+          1. If _n_ &lt; 0 and _n_ &gt; -1, return *-0*.
+          1. If _n_ is an integer, return _n_.
```
michaelficarra, line 28265: We try to use the term "integral Number" for Number values that represent a mathematical integer.

```suggestion
          1. If _n_ is an integral Number, return _n_.
```

spec.html
```diff
@@ -28130,273 +28130,144 @@ <h1>Function Properties of the Math Object</h1>
...
+          1. Return the smallest (closest to *-&infin;*) Number value that is not less than _n_ and is an integer.
```
michaelficarra, line 28266:
```suggestion
          1. Return the smallest (closest to *-&infin;*) integral Number value that is not less than _n_.
```

ryanjduffy:
Updated "integer" to "integral Number" in `floor()` and `ceil()` and removed those two paragraphs.

spec.html
```diff
@@ -28546,23 +28359,27 @@ <h1>Math.fround ( _x_ )</h1>
...
+            1. Let _value_ be the first element of _values_.
+            1. Remove the first element from _values_.
+            1. Let _v_ be ? ToNumber(_value_).
+            1. If _v_ is *+&infin;*, return *+&infin;*.
+            1. If _v_ is *-&infin;*, return *-&infin;*.
+            1. If _v_ is *NaN*, set _hasNaN_ to *true*.
+            1. If _v_ is not *+0* or _v_ is not *-0*, set _onlyZero_ to be *false*.
+          1. If _hasNaN_ is *true*, return *NaN*.
+          1. If _onlyZero_ is *true*, return *+0*.
+          1. Return an implementation-approximated value representing the result of the square root of the sum of squares of its arguments.
```
michaelficarra, line 28381: I don't think this step is sufficient. I think you should construct a List of all the `ToNumber` results like

```
1. Let _numbers_ be a List containing _v1_ and _v2_.
...
1. Repeat, while _values_ is not empty,
  ...
  1. Append _v_ to _numbers_.
```

and then say here something like

```
1. Return an implementation-approximated value representing the square root of the sum of squares of the elements of _numbers_.
```

Does that make sense?

spec.html
```diff
@@ -28546,23 +28359,27 @@ <h1>Math.fround ( _x_ )</h1>
...
-        </ul>
+        <emu-alg>
+          1. Let _v1_ be ? ToNumber(_value1_).
+          1. Let _v2_ be ? ToNumber(_value2_).
+          1. If _v1_ is *NaN* or _v2_ is *NaN*, return *NaN*.
+          1. If _v1_ is *+&infin;* or _v2_ is *+&infin;*, return *+&infin;*.
+          1. If _v1_ is *-&infin;* or _v2_ is *-&infin;*, return *-&infin;*.
+          1. Let _hasNaN_ be *false*.
+          1. Let _onlyZero_ be *true*.
+          1. If _v1_ is not 0 and _v2_ is not 0, set _onlyZero_ to *false*.
```
michaelficarra, line 28370: I think we can skip all of this duplication by constructing a list like

```
1. Let _args_ be a List containing _value1_, _value2_, and the elements of _values_ in order.
```

and then just looping over `_args_` below instead of `_values_`.

spec.html
```diff
@@ -28582,130 +28399,89 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          <li>
-            If any value is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            The comparison of values to determine the largest value is done using the Abstract Relational Comparison algorithm except that *+0* is considered to be larger than *-0*.
-          </li>
-        </ul>
+        <emu-alg>
+          1. Let _highest_ be *-&infin;*.
+          1. If _value1_ &gt; _highest_, set _highest_ to _value1_.
```
michaelficarra, line 28455: We're forgetting to call `ToNumber` on `_value1_` and `_value2_` here. I'm thinking it might be easiest to prepend them to `_values_` and re-use the loop steps below, as I suggested for `Math.hypot`.

spec.html
```diff
@@ -28582,130 +28399,89 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          <li>
-            If any value is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            The comparison of values to determine the smallest value is done using the Abstract Relational Comparison algorithm except that *+0* is considered to be larger than *-0*.
-          </li>
-        </ul>
+        <emu-alg>
+          1. Let _lowest_ be *+&infin;*.
+          1. If _value1_ &lt; _lowest_, set _lowest_ to _value1_.
```
michaelficarra, line 28474: Same comment as `Math.max`.

spec.html
ryanjduffy: I took a slightly different approach that suggested but let me know if it doesn't work or follow convention.

```
        <emu-alg>
          1. Let _numbers_ be a List containing _value1_ and _value2_, and the elements of _values_ in order.
          1. Let _onlyZero_ be *true*.
          1. Repeat, while _numbers_ is not empty,
            1. Let _number_ be the first element of _numbers_.
            1. Remove the first element from _numbers_.
            1. Let _n_ be ? ToNumber(_number_).
            1. If _n_ is *NaN*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
            1. If _n_ is not *+0* or _n_ is not *-0*, set _onlyZero_ to be *false*.
          1. If _onlyZero_ is *true*, return *+0*.
          1. Return an implementation-approximated value representing the result of the square root of the sum of squares of the elements of _numbers_.
        </emu-alg>
```

I used the similar `List` construction sentence proposed below of all the values. Since NaN and +/- infinity all cause a bail out, I thought it was easier to grok by iterating over the original list and returning the exception cases.

ryanjduffy:
I think I've addressed all the feedback so I've moved this out of draft. Thanks for the help getting up to speed with the conventions, @michaelficarra  and @ljharb!

spec.html
```diff
@@ -28360,25 +28360,19 @@ <h1>Math.fround ( _x_ )</h1>
...
+            1. Let _number_ be the first element of _numbers_.
+            1. Remove the first element from _numbers_.
+            1. Let _n_ be ? ToNumber(_number_).
+            1. If _n_ is *+&infin;* or _n_ is *-&infin;*, return _n_.
+            1. If _n_ is *NaN*, set _hasNaN_ to *true*.
+            1. If _n_ is not *+0* or _n_ is not *-0*, set _onlyZero_ to be *false*.
           1. If _hasNaN_ is *true*, return *NaN*.
           1. If _onlyZero_ is *true*, return *+0*.
-          1. Return an implementation-approximated value representing the result of the square root of the sum of squares of its arguments.
+          1. Return an implementation-approximated value representing the result of the square root of the sum of squares of the elements of _numbers_.
```
michaelficarra: Nit:

```suggestion
          1. Return an implementation-approximated value representing the square root of the sum of squares of the elements of _numbers_.
```

Same in the summary.

spec.html
```diff
@@ -28346,7 +28361,7 @@ <h1>Math.floor ( _x_ )</h1>
 
       <emu-clause id="sec-math.fround">
         <h1>Math.fround ( _x_ )</h1>
-        <p>When `Math.fround` is called with argument _x_, the following steps are taken:</p>
+        <p>When the `Math.fround` method is called with argument _x_, the following steps are taken:</p>
```
michaelficarra, line 28364: Every other unary function's summary now says "one argument", so let's keep it consistent.

```suggestion
        <p>When the `Math.fround` method is called with one argument _x_, the following steps are taken:</p>
```

spec.html
```diff
@@ -28132,278 +28130,159 @@ <h1>Function Properties of the Math Object</h1>
...
-          <li>
-            If _x_ is exactly 1, the result is *+0*.
-          </li>
-        </ul>
+        <p>When the `Math.acos` method is called with one argument _x_, the following steps are taken:</p>
+        <emu-alg>
+            1. Let _n_ be ? ToNumber(_x_).
+            1. If _n_ is *NaN*, _n_ &gt; 1, or _n_ &lt; -1, return *NaN*.
+            1. If _n_ is 1, return *+0*.
+            1. Return an implementation-approximated value representing the result of the arc cosine of _n_.
```
ljharb, line 28152:
```suggestion
            1. Return an implementation-approximated value representing the result of the inverse cosine of _n_.
```

spec.html
```diff
@@ -28132,278 +28130,159 @@ <h1>Function Properties of the Math Object</h1>
...
-          <li>
-            If _x_ is *-0*, the result is *-0*.
-          </li>
-        </ul>
+        <p>When the `Math.asin` method is called with one argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
+          1. If _n_ &gt; 1 or _n_ &lt; -1, return *NaN*.
+          1. Return an implementation-approximated value representing the result of the arc sine of _n_.
```
ljharb, line 28177:
```suggestion
          1. Return an implementation-approximated value representing the result of the inverse sine of _n_.
```

spec.html
```diff
@@ -28132,278 +28130,159 @@ <h1>Function Properties of the Math Object</h1>
...
-            If _x_ is *-&infin;*, the result is an implementation-approximated value representing to -&pi; / 2.
-          </li>
-        </ul>
+        <p>When the `Math.atan` method is called with one argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
+          1. If _n_ is *+&infin;*, return an implementation-approximated value representing +&pi; / 2.
+          1. If _n_ is *-&infin;*, return an implementation-approximated value representing -&pi; / 2.
+          1. Return an implementation-approximated value representing the result of the arc tangent of _n_.
```
ljharb, line 28201:
```suggestion
          1. Return an implementation-approximated value representing the result of the inverse tangent of _n_.
```

spec.html
```diff
@@ -28132,278 +28130,159 @@ <h1>Function Properties of the Math Object</h1>
...
+          1. If _nx_ is finite, then
+            1. If _ny_ is *+&infin;*, return an implementation-approximated value representing +&pi; / 2.
+            1. If _ny_ is *-&infin;*, return an implementation-approximated value representing -&pi; / 2.
+          1. If _ny_ is *+&infin;*, then
+            1. If _nx_ is *+&infin;*, return an implementation-approximated value representing +&pi; / 4.
+            1. If _nx_ is *-&infin;*, return an implementation-approximated value representing +3&pi; / 4.
+          1. If _ny_ is *-&infin;*, then
+            1. If _nx_ is *+&infin;*, return an implementation-approximated value representing -&pi; / 4.
+            1. If _nx_ is *-&infin;*, return an implementation-approximated value representing -3&pi; / 4.
+          1. Return an implementation-approximated value representing the result of the arc tangent of the quotient _ny_ / _nx_.
```
ljharb, line 28252:
```suggestion
          1. Return an implementation-approximated value representing the result of the inverse tangent of the quotient _ny_ / _nx_.
```

spec.html
```diff
@@ -28464,78 +28323,47 @@ <h1>Math.cosh ( _x_ )</h1>
...
+        </emu-alg>
         <emu-note>
           <p>The value of `Math.floor(x)` is the same as the value of `-Math.ceil(-x)`.</p>
         </emu-note>
       </emu-clause>
 
       <emu-clause id="sec-math.fround">
         <h1>Math.fround ( _x_ )</h1>
-        <p>When `Math.fround` is called with argument _x_, the following steps are taken:</p>
+        <p>When the `Math.fround` method is called with one argument _x_, the following steps are taken:</p>
```
ljharb, line 28366:
```suggestion
        <p>When the `Math.fround` method is called with one argument _x_, the following steps are taken:</p>
```
or perhaps, remove "one" from all these preambles? "is called with one" appears 36 times in the spec, and "is called with arg" appears 113 times, so it seems like it'd be more consistent to go with the more common form and make them all "is called with argument".

(Additionally, by specifying the number of arguments there's a mild implication that if it's called with zero or 2+ args, different steps are taken, which isn't the case)

spec.html
```diff
@@ -28728,29 +28517,12 @@ <h1>Math.random ( )</h1>
...
-          </li>
-          <li>
-            If _x_ is less than 0 but greater than or equal to -0.5, the result is *-0*.
-          </li>
-        </ul>
+        <p>When the `Math.round` method is called with one argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is an integral Number, return _n_.
+          1. Return the integral Number closest to _n_, preferring the Number closer to *+&infin;* in the case of a tie.
```
ljharb, line 28524: it seems important not to lose the explicit mention of `0.5` here as part of the rounding method.

michaelficarra: Do you have a concrete suggestion? I personally think the current text is sufficient.

ljharb:
```suggestion
          1. If _x_ &gt; 0 and _x_ &lt; 0.5, return *+0*.
          1. If _x_ &lt; 0 and _x_ &ge; -0.5, return *-0*.
          1. Return the integral Number closest to _n_, preferring the Number closer to *+&infin;* in the case of a tie.
```

michaelficarra: Either is fine by me.

spec.html
ljharb: Yes, I would expect and prefer that all these explicit mentions of a count followed by a short list where the reader can easily count for themselves should remove the count :-)

spec.html
ljharb: I think that this PR needs to be overall consistent with either using "arc" or "inverse"; currently it uses both.

spec.html
```diff
@@ -28132,278 +28130,159 @@ <h1>Function Properties of the Math Object</h1>
...
+            1. If _n_ is *-0*, return *+0*.
+            1. If _n_ is *-&infin;*, return *+&infin;*.
+            1. If _n_ < 0, return -_n_.
+            1. Return _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.acos">
         <h1>Math.acos ( _x_ )</h1>
         <p>Returns an implementation-approximated value representing the result of the arc cosine of _x_. The result is expressed in radians and ranges from *+0* to +&pi;.</p>
```
syg, line 28146: If we're unifying on "inverse" instead of "arc" here and below, this paragraph should change too.

spec.html
```diff
@@ -28132,278 +28130,159 @@ <h1>Function Properties of the Math Object</h1>
...
+          1. If _n_ is *NaN* or _n_ is *+&infin;*, return _n_.
+          1. If _n_ &lt; 1, return *NaN*.
+          1. If _n_ is 1, return *+0*.
+          1. Return an implementation-approximated value representing the result of the inverse hyperbolic cosine of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.asin">
         <h1>Math.asin ( _x_ )</h1>
         <p>Returns an implementation-approximated value representing the result of the arc sine of _x_. The result is expressed in radians and ranges from -&pi; / 2 to +&pi; / 2.</p>
```
syg, line 28171: s/arc/inverse

spec.html
```diff
@@ -28132,278 +28130,159 @@ <h1>Function Properties of the Math Object</h1>
...
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+          1. Return an implementation-approximated value representing the result of the inverse hyperbolic sine of _n_.
+        </emu-alg>
       </emu-clause>
 
       <emu-clause id="sec-math.atan">
         <h1>Math.atan ( _x_ )</h1>
         <p>Returns an implementation-approximated value representing the result of the arc tangent of _x_. The result is expressed in radians and ranges from -&pi; / 2 to +&pi; / 2.</p>
```
syg, line 28194: s/arc/inverse

ryanjduffy:
@syg, @michaelficarra  - I simplified the summaries (in 5a29038) in cases where it seemed redundant to the steps. I think it reads better but would appreciate your thoughts.

spec.html
```diff
@@ -28168,7 +28168,7 @@ <h1>Math.acosh ( _x_ )</h1>
 
       <emu-clause id="sec-math.asin">
         <h1>Math.asin ( _x_ )</h1>
-        <p>Returns an implementation-approximated value representing the result of the arc sine of _x_. The result is expressed in radians and ranges from -&pi; / 2 to +&pi; / 2.</p>
+        <p>Returns the inverse sine of _x_. The argument is expressed in radians and ranges from -&pi; / 2 to +&pi; / 2.</p>
```
michaelficarra, line 28171: You changed a lot of these to "the argument", but I think they're supposed to be talking about the result.

ryanjduffy: My mistake! Got a little carried away in the standardization. :/

ryanjduffy: Fixed this up!

spec.html
```diff
@@ -28375,7 +28375,7 @@ <h1>Math.fround ( _x_ )</h1>
 
       <emu-clause id="sec-math.hypot">
         <h1>Math.hypot ( _value1_, _value2_, ..._values_ )</h1>
-        <p>`Math.hypot` returns an implementation-approximated value representing the square root of the sum of squares of its arguments.</p>
+        <p>`Math.hypot` returns the square root of the sum of squares of its arguments.</p>
```
michaelficarra, line 28378: Nit: this summary stands out in that it mentions the method name again at the beginning. The others just start with "Returns ...".

ryanjduffy: Fixed!

michaelficarra:
@ryanjduffy FYI, I pushed a commit to align with #2152.

spec.html
```diff
@@ -28538,135 +28360,96 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          <li>
-            If any value is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            The comparison of values to determine the largest value is done using the Abstract Relational Comparison algorithm except that *+0* is considered to be larger than *-0*.
-          </li>
-        </ul>
+        <p>When the `Math.max` method is called with at least two arguments _value1_ and _value2_ and any number of additional arguments which form the rest parameter ..._values_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _numbers_ be a List containing _value1_, _value2_, and the elements of _values_ in order.
```
ljharb, line 28421:
```suggestion
          1. Let _numbers_ be a List containing _value1_, _value2_, and the elements of _values_ in List order.
```

spec.html
```diff
@@ -28538,135 +28360,96 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          <li>
-            If any value is *NaN*, the result is *NaN*.
-          </li>
-          <li>
-            The comparison of values to determine the smallest value is done using the Abstract Relational Comparison algorithm except that *+0* is considered to be larger than *-0*.
-          </li>
-        </ul>
+        <p>When the `Math.min` method is called with at least two arguments _value1_ and _value2_ and any number of additional arguments which form the rest parameter ..._values_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _numbers_ be a List containing _value1_, _value2_, and the elements of _values_ in order.
```
ljharb, line 28438:
```suggestion
          1. Let _numbers_ be a List containing _value1_, _value2_, and the elements of _values_ in List order.
```

ljharb (approved):

ryanjduffy:
@ljharb - Shall I make some updates based on your feedback or do you plan to take care of them since you self-assigned?

ljharb:
@ryanjduffy none of the suggestions are blockers; feel free to make them yourself :-) i self-assigned since i plan to merge this once enough editors have stamped.

ryanjduffy:
Updated!

spec.html
```diff
@@ -28371,126 +28250,75 @@ <h1>Math.clz32 ( _x_ )</h1>
...
-          <li>
-            If _x_ is *-&infin;*, the result is *+0*.
-          </li>
-        </ul>
+        <p>Returns the exponential function of _x_ (_e_ raised to the power of _x_, where _e_ is the base of the natural logarithms).</p>
+        <p>When the `Math.exp` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN* or _n_ is *+&infin;*, return _n_.
+          1. If _n_ is *-0* or _n_ is *+0*, return 1.
```
bakkot, line 28285:
```suggestion
          1. If _n_ is *+0* or _n_ is *-0*, return 1.
```

spec.html
```diff
@@ -28371,126 +28250,75 @@ <h1>Math.clz32 ( _x_ )</h1>
...
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is -1.
-          </li>
-        </ul>
+        <p>Returns the result of subtracting 1 from the exponential function of _x_ (_e_ raised to the power of _x_, where _e_ is the base of the natural logarithms). The result is computed in a way that is accurate even when the value of x is close 0.</p>
+        <p>When the `Math.expm1` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *-0*, _n_ is *+0*, or _n_ is *+&infin;*, return _n_.
```
bakkot, line 28297:
```suggestion
          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, or _n_ is *+&infin;*, return _n_.
```

spec.html
```diff
@@ -28502,24 +28330,18 @@ <h1>Math.fround ( _x_ )</h1>
...
-        </ul>
+        <p>Returns the square root of the sum of squares of its arguments.</p>
+        <p>When the `Math.hypot` method is called with at least two arguments _value1_ and _value2_ and any number of additional arguments which form the rest parameter ..._values_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _numbers_ be a List containing _value1_, _value2_, and the elements of _values_ in List order.
+          1. Let _onlyZero_ be *true*.
+          1. For each element _number_ of _numbers_, do
+            1. Let _n_ be ? ToNumber(_number_).
+            1. If _n_ is *NaN*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+            1. If _n_ is not *+0* or _n_ is not *-0*, set _onlyZero_ to be *false*.
```
bakkot, line 28341:
```suggestion
            1. If _n_ is neither *+0* nor *-0*, set _onlyZero_ to *false*.
```

spec.html
```diff
@@ -28538,135 +28360,96 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-            If _x_ is *+&infin;*, the result is *+&infin;*.
-          </li>
-        </ul>
+        <p>Returns the natural logarithm of 1 + _x_. The result is computed in a way that is accurate even when the value of x is close to zero.</p>
+        <p>When the `Math.log1p` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, or _n_ is *+&infin;*, return _n_.
+          1. If _n_ &lt; -1, return *NaN*.
+          1. If _n_ is -1, return *-&infin;*.
```
bakkot, line 28383: Better to go in descending order:
```suggestion
          1. If _n_ is -1, return *-&infin;*.
          1. If _n_ &lt; -1, return *NaN*.
```

spec.html
```diff
@@ -28538,135 +28360,96 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          </li>
-        </ul>
+        <p>Returns the natural logarithm of _x_.</p>
+        <p>When the `Math.log` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN* or _n_ is *+&infin;*, return _n_.
+          1. If _n_ &lt; 0, return *NaN*.
+          1. If _n_ is *+0* or _n_ is *-0*, return *-&infin;*.
+          1. If _n_ is 1, return *+0*.
```
bakkot, line 28370: Better to go in descending order:
```suggestion
          1. If _n_ is 1, return *+0*.
          1. If _n_ is *+0* or _n_ is *-0*, return *-&infin;*.
          1. If _n_ &lt; 0, return *NaN*.
```

spec.html
```diff
@@ -28538,135 +28360,96 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          </li>
-        </ul>
+        <p>Returns the base 10 logarithm of _x_.</p>
+        <p>When the `Math.log10` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN* or _n_ is *+&infin;*, return _n_.
+          1. If _n_ &lt; 0, return *NaN*.
+          1. If _n_ is *+0* or _n_ is *-0*, return *-&infin;*.
+          1. If _n_ is 1, return *+0*.
```
bakkot, line 28397:
```suggestion
          1. If _n_ is 1, return *+0*.
          1. If _n_ is *+0* or _n_ is *-0*, return *-&infin;*.
          1. If _n_ &lt; 0, return *NaN*.
```

spec.html
```diff
@@ -28538,135 +28360,96 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-          </li>
-        </ul>
+        <p>Returns the base 2 logarithm of _x_.</p>
+        <p>When the `Math.log2` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN* or _n_ is *+&infin;*, return _n_.
+          1. If _n_ &lt; 0, return *NaN*.
+          1. If _n_ is *+0* or _n_ is *-0*, return *-&infin;*.
+          1. If _n_ is 1, return *+0*.
```
bakkot, line 28411:
```suggestion
          1. If _n_ is 1, return *+0*.
          1. If _n_ is *+0* or _n_ is *-0*, return *-&infin;*.
          1. If _n_ &lt; 0, return *NaN*.
```

spec.html
```diff
@@ -28538,135 +28360,96 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-            The comparison of values to determine the largest value is done using the Abstract Relational Comparison algorithm except that *+0* is considered to be larger than *-0*.
-          </li>
-        </ul>
+        <p>When the `Math.max` method is called with at least two arguments _value1_ and _value2_ and any number of additional arguments which form the rest parameter ..._values_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _numbers_ be a List containing _value1_, _value2_, and the elements of _values_ in List order.
+          1. Let _highest_ be *-&infin;*.
+          1. For each element _number_ of _numbers_, do
+            1. Let _n_ be ? ToNumber(_number_).
+            1. If _n_ &gt; _highest_, set _highest_ to _n_.
```
bakkot, line 28425: We need to explicitly treat `NaN`, `+0`, and `-0` here, I think; it does not make sense to ask if `NaN > 1`, and we have to ensure that `+0` is larger than `-0`.

spec.html
```diff
@@ -28538,135 +28360,96 @@ <h1>Math.imul ( _x_, _y_ )</h1>
...
-            The comparison of values to determine the smallest value is done using the Abstract Relational Comparison algorithm except that *+0* is considered to be larger than *-0*.
-          </li>
-        </ul>
+        <p>When the `Math.min` method is called with at least two arguments _value1_ and _value2_ and any number of additional arguments which form the rest parameter ..._values_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _numbers_ be a List containing _value1_, _value2_, and the elements of _values_ in List order.
+          1. Let _lowest_ be *+&infin;*.
+          1. For each element _number_ of _numbers_, do
+            1. Let _n_ be ? ToNumber(_number_).
+            1. If _n_ &lt; _lowest_, set _lowest_ to _n_.
```
bakkot, line 28442: As above, we need to explicitly treat `NaN`, `+0`, and `-0` here.

spec.html
```diff
@@ -28717,130 +28485,76 @@ <h1>Math.round ( _x_ )</h1>
...
-            If _x_ is negative and not *-0*, the result is -1.
-          </li>
-          <li>
-            If _x_ is positive and not *+0*, the result is +1.
-          </li>
-        </ul>
+        <p>When the `Math.sign` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *-0*, or _n_ is *+0*, return _n_.
```
bakkot, line 28491:
```suggestion
          1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
```

spec.html
```diff
@@ -28717,130 +28485,76 @@ <h1>Math.round ( _x_ )</h1>
...
-          </li>
-          <li>
-            If _x_ is *+&infin;* or *-&infin;*, the result is *NaN*.
-          </li>
-        </ul>
+        <p>Returns the sine of _x_. The argument is expressed in radians.</p>
+        <p>When the `Math.sin` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+            1. Let _n_ be ? ToNumber(_x_).
+            1. If _n_ is *NaN*, _n_ is *-0*, or _n_ is *+0*, return _n_.
```
bakkot, line 28503:
```suggestion
            1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
```

spec.html
```diff
@@ -28717,130 +28485,76 @@ <h1>Math.round ( _x_ )</h1>
...
+            1. If _n_ is *-&infin;* or _n_ is *+&infin;*, return *NaN*.
```
bakkot, line 28504:
```suggestion
            1. If _n_ is *+&infin;* or _n_ is *-&infin;*, return *NaN*.
```

spec.html
```diff
@@ -28717,130 +28485,76 @@ <h1>Math.round ( _x_ )</h1>
...
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is *-&infin;*.
-          </li>
-        </ul>
+        <p>Returns the hyperbolic sine of _x_.</p>
+        <p>When the `Math.sinh` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+            1. Let _n_ be ? ToNumber(_x_).
+            1. If _n_ is *NaN*, _n_ is *-0*, _n_ is *+0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
```
bakkot, line 28515:
```suggestion
            1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
```

spec.html
```diff
@@ -28717,130 +28485,76 @@ <h1>Math.round ( _x_ )</h1>
...
-          </li>
-          <li>
-            If _x_ is *+&infin;*, the result is *+&infin;*.
-          </li>
-        </ul>
+        <p>Returns the square root of _x_.</p>
+        <p>When the `Math.sqrt` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+            1. Let _n_ be ? ToNumber(_x_).
+            1. If _n_ is *NaN*, _n_ is *-0*, _n_ is *+0*, or _n_ is *+&infin;*, return _n_.
```
bakkot, line 28529:
```suggestion
            1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, or _n_ is *+&infin;*, return _n_.
```

spec.html
```diff
@@ -28717,130 +28485,76 @@ <h1>Math.round ( _x_ )</h1>
...
-          </li>
-          <li>
-            If _x_ is *+&infin;* or *-&infin;*, the result is *NaN*.
-          </li>
-        </ul>
+        <p>Returns the tangent of _x_. The argument is expressed in radians.</p>
+        <p>When the `Math.tan` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+            1. Let _n_ be ? ToNumber(_x_).
+            1. If _n_ is *NaN*, _n_ is *-0*, or _n_ is *+0*, return _n_.
```
bakkot, line 28541:
```suggestion
            1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
```

spec.html
```diff
@@ -28717,130 +28485,76 @@ <h1>Math.round ( _x_ )</h1>
...
+            1. If _n_ is *-&infin;*, or _n_ is *+&infin;*, return *NaN*.
```
bakkot, line 28542:
```suggestion
            1. If _n_ is *+&infin;* or _n_ is *-&infin;*, return *NaN*.
```

(note also the stray comma)

spec.html
```diff
@@ -28717,130 +28485,76 @@ <h1>Math.round ( _x_ )</h1>
...
-          </li>
-          <li>
-            If _x_ is *-&infin;*, the result is -1.
-          </li>
-        </ul>
+        <p>Returns the hyperbolic tangent of _x_.</p>
+        <p>When the `Math.tanh` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+            1. Let _n_ be ? ToNumber(_x_).
+            1. If _n_ is *NaN*, _n_ is *-0*, or _n_ is *+0*, return _n_.
```
bakkot, line 28553:
```suggestion
            1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
```

spec.html
```diff
@@ -28849,29 +28563,14 @@ <h1>Math.tanh ( _x_ )</h1>
...
-            If _x_ is greater than 0 but less than 1, the result is *+0*.
-          </li>
-          <li>
-            If _x_ is less than 0 but greater than -1, the result is *-0*.
-          </li>
-        </ul>
+        <p>When the `Math.trunc` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *-0*, _n_ is *+0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
```
bakkot, line 28569:
```suggestion
          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
```

bakkot (commented):
Looks pretty good. I had a few small comments.

spec.html
michaelficarra: Nice catch!

spec.html
```diff
@@ -28087,278 +28085,159 @@ <h1>Function Properties of the Math Object</h1>
...
-            If _y_ is *-&infin;* and _x_ is *+&infin;*, the result is an implementation-approximated value representing -&pi; / 4.
-          </li>
-          <li>
-            If _y_ is *-&infin;* and _x_ is *-&infin;*, the result is an implementation-approximated value representing -3&pi; / 4.
-          </li>
-        </ul>
+        <p>Returns the inverse tangent of the quotient <emu-eqn>_y_ / _x_</emu-eqn> of the arguments _y_ and _x_, where the signs of _y_ and _x_ are used to determine the quadrant of the result. Note that it is intentional and traditional for the two-argument inverse tangent function that the argument named _y_ be first and the argument named _x_ be second. The result is expressed in radians and ranges from -&pi; to +&pi;, inclusive.</p>
+        <p>When the `Math.atan2` method is called with arguments _x_ and _y_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _ny_ be ? ToNumber(_y_).
```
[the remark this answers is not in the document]
ryanjduffy: Thanks! Agree this is more readable.

spec.html
ryanjduffy: I had organized the steps to list all "identity" cases, followed by exceptional cases (e.g. inputs which return NaN or that have specific return values) and then the "regular" cases. 

How do you want to handle cases like `Math.atanh`?

```
          1. Let _n_ be ? ToNumber(_x_).
          1. If _n_ is *NaN*, _n_ is *+0*, or _n_ is *-0*, return _n_.
          1. If _n_ &gt; 1 or _n_ &lt; -1, return *NaN*.
          1. If _n_ is +1, return *+&infin;*.
          1. If _n_ is -1, return *-&infin;*.
          1. Return an implementation-approximated value representing the result of the inverse hyperbolic tangent of _n_.
```

To strictly follow your guidance, I could split the third step so that the order is > 1, = 1, = -1, < -1. Alternatively this could remain as is and we treat multiple conditions as special cases that precede the descending order. Probably other options too. What do you think, @bakkot?

spec.html
ryanjduffy: How does this look? Would it be better to break the second link into two separate conditions or somehow group the two conditions joined with "and" for better clarity?

```
            1. If _n_ is *NaN*, return *NaN*.
            1. If _n_ &gt; _highest_ or _n_ is +0 and _highest_ is *-0*, set _highest_ to _n_.
```

ryanjduffy: Maybe this is a bit better

```
            1. If _n_ is *NaN*, return *NaN*.
            1. If _n_ is +0 and _highest_ is *-0* or _n_ &gt; _highest_, set _highest_ to _n_.
```

ryanjduffy:
Thanks for the feedback, @bakkot. Had a couple questions I shared above but I pushed a batch changes to keep things moving.

spec.html
michaelficarra: @bakkot See https://github.com/tc39/ecma262/pull/2122#discussion_r465876073.

spec.html
michaelficarra: @ryanjduffy Please don't combine things like that. The precedence of and/or in English is ambiguous.

ryanjduffy: :+1: That was my instinct as well. I'll separate them out.

spec.html
bakkot: > How do you want to handle cases like `Math.atanh`?

In cases where there's no natural order because some conditions are combined this preference wouldn't apply. I only intended this comment to apply to the `log` methods, and only for these three cases (1, 0, <0).

@michaelficarra I saw that comment, but I don't understand the relevance here.

spec.html
ryanjduffy: I've split them into separate lines. Seems to read better but open to suggestions as well.

spec.html
ryanjduffy: @bakkot - Thanks! I've made a few related changes to prefer descending order (e.g. [this one](https://github.com/tc39/ecma262/pull/2122/commits/e7c17df9c70b0575e0d483fc09b67fbeb50a452d#diff-3540caefa502006d8a33cb1385720803R28309)) but the `log` entries were indeed the primary ones.

jmdyck (dismissed):
The phrase
`an implementation-approximated value representing the result of the <foo> of _n_`
is rather long. I don't think "the result of" is necessary, so I suggest dropping it:
`an implementation-approximated value representing the <foo> of _n_`

Interestingly, there's nothing about "implementation-approximated value" that says it's a Number. Nor is there an over-arching requirement that allows us to infer that the value must be a Number. (The first para of this clause used to say "The value returned by each function is a Number", but this PR removed it.) So maybe reinstate that sentence, or insert "Number":
`an implementation-approximated Number value representing the <foo> of _n_`

Moreover, I think the "representing" is incorrect. The resulting Number value doesn't *represent* `the <foo> of _n_`; rather it represents a mathematical value that approximates `the <foo> of _n_`. So you could maybe say:
`a Number value that implementation-approximates the <foo> of _n_`
but then you don't get auto-linking to "implementation-approximated".

So you could coin a phrase, e.g.
`a Number value that approximates _x_`
or
`a Number-approximation for _x_`
and then define it in 6.1.6.1 "The Number Type", right after defining [`the Number value for _x_`](https://tc39.es/ecma262/#number-value), and mention there that it's implementation-approximated.

Note that an implementation-approximated facility is supposed to recommend "an ideal behaviour". So you could say that for `a Number value that approximates _x_`, the ideal behaviour is `the Number value for _x_`.

bakkot:
@jmdyck I agree that the "an implementation-approximated value representing the result of <foo> of _n_" phrasing is awkward and a little imprecise, but its use for math predates this PR; I think we shouldn't worry about changing it here, and instead explore alternative phrasings in a different issue.

jmdyck:
> its use for math predates this PR

True, although in preambles rather than algorithm steps.

> I think we shouldn't worry about changing it here, and instead explore alternative phrasings in a different issue.

Okay.

ryanjduffy:
Thanks for the feedback, @jmdyck. Fixed up the suggestions other than the discussion on "implementation-approximated value."

bakkot (approved):
LGTM. Thanks for sticking with it!

spec.html
```diff
@@ -28087,278 +28085,158 @@ <h1>Function Properties of the Math Object</h1>
...
+        <p>When the `Math.ceil` method is called with argument _x_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _n_ be ? ToNumber(_x_).
+          1. If _n_ is *NaN*, _n_ is *+0*, _n_ is *-0*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+          1. If _n_ &lt; 0 and _n_ &gt; -1, return *-0*.
+          1. If _n_ is an integral Number, return _n_.
+          1. Return the smallest (closest to *-&infin;*) integral Number value that is not less than _n_.
+        </emu-alg>
+        <emu-note>
+          <p>The value of `Math.ceil(x)` is the same as the value of `-Math.floor(-x)`.</p>
```
ljharb, line 28233: i wonder if it'd be simpler to just have this algorithm return `- %Math.floor(-x)%` rather than repeating the steps and having the note.

spec.html
```diff
@@ -28502,24 +28329,18 @@ <h1>Math.fround ( _x_ )</h1>
...
+        <p>When the `Math.hypot` method is called with at least two arguments _value1_ and _value2_ and any number of additional arguments which form the rest parameter ..._values_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _numbers_ be a List containing _value1_, _value2_, and the elements of _values_ in List order.
+          1. Let _onlyZero_ be *true*.
+          1. For each element _number_ of _numbers_, do
+            1. Let _n_ be ? ToNumber(_number_).
+            1. If _n_ is *NaN*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+            1. If _n_ is neither *+0* nor *-0*, set _onlyZero_ to *false*.
+          1. If _onlyZero_ is *true*, return *+0*.
+          1. Return an implementation-approximated value representing the square root of the sum of squares of the elements of _numbers_.
```
ljharb, line 28342: Given that the elements of `numbers` aren't actually all numbers - they might be objects that are `ToNumber`ed - it seems like this actually mandates *two* observable ToNumber calls for any object value.

Obviously this is not the intent, and it's likely that all implementations only call ToNumber once per item, but I think we should lock this down in a future PR.

ljharb (approved):
Fully re-reviewed this; my comments are nonblockers and should be addressed in future PRs.

spec.html
bakkot: Nice catch. I think we should fix that here, probably by constructing a second list while iterating over the first.

ljharb:
I'm going to rebase and land this now; for any additional feedback, please submit an issue or a PR :-)

ryanjduffy:
Thanks all! Good to see it land. 141 comments is definitely a new personal record for a PR :)

michaelficarra:
@ryanjduffy Yeah we're pretty thorough here. #2045 is currently sitting at 281 comments. If you want to take up any more open issues, I'd be happy to help out.

ryanjduffy:
@michaelficarra - not sure if 281 comments is supposed to be encouraging or intimidating ;)

PR #2056, opened 2020-06-19
Editorial: remove legacy intrinsics

ljharb:
All uses of these have been removed from HTML and 402:
 - https://github.com/heycam/webidl/pull/898
 - https://github.com/whatwg/html/pull/5655
 - https://github.com/whatwg/infra/pull/316
 - https://github.com/tc39/ecma402/pull/474

Closes #2023. Closes #2124.

PR #2124, opened 2020-08-03
Editorial: remove unused %AggregateErrorPrototype% intrinsic

ljharb:
This is non-normative since it is not a function, and nothing uses it in this or any other spec. It was mistakenly added in #2040.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2773,17 +2773,6 @@
             </tr>
             <tr>
               <td>
-                %AggregateErrorPrototype%
-              </td>
-              <td>
-                `AggregateError.prototype`
-              </td>
-              <td>
-                The initial value of the `prototype` data property of %AggregateError%; i.e., %AggregateError.prototype%
-              </td>
-            </tr>
-            <tr>
-              <td>
                 %Array%
               </td>
               <td>


ljharb:
Just realized why none of these aren't auto-linking - https://tc39.es/ecma262/#sec-nativeerror used to dynamically refer to them, but now it uses the dotted intrinsic form.

We could save this one for #2056, and leave it now for consistency, or merge this now. Up to yall.

michaelficarra:
I say just throw it in to #2056. In the editor call, I would like to talk about the poor usability of our dynamic definitions of errors and typed arrays.

ExE-Boss:
I suggested this in <https://github.com/tc39/ecma262/pull/2040#discussion_r438133010>, but it was rejected by @chicoxyzzy because of <https://github.com/tc39/proposal-promise-any/pull/37#discussion_r328162938>.

ljharb:
Indeed :-) I did some grepping and couldn't find uses of this one, so I made the PR; but then I realized it was the NativeError magic.

caridy:
I'm getting confused by the different comments here. My assumption is that you cannot remove this since it is an undeniable intrinsic. Some objects will be created with `__proto__` set to this intrinsic object.

ljharb:
@caridy i'm not sure why that makes a difference? `%AggregateError.prototype%` will always refer to this object; the non-dotted form is a legacy way of referencing it. An intrinsic no longer needs to be "well-known" to be accessible via the % syntax.

caridy:
@ljharb thanks for the clarification. It makes perfect sense now!

syg:
There's still %ErrorPrototype% use, so should this PR be closed?

PR #2127, opened 2020-08-06
Normative: Add missing case in 'Static Semantics: TV and TRV'

h2oche:
In current [Static Semantics: TV and TRV](https://tc39.es/ecma262/#sec-static-semantics-tv-and-trv), [`NotEscapeSequence :: DecimalDigit but not 0`](https://tc39.es/ecma262/#prod-NotEscapeSequence) case is undefined.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12148,6 +12148,9 @@
             The TRV of <emu-grammar>NotEscapeSequence :: `0` DecimalDigit</emu-grammar> is the sequence consisting of the code unit 0x0030 (DIGIT ZERO) followed by the code units of the TRV of |DecimalDigit|.
           </li>
           <li>
+            The TRV of <emu-grammar>NotEscapeSequence :: DecimalDigit but not `0`</emu-grammar> is the code unit of the TRV of |DecimalDigit|.
+          </li>
+          <li>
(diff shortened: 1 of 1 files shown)


rkirsling:
My reading on this has been that it's not missing but rather auto-expanded and covered by:
> The TRV of DecimalDigit :: one of 0 1 2 3 4 5 6 7 8 9 is the UTF16Encoding of the single code point matched by this production.

While it's possible that the clarity could be improved, the current approach makes sense because this production of NotEscapeSequence is deferring directly to DecimalDigit without adding anything to it.

bakkot:
Thanks for the patch, but as @rkirsling says, this isn't necessary. Specifically, this is covered by the "chain production" rule under [Notational Conventions](https://tc39.es/ecma262/#sec-notational-conventions):

> A [chain production](https://tc39.es/ecma262/#sec-context-free-grammars) is a production that has exactly one nonterminal symbol on its right-hand side along with zero or more terminal symbols.

> [...]

> Unless explicitly specified otherwise, [all chain productions have an implicit definition for every operation that might be applied to that production's left-hand side nonterminal](https://tc39.es/ecma262/#sec-algorithm-conventions-syntax-directed-operations). The implicit definition simply reapplies the same operation with the same parameters, if any, to the chain production's sole right-hand side nonterminal and then returns the result.

h2oche:
Thanks for kind reply. I was not aware of notational convention of chain production. I'll close this PR.

PR #2133, opened 2020-08-11
Editorial: Update clause range references

gibson042:
The "Managing Memory" clause inserted by 71c08971700eeea4f55f49edc6973670c0d4ee2e broke some clause range references. We might want a more automated way to catch this, but such restructuring is rare and in the meantime it's just a handful of changes to merge.


bakkot:
Nice catch.

> We might want a more automated way to catch this

Yeah, if you replace `clause 6` with `clause <emu-xref href="#sec-ecmascript-data-types-and-values"></emu-xref>` (or whatever), then the references can't get stale. The emu-xref will turn into a link whose text is the number of the clause (here "6") and whose target is the actual clause.

I think that's probably better than updating the numbers manually. Do you want to do update all the explicit references to clause numbers yourself? I can take care of it otherwise.

spec.html
```diff
@@ -25189,8 +25189,8 @@ <h1>ECMAScript Standard Built-in Objects</h1>
...
-  <p>Every accessor property described in clauses 18 through 26 and in Annex <emu-xref href="#sec-additional-built-in-properties"></emu-xref> has the attributes { [[Enumerable]]: *false*, [[Configurable]]: *true* } unless otherwise specified. If only a get accessor function is described, the set accessor function is the default value, *undefined*. If only a set accessor is described the get accessor is the default value, *undefined*.</p>
+  <p>Every other data property described in clauses 18 through 27 and in Annex <emu-xref href="#sec-additional-built-in-properties"></emu-xref> has the attributes { [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *true* } unless otherwise specified.</p>
+  <p>Every accessor property described in clauses 18 through 27 and in Annex <emu-xref href="#sec-additional-built-in-properties"></emu-xref> has the attributes { [[Enumerable]]: *false*, [[Configurable]]: *true* } unless otherwise specified. If only a get accessor function is described, the set accessor function is the default value, *undefined*. If only a set accessor is described the get accessor is the default value, *undefined*.</p>
```
ljharb, line 25193: can these numbers use the new "labelled step" mechanism we have, so they're automatically kept up to date?

spec.html
```diff
@@ -7533,7 +7533,7 @@ <h1>CreateIntrinsics ( _realmRec_ )</h1>
...
+        1. Set fields of _intrinsics_ with the values listed in <emu-xref href="#table-7"></emu-xref>. The field names are the names listed in column one of the table. The value of each field is a new object value fully and recursively populated with property values as defined by the specification of each object in clauses 18 through 27. All object property values are newly created object values. All values that are built-in function objects are created by performing CreateBuiltinFunction(_steps_, _slots_, _realmRec_, _prototype_) where _steps_ is the definition of that function provided by this specification, _slots_ is a list of the names, if any, of the function's specified internal slots, and _prototype_ is the specified value of the function's [[Prototype]] internal slot. The creation of the intrinsics and their properties must be ordered to avoid any dependencies upon objects that have not yet been created.
```
ljharb, line 7536: can these numbers use the new "labelled step" mechanism we have, so they're automatically kept up to date?

gibson042:
> Yeah, if you replace `clause 6` with `clause <emu-xref href="#sec-ecmascript-data-types-and-values"></emu-xref>` (or whatever), then the references can't get stale. The emu-xref will turn into a link whose text is the number of the clause (here "6") and whose target is the actual clause.

References wouldn't get stale, but it'd still be baking in an assumption that the relevant clauses appear in order and fully encompass the standard library (which would be violated if a standard library section is added before The Global Object or after Reflection, or if either of those move). It would be best if there were a single link target, but I understand not wanting to incur the additional depth hit for what is probably the most popular 40% of the spec (or wait for TOC construction changes supporting subsection promotion).

Given the above, do you still want the explicit references?

bakkot:
> It would be best if there were a single link target

I'm not sure what you're suggesting here; can you sketch an example?

> Given the above, do you still want the explicit references?

Yeah, I think `emu-xref`s are a strict improvement over writing out the numbers explicitly, even though they don't solve all issues.

jmdyck:
> > It would be best if there were a single link target
> 
> I'm not sure what you're suggesting here; can you sketch an example?

E.g., if you took clauses "The Global Object" through "Reflection" and made them children of a new emu-clause (called, say, "The ECMAScript Standard Library"), then you could xref to that one clause rather than to "clauses 17 through 27".

jmdyck:
There are also clause-range references in [4.5 Organization of This Specification](https://tc39.es/ecma262/#sec-organization-of-this-specification).

gibson042:
Thanks for that find, @jmdyck. @bakkot I've replaced the literal clause numbers with \<emu-xref>s.

spec.html
```diff
@@ -526,7 +526,7 @@ <h1>The Numeric String Grammar</h1>
 
     <emu-clause id="sec-syntactic-grammar">
       <h1>The Syntactic Grammar</h1>
-      <p>The <em>syntactic grammar</em> for ECMAScript is given in clauses 11, 12, 13, 14, and 15. This grammar has ECMAScript tokens defined by the lexical grammar as its terminal symbols (<emu-xref href="#sec-lexical-and-regexp-grammars"></emu-xref>). It defines a set of productions, starting from two alternative goal symbols |Script| and |Module|, that describe how sequences of tokens form syntactically correct independent components of ECMAScript programs.</p>
+      <p>The <em>syntactic grammar</em> for ECMAScript is given in clauses <emu-xref href="#sec-ecmascript-language-expressions"></emu-xref> through <emu-xref href="#sec-ecmascript-language-scripts-and-modules"></emu-xref>. This grammar has ECMAScript tokens defined by the lexical grammar as its terminal symbols (<emu-xref href="#sec-lexical-and-regexp-grammars"></emu-xref>). It defines a set of productions, starting from two alternative goal symbols |Script| and |Module|, that describe how sequences of tokens form syntactically correct independent components of ECMAScript programs.</p>
```
bakkot, line 529: I see that this drops the reference to clause 11 (which has not changed location in some time). I assume that additional change is intentional, since it seems to be a correct change.

spec.html
```diff
@@ -541,7 +541,7 @@ <h1>The Syntactic Grammar</h1>
...
+      <p>The syntactic grammar as presented in clauses <emu-xref href="#sec-ecmascript-language-expressions"></emu-xref> through <emu-xref href="#sec-ecmascript-language-scripts-and-modules"></emu-xref> is not a complete account of which token sequences are accepted as a correct ECMAScript |Script| or |Module|. Certain additional token sequences are also accepted, namely, those that would be described by the grammar if only semicolons were added to the sequence in certain places (such as before line terminator characters). Furthermore, certain token sequences that are described by the grammar are not considered acceptable if a line terminator character appears in certain &ldquo;awkward&rdquo; places.</p>
```
michaelficarra, line 544: I'd rather list them out instead of using a range. Now we're dependent on the clause ordering. If we list them out, we can re-order clauses all we want without breaking these references.

bakkot: @michaelficarra We'd still need to update this sentence because such a reordering would result in these being listed out of order.

More generally, I think readability ought to take precedence over maintainability (especially for things which rarely require maintenance, as here), and I think "as presented in clause 12 through 15" is more readable than "as presented in clauses 12, 13, 14, and 15", so I would prefer to keep the current PR's wording.

spec.html
gibson042: Correct.

spec.html
michaelficarra: I will reluctantly accept this change since it is easy and doesn't make things worse. For posterity, my preference would be to list out an `<emu-xref>` for each clause, and when ecmarkup sees that they represent a range, it could render it as we see here. This would address the issues of finding occurrences of the clause ID within the source and including the reference in a clause's reference list. But the other editors don't agree it's worthwhile.

PR #2134, opened 2020-08-11
Editorial: Use consistent "...generally do not look into function definitions" language

gibson042:
There is a general pattern in which Contains abstract operations do not descend into function bodies, called out with notes like "Static semantic rules that depend upon substructure generally do not look into function definitions." ([example](https://tc39.es/ecma262/#sec-function-definitions-static-semantics-contains)). But the note is missing for [`async function` productions](https://tc39.es/ecma262/#sec-async-function-definitions-static-semantics-Contains) and is strangely worded (with "normally") for concise functions (both [sync](https://tc39.es/ecma262/#sec-arrow-function-definitions-static-semantics-contains) and [async](https://tc39.es/ecma262/#sec-async-arrow-function-definitions-static-semantics-Contains)).

This PR fixes those issues.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19879,7 +19879,7 @@
         1. Return |ConciseBody| Contains _symbol_.
       </emu-alg>
       <emu-note>
-        <p>Normally, Contains does not look inside most function forms. However, Contains is used to detect `new.target`, `this`, and `super` usage within an |ArrowFunction|.</p>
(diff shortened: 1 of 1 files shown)


michaelficarra (approved):
WFM

syg (approved):

PR #2137, opened 2020-08-17
Editorial: replace "and let"/"and return" with better phrasing

michaelficarra:


spec.html
```diff
@@ -27763,7 +27764,8 @@ <h1>Number.prototype.toExponential ( _fractionDigits_ )</h1>
               1. [id="step-number-proto-toexponential-intermediate-values"] Let _e_, _n_, and _f_ be integers such that _f_ &ge; 0, 10<sup>_f_</sup> &le; _n_ &lt; 10<sup>_f_ + 1</sup>, the Number value for ℝ(_n_) &times; 10<sub>ℝ</sub><sup>ℝ(_e_) - ℝ(_f_)</sup> is _x_, and _f_ is as small as possible. Note that the decimal representation of _n_ has _f_ + 1<sub>ℝ</sub> digits, _n_ is not divisible by 10, and the least significant digit of _n_ is not necessarily uniquely determined by these criteria.
             1. Let _m_ be the String value consisting of the digits of the decimal representation of _n_ (in order, with no leading zeroes).
           1. If _f_ &ne; 0, then
-            1. Let _a_ be the first code unit of _m_, and let _b_ be the remaining _f_ code units of _m_.
+            1. Let _a_ be the first code unit of _m_.
+            1. Let _b_ be the last _f_ code units of _m_.
```
bakkot, line 27768: I don't like this change because it loses the context that this is all the code units except the first.

bakkot:
Note that this duplicates some of the work in https://github.com/tc39/ecma262/pull/2112; I'd prefer to land that first, because it is older and this PR will be easier to rebase.

spec.html
syg: +1 to this, here and below where the operations are clearly segmenting code units into a head and tail, a single step seems quite natural compared to two steps.

michaelficarra: Suggested alternative wording?

michaelficarra:
@bakkot Now that I look into #2112 more, I prefer landing this one first. The changes that overlap between the PRs have nothing to do with RegExp, and it will allow #2112 to be more focused on the RegExp related changes.

spec.html
bakkot: The existing wording seems better than the wording in this PR. I'd also be OK with

```
1. Let _a_ be the first code unit of _m_.
1. Let _b_ be the other _f_ code units of _m_.
```

michaelficarra: @bakkot @syg Addressed using Kevin's suggested phrasing.

spec.html
```diff
@@ -43360,13 +43379,13 @@ <h1>Pattern Semantics</h1>
         <p>The production <emu-grammar>ExtendedAtom :: `\` [lookahead == `c`]</emu-grammar> evaluates as follows:</p>
         <emu-alg>
           1. Let _A_ be the CharSet containing the single character `\\` U+005C (REVERSE SOLIDUS).
-          1. Call CharacterSetMatcher(_A_, *false*, _direction_) and return its Matcher result.
+          1. Return CharacterSetMatcher(_A_, *false*, _direction_).
```
gibson042, line 43382: The changes in B.1.4.4 Pattern Semantics ([Annex B] Pattern Semantics) should also be addressed by #2112.

michaelficarra:
Rebased.

michaelficarra:
Rebased.

PR #2139, opened 2020-08-17
Editorial: rest parameters are reified as spec Lists

michaelficarra:
This PR is meant to be used as an example in the next editor call to decide whether we want rest parameters to be automatically reified in the associated algorithm steps as a spec List. We'll probably have to add a sentence to the Notational Conventions section to explicitly describe this.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -772,7 +772,7 @@
   <emu-clause id="sec-algorithm-conventions">
     <h1>Algorithm Conventions</h1>
     <p>The specification often uses a numbered list to specify steps in an algorithm. These algorithms are used to precisely specify the required semantics of ECMAScript language constructs. The algorithms are not intended to imply the use of any specific implementation technique. In practice, there may be more efficient algorithms available to implement a given feature.</p>
-    <p>Algorithms may be explicitly parameterized, in which case the names and usage of the parameters must be provided as part of the algorithm's definition.</p>
+    <p>Algorithms may be explicitly parameterized with an ordered, comma-separated sequence of alias names which may be used within the algorithm steps to reference the argument passed in that position. Optional parameters are denoted with surrounding brackets ([ , _name_ ]) and are no different from required parameters within algorithm steps. A rest parameter may appear at the end of a parameter list, denoted with leading ellipsis (, ..._name_). The rest parameter captures all of the arguments provided following the required and optional parameters into a List. If there are no such additional arguments, that List is empty.</p>
     <p>Algorithm steps may be subdivided into sequential substeps. Substeps are indented and may themselves be further divided into indented substeps. Outline numbering conventions are used to identify substeps with the first level of substeps labelled with lower case alphabetic characters and the second level of substeps labelled with lower case roman numerals. If more than three levels are required these rules repeat with the fourth level using numeric labels. For example:</p>
     <emu-alg type="example">
       1. Top-level step
@@ -25268,7 +25268,7 @@
   <p>Unless otherwise specified every built-in prototype object has the Object prototype object, which is the initial value of the expression `Object.prototype` (<emu-xref href="#sec-properties-of-the-object-prototype-object"></emu-xref>), as the value of its [[Prototype]] internal slot, except the Object prototype object itself.</p>
   <p>Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function.</p>
   <p>Each built-in function defined in this specification is created by calling the CreateBuiltinFunction abstract operation (<emu-xref href="#sec-createbuiltinfunction"></emu-xref>).</p>
-  <p>Every built-in function object, including constructors, has a *"length"* property whose value is an integer. Unless otherwise specified, this value is equal to the largest number of named arguments shown in the subclause headings for the function description. Optional parameters (which are indicated with brackets: `[` `]`) or rest parameters (which are shown using the form &laquo;...name&raquo;) are not included in the default argument count.</p>
+  <p>Every built-in function object, including constructors, has a *"length"* property whose value is an integer. Unless otherwise specified, this value is equal to the number of required parameters shown in the subclause headings for the function description. Optional parameters and rest parameters are not included in the parameter count.</p>
   <emu-note>
     <p>For example, the function object that is the initial value of the *"map"* property of the Array prototype object is described under the subclause heading &laquo;Array.prototype.map (callbackFn [ , thisArg])&raquo; which shows the two named arguments callbackFn and thisArg, the latter being optional; therefore the value of the *"length"* property of that function object is 1.</p>
   </emu-note>
@@ -26184,7 +26184,6 @@
         <emu-alg>
           1. Let _to_ be ? ToObject(_target_).
           1. If only one argument was passed, return _to_.
-          1. Let _sources_ be the List of argument values starting with the second argument.
           1. For each element _nextSource_ of _sources_, in ascending index order, do
             1. If _nextSource_ is neither *undefined* nor *null*, then
               1. Let _from_ be ! ToObject(_nextSource_).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -772,7 +772,7 @@ <h1>Grammar Notation</h1>
...
     <h1>Algorithm Conventions</h1>
     <p>The specification often uses a numbered list to specify steps in an algorithm. These algorithms are used to precisely specify the required semantics of ECMAScript language constructs. The algorithms are not intended to imply the use of any specific implementation technique. In practice, there may be more efficient algorithms available to implement a given feature.</p>
-    <p>Algorithms may be explicitly parameterized, in which case the names and usage of the parameters must be provided as part of the algorithm's definition.</p>
+    <p>Algorithms may be explicitly parameterized, in which case the names and usage of the parameters must be provided as part of the algorithm's definition. <dfn>Optional parameters</dfn> are denoted with surrounding brackets (`[ , _name_ ]`) and are no different from mandatory parameters within algorithm steps. A <dfn>rest parameter</dfn> may appear at the end of a parameter list, denoted with leading elipsis (`...name`). The rest parameter captures all of the arguments provided following the mandatory and optional parameters, if any, into a possibly-empty List.</p>
```
bakkot, line 775: > The rest parameter captures all of the arguments provided following the mandatory and optional parameters, if any, into a possibly-empty List.

I think it'd be good to explicitly say the parameter introduces an alias. Maybe:

> A rest parameter introduces an alias whose name is the name of the parameter and which holds a possibly-empty List whose elements are any arguments provided following the mandatory and optional parameters, in order.

---

also, s/elipsis/ellipsis/

ljharb: it's not actually an ellipsis (`…`) - it's three dots (`...`) for consistency with JS spread/rest syntax.

michaelficarra: @ljharb An ellipsis doesn't need to be written as the precomposed code point. See https://en.wikipedia.org/wiki/Ellipsis. It's still an ellipsis either way.

ljharb: fair enough, but i still find the term confusing since it *includes* the precomposed code point, and that won't appear in our prose.

michaelficarra: But I show an example of how we will write it immediately following it. And if someone searches for the precomposed code point, modern browsers will also find the decomposed results (and vice versa).

ljharb: That last point in particular (something i was unaware of) resolves my concern, thanks.

syg, line 775: Is this section meant to apply to both built-ins and to AOs and ACs? We need to say something about the mandatory nature of parameters, but it's treated differently from built-ins to AOs. I assume it's an error for AOs to use a parameter with an absent argument, but built-ins can and assume it's undefined, right?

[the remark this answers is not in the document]
michaelficarra: While I would prefer that, I don't think that's the Oxford spelling. See https://en.wikipedia.org/wiki/Oxford_spelling#Language_tag_comparison.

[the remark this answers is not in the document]
syg: Are you kidding me, Oxford uses a mixture of z and s for the same phoneme?? Why are we using this yahoo spelling?

michaelficarra:
I updated the notational conventions with a slightly more substantial change. Please re-review.

spec.html
```diff
@@ -9039,7 +9039,7 @@ <h1>FunctionDeclarationInstantiation ( _func_, _argumentsList_ )</h1>
           1. Set the LexicalEnvironment of _calleeContext_ to _env_.
         1. For each String _paramName_ in _parameterNames_, do
           1. Let _alreadyDeclared_ be _env_.HasBinding(_paramName_).
-          1. NOTE: Early errors ensure that duplicate parameter names can only occur in non-strict functions that do not have parameter default values or rest parameters.
+          1. NOTE: Early errors ensure that duplicate parameter names can only occur in non-strict functions that do not have parameter default values or <emu-not-ref>rest parameters</emu-not-ref>.
```
bakkot, line 9042: I'd prefer we just not `dfn` this term, rather than `emu-not-ref`'ing half of its uses.

spec.html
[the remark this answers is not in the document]
michaelficarra: I'd be in favour of transitioning to en-GB. I miss the old "object initialiser" spelling.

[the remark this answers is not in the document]
ljharb: Hard -1 on that :-) i'd prefer moving entirely to American spellings if we're going to make a change :-p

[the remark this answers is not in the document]
michaelficarra: Never! I'll take the horrible amalgam that is Oxford spellings over the misbegotten American spellings any day.

michaelficarra: The handling of missing mandatory parameters is detailed in [section 17](https://tc39.es/ecma262/#sec-ecmascript-standard-built-in-objects). The paragraph here is written in a way that applies equally to built-ins and AOs.

bakkot: > We need to say something about the mandatory nature of parameters, but it's treated differently from built-ins to AOs. I assume it's an error for AOs to use a parameter with an absent argument, but built-ins can and assume it's undefined, right?

I don't think we need anything further: the behavior for built-ins is already spelled out below, in [Clause 17](https://tc39.es/ecma262/#sec-ecmascript-standard-built-in-objects):

> Unless otherwise specified in the description of a particular function, if a built-in function or constructor is given fewer arguments than the function is specified to require, the function or constructor shall behave exactly as if it had been given sufficient additional arguments, each such argument being the undefined value. Such missing arguments are considered to be “not present” and may be identified in that manner by specification algorithms.

And of course for AOs and ACs it would be a _spec_ error; we don't need to spell out the behavior for invoking an AO or AC without all required arguments because the situation does not arise.

(It's not entirely clear to me why the description of editorial conventions for built-in functions is split across [Clause 9.3](https://tc39.es/ecma262/#sec-built-in-function-objects) "Built-In Function Objects" and Clause 17 "ECMAScript Standard Built-in Objects"; that could perhaps be improved in a different PR.)

spec.html
```diff
@@ -25268,7 +25268,7 @@ <h1>ECMAScript Standard Built-in Objects</h1>
...
   <p>Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function.</p>
   <p>Each built-in function defined in this specification is created by calling the CreateBuiltinFunction abstract operation (<emu-xref href="#sec-createbuiltinfunction"></emu-xref>).</p>
-  <p>Every built-in function object, including constructors, has a *"length"* property whose value is an integer. Unless otherwise specified, this value is equal to the largest number of named arguments shown in the subclause headings for the function description. Optional parameters (which are indicated with brackets: `[` `]`) or rest parameters (which are shown using the form &laquo;...name&raquo;) are not included in the default argument count.</p>
+  <p>Every built-in function object, including constructors, has a *"length"* property whose value is an integer. Unless otherwise specified, this value is equal to the number of mandatory parameters shown in the subclause headings for the function description. Optional parameters and rest parameters are not included in the parameter count.</p>
```
bakkot, line 25271:
```suggestion
  <p>Every built-in function object, including constructors, has a *"length"* property whose value is an integer. Unless otherwise specified, this value is equal to the number of required parameters shown in the subclause headings for the function description. Optional parameters and rest parameters are not included in the parameter count.</p>
```

to match the "fewer arguments than the function is specified to require" wording in clause 17.

michaelficarra: I prefer the term "mandatory", but sure.

spec.html
jmdyck: > (It's not entirely clear to me why the description of editorial conventions for built-in functions is split across [Clause 9.3](https://tc39.es/ecma262/#sec-built-in-function-objects) "Built-In Function Objects" and Clause 17 "ECMAScript Standard Built-in Objects"; that could perhaps be improved in a different PR.)

See PR #539.

bakkot: @syg Do you think the parts I've mentioned are sufficient, or would you still like something more here?

syg: They seem sufficient to me.

PR #2140, opened 2020-08-17
Editorial: fix a minor wording oddity in the splice summary

michaelficarra:


spec.html
```diff
@@ -34905,7 +34905,7 @@ <h1>Runtime Semantics: SortCompare ( _x_, _y_ )</h1>
       <emu-clause id="sec-array.prototype.splice">
         <h1>Array.prototype.splice ( _start_, _deleteCount_, ..._items_ )</h1>
         <emu-note>
-          <p>When the `splice` method is called with two or more arguments _start_, _deleteCount_ and zero or more _items_, the _deleteCount_ elements of the array starting at integer index _start_ are replaced by the arguments _items_. An Array object containing the deleted elements (if any) is returned.</p>
+          <p>When the `splice` method is called with two or more arguments _start_, _deleteCount_ and zero or more _items_, the _deleteCount_ elements of the array starting at integer index _start_ are replaced by _items_. An Array object containing the deleted elements (if any) is returned.</p>
```
ljharb, line 34908:
```suggestion
          <p>When the `splice` method is called with two or more arguments _start_, _deleteCount_ and zero or more _items_, the _deleteCount_ elements of the array starting at integer index _start_ are replaced by the contents of _items_. An Array object containing the deleted elements (if any) is returned.</p>
```
otherwise it kind of implies that the elements are replaced by the single array `items`.

michaelficarra:
@ljharb What do you think about "the elements of" instead?

ljharb:
@michaelficarra totally fine; as long as the array itself isn’t implied, my concern is addressed.

PR #2144, opened 2020-08-19
Editorial: Merge IsStringPrefix and SplitMatch abstract ops

shvaikalesh:
This PR merges [`IsStringPrefix`](https://tc39.es/ecma262/#sec-isstringprefix) and [`SplitMatch`](https://tc39.es/ecma262/#sec-splitmatch) abstract ops into single well-defined `StringIncludesAt` helper, improving:
* unobvious parameter order of [`IsStringPrefix`](https://tc39.es/ecma262/#sec-isstringprefix);
* too specific name and polymorphic return type of [`SplitMatch`](https://tc39.es/ecma262/#sec-splitmatch).

We can't use [`StringIndexOf`](https://tc39.es/ecma262/#sec-stringindexof) instead, since it searches until the end of string rather than matching substring only at index.


spec.html
```diff
@@ -5624,19 +5646,6 @@ <h1>IsRegExp ( _argument_ )</h1>
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-isstringprefix" aoid="IsStringPrefix">
-      <h1>IsStringPrefix ( _p_, _q_ )</h1>
```
michaelficarra, line 5628: For readability, we could leave this with an implementation that just passes `0` as the third parameter to `StringIncludesAt`. Not sure if it's worth it.

spec.html
```diff
@@ -1037,6 +1037,28 @@ <h1>Runtime Semantics: StringIndexOf ( _string_, _searchValue_, _fromIndex_ )</h
           <p>This algorithm always returns -1 if _fromIndex_ &gt; the length of _string_.</p>
         </emu-note>
       </emu-clause>
+
+      <emu-clause id="sec-stringincludesat" aoid="StringIncludesAt">
+        <h1>Runtime Semantics: StringIncludesAt ( _string_, _target_, _atIndex_ )</h1>
```
michaelficarra, line 1042: These parameter names make it more immediately obvious to me which is which.

```suggestion
        <h1>Runtime Semantics: StringIncludesAt ( _haystack_, _needle_, _atIndex_ )</h1>
```

spec.html
```diff
@@ -31027,9 +31036,9 @@ <h1>String.prototype.split ( _separator_, _limit_ )</h1>
           1. Let _p_ be 0.
           1. Let _q_ be _p_.
           1. Repeat, while _q_ &ne; _s_,
-            1. Let _e_ be SplitMatch(_S_, _q_, _R_).
-            1. If _e_ is *false*, set _q_ to _q_ + 1.
+            1. If ! StringIncludesAt(_S_, _R_, _q_) is *false*, then set _q_ to _q_ + 1.
             1. Else,
+              1. Let _e_ be _q_ + the number of code units in _R_.
```
michaelficarra, line 31041: This seems just unnecessarily verbose.

```suggestion
              1. Let _e_ be _q_ + the length of _R_.
```

spec.html
shvaikalesh: I came to appreciate haystack/needle naming ever since I saw it in PHP docs, however:
* it implies searching a _needle_ in a _haystack_, yet this op is rather matches a substring at given index;
* it may sound not formal enough for the spec?

spec.html
shvaikalesh: It's maybe just for me, but parameter order of [`IsStringPrefix`](https://tc39.es/ecma262/#sec-isstringprefix) isn't immediately clear: is first parameter a string or a prefix?

michaelficarra: I feel the same way about `StringIncludesAt`.

spec.html
michaelficarra: If you don't like the haystack/needle searching metaphor, there's plenty of other alternatives that are better than what is here right now: whole/part, string/substring, outer/inner, etc.

shvaikalesh: Thank you, switched to string/substring @ 8a80eee.

spec.html
shvaikalesh: It would be great if we came up with a better name. `StringMatch`?

michaelficarra: I don't really think the name will ever be sufficient. I will always need to take a quick look at the parameter names or the summary. Or infer it from context at the usage site.

spec.html
```diff
@@ -5696,8 +5705,8 @@ <h1>Abstract Relational Comparison</h1>
           1. Let _py_ be ? ToPrimitive(_y_, hint Number).
           1. Let _px_ be ? ToPrimitive(_x_, hint Number).
         1. [id="step-arc-string-check"] If Type(_px_) is String and Type(_py_) is String, then
-          1. If IsStringPrefix(_py_, _px_) is *true*, return *false*.
-          1. If IsStringPrefix(_px_, _py_) is *true*, return *true*.
+          1. If ! StringIncludesAt(_px_, _py_, 0) is *true*, return *false*.
+          1. If ! StringIncludesAt(_py_, _px_, 0) is *true*, return *true*.
```
ljharb, line 5709: i don't find this change clearer :-/

spec.html
```diff
@@ -31031,9 +31040,9 @@ <h1>String.prototype.split ( _separator_, _limit_ )</h1>
           1. Let _p_ be 0.
           1. Let _q_ be _p_.
           1. Repeat, while _q_ &ne; _s_,
-            1. Let _e_ be SplitMatch(_S_, _q_, _R_).
-            1. If _e_ is *false*, set _q_ to _q_ + 1.
+            1. If ! StringIncludesAt(_S_, _R_, _q_) is *false*, then set _q_ to _q_ + 1.
             1. Else,
```
ljharb, line 31044: since the "else" is multiline, i think the "if" needs to be too.

ljharb:
Discussed this on the editor call; let's please leave `IsStringPrefix` and its callsites untouched (but update its implementation to call into `StringIncludesAt`). I'm not personally convinced that `StringIncludesAt` is the best name but don't have a better suggestion at this time.

spec.html
```diff
@@ -1038,6 +1038,28 @@ <h1>Runtime Semantics: StringIndexOf ( _string_, _searchValue_, _fromIndex_ )</h
           <p>This algorithm always returns -1 if _fromIndex_ &gt; the length of _string_.</p>
         </emu-note>
       </emu-clause>
+
+      <emu-clause id="sec-stringincludesat" aoid="StringIncludesAt">
```
gibson042, line 1042:
```suggestion
      <emu-clause id="sec-stringincludesat" aoid="StringIncludesAt" oldids="IsStringPrefix,SplitMatch">
```

spec.html
jmdyck: The spec has about 37 examples of an 'inline' `If` followed by a 'multiline' `Else`. And about 18 examples of the reverse.

ljharb: Compared to how many examples are both multiline or both inline, those seem like cases we should make consistent.

ljharb:
I've rebased this; the first commit implements https://github.com/tc39/ecma262/pull/2144#issuecomment-676783183; the second commit reverts that and implements https://github.com/tc39/ecma262/pull/2144#pullrequestreview-471015688, per one editor's preference on Matrix.

Happy to drop the second, or squash the two, as preferred.

The only unaccounted-for piece in the second commit is adding an oldid somewhere for SplitMatch - I'm not sure where it would go.

Issue #1287, opened 2018-08-14
Text encoding of spec.html: ascii or utf-8?


claudepache:
Currently, the file [spec.html](https://github.com/tc39/ecma262/blob/master/spec.html) contains the declaration

    <meta charset="ascii">

which is contradicted by its contents, because it has two non-ascii, utf-8-encoded characters (namely `≥` on line 7524 and `’` on line 38018).

That makes my text editor complain each time I want to edit the spec, either:

![encoding-mismatch-1](https://user-images.githubusercontent.com/3309637/44086117-d9e1ebcc-9fbb-11e8-9e43-6b931c7bb764.png)

or:

![encoding-mismatch-warning](https://user-images.githubusercontent.com/3309637/44086121-df22ae78-9fbb-11e8-89b4-15fafb1dd420.png)

Two options:

1. Officially allow utf-8-encoded contents, and declare formally spec.html’s charset as utf-8;
2. Replace the two offending characters by the proper html entities, and add an automated test in order to reject further insertions of non-ascii characters.

Personally, I prefer option 1 (`≥` is more readable, more writable (although I guess that may depend on your working environment), and I never recall the html entity for that character).

jmdyck:
The U+2019 right single quotation mark on line 38018 should be a U+0027 apostrophe. Commit 64ab8cfc in my "Misc editorial" PR fixes this.

So it's really just the U+2265 greater-than or equal to on line 7524. For option 2, this could be changed to `&ge;`.

annevk:
Note that `<meta charset=ascii>` isn't really conforming HTML anymore (and conforming implementations will treat it as if you wrote windows-1252). Only UTF-8 is allowed.

jmdyck:
The 'rendering' process generates a file with `<meta charset="utf-8">`, so at least we're okay on that front.

claudepache:
Resolved by dc1e21c454bd316810be1c0e7af0131a2d7f38e9, which follows from the introduction of character ℝ in spec.html.

PR #2149, opened 2020-08-20
Editorial: Replace non-ASCII quotes

gibson042:
The resolution of #1287 established a UTF-8 encoding of spec.html, but in practice the file contents are almost exclusively ASCII with the exception of `𝔽` and `ℝ`. This PR remedies the handful of outliers (a pair of `“…”` left/right quotes and some literal `«…»` guillemets), although I would not object to a subsequent sweep that replaces character references and/or "dumb" quotes.


bakkot:
I'm definitely in favor of getting rid of the smart quotes, but [discussion on ecmarkup](https://github.com/tc39/ecmarkup/issues/173#issuecomment-658447485) suggested that some people would prefer to use the actual codepoint rather than HTML entities for other non-ASCII characters like `«`.

gibson042:
Either way, it seems desirable to use the same patterns throughout the spec. This PR makes the minimum necessary changes to achieve that.

ljharb:
I'm fine temporarily using HTML entities for consistency, but I am not fine with using straight quotes in prose, which are always typographically incorrect. I used curly quotes in Annex E quite on purpose.

gibson042:
> I used curly quotes in Annex E quite on purpose.

Annex E contains no curly quotes; literally the only use of them is around "nullish" in the introduction. And there are several phrases wrapped in U+0022 QUOTATION MARK where curly quotes would be appropriate—why should that one be special?

gibson042:
Oh, I misinterpreted your comment. Updating now.

PR #2150, opened 2020-08-21
Editorial: Add notes about the 1-based indexing of regular expression _captures_

gibson042:
Split from #2112; this can be merged as is or built upon/superceded by a switch to zero-based indexing.


michaelficarra:
@gibson042 The editors would like to avoid using the 1-indexed lists for this purpose, as they are poorly specified and this is the only use for them. Our preferred strategy is to use a regular (0-indexed) spec list for this with the first element populated by `~empty~` We also considered an alternative of maintaining an offset of 1 everywhere the lists are indexed, but decided that that strategy had more risk of being incorrectly implemented. Do you want to update this PR? If not, I can take it (though it may not be very soon).

gibson042:
I may not get to it quickly either, so let's just agree that whoever is ready to pick it up first should communicate that by posting here.

PR #2152, opened 2020-08-23
Editorial: consistency around List iteration

bakkot:
Every place that an algorithm talks about iterating a List it now says `1. For each element _name_ of _list_, do`, rather than saying "in" or "that is a member of" or "from" or whatever. Sometimes an actual type is used in place of "element".

I also added a sentence to the description of the List type that explicitly says Lists are iterated in order, and removed the few places in the spec text which explicitly specified that the order of iteration was the order of the List. (Most places just assume it.)

~Marked as a draft until I get feedback on this and, if we are in favor of these changes, add a lint rule to ecmarkup to enforce this style where possible.~ Edit: ready to go. The first commit bumps ecmarkup, which fails linting, and the second commit changes the list iteration style and then passes linting. They should be landed in reverse order.

See also https://github.com/tc39/ecma262/pull/2142.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4023,6 +4023,7 @@
     <emu-clause id="sec-list-and-record-specification-type">
       <h1>The List and Record Specification Types</h1>
       <p>The <dfn>List</dfn> type is used to explain the evaluation of argument lists (see <emu-xref href="#sec-argument-lists"></emu-xref>) in `new` expressions, in function calls, and in other algorithms where a simple ordered list of values is needed. Values of the List type are simply ordered sequences of list elements containing the individual values. These sequences may be of any length. The elements of a list may be randomly accessed using 0-origin indices. For notational convenience an array-like syntax can be used to access List elements. For example, _arguments_[2] is shorthand for saying the 3<sup>rd</sup> element of the List _arguments_.</p>
+      <p>When an algorithm iterates over the elements of a List without specifying an order, the order used is the order of the elements in the List.</p>
       <p>For notational convenience within this specification, a literal syntax can be used to express a new List value. For example, &laquo; 1, 2 &raquo; defines a List value that has two elements each of which is initialized to a specific value. A new empty List can be expressed as &laquo; &raquo;.</p>
       <p>The <dfn>Record</dfn> type is used to describe data aggregations within the algorithms of this specification. A Record type value consists of one or more named fields. The value of each field is either an ECMAScript value or an abstract value represented by a name associated with the Record type. Field names are always enclosed in double brackets, for example [[Value]].</p>
       <p>For notational convenience within this specification, an object literal-like syntax can be used to express a Record value. For example, { [[Field1]]: 42, [[Field2]]: *false*, [[Field3]]: ~empty~ } defines a Record value that has three fields, each of which is initialized to a specific value. Field name order is not significant. Any fields that are not explicitly listed are considered to be absent.</p>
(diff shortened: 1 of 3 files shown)


jmdyck:
Was there ever the idea that, if a "For each" didn't state an explicit order, then the order didn't matter (i.e., either the order made no observable difference, or it did but the different behaviors were all deemed valid)? Looking at past editions, I don't *think* so, but it's hard to be sure.

In cases where different iteration orders would result in different behaviors, does test262 require "List order"?

spec.html
```diff
@@ -14455,7 +14456,7 @@ <h1>Runtime Semantics: Evaluation</h1>
           1. If _importMeta_ is ~empty~, then
             1. Set _importMeta_ to ! OrdinaryObjectCreate(*null*).
             1. Let _importMetaValues_ be ! HostGetImportMetaProperties(_module_).
-            1. For each Record { [[Key]], [[Value]] } _p_ that is an element of _importMetaValues_, do
+            1. For each Record { [[Key]], [[Value]] } _p_ of _importMetaValues_, do
```
michaelficarra, line 14459:
```suggestion
            1. For each Record _p_ of _importMetaValues_, do
```

I don't think mentioning the fields adds anything here.

spec.html
```diff
@@ -39350,7 +39351,7 @@ <h1>FinalizationRegistry.prototype.unregister ( _unregisterToken_ )</h1>
           1. Perform ? RequireInternalSlot(_finalizationRegistry_, [[Cells]]).
           1. If Type(_unregisterToken_) is not Object, throw a *TypeError* exception.
           1. Let _removed_ be *false*.
-          1. For each Record { [[WeakRefTarget]], [[HeldValue]], [[UnregisterToken]] } _cell_ that is an element of _finalizationRegistry_.[[Cells]], do
+          1. For each Record { [[WeakRefTarget]], [[HeldValue]], [[UnregisterToken]] } _cell_ of _finalizationRegistry_.[[Cells]], do
```
michaelficarra, line 39354: Same here.

```suggestion
          1. For each Record _cell_ of _finalizationRegistry_.[[Cells]], do
```

spec.html
bakkot: Since these are anonymous Records, the fields are essentially part of the type. I'd lightly prefer to keep them.

bakkot:
> You should also look into instances of iteration that empty the list

Good catch, will do.

> There's also some that do indexing manually that we could probably convert.

Yes:
- in [RegExpBuiltinExec](https://tc39.es/ecma262/#sec-regexpbuiltinexec) step 27, for the captures list discussed in https://github.com/tc39/ecma262/pull/2150, which I don't want to touch until we've sorted that out
- in step 15 of [CreateDynamicFunction](https://tc39.es/ecma262/#sec-createdynamicfunction)
- in [String.fromCharCode](https://tc39.es/ecma262/#sec-string.fromcharcode)
- in [String.fromCodePoint](https://tc39.es/ecma262/#sec-string.fromcodepoint)

I'll address the latter three.

michaelficarra:
Be careful, the CreateDynamicFunction one doesn't cover the whole list.

spec.html
ljharb: could we make them proper named records instead?

bakkot: We could, but this and [the host-defined AO which provides this list](https://tc39.es/ecma262/#sec-hostgetimportmetaproperties) are the only two places which talk about this particular kind of record, so I think adding another layer of indirection there would hurt rather than help clarity.

bakkot:
@michaelficarra I added a pair of commits converting those two styles of iteration into the standard style.

There's still an instance of the first style in [ForInIteratorPrototype.next](https://tc39.es/ecma262/#sec-%foriniteratorprototype%.next) step 7.b, but that one is relying on the list mutation to persist state across invocations, so it shouldn't change.

I also left CreateDynamicFunction alone, so the only instances of the second style I changed were in String.fromCharCode and String.fromCodePoint.

PR #2153, opened 2020-08-26
Editorial: Clarify that "length" value of built-in functions is non-negative

shvaikalesh:


spec.html
```diff
@@ -25268,7 +25268,7 @@ <h1>ECMAScript Standard Built-in Objects</h1>
...
   <p>Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function.</p>
   <p>Each built-in function defined in this specification is created by calling the CreateBuiltinFunction abstract operation (<emu-xref href="#sec-createbuiltinfunction"></emu-xref>).</p>
-  <p>Every built-in function object, including constructors, has a *"length"* property whose value is an integer. Unless otherwise specified, this value is equal to the number of required parameters shown in the subclause headings for the function description. Optional parameters and rest parameters are not included in the parameter count.</p>
+  <p>Every built-in function object, including constructors, has a *"length"* property whose value is a non-negative integer. Unless otherwise specified, this value is equal to the number of required parameters shown in the subclause headings for the function description. Optional parameters and rest parameters are not included in the parameter count.</p>
```
michaelficarra, line 25271:
```suggestion
  <p>Every built-in function object, including constructors, has a *"length"* property whose value is a non-negative integral Number. Unless otherwise specified, this value is equal to the number of required parameters shown in the subclause headings for the function description. Optional parameters and rest parameters are not included in the parameter count.</p>
```

We're moving to the terminology "integral Number" for JavaScript Number values that represent an integer.

shvaikalesh: nit: There are 3 occurrences of "integ**er** Number" (`IsInteger`, `IsNonNegativeInteger`, and `String.prototype.codePointAt`), and only one usage of "integ**ral** Number" in `ToInteger`.
Should we align it?

bakkot: I will align those as part of #2007. I'd prefer we not touch any such occurrences which we weren't already going to change so that rebasing is easier.

bakkot (approved):

PR #2154, opened 2020-08-26
Normative: Arbitrary module namespace identifier names

bmeck:
I do not believe this affects any integration downstream.

This PR aims to allow:

```mjs
import {"\0 any unicode" as foo} from "";
export {foo as "\0 any unicode"};
```

This matches the restriction that WASM has that the string names for exports must be valid Unicode.

Depends on #2155

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6065,6 +6065,27 @@
       </emu-note>
     </emu-clause>
 
+    <emu-clause id="sec-isstringwellformedunicode" type="abstract operation">
+      <h1>
+        Static Semantics: IsStringWellFormedUnicode (
+          _string_: a String,
+        )
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It interprets _string_ as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, and determines whether it is a <a href="http://www.unicode.org/glossary/#well_formed_code_unit_sequence">well formed</a> UTF-16 sequence.</dd>
+      </dl>
+      <emu-alg>
+        1. Let _strLen_ be the number of code units in _string_.
+        1. Let _k_ be 0.
+        1. Repeat, while _k_ &ne; _strLen_,
+          1. Let _cp_ be ! CodePointAt(_string_, _k_).
+          1. If _cp_.[[IsUnpairedSurrogate]] is *true*, return *false*.
+          1. Set _k_ to _k_ + _cp_.[[CodeUnitCount]].
+        1. Return *true*.
+      </emu-alg>
+    </emu-clause>
+
     <emu-clause id="sec-samevalue" type="abstract operation">
       <h1>
         SameValue (
@@ -7424,7 +7445,7 @@
         1. Let _names2_ be the BoundNames of |ImportSpecifier|.
         1. Return the list-concatenation of _names1_ and _names2_.
       </emu-alg>
-      <emu-grammar>ImportSpecifier : IdentifierName `as` ImportedBinding</emu-grammar>
+      <emu-grammar>ImportSpecifier : ModuleExportName `as` ImportedBinding</emu-grammar>
       <emu-alg>
         1. Return the BoundNames of |ImportedBinding|.
       </emu-alg>
@@ -17975,6 +17996,12 @@
       <emu-alg>
         1. Return the string-concatenation of 0x0023 (NUMBER SIGN) and the StringValue of |IdentifierName|.
       </emu-alg>
+      <emu-grammar>
+        ModuleExportName : StringLiteral
+      </emu-grammar>
+      <emu-alg>
+        1. Return the SV of |StringLiteral|.
+      </emu-alg>
     </emu-clause>
 
     <emu-clause id="sec-identifiers-runtime-semantics-evaluation">
@@ -25934,6 +25961,10 @@
         ImportDeclaration
         ExportDeclaration
         StatementListItem[~Yield, +Await, ~Return]
+
+      ModuleExportName :
+        IdentifierName
+        StringLiteral
     </emu-grammar>
 
     <emu-clause id="sec-module-semantics">
@@ -25977,6 +26008,10 @@
         <emu-note>
           <p>The duplicate ExportedNames rule implies that multiple `export default` |ExportDeclaration| items within a |ModuleBody| is a Syntax Error. Additional error conditions relating to conflicting or duplicate declarations are checked during module linking prior to evaluation of a |Module|. If any such errors are detected the |Module| is not evaluated.</p>
         </emu-note>
+        <emu-grammar>ModuleExportName : StringLiteral</emu-grammar>
+        <ul>
+          <li>It is a Syntax Error if IsStringWellFormedUnicode(the SV of |StringLiteral|) is *false*.</li>
+        </ul>
       </emu-clause>
 
       <emu-clause id="sec-importedlocalnames" type="abstract operation">
@@ -28023,7 +28058,7 @@
 
         ImportSpecifier :
           ImportedBinding
-          IdentifierName `as` ImportedBinding
+          ModuleExportName `as` ImportedBinding
 
         ModuleSpecifier :
           StringLiteral
@@ -28123,9 +28158,9 @@
           1. Let _entry_ be the ImportEntry Record { [[ModuleRequest]]: _module_, [[ImportName]]: _localName_, [[LocalName]]: _localName_ }.
           1. Return a List whose sole element is _entry_.
         </emu-alg>
-        <emu-grammar>ImportSpecifier : IdentifierName `as` ImportedBinding</emu-grammar>
+        <emu-grammar>ImportSpecifier : ModuleExportName `as` ImportedBinding</emu-grammar>
         <emu-alg>
-          1. Let _importName_ be the StringValue of |IdentifierName|.
+          1. Let _importName_ be the StringValue of |ModuleExportName|.
           1. Let _localName_ be the StringValue of |ImportedBinding|.
           1. Let _entry_ be the ImportEntry Record { [[ModuleRequest]]: _module_, [[ImportName]]: _importName_, [[LocalName]]: _localName_ }.
           1. Return a List whose sole element is _entry_.
@@ -28148,7 +28183,7 @@
 
         ExportFromClause :
           `*`
-          `*` `as` IdentifierName
+          `*` `as` ModuleExportName
           NamedExports
 
         NamedExports :
@@ -28161,14 +28196,17 @@
           ExportsList `,` ExportSpecifier
 
         ExportSpecifier :
-          IdentifierName
-          IdentifierName `as` IdentifierName
+          ModuleExportName
+          ModuleExportName `as` ModuleExportName
       </emu-grammar>
 
       <emu-clause id="sec-exports-static-semantics-early-errors">
         <h1>Static Semantics: Early Errors</h1>
         <emu-grammar>ExportDeclaration : `export` NamedExports `;`</emu-grammar>
         <ul>
+          <li>
+            It is a Syntax Error if ReferencedBindings of |NamedExports| contains any |StringLiteral|s.
+          </li>
           <li>
             For each |IdentifierName| _n_ in ReferencedBindings of |NamedExports|: It is a Syntax Error if StringValue of _n_ is a |ReservedWord| or if the StringValue of _n_ is one of: *"implements"*, *"interface"*, *"let"*, *"package"*, *"private"*, *"protected"*, *"public"*, or *"static"*.
           </li>
@@ -28237,13 +28275,13 @@
           1. Let _names2_ be the ExportedBindings of |ExportSpecifier|.
           1. Return the list-concatenation of _names1_ and _names2_.
         </emu-alg>
-        <emu-grammar>ExportSpecifier : IdentifierName</emu-grammar>
+        <emu-grammar>ExportSpecifier : ModuleExportName</emu-grammar>
         <emu-alg>
-          1. Return a List whose sole element is the StringValue of |IdentifierName|.
+          1. Return a List whose sole element is the StringValue of |ModuleExportName|.
         </emu-alg>
-        <emu-grammar>ExportSpecifier : IdentifierName `as` IdentifierName</emu-grammar>
+        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
         <emu-alg>
-          1. Return a List whose sole element is the StringValue of the first |IdentifierName|.
+          1. Return a List whose sole element is the StringValue of the first |ModuleExportName|.
         </emu-alg>
       </emu-clause>
 
@@ -28280,9 +28318,9 @@
         <emu-alg>
           1. Return a new empty List.
         </emu-alg>
-        <emu-grammar>ExportFromClause : `*` `as` IdentifierName</emu-grammar>
+        <emu-grammar>ExportFromClause : `*` `as` ModuleExportName</emu-grammar>
         <emu-alg>
-          1. Return a List whose sole element is the StringValue of |IdentifierName|.
+          1. Return a List whose sole element is the StringValue of |ModuleExportName|.
         </emu-alg>
         <emu-grammar>ExportFromClause : NamedExports</emu-grammar>
         <emu-alg>
@@ -28315,13 +28353,13 @@
           1. Let _names2_ be the ExportedNames of |ExportSpecifier|.
           1. Return the list-concatenation of _names1_ and _names2_.
         </emu-alg>
-        <emu-grammar>ExportSpecifier : IdentifierName</emu-grammar>
+        <emu-grammar>ExportSpecifier : ModuleExportName</emu-grammar>
         <emu-alg>
-          1. Return a List whose sole element is the StringValue of |IdentifierName|.
+          1. Return a List whose sole element is the StringValue of |ModuleExportName|.
         </emu-alg>
-        <emu-grammar>ExportSpecifier : IdentifierName `as` IdentifierName</emu-grammar>
+        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
         <emu-alg>
-          1. Return a List whose sole element is the StringValue of the second |IdentifierName|.
+          1. Return a List whose sole element is the StringValue of the second |ModuleExportName|.
         </emu-alg>
       </emu-clause>
 
@@ -28407,9 +28445,9 @@
           1. Let _entry_ be the ExportEntry Record { [[ModuleRequest]]: _module_, [[ImportName]]: ~all-but-default~, [[LocalName]]: *null*, [[ExportName]]: *null* }.
           1. Return a List whose sole element is _entry_.
         </emu-alg>
-        <emu-grammar>ExportFromClause : `*` `as` IdentifierName</emu-grammar>
+        <emu-grammar>ExportFromClause : `*` `as` ModuleExportName</emu-grammar>
         <emu-alg>
-          1. Let _exportName_ be the StringValue of |IdentifierName|.
+          1. Let _exportName_ be the StringValue of |ModuleExportName|.
           1. Let _entry_ be the ExportEntry Record { [[ModuleRequest]]: _module_, [[ImportName]]: ~all~, [[LocalName]]: *null*, [[ExportName]]: _exportName_ }.
           1. Return a List whose sole element is _entry_.
         </emu-alg>
@@ -28423,9 +28461,9 @@
           1. Let _specs2_ be the ExportEntriesForModule of |ExportSpecifier| with argument _module_.
           1. Return the list-concatenation of _specs1_ and _specs2_.
         </emu-alg>
-        <emu-grammar>ExportSpecifier : IdentifierName</emu-grammar>
+        <emu-grammar>ExportSpecifier : ModuleExportName</emu-grammar>
         <emu-alg>
-          1. Let _sourceName_ be the StringValue of |IdentifierName|.
+          1. Let _sourceName_ be the StringValue of |ModuleExportName|.
           1. If _module_ is *null*, then
             1. Let _localName_ be _sourceName_.
             1. Let _importName_ be *null*.
@@ -28434,10 +28472,10 @@
             1. Let _importName_ be _sourceName_.
           1. Return a List whose sole element is the ExportEntry Record { [[ModuleRequest]]: _module_, [[ImportName]]: _importName_, [[LocalName]]: _localName_, [[ExportName]]: _sourceName_ }.
         </emu-alg>
-        <emu-grammar>ExportSpecifier : IdentifierName `as` IdentifierName</emu-grammar>
+        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
         <emu-alg>
-          1. Let _sourceName_ be the StringValue of the first |IdentifierName|.
-          1. Let _exportName_ be the StringValue of the second |IdentifierName|.
+          1. Let _sourceName_ be the StringValue of the first |ModuleExportName|.
+          1. Let _exportName_ be the StringValue of the second |ModuleExportName|.
           1. If _module_ is *null*, then
             1. Let _localName_ be _sourceName_.
             1. Let _importName_ be *null*.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -5638,6 +5638,23 @@ <h1>IsStringPrefix ( _p_, _q_ )</h1>
       </emu-note>
     </emu-clause>
 
+    <emu-clause id="sec-isstringvalidunicode" aoid="IsStringValidUnicode">
+      <h1>Static Semantics: IsStringValidUnicode ( _string_ )</h1>
+      <p>
+        The abstract operation IsStringValidUnicode takes an argument _string_ ( a String ). It interprets _string_ as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, and returns if it is a complete valid UTF-16 sequence. It performs the following steps when called:
```
ljharb, line 5644:
```suggestion
        The abstract operation IsStringValidUnicode takes an argument _string_ (a String). It interprets _string_ as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, and returns if it is a complete valid UTF-16 sequence. It performs the following steps when called:
```

spec.html
```diff
@@ -5638,6 +5638,23 @@ <h1>IsStringPrefix ( _p_, _q_ )</h1>
...
+      </p>
+      <emu-alg>
+        1. Let _strLen_ be the number of code units in string.
+        1. Let _k_ be 0.
+        1. Repeat,
+          1. If _k_ equals _strLen_, return.
```
ljharb, line 5650: should this return true or false?

spec.html
```diff
@@ -21877,9 +21894,9 @@ <h1>Static Semantics: BoundNames</h1>
         AsyncFunctionDeclaration : `async` `function` `(` FormalParameters `)` `{` AsyncFunctionBody `}`
       </emu-grammar>
       <emu-alg>
-        1. Return &laquo; *"\*default\*"* &raquo;.
+        1. Return &laquo; ~default~ &raquo;.
```
ljharb, line 21897: this seems like a good editorial PR to land separately and first :-)

spec.html
```diff
@@ -24783,7 +24821,7 @@ <h2>Syntax</h2>
 
         ExportFromClause :
           `*`
-          `*` `as` IdentifierName
+          `*` `as` ExportSpecifier
```
devsnek, line 24824: this allows `* as a as "b"`

spec.html
bmeck: I'm only doing this here because it needs to be done for this PR, feel free to rip it out to a separate PR.

spec.html
bmeck: Ah, good catch

bmeck:
```suggestion
          `*` `as` ModuleExportName
```
?

spec.html
bmeck: ah, that should be breaking out of the loop, it should make the whole algorithm return `true`

ljharb: In that case, I'd expect `Repeat, while _k_ does not equal _strLen_,` or similar.

bmeck: should be fixed

spec.html
bmeck: should be fixed

spec.html
ljharb: thanks, this works, but i think it'd be clearer with a `return true` at the end, and the condition as part of the `repeat`

bmeck: redone

spec.html
```diff
@@ -23063,6 +23083,12 @@ <h1>Static Semantics: Early Errors</h1>
         <emu-note>
           <p>The duplicate ExportedNames rule implies that multiple `export default` |ExportDeclaration| items within a |ModuleBody| is a Syntax Error. Additional error conditions relating to conflicting or duplicate declarations are checked during module linking prior to evaluation of a |Module|. If any such errors are detected the |Module| is not evaluated.</p>
         </emu-note>
+
+        <h1>Static Semantics: Early Errors</h1>
+        <emu-grammar>ModuleExportName : StringLiteral</emu-grammar>
+        <ul>
+          <li>It is a Syntax Error if IsStringValidUnicode(StringLiteral) is *false*.</li>
```
devsnek, line 23090:
```suggestion
          <li>It is a Syntax Error if IsStringValidUnicode(StringValue(StringLiteral)) is *false*.</li>
```

spec.html
ljharb: Went ahead and filed #2155.

spec.html
```diff
@@ -25176,9 +25242,9 @@ <h1>Runtime Semantics: Evaluation</h1>
         <emu-alg>
           1. Let _value_ be ? BindingClassDeclarationEvaluation of |ClassDeclaration|.
           1. Let _className_ be the sole element of BoundNames of |ClassDeclaration|.
-          1. If _className_ is *"\*default\*"*, then
+          1. If _className_ is ~default~, then
             1. Let _env_ be the running execution context's LexicalEnvironment.
-            1. Perform ? InitializeBoundName(*"\*default\*"*, _value_, _env_).
+            1. Perform ? InitializeBoundName(~default~, _value_, _env_).
```
devsnek, line 25247: `InitializeBoundName` only accepts strings

spec.html
ljharb: Thanks! now this line isn't needed anymore tho:
```suggestion
```

bmeck:
Updated, now depends on #2155

spec.html
```diff
@@ -5638,6 +5638,22 @@ <h1>IsStringPrefix ( _p_, _q_ )</h1>
       </emu-note>
     </emu-clause>
 
+    <emu-clause id="sec-isstringvalidunicode" aoid="IsStringValidUnicode">
+      <h1>Static Semantics: IsStringValidUnicode ( _string_ )</h1>
+      <p>
+        The abstract operation IsStringValidUnicode takes an argument _string_ (a String). It interprets _string_ as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, and returns if it is a complete valid UTF-16 sequence. It performs the following steps when called:
```
bakkot, line 5644:
```suggestion
        The abstract operation IsStringValidUnicode takes an argument _string_ (a String). It interprets _string_ as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, and determines whether it is a complete valid UTF-16 sequence. It performs the following steps when called:
```

spec.html
```diff
@@ -5638,6 +5638,22 @@ <h1>IsStringPrefix ( _p_, _q_ )</h1>
...
+      </p>
+      <emu-alg>
+        1. Let _strLen_ be the number of code units in string.
```
bakkot, line 5647:
```suggestion
        1. Let _strLen_ be the number of code units in _string_.
```

spec.html
```diff
@@ -23063,6 +23082,12 @@ <h1>Static Semantics: Early Errors</h1>
         <emu-note>
           <p>The duplicate ExportedNames rule implies that multiple `export default` |ExportDeclaration| items within a |ModuleBody| is a Syntax Error. Additional error conditions relating to conflicting or duplicate declarations are checked during module linking prior to evaluation of a |Module|. If any such errors are detected the |Module| is not evaluated.</p>
         </emu-note>
+
+        <h1>Static Semantics: Early Errors</h1>
+        <emu-grammar>ModuleExportName : StringLiteral</emu-grammar>
+        <ul>
+          <li>It is a Syntax Error if IsStringValidUnicode(StringLiteral) is *false*.</li>
```
bakkot, line 23089:
```suggestion
          <li>It is a Syntax Error if IsStringValidUnicode of the StringValue of StringLiteral is *false*.</li>
```

spec.html
```diff
@@ -5638,6 +5638,22 @@ <h1>IsStringPrefix ( _p_, _q_ )</h1>
       </emu-note>
     </emu-clause>
 
+    <emu-clause id="sec-isstringvalidunicode" aoid="IsStringValidUnicode">
+      <h1>Static Semantics: IsStringValidUnicode ( _string_ )</h1>
```
gibson042, line 5642: The Unicode concept is "well-formed", and note that it applies to sequences of code _units_ purporting to be in a particular encoding form: http://www.unicode.org/glossary/#well_formed_code_unit_sequence

JLHwung:
Should we allow `{ ModuleExportName as ModuleExportName }` in `ExportFromtClause`?

The goal example
```js
import {"\0 any unicode" as foo} from "";
export {foo as "\0 any unicode"};
```
reminds me of the following one-liner:

```js
export {"\0 any unicode" as "\0 any unicode"} from "some-module";
```

ExE-Boss:
It should also be possible to do:
```js
export { "\0 any unicode" } from "some-module";
```

spec.html
```diff
@@ -24798,6 +24836,7 @@ <h2>Syntax</h2>
         ExportSpecifier :
           IdentifierName
           IdentifierName `as` IdentifierName
+          IdentifierName `as` ModuleExportName
```
ExE-Boss, line 24839: The following change will address both <https://github.com/tc39/ecma262/pull/2154#issuecomment-696362312> and <https://github.com/tc39/ecma262/pull/2154#issuecomment-696379292>:

```suggestion
        ExportSpecifier :
          IdentifierName
          IdentifierName `as` IdentifierName
          IdentifierName `as` ModuleExportName
          ModuleExportName
          ModuleExportName `as` ModuleExportName
```

---

Also, remember to add an early error for when the following is not a part of an `ExportFromClause`:

```emu-grammar
ExportSpecifier :
  ModuleExportName
  ModuleExportName `as` ModuleExportName
```

spec.html
```diff
@@ -25148,6 +25247,10 @@ <h1>Static Semantics: ReferencedBindings</h1>
         <emu-alg>
           1. Return a List containing the first |IdentifierName|.
         </emu-alg>
+        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
```
ExE-Boss, line 25250: This also needs to handle the non‑renaming `ExportSpecifier : ModuleExportName` shorthand.

```suggestion
        <emu-grammar>
          ExportSpecifier :
            ModuleExportName
            ModuleExportName `as` ModuleExportName
        </emu-grammar>
```

spec.html
```diff
@@ -25247,7 +25247,10 @@ <h1>Static Semantics: ReferencedBindings</h1>
         <emu-alg>
           1. Return a List containing the first |IdentifierName|.
         </emu-alg>
-        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
+        <emu-grammar>
+          ExportSpecifier :
+            ModuleExportName
+            ModuleExportName `as` ModuleExportName</emu-grammar>
```
ExE-Boss, line 25253: This is missing a newline:
```suggestion
            ModuleExportName `as` ModuleExportName
          </emu-grammar>
```

bmeck:
Made a tests PR at https://github.com/tc39/test262/pull/2836

devsnek:
Isn't choosing between `+From` and `~From` forms impossible without infinite lookahead?

bmeck:
@devsnek ah, yes that would be true due to `from` being in a trailing position, we could convert to a cover grammar I think?

bmeck:
alternatively we could remove the `{ "x" }` form that only works in a `export ... from` form that would get rid of the issue.

devsnek:
Cover grammar seems unnecessary. I would just add an early error saying "it is an early syntax error if `export { 'x' };` happens"

spec.html
```diff
@@ -24798,12 +24836,18 @@ <h2>Syntax</h2>
         ExportSpecifier :
           IdentifierName
           IdentifierName `as` IdentifierName
+          IdentifierName `as` ModuleExportName
+          ModuleExportName
+          ModuleExportName `as` ModuleExportName
```
ExE-Boss, line 24841: It should also be possible to rename `ModuleExportName` to `IdentifierName` when re‑exporting:
```suggestion
          ModuleExportName `as` IdentifierName
          ModuleExportName `as` ModuleExportName
```

ExE-Boss, line 24841:
```suggestion
          ModuleExportName `as` IdentifierName
          ModuleExportName `as` ModuleExportName
```

spec.html
```diff
@@ -24900,6 +24944,14 @@ <h1>Static Semantics: ExportedBindings</h1>
...
           1. Return a List containing the StringValue of the first |IdentifierName|.
         </emu-alg>
+        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
+        <emu-alg>
+          1. Return a List containing the StringValue of the first |ModuleExportName|.
+        </emu-alg>
+        <emu-grammar>ExportSpecifier : ModuleExportName</emu-grammar>
+        <emu-alg>
+          1. Return a List containing the StringValue of |ModuleExportName|.
+        </emu-alg>
```
ExE-Boss, line 24954: This should also be sorted:
```suggestion
        <emu-grammar>
          ExportSpecifier :
            ModuleExportName
            ModuleExportName `as` IdentifierName
        </emu-grammar>
        <emu-alg>
          1. Return a List containing the StringValue of |ModuleExportName|.
        </emu-alg>
        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
        <emu-alg>
          1. Return a List containing the StringValue of the first |ModuleExportName|.
        </emu-alg>
```

spec.html
```diff
@@ -24957,6 +25013,10 @@ <h1>Static Semantics: ExportedNames</h1>
         <emu-alg>
           1. Return a List containing the StringValue of the second |IdentifierName|.
         </emu-alg>
+        <emu-grammar>ExportSpecifier : IdentifierName `as` ModuleExportName</emu-grammar>
+        <emu-alg>
+          1. Return a List containing the StringValue of ModuleExportName.
+        </emu-alg>
```
ExE-Boss, line 25019:
```suggestion
        <emu-grammar>
          ExportSpecifier :
            IdentifierName `as` ModuleExportName
            ModuleExportName
        </emu-grammar>
        <emu-alg>
          1. Return a List containing the StringValue of |ModuleExportName|.
        </emu-alg>
        <emu-grammar>ExportSpecifier : ModuleExportName `as` IdentifierName</emu-grammar>
        <emu-alg>
          1. Return a List containing the StringValue of |IdentifierName|.
        </emu-alg>
        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
        <emu-alg>
          1. Return a List containing the StringValue of the second |ModuleExportName|.
        </emu-alg>
```

spec.html
```diff
@@ -25056,6 +25122,42 @@ <h1>Static Semantics: ExportEntriesForModule</h1>
...
+          1. Let _sourceName_ be the StringValue of |ModuleExportName|.
+          1. Let _exportName_ be the StringValue of |ModuleExportName|.
+          1. If _module_ is *null*, then
+            1. Let _localName_ be _name_.
+            1. Let _importName_ be *null*.
+          1. Else,
+            1. Let _localName_ be *null*.
+            1. Let _importName_ be _name_.
+          1. Return a new List containing the ExportEntry Record { [[ModuleRequest]]: _module_, [[ImportName]]: _importName_, [[LocalName]]: _localName_, [[ExportName]]: _exportName_ }.
+        </emu-alg>
```
ExE-Boss, line 25160:
```suggestion
        <emu-grammar>ExportSpecifier : ModuleExportName</emu-grammar>
        <emu-alg>
          1. Let _sourceName_ be the StringValue of |ModuleExportName|.
          1. If _module_ is *null*, then
            1. Let _localName_ be _sourceName_.
            1. Let _importName_ be *null*.
          1. Else,
            1. Let _localName_ be *null*.
            1. Let _importName_ be _sourceName_.
          1. Return a new List containing the ExportEntry Record { [[ModuleRequest]]: _module_, [[ImportName]]: _importName_, [[LocalName]]: _localName_, [[ExportName]]: _sourceName_ }.
        </emu-alg>
        <emu-grammar>ExportSpecifier : ModuleExportName `as` IdentifierName</emu-grammar>
        <emu-alg>
          1. Let _sourceName_ be the StringValue of |ModuleExportName|.
          1. Let _exportName_ be the StringValue of |IdentifierName|.
          1. If _module_ is *null*, then
            1. Let _localName_ be _sourceName_.
            1. Let _importName_ be *null*.
          1. Else,
            1. Let _localName_ be *null*.
            1. Let _importName_ be _sourceName_.
          1. Return a new List containing the ExportEntry Record { [[ModuleRequest]]: _module_, [[ImportName]]: _importName_, [[LocalName]]: _localName_, [[ExportName]]: _exportName_ }.
        </emu-alg>
        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
        <emu-alg>
          1. Let _sourceName_ be the StringValue of the first |ModuleExportName|.
          1. Let _exportName_ be the StringValue of the second |ModuleExportName|.
          1. If _module_ is *null*, then
            1. Let _localName_ be _sourceName_.
            1. Let _importName_ be *null*.
          1. Else,
            1. Let _localName_ be *null*.
            1. Let _importName_ be _sourceName_.
          1. Return a new List containing the ExportEntry Record { [[ModuleRequest]]: _module_, [[ImportName]]: _importName_, [[LocalName]]: _localName_, [[ExportName]]: _exportName_ }.
        </emu-alg>
```

spec.html
```diff
@@ -25148,6 +25250,18 @@ <h1>Static Semantics: ReferencedBindings</h1>
...
+          1. Return a List containing the |IdentifierName|.
+        </emu-alg>
+        <emu-grammar>
+          ExportSpecifier :
+            ModuleExportName
+            ModuleExportName `as` ModuleExportName
+        </emu-grammar>
+        <emu-alg>
+          1. Return a List containing the first |ModuleExportName|.
+        </emu-alg>
```
ExE-Boss, line 25264:
```suggestion
        <emu-grammar>
          ExportSpecifier :
            ModuleExportName
            ModuleExportName `as` IdentifierName
        </emu-grammar>
        <emu-alg>
          1. Return a List containing the |ModuleExportName|.
        </emu-alg>
        <emu-grammar>ExportSpecifier : ModuleExportName `as` ModuleExportName</emu-grammar>
        <emu-alg>
          1. Return a List containing the first |ModuleExportName|.
        </emu-alg>
```

ExE-Boss (commented):
It should also be possible to rename `ModuleExportName` to `IdentifierName` when re‑exporting:

Qawsedrf0501:
H

loksta213 (commented):
[spec.html](https://blog.twitter.com/developer/en_us.html)

nicolo-ribaudo:
For the `[needs implementation]` label: this now works in Node.js 16 (and thus Chrome 90 I believe).

ljharb:
@bmeck there's some outstanding review items, and it's still blocked on #2155, which i'm a bit stuck on.

Help on #2155 (rebasing it and accounting for review comments) would be appreciated, after which this PR could be rebased, and whether it has sufficient implementations can be re-evaulated.

spec.html
```diff
@@ -5638,6 +5638,22 @@ <h1>IsStringPrefix ( _p_, _q_ )</h1>
       </emu-note>
     </emu-clause>
 
+    <emu-clause id="sec-isstringwellformedunicode" aoid="IsStringWellFormedUnicode">
+      <h1>Static Semantics: IsStringWellFormedUnicode ( _string_ )</h1>
+      <p>
```
syg, line 5643: This needs a structured header now.

syg:
Note that this remains blocked on #2155.

bergus:
Here's where the  came from: https://github.com/tc39/notes/blob/master/meetings/2020-06/june-2.md#arbitrary-module-namespace-names

bakkot:
Rebased and pushed a commit addressing comments from @ExE-Boss.

That said, I think this can (and should) be substantially simplified by having a single |ExportedBinding| nonterminal which produces both |IdentifierName| and |StringLiteral|. I'll get to that at some point.

spec.html
```diff
@@ -6065,6 +6065,27 @@ <h1>
       </emu-note>
     </emu-clause>
 
+    <emu-clause id="sec-isstringwellformedunicode" type="abstract operation">
+      <h1>
+        Static Semantics: IsStringWellFormedUnicode (
```
arhadthedev, line 6070: An unpaired surrogate makes an invalid UTF-16 string not only in module export names, but everywhere in source code. I think that IsStringWellFormedUnicode should be called either from |DoubleStringCharacters|-|SingleStringCharacters|-|IdentifierName|, or, better, reimplemented as a lookahead inside |SourceCharacter|.

bakkot: No, in JavaScript you can have ill-formed UTF-16 in any other string - for example, the source text consisting of the three code units U+0022 U+D800 U+0022 is a legal program.

It really is only in module export names where we intend to apply this restriction.

bakkot:
OK, refactored per [above comment](https://github.com/tc39/ecma262/pull/2154#issuecomment-962295962), though I called the new nonterminal |ModuleExportName|. I think this is ready to go now (pending review).

spec.html
arhadthedev: > No, in JavaScript you can have ill-formed UTF-16 in any other string - for example, the source text consisting of the three code units U+0022 U+D800 U+0022 is a legal program.

Formally yes, but how can it be (ab)used in practice without low-level work with already encoded UTF-16 byte stream? I believe it can cause more trouble than use because it may be a sign that an input sequence is damaged somehow or tampered by some exploit of a transport protocol.

arhadthedev: Addition: I'm aware of `All Unicode code point values from U+0000 to U+10FFFF, including surrogate code points` in #sec-source-text. I just express my concern about its pros and cons.

bakkot: That's really not relevant to this PR. If you'd like to make a proposal for changing the language, see [CONTRIBUTING.md](https://github.com/tc39/ecma262/blob/3da56831eaa4c8029d0c69445d5099ce6c2b0c5e/CONTRIBUTING.md), though I note that you are unlikely to have much luck pursuing that particular change.

spec.html
```diff
@@ -6065,6 +6065,27 @@ <h1>
...
+        )
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It interprets _string_ as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, and determines whether it is a <a href="http://www.unicode.org/glossary/#well_formed_code_unit_sequence">well formed</a> UTF-16 sequence.</dd>
+      </dl>
+      <emu-alg>
+        1. Let _strLen_ be the number of code units in _string_.
+        1. Let _k_ be 0.
+        1. Repeat, while _k_ does not equal _strLen_,
```
jmdyck, line 6081:
```suggestion
        1. Repeat, while _k_ &ne; _strLen_,
```

spec.html
```diff
@@ -25977,6 +26008,12 @@ <h1>Static Semantics: Early Errors</h1>
         <emu-note>
           <p>The duplicate ExportedNames rule implies that multiple `export default` |ExportDeclaration| items within a |ModuleBody| is a Syntax Error. Additional error conditions relating to conflicting or duplicate declarations are checked during module linking prior to evaluation of a |Module|. If any such errors are detected the |Module| is not evaluated.</p>
         </emu-note>
+
+        <h1>Static Semantics: Early Errors</h1>
```
jmdyck, line 26012: Delete this mid-clause heading?

spec.html
```diff
@@ -25977,6 +26008,12 @@ <h1>Static Semantics: Early Errors</h1>
...
+        <emu-grammar>ModuleExportName : StringLiteral</emu-grammar>
+        <ul>
+          <li>It is a Syntax Error if IsStringWellFormedUnicode of the SV of |StringLiteral| is *false*.</li>
```
jmdyck, line 26015:
```suggestion
          <li>It is a Syntax Error if IsStringWellFormedUnicode(the SV of |StringLiteral|) is *false*.</li>
```
It's unusual to invoke an AO (IsStringWellFormedUnicode) with the `<name> of <arg>` syntax instead of `<name>(<arg>)`. (And yeah, it's also unusual to have an SDO-invocation as an arg. I'm not sure which is weirder.)

jmdyck (dismissed):

michaelficarra (approved):

syg (approved):
lgtm

bakkot:
Sorry for the delay on getting this landed, all!

PR #2158, opened 2020-08-29
Editorial: Remove notes on well-formed rest arguments

shvaikalesh:
#2139 clarified rest parameters, making these notes unnecessary.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33651,9 +33651,6 @@ THH:mm:ss.sss
           1. Return _A_.
         </emu-alg>
         <emu-note>
-          <p>The _items_ argument is assumed to be a well-formed rest argument value.</p>
-        </emu-note>
-        <emu-note>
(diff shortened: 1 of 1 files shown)


jmdyck:
> It is not at all clear to me what these notes were trying to say in the first place, but they certainly do not seem to be necessary now.

When (e.g.) `Array.of ( ..._items_ )` was added to the spec (in an early draft of ES6), the first step of the algorithm was:
```
Let _lenValue_ be the result of calling the [[Get]] internal method of _items_ with the argument *"length"*.
```
That is, for the purpose of the algorithm, `_items_` was deemed to be an array-like object. So it made sense to have a note clarifying its nature (although there wasn't a definition of "well-formed rest argument value" yet).

In a later draft of of ES6, this step changed to:
```
Let _items_ be the List of arguments passed to this function.
```
That was probably the point at which the Note became obsolete.

PR #2160, opened 2020-08-30
Editorial: Add note on ignored argument of Symbol.prototype[@@toPrimitive]

shvaikalesh:
Aligns it with [`Date.prototype.toJSON`](https://tc39.es/ecma262/#sec-date.prototype.tojson) and removes incorrect "allowed values" note.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -26704,13 +26704,16 @@
 
       <emu-clause id="sec-symbol.prototype-@@toprimitive">
         <h1>Symbol.prototype [ @@toPrimitive ] ( _hint_ )</h1>
-        <p>This function is called by ECMAScript language operators to convert a Symbol object to a primitive value. The allowed values for _hint_ are *"default"*, *"number"*, and *"string"*.</p>
+        <p>This function is called by ECMAScript language operators to convert a Symbol object to a primitive value.</p>
         <p>When the `@@toPrimitive` method is called with argument _hint_, the following steps are taken:</p>
         <emu-alg>
           1. Return ? thisSymbolValue(*this* value).
         </emu-alg>
         <p>The value of the *"name"* property of this function is *"[Symbol.toPrimitive]"*.</p>
         <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
+        <emu-note>
+          <p>The argument is ignored.</p>
+        </emu-note>
       </emu-clause>
 
       <emu-clause id="sec-symbol.prototype-@@tostringtag">


ExE-Boss:
The specification of the `%Symbol.toPrimitive%` well‑known symbol is that the <code><var>hint</var></code> parameter is expected to be one of `"default"`, `"number"`, or `"string"`.

bathos:
@ExE-Boss It appears that `Symbol.prototype[Symbol.toPrimitive]` is already specced to ignore the hint arg entirely, unlike `Date.prototype[Symbol.toPrimitive]` (which throws if it’s not given a valid hint), so I think this editorial update is removing misleading noise.

ljharb:
It kind of seems like there should be a normative PR to make `Symbol.prototype[Symbol.toPrimitive]` throw when it receives an invalid hint - iow, the runtime exception is the bug, not the redundant note?

michaelficarra (approved):

ljharb (approved):
Since it's unlikely there will be an appetite for a normative change here (either making a bad hint throw, or, removing the hint arg entirely and changing the length), this is fine.

shvaikalesh:
I'm up for submitting a normative PR + tests + JSC patch for throwing on a bad hint, but it might not be web-compatible if there is code like `symbol[Symbol.toPrimitive]()`.

> removing the hint arg entirely and changing the length

Isn't having the `"length"` of `1` while ignoring the argument prevents implementors from extending this method? Pretty much like `Object.prototype.toLocaleString ( [ reserved1 [ , reserved2 ] ] )`.

bakkot:
I'd prefer not to change the behavior. I don't think there's much user benefit to throwing; certainly not enough to justify the runtime overhead and web-compat risk.

ExE-Boss:
> I'm up for submitting a normative PR + tests + JSC patch for throwing on a bad hint, but it might not be web-compatible if there is code like `symbol[Symbol.toPrimitive]()`.

We could make that specific case behave the same as `hint === "default"`.

ljharb:
At that point we'd lack the consistency angle :-/

bakkot (approved):

dabonneur (changes requested):
![Uploading 93009981...]

dabonneur:
> ![Uploading 93009981...]

dabonneur:
stop please

dabonneur (changes requested):
image

dabonneur (changes requested):
jjb

dabonneur (changes requested):
[](https://github.com/tc39/ecma262/pull/2160)

dabonneur:
####-

dabonneur (changes requested):
ok

dabonneur (changes requested):
-

spec.html
```diff
@@ -27204,13 +27204,16 @@ <h1>Symbol.prototype.valueOf ( )</h1>
...
         <p>When the `@@toPrimitive` method is called with argument _hint_, the following steps are taken:</p>
         <emu-alg>
           1. Return ? thisSymbolValue(*this* value).
         </emu-alg>
         <p>The value of the *"name"* property of this function is *"[Symbol.toPrimitive]"*.</p>
         <p>This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.</p>
+        <emu-note>
+          <p>The argument is ignored.</p>
+        </emu-note>
       </emu-clause>
```
dabonneur, line 27217: an[](https://github.com/tc39/ecma262/pull/2160)

dabonneur (approved):

PR #2163, opened 2020-09-03
Meta: add "Require Allow Edits" action

ljharb:
This adds a github action to require "allow edits" to be checked.

I've created the PR while unchecking the box, so i can get a screenshot of it failing, and then i'll check it and rerun it, and then update the OP with the comparison.

---

Screen Shot 2020-09-03 at 10 55 15 AM
Screen Shot 2020-09-03 at 10 55 26 AM
Screen Shot 2020-09-03 at 10 56 26 AM

---

Note that the job has to be manually rerun (by rebasing or pushing to the PR, one of us clicking "rerun job" on the action, or by closing/reopening the PR); when I figure out how to make it automatically rerun when the box is checked, I'll update it to do so. imo this is Fine, though, because the benefit is automatically signaling to the PR author that they need to check the box.

### DIFF
--- a/.github/workflows/require-allow-edits.yml
+++ b/.github/workflows/require-allow-edits.yml
@@ -1,0 +1,14 @@
+name: Require “Allow Edits”
+
+on: [pull_request]
+
+jobs:
+  _:
+    name: "Require “Allow Edits”"
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: ljharb/require-allow-edits@main
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,4 +3,6 @@ If you are changing the signature or behavior of an existing construct, please c
 
 * [Web IDL](https://heycam.github.io/webidl/) — [file an issue](https://github.com/heycam/webidl/issues/new)
 * [HTML Standard](https://html.spec.whatwg.org/) — [file an issue](https://github.com/whatwg/html/issues/new)
+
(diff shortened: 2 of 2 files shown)


bakkot (approved):
Sure.

.github/workflows/require-allow-edits.yml
```diff
@@ -0,0 +1,14 @@
...
+on: [pull_request]
+
+jobs:
+  _:
+    name: "Require “Allow Edits”"
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: ljharb/require-allow-edits@main
```
michaelficarra, line 12: Is it typical to use a branch name instead of a SHA? Doesn't this allow you to arbitrarily change the functionality without a PR here? I don't know if I like that.

ljharb: It's very typical, yes; it does allow that (it's how bugfixes and improvements can be landed automatically); it's totally fine to use a sha if you're concerned about it.

.github/workflows/require-allow-edits.yml
```diff
@@ -0,0 +1,14 @@
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
septs, line 14: use new version, remove this line?

.github/workflows/require-allow-edits.yml
codehag: This is a bit weird given what we have said elsewhere. Since it is a private repository rather than a TC39 repository, can we use a SHA?

ljharb: @codehag it's a repository owned by a current editor, and this repo is owned by the editors. When the action isn't owned by a current editor, we'd immediately update this to use the SHA instead of the branch.

PR #2164, opened 2020-09-04
Normative: Align detached buffer semantics with web reality

### LINKED ISSUE #678: detached ArrayBuffers are specced to throw often, but implementations do not
Previously discussed several times. @annevk brought it up 2 years ago in https://esdiscuss.org/topic/arraybuffer-neutering.

My understanding is that TC39 has decided to hope that one day implementations start throwing here. I am not aware of any implementation plans to do so, but maybe my info is out of date.

In any case, it's worth having an open tracking issue under "web reality" so that people trying to implement the ES spec realize that it doesn't match the reality of the web and implementations.

rkirsling:
Resolves #678.

Specifically, web reality is such that IsDetachedBuffer checks should not cause the following to throw:
- `ArrayBuffer.prototype.byteLength`
- [[HasProperty]] for typed arrays
- IntegerIndexedElementGet (i.e. [[GetOwnProperty]] and [[Get]] for typed arrays)
- IntegerIndexedElementSet (i.e. [[DefineOwnProperty]] and [[Set]] for typed arrays)

To ensure that we don't violate EIM invariants for typed arrays, we must also:
- Make integer-indexed elements [[Configurable]]
- Override [[Delete]] such that `false` is returned upon attempted deletion of an integer-indexed element

(CC @syg, @rwaldron)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9139,7 +9139,7 @@
       <h1>Integer-Indexed Exotic Objects</h1>
       <p>An Integer-Indexed exotic object is an exotic object that performs special handling of integer index property keys.</p>
       <p><emu-xref href="#integer-indexed-exotic-object">Integer-Indexed exotic objects</emu-xref> have the same internal slots as ordinary objects and additionally [[ViewedArrayBuffer]], [[ArrayLength]], [[ByteOffset]], [[ContentType]], and [[TypedArrayName]] internal slots.</p>
-      <p>An object is an <dfn id="integer-indexed-exotic-object">Integer-Indexed exotic object</dfn> if its [[GetOwnProperty]], [[HasProperty]], [[DefineOwnProperty]], [[Get]], [[Set]], and [[OwnPropertyKeys]] internal methods use the definitions in this section, and its other essential internal methods use the definitions found in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>. These methods are installed by IntegerIndexedObjectCreate.</p>
+      <p>An object is an <dfn id="integer-indexed-exotic-object">Integer-Indexed exotic object</dfn> if its [[GetOwnProperty]], [[HasProperty]], [[DefineOwnProperty]], [[Get]], [[Set]], [[Delete]], and [[OwnPropertyKeys]] internal methods use the definitions in this section, and its other essential internal methods use the definitions found in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>. These methods are installed by IntegerIndexedObjectCreate.</p>
 
       <emu-clause id="sec-integer-indexed-exotic-objects-getownproperty-p">
         <h1>[[GetOwnProperty]] ( _P_ )</h1>
@@ -9150,9 +9150,9 @@
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
-              1. Let _value_ be ? IntegerIndexedElementGet(_O_, _numericIndex_).
+              1. Let _value_ be ! IntegerIndexedElementGet(_O_, _numericIndex_).
               1. If _value_ is *undefined*, return *undefined*.
-              1. Return the PropertyDescriptor { [[Value]]: _value_, [[Writable]]: *true*, [[Enumerable]]: *true*, [[Configurable]]: *false* }.
+              1. Return the PropertyDescriptor { [[Value]]: _value_, [[Writable]]: *true*, [[Enumerable]]: *true*, [[Configurable]]: *true* }.
           1. Return OrdinaryGetOwnProperty(_O_, _P_).
         </emu-alg>
       </emu-clause>
@@ -9167,7 +9167,7 @@
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
               1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
-              1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
+              1. If IsDetachedBuffer(_buffer_) is *true*, return *false*.
               1. If ! IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *false*.
               1. Return *true*.
           1. Return ? OrdinaryHasProperty(_O_, _P_).
@@ -9185,7 +9185,7 @@
             1. If _numericIndex_ is not *undefined*, then
               1. If ! IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *false*.
               1. If IsAccessorDescriptor(_Desc_) is *true*, return *false*.
-              1. If _Desc_ has a [[Configurable]] field and if _Desc_.[[Configurable]] is *true*, return *false*.
+              1. If _Desc_ has a [[Configurable]] field and if _Desc_.[[Configurable]] is *false*, return *false*.
               1. If _Desc_ has an [[Enumerable]] field and if _Desc_.[[Enumerable]] is *false*, return *false*.
               1. If _Desc_ has a [[Writable]] field and if _Desc_.[[Writable]] is *false*, return *false*.
               1. If _Desc_ has a [[Value]] field, then
@@ -9204,7 +9204,7 @@
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
-              1. Return ? IntegerIndexedElementGet(_O_, _numericIndex_).
+              1. Return ! IntegerIndexedElementGet(_O_, _numericIndex_).
           1. Return ? OrdinaryGet(_O_, _P_, _Receiver_).
         </emu-alg>
       </emu-clause>
@@ -9219,6 +9219,22 @@
             1. If _numericIndex_ is not *undefined*, then
               1. Return ? IntegerIndexedElementSet(_O_, _numericIndex_, _V_).
           1. Return ? OrdinarySet(_O_, _P_, _V_, _Receiver_).
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-integer-indexed-exotic-objects-delete-p">
+        <h1>[[Delete]] ( _P_ )</h1>
+        <p>When the [[Delete]] internal method of an Integer-Indexed exotic object _O_ is called with property key _P_, the following steps are taken:</p>
+        <emu-alg>
+          1. Assert: IsPropertyKey(_P_) is *true*.
+          1. Assert: _O_ is an Integer-Indexed exotic object.
+          1. If Type(_P_) is String, then
+            1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
+            1. If _numericIndex_ is not *undefined*, then
+              1. If IsDetachedBuffer(_O_.[[ViewedArrayBuffer]]) is *true*, return *true*.
+              1. If ! IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *true*.
+              1. Return *false*.
+          1. Return ? OrdinaryDelete(_O_, _P_).
         </emu-alg>
       </emu-clause>
 
@@ -9250,6 +9266,7 @@
           1. Set _A_.[[DefineOwnProperty]] as specified in <emu-xref href="#sec-integer-indexed-exotic-objects-defineownproperty-p-desc"></emu-xref>.
           1. Set _A_.[[Get]] as specified in <emu-xref href="#sec-integer-indexed-exotic-objects-get-p-receiver"></emu-xref>.
           1. Set _A_.[[Set]] as specified in <emu-xref href="#sec-integer-indexed-exotic-objects-set-p-v-receiver"></emu-xref>.
+          1. Set _A_.[[Delete]] as specified in <emu-xref href="#sec-integer-indexed-exotic-objects-delete-p"></emu-xref>.
           1. Set _A_.[[OwnPropertyKeys]] as specified in <emu-xref href="#sec-integer-indexed-exotic-objects-ownpropertykeys"></emu-xref>.
           1. Set _A_.[[Prototype]] to _prototype_.
           1. Return _A_.
@@ -9274,7 +9291,7 @@
         <emu-alg>
           1. Assert: _O_ is an Integer-Indexed exotic object.
           1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
-          1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
+          1. If IsDetachedBuffer(_buffer_) is *true*, return *undefined*.
           1. If ! IsValidIntegerIndex(_O_, _index_) is *false*, return *undefined*.
           1. Let _offset_ be _O_.[[ByteOffset]].
           1. Let _arrayTypeName_ be the String value of _O_.[[TypedArrayName]].
@@ -9293,7 +9310,7 @@
           1. If _O_.[[ContentType]] is ~BigInt~, let _numValue_ be ? ToBigInt(_value_).
           1. Otherwise, let _numValue_ be ? ToNumber(_value_).
           1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
-          1. If IsDetachedBuffer(_buffer_) is *true*, throw a *TypeError* exception.
+          1. If IsDetachedBuffer(_buffer_) is *true*, return *false*.
           1. If ! IsValidIntegerIndex(_O_, _index_) is *false*, return *false*.
           1. Let _offset_ be _O_.[[ByteOffset]].
           1. Let _arrayTypeName_ be the String value of _O_.[[TypedArrayName]].
@@ -34610,7 +34627,7 @@ THH:mm:ss.sss
           1. Let _captured_ be 0.
           1. Repeat, while _k_ &lt; _len_,
             1. Let _Pk_ be ! ToString(𝔽(_k_)).
-            1. Let _kValue_ be ? Get(_O_, _Pk_).
+            1. Let _kValue_ be ! Get(_O_, _Pk_).
             1. Let _selected_ be ! ToBoolean(? Call(_callbackfn_, _thisArg_, &laquo; _kValue_, 𝔽(_k_), _O_ &raquo;)).
             1. If _selected_ is *true*, then
               1. Append _kValue_ to the end of _kept_.
@@ -34706,7 +34723,7 @@ THH:mm:ss.sss
           1. Let _k_ be 0.
           1. Repeat, while _k_ &lt; _len_,
             1. Let _Pk_ be ! ToString(𝔽(_k_)).
-            1. Let _kValue_ be ? Get(_O_, _Pk_).
+            1. Let _kValue_ be ! Get(_O_, _Pk_).
             1. Let _mappedValue_ be ? Call(_callbackfn_, _thisArg_, &laquo; _kValue_, 𝔽(_k_), _O_ &raquo;).
             1. Perform ? Set(_A_, _Pk_, _mappedValue_, *true*).
             1. Set _k_ to _k_ + 1.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -36833,7 +36833,7 @@ <h1>get ArrayBuffer.prototype.byteLength</h1>
           1. Let _O_ be the *this* value.
           1. Perform ? RequireInternalSlot(_O_, [[ArrayBufferData]]).
           1. If IsSharedArrayBuffer(_O_) is *true*, throw a *TypeError* exception.
-          1. If IsDetachedBuffer(_O_) is *true*, throw a *TypeError* exception.
+          1. If IsDetachedBuffer(_O_) is *true*, return 0.
```
ljharb, line 36836: this isn't a math value, it's a Number; not sure if this is fine now and can wait for #2007, or if this should be:
```suggestion
          1. If IsDetachedBuffer(_O_) is *true*, return *+0*.
```

ljharb (approved):

spec.html
michaelficarra: We'll probably have #2007 merged before this reaches consensus, so I'd say it's best to match the conventions from there. But feel free to wait until it lands.

anba:
This change will also require updates to the TypedArray built-ins:
1. Some operations are no longer fallible and should be updated to use the `!` prefix instead of `?`.
2. Previous implicit detached ArrayBuffer checks now need to be explicit, for example in `%TypedArray%.prototype.slice()` to ensure we get the same behaviour for both code-paths (steps 14-15).
    - `slice()` might be a special-case, but it probably doesn't hurt to verify this for all built-ins and compare against actual implementations.

spec.html
rkirsling: Note though, that #2007 doesn't currently change the existing cases of `If IsDetachedBuffer(_buffer_) is *true*, return 0.` in %TypedArray%.prototype.{byteLength, byteOffset}.

michaelficarra: Yeah, we will be doing a lot of work on it this week.

rkirsling:
@anba [%TypedArray%.prototype.slice](https://tc39.es/ecma262/#sec-%typedarray%.prototype.slice) does seem unique in this respect, but I'm also not sure a change is required? Step 15.b checks that _O_.[[ViewedArrayBuffer]] is detached before making a direct call to GetValueFromBuffer (and right after giving it a local name), but it seems that this is effectively an assertion, since we already verified this on step 2 ([ValidateTypedArray](https://tc39.es/ecma262/#sec-validatetypedarray) step 4)?

That said, we _do_ double-down on this check in [%TypedArray%.prototype.fill](https://tc39.es/ecma262/#sec-%typedarray%.prototype.fill) step 10—is there some reason that this isn't redundant?

anba:
`TypedArraySpeciesCreate()` in step 9 can detach `O`. Then if `count > 0`, `? Get(O, Pk)` in step 14.b.ii throws for detached buffers. And step 15.b has an explicit detached buffer check. With the current PR, step 14 will no longer throw, whereas step 15 will still throw, because there's the explicit check. For consistency both code paths should either both throw or not throw.

rkirsling:
> TypedArraySpeciesCreate() in step 9 can detach O.

Ah, thanks. I think I've (hopefully) covered your concerns now.

spec.html
rkirsling: Sounds good, I've gone ahead and addressed this then.

spec.html
```diff
@@ -9201,7 +9201,7 @@ <h1>[[DefineOwnProperty]] ( _P_, _Desc_ )</h1>
               1. If _Desc_ has a [[Writable]] field and if _Desc_.[[Writable]] is *false*, return *false*.
               1. If _Desc_ has a [[Value]] field, then
                 1. Let _value_ be _Desc_.[[Value]].
-                1. Return ? IntegerIndexedElementSet(_O_, _numericIndex_, _value_).
+                1. Return ! IntegerIndexedElementSet(_O_, _numericIndex_, _value_).
```
anba, line 9204: Still fallible b/c of ToNumber/ToBigInt.

spec.html
```diff
@@ -9228,7 +9228,7 @@ <h1>[[Set]] ( _P_, _V_, _Receiver_ )</h1>
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
-              1. Return ? IntegerIndexedElementSet(_O_, _numericIndex_, _V_).
+              1. Return ! IntegerIndexedElementSet(_O_, _numericIndex_, _V_).
```
anba, line 9231: Ditto.

spec.html
```diff
@@ -35129,6 +35129,7 @@ <h1>%TypedArray%.prototype.slice ( _start_, _end_ )</h1>
           1. If _relativeEnd_ &lt; 0, let _final_ be max((_len_ + _relativeEnd_), 0); else let _final_ be min(_relativeEnd_, _len_).
           1. Let _count_ be max(_final_ - _k_, 0).
           1. Let _A_ be ? TypedArraySpeciesCreate(_O_, &laquo; _count_ &raquo;).
+          1. If IsDetachedBuffer(_O_.[[ViewedArrayBuffer]]) is *true*, throw a *TypeError* exception.
```
anba, line 35132: Previous behaviour was to throw iff `count > 0`. Similar to how for example [%TypedArray%.prototype.copyWithin](https://tc39.es/ecma262/#sec-%typedarray%.prototype.copywithin) works (cf. step 11.c). [%TypedArray%.prototype.fill](https://tc39.es/ecma262/#sec-%typedarray%.prototype.fill) was the only outlier which always throws even when no elements are touched (cf. step 10).

anba (changes requested):

anba:
Is there agreement to keep the explicit detached ArrayBuffer checks in `%TypedArray%.prototype.copyWithin`, `%TypedArray%.prototype.fill`, and `%TypedArray%.prototype.slice`? Not all engines include these checks.

`detach` function:
<details>

```
function detach(ab) {
  if (ArrayBuffer.transfer) {
    ArrayBuffer.transfer(ab);
    ArrayBuffer.detach(ab);
    detachArrayBuffer(ab);
    transferArrayBuffer(ab)
      w.postMessage(ab, [ab]);
      w.terminate();
    }
  } else {
  }
}
```
</details>

Correctly throws a TypeError in JSC and SM, but no error thrown in V8:
```
var ta = new Int32Array(10);
ta.fill({
  valueOf() {
    detach(ta.buffer);
    return 0;
  }
});
```

Same for `copyWithin()`:
```
var ta = new Int32Array(10);
ta.copyWithin({
  valueOf() {
    detach(ta.buffer);
    return 0;
  }
});
```

And for the explicit check in `%ArrayIteratorPrototype%.next ( )`:

Correctly prints "0" and then throws TypeError in V8 and SM, but doesn't throw an error in JSC (, still stops after the first element):
```
var ta = new Int32Array(10);
for (var k of ta) {
  detach(ta.buffer);
  print(k);
}
```

Shouldn't throw a TypeError, but incorrectly throws one in V8:
```
var ta = new Int32Array(1);
var vals = ta.values();
print(JSON.stringify(vals.next())); // {"value":0,"done":false}
print(JSON.stringify(vals.next())); // {"done":true}
detach(ta.buffer);
print(JSON.stringify(vals.next())); // {"done":true}
```



There are probably still more differences in detached ArrayBuffer across engines. Do we attempt to find and fix them as part of this bug? For example in `%TypedArray%.prototype.set`:

Expected: Prints "get length", "get one", and then throws a TypeError.
Actual:
- JSC: Prints "get length" and then throws a RangeError
- V8: Prints "get length" and then crashes with `Check failed: !destination_ta->WasDetached().`. (Fortunately that's a release crash.)
```
var ta = new Int32Array(10);
ta.set({
  get length() {
    print("get length");
    detach(ta.buffer);
    return 1;
  },
  get 0() {
    print("get one");
    return 100;
  },
});
```

---

anba:
And more importantly [[GetOwnProperty]] needs to be changed to use `configurable = true` for indexed properties. Engines currently don't do this, but it's required before this can land in the spec.

anba:
`for-in` loops also show differences across engines. Are no longer existent indexed properties considered as *deleted* for [EnumerateObjectProperties](https://tc39.es/ecma262/#sec-enumerate-object-properties)?
```
var ta = new Int8Array(4);
for (var k in ta) {
  print(k);
  detach(ta.buffer);
}
```

syg:
My thinking is that we align with reality on what's here in this PR, and punt on the method incompats that @anba listed.

https://github.com/tc39/ecma262/pull/2164#issuecomment-689483167 is an interesting point. As it stands, TAs break the non-configurable invariant in that detached buffers cause TAs to return `undefined` when getting own property descriptors on the indexed properties. But changing configurable to true doesn't seem possible either: what we expect `delete` to do? TAs can't have holes -- making them have holes is a nonstarter.

What are our options here? I don't see a way to unbreak the non-configurable invariant as well as keeping with reality around detached buffers.

spec.html
rkirsling: Oops, yes, my bad. 🤦

bakkot:
Strictly speaking it is not an [essential invariant](https://tc39.es/ecma262/#sec-invariants-of-the-essential-internal-methods) that a `configurable: true` property can be `delete`'d (only the converse). You can make a Proxy which reports `configurable: true` for some property but refuses to let it be deleted, for example: `new Proxy({}, { deleteProperty: (t, k) => k === 'x' ? false : Reflect.deleteProperty(t, k) })`. I wouldn't be too surprised to find that there are existing web-platform objects which display similar behavior.

syg:
Ah, good call on `delete`. What about [[HasProperty]]? It has: 

- If P was previously observed as a non-configurable own data or accessor property of the target, [[HasProperty]] must return true.

spec.html
rkirsling: Addressed.

bakkot:
> If P was previously observed as a non-configurable own data or accessor property of the target, [[HasProperty]] must return true.

As with `delete`, this is only a constraint on properties which are _not_ configurable, and places no constraints on those which are.

syg:
Ah, of course, then there's no issue here.

rkirsling:
> My thinking is that we align with reality on what's here in this PR, and punt on the method incompats that @anba listed.

Agreed. The crucial thing here is fixing the spec where reality is consistent (and has been for a long time). Ending up with as much consistency as possible would be ideal, but it's less urgent since (1) behavior that engines don't agree upon is behavior that users can't rely upon and (2) removing errors is web-compatible.

anba:
Integer indexed object will need to have their own [[Delete]] method. Something along the lines of:

1. Assert: IsPropertyKey(P) is true.
1. Assert: O is an Integer-Indexed exotic object.
1. If Type(P) is String, then
    1. Let numericIndex be ! CanonicalNumericIndexString(P).
    1. If numericIndex is not undefined, then
        1. If IsDetachedBuffer(buffer) is true, return true.
        1. If ! IsValidIntegerIndex(O, numericIndex) is false, return true.
        1. Return false.
1. Return ? OrdinaryDelete(O, P).

And the configurable check in [[DefineOwnProperty]] needs to be flipped.

syg:
@rkirsling My understanding of this PR currently is the following. Is this correct?

- Make `byteLength` return 0 for detached buffers.
- Make `length` return 0 for detached buffers.
- Make [[Get]] return undefined for detached buffers.
- Make [[Set]] do nothing in sloppy mode for detached buffers.
- Make `slice()` only throw when it would return a non-empty array.

Unfortunately not all of these are web reality. Only `byteLength` and `length` seem to be.

I'd still prefer that we align on non-throwing behavior for [[Get]] and [[Set]], but would need input from Apple.

Here's a test program and the output:

## `test.js`
<details>

```javascript
function detach(ab) {
  if (ArrayBuffer.transfer) {
    ArrayBuffer.transfer(ab);
    ArrayBuffer.detach(ab);
    detachArrayBuffer(ab);
    transferArrayBuffer(ab)
      w.postMessage(ab, [ab]);
      w.terminate();
    }
  } else {
  }
}

class MyArray extends Uint32Array {
  static get [Symbol.species]() { detach(gToDetach.buffer); return Uint8Array; }
}

function test(msg, cb) {
  print(msg);
  try { print("  >> " + cb()); } catch (e) { print("  >> caught " + e); }
  print("");
}

var ta = new Uint8Array(512);
detach(ta.buffer);

test(".byteLength", () => ta.byteLength);
test(".length", () => ta.length);
test("[[Get]]", () => ta[42]);
test("[[Set]]", () => ta[42] = 42);
test("[[Has]]", () => ta.hasOwnProperty("42"));
test(".slice() @@species count>0 path", () => {
  gToDetach = new MyArray(512);
  gToDetach.slice(0, 128);
});
test(".slice() @@species count=0 path", () => {
  gToDetach = new MyArray(512);
  gToDetach.slice(42, 42);
});
test(".slice() count>0 path", () => {
  ta.slice(0, 128);
});
test(".slice() count=0 path", () => {
  ta.slice(42, 42);
});
```
</details>

## Output
#### jsc
```
.byteLength
  >> 0

.length
  >> 0

[[Get]]
  >> caught TypeError: Underlying ArrayBuffer has been detached from the view

[[Set]]
  >> caught TypeError: Underlying ArrayBuffer has been detached from the view

[[Has]]
  >> true

.slice() @@species count>0 path
  >> caught TypeError: Underlying ArrayBuffer has been detached from the view

.slice() @@species count=0 path
  >> caught TypeError: Underlying ArrayBuffer has been detached from the view

.slice() count>0 path
  >> caught TypeError: Underlying ArrayBuffer has been detached from the view

.slice() count=0 path
  >> caught TypeError: Underlying ArrayBuffer has been detached from the view
```

#### sm
```
.byteLength
  >> 0

.length
  >> 0

[[Get]]
  >> undefined

[[Set]]
  >> 42

[[Has]]
  >> false

.slice() @@species count>0 path
  >> caught TypeError: attempting to access detached ArrayBuffer

.slice() @@species count=0 path
  >> undefined

.slice() count>0 path
  >> caught TypeError: attempting to access detached ArrayBuffer

.slice() count=0 path
  >> caught TypeError: attempting to access detached ArrayBuffer
```

#### v8
```
.byteLength
  >> 0

.length
  >> 0

[[Get]]
  >> undefined

[[Set]]
  >> 42

[[Has]]
  >> false

.slice() @@species count>0 path
  >> caught TypeError: Cannot perform %TypedArray%.prototype.slice on a detached ArrayBuffer

.slice() @@species count=0 path
  >> undefined

.slice() count>0 path
  >> caught TypeError: Cannot perform %TypedArray%.prototype.slice on a detached ArrayBuffer

.slice() count=0 path
  >> caught TypeError: Cannot perform %TypedArray%.prototype.slice on a detached ArrayBuffer
```

rkirsling:
Note that this came up because I was going through JSC's test262 failures and looking for things to fix, so it should be understood that we have a slew of spec bugs yet to fix in ArrayBuffer#slice and the EIMs for integer-indexed objects. (We can get explicit confirmation from an Appler if you're worried though. 😅)

syg:
> Note that this came up because I was going through JSC's test262 failures and looking for things to fix, so it should be understood that we have a slew of conformance bugs yet to fix in ArrayBuffer#slice and the EIMs for integer-indexed objects. (We can get explicit confirmation from an Appler if you're worried though. 😅)

Good to hear! I'm not worried per se, but would nevertheless like explicit confirmation.

syg (approved):
This PR lgtm editorially.

ExE-Boss:
Personally, I’d prefer if the indexed properties were kept as `[[Configurable]]: false` and were always returned by `[[OwnPropertyKeys]]()` and `[[GetOwnProperty]]()`, like what **jsc** does in the `hasOwnProperty` case.

syg:
> Personally, I’d prefer if the indexed properties were kept as [[Configurable]]: false and were always returned by [[OwnPropertyKeys]]() and [[GetOwnProperty]](), like what jsc does in the hasOwnProperty case.

Could you say more for the motivation?

rwaldron:
While reviewing & writing tests for this, I noticed that `DataView.prototype.byteLength` is not included in the changes, but is currently spec'ed to throw on detached array buffers. Implementations currently do not agree: 

<details>
 <summary>detach-ab.js</summary>

```js
function detach(ab) {
  if (ArrayBuffer.transfer) {
    ArrayBuffer.transfer(ab);
    ArrayBuffer.detach(ab);
    detachArrayBuffer(ab);
    transferArrayBuffer(ab)
    try {
      const w = new Worker("", {type: "string"});
      w.postMessage(ab, [ab]);
      w.terminate();
    } catch (e) {
      const blob = new Blob([""], {type: "application/javascript"});
      const blobURL = URL.createObjectURL(blob);
      const w = new Worker(blobURL);
      w.postMessage(ab, [ab]);
      w.terminate();
    }
  } else {
  }
}

function test(msg, cb) {
  print(msg);
  try { print("  >> " + cb()); } catch (e) { print("  >> caught " + e); }
  print("");
}

const buffer = new ArrayBuffer(1);
const dv = new DataView(buffer, 0);
detach(dv.buffer);

test(".byteLength", () => dv.byteLength);
```
</details>



```
$ eshost -st detach-ab.js
┌────────────────┬───────────────────────────────────────────────────────────────────────────────┐
│ JavaScriptCore │ .byteLength                                                                   │
│                │   >> caught TypeError: Underlying ArrayBuffer has been detached from the view │
├────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ SpiderMonkey   │ .byteLength                                                                   │
│                │   >> caught TypeError: attempting to access detached ArrayBuffer              │
├────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ V8             │ .byteLength                                                                   │
│                │   >> 0                                                                        │
└────────────────┴───────────────────────────────────────────────────────────────────────────────┘
```

rwaldron:
What should happen if `Let targetBuffer be A.[[ViewedArrayBuffer]].` is reached, but `A.[[ViewedArrayBuffer]]` is detached?

rkirsling:
@rwaldron That shouldn't be possible since `Let A be ? TypedArraySpeciesCreate(O, « count »).` does `ValidateTypedArray` after creation and nothing in between is effectful?

rwaldron:
@rkirsling at some point after I posted that, I figured out where I was getting turned around.

rwaldron:
Tests for this change are now in Test262. The feature flag is `align-detached-buffer-semantics-with-web-reality`

bakkot:
For `.slice`, it looks like the behavior specified in this PR is for

```
.slice() @@species count>0 path
  >> TypeError

.slice() @@species count=0 path
  >> undefined (but the `.slice` does return a new empty Uint8Array)

.slice() count>0 path
  >> TypeError

.slice() count=0 path
  >> TypeError
```

That doesn't match what @syg said above:

> Make slice() only throw when it would return a non-empty array.

Specifically, it's still throwing as a consequence of step 2, "Perform ? ValidateTypedArray(O).".

Is that intentional?

bakkot (approved):
LGTM assuming the behavior described in my above comment is intentional.

michaelficarra (approved):
This LGTM, though I would also be okay with delaying the `ValidateTypedArray` call until after `_count_` was determined so that we can conform to @shu's preference.

rkirsling:
The change to `slice` in this PR was just meant to [ensure its behavior wouldn't be inadvertently changed due to implicit checks](https://github.com/tc39/ecma262/pull/2164#issuecomment-689081373), though we could certainly improve consistency here if desired.

ljharb (approved):
editorial LGTM

michaelficarra:
@rkirsling If you'd like to make that change (even though all implementations agree on throwing), please open a separate PR so we can get consensus on that.

syg:
@bakkot Thanks for catching my misunderstanding re: https://github.com/tc39/ecma262/pull/2164#issuecomment-709569671. I'm pretty sure I misread the PR at the time. The diff showed a bunch of steps moving under "If count &gt; 0", which I must've misunderstood to mean it was moving the only detached throwing case to under that condition.

For https://github.com/tc39/ecma262/pull/2164#issuecomment-709615062, I didn't intend nor currently think we should be delaying `ValidateTypedArray` to be after checking the count. I'm happy with the merged PR reflecting implementation reality.

bakkot:
Looking again, I think this ought to also have added a `IsDetachedBuffer` check to [the `[[OwnPropertyKeys]]` internal method of integer-indexed exotic objects](https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects-ownpropertykeys), so that

```js
let x = new Uint8Array(20);
detach(x.buffer);
Object.getOwnPropertyNames(x)
```

would produce `[]`, as it does in V8, SM, and JSC.

(Aside: a simple `detach` implementation for browsers is `let detach = b => postMessage(null, '*', [b])`, if you want to try it yourself.)

PR #2172, opened 2020-09-12
Editorial: a couple of fixes for Math

bakkot:
Followup to https://github.com/tc39/ecma262/pull/2122.

The first commit, https://github.com/tc39/ecma262/pull/2172/commits/75c3c480aec360a013e98890fb94bc204f1b7451, makes it explicit that coercion happens exactly once for each argument to Math.hypot. See https://github.com/tc39/ecma262/pull/2122#discussion_r487365913.

The second commit, https://github.com/tc39/ecma262/pull/2172/commits/fb2811b5b18e9d5cb9517554278acf015ec4ae81, ensures that `Math.hypot(-Infinity)` returns `Infinity`, not `-Infinity`.

The third commit, https://github.com/tc39/ecma262/pull/2172/commits/464d064159606b75f736d0e79b4269186762fd09, ensures that all possible side effects in `Math.hypot`, `Math.max`, and `Math.min` happen before the method returns.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27842,7 +27842,8 @@
           1. Let _onlyZero_ be *true*.
           1. For each element _number_ of _numbers_, do
             1. Let _n_ be ? ToNumber(_number_).
-            1. If _n_ is *NaN*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
+            1. If _n_ is *NaN* or _n_ is *+&infin;*, return _n_.
+            1. If _n_ is *-&infin;*, return *+&infin;*.
             1. If _n_ is neither *+0* nor *-0*, set _onlyZero_ to *false*.
             1. Append _n_ to _coerced_.
           1. If _onlyZero_ is *true*, return *+0*.
--- a/spec.html
+++ b/spec.html
@@ -27838,13 +27838,15 @@
         <p>When the `Math.hypot` method is called with at least two arguments _value1_ and _value2_ and any number of additional arguments which form the rest parameter ..._values_, the following steps are taken:</p>
         <emu-alg>
           1. Let _numbers_ be a List containing _value1_, _value2_, and the elements of _values_ in List order.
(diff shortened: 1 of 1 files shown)


ljharb (approved):
LGTM, thanks

spec.html
```diff
@@ -27834,14 +27834,18 @@ <h1>Math.hypot ( _value1_, _value2_, ..._values_ )</h1>
...
+            1. Append _n_ to _coerced_.
           1. Let _onlyZero_ be *true*.
-          1. For each element _number_ of _numbers_, do
-            1. Let _n_ be ? ToNumber(_number_).
-            1. If _n_ is *NaN*, _n_ is *+&infin;*, or _n_ is *-&infin;*, return _n_.
-            1. If _n_ is neither *+0* nor *-0*, set _onlyZero_ to *false*.
+          1. For each element _number_ of _coerced_, do
+            1. If _number_ is *NaN* or _number_ is *+&infin;*, return _number_.
+            1. If _number_ is *-&infin;*, return *+&infin;*.
+            1. If _number_ is neither *+0* nor *-0*, set _onlyZero_ to *false*.
```
ljharb, line 27846: this now eagerly ToNumbers everything, even if the previous logic would have had it return early - I believe the ToNumbers *have* to be interleaved with the checking logic, as in your first commit.

bakkot: Nope, the checks must come after ToNumber, hence this commit. You can confirm this in implementations. The change in #2122 was incorrect (and presumably accidental).

The previous wording was

> Each of the following `Math` object functions applies the ToNumber abstract operation to each of its arguments (in left-to-right order if there is more than one). If ToNumber returns an abrupt completion, that Completion Record is immediately returned. Otherwise, the function performs a computation on the resulting Number value(s).

ljharb: aha, you're correct.

ljharb (approved):

michaelficarra (approved):

PR #2177, opened 2020-09-18
Editorial: replace some more internal usage of strings with spec enums

michaelficarra:
This eliminates the weird unspecified "hint X" form and replaces some internal-only usage of strings with spec `~enums~`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4078,27 +4078,27 @@
     <p>The BigInt type has no implicit conversions in the ECMAScript language; programmers must call BigInt explicitly to convert values from other types.</p>
 
     <emu-clause id="sec-toprimitive" aoid="ToPrimitive" oldids="table-9">
-      <h1>ToPrimitive ( _input_ [ , _PreferredType_ ] )</h1>
-      <p>The abstract operation ToPrimitive takes argument _input_ and optional argument _PreferredType_. It converts its _input_ argument to a non-Object type. If an object is capable of converting to more than one primitive type, it may use the optional hint _PreferredType_ to favour that type. It performs the following steps when called:</p>
+      <h1>ToPrimitive ( _input_ [ , _preferredType_ ] )</h1>
+      <p>The abstract operation ToPrimitive takes argument _input_ and optional argument _preferredType_. It converts its _input_ argument to a non-Object type. If an object is capable of converting to more than one primitive type, it may use the optional hint _preferredType_ to favour that type. It performs the following steps when called:</p>
       <emu-alg>
         1. Assert: _input_ is an ECMAScript language value.
         1. If Type(_input_) is Object, then
-          1. If _PreferredType_ is not present, let _hint_ be *"default"*.
-          1. Else if _PreferredType_ is hint String, let _hint_ be *"string"*.
-          1. Else,
-            1. Assert: _PreferredType_ is hint Number.
-            1. Let _hint_ be *"number"*.
           1. Let _exoticToPrim_ be ? GetMethod(_input_, @@toPrimitive).
           1. If _exoticToPrim_ is not *undefined*, then
+            1. If _preferredType_ is not present, let _hint_ be *"default"*.
+            1. Else if _preferredType_ is ~string~, let _hint_ be *"string"*.
+            1. Else,
+              1. Assert: _preferredType_ is ~number~.
+              1. Let _hint_ be *"number"*.
             1. Let _result_ be ? Call(_exoticToPrim_, _input_, &laquo; _hint_ &raquo;).
             1. If Type(_result_) is not Object, return _result_.
             1. Throw a *TypeError* exception.
-          1. If _hint_ is *"default"*, set _hint_ to *"number"*.
-          1. Return ? OrdinaryToPrimitive(_input_, _hint_).
+          1. If _preferredType_ is not present, let _preferredType_ be ~number~.
+          1. Return ? OrdinaryToPrimitive(_input_, _preferredType_).
         1. Return _input_.
       </emu-alg>
       <emu-note>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -4078,36 +4078,36 @@ <h1>Type Conversion</h1>
...
-      <h1>ToPrimitive ( _input_ [ , _PreferredType_ ] )</h1>
-      <p>The abstract operation ToPrimitive takes argument _input_ and optional argument _PreferredType_. It converts its _input_ argument to a non-Object type. If an object is capable of converting to more than one primitive type, it may use the optional hint _PreferredType_ to favour that type. It performs the following steps when called:</p>
+      <h1>ToPrimitive ( _input_ [ , _preferredType_ ] )</h1>
+      <p>The abstract operation ToPrimitive takes argument _input_ and optional argument _preferredType_. It converts its _input_ argument to a non-Object type. If an object is capable of converting to more than one primitive type, it may use the optional hint _preferredType_ to favour that type. It performs the following steps when called:</p>
       <emu-alg>
         1. Assert: _input_ is an ECMAScript language value.
         1. If Type(_input_) is Object, then
-          1. If _PreferredType_ is not present, let _hint_ be *"default"*.
-          1. Else if _PreferredType_ is hint String, let _hint_ be *"string"*.
+          1. If _preferredType_ is not present, let _hint_ be *"default"*.
```
bakkot, line 4086: Since `_hint_` is now only used in the `If _exoticToPrim_ is not *undefined*` branch, this if/else if/else should probably be moved under that.

michaelficarra: Done.

bakkot (approved):

spec.html
```diff
@@ -4078,36 +4078,36 @@ <h1>Type Conversion</h1>
...
+          1. Return ? OrdinaryToPrimitive(_input_, _preferredType_).
         1. Return _input_.
       </emu-alg>
       <emu-note>
-        <p>When ToPrimitive is called with no hint, then it generally behaves as if the hint were Number. However, objects may over-ride this behaviour by defining a @@toPrimitive method. Of the objects defined in this specification only Date objects (see <emu-xref href="#sec-date.prototype-@@toprimitive"></emu-xref>) and Symbol objects (see <emu-xref href="#sec-symbol.prototype-@@toprimitive"></emu-xref>) over-ride the default ToPrimitive behaviour. Date objects treat no hint as if the hint were String.</p>
+        <p>When ToPrimitive is called with no hint, then it generally behaves as if the hint was ~number~. However, objects may over-ride this behaviour by defining a @@toPrimitive method. Of the objects defined in this specification only Date objects (see <emu-xref href="#sec-date.prototype-@@toprimitive"></emu-xref>) and Symbol objects (see <emu-xref href="#sec-symbol.prototype-@@toprimitive"></emu-xref>) over-ride the default ToPrimitive behaviour. Date objects treat no hint as if the hint was ~string~.</p>
```
bakkot, line 4101: I notice you also changed "as if the hint _were_"  to "as if the hint _was_". Both seem correct to me, so I don't care either way, just want to call attention to it.

ljharb, line 4101:
```suggestion
        <p>When ToPrimitive is called with no hint, then it generally behaves as if the hint was ~number~. However, objects may override this behaviour by defining a @@toPrimitive method. Of the objects defined in this specification only Date objects (see <emu-xref href="#sec-date.prototype-@@toprimitive"></emu-xref>) and Symbol objects (see <emu-xref href="#sec-symbol.prototype-@@toprimitive"></emu-xref>) override the default ToPrimitive behaviour. Date objects treat no hint as if the hint was ~string~.</p>
```
"override" with a hyphen stuck in the middle seems bizarre to me, is that a common thing?

re was vs were, when it's referring to a single variable (`_hint_`), `was` seems right to me; in the case where it's prose, "were" seems like a better fit to me.

spec.html
```diff
@@ -4078,36 +4078,36 @@ <h1>Type Conversion</h1>
...
-          1. Else,
-            1. Assert: _PreferredType_ is hint Number.
-            1. Let _hint_ be *"number"*.
           1. Let _exoticToPrim_ be ? GetMethod(_input_, @@toPrimitive).
           1. If _exoticToPrim_ is not *undefined*, then
+            1. If _preferredType_ is not present, let _hint_ be *"default"*.
+            1. Else if _preferredType_ is ~string~, let _hint_ be *"string"*.
+            1. Else,
+              1. Assert: _preferredType_ is ~number~.
+              1. Let _hint_ be *"number"*.
```
ljharb, line 4092: kind of seems like it'd be simpler to just pass the string instead of doing a conversion from "not a string" to "basically the equivalent string"

spec.html
michaelficarra: Pretty sure the over-use (see what I did there?) of hyphens is a British-ism. I could go either way with this one, but I think I prefer it for consistency with "non-zero" and "non-normative". Either way, I don't think it's in scope for this PR.

Regarding was/were, "were" was the adjacent word and sounded wrong to me, so I changed it. But if it's controversial, I can back it out and we can address it in another PR.

spec.html
michaelficarra: Eh, I'd weakly prefer to pass the spec enum at all the call sites and create a string here. If we used a string, it'd be serving a dual purpose: signalling the desired behaviour of OrdinaryToPrimitive and passing to user functions. Having two separate values (even if isomorphic) seems like it separates those concepts better.

spec.html
jmdyck: (Here, "were" uses the [subjunctive mood](https://en.wikipedia.org/wiki/English_subjunctive), which seems to be gradually fading from use in English.)

spec.html
ljharb: Fair enough.

michaelficarra:
@ljharb Can I get a review on this? Your only comment so far was about content that was not a significant part of this PR.

spec.html
ljharb: I lightly prefer "were".

ljharb: Agreed that changing "over-ride" is out of scope of this particular PR. We can discuss "were" on the other thread.

ljharb (approved):

spec.html
michaelficarra: Reverted back to "were".

PR #2182, opened 2020-09-18
Editorial: normalise date formatting/punctuation

michaelficarra:


spec.html
```diff
@@ -28164,7 +28164,7 @@ <h1>Month Number</h1>
         </emu-eqn>
         <p>where</p>
         <emu-eqn id="eqn-DayWithinYear" aoid="DayWithinYear">DayWithinYear(_t_) = Day(_t_) - DayFromYear(YearFromTime(_t_))</emu-eqn>
-        <p>A month value of 0 specifies January; 1 specifies February; 2 specifies March; 3 specifies April; 4 specifies May; 5 specifies June; 6 specifies July; 7 specifies August; 8 specifies September; 9 specifies October; 10 specifies November; and 11 specifies December. Note that <emu-eqn>MonthFromTime(0) = 0</emu-eqn>, corresponding to Thursday, 01 January, 1970.</p>
+        <p>A month value of 0 specifies January; 1 specifies February; 2 specifies March; 3 specifies April; 4 specifies May; 5 specifies June; 6 specifies July; 7 specifies August; 8 specifies September; 9 specifies October; 10 specifies November; and 11 specifies December. Note that <emu-eqn>MonthFromTime(0) = 0</emu-eqn>, corresponding to Thursday, 1 January 1970.</p>
```
bakkot, line 28167: I'm not sure the "Thursday" before "1 January 1970" is doing much good here. I'd keep it in `WeekDay`, though.

michaelficarra: Is it worth making that change in this PR? I was only expecting to make a formatting change here.

spec.html
```diff
@@ -1913,7 +1913,7 @@ <h1>Number::toString ( _x_ )</h1>
           </emu-note>
           <emu-note>
             <p>Implementers of ECMAScript may find useful the paper and code written by David M. Gay for binary-to-decimal conversion of floating-point numbers:</p>
-            <p>Gay, David M. Correctly Rounded Binary-Decimal and Decimal-Binary Conversions. Numerical Analysis, Manuscript 90-10. AT&amp;T Bell Laboratories (Murray Hill, New Jersey). November 30, 1990. Available as
+            <p>Gay, David M. Correctly Rounded Binary-Decimal and Decimal-Binary Conversions. Numerical Analysis, Manuscript 90-10. AT&amp;T Bell Laboratories (Murray Hill, New Jersey). 30 November 1990. Available as
```
ljharb, line 1916: if we're going to normalize on a format, why not YMD?

michaelficarra: Because they're not fully numeric.

spec.html
bakkot: Seems like a formatting question to me, but I don't really care either way.

spec.html
ljharb: and why normalize on `30 November 1990` instead of `November 30, 1990`?

michaelficarra: Arbitrary choice. I guess it's better because it's still in order?

This article shows many more people use big-endian/little-endian ordering than use middle-endian ordering: https://en.wikipedia.org/wiki/Date_format_by_country.

ljharb: Sure, but that ordering applies to triplicate forms like `Y/M/D` or `Y-M-D`, whereas "M D, Y" is prose - iow, the DMY/MDY ambiguity doesn't apply when the month name is non-numeric, so there doesn't seem to be a clear advantage in that regard.

To me, the "30 November 1990" form feels stilted in a sentence and like it belongs in a bibliography, where the "November 30, 1990" form feels like a better fit for a sentence.

michaelficarra: image

ljharb: Right, but this is in prose, where none of those scenarios applies.

syg (approved):

ljharb (approved):
sigh, i hate oxford style (but i love their comma)

PR #2185, opened 2020-09-21
Editorial: handle NaN and infinities in Math.round

bakkot:
Followup to #2122.


ljharb (approved):

ljharb:
Technically i think this is a normative spec bug, which means it doesn't need consensus?

bakkot:
I marked it as editorial because #2122 did not intend to change it.

michaelficarra (approved):

Issue #2183, opened 2020-09-18
inconsistent notation for multiplication and division


michaelficarra:
We inconsistently use `*` and &times; for multiplication and `/` and &divide; for division. We should use consistent notation.

jmdyck:
The only place I see us using `*` for multiplication is in [MakeTime](https://tc39.es/ecma262/#sec-maketime), where it says "(that is, as if using the ECMAScript operators `*` and `+`)", so that's a conscious decision.

But yes re the inconsistency on division: we use `&divide;` twice (in [BigInt::leftShift](https://tc39.es/ecma262/#sec-numeric-types-bigint-leftShift) and [Number.prototype.toFixed](https://tc39.es/ecma262/#sec-number.prototype.tofixed)), and use `/` everywhere else.

michaelficarra:
@jmdyck #2007 should be removing that first case soon.

michaelficarra:
Discussed in the editor call today. Multiplication will use `×` and division will use `/`, in accordance with ISO 80000-2:2019.

@bakkot will double check that MakeTime uses the correct notation in #2007.

bakkot:
Relevant bits of ISO 80000-2:

<img width="823" alt="Screen Shot 2020-09-30 at 3 26 10 PM" src="https://user-images.githubusercontent.com/1653598/94746303-4e7c9380-0331-11eb-8ed1-6471d33349c5.png">


<img width="817" alt="Screen Shot 2020-09-30 at 3 25 18 PM" src="https://user-images.githubusercontent.com/1653598/94746256-33aa1f00-0331-11eb-92eb-af3c91e24088.png">


<img width="819" alt="Screen Shot 2020-09-30 at 3 23 57 PM" src="https://user-images.githubusercontent.com/1653598/94746189-0cebe880-0331-11eb-81a0-a5b7043d8873.png">

PR #2189, opened 2020-09-23
Editorial: convert Number abstract operations to algorithmic steps

### LINKED ISSUE #2179: convert some Number:: AOs to use algorithm steps
In the same way we recently rewrote all of the Math built-ins to use algorithm steps (#2122), some of the Number:: AOs can be rewritten in the more typical form, as a sequence of algorithm steps. I've identified the following AOs:

* Number::exponentiate
* Number::multiply
* Number::divide
* Number::remainder
* Number::add

/cc @ryanjduffy in case he has any interest doing another one of these.

ryanjduffy:
_Draft_ to fix #2179

* Number::exponentiate
* Number::multiply
* Number::divide
* Number::remainder
* Number::add

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1577,32 +1577,38 @@
 
         <emu-clause id="sec-numeric-types-number-exponentiate" oldids="sec-applying-the-exp-operator">
           <h1>Number::exponentiate ( _base_, _exponent_ )</h1>
-          <p>The abstract operation Number::exponentiate takes arguments _base_ (a Number) and _exponent_ (a Number). It returns an implementation-approximated value representing the result of raising _base_ to the power _exponent_, subject to the following requirements:</p>
-          <ul>
-            <li>If _exponent_ is *NaN*, the result is *NaN*.</li>
-            <li>If _exponent_ is *+0*<sub>𝔽</sub>, the result is *1*<sub>𝔽</sub>, even if _base_ is *NaN*.</li>
-            <li>If _exponent_ is *-0*<sub>𝔽</sub>, the result is *1*<sub>𝔽</sub>, even if _base_ is *NaN*.</li>
-            <li>If _base_ is *NaN* and _exponent_ is non-zero, the result is *NaN*.</li>
-            <li>If abs(ℝ(_base_)) &gt; 1 and _exponent_ is *+&infin;*<sub>𝔽</sub>, the result is *+&infin;*<sub>𝔽</sub>.</li>
-            <li>If abs(ℝ(_base_)) &gt; 1 and _exponent_ is *-&infin;*<sub>𝔽</sub>, the result is *+0*<sub>𝔽</sub>.</li>
-            <li>If abs(ℝ(_base_)) is 1 and _exponent_ is *+&infin;*<sub>𝔽</sub>, the result is *NaN*.</li>
-            <li>If abs(ℝ(_base_)) is 1 and _exponent_ is *-&infin;*<sub>𝔽</sub>, the result is *NaN*.</li>
-            <li>If abs(ℝ(_base_)) &lt; 1 and _exponent_ is *+&infin;*<sub>𝔽</sub>, the result is *+0*<sub>𝔽</sub>.</li>
-            <li>If abs(ℝ(_base_)) &lt; 1 and _exponent_ is *-&infin;*<sub>𝔽</sub>, the result is *+&infin;*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *+&infin;*<sub>𝔽</sub> and ℝ(_exponent_) &gt; 0, the result is *+&infin;*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *+&infin;*<sub>𝔽</sub> and ℝ(_exponent_) &lt; 0, the result is *+0*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *-&infin;*<sub>𝔽</sub> and ℝ(_exponent_) &gt; 0 and _exponent_ is an odd integral Number, the result is *-&infin;*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *-&infin;*<sub>𝔽</sub> and ℝ(_exponent_) &gt; 0 and _exponent_ is not an odd integral Number, the result is *+&infin;*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *-&infin;*<sub>𝔽</sub> and ℝ(_exponent_) &lt; 0 and _exponent_ is an odd integral Number, the result is *-0*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *-&infin;*<sub>𝔽</sub> and ℝ(_exponent_) &lt; 0 and _exponent_ is not an odd integral Number, the result is *+0*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *+0*<sub>𝔽</sub> and ℝ(_exponent_) &gt; 0, the result is *+0*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *+0*<sub>𝔽</sub> and ℝ(_exponent_) &lt; 0, the result is *+&infin;*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *-0*<sub>𝔽</sub> and ℝ(_exponent_) &gt; 0 and _exponent_ is an odd integral Number, the result is *-0*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *-0*<sub>𝔽</sub> and ℝ(_exponent_) &gt; 0 and _exponent_ is not an odd integral Number, the result is *+0*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *-0*<sub>𝔽</sub> and ℝ(_exponent_) &lt; 0 and _exponent_ is an odd integral Number, the result is *-&infin;*<sub>𝔽</sub>.</li>
-            <li>If _base_ is *-0*<sub>𝔽</sub> and ℝ(_exponent_) &lt; 0 and _exponent_ is not an odd integral Number, the result is *+&infin;*<sub>𝔽</sub>.</li>
-            <li>If ℝ(_base_) &lt; 0 and _base_ is finite and _exponent_ is finite and _exponent_ is not an integral Number, the result is *NaN*.</li>
-          </ul>
+          <p>The abstract operation Number::exponentiate takes arguments _base_ (a Number) and _exponent_ (a Number). It returns an implementation-approximated value representing the result of raising _base_ to the _exponent_ power. It performs the following steps when called:</p>
+          <emu-alg>
+            1. If _exponent_ is *NaN*, return *NaN*.
+            1. If _exponent_ is *+0*<sub>𝔽</sub> or _exponent_ is *-0*<sub>𝔽</sub>, return *1*<sub>𝔽</sub>.
+            1. If _base_ is *NaN*, return *NaN*.
+            1. If _base_ is *+&infin;*<sub>𝔽</sub>, then
+              1. If _exponent_ &gt; *+0*<sub>𝔽</sub>, return *+&infin;*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
+            1. If _base_ is *-&infin;*<sub>𝔽</sub>, then
+              1. If _exponent_ &gt; *+0*<sub>𝔽</sub>, then
+                1. If _exponent_ is an odd integral Number, return *-&infin;*<sub>𝔽</sub>. Otherwise, return *+&infin;*<sub>𝔽</sub>.
+              1. Else,
+                1. If _exponent_ is an odd integral Number, return *-0*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
+            1. If _base_ is *+0*<sub>𝔽</sub>, then
+              1. If _exponent_ &gt; *+0*<sub>𝔽</sub>, return *+0*<sub>𝔽</sub>. Otherwise, return *+&infin;*<sub>𝔽</sub>.
+            1. If _base_ is *-0*<sub>𝔽</sub>, then
+              1. If _exponent_ &gt; *+0*<sub>𝔽</sub>, then
+                1. If _exponent_ is an odd integral Number, return *-0*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
+              1. Else,
+                1. If _exponent_ is an odd integral Number, return *-&infin;*<sub>𝔽</sub>. Otherwise, return *+&infin;*<sub>𝔽</sub>.
+            1. Assert: _base_ is finite and is neither *+0*<sub>𝔽</sub> nor *-0*<sub>𝔽</sub>.
+            1. If _exponent_ is *+&infin;*<sub>𝔽</sub>, then
+              1. If abs(ℝ(_base_)) &gt; 1, return *+&infin;*<sub>𝔽</sub>.
+              1. If abs(ℝ(_base_)) is 1, return *NaN*.
+              1. If abs(ℝ(_base_)) &lt; 1, return *+0*<sub>𝔽</sub>.
+            1. If _exponent_ is *-&infin;*<sub>𝔽</sub>, then
+              1. If abs(ℝ(_base_)) &gt; 1, return *+0*<sub>𝔽</sub>.
+              1. If abs(ℝ(_base_)) is 1, return *NaN*.
+              1. If abs(ℝ(_base_)) &lt; 1, return *+&infin;*<sub>𝔽</sub>.
+            1. Assert: _exponent_ is finite and is neither *+0*<sub>𝔽</sub> nor *-0*<sub>𝔽</sub>.
+            1. If _base_ &lt; *+0*<sub>𝔽</sub> and _exponent_ is not an integral Number, return *NaN*.
+            1. Return an implementation-approximated value representing the result of raising ℝ(_base_) to the ℝ(_exponent_) power.
+          </emu-alg>
           <emu-note>
             <p>The result of _base_ `**` _exponent_ when _base_ is *1*<sub>𝔽</sub> or *-1*<sub>𝔽</sub> and _exponent_ is *+&infin;*<sub>𝔽</sub> or *-&infin;*<sub>𝔽</sub>, or when _base_ is *1*<sub>𝔽</sub> and _exponent_ is *NaN*, differs from IEEE 754-2019. The first edition of ECMAScript specified a result of *NaN* for this operation, whereas later versions of IEEE 754-2019 specified *1*<sub>𝔽</sub>. The historical ECMAScript behaviour is preserved for compatibility reasons.</p>
           </emu-note>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -1691,32 +1691,26 @@ <h1>Number::remainder ( _n_, _d_ )</h1>
         <emu-clause id="sec-numeric-types-number-add" oldids="sec-applying-the-additive-operators-to-numbers">
           <h1>Number::add ( _x_, _y_ )</h1>
           <p>The abstract operation Number::add takes arguments _x_ (a Number) and _y_ (a Number). It performs addition, producing the sum of _x_ and _y_ as determined using the rules of IEEE 754-2019 binary double-precision arithmetic:</p>
```
michaelficarra, line 1693: We should make this read more like the typical form:

> The abstract operation Number::add takes arguments _x_ (a Number) and _y_ (a Number). It performs addition according to the rules of IEEE 754-2019 binary double-precision arithmetic, producing the sum of its arguments. It performs the following steps when called:

spec.html
```diff
@@ -1691,32 +1691,26 @@ <h1>Number::remainder ( _n_, _d_ )</h1>
...
+              1. If _ny_ is *-&infin;*, return *NaN*; else return *+&infin;*.
+            1. If _nx_ is *-&infin;*, then
+              1. If _ny_ is *+&infin;*, return *NaN*; else return *-&infin;*.
+            1. Assert: _nx_ is finite
+            1. If _ny_ is *+&infin;* or _ny_ is *-&infin;*, return _ny_.
+            1. If _nx_ is *+0*, then
+              1. If _ny_ is *-0*, return *+0*; else return _ny_.
+            1. If _nx_ is *-0*, then
+              1. If _ny_ is *+0*, return *+0*; else return _ny_.
+            1. If _nx_ is *-0* and _ny_ is *-0*, return *-0*.
```
michaelficarra, line 1708:
```suggestion
            1. If _nx_ is *-0* and _ny_ is *-0*, return *-0*.
            1. If _nx_ is *+0* or _nx_ is *-0*, return _ny_.
```

spec.html
```diff
@@ -1691,32 +1691,26 @@ <h1>Number::remainder ( _n_, _d_ )</h1>
...
+            1. Assert: _nx_ is neither *+0* nor *-0*.
+            1. If _ny_ is *+0* or _ny_ is *-0*, return _nx_.
+            1. If _nx_ does not equal _ny_ and abs(_nx) equals _ny_, return *+0*.
```
michaelficarra, line 1711: Is this step necessary? I don't think it is.

spec.html
```diff
@@ -1691,32 +1691,26 @@ <h1>Number::remainder ( _n_, _d_ )</h1>
...
+            1. Return the nearest representable value of the sum of _nx_ and _ny_.
```
michaelficarra, line 1712: We should continue to specify the rounding mode here.

```suggestion
            1. Return the nearest representable value of the sum of _nx_ and _ny_ using IEEE 754-2019 roundTiesToEven mode.
```

@tc39/ecma262-editors do we need the mentions of overflow producing an infinity and gradual underflow support here?

spec.html
```diff
@@ -1691,32 +1691,26 @@ <h1>Number::remainder ( _n_, _d_ )</h1>
...
-            </li>
-            <li>
-              The sum of two nonzero finite values of the same magnitude and opposite sign is *+0*.
-            </li>
-            <li>
-              In the remaining cases, where neither an infinity, nor a zero, nor *NaN* is involved, and the operands have the same sign or have different magnitudes, the sum is computed and rounded to the nearest representable value using IEEE 754-2019 roundTiesToEven mode. If the magnitude is too large to represent, the operation overflows and the result is then an infinity of appropriate sign. The ECMAScript language requires support of gradual underflow as defined by IEEE 754-2019.
-            </li>
-          </ul>
+          <emu-alg>
+            1. Let _nx_ be ? ToNumber(_x_).
```
michaelficarra, line 1695: We don't actually need to ToNumber these since they come in as Numbers already.

spec.html
```diff
@@ -1691,32 +1691,26 @@ <h1>Number::remainder ( _n_, _d_ )</h1>
...
+            1. Let _ny_ be ? ToNumber(_y_).
+            1. If _nx_ is *NaN* or _ny_ is *NaN*, return *NaN*.
+            1. If _nx_ is *+&infin;*, then
+              1. If _ny_ is *-&infin;*, return *NaN*; else return *+&infin;*.
+            1. If _nx_ is *-&infin;*, then
+              1. If _ny_ is *+&infin;*, return *NaN*; else return *-&infin;*.
+            1. Assert: _nx_ is finite
```
michaelficarra, line 1702: We could move this assert down one and assert that they're both finite. That might be more useful for a reader.

spec.html
ljharb: probably best to keep it.

michaelficarra:
```suggestion
            1. Return the nearest representable value of the sum of _nx_ and _ny_ using IEEE 754-2019 roundTiesToEven mode. If the magnitude is too large to represent, the operation overflows and the result will be +&infin; or -&infin;. IEEE 754-2019 gradual underflow must be supported.
```

spec.html
```diff
@@ -1691,32 +1691,26 @@ <h1>Number::remainder ( _n_, _d_ )</h1>
...
+            1. If _ny_ is *+&infin;* or _ny_ is *-&infin;*, return _ny_.
+            1. If _nx_ is *+0*, then
+              1. If _ny_ is *-0*, return *+0*; else return _ny_.
+            1. If _nx_ is *-0*, then
+              1. If _ny_ is *+0*, return *+0*; else return _ny_.
+            1. If _nx_ is *-0* and _ny_ is *-0*, return *-0*.
+            1. Assert: _nx_ is neither *+0* nor *-0*.
```
michaelficarra, line 1709: I don't think this assert helps the reader like the finite assert above does.

michaelficarra:
@ryanjduffy Are you still planning on finishing this PR?

spec.html
```diff
@@ -1618,125 +1618,80 @@ <h1>Number::exponentiate ( _base_, _exponent_ )</h1>
...
+              1. If _x_ is *+0*<sub>𝔽</sub> or _x_ is positive, return *-0*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
+            1. If _x_ is *+0*<sub>𝔽</sub> or _x_ is *-0*<sub>𝔽</sub>, then
+              1. If _y_ is *+0*<sub>𝔽</sub> or _y_ is *-0*<sub>𝔽</sub>, return *NaN*.
+              1. If _y_ is positive, return _x_.
+              1. If _x_ is *+0*<sub>𝔽</sub>, return *-0*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
+            1. If _y_ is *+0*<sub>𝔽</sub>, then
+              1. If _x_ is positive, return *+&infin;*<sub>𝔽</sub>. Otherwise, return *-&infin;*<sub>𝔽</sub>.
+            1. If _y_ is *-0*<sub>𝔽</sub>, then
+              1. If _x_ is positive, return *-&infin;*<sub>𝔽</sub>. Otherwise, return *+&infin;*<sub>𝔽</sub>.
+            1. Return 𝔽(ℝ(_x_) &divide; ℝ(_y_)).
```
bakkot, line 1660: We shouldn't use `÷`, as [previously discussed](https://github.com/tc39/ecma262/issues/2183#issuecomment-701677977).

michaelficarra: That issue hasn't been resolved yet, so it's still inconsistent in the draft. But I'll switch it out.

bakkot: I'll [enforce it in the linter](https://github.com/tc39/ecmarkup/pull/280) and we can fix it here as part of upstreaming that change.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
-            <li>
-              In the remaining cases, where neither an infinity nor *NaN* is involved, the product is computed and rounded to the nearest representable value using IEEE 754-2019 roundTiesToEven mode. If the magnitude is too large to represent, the result is then an infinity of appropriate sign. If the magnitude is too small to represent, the result is then a zero of appropriate sign. The ECMAScript language requires support of gradual underflow as defined by IEEE 754-2019.
-            </li>
-          </ul>
+          <p>The abstract operation Number::multiply takes arguments _x_ (a Number) and _y_ (a Number). It performs multiplication, producing the product of _x_ and _y_. It performs the following steps when called:</p>
+          <emu-alg>
+            1. If _x_ is *NaN* or _y_ is *NaN*, return *NaN*.
+            1. If _x_ is *+&infin;*<sub>𝔽</sub> or _x_ is *-&infin;*<sub>𝔽</sub>, then
+              1. If _y_ is *+0*<sub>𝔽</sub> or _y_ is *-0*<sub>𝔽</sub>, return *NaN*.
+              1. If _y_ is positive, return _x_.
```
bakkot, line 1634:
```suggestion
              1. If _y_ &gt; *+0*<sub>𝔽</sub>, return _x_.
```

for consistency with `::exponentiate`.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
+              1. If _x_ is *+&infin;*<sub>𝔽</sub>, return *-&infin;*<sub>𝔽</sub>. Otherwise, return *+&infin;*<sub>𝔽</sub>.
```
bakkot, line 1635: I'd prefer the `Otherwise` get its own line, here. Putting it inline obscures that there are four independent cases: y is 0, y is positive, x is +∞, otherwise (i.e. x is -∞).

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
-              Multiplication of an infinity by an infinity results in an infinity. The sign is determined by the rule already stated above.
-            </li>
-            <li>
-              Multiplication of an infinity by a finite non-zero value results in a signed infinity. The sign is determined by the rule already stated above.
-            </li>
-            <li>
-              In the remaining cases, where neither an infinity nor *NaN* is involved, the product is computed and rounded to the nearest representable value using IEEE 754-2019 roundTiesToEven mode. If the magnitude is too large to represent, the result is then an infinity of appropriate sign. If the magnitude is too small to represent, the result is then a zero of appropriate sign. The ECMAScript language requires support of gradual underflow as defined by IEEE 754-2019.
-            </li>
-          </ul>
+          <p>The abstract operation Number::multiply takes arguments _x_ (a Number) and _y_ (a Number). It performs multiplication, producing the product of _x_ and _y_. It performs the following steps when called:</p>
```
bakkot, line 1629: I would prefer to keep the "as determined by the rules of IEEE 754-2019 binary double-precision arithmetic", here and below.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
-              In the remaining cases, where neither an infinity, nor a zero, nor *NaN* is involved, the quotient is computed and rounded to the nearest representable value using IEEE 754-2019 roundTiesToEven mode. If the magnitude is too large to represent, the operation overflows; the result is then an infinity of appropriate sign. If the magnitude is too small to represent, the operation underflows and the result is a zero of the appropriate sign. The ECMAScript language requires support of gradual underflow as defined by IEEE 754-2019.
-            </li>
-          </ul>
+          <p>The abstract operation Number::divide takes arguments _x_ (a Number) and _y_ (a Number). It performs division, producing the quotient of _x_ and _y_ where _x_ is the dividend and _y_ is the divisor. It performs the following steps when called:</p>
+          <emu-alg>
+            1. If _x_ is *NaN* or _y_ is *NaN*, return *NaN*.
+            1. If _x_ is *+&infin;*<sub>𝔽</sub> or _x_ is *-&infin;*<sub>𝔽</sub>, then
+              1. If _y_ is *+&infin;*<sub>𝔽</sub> or _y_ is *-&infin;*<sub>𝔽</sub>, return *NaN*.
+              1. If _y_ is *+0*<sub>𝔽</sub> or _y_ is positive, return _x_.
```
bakkot, line 1654:
```suggestion
              1. If _y_ is *+0*<sub>𝔽</sub> or _y_ &gt; *+0*<sub>𝔽</sub>, return _x_.
```
For consistency with above. (Ditto all other uses of "is positive".)

spec.html
bakkot: Also it's probably clearer to write `Return -_x_.`, rather than this branch.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
+              1. If _x_ is *+&infin;*<sub>𝔽</sub>, return *-&infin;*<sub>𝔽</sub>. Otherwise, return *+&infin;*<sub>𝔽</sub>.
```
bakkot, line 1655: It's probably clearer to write `Return -_x_`.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
+            1. If _y_ is *+&infin;*<sub>𝔽</sub>, then
+              1. If _x_ is *+0*<sub>𝔽</sub> or _x_ is positive, return *+0*<sub>𝔽</sub>. Otherwise, return *-0*<sub>𝔽</sub>.
+            1. If _y_ is *-&infin;*<sub>𝔽</sub>, then
+              1. If _x_ is *+0*<sub>𝔽</sub> or _x_ is positive, return *-0*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
+            1. If _x_ is *+0*<sub>𝔽</sub> or _x_ is *-0*<sub>𝔽</sub>, then
+              1. If _y_ is *+0*<sub>𝔽</sub> or _y_ is *-0*<sub>𝔽</sub>, return *NaN*.
+              1. If _y_ is positive, return _x_.
+              1. If _x_ is *+0*<sub>𝔽</sub>, return *-0*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
```
bakkot, line 1663: It's probably clearer to write `Return -_x_`. Ditto similar cases elsewhere; I'm going to stop commenting on all of them.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
         <emu-clause id="sec-numeric-types-number-remainder" oldids="sec-applying-the-mod-operator">
           <h1>Number::remainder ( _n_, _d_ )</h1>
-          <p>The abstract operation Number::remainder takes arguments _n_ (a Number) and _d_ (a Number). It yields the remainder of its operands from an implied division; _n_ is the dividend and _d_ is the divisor.</p>
+          <p>The abstract operation Number::remainder takes arguments _n_ (a Number) and _d_ (a Number). It yields the remainder from an implied division of its operands where _n_ is the dividend and _d_ is the divisor. It performs the following steps when called:</p>
+          <emu-alg>
+            1. If _n_ is *NaN* or _d_ is *NaN*, return *NaN*.
+            1. If _n_ is *+&infin;*<sub>𝔽</sub> or _n_ is *-&infin;*<sub>𝔽</sub>, return *NaN*.
+            1. If _d_ is *+0*<sub>𝔽</sub> or _d_ is *-0*<sub>𝔽</sub>, return *NaN*.
+            1. If _d_ is *+&infin;*<sub>𝔽</sub> or _d_ is *-&infin;*<sub>𝔽</sub>, return _n_.
+            1. If _n_ is *+0*<sub>𝔽</sub> or _n_ is *-0*<sub>𝔽</sub>, return _n_.
```
bakkot, line 1680: Nit: if you switch this line and the previous, then we check in the order n, d, n, d, instead of (as here) n, d, d, n.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
+            1. Let _r_ be ℝ(_n_) - (ℝ(_d_) &times; _q_) where _q_ is an integer that is negative if and only if _n_ and _d_ have opposite sign, and whose magnitude is as large as possible without exceeding the magnitude of ℝ(_n_) / ℝ(_d_).
```
bakkot, line 1681: It is maybe worth asserting that `_n_` and `_d_` are finite and nonzero before this line.

The math looks right.

bakkot (commented):
Some editorial comments, but otherwise looks good to me.

ryanjduffy:
@michaelficarra - Thanks for picking this up from me. Life got in the way over the past few months. :) I'll keep an eye out for other issues I can pick up in the near future.

michaelficarra:
No worries, @ryanjduffy. Thanks for the work you had already done. I'd love to see more contributions from you in the future, but don't feel pressured.

spec.html
michaelficarra: I did a little more swapping here to group the infinity/0 cases. I *think* it was okay to do that, but double check for me.

spec.html
michaelficarra: I don't think it helps much, but I added it.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
+                1. If _exponent_ is an odd integral Number, return *-0*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
+              1. Else,
+                1. If _exponent_ is an odd integral Number, return *-&infin;*<sub>𝔽</sub>. Otherwise, return *+&infin;*<sub>𝔽</sub>.
+            1. Assert: _base_ is finite and is neither *+0*<sub>𝔽</sub> nor *-0*<sub>𝔽</sub>.
+            1. If _exponent_ is *+&infin;*<sub>𝔽</sub>, then
+              1. If abs(ℝ(_base_)) &lt; 1, return *+0*<sub>𝔽</sub>.
+              1. If abs(ℝ(_base_)) is 1, return *NaN*.
+              1. If abs(ℝ(_base_)) &gt; 1, return *+&infin;*<sub>𝔽</sub>.
+            1. If _exponent_ is *-&infin;*<sub>𝔽</sub>, then
+              1. If abs(ℝ(_base_)) &lt; 1, return *+&infin;*<sub>𝔽</sub>.
```
bakkot, line 1613: Nit: I would prefer we stick to consistently treating cases like this in descending order: so `> 1`, `= 1`, `< 1`, rather than the reverse.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
+              1. If abs(ℝ(_base_)) is 1, return *NaN*.
+              1. If abs(ℝ(_base_)) &gt; 1, return *+0*<sub>𝔽</sub>.
+            1. Assert: _exponent_ is finite and is neither *+0*<sub>𝔽</sub> nor *-0*<sub>𝔽</sub>.
+            1. If ℝ(_base_) &lt; 0 and _exponent_ is not an integral Number, return *NaN*.
```
bakkot, line 1617: For consistency:

```suggestion
            1. If _base_ &lt; *+0*<sub>𝔽</sub> and _exponent_ is not an integral Number, return *NaN*.
```

spec.html
bakkot: No, it's wrong now: this now says that `0 % 0` is `0` instead of (as it should be) `NaN`. The check for `d = 0` needs to precede the check for `n = 0`.

michaelficarra:
@bakkot Addressed.

bakkot (approved):

jmdyck (dismissed):

syg (approved):

Activate256 (approved):

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
+            1. Let _r_ be an implementation-approximated value representing the result of raising ℝ(_base_) to the ℝ(_exponent_) power.
+            1. Return 𝔽(_r_).
```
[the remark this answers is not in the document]
bakkot: Done.

jmdyck (commented):
No complaints, but I didn't check that the changes were semantics-preserving.

ljharb (approved):
Verified all changes against my es-abstract implementations and tests.

spec.html
```diff
@@ -1585,158 +1585,121 @@ <h1>Number::bitwiseNOT ( _x_ )</h1>
...
+                1. If _exponent_ is an odd integral Number, return *-&infin;*<sub>𝔽</sub>. Otherwise, return *+&infin;*<sub>𝔽</sub>.
+            1. If _exponent_ is *+&infin;*<sub>𝔽</sub>, then
+              1. If abs(ℝ(_base_)) &lt; 1, return *+0*<sub>𝔽</sub>.
+              1. If abs(ℝ(_base_)) is 1, return *NaN*.
+              1. If abs(ℝ(_base_)) &gt; 1, return *+&infin;*<sub>𝔽</sub>.
+            1. If _exponent_ is *-&infin;*<sub>𝔽</sub>, then
+              1. If abs(ℝ(_base_)) &lt; 1, return *+&infin;*<sub>𝔽</sub>.
+              1. If abs(ℝ(_base_)) is 1, return *NaN*.
+              1. If abs(ℝ(_base_)) &gt; 1, return *+0*<sub>𝔽</sub>.
+            1. If ℝ(_base_) &lt; 0 and _exponent_ is not an integral Number, return *NaN*.
```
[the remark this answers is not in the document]
Activate256: Trott:lint-preset-2.0.1

PR #2201, opened 2020-10-09
Editorial: normalise phrasing used to create Lists

michaelficarra:
As suggested by @bakkot as a follow-on to #2142. *edit*: Added the rephrasing that @gibson042 suggested in that thread as well.

Note that this only addresses the remaining phrasing inconsistencies, and *does not* attempt to address the inconsistency with when we use double angle brackets to denote List literals.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4046,7 +4046,7 @@
             1. If _fromBlock_ is a Shared Data Block, then
               1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
               1. Let _eventList_ be the [[EventList]] field of the element in _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
-              1. Let _bytes_ be a List of length 1 that contains a nondeterministically chosen byte value.
+              1. Let _bytes_ be a List whose sole element is a nondeterministically chosen byte value.
               1. NOTE: In implementations, _bytes_ is the result of a non-atomic read instruction on the underlying hardware. The nondeterminism is a semantic prescription of the memory model to describe observable behaviour of hardware with weak consistency.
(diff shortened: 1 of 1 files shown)


bakkot:
There's a couple instances of "a list containing" in [Table 7](https://ci.tc39.es/preview/tc39/ecma262/sha/d41ae03082aed8c9526db6ae212b37bdd3127881/#table-additional-essential-internal-methods-of-function-objects), an instance of "a List containing" in [Table 30](https://ci.tc39.es/preview/tc39/ecma262/sha/d41ae03082aed8c9526db6ae212b37bdd3127881/#table-internal-slots-of-module-namespace-exotic-objects), and an instance of "a List consisting of" in the [notation section for RegExp](https://ci.tc39.es/preview/tc39/ecma262/sha/d41ae03082aed8c9526db6ae212b37bdd3127881/#sec-notation). These aren't technically constructing lists, but we might as well be consistent there too.

Also, sometimes we make lists by length, and we're inconsistent about that as well. [`GetModifySetValueInBuffer`](https://ci.tc39.es/preview/tc39/ecma262/sha/d41ae03082aed8c9526db6ae212b37bdd3127881/#sec-notation) 9.c. makes "a List of length [...]", and 10.a. of the same algorithm makes "a List of size [...]". Looks like other instances all use "length", so maybe just fix 10.a. here.

michaelficarra:
@bakkot I've addressed your comments.

bakkot:
@michaelficarra There's still an instance of "a List containing" in Table 30.

michaelficarra:
@bakkot Oops. Fixed.

bakkot (approved):

ljharb (approved):

PR #2207, opened 2020-10-17
Normative: IsDetachedBuffer should be checked before accessing [[ArrayLength]]

rkirsling:
It's probably wrong to call this editorial, but it's a fix for an oversight in #2164.

Integer-indexed [[[DefineOwnProperty]]](https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects-defineownproperty-p-desc) is the only place where IsValidIntegerIndex is checked without being immediately preceded by an IsDetachedBuffer check, and I believe this to be a mistake.

The issue is that [IsValidIntegerIndex](https://tc39.es/ecma262/#sec-isvalidintegerindex) checks [[ArrayLength]] of the typed array itself, which is unaffected by calling [DetachArrayBuffer](https://tc39.es/ecma262/#sec-detacharraybuffer). This seems very silly and may warrant a more significant refactor, but the immediate fix is to avoid calling IsValidIntegerIndex on a typed array with a detached buffer.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9159,11 +9159,7 @@
           1. Assert: _O_ is an Integer-Indexed exotic object.
           1. If Type(_P_) is String, then
             1. Let _numericIndex_ be ! CanonicalNumericIndexString(_P_).
-            1. If _numericIndex_ is not *undefined*, then
-              1. Let _buffer_ be _O_.[[ViewedArrayBuffer]].
-              1. If IsDetachedBuffer(_buffer_) is *true*, return *false*.
-              1. If ! IsValidIntegerIndex(_O_, _numericIndex_) is *false*, return *false*.
-              1. Return *true*.
+            1. If _numericIndex_ is not *undefined*, return ! IsValidIntegerIndex(_O_, _numericIndex_).
           1. Return ? OrdinaryHasProperty(_O_, _P_).
         </emu-alg>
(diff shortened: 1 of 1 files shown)


bakkot (approved):
LGTM. Every other call to `IsValidIntegerIndex` is preceded by this call, so it makes sense to do it here as well. That said, it would perhaps make sense to move the `IsDetachedBuffer` check into `IsValidIntegerIndex`, so as to avoid this problem.

rkirsling:
Added a commit to address https://github.com/tc39/ecma262/pull/2164#issuecomment-711092235 and temporarily removed `Editorial:` from the title (feel free to re-add if appropriate).

syg (approved):
Looks great, these both look right to me.

michaelficarra (approved):

VR0ME (approved):
[Name](BrechtVrolix)

bakkot:
Per editor call: let's call this normative, but we'll land it without consensus on the belief that it was intended to be a part of #2164. @rkirsling we also think it makes sense to move move the `IsDetachedBuffer` calls which precede each call to `IsValidIntegerIndex` into `IsValidIntegerIndex` itself; do you want to do this or shall I take care of it?

rkirsling:
Done. Thanks for your patience. :bow:

PR #2210, opened 2020-10-21
Normative: IntegerIndexedElementSet should always indicate success

### LINKED ISSUE #2208: Is IntegerIndexedElementSet returning `false` for OOB access web-compatible?
Please consider the following code:

```js
"use strict";
const tA = new Int8Array([0, 1, 2]);
tA[42] = 1; // normal completion
```

Step 6 of [`IntegerIndexedElementSet`](https://tc39.es/ecma262/#sec-integerindexedelementset) returns `false` for out-of-bounds indices, which [`Set`](https://tc39.es/ecma262/#sec-set-o-p-v-throw) should throw a `TypeError` for.
However, none of the tested runtimes (V8, SM, JSC) throws.

```js
const tA = new Int8Array([0, 1, 2]);
Reflect.set(tA, 42, 1);
// V8: true
// SM: false
// JSC: false
```

If throwing isn't web-compatible, should `true` be returned instead to maintain strict-mode assignment semantics?

Note: this issue wasn't affected by / isn't related to recent detached buffers change.

_cc_ @rkirsling @syg

rkirsling:
Resolves #2208.

This behavior is necessitated by web reality and tied to #2164 (although the OOB path is technically separate).

@syg and I chatted about whether the (non-normative) description of [[Set]] in [Table 6](https://tc39.es/ecma262/#table-essential-internal-methods) needs to be updated and concluded "no" for two reasons:
1. It is extremely difficult to update this verbiage to be exhaustive without explicit mention of typed arrays.
2. The meaning of the verb "set" is already ambiguous thanks to userland setters. In other words, it can be argued that the property value was "set" just by virtue of having entered IntegerIndexedElementSet.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9180,9 +9180,7 @@
               1. If _Desc_ has a [[Configurable]] field and if _Desc_.[[Configurable]] is *false*, return *false*.
               1. If _Desc_ has an [[Enumerable]] field and if _Desc_.[[Enumerable]] is *false*, return *false*.
               1. If _Desc_ has a [[Writable]] field and if _Desc_.[[Writable]] is *false*, return *false*.
-              1. If _Desc_ has a [[Value]] field, then
-                1. Let _value_ be _Desc_.[[Value]].
-                1. Return ? IntegerIndexedElementSet(_O_, _numericIndex_, _value_).
+              1. If _Desc_ has a [[Value]] field, perform ? IntegerIndexedElementSet(_O_, _numericIndex_, _Desc_.[[Value]]).
               1. Return *true*.
           1. Return ! OrdinaryDefineOwnProperty(_O_, _P_, _Desc_).
         </emu-alg>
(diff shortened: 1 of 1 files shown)


bakkot:
I note that this operation can no longer return any normal completion except `true`. I wonder if it should be refactored to not return anything.

Also, this probably merits a NOTE to the effect of "This operation always appears to succeed, although it does not have any effect when writing past the end of the TypedArray or to a TypedArray which is backed by a detached ArrayBuffer".

rkirsling:
@bakkot I definitely agree about the note, but I'm not certain that the return value refactor is clearer. I've added it here as a third commit, so we can discuss whether we want to keep it or not.

bakkot:
@rkirsling This got consensus, but I'm not sure if there's tests for it yet. Do you know?

rkirsling:
@rwaldron's been helping me with the tests for #2164 and its various ripple effects. Rick, would you have time to look at this (and #2221)? Lemme know if it's getting to be too much. 😅

bakkot (approved):

ljharb (approved):

syg (approved):
lgtm, thanks!

PR #2211, opened 2020-10-22
Editorial: Move and fold the super constructor check into GetSuperConstructor

syg:
The separate GetSuperConstructor call and IsConstructor check is
confusing, and caused a bug in V8, where the IsConstructor check was
folded into GetSuperConstructor and is thrown before argument
evaluation.

The actual property get of the active function's [[Prototype]] in
GetSuperConstructor is not observable, so it is editorial to fold the
IsConstructor check into GetSuperConstructor and move the call to be
after the argument evaluation.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13689,9 +13689,8 @@
         <emu-alg>
           1. Let _newTarget_ be GetNewTarget().
           1. Assert: Type(_newTarget_) is Object.
-          1. Let _func_ be ! GetSuperConstructor().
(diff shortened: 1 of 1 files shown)


michaelficarra:
I don't think this is editorial. `ArgumentListEvaluation of |Arguments|` can have effects (or throw), and it wouldn't have previously been reached.

syg:
Oh wait, no, this is wrong in another way, because argument evaluation can set the __proto__ of the function object. Hmm...

syg:
Closing for now, this is incorrect.

PR #2215, opened 2020-10-27
Normative: add normative optional styling css override

### LINKED ISSUE #2214: WeakRef as specified violates tc39 consensus
https://tc39.es/ecma262/#sec-weak-ref.prototype.constructor does not mention that the `WeakRef.prototype.constructor` property itself is normative optional. This is a bug. The tc39 consensus was that it is normative optional, i.e., that a conforming implementation may omit it. 

Making it mandatory violates a safety property: The `WeakRef` constructor is designed to be separately deniable or virtualizable because it creates a huge side channel through the garbage collector. Unlike most side channels, this one does *not* rely on being able to measure duration. Access to the `WeakRef` constructor enables detecting of the collection of *any* object. A weakref instance should enable only the detection of whether that weakref's target is collected. A conforming platform must be able to omit the `WeakRef.prototype.constructor` property in order to prevent the privilege escalation from detecting the collection of one object to detecting the collection of any object.

During the process from consensus to final merging into the spec, it would be interesting to know when this got dropped. Adding it back in will not invalidate any currently valid engine implementations. Any implementation that conforms to the spec as currently written will also conform to the spec that we agreed on. (Since we need to add it back in anyway, there's a different configuration of `WeakRef.prototype.constructor` that I would now prefer, that would be closer to the semantics of the spec as currently written. But we should first restore the consensus state before considering alternatives.)

Attn @dtribble @kriskowal @FUDCo @kumavis @danfinlay @phoddie @patrick-soquet @warner @bmeck @guybedford @tschneidereit @syg @littledan @codehag

codehag:
This addresses the problem brought up in #2214 -- where the styles for normative optional changes were forgotten (by me) when the weakrefs merge pr was made. 

This likely isn't the best way to do this. This pr is here in case we need a stop gap solution until we clarify how best to add these styles in an accessible way. cc @bterlson, @ljharb, @michaelficarra, @syg and @bakkot.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -57,6 +57,17 @@
   emu-table td {
     background: #fff;
   }
+  [normative-optional] {
+    border-left: 5px solid #ff6600;
+    padding: .5em;
+    display: block;
+    background: #ffeedd;
+  }
+  [normative-optional]:before {
+    display: block;
+    color: #884400;
+    content: "NORMATIVE OPTIONAL";
+  }
 </style>
 </head>
 <body>
@@ -142,6 +153,11 @@
   <p>A conforming implementation of ECMAScript may support program and regular expression syntax not described in this specification. In particular, a conforming implementation of ECMAScript may support program syntax that makes use of any &ldquo;future reserved words&rdquo; noted in subclause <emu-xref href="#sec-keywords-and-reserved-words"></emu-xref> of this specification.</p>
   <p>A conforming implementation of ECMAScript must not implement any extension that is listed as a Forbidden Extension in subclause <emu-xref href="#sec-forbidden-extensions"></emu-xref>.</p>
   <p>A conforming implementation of ECMAScript must not redefine any facilities that are not implementation-defined, implementation-approximated, or host-defined.</p>
+  <p>A conforming implementation of ECMAScript may choose to implement or not implement Normative Optional subclauses. If any Normative Optional behaviour is implemented, all of the behaviour in the containing Normative Optional clause must be implemented. A Normative Optional clause is denoted in this specification with the words "Normative Optional" in a coloured box, as shown below.</p>
+  <emu-clause id="sec-conformance.normative-optional" type="example" normative-optional>
+    <h1>Example Clause Heading</h1>
+    <p>Example clause contents.</p>
+  </emu-clause>
 </emu-clause>
 
 <emu-clause id="sec-normative-references">


michaelficarra:
Screenshot:

image

syg:
We should be explicit about what the unit of optionality is. @codehag Could you add a sentence explicitly defining the unit of optionality to be everything within a Normative Optional colored box?

codehag:
let me know if that is what you had in mind...

Screenshot 2020-12-03 at 16 58 58

Also, were there any thoughts about the point @erights raised about this needing to be visible even without css, or is this not likely to be an issue? One suggestion I can think of is that any section (ie 2.1 here) is followed by "Normative" or something to this effect. Then we can drop the prepend text from the css?

ljharb:
We were thinking that ecmarkup can insert the full “normative optional” text explicitly.

spec.html
```diff
@@ -142,6 +153,11 @@ <h1>Conformance</h1>
   <p>A conforming implementation of ECMAScript may support program and regular expression syntax not described in this specification. In particular, a conforming implementation of ECMAScript may support program syntax that makes use of any &ldquo;future reserved words&rdquo; noted in subclause <emu-xref href="#sec-keywords-and-reserved-words"></emu-xref> of this specification.</p>
   <p>A conforming implementation of ECMAScript must not implement any extension that is listed as a Forbidden Extension in subclause <emu-xref href="#sec-forbidden-extensions"></emu-xref>.</p>
   <p>A conforming implementation of ECMAScript must not redefine any facilities that are not implementation-defined, implementation-approximated, or host-defined.</p>
+  <p>A conforming implementation of ECMAScript may choose to implement or not implement Normative Optional subclauses. However if Normative Optional behavior is implemented in some form, it must be implemented in accordance to the Normative Optional clause. A Normative Optional clause is defined in the spec within a Normative Optional colored box.</p>
```
michaelficarra, line 156:
```suggestion
  <p>A conforming implementation of ECMAScript may choose to implement or not implement Normative Optional subclauses. If any Normative Optional behaviour is implemented, all of the behaviour in the containing Normative Optional clause must be implemented. A Normative Optional clause is denoted in the spec with the words "Normative Optional" in a coloured box, as shown below.</p>
```

jmdyck: The spec doesn't refer to itself as "the spec". Usually it says "this specification", occasionally "this document".

michaelficarra: @jmdyck Updated.

syg: +1 to @michaelficarra's edit.

spec.html
```diff
@@ -142,6 +153,11 @@ <h1>Conformance</h1>
...
+  <emu-clause id="sec-conformance.normative-optional" normative-optional>
+    <h1>Normative Optional Example Clause</h1>
+    <p>Example of Normative Optional clause</p>
```
michaelficarra, line 159:
```suggestion
    <h1>Example Clause Heading</h1>
    <p>Example clause contents.</p>
```

Just to keep the words "normative optional" out of the example otherwise.

spec.html
```diff
@@ -142,6 +153,11 @@ <h1>Conformance</h1>
   <p>A conforming implementation of ECMAScript may support program and regular expression syntax not described in this specification. In particular, a conforming implementation of ECMAScript may support program syntax that makes use of any &ldquo;future reserved words&rdquo; noted in subclause <emu-xref href="#sec-keywords-and-reserved-words"></emu-xref> of this specification.</p>
   <p>A conforming implementation of ECMAScript must not implement any extension that is listed as a Forbidden Extension in subclause <emu-xref href="#sec-forbidden-extensions"></emu-xref>.</p>
   <p>A conforming implementation of ECMAScript must not redefine any facilities that are not implementation-defined, implementation-approximated, or host-defined.</p>
+  <p>A conforming implementation of ECMAScript may choose to implement or not implement Normative Optional subclauses. However if Normative Optional behavior is implemented in some form, it must be implemented in accordance to the Normative Optional clause. A Normative Optional clause is defined in the spec within a Normative Optional colored box.</p>
+  <emu-clause id="sec-conformance.normative-optional" normative-optional>
```
michaelficarra, line 157:
```suggestion
  <emu-clause id="sec-conformance.normative-optional" type="example" normative-optional>
```

@bakkot I know we use the `type="example"` attribute for `emu-grammar` elements. Should we also be using it here?

bakkot: I believe it wouldn't affect anything the current setup, but it wouldn't hurt.

codehag:
I applied the changes manually. github seems to have some issues applying comment suggestions atm.

syg:
> I applied the changes manually. github seems to have some issues applying comment suggestions atm.

That feature has never worked right for this repo, something to do with permissions like if the suggester has write access or something, it breaks.

I also find that suggestions generate these unreadable emails since the spec source has very long lines.

michaelficarra:
@syg @codehag I think it might have something to do with `spec.html` being very large, in the same way that expanding the context around a diff to `spec.html` fails.

michaelficarra (approved):
This WFM for now. When we get ecmarkup generating the "normative optional" text for these sections, we can remove the `:before` pseudo-element selector.

bakkot (approved):
LGTM, thanks! I'll upstream the CSS into ecmarkup at some point.

syg (approved):

codehag:
I'm not quite used to the work flow: do I merge after 3 approvals, or do you? Or do we need one more?

michaelficarra:
@codehag We will handle it.

ljharb (approved):

PR #2219, opened 2020-10-30
Editorial: Simplify Promise.all and Promise.any

chicoxyzzy:
`[[AlreadyCalled]]` fields in `Promise.all` and `Promise.any` doesn't have to be a Record

Discussion https://github.com/tc39/ecma262/pull/2040#discussion_r454699178

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -39947,7 +39947,7 @@ THH:mm:ss.sss
               1. Let _nextPromise_ be ? Call(_promiseResolve_, _constructor_, &laquo; _nextValue_ &raquo;).
               1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-promise.all-resolve-element-functions" title></emu-xref>.
               1. Let _resolveElement_ be ! CreateBuiltinFunction(_steps_, &laquo; [[AlreadyCalled]], [[Index]], [[Values]], [[Capability]], [[RemainingElements]] &raquo;).
-              1. Set _resolveElement_.[[AlreadyCalled]] to the Record { [[Value]]: *false* }.
+              1. Set _resolveElement_.[[AlreadyCalled]] to *false*.
               1. Set _resolveElement_.[[Index]] to _index_.
               1. Set _resolveElement_.[[Values]] to _values_.
               1. Set _resolveElement_.[[Capability]] to _resultCapability_.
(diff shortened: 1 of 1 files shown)


Jack-Works:
It seems like Editorial not normative

mathiasbynens (approved):
LGTM but I don't think this is a normative/observable change, in which case the commit message needs changing

spec.html
```diff
@@ -39912,8 +39912,8 @@ <h1>`Promise.all` Resolve Element Functions</h1>
           <emu-alg>
             1. Let _F_ be the active function object.
             1. Let _alreadyCalled_ be _F_.[[AlreadyCalled]].
-            1. If _alreadyCalled_.[[Value]] is *true*, return *undefined*.
-            1. Set _alreadyCalled_.[[Value]] to *true*.
+            1. If _alreadyCalled_ is *true*, return *undefined*.
+            1. Set _alreadyCalled_ to *true*.
```
shvaikalesh, line 39916: Since `alreadyCalled` is now a primitive, setting it won't affect `F.[[AlreadyCalled]]`, right?
I believe `F.[[AlreadyCalled]]` should be assigned to directly.

spec.html
```diff
@@ -39911,9 +39911,8 @@ <h1>`Promise.all` Resolve Element Functions</h1>
           <p>When a `Promise.all` resolve element function is called with argument _x_, the following steps are taken:</p>
           <emu-alg>
             1. Let _F_ be the active function object.
-            1. Let _alreadyCalled_ be _F_.[[AlreadyCalled]].
-            1. If _alreadyCalled_.[[Value]] is *true*, return *undefined*.
-            1. Set _alreadyCalled_.[[Value]] to *true*.
+            1. If _F_.[[AlreadyCalled]]. is *true*, return *undefined*.
```
shvaikalesh, line 39914: nit:
```suggestion
            1. If _F_.[[AlreadyCalled]] is *true*, return *undefined*.
```

chicoxyzzy: thanks!

bakkot (approved):

ljharb (approved):

syg (approved):

michaelficarra (approved):

PR #2221, opened 2020-11-03
Normative: Give %TypedArray% methods explicit algorithms.

rkirsling:
This ought to be editorial, but it's possible that it requires consensus due to the underspecification at play.

There are 12 %TypedArray% methods which refer to their Array counterparts without giving an explicit algorithm. This isn't so much delegation as it is hand-waving: [`map`](https://tc39.es/ecma262/#sec-%typedarray%.prototype.map) and [`filter`](https://tc39.es/ecma262/#sec-%typedarray%.prototype.filter) happen to be spelled out and, in particular, do not perform per-iteration HasProperty checks (because typed arrays can't have holes). This is not called out in the current explanatory text for, e.g., [`reduce`](https://tc39.es/ecma262/#sec-%typedarray%.prototype.reduce), but I argue that it is being assumed and not consciously omitted.

The foregoing behavior in SM and V8 diverges here—neither were throwing on set/get for integer-indexed elements with a detached buffer, but V8 furthermore chose to stop iterating (even for the fully specified `map` and `filter`). @syg has suggested that V8 is willing to change this behavior.

As such, this PR replaces the explanatory text of the following %TypedArray% methods with explicit algorithms.

- callback-taking methods (iteration is observable):
    `every`, `find`, `findIndex`, `forEach`, `reduce`, `reduceRight`, `some`

- other methods (iteration not observable):
    `includes`, `indexOf`, `join`, `lastIndexOf`, `reverse`

The changes from their Array counterparts are nothing other than calling ValidateTypedArray on the `this` value, using [[ArrayLength]] for length, and removing all HasProperty checks.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -32064,7 +32064,7 @@ THH:mm:ss.sss
             1. Let _k_ be _len_ + _n_.
             1. If _k_ &lt; 0, set _k_ to 0.
           1. Repeat, while _k_ &lt; _len_,
-            1. Let _elementK_ be the result of ? Get(_O_, ! ToString(𝔽(_k_))).
+            1. Let _elementK_ be ? Get(_O_, ! ToString(𝔽(_k_))).
             1. If SameValueZero(_searchElement_, _elementK_) is *true*, return *true*.
             1. Set _k_ to _k_ + 1.
           1. Return *false*.
@@ -33328,8 +33328,23 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-%typedarray%.prototype.every">
         <h1>%TypedArray%.prototype.every ( _callbackfn_ [ , _thisArg_ ] )</h1>
-        <p>%TypedArray%`.prototype.every` is a distinct function that implements the same algorithm as `Array.prototype.every` as defined in <emu-xref href="#sec-array.prototype.every"></emu-xref> except that the *this* value's [[ArrayLength]] internal slot is accessed in place of performing a [[Get]] of *"length"*. The implementation of the algorithm may be optimized with the knowledge that the *this* value is an object that has a fixed length and whose integer-indexed properties are not sparse. However, such optimization must not introduce any observable changes in the specified behaviour of the algorithm and must take into account the possibility that calls to _callbackfn_ may cause the *this* value to become detached.</p>
-        <p>This function is not generic. ValidateTypedArray is applied to the *this* value prior to evaluating the algorithm. If its result is an abrupt completion that exception is thrown instead of evaluating the algorithm.</p>
+        <p>The interpretation and use of the arguments of %TypedArray%`.prototype.every` are the same as for `Array.prototype.every` as defined in <emu-xref href="#sec-array.prototype.every"></emu-xref>.</p>
+        <p>When the `every` method is called with one or two arguments, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If IsCallable(_callbackfn_) is *false*, throw a *TypeError* exception.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Let _kValue_ be ! Get(_O_, _Pk_).
+            1. Let _testResult_ be ! ToBoolean(? Call(_callbackfn_, _thisArg_, &laquo; _kValue_, 𝔽(_k_), _O_ &raquo;)).
+            1. If _testResult_ is *false*, return *false*.
+            1. Set _k_ to _k_ + 1.
+          1. Return *true*.
+        </emu-alg>
+        <p>This function is not generic. The *this* value must be an object with a [[TypedArrayName]] internal slot.</p>
       </emu-clause>
 
       <emu-clause id="sec-%typedarray%.prototype.fill">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -34622,38 +34637,142 @@ <h1>%TypedArray%.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If _separator_ is *undefined*, let _sep_ be the single-element String *","*.
+          1. Else, let _sep_ be ? ToString(_separator_).
+          1. Let _R_ be the empty String.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. If _k_ &gt; 0, set _R_ to the string-concatenation of _R_ and _sep_.
+            1. Let _element_ be ? Get(_O_, ! ToString(𝔽(_k_))).
```
syg, line 34769: `! Get`

spec.html
```diff
@@ -34622,38 +34637,142 @@ <h1>%TypedArray%.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. If _element_ is *undefined* or *null*, let _next_ be the empty String; otherwise, let _next_ be ? ToString(_element_).
```
syg, line 34770: `! ToString`

spec.html
```diff
@@ -34622,38 +34637,142 @@ <h1>%TypedArray%.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
-        <p>This function is not generic. ValidateTypedArray is applied to the *this* value prior to evaluating the algorithm. If its result is an abrupt completion that exception is thrown instead of evaluating the algorithm.</p>
+        <p>The interpretation and use of the arguments of %TypedArray%`.prototype.join` are the same as for `Array.prototype.join` as defined in <emu-xref href="#sec-array.prototype.join"></emu-xref>.</p>
+        <p>When the `join` method is called with one argument _separator_, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If _separator_ is *undefined*, let _sep_ be the single-element String *","*.
+          1. Else, let _sep_ be ? ToString(_separator_).
```
syg, line 34764: In a similar vein, with this PR, `ta.join(detachPoison)` is now specced to return a string with a length-1 number of `,`s. V8 already matches this, SM prints a bunch of `undefined`.

```javascript
function detach(ab) {
  if (ArrayBuffer.transfer) {
    ArrayBuffer.transfer(ab);
    ArrayBuffer.detach(ab);
    detachArrayBuffer(ab);
    transferArrayBuffer(ab)
      w.postMessage(ab, [ab]);
      w.terminate();
    }
  } else {
  }
}

let ta = new Uint8Array(2);
print(ta.join({ toString() { detach(ta.buffer); return ','; } }));
```

```
#### jsc

TypeError: Underlying ArrayBuffer has been detached from the view

#### sm
undefined,undefined

#### v8
,
```
```

File: spec.html [hunk 4]
```diff
@@ -34668,8 +34787,27 @@ <h1>%TypedArray%.prototype.keys ( )</h1>
...
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If _len_ is 0, return *-1*<sub>𝔽</sub>.
+          1. If _fromIndex_ is present, let _n_ be ? ToIntegerOrInfinity(_fromIndex_); else let _n_ be _len_ - 1.
+          1. If _n_ is -&infin;, return *-1*<sub>𝔽</sub>.
+          1. If _n_ &ge; 0, then
+            1. Let _k_ be min(_n_, _len_ - 1).
+          1. Else,
+            1. Let _k_ be _len_ + _n_.
+          1. Repeat, while _k_ &ge; 0,
+            1. Let _elementK_ be ? Get(_O_, ! ToString(𝔽(_k_))).
```
syg, line 34804: `! Get`

spec.html
```diff
@@ -34668,8 +34787,27 @@ <h1>%TypedArray%.prototype.keys ( )</h1>
...
-        <p>This function is not generic. ValidateTypedArray is applied to the *this* value prior to evaluating the algorithm. If its result is an abrupt completion that exception is thrown instead of evaluating the algorithm.</p>
+        <p>The interpretation and use of the arguments of %TypedArray%`.prototype.lastIndexOf` are the same as for `Array.prototype.lastIndexOf` as defined in <emu-xref href="#sec-array.prototype.lastindexof"></emu-xref>.</p>
+        <p>When the `lastIndexOf` method is called with one or two arguments, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If _len_ is 0, return *-1*<sub>𝔽</sub>.
+          1. If _fromIndex_ is present, let _n_ be ? ToIntegerOrInfinity(_fromIndex_); else let _n_ be _len_ - 1.
```
syg, line 34797: Similar issue as with `indexOf`. For TA of length 8,

```
#### jsc

TypeError: Underlying ArrayBuffer has been detached from the view

#### sm
7

#### v8
-1
```

spec.html
```diff
@@ -34711,20 +34849,80 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _k_ be 0.
+          1. Let _accumulator_ be *undefined*.
+          1. If _initialValue_ is present, then
+            1. Set _accumulator_ to _initialValue_.
+          1. Else,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Set _accumulator_ to ? Get(_O_, _Pk_).
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Let _kValue_ be ? Get(_O_, _Pk_).
```
syg, line 34869: `! Get`

spec.html
```diff
@@ -34711,20 +34849,80 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If IsCallable(_callbackfn_) is *false*, throw a *TypeError* exception.
+          1. If _len_ = 0 and _initialValue_ is not present, throw a *TypeError* exception.
+          1. Let _k_ be 0.
+          1. Let _accumulator_ be *undefined*.
+          1. If _initialValue_ is present, then
+            1. Set _accumulator_ to _initialValue_.
+          1. Else,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Set _accumulator_ to ? Get(_O_, _Pk_).
```
syg, line 34866: `! Get`

spec.html
```diff
@@ -34711,20 +34849,80 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If IsCallable(_callbackfn_) is *false*, throw a *TypeError* exception.
+          1. If _len_ is 0 and _initialValue_ is not present, throw a *TypeError* exception.
+          1. Let _k_ be _len_ - 1.
+          1. Let _accumulator_ be *undefined*.
+          1. If _initialValue_ is present, then
+            1. Set _accumulator_ to _initialValue_.
+          1. Else,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Set _accumulator_ to ? Get(_O_, _Pk_).
```
syg, line 34893: `! Get`

spec.html
```diff
@@ -34711,20 +34849,80 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Repeat, while _k_ &ge; 0,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Let _kValue_ be ? Get(_O_, _Pk_).
```
syg, line 34896: `! Get`

spec.html
```diff
@@ -34711,20 +34849,80 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. Let _middle_ be floor(_len_ / 2).
+          1. Let _lower_ be 0.
+          1. Repeat, while _lower_ &ne; _middle_,
+            1. Let _upper_ be _len_ - _lower_ - 1.
+            1. Let _upperP_ be ! ToString(𝔽(_upper_)).
+            1. Let _lowerP_ be ! ToString(𝔽(_lower_)).
+            1. Let _lowerValue_ be ? Get(_O_, _lowerP_).
```
syg, line 34918: `! Get`

spec.html
```diff
@@ -34711,20 +34849,80 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+            1. Let _upperValue_ be ? Get(_O_, _upperP_).
```
syg, line 34919: `! Get`

spec.html
```diff
@@ -34878,8 +35076,23 @@ <h1>%TypedArray%.prototype.slice ( _start_, _end_ )</h1>
...
+        <p>When the `some` method is called with one or two arguments, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If IsCallable(_callbackfn_) is *false*, throw a *TypeError* exception.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Let _kValue_ be ? Get(_O_, _Pk_).
```
syg, line 35089: `! Get`

syg (commented):
This is really great, thank you very much @rkirsling.

IMO this is normative because we should make a conscious decision what methods like `indexOf` ought to return if the TA is sneakily detached.

rkirsling:
@syg brought up `array.length = 0;` as an analogous operation to buffer detachment, and I'm thinking that's probably the best way to view the intent here. Thus in the above examples, `includes` should return `true` and `join` should return `","`, but we should restore the HasProperty checks for `indexOf` / `lastIndexOf` such that they return `-1`.

(It is also possible to argue for leaving those HasProperty checks off, on the basis that the length is cached at that moment anyway; I wouldn't object either way but I feel like this would be a more "unique" behavior.)

syg (approved):
lgtm editorially

syg:
This got consensus.

ExE-Boss:
How about extracting the common portions into abstract operations and calling that instead?

---

That is sort‑of how @engine262 [implements this](https://github.com/engine262/engine262/blob/main/src/intrinsics/ArrayPrototypeShared.mjs),

spec.html
```diff
@@ -34622,38 +34637,144 @@ <h1>%TypedArray%.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If IsCallable(_predicate_) is *false*, throw a *TypeError* exception.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Let _kValue_ be ! Get(_O_, _Pk_).
+            1. Let _testResult_ be ! ToBoolean(? Call(_predicate_, _thisArg_, &laquo; _kValue_, 𝔽(_k_), _O_ &raquo;)).
+            1. If _testResult_ is *true*, return _k_.
+            1. Set _k_ to _k_ + 1.
+          1. Return -1.
```
bakkot, line 34675:
```suggestion
          1. Return *-1*<sub>𝔽</sub>.
```

spec.html
```diff
@@ -34622,38 +34637,144 @@ <h1>%TypedArray%.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If IsCallable(_predicate_) is *false*, throw a *TypeError* exception.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Let _kValue_ be ! Get(_O_, _Pk_).
+            1. Let _testResult_ be ! ToBoolean(? Call(_predicate_, _thisArg_, &laquo; _kValue_, 𝔽(_k_), _O_ &raquo;)).
+            1. If _testResult_ is *true*, return _k_.
```
bakkot, line 34673:
```suggestion
            1. If _testResult_ is *true*, return 𝔽(_k_).
```

spec.html
```diff
@@ -34711,20 +34853,80 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If IsCallable(_callbackfn_) is *false*, throw a *TypeError* exception.
+          1. If _len_ = 0 and _initialValue_ is not present, throw a *TypeError* exception.
+          1. Let _k_ be 0.
+          1. Let _accumulator_ be *undefined*.
+          1. If _initialValue_ is present, then
+            1. Set _accumulator_ to _initialValue_.
+          1. Else,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Set _accumulator_ to ! Get(_O_, _Pk_).
```
bakkot, line 34870: You need to set `_k_` to `1` at the end of this branch, so that the first element does not get used twice.

spec.html
```diff
@@ -34711,20 +34853,80 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If IsCallable(_callbackfn_) is *false*, throw a *TypeError* exception.
+          1. If _len_ is 0 and _initialValue_ is not present, throw a *TypeError* exception.
+          1. Let _k_ be _len_ - 1.
+          1. Let _accumulator_ be *undefined*.
+          1. If _initialValue_ is present, then
+            1. Set _accumulator_ to _initialValue_.
+          1. Else,
+            1. Let _Pk_ be ! ToString(𝔽(_k_)).
+            1. Set _accumulator_ to ! Get(_O_, _Pk_).
```
bakkot, line 34897: As with `reduce`, you need to set `_k_` to `_k_ - 1` at the end of this branch.

spec.html
```diff
@@ -34711,20 +34853,80 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Let _middle_ be floor(_len_ / 2).
+          1. Let _lower_ be 0.
+          1. Repeat, while _lower_ &ne; _middle_,
+            1. Let _upper_ be _len_ - _lower_ - 1.
+            1. Let _upperP_ be ! ToString(𝔽(_upper_)).
+            1. Let _lowerP_ be ! ToString(𝔽(_lower_)).
+            1. Let _lowerValue_ be ! Get(_O_, _lowerP_).
+            1. Let _upperValue_ be ! Get(_O_, _upperP_).
+            1. Perform ? Set(_O_, _lowerP_, _upperValue_, *true*).
+            1. Perform ? Set(_O_, _upperP_, _lowerValue_, *true*).
```
bakkot, line 34925: As of https://github.com/tc39/ecma262/pull/2210 these two `Set`s will be infallible, right?

bakkot (commented):
Couple of things need addressing. Otherwise LGTM.

spec.html
rkirsling: Oh wow, oops! Thanks for catching that.

spec.html
rkirsling: That's correct. May as well address it right away here then.

bakkot:
Note that https://github.com/tc39/ecma262/pull/2210 should land first, so that the `! Set`s in this are correct.

spec.html
```diff
@@ -34622,38 +34637,144 @@ <h1>%TypedArray%.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If _separator_ is *undefined*, let _sep_ be the single-element String *","*.
+          1. Else, let _sep_ be ? ToString(_separator_).
+          1. Let _R_ be the empty String.
+          1. Let _k_ be 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. If _k_ &gt; 0, set _R_ to the string-concatenation of _R_ and _sep_.
+            1. Let _element_ be ! Get(_O_, ! ToString(𝔽(_k_))).
+            1. If _element_ is *undefined* or *null*, let _next_ be the empty String; otherwise, let _next_ be ! ToString(_element_).
```
bakkot, line 34772: Is it actually be possible for `_element_` to be `*null*`? It looks to me like `_O_` is always an integer-indexed exotic object and `ToString(𝔽(_k_))` is always a [CanonicalNumericIndexString](https://tc39.es/ecma262/#sec-canonicalnumericindexstring), which means the above `Get` is [going to](https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects-get-p-receiver) go through [IntegerIndexedElementGet](https://tc39.es/ecma262/#sec-integerindexedelementget), which can only return numerics or `*undefined*`.

If that's correct, the `or *null*` here should be removed, since it's misleading.

spec.html
```diff
@@ -34711,20 +34853,82 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+        <p>The interpretation and use of the arguments of %TypedArray%`.prototype.reduce` are the same as for `Array.prototype.reduce` as defined in <emu-xref href="#sec-array.prototype.reduce"></emu-xref>.</p>
+        <p>When the `reduce` method is called with one or two arguments, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. If IsCallable(_callbackfn_) is *false*, throw a *TypeError* exception.
+          1. If _len_ = 0 and _initialValue_ is not present, throw a *TypeError* exception.
+          1. Let _k_ be 0.
+          1. Let _accumulator_ be *undefined*.
```
bakkot, line 34865: There's a bit of weirdness here: this never gets used, since both branches below set it. The prevailing convention in the spec is that it should just be declared in those branches instead.

However, that matches Array.prototype.reduce. In that case the situation is more complicated, because of holes. (Fun fact: `[,0].reduce(() => console.log('called'))` returns `0` and does not log `called`.) You still end up having `_accumulator_` be set in both branches if control exits normally, it's just that you have to reason about it a little more to know that.

I don't think this warrants changes, I just wanted to note it for posterity.

bakkot (approved):
LGTM other than my thing about `*null*`.

spec.html
```diff
@@ -34559,8 +34559,23 @@ <h1>%TypedArray%.prototype.entries ( )</h1>
...
-        <p>This function is not generic. ValidateTypedArray is applied to the *this* value prior to evaluating the algorithm. If its result is an abrupt completion that exception is thrown instead of evaluating the algorithm.</p>
+        <p>The interpretation and use of the arguments of %TypedArray%`.prototype.every` are the same as for `Array.prototype.every` as defined in <emu-xref href="#sec-array.prototype.every"></emu-xref>.</p>
+        <p>When the `every` method is called with one or two arguments, the following steps are taken:</p>
```
michaelficarra, line 34563: I prefer phrasing like

```suggestion
        <p>The `every` method performs the following steps when called:</p>
```

since it's true regardless of how many arguments are given. I see that you've copied this phrasing from the `Array.prototype` methods. We should change those, too. Not sure whether we should do that before or after this PR. Either way, I don't think we should land the phrasing you have here.

spec.html
```diff
@@ -34622,38 +34637,144 @@ <h1>%TypedArray%.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Assert: If _fromIndex_ is *undefined*, then _n_ is 0.
+          1. If _n_ is +&infin;, return *false*.
+          1. Else if _n_ is -&infin;, set _n_ to 0.
+          1. If _n_ &ge; 0, then
+            1. Let _k_ be _n_.
+          1. Else,
+            1. Let _k_ be _len_ + _n_.
+            1. If _k_ &lt; 0, set _k_ to 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _elementK_ be the result of ! Get(_O_, ! ToString(𝔽(_k_))).
```
michaelficarra, line 34719:
```suggestion
            1. Let _elementK_ be ! Get(_O_, ! ToString(𝔽(_k_))).
```

spec.html
rkirsling: > Either way, I don't think we should land the phrasing you have here.

I don't follow. What you're describing would constitute a wide-reaching editorial PR so it doesn't seem like it should hold up this one?

michaelficarra: It's not wide-reaching; we use the phrasing I suggest extensively throughout the spec. What I'm asking is that we change it in this PR because it's better to have the better phrasing even if we are inconsistent with the phrasing used in Array until we fix it there.

bakkot: @michaelficarra I would strongly prefer to match the Array methods for now and change them all later. This isn't the appropriate place to do that kind of revision, even just scoped to the parts touched by this PR.

rkirsling: `When [^\.\n]+ is called with` occurs 210 times in the spec.

spec.html
```diff
@@ -34622,38 +34637,144 @@ <h1>%TypedArray%.prototype.filter ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+          1. Assert: If _fromIndex_ is *undefined*, then _n_ is 0.
+          1. If _n_ is +&infin;, return *-1*<sub>𝔽</sub>.
+          1. Else if _n_ is -&infin;, set _n_ to 0.
+          1. If _n_ &ge; 0, then
+            1. Let _k_ be _n_.
+          1. Else,
+            1. Let _k_ be _len_ + _n_.
+            1. If _k_ &lt; 0, set _k_ to 0.
+          1. Repeat, while _k_ &lt; _len_,
+            1. Let _kPresent_ be ! HasProperty(_O_, ! ToString(𝔽(_k_))).
```
michaelficarra, line 34746: Why are we doing this `HasProperty` check? In the Array implementations that use it, it's to skip holes. What is its purpose for TypedArrays? And why only `indexOf`/`lastIndexOf`?

rkirsling: Because that's what we asked for consensus on, as described [above](https://github.com/tc39/ecma262/pull/2221#issuecomment-721367858).

spec.html
michaelficarra: @rkirsling Did you look at those occurrences? The array methods are unique in the way they number the arguments. If you want to want name the arguments here, that would be fine too:

```suggestion
        <p>When the `every` method is called with argument _callbackfn_ and optional argument _thisArg_, the following steps are taken:</p>
```

@bakkot I don't know why we would prefer to introduce imprecise wording, even temporarily. We would match basically everything but the Array method phrasing, which is fine by me.

rkirsling: I see; I thought you were taking issue with listing off the arguments in general.

bakkot: @michaelficarra Using identical wording to the corresponding sections in Array means that readers don't need to consider why the wording is different, which is good.

Also, the decision of how to phrase this is an editorial one which unrelated to the focus of the PR. I would in general prefer not to pull in unrelated changes; we can fix this up *later* as part of fixing up [the rest](https://github.com/tc39/ecma262/pull/2294#issuecomment-773000412).

spec.html
```diff
@@ -34711,20 +34853,82 @@ <h1>%TypedArray%.prototype.map ( _callbackfn_ [ , _thisArg_ ] )</h1>
...
+        <p>The interpretation and use of the arguments of %TypedArray%`.prototype.reverse` are the same as for `Array.prototype.reverse` as defined in <emu-xref href="#sec-array.prototype.reverse"></emu-xref>.</p>
+        <p>When the `reverse` method is called, the following steps are taken:</p>
+        <emu-alg>
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. Let _middle_ be floor(_len_ / 2).
+          1. Let _lower_ be 0.
+          1. Repeat, while _lower_ &ne; _middle_,
+            1. Let _upper_ be _len_ - _lower_ - 1.
```
michaelficarra, line 34921: I would parenthesise this since it's not associative.

```suggestion
            1. Let _upper_ be (_len_ - _lower_) - 1.
```

But if the other editors would prefer to make this change in Array.prototype.reverse as well at the same time, we can leave it out for now.

spec.html
bakkot: Discussed this in the editor call. We'll try to fix all these up at some future time (opened https://github.com/tc39/ecma262/issues/2304 to track) and land this as-is for now.

michaelficarra (approved):
Approving since we'd like this to get in for the 2021 cut and it is an improvement over not using algorithm steps. Still, I feel it's 2 steps forward and 1 step backward due to the wording in the summary. Hopefully we can get at least the cases like these fixed up before 2021 is cut, but at least we'll for sure have these algorithms more clearly defined. Thanks for the contribution, @rkirsling.

PR #2227, opened 2020-11-13
Editorial: define how lookahead restrictions work at EOF

michaelficarra:
In #1966, we introduced a usage of `lookahead` at the end of a production without the behaviour of lookaheads being properly defined when there are no remaining tokens. I've defined them for both the positive and negative cases here. This may be overly pedantic, since I think there's only really one way that someone could've interpreted them.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -729,7 +729,7 @@
           `9`
       </emu-grammar>
       <p>If the phrase &ldquo;[empty]&rdquo; appears as the right-hand side of a production, it indicates that the production's right-hand side contains no terminals or nonterminals.</p>
-      <p>If the phrase &ldquo;[lookahead &notin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may not be used if the immediately following input token sequence is a member of the given _set_. The _set_ can be written as a comma separated list of one or two element terminal sequences enclosed in curly brackets. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all terminals to which that nonterminal could expand. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead &ne; _terminal_]&rdquo; may be used.</p>
+      <p>If the phrase &ldquo;[lookahead &notin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may not be used if there are remaining input tokens and the immediately following input token sequence is a member of the given _set_. The _set_ can be written as a comma separated list of one or two element terminal sequences enclosed in curly brackets. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all terminals to which that nonterminal could expand. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead &ne; _terminal_]&rdquo; may be used.</p>
       <p>For example, given the definitions:</p>
       <emu-grammar type="example">
         DecimalDigit :: one of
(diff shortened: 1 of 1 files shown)


ljharb (approved):
This seems correct but completely redundant; if there are no remaining tokens, then there is no following input sequence that can be a member of the given set (altho i suppose you'll reply with some kind of "the null set is a member of every set" argument)

claudepache:
The added precision is redundant in case of non-empty set and incorrect in case of empty set, although, of course, lookahead with empty set is not useful. If you want to split hairs, you can say that the definition is for non-empty sets.

claudepache:
[correction from my previous deleted comment]

The added precision is redundant in case of sets not containing the zero-length sequence of tokens, and incorrect in case of sets containing the zero-length sequence of tokens, although, of course, looking ahead the zero-length sequence of tokens is not useful. If you want to split hairs, you can add that the considered sets must contain only non-empty sequences of tokens.

michaelficarra:
@claudepache In what way is it incorrect? Also, I'm not understanding what you mean about the sets containing empty sequences of tokens. This PR is accounting for no tokens remaining in the input token sequence (so the lookahead, positive or negative, refers to nothing).

claudepache:
@michaelficarra Let **(void)** denote the sequence of zero tokens. Then  or <code>[ lookahead = **(void)** ]</code> or  
<code>[ lookahead ∈ { **(void)** } ]</code> is always trivially true.

michaelficarra:
@claudepache Yeah, and? I don't think that was disputed.

waldemarhorwat:
Lookaheads used to be for exactly one token, and this paragraph was worded correctly then. The wording got broken when we added multi-token lookaheads. It's still broken today with or without this pull.

Consider this: You have a lookahead restriction for the one-element set containing the token sequence `let [`.
The actual immediately following token sequence is: `let [ x ] …`
The lookahead restriction doesn't hold because `let [ x ] …` is _not_ a member of the set { `let [` }.

Please fix that problem.

waldemarhorwat:
Also, when constructing the fix to the issue I mentioned above, be aware that the concept of multiple-token lookahead in the syntactic grammar is not well-defined. If the lookahead sequence of characters is `await/"bc/;"…`, you don't know whether the `/` is a division symbol followed by the string `"bc/;"` or starts the regular expression `/"bc/` followed by a semicolon and the start of another string.

bakkot:
There's some discussion of the issue @waldemarhorwat describes in https://github.com/tc39/ecma262/pull/1828.

jmdyck:
> Also, when constructing the fix to the issue I mentioned above, be aware that the concept of multiple-token lookahead in the syntactic grammar is not well-defined. If the lookahead sequence of characters is `await/"bc/;"…`, you don't know whether the `/` is a division symbol followed by the string `"bc/;"` or starts the regular expression `/"bc/` followed by a semicolon and the start of another string.

It's true that, without any further information, you don't know. But a parser, parsing a particular source text, has lots more information. It knows the current "syntactic grammar context", and thus what InputElement goal symbol to use to get the next token. And, depending on what that token is, it knows what the context will be after it, and thus what InputElement goal symbol to use to get the next+1 token.

That is, I think "[lookahead != X Y]" means (roughly) you can't use this production if the next token is X and the token after that (for the context that would result from reading X) is Y.

Issue #800, opened 2017-02-06
Atomics and memory model specification questions/issues


anba:
24.4.1.3 GetWaiterList
- The term "semantic object" is not defined.
- The "store of WaiterList objects" and the lookup/retrieval operations could be more explicitly defined, e.g. similar to the Map.prototype methods.

24.4.*
- The term "semantic function" is not defined.
- The various operations should work on number values instead of list of byte values, e.g. the result of `add` when called with &laquo; 255 &raquo; and &laquo; 255 &raquo; is currently somewhat ambiguous, because the element type is not given which means i.a. that the length of the returned list of byte values isn't specified. 
- Also: It probably wouldn't hurt to have an explicit definition of the term "byte value".

27.7.3 Tear Free Reads
- Has "See Access Atomicity below.", but there's no "Access Atomicity" in the following paragraphs. Can we simply remove this reference and should it point to another section?

jmdyck:
re "semantic object": Probably a different term should be used. (Elsewhere, "specification type value" is used, but it's kind of clunky.)

At first glance, it seems like a WaiterList could be modelled as just a List. But apparently it also has a 'critical section' (which is never defined, by the way), so perhaps it would be better modelled as a Record with 2 fields.

jmdyck:
re "semantic function": I think this just means a function in the mathematical sense. The 'Algorithm Conventions' clause uses the term "mathematical function" for this. The RegExp clause uses the term "internal procedure" for something similar.

Alternatively, I think you could argue that the things referred to as "semantic functions" qualify as "abstract operations".

lars-t-hansen:
Related background:
https://github.com/tc39/ecmascript_sharedmem/issues/118
https://github.com/tc39/ecmascript_sharedmem/issues/120

I have no clear memory of where "semantic function" came from, but ISTR it is the result of an earlier bikeshed aimed at ensuring that there was no confusion between these functions and ES functions, at the time when the higher-order operation was introduced (the original spec draft just spelled out the algorithms individually).  Any term will suffice as long as that confusion is avoided.

(SIMD has the same concern, presumably.)

lars-t-hansen:
The critical section is orthogonal to the list, it just needs to be clear that the critical section protects that list.  The list can be any data structure, since it's only accessed inside the critical section.

[...]

PR #2240, opened 2020-12-02
Editorial: Describe `WaiterList` using `Record`s

ExE-Boss:
This makes `WaiterList`s use existing constructs, rather than the not‑used‑elsewhere “semantic object”.

This also makes [the `Atomics.waitAsync(…)` proposal](https://tc39.es/proposal-atomics-wait-async/) diff more cleanly.

## See also:
- <https://github.com/tc39/ecma262/issues/800>
- <https://github.com/tc39/ecma262/issues/1742>

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37244,33 +37244,53 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-getwaiterlist" aoid="GetWaiterList">
         <h1>GetWaiterList ( _block_, _i_ )</h1>
-        <p>A <dfn>WaiterList</dfn> is a semantic object that contains an ordered list of those agents that are waiting on a location (_block_, _i_) in shared memory; _block_ is a Shared Data Block and _i_ a byte offset into the memory of _block_. A WaiterList object also optionally contains a Synchronize event denoting the previous leaving of its critical section.</p>
-        <p>Initially a WaiterList object has an empty list and no Synchronize event.</p>
-        <p>The agent cluster has a store of WaiterList objects; the store is indexed by (_block_, _i_). WaiterLists are agent-independent: a lookup in the store of WaiterLists by (_block_, _i_) will result in the same WaiterList object in any agent in the agent cluster.</p>
-        <p>Each WaiterList has a <dfn>critical section</dfn> that controls exclusive access to that WaiterList during evaluation. Only a single agent may enter a WaiterList's critical section at one time. Entering and leaving a WaiterList's critical section is controlled by the abstract operations EnterCriticalSection and LeaveCriticalSection. Operations on a WaiterList&mdash;adding and removing waiting agents, traversing the list of agents, suspending and notifying agents on the list, setting and retrieving the Synchronize event&mdash;may only be performed by agents that have entered the WaiterList's critical section.</p>
+        <p>A <dfn id="waiterlist-record">WaiterList Record</dfn> is a Record value used to explain waiting and notification of agents via `Atomics.wait` and `Atomics.notify`. It has fields as defined by <emu-xref href="#table-waiterlist-record-fields"></emu-xref>.</p>
+        <emu-table id="table-waiterlist-record-fields" caption="WaiterList Record Fields">
+          <table>
+            <tbody>
+            <tr>
+              <th>Field Name</th>
+              <th>Value</th>
+              <th>Meaning</th>
+            </tr>
+            <tr>
+              <td>[[Waiters]]</td>
+              <td>A list of <emu-xref href="#agent">agents</emu-xref>.</td>
+              <td>The list of agents waiting on a memory location.</td>
+            </tr>
+            <tr>
+              <td>[[MostRecentLeaveEvent]]</td>
+              <td>Synchronize event | *undefined*</td>
+              <td>The event of the most recent leaving of its critical section, or *undefined* if its critical section has never been entered. </td>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -37244,76 +37244,96 @@ <h1>ValidateAtomicAccess ( _typedArray_, _requestIndex_ )</h1>
...
+          <table>
+            <tbody>
+            <tr>
+              <th>Field Name</th>
+              <th>Value</th>
+              <th>Meaning</th>
+            </tr>
+            <tr>
+              <td>[[Waiters]]</td>
+              <td>A list of <emu-xref href="#agent">agents</emu-xref>.</td>
```
syg, line 37258: This is a list of agent signifiers.

spec.html
```diff
@@ -37244,76 +37244,96 @@ <h1>ValidateAtomicAccess ( _typedArray_, _requestIndex_ )</h1>
...
+              <td>The list of agents waiting on a memory location.</td>
```
syg, line 37259:
```suggestion
              <td>The list of agent signifiers whose corresponding agents are waiting on the memory location.</td>
```

spec.html
```diff
@@ -37244,76 +37244,96 @@ <h1>ValidateAtomicAccess ( _typedArray_, _requestIndex_ )</h1>
...
+            </tr>
+            <tr>
+              <td>[[MostRecentLeaveEvent]]</td>
+              <td>Synchronize event | *undefined*</td>
```
syg, line 37263:
```suggestion
              <td>Synchronize event | ~empty~</td>
```

spec.html
```diff
@@ -37244,76 +37244,96 @@ <h1>ValidateAtomicAccess ( _typedArray_, _requestIndex_ )</h1>
...
       <emu-clause id="sec-entercriticalsection" aoid="EnterCriticalSection">
         <h1>EnterCriticalSection ( _WL_ )</h1>
-        <p>The abstract operation EnterCriticalSection takes argument _WL_ (a WaiterList). It performs the following steps when called:</p>
+        <p>The abstract operation EnterCriticalSection takes argument _WL_ (a WaiterList Record). It performs the following steps when called:</p>
         <emu-alg>
-          1. Assert: The calling agent is not in the critical section for any WaiterList.
+          1. Assert: The calling agent is not in the critical section for any WaiterList Record.
           1. Wait until no agent is in the critical section for _WL_, then enter the critical section for _WL_ (without allowing any other agent to enter).
-          1. If _WL_ has a Synchronize event, then
+          1. If _WL_.[[MostRecentLeaveEvent]] is not *undefined*, then
```
syg, line 37286:
```suggestion
          1. If _WL_.[[MostRecentLeaveEvent]] is not ~empty~, then
```

spec.html
```diff
@@ -37244,76 +37244,96 @@ <h1>ValidateAtomicAccess ( _typedArray_, _requestIndex_ )</h1>
 
       <emu-clause id="sec-getwaiterlist" aoid="GetWaiterList">
         <h1>GetWaiterList ( _block_, _i_ )</h1>
-        <p>A <dfn>WaiterList</dfn> is a semantic object that contains an ordered list of those agents that are waiting on a location (_block_, _i_) in shared memory; _block_ is a Shared Data Block and _i_ a byte offset into the memory of _block_. A WaiterList object also optionally contains a Synchronize event denoting the previous leaving of its critical section.</p>
-        <p>Initially a WaiterList object has an empty list and no Synchronize event.</p>
```
syg, line 37248: This paragraph is important and need to be captured, see comment below.

spec.html
```diff
@@ -37244,76 +37244,96 @@ <h1>ValidateAtomicAccess ( _typedArray_, _requestIndex_ )</h1>
...
         <p>The abstract operation GetWaiterList takes arguments _block_ (a Shared Data Block) and _i_ (a non-negative integer). It performs the following steps when called:</p>
         <emu-alg>
           1. Assert: _block_ is a Shared Data Block.
           1. Assert: _i_ and _i_ + 3 are valid byte offsets within the memory of _block_.
           1. Assert: _i_ is divisible by 4.
-          1. Return the WaiterList that is referenced by the pair (_block_, _i_).
+          1. Return the WaiterList Record that is referenced by the pair (_block_, _i_).
```
syg, line 37276: This part is a tricky. The "nice" thing about WaiterList being a handwavy magical "semantic object" was that we sidestepped making synchronization around this agent cluster-wide WaiterList store explicit. We can kind of pretend that they magically appear, or that they always exist for any possible (_block_, _i_).

Now, since for Records, the creation of new Records is an explicit thing, we now need to also explicitly capture the creation of these WaiterList records in the agent-wide store. Importantly, we need to synchronize the creation. If two agents are concurrently trying to get a WL corresponding to (_b_, _i_), we need to ensure both agents get the same WL.

To capture this, we need something like the following:

- The agent cluster-wide WL store itself needs a critical section that GetWaiterList enters.
- If the store doesn't have a record for (_b_, _i_), one needs to be created and explicitly set [[Waiters]] to the empty List and [[MostRecentLeaveEvent]] to ~empty~.

That's more machinery for not a lot of extra clarity of understanding, so I'm on the fence for this PR.

syg (changes requested):
I'm on the fence for this PR, given the comments below, but it could be made to work.

PR #2250, opened 2020-12-11
Normative: Relax the strength of Atomics.compareExchange

### LINKED ISSUE #2231: (memory model) Atomics.compareExchange is specified too strongly
Currently, [Atomics.compareExchange](https://tc39.es/ecma262/#sec-atomics.compareexchange) is specified as performing an idempotent RMW in the case that the comparison fails. This is stronger than the [analogous C++11 operation](https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange), which only performs a load if the comparison fails.

Currently, V8 appears to implement compareExchange on ARMv8 as a [pair of load-store exclusives](https://github.com/v8/v8/blob/4b9b23521e6fd42373ebbcb20ebe03bf445494f9/src/compiler/backend/arm64/code-generator-arm64.cc#L429), and only a load is performed if the comparison fails.

The difference is visible in the following program, where `b` is some integer TypedArray wrapping a zero-initialised SAB:

| Thread 1 | Thread 2 |
| ------------- | ------------- |
| b[1] = 1; | b[2] = 2; |
| Atomics.compareExchange(b, 0, 42, 42);  | Atomics.compareExchange(b, 0, 42, 42);  |
| console.log(b[2]); | console.log(b[1]);  |

If it is guaranteed that Atomics.compareExchange will always perform an RMW, then the outcome 0, 0 is forbidden by the JS memory model, as the total ordering of atomics forces one thread to synchronise with the other, even though both Atomics.compareExchange comparisons will clearly fail. Both the C++11 model, and the V8 implementation above, permit this outcome.

To model this, instead of unconditionally performing an RMW (via the spec function GetModifySetValueInBuffer), there should be two possible cases in the operational semantics. Given the provided expectedValue, either a load event with a value that is `!=` the expectedValue must be emitted, or an RMW event with a load value that is `==` the expectedValue must be emitted. The choice is non-deterministic (constrained by the axiomatic memory model).

syg:
Closes #2231

@conrad-watt does this look like what you had in mind?

I am labeling this normative **but not** needs consensus because I don't believe there is a reasonable way to implement the currently specified strong behavior, nor do implementations implement the strong version. Neither ARM's load and store exclusive instructions nor x86's `cmpxchg` unconditionally perform a read-modify-write as the spec currently requires.

Let's merge this now and inform the committee at the next plenary.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37419,6 +37419,7 @@ THH:mm:ss.sss
       <p>The following steps are taken:</p>
       <emu-alg>
         1. Let _buffer_ be ? ValidateIntegerTypedArray(_typedArray_).
+        1. Let _block_ be _buffer_.[[ArrayBufferData]].
         1. Let _indexedPosition_ be ? ValidateAtomicAccess(_typedArray_, _index_).
         1. Let _arrayTypeName_ be _typedArray_.[[TypedArrayName]].
         1. If _typedArray_.[[ContentType]] is ~BigInt~, then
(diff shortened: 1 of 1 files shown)


conrad-watt (commented):
I might be totally misreading, but I think this change is in the wrong place? `GetModifySetValueInBuffer` is used to define regular RMWs too, and it shouldn't be possible for these to emit `read` events.

Moreover, the `value` parameter to `GetModifySetValueInBuffer` is the _replacement_ value in the case of `Atomics.compareExchange`, not the expected value, so its byte representation shouldn't be being compared to `rawBytesRead`.

syg:
> I might be totally misreading, but I think this change is in the wrong place? GetModifySetValueInBuffer is used to define regular RMWs too, and it shouldn't be possible for these to emit read events.

Oh ha, yes.

syg:
This should be more like it.

@conrad-watt An interesting part of this change is that the comparison is now redundant in _compareExchange_ closure. I *think* removing the comparison there doesn't add any adverse side-effects but I'm not 100% confident. Moving the comparison of the expected value and the read value into the operational semantics means Valid Chosen Reads is what ensures that there is no valid execution whose compareExchange RMW event can witness a non-expected value. Those executions are instead now captured by the non-equal load event paths. Does that sound right to you?

spec.html
```diff
@@ -37455,12 +37456,29 @@ <h1>Atomics.compareExchange ( _typedArray_, _index_, _expectedValue_, _replaceme
...
+        1. Let _elementSize_ be the Element Size value specified in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for Element Type _elementType_.
         1. Let _isLittleEndian_ be the value of the [[LittleEndian]] field of the surrounding agent's Agent Record.
         1. Let _expectedBytes_ be NumericToRawBytes(_elementType_, _expected_, _isLittleEndian_).
-        1. Let _compareExchange_ be a new read-modify-write modification function with parameters (_oldBytes_, _newBytes_) that captures _expectedBytes_ and performs the following steps atomically when called:
-          1. If ByteListEqual(_oldBytes_, _expectedBytes_) is *true*, return _newBytes_.
-          1. Return _oldBytes_.
-        1. Return GetModifySetValueInBuffer(_buffer_, _indexedPosition_, _elementType_, _replacement_, _compareExchange_).
+        1. Let _replacementBytes_ be NumericToRawBytes(_elementType_, _replacement_, _isLittleEndian_).
+        1. Let _second_ be a new read-modify-write modification function with parameters (_oldBytes_, _newBytes_) that captures nothing and performs the following steps atomically when called:
```
bakkot, line 37463: Editorial nit: I'd prefer this to be defined closer to its single use site, rather than outside of the `If`.

acardoz (approved):

spec.html
syg: Done.

conrad-watt (approved):
LGTM

sorry for the delay

conrad-watt:
> Moving the comparison of the expected value and the read value into the operational semantics means Valid Chosen Reads is what ensures that there is no valid execution whose compareExchange RMW event can witness a non-expected value. Those executions are instead now captured by the non-equal load event paths. Does that sound right to you?

This should work. It is exposing the weird interplay between the operational and axiomatic parts of the semantics, but it isn't any worse than existing situations when there's conditional control flow (like an `if` statement) based on a concurrently loaded value.

michaelficarra (approved):

bakkot (approved):

Issue #2206, opened 2020-10-15
JSON.parse needs to not apply the early error rule from B.3.1


bakkot:
[`JSON.parse`](https://tc39.es/ecma262/#sec-json.parse) says

> Let _completion_ be the result of parsing and evaluating ! StringToCodePoints(scriptString) as if it was the source text of an ECMAScript Script. The extended PropertyDefinitionEvaluation semantics defined in [B.3.1](https://tc39.es/ecma262/#sec-__proto__-property-names-in-object-initializers) must not be used during the evaluation.

B.3.1 deals with `__proto__` in object literals. But it has two parts: the extended evaluation semantics, and a new early error rule forbidding multiple occurrences of `__proto__` in a single object literal. The early error rule is also not applied in implementations (and it would be weird to do so), but the spec text doesn't mention that the rule should not be applied.

michaelficarra:
As per today's editor call, we will mark this change as normative but make the change without asking for consensus (as a spec bugfix) since it matches with what we believe the intent of the committee to have been and with all known implementations.

PR #2252, opened 2020-12-13
Normative: do not apply Early Error for duplicate __proto__ properties when performing JSON.parse

bakkot:
Fixes https://github.com/tc39/ecma262/issues/2206. Quoting @michaelficarra there:

> As per today's editor call, we will mark this change as normative but make the change without asking for consensus (as a spec bugfix) since it matches with what we believe the intent of the committee to have been and with all known implementations.

I hope "when parsing text for JSON.parse" is sufficiently clear. I could make it more precise by labelling the step in JSON.parse which calls ParseText and then referring directly to that step, but that seemed like overkill.

This PR has two commits: the first makes the change, and the second rewords the prose list of conditions to be an HTML list, which I think reads better. I'm happy to back out the second if people dislike the style. If we keep it, the two commits should be combined before landing.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -42797,7 +42797,13 @@ THH:mm:ss.sss
 
     <emu-annex id="sec-__proto__-property-names-in-object-initializers">
       <h1>__proto__ Property Names in Object Initializers</h1>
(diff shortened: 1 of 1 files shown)


ljharb (approved):

michaelficarra (approved):
I still don't love the wording of this section, but this works for now.

PR #2254, opened 2020-12-13
Editorial: reword definition of lookahead restrictions

bakkot:
There's a few differences here:
- It talks about the referenced token sequences being a _prefix_ of the following sequence, which I believe was always the intent (resolving the issue mentioned in [this comment](https://github.com/tc39/ecma262/pull/2227#issuecomment-727093111)).
- It describes how to handle the case where the following token sequence is ambiguous, namely, by only allowing lookahead restrictions for which the ambiguity cannot matter (resolving the issue mentioned in [this comment](https://github.com/tc39/ecma262/pull/2227#issuecomment-727095436)).
- It only ever talks about sequences of tokens, never individual tokens (and therefore subsumes https://github.com/tc39/ecma262/pull/1828). 
- It removes the restriction that sequences be of length 1 or 2. I think the point of that restriction was to emphasize that the grammar is LR(2), but I think emphasizing this here is of dubious value. And there are cases where we would want [longer restrictions](https://github.com/tc39/ecma262/pull/1727#discussion_r332641200).
- It introduces the phrases in the order `=`, `∈`, `≠`, `∉`, which I think is clearest.

@waldemarhorwat @gibson042 @jmdyck, since each of you have commented on this section in the past, I'd appreciate any feedback you have about this phrasing.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -745,8 +745,9 @@
           `9`
       </emu-grammar>
       <p>If the phrase &ldquo;[empty]&rdquo; appears as the right-hand side of a production, it indicates that the production's right-hand side contains no terminals or nonterminals.</p>
-      <p>If the phrase &ldquo;[lookahead &notin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may not be used if the immediately following input token sequence is a member of the given _set_. The _set_ can be written as a comma separated list of one or two element terminal sequences enclosed in curly brackets. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all terminals to which that nonterminal could expand. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead &ne; _terminal_]&rdquo; may be used.</p>
-      <p>For example, given the definitions:</p>
+      <p>If the phrase &ldquo;[lookahead = _seq_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the token sequence _seq_ is a prefix of the immediately following input token sequence. Similarly, &ldquo;[lookahead &isin; _set_]&rdquo;, where _set_ is a finite nonempty set of token sequences, indicates that the production may only be used if some element of _set_ is a prefix of the immediately following token sequence. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all token sequences to which that nonterminal could expand. It is considered an editorial error if the nonterminal could expand to infinitely many distinct token sequences.</p>
+      <p>These conditions may be negated. &ldquo;[lookahead &ne; _seq_]&rdquo; indicates that the containing production may only be used if _seq_ is <em>not</em> a prefix of the immediately following input token sequence, and &ldquo;[lookahead &notin; _set_]&rdquo; indicates that the production may only be used if <em>no</em> element of _set_ is a prefix of the immediately following token sequence.</p>
+      <p>As an example, given the definitions:</p>
       <emu-grammar type="example">
         DecimalDigit :: one of
           `0` `1` `2` `3` `4` `5` `6` `7` `8` `9`
@@ -762,7 +763,7 @@
           DecimalDigit [lookahead &lt;! DecimalDigit]
       </emu-grammar>
       <p>matches either the letter `n` followed by one or more decimal digits the first of which is even, or a decimal digit not followed by another decimal digit.</p>
-      <p>Similarly, if the phrase &ldquo;[lookahead &isin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the immediately following input token sequence is a member of the given _set_. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead = _terminal_]&rdquo; may be used.</p>
+      <p>Note that when these phrases are used in the syntactic grammar, it may not be possible to unambiguously identify the immediately following token sequence because determining later tokens requires knowing which lexical goal symbol to use at later positions. As such, when these are used in the syntactic grammar, it is considered an editorial error for a token sequence _seq_ to appear in a lookahead restriction (including as part of a set of sequences) if the choices of lexical goal symbols to use could change whether or not _seq_ would be a prefix of the resulting token sequence.</p>
       <p>If the phrase &ldquo;[no |LineTerminator| here]&rdquo; appears in the right-hand side of a production of the syntactic grammar, it indicates that the production is <em>a restricted production</em>: it may not be used if a |LineTerminator| occurs in the input stream at the indicated position. For example, the production:</p>
       <emu-grammar type="example">
         ThrowStatement :


spec.html
```diff
@@ -745,8 +745,9 @@ <h1>Grammar Notation</h1>
...
-      <p>For example, given the definitions:</p>
+      <p>If the phrase &ldquo;[lookahead = _seq_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the token sequence _seq_ is a prefix of the immediately following input token sequence. Similarly, &ldquo;[lookahead &isin; _set_]&rdquo;, where _set_ is a finite nonempty set of token sequences, indicates that the production may only be used if some element of _set_ is a prefix of the immediately following token sequence. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all sequences of tokens to which that nonterminal could expand.</p>
```
jmdyck, line 748:
```suggestion
      <p>If the phrase &ldquo;[lookahead = _seq_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the token sequence _seq_ is a prefix of the immediately following input token sequence. Similarly, &ldquo;[lookahead &isin; _set_]&rdquo;, where _set_ is a finite nonempty set of token sequences, indicates that the production may only be used if some element of _set_ is a prefix of the immediately following token sequence. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all token sequences to which that nonterminal could expand.</p>
```
(for consistency with the rest of the sentence)

spec.html
```diff
@@ -762,7 +763,7 @@ <h1>Grammar Notation</h1>
           DecimalDigit [lookahead &lt;! DecimalDigit]
       </emu-grammar>
       <p>matches either the letter `n` followed by one or more decimal digits the first of which is even, or a decimal digit not followed by another decimal digit.</p>
-      <p>Similarly, if the phrase &ldquo;[lookahead &isin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the immediately following input token sequence is a member of the given _set_. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead = _terminal_]&rdquo; may be used.</p>
+      <p>Note that when these phrases are used in the syntactic grammar it may not be possible to unambiguously identify the immediately following token sequence, because determining later tokens requires knowing which lexical goal symbol to use at later positions. As such, when these are used in the syntatic grammar each token sequence _seq_ must be such that the choices of lexical goal symbols to use when scanning the following input text cannot change whether or not _seq_ is a prefix of the resulting token sequence.</p>
```
jmdyck, line 766:
```suggestion
      <p>Note that when these phrases are used in the syntactic grammar it may not be possible to unambiguously identify the immediately following token sequence, because determining later tokens requires knowing which lexical goal symbol to use at later positions. As such, when these are used in the syntactic grammar, each token sequence _seq_ is such that the choices of lexical goal symbols to use when scanning the following input text cannot change whether or not _seq_ is a prefix of the resulting token sequence.</p>
```
("must be" sounds like a requirement on implementations, whereas this is stating a self-imposed constraint on the spec.)

jmdyck (changes requested):

bakkot:
@jmdyck Done, thanks.

spec.html
```diff
@@ -762,7 +763,7 @@ <h1>Grammar Notation</h1>
           DecimalDigit [lookahead &lt;! DecimalDigit]
       </emu-grammar>
       <p>matches either the letter `n` followed by one or more decimal digits the first of which is even, or a decimal digit not followed by another decimal digit.</p>
-      <p>Similarly, if the phrase &ldquo;[lookahead &isin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the immediately following input token sequence is a member of the given _set_. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead = _terminal_]&rdquo; may be used.</p>
+      <p>Note that when these phrases are used in the syntactic grammar it may not be possible to unambiguously identify the immediately following token sequence, because determining later tokens requires knowing which lexical goal symbol to use at later positions. As such, when these are used in the syntactic grammar, each token sequence _seq_ is such that the choices of lexical goal symbols to use when scanning the following input text cannot change whether or not _seq_ is a prefix of the resulting token sequence.</p>
```
syg, line 766: This still reads awkwardly for the reason @jmdyck said. Maybe something like "As such, when these are used in the syntactic grammar, for each token sequence _seq_, it is considered an editorial error unless the choices of lexical goal symbols to use cannot change whether or not _seq_ is a prefix of the resulting token sequence."?

syg (commented):
lgtm modulo wordsmithing

syg:
This subsumes #2227, right?

spec.html
bakkot: Hm, that's still a bit awkward to my ear. Maybe

> As such, when these are used in the syntactic grammar it is considered an editorial error for a token sequence _seq_ to appear in a lookahead restriction if the choices of lexical goal symbols to use could change whether or not _seq_ would be a prefix of the resulting token sequence.

?

bakkot:
> This subsumes #2227, right?

Right. To be precise, we decided #2227 wasn't exactly an issue, but there were related issues which needed addressing, which this PR does.

spec.html
syg: The tweak sounds good to me.

spec.html
```diff
@@ -762,7 +763,7 @@ <h1>Grammar Notation</h1>
           DecimalDigit [lookahead &lt;! DecimalDigit]
       </emu-grammar>
       <p>matches either the letter `n` followed by one or more decimal digits the first of which is even, or a decimal digit not followed by another decimal digit.</p>
-      <p>Similarly, if the phrase &ldquo;[lookahead &isin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the immediately following input token sequence is a member of the given _set_. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead = _terminal_]&rdquo; may be used.</p>
+      <p>Note that when these phrases are used in the syntactic grammar it may not be possible to unambiguously identify the immediately following token sequence, because determining later tokens requires knowing which lexical goal symbol to use at later positions. As such, when these are used in the syntactic grammar it is considered an editorial error for a token sequence _seq_ to appear in a lookahead restriction if the choices of lexical goal symbols to use could change whether or not _seq_ would be a prefix of the resulting token sequence.</p>
```
michaelficarra, line 766:
```suggestion
      <p>Note that when these phrases are used in the syntactic grammar, it may not be possible to unambiguously identify the immediately following token sequence because determining later tokens requires knowing which lexical goal symbol to use at later positions. As such, when these are used in the syntactic grammar, it is considered an editorial error for a token sequence _seq_ to appear in a lookahead restriction if the choices of lexical goal symbols to use could change whether or not _seq_ would be a prefix of the resulting token sequence.</p>
```

michaelficarra (approved):

spec.html
bakkot: Done.

claudepache:
I’m not fan of formulations like:

> It is considered an editorial error

used twice in this PR, because you are mixing at the same level considerations about the language (ECMAScript) and the metalanguage (ECMASpeak). It may not be obvious to a casual reader that this phrase talk about a completely different subject from “It is a Syntax Error...” found elsewhere in the spec. I find it a better style to make this sort of “meta-error” implicit, i.e., instead of writing:

> ... the set can also be written as a nonterminal, in which case it represents the set of all token sequences to which that nonterminal could expand. <ins>It is considered an editorial error if the nonterminal could expand to infinitely many</ins> distinct token sequences.

write:

> ... the set can also be written as a nonterminal <ins>which could expand to finitely many</ins> distinct token sequences, in which case it represents the set of all token sequences to which that nonterminal could expand.

Similarly, instead of blaming ourselves when this will happen:

> Note that when these phrases are used in the syntactic grammar, <ins>it may not be possible to unambiguously identify</ins> the immediately following token sequence because determining later tokens requires knowing which lexical goal symbol to use at later positions. As such, when these are used in the syntactic grammar, <ins>it is considered an editorial error for a token sequence _seq_ to appear</ins> in a lookahead restriction if the choices of lexical goal symbols to use could change whether or not _seq_ would be a prefix of the resulting token sequence.

just state that this will not happen:

> These phrases are used in the syntactic grammar <ins>only when it is possible to unambiguously identify</ins> the immediately following token sequence, therefore when determining later tokens does not require knowing which lexical goal symbol to use at later positions.  As such, <ins>a token sequence _seq_ will not appear</ins> in a lookahead restriction if the choices of lexical goal symbols to use could change whether or not _seq_ would be a prefix of the resulting token sequence.

As a bonus, we gain in conciseness.

jmdyck:
I agree with @claudepache's general idea. However, in the latter case, I think the suggested replacement is incorrect. The meta-restriction doesn't ensure that lookahead-constraints are used "only when it is possible to unambiguously identify the immediately following token sequence". Rather, we're saying that even though that 'ambiguity' exists, it doesn't matter for the purposes of the lookahead-constraint. (Even though you don't know what the immediately following token sequence *is*, we guarantee that you'll be able to discern whether the lookahead-constraint is satisfied.)

(Strictly speaking, this isn't an "ambiguity" in the grammatical sense. It's more a like a circularity ("catch 22") in the definition:
- To know whether you can use this production,
- you need to know if its lookahead-constraint is satisfied,
- and for that, you (seem to) need to know the immediately following token sequence,
- and for that, you (in general) need to know what lexical goal symbol to use for each of the next few tokens,
- and for that, you need to know what the next few "syntactic grammar contexts" are,
- and for that, you need to know what production to use now,
- which is where we started.

We break this apparent circularity at step 3: to know if a lookahead-constraint is satisfied, you *don't* need to know exactly what the following token sequence is.)

bakkot:
I'm fine with either construction, though as @jmdyck notes, the "it may not be possible to unambiguously identify" phrasing needs to stay (or be replaced with an equivalent).

@syg @michaelficarra, thoughts on @claudepache's suggested wordings?

syg:
I can live with @claudepache's suggestion, but prefer "it is an editorial error" for some semblance of "coherence in depth" (which one may disagree with philosophically similarly to disagreeing with defense in depth). Our spec is a huge document and I like the explicitness of the affordance that "it is an editorial error" gives for future bugs.

I think @claudepache's formulation does indeed read nicer, and in other contexts than a living standard draft (e.g. a paper) I'd prefer it.

spec.html
```diff
@@ -762,7 +763,7 @@ <h1>Grammar Notation</h1>
           DecimalDigit [lookahead &lt;! DecimalDigit]
       </emu-grammar>
       <p>matches either the letter `n` followed by one or more decimal digits the first of which is even, or a decimal digit not followed by another decimal digit.</p>
-      <p>Similarly, if the phrase &ldquo;[lookahead &isin; _set_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the immediately following input token sequence is a member of the given _set_. If the _set_ consists of a single terminal the phrase &ldquo;[lookahead = _terminal_]&rdquo; may be used.</p>
+      <p>Note that when these phrases are used in the syntactic grammar, it may not be possible to unambiguously identify the immediately following token sequence because determining later tokens requires knowing which lexical goal symbol to use at later positions. As such, when these are used in the syntactic grammar, it is considered an editorial error for a token sequence _seq_ to appear in a lookahead restriction if the choices of lexical goal symbols to use could change whether or not _seq_ would be a prefix of the resulting token sequence.</p>
```
waldemarhorwat, line 766: _seq_ should also include _set_ in the last paragraph.

waldemarhorwat:
I like the wording. Minor correction: _seq_ should also include _set_ in the last paragraph.

Also, the positive set lookahead is never used in the spec. It doesn't bother me to keep it though.

spec.html
bakkot: I've [clarified](https://github.com/tc39/ecma262/pull/2254/commits/7351ba7bca46b8552f08797c68b450ac471567fe); let me know if that doesn't address your concern.

spec.html
```diff
@@ -745,8 +745,9 @@ <h1>Grammar Notation</h1>
...
+      <p>If the phrase &ldquo;[lookahead = _seq_]&rdquo; appears in the right-hand side of a production, it indicates that the production may only be used if the token sequence _seq_ is a prefix of the immediately following input token sequence. Similarly, &ldquo;[lookahead &isin; _set_]&rdquo;, where _set_ is a finite nonempty set of token sequences, indicates that the production may only be used if some element of _set_ is a prefix of the immediately following token sequence. For convenience, the set can also be written as a nonterminal, in which case it represents the set of all token sequences to which that nonterminal could expand. It is considered an editorial error if the nonterminal could expand to infinitely many distinct token sequences.</p>
+      </p>These conditions may be negated. &ldquo;[lookahead &ne; _seq_]&rdquo; indicates that the containing production may only be used if _seq_ is <em>not</em> a prefix of the immediately following input token sequence, and &ldquo;[lookahead &notin; _set_]&rdquo; indicates that the production may only be used if <em>no</em> element of _set_ is a prefix of the immediately following token sequence.</p>
```
TimothyGu, line 749:
```suggestion
      <p>These conditions may be negated. &ldquo;[lookahead &ne; _seq_]&rdquo; indicates that the containing production may only be used if _seq_ is <em>not</em> a prefix of the immediately following input token sequence, and &ldquo;[lookahead &notin; _set_]&rdquo; indicates that the production may only be used if <em>no</em> element of _set_ is a prefix of the immediately following token sequence.</p>
```

bakkot: Whoops, nice catch. Fixed.

syg (approved):

bakkot:
OK, I'm going to leave it as-is. I think this is ready to go.

PR #2258, opened 2020-12-26
Editorial: reword StringIndexOf in terms of substring

bakkot:
This is a very small tweak. I find the explicit loop to be easier to understand than the previous "if there exists ... such that for all ..." prose.


ljharb (approved):
This is a massive improvement, thanks

mathiasbynens (approved):
Although `git blame` points to me here, this was inherited from the old phrasing (from before `StringIndexOf` existed). I agree this patch makes things much easier to follow!

ljharb:
(added in #2009; i commented on this [here](https://github.com/tc39/proposal-string-replaceall/issues/32#issuecomment-547750076) :-p )

syg (approved):
Very nice.

PR #2260, opened 2020-12-28
Meta: migrate from travis-ci to Github Actions

ljharb:
(ref https://github.com/ljharb/object.assign/pull/81)

travis-ci's new pricing plan has effectively dropped support for open source repos, so we need to migrate off of it.

(To work properly, these workflows must be set to `pull_request_target`, not `pull_request`, and must have a GOOGLE_API_KEY secret available, as well as GH_TOKEN - still working on this)


leobalter:
to make this work properly, you eventually need to manually disable the old expected checks from TravisCI or they will remain eternally waiting for status.

Disregard if you already know this, it's just trauma from an event in the past.

ljharb:
Yes, as soon as this PR is ready to land, I'll make the actions checks "required" and the travis checks no longer required, but as soon as I do that, every PR will be unmergeable until they're rebased, so it's the last step before/during merge.

.github/workflows/build.yml
```diff
@@ -0,0 +1,17 @@
+name: 'ecma-262'
+
+on: [pull_request, push]
+
+jobs:
+  build:
+    name: 'build & lint ecmarkup'
+    runs-on: ubuntu-latest
```
septs, line 8:
```suggestion
    runs-on: ubuntu-20.04
```

> Ubuntu-20.04 is ready to be the default version for the “ubuntu-latest” label in GitHub Actions and Azure DevOps. This change will be rolled out over a period of several weeks beginning on November, 30. (December, 22 - January, 4 will be a migration break due to holidays)

from https://github.com/actions/virtual-environments/issues/1816

septs: image

ljharb: the ubuntu version should never make a difference for the majority of node project's CI, there's no benefit in changing from the default.

.github/workflows/build.yml
```diff
@@ -0,0 +1,17 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ljharb/actions/node/run@125357c
+        name: 'npm ci && npm run build-master'
+        with:
+          node-version: lts/*
+          use-npm-ci: true
+          command: 'build-master'
```
ExE-Boss, line 17: This should probably include the trailing newline.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ljharb/actions/node/run@125357c
+        name: 'npm ci && npm run build-only ./scripts/auto-deploy.sh'
+        with:
+          node-version: lts/*
+          use-npm-ci: true
+          shell-command: |
+            npm run build-only && \
+            ./scripts/auto-deploy.sh
```
ExE-Boss, line 23: ditto

.github/workflows/preview.yml
```diff
@@ -0,0 +1,22 @@
...
+      - uses: ljharb/actions/node/run@125357c
+        name: 'npm ci && npm run build-only && …'
+        env:
+          CI_PREVIEW_TOKEN: ${{ secrets.CI_PREVIEW_TOKEN }}
+          PULL_REQUEST: ${{ github.event.number }}
+        with:
+          node-version: lts/*
+          use-npm-ci: true
+          shell-command: |
+            npm run build-only && node scripts/insert_snapshot_warning && node scripts/publish-preview
```
ExE-Boss, line 22: ditto

.github/workflows/build.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ljharb/actions/node/install@56f291b2a0d9e3bce72634356ae949b1e054b9f8
+        name: 'nvm install lts/* && npm install'
+        with:
+          node-version: lts/*
+          use-npm-ci: true
+      - run: 'npm ci && npm run build-master'
```
septs, line 17: why not split?

```suggestion
      - run: npm ci
      - run: npm run build-master
```

ljharb: sure, thanks

ljharb:
This is ready to go! See https://github.com/ljharb/ecma262/pull/13 for proof that the Preview status shows up on a PR (but that won't happen on this repo until after this PR is landed, and thus, won't ever happen on this PR).

Once this is landed / while landing it, i'll update the required branch protections so that any other PR must rebase in order to be mergeable.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,21 @@
...
+    if: ${{ github.repository == 'tc39/ecma262' }}
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ljharb/actions/node/install@56f291b2a0d9e3bce72634356ae949b1e054b9f8
+        name: 'nvm install lts/* && npm ci'
+        with:
+          node-version: lts/*
+          use-npm-ci: true
+      - run: npm run build-only ./scripts/auto-deploy.sh
```
bakkot, line 21: Does `npm run` allow you to specify multiple commands, or is this some other syntax I don't recognize?

ljharb: whoops, it is a typo. good catch! saves us a failed master build :-)

.github/workflows/preview.yml
```diff
@@ -0,0 +1,51 @@
...
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const artifacts = await github.actions.listWorkflowRunArtifacts({
+               owner,
+               repo,
+               run_id: ${{ github.event.workflow_run.id }},
+            });
+            const [{ id: artifact_id }] = artifacts.data.artifacts.filter((artifact) => artifact.name == 'out');
```
bakkot, line 37: Random nit:

```suggestion
            const { id: artifact_id } = artifacts.data.artifacts.find((artifact) => artifact.name == 'out');
```

bakkot (approved):
Hard to review this for correctness, since it interacts with a bunch of other systems, so take my approval as "this looks reasonable as far as I can tell, let's land it and see what happens".

.github/workflows/preview.yml
ljharb: ha, thanks. this was tweaked copy-pasta; i'll pull in your suggestion.

PR #2266, opened 2021-01-04
Editorial: Fix author listing

chrikrah:
1. Changing the listing of authors to be in format `<firstname lastname>` consistently. Previously `<firstname lastname>` and `<lastname, firstname>` were both used. Choosing `<firstname lastname>` to prevent too many commas in a listing of multiple authors in a row which ultimately hinders readability.
2. Adding comma after the penultimate item in author listing (to be consistent with the subsequent line which was already doing it that way).


bakkot:
Hm. I suspect this is intended to follow a specific citation format, although I'm not sure which it would be.

syg:
It would be good to pick a bibliography style and stick to it.

syg:
The editors chatted about this, and we realized that that section is significantly outdated. For example, it references a very, very old version of the Java standard. We might rewrite that section and sidestep this bibliography style question, but we're still thinking about it.

syg:
As in https://github.com/tc39/ecma262/pull/2303#issuecomment-780912446, the editors have decided to remove this paragraph entirely.

PR #2270, opened 2021-01-08
Meta: bump ecmarkup (and related tweaks)

### LINKED ISSUE #2183: inconsistent notation for multiplication and division

bakkot:
See commits for details. These commits can be landed individually.

This also fixes #2183 by replacing the two usages of `÷` with `/`. I couldn't find any usages of `*` for multiplication, so I think we're good there.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,17 +11,17 @@
       "dev": true
     },
     "@babel/code-frame": {
-      "version": "7.10.4",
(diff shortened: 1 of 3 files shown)


ljharb (approved):

bakkot:
Also note that this includes https://github.com/tc39/ecmarkup/pull/275, which affects rendering. You can play with it in the preview.

michaelficarra (approved):

PR #2282, opened 2021-01-18
Editorial: add missing assertion to ApplyStringOrNumericBinaryOperator

### LINKED ISSUE #2281: `ApplyStringOrNumericBinaryOperator` does not validate `opText`
https://tc39.es/ecma262/#sec-applystringornumericbinaryoperator asserts that `opText` is a sequence of Unicode code points, but does not validate that it is present in the table in step 7.

I assume we want an Assert that `opText` is in the relevant table?

ljharb:
Fixes #2281

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -17157,6 +17157,7 @@
       <h1>ApplyStringOrNumericBinaryOperator ( _lval_, _opText_, _rval_ )</h1>
       <p>The abstract operation ApplyStringOrNumericBinaryOperator takes arguments _lval_ (an ECMAScript language value), _opText_ (a sequence of Unicode code points), and _rval_ (an ECMAScript language value). It performs the following steps when called:</p>
       <emu-alg>
+        1. Assert: _opText_ is present in the table in step <emu-xref href="#step-applystringornumericbinaryoperator-operations-table"></emu-xref>.
         1. If _opText_ is `+`, then
           1. [id="step-binary-op-toprimitive-lval"] Let _lprim_ be ? ToPrimitive(_lval_).
           1. [id="step-binary-op-toprimitive-rval"] Let _rprim_ be ? ToPrimitive(_rval_).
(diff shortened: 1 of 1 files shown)


michaelficarra (approved):

spec.html
```diff
@@ -17171,28 +17172,28 @@ <h1>ApplyStringOrNumericBinaryOperator ( _lval_, _opText_, _rval_ )</h1>
...
-                <tr><td> `&amp;`        </td><td> _T_::bitwiseAND         </td></tr>
-                <tr><td> `^`            </td><td> _T_::bitwiseXOR         </td></tr>
-                <tr><td> `|`            </td><td> _T_::bitwiseOR          </td></tr>
-              </tbody>
-            </table>
-          </figure>
+        1. Let _operation_ be the abstract operation associated with _opText_ in the following <emu-xref href="#table-applystringornumericbinaryoperator-operations">table of operations</emu-xref>.
         1. Return ? _operation_(_lnum_, _rnum_).
       </emu-alg>
+      <emu-table id="table-applystringornumericbinaryoperator-operations">
```
jmdyck, line 17178: This table has no caption.

spec.html
```diff
@@ -17171,28 +17172,28 @@ <h1>ApplyStringOrNumericBinaryOperator ( _lval_, _opText_, _rval_ )</h1>
...
+        <table class="lightweigt-table">
```
jmdyck, line 17179:
```suggestion
        <table class="lightweight-table">
```

jmdyck (changes requested):
Why move the table out of the algorithm? In the context of the algorithm, the aliases `_opText_`, `_operation_`, and `_T_` all have meaning. Outside the algorithm, they don't.

If the point is just to be able to refer to the table from the Assertion, you don't need to make it a 'free-standing' table, you could just give step 8 an `[id="..."]` and say:
`Assert: _opText_ is present in the table in [step 8].`

ljharb:
@jmdyck because it wouldn't render properly with an ID when it's inside the algorithm - i didn't think of using step IDs tho, i'll try that.

ljharb:
Updated to use a step xref instead of altering the table.

michaelficarra:
I kinda preferred moving the table out. We could use better column headers if the variable names would be confusing there.

ljharb:
cool, let's figure it out tomorrow on the call; i can easily revert back.

jmdyck:
> I kinda preferred moving the table out. We could use better column headers if the variable names would be confusing there.

Yeah, but it's unclear what you'd do about the references to `_T_`.

jmdyck (approved):
I mean, I'm not convinced of the need (usefulness) of this assertion, but I recognize that that's a matter of taste; and if you want an assertion, this (with the table embedded in the algorithm) seems like the easiest (least problematic) way to achieve that.

bakkot (approved):
Inline LGTM.

syg (approved):

PR #1668, opened 2019-08-14
Normative: Add public and private class fields, static class features, private methods & getters/setters

Ms2ger:
This includes changes from https://tc39.es/proposal-static-class-features/.

PR #2290, opened 2021-01-27
Normative: add `#x in obj` syntax

ljharb:
~~This PR is based on top of #1668. Please review by commit. As #1668 is updated and/or merged, I'll rebase this one.~~

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7708,6 +7708,7 @@
           RelationalExpression `&gt;=` ShiftExpression
           RelationalExpression `instanceof` ShiftExpression
           RelationalExpression `in` ShiftExpression
+          PrivateIdentifier `in` ShiftExpression
 
         EqualityExpression :
           EqualityExpression `==` RelationalExpression
@@ -8456,6 +8457,7 @@
           RelationalExpression `&gt;=` ShiftExpression
           RelationalExpression `instanceof` ShiftExpression
           RelationalExpression `in` ShiftExpression
+          PrivateIdentifier `in` ShiftExpression
 
         EqualityExpression :
           EqualityExpression `==` RelationalExpression
@@ -17124,6 +17126,7 @@
         RelationalExpression[?In, ?Yield, ?Await] `&gt;=` ShiftExpression[?Yield, ?Await]
         RelationalExpression[?In, ?Yield, ?Await] `instanceof` ShiftExpression[?Yield, ?Await]
         [+In] RelationalExpression[+In, ?Yield, ?Await] `in` ShiftExpression[?Yield, ?Await]
+        [+In] PrivateIdentifier `in` ShiftExpression[?Yield, ?Await]
     </emu-grammar>
     <emu-note>
       <p>The <sub>[In]</sub> grammar parameter is needed to avoid confusing the `in` operator in a relational expression with the `in` operator in a `for` statement.</p>
@@ -17183,6 +17186,17 @@
         1. Let _rval_ be ? GetValue(_rref_).
         1. If Type(_rval_) is not Object, throw a *TypeError* exception.
         1. Return ? HasProperty(_rval_, ? ToPropertyKey(_lval_)).
+      </emu-alg>
+      <emu-grammar>RelationalExpression : PrivateIdentifier `in` ShiftExpression</emu-grammar>
+      <emu-alg>
+        1. Let _privateIdentifier_ be the StringValue of |PrivateIdentifier|.
+        1. Let _rref_ be the result of evaluating |ShiftExpression|.
+        1. Let _rval_ be ? GetValue(_rref_).
+        1. If Type(_rval_) is not Object, throw a *TypeError* exception.
+        1. Let _privateEnv_ be the running execution context's PrivateEnvironment.
+        1. Let _privateName_ be ! ResolvePrivateIdentifier(_privateEnv_, _privateIdentifier_).
+        1. If ! PrivateElementFind(_privateName_, _rval_) is not ~empty~, return *true*.
+        1. Return *false*.
       </emu-alg>
     </emu-clause>
 
@@ -21226,6 +21240,13 @@
         1. Let _newNames_ be the list-concatenation of _names_ and PrivateBoundIdentifiers of |ClassBody|.
         1. Return AllPrivateIdentifiersValid of |ClassElementList| with argument _newNames_.
       </emu-alg>
+
+      <emu-grammar>RelationalExpression : PrivateIdentifier `in` ShiftExpression</emu-grammar>
+      <emu-alg>
+        1. If _names_ contains the StringValue of |PrivateIdentifier|, then
+          1. Return AllPrivateIdentifiersValid of |ShiftExpression| with argument _names_.
+        1. Return *false*.
+      </emu-alg>
     </emu-clause>
 
     <emu-clause id="sec-static-semantics-privateboundidentifiers" type="sdo" aoid="PrivateBoundIdentifiers">
@@ -22088,6 +22109,7 @@
             RelationalExpression `&gt;=` ShiftExpression
             RelationalExpression `instanceof` ShiftExpression
             RelationalExpression `in` ShiftExpression
+            PrivateIdentifier `in` ShiftExpression
 
           ShiftExpression :
             ShiftExpression `&lt;&lt;` AdditiveExpression


spec.html
```diff
@@ -20739,6 +21579,26 @@ <h1>Runtime Semantics: Evaluation</h1>
...
       </emu-alg>
+      <emu-grammar>ClassElementName : PropertyName</emu-grammar>
+      <emu-alg>
+        1. Return the result of evaluating PropertyName.
+      </emu-alg>
+      <emu-grammar>ClassElementName : PrivateIdentifier</emu-grammar>
+      <emu-alg>
+        1. Let _privateIdentifier_ be StringValue of |PrivateIdentifier|.
+        1. Let _scope_ be the running execution context's PrivateEnvironment.
+        1. Let _scopeEnvRec_ be _scope_'s EnvironmentRecord.
```
jmdyck, line 21590:
```suggestion
        1. Let _scopeEnvRec_ be the running execution context's PrivateEnvironment.
```
(but a better name would probably be `_privateEnvRec_`)

spec.html
```diff
@@ -23685,8 +24566,9 @@ <h1>PerformEval ( _x_, _callerRealm_, _strictCaller_, _direct_ )</h1>
           1. Set _evalContext_'s ScriptOrModule to _runningContext_'s ScriptOrModule.
           1. Set _evalContext_'s VariableEnvironment to _varEnv_.
           1. Set _evalContext_'s LexicalEnvironment to _lexEnv_.
+          1. Set _evalContext_'s PrivateEnvironment to _privateEnv_.
           1. Push _evalContext_ onto the execution context stack; _evalContext_ is now the running execution context.
-          1. Let _result_ be EvalDeclarationInstantiation(_body_, _varEnv_, _lexEnv_, _strictEval_).
+          1. Let _result_ be EvalDeclarationInstantiation(_body_, _varEnv_, _lexEnv_, _privateEnv_ _strictEval_).
```
jmdyck, line 24571:
```suggestion
          1. Let _result_ be EvalDeclarationInstantiation(_body_, _varEnv_, _lexEnv_, _privateEnv_, _strictEval_).
```

spec.html
```diff
@@ -23734,6 +24616,13 @@ <h1>EvalDeclarationInstantiation ( _body_, _varEnv_, _lexEnv_, _strict_ )</h1>
                     1. NOTE: Annex <emu-xref href="#sec-variablestatements-in-catch-blocks"></emu-xref> defines alternate semantics for the above step.
                   1. NOTE: A direct eval will not hoist var declaration over a like-named lexical declaration.
               1. Set _thisEnv_ to _thisEnv_.[[OuterEnv]].
+          1. Let _privateIdentifiers_ be a new empty List.
+          1. Let _privEnv_ be _privateEnv_.
+          1. Repeat, while _privEnv_ is not *null*,
+            1. For each binding named _N_ in _privEnv_'s EnvironmentRecord, do
```
jmdyck, line 24622:
```suggestion
            1. For each binding named _N_ in _privEnv_, do
```

spec.html
```diff
@@ -23734,6 +24616,13 @@ <h1>EvalDeclarationInstantiation ( _body_, _varEnv_, _lexEnv_, _strict_ )</h1>
...
+              1. If _privateIdentifiers_ does not contain _N_, append _N_ to _privateIdentifiers_.
+            1. Set _privEnv_ to _privEnv_'s outer environment reference.
```
jmdyck, line 24624:
```suggestion
            1. Set _privEnv_ to _privEnv_.[[OuterEnv]].
```

jmdyck (changes requested):

spec.html
ljharb: I think this one is for #1668? i fixed it here tho, just in case.

spec.html
ljharb: this also is for #1668, i think?

ljharb:
I've split out all the fixes you requested that aren't part of this PR, but are instead part of #1668, into a separate commit. Please only review the last commit here :-)

jmdyck:
Sorry, I wasn't paying enough attention (re #1668).

spec.html
```diff
@@ -3766,7 +3777,7 @@ <h1>The Reference Record Specification Type</h1>
             </tr>
             <tr>
               <td oldids="sec-getreferencedname,ao-getreferencedname">[[ReferencedName]]</td>
-              <td>String or Symbol</td>
+              <td>String, Symbol or Private Name</td>
```
jridgewell, line 3780: :ahem:
```suggestion
              <td>String, Symbol, or Private Name</td>
```

jmdyck: (That's already been [suggested](https://github.com/tc39/ecma262/pull/1668/files#r560402082) in PR #1668.)

ljharb: Done in 53322a0, and added to #1668.

spec.html
```diff
@@ -17236,6 +17239,19 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+      <emu-alg>
+        1. Let _privateIdentifier_ be the StringValue of |PrivateIdentifier|.
+        1. Let _rref_ be the result of evaluating |ShiftExpression|.
+        1. Let _rval_ be ? GetValue(_rref_).
+        1. If Type(_rval_) is not Object, throw a *TypeError* exception.
+        1. Let _env_ be the running execution context's PrivateEnvironment.
+        1. Let _privateNameBinding_ be ? ResolveBinding(_privateIdentifier_, _env_).
+        1. Let _privateName_ be ? GetValue(_privateNameBinding_).
+        1. Assert: _privateName_ is a Private Name.
+        1. If ! PrivateElementFind(_privateName_, _rval) is not ~empty~, return *true*.
```
nicolo-ribaudo, line 17252:
```suggestion
        1. If ! PrivateElementFind(_privateName_, _rval_) is not ~empty~, return *true*.
```

spec.html
```diff
@@ -21281,6 +21297,14 @@ <h1>Static Semantics: AllPrivateIdentifiersValid</h1>
...
         1. Return AllPrivateIdentifiersValid of |ClassElementList| with argument _newNames_.
       </emu-alg>
+
+      <emu-grammar>RelationalExpression : PrivateIdentifier `in` ShiftExpression</emu-grammar>
+      <emu-alg>
+        1. If StringValue of |PrivateIdentifier| is in _names_, return *true*.
+        1. Return *false*.
+      </emu-alg>
+
+      <p>For all other grammatical productions, recurse on subexpressions/substatements, passing in the _names_ of the caller. If all pieces return *true*, then return *true*. If any returns *false*, return *false*.</p>
```
nicolo-ribaudo, line 21307: I think this was accidentally left here while rebasing, but it has been replaced by the first catch-all algorithm of https://tc39.es/ecma262/#sec-static-semantics-allprivateidentifiersvalid.

spec.html
```diff
@@ -17236,6 +17239,19 @@ <h1>Runtime Semantics: Evaluation</h1>
...
       </emu-alg>
+      <emu-grammar>RelationalExpression : PrivateIdentifier `in` ShiftExpression</emu-grammar>
+      <emu-alg>
+        1. Let _privateIdentifier_ be the StringValue of |PrivateIdentifier|.
+        1. Let _rref_ be the result of evaluating |ShiftExpression|.
+        1. Let _rval_ be ? GetValue(_rref_).
+        1. If Type(_rval_) is not Object, throw a *TypeError* exception.
+        1. Let _env_ be the running execution context's PrivateEnvironment.
+        1. Let _privateNameBinding_ be ? ResolveBinding(_privateIdentifier_, _env_).
+        1. Let _privateName_ be ? GetValue(_privateNameBinding_).
```
jmdyck, line 17250:
```suggestion
        1. Let _privateName_ be ! ResolvePrivateIdentifier(_env_, _privateIdentifier_).
```
(I *think* that's right. @Bakkot?)

Also, you might want to rename `_env_` to `_privateEnv_`.

jmdyck (changes requested):

spec.html
bakkot: Yup. And no particular reason for the assertion afterwards, since `ResolvePrivateIdentifier` always returns a Private Name.

spec.html
```diff
@@ -21281,6 +21297,12 @@ <h1>Static Semantics: AllPrivateIdentifiersValid</h1>
         1. Append to _newNames_ the elements of PrivateBoundIdentifiers of |ClassBody|.
         1. Return AllPrivateIdentifiersValid of |ClassElementList| with argument _newNames_.
       </emu-alg>
+
+      <emu-grammar>RelationalExpression : PrivateIdentifier `in` ShiftExpression</emu-grammar>
+      <emu-alg>
+        1. If StringValue of |PrivateIdentifier| is in _names_, return *true*.
```
jridgewell, line 21303: I think I made a mistake here when I originally suggested this. This sees a an expression `#foo in SHIFT_EXPRESSION`, and decides that if `foo` is a known private, all is good. Except, we forgot to check whether there are any unknown privates in `SHIFT_EXPRESSION`. For instance, I don't believe we fail like we should with `#foo in obj.#bar` (given `#foo` is declared and `#bar` is not), because `#bar` is in the `SHIFT_EXPRESSION` position and we stop at the `RELATIONAL_EXPRESSION` position.

We can see that we still need to recurse by looking at the other `AllPrivateIdentifiersValid` defined here:

```text
MemberExpression : MemberExpression . PrivateIdentifier
  1. If names contains the StringValue of PrivateIdentifier, then
    a. Return AllPrivateIdentifiersValid of MemberExpression with argument names.
  2. Return false.
```

Here, we found `MEMBER_EXPRESION.#foo`, `foo` was a known private and then continue to recurse into `MEMBER_EXPRESSION`. Because there could be some `#bar` deeper down (eg, `obj.#bar.#foo`) that is not declared.

```suggestion
        1. If StringValue of |PrivateIdentifier| is in _names_, then
          1. Return AllPrivateIdentifiersValid of |ShiftExpression| with argument _names_.
```

jridgewell (changes requested):

spec.html
bakkot: Agreed. #1668 originally had the same issue, in the other parts of AllPrivateIdentifiersValid.

While you're at it:

```suggestion
        1. If _names_ contains the StringValue of |PrivateIdentifier|, then
```

ljharb: Thanks! Both done.

jridgewell (approved):

nicolo-ribaudo (approved):

jmdyck (commented):
No complaints.

jugglinmike:
[Test262 now includes tests for this proposal.](https://github.com/tc39/test262/pull/2963)

WebReflection:
as `#x` *has* to be declared at the class level, hence always known, in which circumstance `#x in this` would fail, if I might ask?

bakkot:
`#x in this` will never throw, but it can return `false` if you do something like

```js
class A {
  #x = 0;
  hasX() {
    return #x in this;
  }
}
A.prototype.hasX.call({}); // false
```

bakkot:
No? It's only the class which has is able to do this check; you can't write `#x in this` _outside_ of `A`.

ljharb:
@WebReflection `this` isn't the only object that can hold a private field; imagine `static compare(a, b) { return #x in a && #x in b && a.#x === b.#x; }`

WebReflection:
OK, the class “owns” the private field access, but I wonder what a Proxy trap would do there, and if this check could lead to “leaking” exploits. If Proxy can’t intercept these check then fine, but I didn’t read anything about proxies in the text. If I missed that, no issue, if I didn’t, are private fields leaking via Proxy traps?

ljharb:
@WebReflection Proxy has no ability to detect or intercept private fields, or internal slots.

WebReflection:
Thanks! Please ignore my comments then, I had a mix of a-aha and “the heck?” moment there. All good 👍

syg (approved):

bakkot (approved):

PR #2294, opened 2021-02-04
Editorial: Remove unnecessary statement following heading for `find`.

DanielRosenwasser:
I found this while reading over the proposed spec text for `findLast`. It seems like no other function/method contains text like this.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -31932,7 +31932,6 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-array.prototype.find">
         <h1>Array.prototype.find ( _predicate_ [ , _thisArg_ ] )</h1>
-        <p>The `find` method is called with one or two arguments, _predicate_ and _thisArg_.</p>
         <emu-note>
           <p>_predicate_ should be a function that accepts three arguments and returns a value that is coercible to a Boolean value. `find` calls _predicate_ once for each element of the array, in ascending order, until it finds one where _predicate_ returns *true*. If such an element is found, `find` immediately returns that element value. Otherwise, `find` returns *undefined*.</p>
           <p>If a _thisArg_ parameter is provided, it will be used as the *this* value for each invocation of _predicate_. If it is not provided, *undefined* is used instead.</p>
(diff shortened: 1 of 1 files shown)


bakkot:
There's a few categories of these, it looks like:


- `.concat`/`.every`/`.filter`/`.findIndex`/`.flat`/`.flatMap`/`.forEach`/`.indexOf`/`.lastIndexOf`/`.map`/`.push`/`.reduce`/`.reduceRight`/`.some`/`.unshift` have "When the X method is called with [arguments], [optional description]. The following steps are taken:"
  - as a variant, `.includes` does not include the "with [arguments]" part, despite taking an argument
- `.slice` and `.splice` have a bare "The following steps are taken:"
- `.copyWithin`/`.fill` have "The X method takes up to X arguments A, B, and C." [yes, there is no punctuation after "arguments"]
- `.find` has both "The find method is called with one or two arguments, predicate and thisArg." and "When the find method is called, the following steps are taken:"
- `.join` has "The join method takes one argument, separator, and performs the following steps:"

(I skipped a few.) And that's just the Array prototype methods; I imagine there's other variants lurking out there. At a quick skim, most do seem to use either the "When the X is called" form or a bare "The following steps are taken:".

Clearly there's more work to do. Ideally someday we'd have something along the lines of https://github.com/tc39/ecma262/pull/1914 / https://github.com/tc39/ecma262/pull/1994 / https://github.com/tc39/ecma262/pull/2253, at least where it makes sense to do so.

---

Concretely, I think for this PR it would be good to additionally reword the sentence following the NOTE to add "with one or two arguments", as in "When the find method is called with one or two arguments, the following steps are taken:", which will make it match most of the rest of the Array prototype methods.

DanielRosenwasser:
I think we're good to go now.

bakkot (approved):

michaelficarra:
I don't know why we're bothering with this. We're going to just remove all this nonsense from the Array.prototype section anyway. I'll send a PR.

bakkot:
Thanks for kicking this off, @DanielRosenwasser. I believe it's now addressed a little more comprehensively by https://github.com/tc39/ecma262/pull/2346, so closing this, but feel free to reopen if you think there's more to do here.

PR #2296, opened 2021-02-06
Editorial: remove unnecessary assertion

ljharb:
Both S and R are Strings, and both are asserted in the prose, so either both or neither should be asserted in the steps. I chose neither.


jmdyck:
That's consistent with [this comment](https://github.com/tc39/ecma262/pull/545#issuecomment-559292107) in PR #545, which proposes that:
- a structured header (currently reflected by a standardized preamble) would give parameter invariants,
- ecmarkup would render these as Assert steps,
- and so any equivalent assertions that are already present should be removed.

michaelficarra (approved):

syg (approved):

PR #2297, opened 2021-02-06
Editorial: add <dfn> for "TypedArray element type"

ljharb:
This is solving the problem that the term "TypedArray element type" appears 8 times in the spec as an implied assertion on AO arguments and similar, but is not auto-linked or explicitly defined anywhere.

This first attempt is making the table column name a `<dfn>`, but this might look a little weird since it italicizes the column name.

The other alternative is to change the prose from `supported element types` to `supported <dfn>TypedArray element type<dfn>s` or something similar, but i'm not sure of the cleanest way to do that.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -32870,7 +32870,7 @@ THH:mm:ss.sss
 
   <emu-clause id="sec-typedarray-objects">
     <h1>TypedArray Objects</h1>
-    <p>_TypedArray_ objects present an array-like view of an underlying binary data buffer (<emu-xref href="#sec-arraybuffer-objects"></emu-xref>). Each element of a _TypedArray_ instance has the same underlying binary scalar data type. There is a distinct _TypedArray_ constructor, listed in <emu-xref href="#table-the-typedarray-constructors"></emu-xref>, for each of the supported element types. Each constructor in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> has a corresponding distinct prototype object.</p>
+    <p>_TypedArray_ objects present an array-like view of an underlying binary data buffer (<emu-xref href="#sec-arraybuffer-objects"></emu-xref>). A <dfn>TypedArray element type</dfn> is the underlying binary scalar data type that all elements of a _TypedArray_ instance have. There is a distinct _TypedArray_ constructor, listed in <emu-xref href="#table-the-typedarray-constructors"></emu-xref>, for each of the supported element types. Each constructor in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> has a corresponding distinct prototype object.</p>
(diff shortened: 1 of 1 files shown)


jmdyck:
There are 27 occurrences of "Element Type" that are referring to this column of this table, so if you change the column's name/header, you'd presumably have to change those too.

ljharb:
ah, i see what you mean. That's a good enough reason to rework the prose, instead.

spec.html
```diff
@@ -32933,7 +32933,7 @@ <h1>%ArrayIteratorPrototype% [ @@toStringTag ]</h1>
...
     <h1>TypedArray Objects</h1>
-    <p>_TypedArray_ objects present an array-like view of an underlying binary data buffer (<emu-xref href="#sec-arraybuffer-objects"></emu-xref>). Each element of a _TypedArray_ instance has the same underlying binary scalar data type. There is a distinct _TypedArray_ constructor, listed in <emu-xref href="#table-the-typedarray-constructors"></emu-xref>, for each of the supported element types. Each constructor in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> has a corresponding distinct prototype object.</p>
+    <p>_TypedArray_ objects present an array-like view of an underlying binary data buffer (<emu-xref href="#sec-arraybuffer-objects"></emu-xref>). Each element of a _TypedArray_ instance has the same underlying binary scalar data type. There is a distinct _TypedArray_ constructor, listed in <emu-xref href="#table-the-typedarray-constructors"></emu-xref>, for each of the supported element types. Each constructor in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> has a corresponding distinct prototype object, and each Element Type value is a <dfn>TypedArray element type</dfn>.</p>
```
syg, line 32936: It's a little weird to have the definition at the end of the paragraph.

Maybe something like

```suggestion
    <p>_TypedArray_ objects present an array-like view of an underlying binary data buffer (<emu-xref href="#sec-arraybuffer-objects"></emu-xref>). A <dfn>TypedArray element type</dfn> is the underlying binary scalar data type that all elements of a _TypedArray_ instance have. There is a distinct _TypedArray_ constructor, listed in <emu-xref href="#table-the-typedarray-constructors"></emu-xref>, for each of the supported element types. Each constructor in <emu-xref href="#table-the-typedarray-constructors"></emu-xref> has a corresponding distinct prototype object.</p>
```

michaelficarra: I like this. Would it be better to explicitly list them all here? It's just a spec enum, right?

syg: No, we shouldn't explicitly list them in prose. This paragraph is right above the table where there's an Element Type column that lists them all.

michaelficarra: Oh I see, that's great.

syg (approved):

michaelficarra (approved):

PR #2303, opened 2021-02-10
Editorial: Modernize overview paragraph comparing ES to other PLs

syg:
Closes #2266

References to other languages' specifications and papers are also
removed, as the editor group does not wish to keep an up-to-date
non-normative bibliography to other specifications (some of which are
withdrawn, like Scheme's). Additionally, web searching technology has
advanced sufficiently that finding information on the other PLs should
be easy.


syg:
Any thoughts on other languages to cite similarity for? I know @michaelficarra mentioned CoffeeScript, though it feels odd to cite similarity to a language to compiles to this one.

ljharb (approved):
I don't think we need to exhaustively list all the languages that have informed JS' later design; that would be a very long and thus not very useful list.

syg:
The conclusion on the editor call is to remove this paragraph entirely -- its value add isn't clear, and those more interested in the history and inspiration (instead of mere similarities) should read [Allen's history paper](http://www.wirfs-brock.com/allen/posts/866).

I'll open a new PR to remove this paragraph.

PR #2312, opened 2021-02-17
Editorial: Remove paragraph listing similarities to other languages

syg:
As discussed on the editor call, the feeling is that there isn't enough added value here for this paragraph in a technical specification.


ljharb (approved):

bakkot (approved):
In particular, there's now a great many languages which have influenced ecmascript, and/or been influenced in turn, and it doesn't really make sense to try to enumerate all of them nor any particular subset.

PR #2316, opened 2021-02-19
Normative: Use HostJobCallbacks for FinalizationRegistry callbacks

syg:
This aligns FinalizationRegistry callbacks with Promise callbacks in
allowing web browsers to perform additional steps to save and restore
the correct incumbent settings object.

Context is that one of the blockers in the HTML integration PR is proper handling of backup incumbent settings objects. I won't recapture that horror here, but we went over it for the Promise machinery and got consensus to add the HostJobCallback machinery. This PR just makes use of that machinery too.

/cc @codehag since Firefox is the only one that implements this behavior. Chrome is interested in aligning with the spec here, but is not prioritizing the implementation work.

(I meant to do this before Stage 4, but I... guess I forgot.)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -10171,7 +10171,7 @@
       1. While _finalizationRegistry_.[[Cells]] contains a Record _cell_ such that _cell_.[[WeakRefTarget]] is ~empty~, an implementation may perform the following steps:
         1. Choose any such _cell_.
         1. Remove _cell_ from _finalizationRegistry_.[[Cells]].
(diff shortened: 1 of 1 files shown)


ljharb (approved):

codehag:
We support this, and iirc have already implemented the intended behavior.

ljharb:
(presumably this doesn't need tests because it's not really testable?)

syg:
This is not testable in test262, that's right. There can be WPT tests for this, though.

syg:
This should be good to merge.

bakkot (approved):

michaelficarra (approved):

PR #2323, opened 2021-02-23
Editorial: nest ASI sections under "ASI" section

ljharb:
I noticed that https://tc39.es/ecma262/#sec-asi-cases-with-no-lineterminator-here seems oddly sitting as a sibling of https://tc39.es/ecma262/#sec-interesting-cases-of-automatic-semicolon-insertion, whereas it feels like 12.10 should be the container section for all ASI-related content.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -14513,43 +14513,43 @@
       <pre><code class="javascript">a = b + c(d + e).print()</code></pre>
       <p>In the circumstance that an assignment statement must begin with a left parenthesis, it is a good idea for the programmer to provide an explicit semicolon at the end of the preceding statement rather than to rely on automatic semicolon insertion.</p>
     </emu-clause>
-  </emu-clause>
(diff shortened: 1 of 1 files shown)


jmdyck:
> it feels like 12.10 should be the container section for all ASI-related content.

... apart from the content in 12.9, you mean?

Maybe 12.9 should contain all ASI-related content, and 12.10 + 12.11 should become 12.9.x?

ljharb:
hm, yes that would be an alternative - I'm fine with that, but either way 12.11 didn't seem like it should stand by itself, and 12.10 seems like it was intended to contain it.

bmeck:
I am neutral to either solution

ljharb:
Discussed on the editor call; I'll move 12.10 and 12.11 to be underneath 12.9, and will mark this PR as ready for review once done.

michaelficarra (approved):

bakkot (approved):

PR #2335, opened 2021-03-01
Editorial: Define Module Namespace's [[GetPrototypeOf]] method

shvaikalesh:
Module Namespace's [`[[GetPrototypeOf]]`](https://262.ecma-international.org/6.0/#sec-module-namespace-exotic-objects-getprototypeof) method, introduced in ES6, was replaced with a `[[Prototype]]` slot in #748 because [`SetImmutablePrototype`](https://tc39.es/ecma262/#sec-set-immutable-prototype) didn't call `[[GetPrototypeOf]]` at the time (fixed in #841).

This change makes definition of Module Namespace's `prototype` consistent with its definition of (non-)extensibility.
AWB argued for this approach in https://github.com/tc39/ecma262/pull/858#discussion_r136681103.


ljharb (approved):

syg (approved):
Seems like a good change to me to use the internal methods instead of directly accessing slots, if the reasons for the original workaround are now fixed.

michaelficarra (approved):

bakkot (approved):
Rebased. LGTM.

PR #2336, opened 2021-03-01
Editorial: Rework Module Namespace's [[Get]] method

shvaikalesh:
This PR merely inlines Module Namespace's [`[[Get]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver) into its [`[[GetOwnProperty]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-getownproperty-p) method because:
* [`[[Get]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver) doesn't (need to) terminate prototype chain traversal, unlike TypedArray's [`[[Get]]`](https://tc39.es/ecma262/#sec-integer-indexed-exotic-objects-get-p-receiver), since module namespace object's prototype is `null` and it's non-extensible.
* Some Module Namespace's methods like [`[[HasProperty]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-hasproperty-p) or [`[[Delete]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-delete-p) are subtly different from ordinary ones: they prevent from calling into [`[[GetOwnProperty]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-getownproperty-p) that may throw. We should avoid having methods that are totally equivalent to ordinary ones, like [`[[Get]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver), so they won't be accidentally ignored by implementors.
* It relieves runtimes from calling into [`[[Get]]`](https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver) override, and continue performing [`OrdinaryGet`](https://tc39.es/ecma262/#sec-ordinaryget), which is preferred for simplicity / performance reasons.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -11991,9 +11991,21 @@
           1. If Type(_P_) is Symbol, return OrdinaryGetOwnProperty(_O_, _P_).
           1. Let _exports_ be _O_.[[Exports]].
           1. If _P_ is not an element of _exports_, return *undefined*.
-          1. Let _value_ be ? _O_.[[Get]](_P_, _O_).
+          1. Let _m_ be _O_.[[Module]].
+          1. Let _binding_ be ! _m_.ResolveExport(_P_).
+          1. Assert: _binding_ is a ResolvedBinding Record.
+          1. Let _targetModule_ be _binding_.[[Module]].
+          1. Assert: _targetModule_ is not *undefined*.
+          1. If _binding_.[[BindingName]] is *"\*namespace\*"*, then
+            1. Return ? GetModuleNamespace(_targetModule_).
+          1. Let _targetEnv_ be _targetModule_.[[Environment]].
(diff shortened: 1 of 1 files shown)


syg:
We discussed this on the editor call, and the conclusion is:

- We like the change to get [[Get]] delegating to [[GetOwnProperty]], since that's the usual state of things.
- We'd like to keep an explicit [[Get]] instead of removing it for clarity. We recommend copy/pasting the relevant parts of ordinary objects' [[Get]], with the irrelevant parts omitted. A note that says this is just ordinary [[Get]] with irrelevant parts omitted would be helpful.
- Keep the steps inlined into [[GetOwnProperty]] as you currently have.

PR #2337, opened 2021-03-01
Meta: refactor scripts to ensure /2020/ build works properly

ljharb:


bakkot (approved):

package.json
```diff
@@ -6,17 +6,17 @@
   "scripts": {
     "ipr-check": "node scripts/check-form tc39/ecma262 \"$TRAVIS_COMMIT\"",
     "prebuild-spec": "git remote rm travis-origin 2>/dev/null ; git remote add travis-origin \"git@github.com:${TRAVIS_REPO_SLUG:-tc39/ecma262}.git\" && git fetch --quiet travis-origin",
-    "build-spec": "YEAR=2020 && git checkout --quiet \"es${YEAR}\" && mkdir -p \"out/${YEAR}\" && cp -R img \"out/${YEAR}\" && ecmarkup --verbose spec.html \"out/${YEAR}/index.html\" --css \"out/${YEAR}/ecmarkup.css\" --js \"out/${YEAR}/ecmarkup.js\" && git checkout --quiet travis-origin/test-travis",
+    "build-spec": "YEAR=2020 && OUT=\"out/${YEAR}\" && git checkout --quiet \"es${YEAR}\" && npm run build-only && git checkout --quiet travis-origin/test-travis",
```
bakkot, line 9: Actually, wait, this doesn't work, I think. Once `git checkout` has run, `npm run build-only` is going to attempt to pick up the build-only script from the package.json _in that branch_, which [doesn't exist](https://github.com/tc39/ecma262/blob/dfd5ea2ec5862de21b005737650ba08bc57271fa/package.json) (and even if it did it would not know about the `OUT` environment variable).

ljharb: yes, that's a good point.

SE-S-PI (commented):
Klass

PR #2341, opened 2021-03-09
Editorial: use target alias name instead of "it"

michaelficarra:
I noticed this when reviewing [the ResizableArrayBuffer proposal](https://tc39.es/proposal-resizablearraybuffer/), which has this same pattern.


bakkot (approved):

ljharb (approved):

jmdyck:
Another syntax to convey this is (e.g.):
```_arrayBuffer_ is an Object that has an [[ArrayBufferData]] internal slot```
(There are 12 occurrences of this usage in the spec.)

PR #2342, opened 2021-03-09
Editorial: Clarify ES2021 Array.prototype.sort change

mathiasbynens:
`Array.prototype.sort` was already made stable in ECMAScript 2019 (see two paragraphs above). The change in https://github.com/tc39/ecma262/pull/1585 made `Array.prototype.sort` more precise, reducing the amount of cases that result in an implementation-defined sort order.


ljharb (approved):
Thanks, I'll pull this into the candidate.

PR #2344, opened 2021-03-15
Editorial: remove redundant `index.html`

mathiasbynens:
There was some discussion on removing it here: https://github.com/tc39/ecma262/pull/2339#discussion_r585946339 So I wanted to offer this patch in case this was overlooked.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,9 +24,9 @@
       "integrity": "sha512-np/lG3uARFybkoHokJUmf1QfEvRVCPbmQeUQpKow5cQ3xWrV9i3rUHodKDJPQfTVX61qKi+UdYk8kik84n7XOw=="
     },
     "@babel/highlight": {
-      "version": "7.13.8",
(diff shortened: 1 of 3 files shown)


ljharb:
I'm super on board with this, but I don't think it will work with the deploy preview yet.

bakkot:
Yeah, this will need to be changed in ecmarkup first, not here. I was going to wait on that until the previews worked (which it looks like they maybe do now?).

mathiasbynens:
> Yeah, this will need to be changed in ecmarkup first, not here.

IIUC the two changes are separate, and we want both to happen:

- [ ] one particular instance of a hardcoded `index.html` should be removed (i.e. this patch)
- [ ] ecmarkup should be changed to emit `multipage/` instead of `multipage/index.html`

bakkot:
Right. I want both to happen at the same time, so that we're never simultaneously linking to the same content at two different URLs.

bakkot (approved):
I pushed a commit which bumps ecmarkup.

Issue #2304, opened 2021-02-11
Editorial: make introductions for built-in functions use consistent wording


bakkot:
There are a good number of different phrasings we use when introducing built-in functions, some of which I've documented [here](https://github.com/tc39/ecma262/pull/2294#issuecomment-773000412).

We should be more consistent.

michaelficarra:
My preferred phrasing is

> The &lt;name of function as written in header&gt; &lt;{function,method}&gt; performs the following steps when called:

If we want an additional informative description of the function's purpose, it can precede the phrase

> It performs the following steps when called:

instead.

bakkot:
Works for me.

I'd also be fine with listing the arguments as we do for AOs, but I can see the argument for not doing so: AOs are only ever called with the correct number of arguments, so when we speak of them being called with a particular number of arguments we are covering all possible cases, whereas these can in fact be called with any number of arguments and should follow the specified behavior regardless.

michaelficarra:
We also use "Its get accessor function performs the following steps:".

jmdyck:
Note that PR #2592 addresses this, among other things.

michaelficarra:
@jmdyck Can you put "Fixes #XXX" or "Closes #XXX" in your PR descriptions so they link properly?

jmdyck:
"descriptions" plural?

michaelficarra:
@jmdyck The descriptions of any PRs you either have open or open in the future.

jmdyck:
I didn't create #2592 to address this issue. When I submitted #2592, this PR was 9 months in the past, and I'd forgotten about it until an hour ago. Moreover, nobody else commenting on #2592 (or on issue #2576 that preceded it) brought up this issue, so I'm guessing they'd forgotten about it too. I can't promise to remember, now and in the future, every open issue that might be relevant to a PR, but I'm happy to make connections in my PRs if someone else remembers a relevant issue and brings it to my attention.

michaelficarra:
@jmdyck Of course, I don't expect you to go through every single open issue. Just when you become aware of them. Though performing a search for relevant issues or duplicates before opening a PR is usually a good practice.

ptomato:
@michaelficarra @bakkot Is the wording suggested in https://github.com/tc39/ecma262/issues/2304#issuecomment-777139392 with its two variations now the official editors' guidance, or does it need to be discussed further?

PR #2346, opened 2021-03-15
Editorial: clean up wording of Array methods

michaelficarra:
Obsoletes #2294. I moved a few paragraphs outside of the `<emu-note>` because they weren't notes, they were summaries of the method.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -31601,7 +31601,7 @@ THH:mm:ss.sss
 
       <emu-clause id="sec-array.from">
         <h1>Array.from ( _items_ [ , _mapfn_ [ , _thisArg_ ] ] )</h1>
-        <p>When the `from` method is called with argument _items_ and optional arguments _mapfn_ and _thisArg_, the following steps are taken:</p>
+        <p>When the `from` method is called, the following steps are taken:</p>
         <emu-alg>
           1. Let _C_ be the *this* value.
           1. If _mapfn_ is *undefined*, let _mapping_ be *false*.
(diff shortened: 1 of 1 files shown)


bakkot:
I note that this isn't quite [the wording you'd suggested earlier](https://github.com/tc39/ecma262/issues/2304#issuecomment-777139392).

michaelficarra:
It's good enough for me for now until we do a more strict normalisation across the whole spec.

bakkot:
You should probably make the corresponding changes to the corresponding TypedArray methods - not all of them need it, but [some of them do](https://tc39.es/ecma262/#sec-%typedarray%.prototype.every).

michaelficarra:
@bakkot Done.

jmdyck:
> I moved a few paragraphs outside of the `<emu-note>` because they weren't notes, they were summaries of the method.

I think there are a few more cases of this:
- A.p.every: "`every` calls _callbackfn_ once for each element..."
- A.p.filter: "`filter` calls _callbackfn_ once for each element..."
- A.p.find
- A.p.findIndex
- A.p.forEach
- A.p.includes
- A.p.lastIndexOf
- A.p.map
- A.p.pop
- A.p.push
- A.p.reduce
- A.p.reduceRight
- A.p.reverse
- A.p.some
- A.p.splice

(or they could wait for the "more strict normalisation")

michaelficarra:
@jmdyck Some of those are not quite summaries, so I'd rather leave them out for now. The main goal of this PR was to get rid of the numbering of the arguments.

jmdyck:
Yup, no problem.

bakkot (approved):

PR #2359, opened 2021-03-24
Editorial: Fix early error rules of 'IfStatement'

h2oche:
In [14.6.1 Static Semantics: Early Errors](https://tc39.es/ecma262/#sec-if-statement-static-semantics-early-errors), first production of `IfStatement` contains two `Statement`. However, early error rule of `IfStatement` only mentions about single `Statement`, which seems to be fixed by this PR.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -18048,11 +18048,16 @@
 
     <emu-clause id="sec-if-statement-static-semantics-early-errors">
       <h1>Static Semantics: Early Errors</h1>
-      <emu-grammar>
-        IfStatement :
-          `if` `(` Expression `)` Statement `else` Statement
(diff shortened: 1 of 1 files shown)


ljharb:
Duplicate of #2358. Not sure what happened there, but it'd be great to make sure we don't end up with a third PR as well :-)

h2oche:
Sorry for duplicated PR. There was some mistake while I made new branch. I'll be more careful next time.

bakkot (approved):

michaelficarra (approved):
In the longer term, I'd prefer superscript labels or some other marker for these kinds of ambiguities, but this works for now.

PR #2360, opened 2021-03-25
Editorial: Remove 'designed to be subclassable'

ptomato:
As discussed during the March 2021 plenary, this note on "is designed to
be subclassable" can be read as an encouragement or value judgement on
whether it is desirable to subclass built-in classes like Boolean. We
don't want to actively encourage this.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27944,7 +27944,7 @@
         <li>is the initial value of the *"Object"* property of the global object.</li>
         <li>creates a new ordinary object when called as a constructor.</li>
         <li>performs a type conversion when called as a function rather than as a constructor.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition.</li>
+        <li>may be used as the value of an `extends` clause of a class definition.</li>
       </ul>
 
       <emu-clause id="sec-object-value">
@@ -28388,7 +28388,7 @@
         <li>is <dfn>%Function%</dfn>.</li>
         <li>is the initial value of the *"Function"* property of the global object.</li>
         <li>creates and initializes a new function object when called as a function rather than as a constructor. Thus the function call `Function(&hellip;)` is equivalent to the object creation expression `new Function(&hellip;)` with the same arguments.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Function behaviour must include a `super` call to the Function constructor to create and initialize a subclass instance with the internal slots necessary for built-in function behaviour. All ECMAScript syntactic forms for defining function objects create instances of Function. There is no syntactic means to create instances of Function subclasses except for the built-in GeneratorFunction, AsyncFunction, and AsyncGeneratorFunction subclasses.</li>
+        <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Function behaviour must include a `super` call to the Function constructor to create and initialize a subclass instance with the internal slots necessary for built-in function behaviour. All ECMAScript syntactic forms for defining function objects create instances of Function. There is no syntactic means to create instances of Function subclasses except for the built-in GeneratorFunction, AsyncFunction, and AsyncGeneratorFunction subclasses.</li>
       </ul>
 
       <emu-clause id="sec-function-p1-p2-pn-body">
@@ -28723,7 +28723,7 @@
         <li>is the initial value of the *"Boolean"* property of the global object.</li>
         <li>creates and initializes a new Boolean object when called as a constructor.</li>
         <li>performs a type conversion when called as a function rather than as a constructor.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Boolean behaviour must include a `super` call to the Boolean constructor to create and initialize the subclass instance with a [[BooleanData]] internal slot.</li>
+        <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Boolean behaviour must include a `super` call to the Boolean constructor to create and initialize the subclass instance with a [[BooleanData]] internal slot.</li>
       </ul>
 
       <emu-clause id="sec-boolean-constructor-boolean-value">
@@ -29095,7 +29095,7 @@
         <li>is <dfn>%Error%</dfn>.</li>
         <li>is the initial value of the *"Error"* property of the global object.</li>
         <li>creates and initializes a new Error object when called as a function rather than as a constructor. Thus the function call `Error(&hellip;)` is equivalent to the object creation expression `new Error(&hellip;)` with the same arguments.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Error behaviour must include a `super` call to the Error constructor to create and initialize subclass instances with an [[ErrorData]] internal slot.</li>
+        <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Error behaviour must include a `super` call to the Error constructor to create and initialize subclass instances with an [[ErrorData]] internal slot.</li>
       </ul>
 
       <emu-clause id="sec-error-message">
@@ -29226,7 +29226,7 @@
         <p>Each _NativeError_ constructor:</p>
         <ul>
           <li>creates and initializes a new _NativeError_ object when called as a function rather than as a constructor. A call of the object as a function is equivalent to calling it as a constructor with the same arguments. Thus the function call <code><var>NativeError</var>(&hellip;)</code> is equivalent to the object creation expression <code>new <var>NativeError</var>(&hellip;)</code> with the same arguments.</li>
-          <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified _NativeError_ behaviour must include a `super` call to the _NativeError_ constructor to create and initialize subclass instances with an [[ErrorData]] internal slot.</li>
+          <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified _NativeError_ behaviour must include a `super` call to the _NativeError_ constructor to create and initialize subclass instances with an [[ErrorData]] internal slot.</li>
         </ul>
 
         <emu-clause id="sec-nativeerror">
@@ -29302,7 +29302,7 @@
           <li>is <dfn>%AggregateError%</dfn>.</li>
           <li>is the initial value of the *"AggregateError"* property of the global object.</li>
           <li>creates and initializes a new AggregateError object when called as a function rather than as a constructor. Thus the function call `AggregateError(&hellip;)` is equivalent to the object creation expression `new AggregateError(&hellip;)` with the same arguments.</li>
-          <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified AggregateError behaviour must include a `super` call to the AggregateError constructor to create and initialize subclass instances with an [[ErrorData]] internal slot.</li>
+          <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified AggregateError behaviour must include a `super` call to the AggregateError constructor to create and initialize subclass instances with an [[ErrorData]] internal slot.</li>
         </ul>
 
         <emu-clause id="sec-aggregate-error">
@@ -29385,7 +29385,7 @@
         <li>is the initial value of the *"Number"* property of the global object.</li>
         <li>creates and initializes a new Number object when called as a constructor.</li>
         <li>performs a type conversion when called as a function rather than as a constructor.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Number behaviour must include a `super` call to the Number constructor to create and initialize the subclass instance with a [[NumberData]] internal slot.</li>
+        <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Number behaviour must include a `super` call to the Number constructor to create and initialize the subclass instance with a [[NumberData]] internal slot.</li>
       </ul>
 
       <emu-clause id="sec-number-constructor-number-value">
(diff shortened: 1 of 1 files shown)


ptomato:
Hi, I thought it'd be good to start a discussion around the removal of this phrase (which was mentioned in plenary) in order to provide a basis for the language in Temporal to be consistent with. Please do suggest alternatives :smile:

spec.html
```diff
@@ -24492,7 +24492,7 @@ <h1>The Object Constructor</h1>
         <li>is the initial value of the *"Object"* property of the global object.</li>
         <li>creates a new ordinary object when called as a constructor.</li>
         <li>performs a type conversion when called as a function rather than as a constructor.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition.</li>
+        <li>may be used as the value of an `extends` clause of a class definition.</li>
```
ljharb, line 24495: this is definitely better! but:
```suggestion
        <li>may be used as the value of an `extends` clause of a class definition, but is not intended to be subclassed.</li>
```
perhaps? (at least for all the non-nullish primitive constructors, where it just doesn't make sense to ever subclass them)

spec.html
```diff
@@ -27366,7 +27366,7 @@ <h1>The Date Constructor</h1>
         <li>creates and initializes a new Date object when called as a constructor.</li>
         <li>returns a String representing the current time (UTC) when called as a function rather than as a constructor.</li>
         <li>is a function whose behaviour differs based upon the number and types of its arguments.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Date behaviour must include a `super` call to the Date constructor to create and initialize the subclass instance with a [[DateValue]] internal slot.</li>
+        <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Date behaviour must include a `super` call to the Date constructor to create and initialize the subclass instance with a [[DateValue]] internal slot.</li>
```
ljharb, line 27369: Date seems like it works fine as being subclassable

spec.html
```diff
@@ -30767,7 +30767,7 @@ <h1>The RegExp Constructor</h1>
         <li>is <dfn>%RegExp%</dfn>.</li>
         <li>is the initial value of the *"RegExp"* property of the global object.</li>
         <li>creates and initializes a new RegExp object when called as a function rather than as a constructor. Thus the function call `RegExp(&hellip;)` is equivalent to the object creation expression `new RegExp(&hellip;)` with the same arguments.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified RegExp behaviour must include a `super` call to the RegExp constructor to create and initialize subclass instances with the necessary internal slots.</li>
+        <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified RegExp behaviour must include a `super` call to the RegExp constructor to create and initialize subclass instances with the necessary internal slots.</li>
```
ljharb, line 30770: sadly, RegExp was explicitly designed _to_ be subclassable

Jack-Works: I'm still curious. Does anyone use subclassable RegExp in wild? I never seen that

spec.html
```diff
@@ -25626,7 +25626,7 @@ <h1>The Error Constructor</h1>
         <li>is <dfn>%Error%</dfn>.</li>
         <li>is the initial value of the *"Error"* property of the global object.</li>
         <li>creates and initializes a new Error object when called as a function rather than as a constructor. Thus the function call `Error(&hellip;)` is equivalent to the object creation expression `new Error(&hellip;)` with the same arguments.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Error behaviour must include a `super` call to the Error constructor to create and initialize subclass instances with an [[ErrorData]] internal slot.</li>
+        <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Error behaviour must include a `super` call to the Error constructor to create and initialize subclass instances with an [[ErrorData]] internal slot.</li>
```
Jack-Works, line 25629: Actually, I do see and use subclassing on the Error constructor. Isn't it _is_ designed subclassable?

spec.html
```diff
@@ -31492,7 +31492,7 @@ <h1>The Array Constructor</h1>
...
         <li>is a function whose behaviour differs based upon the number and types of its arguments.</li>
-        <li>is designed to be subclassable. It may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the exotic Array behaviour must include a `super` call to the Array constructor to initialize subclass instances that are Array exotic objects. However, most of the `Array.prototype` methods are generic methods that are not dependent upon their *this* value being an Array exotic object.</li>
+        <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the exotic Array behaviour must include a `super` call to the Array constructor to initialize subclass instances that are Array exotic objects. However, most of the `Array.prototype` methods are generic methods that are not dependent upon their *this* value being an Array exotic object.</li>
```
Jack-Works, line 31495: The same, subclassing Array is really an important feature in ES6, why discourage that

spec.html
```diff
@@ -34249,7 +34249,7 @@ <h1>The Map Constructor</h1>
         <li>is the initial value of the *"Map"* property of the global object.</li>
         <li>creates and initializes a new Map object when called as a constructor.</li>
         <li>is not intended to be called as a function and will throw an exception when called in that manner.</li>
-        <li>is designed to be subclassable. It may be used as the value in an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Map behaviour must include a `super` call to the Map constructor to create and initialize the subclass instance with the internal state necessary to support the `Map.prototype` built-in methods.</li>
+        <li>may be used as the value in an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified Map behaviour must include a `super` call to the Map constructor to create and initialize the subclass instance with the internal state necessary to support the `Map.prototype` built-in methods.</li>
```
Jack-Works, line 34252: Same as Array/Set/Error case

Jack-Works:
In my personal programming use case, I have subclasses the following classes and found they're really useful. I don't think we should discourage those classes to be extended.

- Array
- Error
- Map
- Set
- WeakMap
- WeakSet

spec.html
ljharb: Neither have i, but the committee decided not to remove that plumbing, and it was definitely designed for it.

bakkot: > sadly, RegExp was explicitly designed _to_ be subclassable

Doesn't mean we need to mention that in the specification, though.

ljharb: I think it’s reasonable to denote the things that are built for it, to separate them from the ones that aren’t.

bakkot:
Many of these were originally _designed_ to be subclassable. All of them work to at least some extent in an `extends` clause (though good luck extending `Function` if you might be running on webpages with CSP). Some of them it makes more sense to subclass, others less, though exactly which and to what extend will depend on your point of view.

However, we aren't obligated to mention everything design decision we've ever made. My preference is to just drop mention of subclassing from all built-in constructors. We don't need to take a position on whether it's a good idea to subclass these, whether in general or for some particular subset. We can simply not discuss it at all.

ljharb:
I think it’s very important that we actively discourage subclassing of the things that aren’t meant for it. I’m not sure how to do that without explicitly mentioning that some things are meant for it.

ptomato:
Personally I agree with @bakkot here: removing the phrase doesn't discourage subclassing. But I'm not an editor, so having given my opinion I'll happily change this to reflect whatever comes out of the discussion, and then change Temporal accordingly.

ljharb:
I think this goes beyond a purely editorial question, personally, since user expectations from prose in the spec do often impact normative behaviors - just like the same discussion around Temporal.

spec.html
syg: Error doesn't have any methods outside of `toString`, which side steps the issues with subclassing. Still I don't see any reason to deviate from the proposed language here. I see this PR as removing encouragement, and IMO, the fact that Error can be successfully subclassed doesn't mean the spec needs to encourage it over other built-ins that can't be easily successfully subclassed.

Further, by removing the encouragement, we're also saying something about future proposals for extending Error: that they shouldn't accommodate subclassing beyond the constructor being usable in `extends`.

spec.html
syg: Similar response here as above for Error.

spec.html
syg: I disagree that we should explicitly mark things that are built for it, for the reason stated above: that we're nudging future proposals. Yes, it was explicitly designed for it in the past, a past that we regret.

spec.html
syg: I disagree? I guess it depends on what you mean by subclassing, since in JS it encompasses stuff like species construction, and that methods can deal with non-Array objects that still have the Array exotic objects' slots and essential methods.

What about subclassing Arrays is important beyond being able to `extend` them?

spec.html
syg: Same question as above then: what about the subclassing facilities, aside from being able to extend Maps, is important?

syg:
Strong agree with @bakkot. "Subclassing" as an umbrella term encompasses several mechanisms, some of which I actively despise, and some of which are fine. I like this PR in that it calls out a particular mechanism as being mechanism, and removing encouragement for using the whole set of mechanisms.

@ljharb and @Jack-Works, if you feel some of those mechanisms _should_ be encouraged, then let's discuss those. I'm otherwise willing to die on the hill of removing blanket encouragement of "subclassing".

ljharb:
@syg I'm not attached to the term "subclassing" for the precise reason you mention - that it conflates multiple mechanisms. I think it is important to explicitly distinguish the builtins that can meaningfully be the RHS of an `extends` - either by calling them out, or by calling out the rest. Simply removing the terminology leaves us in the same unfortunate position we're in now: users receive no guidance as to which builtin subclasses make sense (even if only a little) and which make zero sense.

bathos:
I agree with @bakkot that this is very subjective and it's apparent that people don't share the same definition of "subclass". Some of it is provably untrue ("must include a `super` call"). Why replace one iffy/fuzzy text with an iffier, fuzzier one?

syg:
> do you mean https://github.com/tc39/proposal-rm-builtin-subclassing#taxonomy-of-subclassing?

Yes.

syg:
> I agree with @bakkot that this is very subjective and it's apparent that people don't share the same definition of "subclass". Some of it is provably untrue ("must include a super call"). Why replace one iffy/fuzzy text with an iffier, fuzzier one?

What's being replaced? This PR drops the "designed to be subclassable" language, the other language is existing.

bathos:
@syg You’re right, sorry — that stemmed from misreading. The real changes here seem like improvements to me.

ptomato:
Have the editors had a chance to discuss this? It seems like it is blocked on whether the removal of the phrase "is designed to be subclassable" is interpreted as a deprecation, or as a neutral removal of encouragement.

bakkot:
Editors discussed this and are happy about it as-is among ourselves. Since there seems to be disagreement about whether this is strictly an editorial decision, I'm going to put together a very short presentation and ask that we get consensus on this particular change at the next meeting, emphasizing that a.) this PR does not take a position against (or for) subclassing any particular constructors and b.) the current state of affairs, which says that Boolean is "designed to be subclassable", is silly. If people want to call out some particular subset as being "designed to be subclassable", and/or call out some other subset as explicitly not for subclassing, they are welcome to drive that through committee as a followup.

allenwb:
> which says that Boolean is "designed to be subclassable", is silly.

"designed to be subclassable" was used in the ES6 specification with a very specific meaning:

- if the constructor is correctly `super` invoked by a subclass constructor
  - subclass instances will be the same kind of ordinary/exotic object as those created by the "superclass" constructor
  - subclass instances will have the same internal slots as defined by the "superclass"
  - prototype methods defined by the "superclass" that have dependencies upon internal slots or exotic internal methods will function correctly when invoked upon subclass instances 

ES6 Boolean was, in fact, "defined to be subclassable" exactly according to those criteria. Specifically, Boolean.prototype.toString and valueOf are specified such that they make use of [[BooleanData]] internal slot of a Boolean subclass. 

There may be very little utility in subclassing Boolean, but if somebody does it will work. The incorrect operation of ad hoc subclasses of Array and other builtins was something that developers complained about in the ES3/ES5 era. Fixing that general problem was an ES6 objective.  For both internal and external consistency "defined to be subclassable" was the default design principle for ES6 builtins, even with there was minimal utility in subclassing some builtins.  Only in a few cases such as Symbol, were there strong reasons to violate the consistency and design principle and make a builtin non-"subclassable".

Because "designed to be subclassable" was the ES6 norm, that label arguably  was not need on every builtin constructor.  But as a change from prior editions of the specification the label served as an important notice to both ES implementors and developers.  Perhaps it is no longer needed, but the fact that has been this ongoing discussion make me think it still has value.

bakkot:
@allenwb, thanks for giving the history. It's the opinion of the current editor group that the sense of "designed to be subclassable" used in your comment is adequately captured by the "may be used as the value of an `extends` clause of a class definition" part, which will remain. The advantage of removing specifically the "subclassable" part is that some people read it to imply more things than are present in the definition you intended (for example, people have read it as implicitly encouraging subclassing of built-ins, rather than merely noting that it would work).

spec.html
zloirock: > I'm still curious. Does anyone use subclassable RegExp in wild? I never seen that

[At least Babel.](https://babeljs.io/repl#?browsers=ie%2011&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=G4QwTgBGCmELwQPQAoD8AeAntcA-AOgCYDeALAL4CUAtGugLYD2AdgC4AWBJATFbRoRCYuxXpUQBzANxA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=true&targets=&version=7.14.4&externalPlugins=)

ljharb: From reading that output, it seems like Babel is doing that primarily so that it interoperates properly just in case someone else *is* using subclassable regexps - i'd bet they could drastically simplify their output if they didn't have to worry about all the regexp symbols.

bakkot (approved):

michaelficarra (approved):

bigwill1200 (approved):

spec.html
zloirock: @ljharb could you show an example that will properly work with `.exec`, `.match`, `.replace` without it?

ljharb: If “properly work” means “RegExp subclasses don’t work, which is fine because nobody actually uses them except for core-js feature detection”, then they’d just use those methods directly. If by “properly work” you mean that RegExp subclasses continue to work, then no, of course those are required.

zloirock: @ljharb by "properly" work I mean how to make work extended transpiled functionality of `RegExp` - NCG in this case work with `String.prototype.match` and `String.prototype.replace`. It's absolutely not related to polyfills - it's the implementation of ES2018 feature on ES2015.

ljharb: For that, you’re probably right - but “making transpilers possible” isn’t a primary concern when designing features.

zloirock: However, it's a significant plus.

ljharb: When possible, yes, but it's not worth the complexity and performance hits that the RegExp system takes. Unfortunately, we can't remove it because there's no practical way to tease apart polyfill feature detections from real code when gathering usage data.

PR #2366, opened 2021-04-01
Editorial: add usage note

bakkot:
I felt that Annex B was not sufficiently clear about its contents, so I've added a [note](https://en.wikipedia.org/wiki/Long-time_nuclear_waste_warning_messages#Message) about the intent of that section.

We can discuss in plenary the possibility of having this to apply to the whole spec. I'm sure everyone will think it's a good idea; I'll add a five-minute agenda item.

<details>
<summary>rendered</summary>
Screen Shot 2021-04-01 at 8 40 13 AM
</details>

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -40814,6 +40814,17 @@ THH:mm:ss.sss
 
 <emu-annex id="sec-additional-ecmascript-features-for-web-browsers" namespace="annexB" normative>
   <h1>Additional ECMAScript Features for Web Browsers</h1>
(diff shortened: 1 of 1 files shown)


bakkot:
On reflection, the phrasing in this PR may slightly overstate the risks of Annex B.

ljharb:
Only very, very slightly.

syg:
Sorry for late review, lgtm.

PR #2374, opened 2021-04-05
Editorial: more explicit validation in CreateDynamicFunction

### LINKED ISSUE #2373: URGENT SECURITY Function constructor spec creates massive code injection vulnerability
Somehow the text of the Function constructor specs was accidentally changed in a way that, if implemented, would create massive code injection vulnerabilities for code already deployed.

https://tc39.es/ecma262/#sec-function-p1-p2-pn-body

The old spec, and the behavior of v8 (Brave, Chromium, Chrome, Node), SpiderMonkey (Firefox), and JSC (Safari), (and IIRC Edge, though I cannot currently test that) all agree that the code

```js
new Function(attackerProvidedString);
```
would validate whether the `attackerProvidedString` is a syntactically valid *FunctionBody* (or *GeneratorBody*, ...) without any possibility that attacker provided code would get run as a consequence. If the string is a valid FunctionBody, it would return a function, and otherwise it would throw a SyntaxError.

The current definition creates a source string by *concatenation* and then parses only the concatenated string. Thus
```js
new Function('}), a=1, (function(){');
```
would be concatenated into the source string
```js
function anonymous(
)) {}), a=1, (function(){}
```
which parses, with the resulting ast meeting all the requirements set out in the rest of that spec. The Function constructor spec would not itself cause the injected code to execute. But other code treats this success as validating that this string is a syntactically valid function body, that may then be safely concatenated in exactly this way without creating an injection attack.

See https://github.com/Moddable-OpenSource/moddable/issues/623

Attn @allenwb @Waldemar @littledan @ljharb @codehag When and how did this get broken?

bakkot:
Fixes #2373. cc @erights

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -24956,20 +24956,24 @@
             1. If _newTarget_ is *undefined*, set _newTarget_ to _constructor_.
             1. If _kind_ is ~normal~, then
               1. Let _exprSym_ be the grammar symbol |FunctionExpression|.
-              1. Let _bodySym_ be the grammar symbol |FunctionBody|.
+              1. Let _bodySym_ be the grammar symbol |FunctionBody[~Yield, ~Await]|.
+              1. Let _parameterSym_ be the grammar symbol |FormalParameters[~Yield, ~Await]|.
               1. Let _fallbackProto_ be *"%Function.prototype%"*.
             1. Else if _kind_ is ~generator~, then
               1. Let _exprSym_ be the grammar symbol |GeneratorExpression|.
               1. Let _bodySym_ be the grammar symbol |GeneratorBody|.
+              1. Let _parameterSym_ be the grammar symbol |FormalParameters[+Yield, ~Await]|.
               1. Let _fallbackProto_ be *"%GeneratorFunction.prototype%"*.
             1. Else if _kind_ is ~async~, then
               1. Let _exprSym_ be the grammar symbol |AsyncFunctionExpression|.
               1. Let _bodySym_ be the grammar symbol |AsyncFunctionBody|.
+              1. Let _parameterSym_ be the grammar symbol |FormalParameters[~Yield, +Await]|.
               1. Let _fallbackProto_ be *"%AsyncFunction.prototype%"*.
             1. Else,
               1. Assert: _kind_ is ~asyncGenerator~.
               1. Let _exprSym_ be the grammar symbol |AsyncGeneratorExpression|.
               1. Let _bodySym_ be the grammar symbol |AsyncGeneratorBody|.
+              1. Let _parameterSym_ be the grammar symbol |FormalParameters[+Yield, +Await]|.
               1. Let _fallbackProto_ be *"%AsyncGeneratorFunction.prototype%"*.
             1. Let _argCount_ be the number of elements in _args_.
             1. Let _P_ be the empty String.
@@ -24990,10 +24994,14 @@
             1. Let _prefix_ be the prefix associated with _kind_ in <emu-xref href="#table-dynamic-function-sourcetext-prefixes"></emu-xref>.
             1. Let _sourceString_ be the string-concatenation of _prefix_, *" anonymous("*, _P_, 0x000A (LINE FEED), *") {"*, _bodyString_, and *"}"*.
             1. Let _sourceText_ be ! StringToCodePoints(_sourceString_).
+            1. Let _parameters_ be ParseText(! StringToCodePoints(_P_), _parameterSym_).
+            1. If _parameters_ is a List of errors, throw a *SyntaxError* exception.
+            1. Let _body_ be ParseText(! StringToCodePoints(_bodyString_), _bodySym_).
+            1. If _body_ is a List of errors, throw a *SyntaxError* exception.
+            1. NOTE: The parameters and body are parsed separately to ensure that each is valid alone. For example, `new Function("/*", "*/ ) {")` is not legal.
+            1. NOTE: If this step is reached, _sourceText_ must match _exprSym_ (although the reverse implication does not hold). The purpose of the next two steps is to enforce any Early Error rules which apply to _exprSym_ directly.
             1. Let _expr_ be ParseText(_sourceText_, _exprSym_).
             1. If _expr_ is a List of errors, throw a *SyntaxError* exception.
-            1. Let _parameters_ be the |FormalParameters| of _expr_.
-            1. Let _body_ be the child of _expr_ that is an instance of _bodySym_.
             1. Let _proto_ be ? GetPrototypeFromConstructor(_newTarget_, _fallbackProto_).
             1. Let _realmF_ be the current Realm Record.
             1. Let _scope_ be _realmF_.[[GlobalEnv]].


spec.html
```diff
@@ -24990,10 +24994,13 @@ <h1>CreateDynamicFunction ( _constructor_, _newTarget_, _kind_, _args_ )</h1>
...
             1. Let _sourceText_ be ! StringToCodePoints(_sourceString_).
+            1. Let _parameters_ be ParseText(! StringToCodePoints(_P_), _parameterSym_).
+            1. If _parameters_ is a List of errors, throw a *SyntaxError* exception.
+            1. Let _body_ be ParseText(! StringToCodePoints(_bodyString_), _bodySym_).
+            1. If _body_ is a List of errors, throw a *SyntaxError* exception.
             1. Let _expr_ be ParseText(_sourceText_, _exprSym_).
             1. If _expr_ is a List of errors, throw a *SyntaxError* exception.
-            1. Let _parameters_ be the |FormalParameters| of _expr_.
-            1. Let _body_ be the child of _expr_ that is an instance of _bodySym_.
+            1. NOTE: the above two steps ensure that Early Error rules which apply to _exprSym_ directly are enforced.
```
ljharb, line 25003: this note references "above two steps" which references steps 2 and 3, but it seems like it should be referencing steps 1 and 2? (ie, move this note 2 lines up?)

bakkot: The two steps immediately preceding this NOTE are

```
1. Let _expr_ be ParseText(_sourceText_, _exprSym_).
1. If _expr_ is a List of errors, throw a *SyntaxError* exception.
```

which are indeed the steps I intended to reference. Why does it seem like it should be referencing other steps?

ljharb: ahh, i misunderstood, this change makes it discard `expr` since `body` is produced above.

erights:
I am not a reviewer, but FWIW, LGTM. Thanks much!

syg (commented):
I understand where @erights is coming from in #2373, but I find this piecemeal parsing, then re-parsing the entire text only for the Early Error application to be misleading. It's misleading in that it suggests that parsing _expr_ using _exprSym_ alone is somehow insufficient to do all the validation while it in fact is.

I'd prefer to keep to a single parse and have a NOTE that says the syntactic constraint that @erights wants highlighted is applied post-concatenation. Then again, I don't find that state of affairs particularly surprising or indirect as he does.

bakkot:
I think this might actually be normatively necessary - that is, I think #2348 might accidentally not have been editorial after all.

Consider `new Function('/*', '*/ {')`. This shouldn't be legal, I believe that #2348 accidentally made it so.

erights:
@bakkot Nice example! I tried to come up with one but did not. Thanks!

syg:
> Consider new Function('/*', '*/ ) {'). This shouldn't be legal, I believe that #2348 accidentally made it so.

Ah ha, I retract my above comment then.

syg (approved):
lgtm with comment

spec.html
```diff
@@ -24990,10 +24994,13 @@ <h1>CreateDynamicFunction ( _constructor_, _newTarget_, _kind_, _args_ )</h1>
...
             1. Let _sourceText_ be ! StringToCodePoints(_sourceString_).
+            1. Let _parameters_ be ParseText(! StringToCodePoints(_P_), _parameterSym_).
+            1. If _parameters_ is a List of errors, throw a *SyntaxError* exception.
+            1. Let _body_ be ParseText(! StringToCodePoints(_bodyString_), _bodySym_).
+            1. If _body_ is a List of errors, throw a *SyntaxError* exception.
             1. Let _expr_ be ParseText(_sourceText_, _exprSym_).
             1. If _expr_ is a List of errors, throw a *SyntaxError* exception.
-            1. Let _parameters_ be the |FormalParameters| of _expr_.
-            1. Let _body_ be the child of _expr_ that is an instance of _bodySym_.
+            1. NOTE: The above two steps ensure that Early Error rules which apply to _exprSym_ directly are enforced.
```
syg, line 25003: I'd like this note to be even more explicit about a reparse being unnecessary.

michaelficarra (approved):

spec.html
bakkot: Rephrased to

>             1. NOTE: If the above two steps are reached, _sourceText_ is guaranteed to match _exprSym_. The purpose of these steps is to enforce any Early Error rules which apply to _exprSym_ directly.

Does that work for you?

syg: Works for me, thanks.

erights:
I like it. But your example is worth a thousand such explanations since most of us didn't see how the original text was erroneous even after we started looking for a counter-example. Your example is tiny and clarifying. It would be a good addition.

bakkot:
Added my above example as another NOTE.

spec.html
```diff
@@ -24990,10 +24994,14 @@ <h1>CreateDynamicFunction ( _constructor_, _newTarget_, _kind_, _args_ )</h1>
...
+            1. Let _parameters_ be ParseText(! StringToCodePoints(_P_), _parameterSym_).
+            1. If _parameters_ is a List of errors, throw a *SyntaxError* exception.
+            1. Let _body_ be ParseText(! StringToCodePoints(_bodyString_), _bodySym_).
+            1. If _body_ is a List of errors, throw a *SyntaxError* exception.
+            1. NOTE: The parameters and body are parsed separately to ensure that each is valid alone. For example, `new Function("/*", "*/ ) {")` is not legal.
             1. Let _expr_ be ParseText(_sourceText_, _exprSym_).
             1. If _expr_ is a List of errors, throw a *SyntaxError* exception.
-            1. Let _parameters_ be the |FormalParameters| of _expr_.
-            1. Let _body_ be the child of _expr_ that is an instance of _bodySym_.
+            1. NOTE: If the above two steps are reached, _sourceText_ must match _exprSym_ (although the reverse implication does not hold). The purpose of these steps is to enforce any Early Error rules which apply to _exprSym_ directly.
```
michaelficarra, line 25004:
```suggestion
            1. NOTE: If this step is reached, _sourceText_ must match _exprSym_ (although the reverse implication does not hold). The purpose of the next two steps is to enforce any Early Error rules which apply to _exprSym_ directly.
```

and move it up two steps.

bakkot: Done.

spec.html
```diff
@@ -24990,10 +24994,14 @@ <h1>CreateDynamicFunction ( _constructor_, _newTarget_, _kind_, _args_ )</h1>
             1. Let _prefix_ be the prefix associated with _kind_ in <emu-xref href="#table-dynamic-function-sourcetext-prefixes"></emu-xref>.
             1. Let _sourceString_ be the string-concatenation of _prefix_, *" anonymous("*, _P_, 0x000A (LINE FEED), *") {"*, _bodyString_, and *"}"*.
             1. Let _sourceText_ be ! StringToCodePoints(_sourceString_).
+            1. Let _parameters_ be ParseText(! StringToCodePoints(_P_), _parameterSym_).
+            1. If _parameters_ is a List of errors, throw a *SyntaxError* exception.
+            1. Let _body_ be ParseText(! StringToCodePoints(_bodyString_), _bodySym_).
+            1. If _body_ is a List of errors, throw a *SyntaxError* exception.
+            1. NOTE: The parameters and body are parsed separately to ensure that each is valid alone. For example, `new Function("/*", "*/ ) {")` is not legal.
```
michaelficarra, line 25001: This note doesn't need to be inline in the algorithm steps. Probably better as an `<emu-note>`.

bakkot: Eh, I mildly prefer it here, but I'm willing to move it if you feel strongly or @syg agrees.

PR #2379, opened 2021-04-09
Editorial: Explicitly define default value of 'internalSlotsList' in OrdinaryCreateFromConstructor

h2oche:
In [10.1.13 OrdinaryCreateFromConstructor](https://tc39.es/ecma262/#sec-ordinarycreatefromconstructor), it might be better to explicitly define default value of `internalSlotsList` when it is not provided.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -10569,6 +10569,7 @@
       <emu-alg>
         1. Assert: _intrinsicDefaultProto_ is a String value that is this specification's name of an intrinsic object. The corresponding object must be an intrinsic that is intended to be used as the [[Prototype]] value of an object.
         1. Let _proto_ be ? GetPrototypeFromConstructor(_constructor_, _intrinsicDefaultProto_).
+        1. If _internalSlotsList_ is not present, set _internalSlotsList_ to a new empty List.
         1. Return ! OrdinaryObjectCreate(_proto_, _internalSlotsList_).
       </emu-alg>
     </emu-clause>


spec.html
```diff
@@ -10569,6 +10569,7 @@ <h1>OrdinaryCreateFromConstructor ( _constructor_, _intrinsicDefaultProto_ [ , _
       <emu-alg>
         1. Assert: _intrinsicDefaultProto_ is a String value that is this specification's name of an intrinsic object. The corresponding object must be an intrinsic that is intended to be used as the [[Prototype]] value of an object.
         1. Let _proto_ be ? GetPrototypeFromConstructor(_constructor_, _intrinsicDefaultProto_).
+        1. If _internalSlotsList_ is not present, set _internalSlotsList_ to a new empty List.
```
ljharb, line 10572: The prose for this AO already says “ If internalSlotsList is not provided, a new empty List is used.”.

bakkot: While that's true, I think it's nicer to have it in the algorithm steps as well. There's a few other places we do this, like in [Abstract Relational Comparison](https://github.com/tc39/ecma262/pull/2378#discussion_r610333389), and I think we should change all of them.

ljharb: In that case, should it be removed from the prose?

bakkot: I think it's fine to keep it in the prose as well - it's an accurate and potentially useful description of the behavior of the algorithm.

ljharb: would you expect, then, that it'd be duplicated in prose and an algorithm step in every case where there's a default value on an optional AO argument?

bakkot: No, the preambles are just general descriptions of the behavior; they don't need to all give exactly the same level of details.

linusg:
This is now:

> 3. If internalSlotsList is present, let slotsList be internalSlotsList.
> 4. Else, let slotsList be a new empty List.
> 5. Return OrdinaryObjectCreate(proto, slotsList).

PR #2380, opened 2021-04-09
Editorial: More info in 'PromiseCapability Record' table

h2oche:
In [27.2.1.1 PromiseCapability Records](https://tc39.es/ecma262/#sec-promisecapability-records), all properties of `PromiseCapability Record` are initialized to `undefined` in `NewPromiseCapability` operation. It might be better to explicitly describe them in table.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37573,10 +37573,10 @@ THH:mm:ss.sss
                 [[Promise]]
               </td>
               <td>
-                An object
+                An object | *undefined*
               </td>
               <td>
-                An object that is usable as a promise.
+                An object that is usable as a promise. *undefined* if not yet assigned.
               </td>
             </tr>
             <tr>
@@ -37584,10 +37584,10 @@ THH:mm:ss.sss
                 [[Resolve]]
               </td>
               <td>
-                A function object
+                A function object | *undefined*
               </td>
               <td>
-                The function that is used to resolve the given promise object.
+                The function that is used to resolve the given promise object. *undefined* if not yet assigned.
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


ljharb:
Isn’t this true of every single internal slot? Why call it out specifically on these?

bakkot:
[NewPromiseCapability](https://tc39.es/ecma262/#sec-newpromisecapability) is the thing which is used to create PromiseCapability Records, and it ensures the `[[Resolve]]` and `[[Reject]]` slots are not undefined* I don't think this table needs to call out that the description does not apply during that algorithm.

h2oche:
Actually, we are developing type analysis tool for ECMAScript. While building our type model, we didn't notice that those properties can be `undefined` since we built that based on table description. We thought that more informative table may be helpful to someone who are interested in ECMAScript internal.
(Anyway, we agree with all your opinions)

doehyunbaek:
@bakkot I think we can close this PR with the merge of #2980 .

By the way, is there some policy change with regards to transient field values? With #2980 and changes like [this](https://github.com/tc39/ecma262/pull/2547/files#diff-181371b08d71216599b0acccbaabd03c306da6de142ea6275c2135810999805aL11603-R11641)(it locates when you clicks load diff), it seems transient field values are to be discouraged, but not as much to include them in the fields table, like the one for [realm records](https://tc39.es/ecma262/#table-realm-record-fields). Is it right?

If so, what is the recommended way for the current transient assignment to `realmRec.[[GlobalEnv]]` in step 4 of [CreateRealm](https://tc39.es/ecma262/2023/#sec-createrealm)? I guess refactor as in #2980?

PR #2397, opened 2021-04-29
Editorial: describe behavior for algorithms without a "Return"

### LINKED ISSUE #2394: "Return" at end of algorithm: implicit or explicit?
Currently, there are some algorithms where, if an 'early' return doesn't occur, control just falls off the end of the steps. The spec doesn't define what this means. Should we define what it means (implicit `Return` step), or add an explicit `Return` to such cases?

bakkot:
Fixes #2394.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -824,12 +824,7 @@
           1. Assert: _completionRecord_ is a Completion Record.
           1. Return _completionRecord_ as the Completion Record of this abstract operation.
         </emu-alg>
-        <p>A &ldquo;return&rdquo; statement without a value in an algorithm step means the same thing as:</p>
-        <emu-alg>
-          1. Return NormalCompletion(~unused~).
-        </emu-alg>
-        <p>In the above step, ~unused~ is a special value which is never directly consumed.</p>
-        <p>If all steps in an abstract operation have been executed without returning, the abstract operation behaves as though there were an additional step reading "Return." after the existing steps.</p>
+        <p>If all steps in an abstract operation have been executed without returning, the abstract operation behaves as though there were an additional step reading "Return NormalCompletion(~unused~)." after the existing steps. ~unused~ is a special value which is never directly consumed.</p>
         <p>Any reference to a Completion Record value that is in a context that does not explicitly require a complete Completion Record value is equivalent to an explicit reference to the [[Value]] field of the Completion Record value unless the Completion Record is an abrupt completion.</p>
       </emu-clause>
 
@@ -3574,7 +3569,7 @@
           1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
           1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
-          1. Return.
+          1. Return NormalCompletion(~unused~).
           1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
         </emu-alg>
 
@@ -39482,7 +39477,7 @@ THH:mm:ss.sss
             1. Else,
               1. Assert: _result_.[[Type]] is ~throw~.
               1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _result_.[[Value]] &raquo;).
-            1. [id="step-asyncfunctionstart-return-undefined"] Return.
+            1. [id="step-asyncfunctionstart-return-undefined"] Return NormalCompletion(~unused~).
           1. Push _asyncContext_ onto the execution context stack; _asyncContext_ is now the running execution context.
           1. Resume the suspended evaluation of _asyncContext_. Let _result_ be the value returned by the resumed computation.
           1. Assert: When we return here, _asyncContext_ has already been removed from the execution context stack and _runningContext_ is the currently running execution context.
--- a/spec.html
+++ b/spec.html
@@ -826,8 +826,10 @@
         </emu-alg>
         <p>A &ldquo;return&rdquo; statement without a value in an algorithm step means the same thing as:</p>
         <emu-alg>
-          1. Return NormalCompletion(*undefined*).
+          1. Return NormalCompletion(~unused~).
         </emu-alg>
+        <p>In the above step, ~unused~ is a special value which is never directly consumed.</p>
+        <p>If all steps in an abstract operation have been executed without returning, the abstract operation behaves as though there were an additional step reading "Return." after the existing steps.</p>
         <p>Any reference to a Completion Record value that is in a context that does not explicitly require a complete Completion Record value is equivalent to an explicit reference to the [[Value]] field of the Completion Record value unless the Completion Record is an abrupt completion.</p>
       </emu-clause>
 
@@ -3525,7 +3527,7 @@
               [[Value]]
             </td>
             <td>
-              any ECMAScript language value or ~empty~
+              any ECMAScript language value, ~empty~, or ~unused~
             </td>
             <td>
               The value that was produced.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -826,8 +826,10 @@ <h1>Implicit Completion Values</h1>
         </emu-alg>
         <p>A &ldquo;return&rdquo; statement without a value in an algorithm step means the same thing as:</p>
```
michaelficarra, line 827: I only see 2 instances of `Return.` in the spec. I'd rather eliminate this rule and replace them with `Return NormalCompletion(~unused~).`

bakkot: Sure, done.

spec.html
```diff
@@ -3574,7 +3569,7 @@ <h1>Await</h1>
           1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
           1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
-          1. Return.
+          1. Return NormalCompletion(~unused~).
```
ljharb, line 3572: why does this one need an explicit return statement at all?

spec.html
```diff
@@ -39482,7 +39477,7 @@ <h1>AsyncFunctionStart ( _promiseCapability_, _asyncFunctionBody_ )</h1>
             1. Else,
               1. Assert: _result_.[[Type]] is ~throw~.
               1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _result_.[[Value]] &raquo;).
-            1. [id="step-asyncfunctionstart-return-undefined"] Return.
+            1. [id="step-asyncfunctionstart-return-undefined"] Return NormalCompletion(~unused~).
           1. Push _asyncContext_ onto the execution context stack; _asyncContext_ is now the running execution context.
           1. Resume the suspended evaluation of _asyncContext_. Let _result_ be the value returned by the resumed computation.
```
ljharb, line 39482: is this correct, given that 2 lines down, the returned value *is* used?

spec.html
bakkot: It's not strictly necessary, but I think it's helpful so that the "this" in the immediately following note is clear.

spec.html
bakkot: I would not regard the assertion as using the value.

ljharb: additionally, the step ID says "return undefined" - is it not important that it's "undefined" versus an unused value?

bakkot: No, that's just an arbitrary label for making the later step reference work. I didn't change it because step IDs don't yet support `oldids` and I didn't want to break existing links.

bakkot:
I note that this races with two new usages of `Return.` in https://github.com/tc39/ecma262/pull/1668, and whichever lands second will need to be rebased. I'm happy to do that myself, I'm just commenting so we don't lose track.

michaelficarra:
Thinking about this more, we could instead use a new kind of completion record here (just for procedure early exits) that doesn't have a `[[Value]]` slot, making the attempt to use the value of an AO that returns it an editorial type error through existing means.

bakkot:
I would prefer not to further complicate the distinction between normal and abrupt completions, I think.

michaelficarra:
Yeah, I think I agree. We might be able to make some improvements in this area once completion value reform is complete.

spec.html
jmdyck: Because the subsequent NOTE-step needs something to refer to?

ljharb: I'm sure the note could be reworded to refer to the implicit return behavior without the need for a placeholder step, if that were desired.

bakkot: > Because the subsequent NOTE-step needs something to refer to?

Yes. It could probably be rephrased, but I like this wording.

spec.html
```diff
@@ -824,10 +824,7 @@ <h1>Implicit Completion Values</h1>
           1. Assert: _completionRecord_ is a Completion Record.
           1. Return _completionRecord_ as the Completion Record of this abstract operation.
         </emu-alg>
-        <p>A &ldquo;return&rdquo; statement without a value in an algorithm step means the same thing as:</p>
-        <emu-alg>
-          1. Return NormalCompletion(*undefined*).
-        </emu-alg>
+        <p>If all steps in an abstract operation have been executed without returning, the abstract operation behaves as though there were an additional step reading "Return NormalCompletion(~unused~)." after the existing steps. ~unused~ is a special value which is never directly consumed.</p>
```
jmdyck, line 827: "If all steps in an abstract operation have been executed without returning..."

Well, not necessarily *all* steps, because if there are If-else-steps, we might have executed only one of the arms. Maybe "... completes execution without returning ..." ?

Also, it's not just abstract operations, there's at least one syntax-directed operation (ForDeclarationBindingInstantiation) that this applies to. (Or maybe you consider SDOs a subset of AOs, which the spec isn't entirely clear on.)

---

"`~unused~` is a special value which is never directly consumed."

So it might be *indirectly* consumed? What does that even mean?

Maybe "`~unused~` is an ad hoc value that is only used for this purpose."

---

Also maybe "Operations that return `~unused~` are typically invoked in a `Perform` step." (I only found one exception, where `AsyncGeneratorEnqueue` has `Let _check_ be AsyncGeneratorValidate(...).`)

ljharb: I'd say "Operations that return `~unused~` must only be invoked in a `Perform` step.`

jmdyck: So then what about the `AsyncGeneratorEnqueue` case?

bakkot: @jmdyck

> Well, not necessarily all steps, because if there are If-else-steps, we might have executed only one of the arms.

I was modeling the if-else as a single step (with the arms being substeps), but I can see how your interpretation is plausible.

> Maybe "... completes execution without returning ..." ?

I'd be OK with that, though I don't love it. Open to other phrasings here.

I originally said something like "if control reaches the end of the algorithm", but I felt that was maybe too confusing, since we normally talk about control-flow of the JS being executed rather than of spec steps.

> Also, it's not just abstract operations, there's at least one syntax-directed operation (ForDeclarationBindingInstantiation) that this applies to. (Or maybe you consider SDOs a subset of AOs, which the spec isn't entirely clear on.)

For avoidance of doubt I'll list SDOs explicitly.

> So it might be _indirectly_ consumed? What does that even mean?

It means the completion containing it is consumed, or it's stored in a binding within a macro but the binding is not used after the macro, etc. I am content with this phrasing personally; does anyone else want to weigh in?

@ljharb 
> Operations that return `~unused~` must only be invoked in a `Perform` step

That's not a guarantee I want to make: it is reasonable to have instances of the pattern

```
1. Let _result_ be SomeProcedure(whatever).
1. If _result_ is an abrupt completion, then
  1. Do some cleanup. 
```

ljharb: Hmm, that's a good question. I don't have a suggestion.

jmdyck: > I was modeling the if-else as a single step (with the arms being substeps),

Indeed, that's how structured programming wants us to think of it, but when each `Else` has a different step-number, it's harder to suggest that they're not different steps. 

> > Maybe "... completes execution without returning ..." ?
> 
> I'd be OK with that, though I don't love it.

Yeah, I'm not thrilled with it either. The problem is, the existing spec gives us almost no terminology/model for the execution of algorithms. 

> I originally said something like "if control reaches the end of the algorithm", but I felt that was maybe too confusing, since we normally talk about control-flow of the JS being executed rather than of spec steps.

Yeah, I don't see any examples of the spec using "control" in the context of performing spec steps. At least for "execute" there are a handful of examples.

radar210love007 (approved):

syg (approved):

michaelficarra (approved):

bakkot:
This should wait on https://github.com/tc39/ecma262/pull/2413, which touches several of the same places.

jmdyck:
I think, once PR #2547 lands, you won't need this PR any more, because there won't be any algorithms that complete without a Return.

Or I suppose you could say that if it happens it's an editorial error.

jmdyck:
If you rebase to main (now that #2547 has landed), I think you'll find that there isn't much left in this PR.

Note that [9.10.3 Execution](https://tc39.es/ecma262/#sec-weakref-execution) doesn't have a "Return" step, so you might want to add a `Return ~unused~`. On the other hand, it's a weird case, so maybe "Return" doesn't mean anything?

bakkot:
Fixed by #2547 instead.

Issue #2399, opened 2021-05-02
Ambiguous namespace export access returns undefined


guybedford:
Currently ambiguous export bindings that have no other importer but that are available on a module namespace will always return `undefined`.

Example scenario from https://github.com/rollup/rollup/issues/4049#issuecomment-830745772:

```js
// index.js
import * as ns from './reexport.js';
console.log(ns.foo, ns.bar, ns.baz);

// reexport.js
export * from './first.js';
export * from './second.js';

// first.js
export const foo = 1;
export const bar = 1;

// second.js
export const foo = 2;
export const baz = 2;
```

Output:

```
undefined 1 2
```

The `undefined` here seems to be weakly specified in the modules specification per https://tc39.es/ecma262/#sec-module-namespace-exotic-objects-get-p-receiver, where `[[BindingName]]` `"*ambiguous*"` is not explicitly handled in the steps but implicitly returns undefined when passed into `targetEnv.[[GetBindingValue]]`.

The scenario above directly corresponds to the last point mentioned in the spec note in https://tc39.es/ecma262/#sec-getmodulenamespace:

> The only way GetModuleNamespace can throw is via one of the triggered HostResolveImportedModule calls. Unresolvable names are simply excluded from the namespace at this point. They will lead to a real linking error later unless they are all ambiguous star exports that are not explicitly requested anywhere.

Because all of the ambiguous exports have no other direct importer, and only a namespace representation, the ambiguous error is not thrown during initial linking.

Ideally this namespace `Get` undefined return should be explicitly specified here. Alternatively an ambiguous error could be thrown on access, although it's unclear whether that would be web compatible at this point.

guybedford:
I've posted an editorial PR in https://github.com/tc39/ecma262/pull/2401 to more clearly define this behaviour.

guybedford:
Closing per https://github.com/tc39/ecma262/pull/2401#issuecomment-831516529. Apologies for noise. The good news at least is that RollupJS is now fully compatible with these behaviours down to some warnings provided during externalization where ambiguity can't be known ahead of time.

PR #2401, opened 2021-05-03
Editorial: Explicitly define ambiguous namespace properties as undefined

guybedford:
This ensures that the spec is explicit about ambiguous bindings on module namespaces that have not otherwise been imported in a way that would throw an error will always be `undefined` on the namespace, effectively skipping the unnecessary call to `_targetEnv_.getBindingValue`.

See issue https://github.com/tc39/ecma262/issues/2399 for the exact details and scenario to which this applies.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12039,6 +12039,8 @@
           1. Assert: _targetModule_ is not *undefined*.
           1. If _binding_.[[BindingName]] is *"\*namespace\*"*, then
             1. Return ? GetModuleNamespace(_targetModule_).
+          1. If _binding_.[[BindingName]] is *"\*ambiguous\*"*, then
+            1. Return *undefined*.
           1. Let _targetEnv_ be _targetModule_.[[Environment]].
           1. If _targetEnv_ is *undefined*, throw a *ReferenceError* exception.
           1. Return ? _targetEnv_.GetBindingValue(_binding_.[[BindingName]], *true*).


spec.html
```diff
@@ -12039,6 +12039,8 @@ <h1>[[Get]] ( _P_, _Receiver_ )</h1>
           1. Assert: _targetModule_ is not *undefined*.
           1. If _binding_.[[BindingName]] is *"\*namespace\*"*, then
             1. Return ? GetModuleNamespace(_targetModule_).
+          1. If _binding_.[[BindingName]] is *"\*ambiguous\*"*, then
+            1. Return *undefined*.
```
ljharb, line 12043: would it be worth exploring the web compatibility of making this throw?

guybedford: It's difficult to say, but it likely is a web compat issue at this point I'd guess. It could still be worth discussing further though. Another concern about throwing would be even just logging a namespace to the console would be enough to cause the error.

anba:
There seems to be some kind of misunderstanding around ambiguous exports:
1. The string `"*ambiguous*"` is used nowhere within the spec.
2. The spec does use the string `"ambiguous"` (so without the asterisk).
3. `"ambiguous"` is a possible return value in [ResolveExport](https://tc39.es/ecma262/#resolvedbinding-record), but in that position it's never used as a [[BindingName]] value in a [ResolvedBinding record](https://tc39.es/ecma262/#resolvedbinding-record).
    - ResolveExport has three different return values:
      1. A ResolvedBinding record.
      2. `null`
      3. and `"ambiguous"`.
4. Module namespace objects return `undefined` for ambiguous exports due to steps 4-5:
> 4. Let exports be O.[[Exports]].
> 5. If P is not an element of exports, return undefined.

5. The [[Exports]] slot of module namespace exports is set in [ModuleNamespaceCreate](https://tc39.es/ecma262/#sec-modulenamespacecreate). This list doesn't contain any ambiguous exports by definition, cf. [GetModuleNamespace](https://tc39.es/ecma262/#sec-getmodulenamespace) where the exports list is created.

guybedford:
@anba thanks for the very clear explanation here, you are entirely correct! Yes, I missed that these were being filtered off the namespace by the string value `"ambiguous"` not being a ResolveBinding record.

PR #2411, opened 2021-05-20
Editorial: Clarify RegExp grammar parameter U → UnicodeMode

mathiasbynens:
This makes the parameter more easily searchable while clarifying its meaning.

This came up while working on the draft spec for [the RegExp set notation proposal](https://github.com/tc39/proposal-regexp-set-notation), which introduces another RegExp grammar parameter that is related to Unicode. Rather than sticking to single letters, it seems nicer to move to more descriptive names. 

If the editors are open to this change, I’d be happy to work on a patch that renames `N` as well (perhaps to `WithNamedCaptureGroups`).

cc @markusicu @markusicu @sffc

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27789,7 +27789,7 @@
         The behaviour of built-in methods which are specified in ECMA-402, such as those named `toLocaleString`, must not be extended except as specified in ECMA-402.
       </li>
       <li>
-        The RegExp pattern grammars in <emu-xref href="#sec-patterns"></emu-xref> and <emu-xref href="#sec-regular-expressions-patterns"></emu-xref> must not be extended to recognize any of the source characters A-Z or a-z as |IdentityEscape[+U]| when the <sub>[U]</sub> grammar parameter is present.
+        The RegExp pattern grammars in <emu-xref href="#sec-patterns"></emu-xref> and <emu-xref href="#sec-regular-expressions-patterns"></emu-xref> must not be extended to recognize any of the source characters A-Z or a-z as |IdentityEscape[+UnicodeMode]| when the <sub>[UnicodeMode]</sub> grammar parameter is present.
       </li>
       <li>
         The Syntactic Grammar must not be extended in any manner that allows the token `:` to immediately follow source text that matches the |BindingIdentifier| nonterminal symbol.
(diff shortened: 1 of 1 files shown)


markusicu:
Bike-shedding opinion:
- Longer grammar parameters would be more searchable (as you say) and more guessable, e.g., `U` vs. the *Unicode* boolean in the semantics
- We *could* use the same terms for grammar parameters and semantics booleans (e.g., *Unicode*), but it might also be confusing.
- For the syntax grammar, very long parameter names could be tedious to read & write, especially where multiple parameters are used on a line.
- Maybe intermediate abbreviations for the grammar parameters? E.g., `U`→`Uni`, (proposed) `V`→`UniSet`, `N`→`Named`

mathiasbynens:
> * Longer grammar parameters would be more searchable (as you say) and more guessable, e.g., `U` vs. the _Unicode_ boolean in the semantics

+1

> * We _could_ use the same terms for grammar parameters and semantics booleans (e.g., _Unicode_), but it might also be confusing.

I think it would mostly be confusing, as it would imply a connection that isn't there.

> * For the syntax grammar, very long parameter names could be tedious to read & write, especially where multiple parameters are used on a line.
> * Maybe intermediate abbreviations for the grammar parameters? E.g., `U`→`Uni`, (proposed) `V`→`UniSet`, `N`→`Named`

That reduces searchability. Ctrl+F "Uni" would return lots of unrelated hits.

spec.html
```diff
@@ -41915,40 +41915,40 @@ <h2>Syntax</h2>
...
       <p>The syntax of <emu-xref href="#sec-patterns"></emu-xref> is modified and extended as follows. These changes introduce ambiguities that are broken by the ordering of grammar productions and by contextual information. When parsing using the following grammar, each alternative is considered only if previous production alternatives do not match.</p>
-      <p>This alternative pattern grammar and semantics only changes the syntax and semantics of BMP patterns. The following grammar extensions include productions parameterized with the [U] parameter. However, none of these extensions change the syntax of Unicode patterns recognized when parsing with the [U] parameter present on the goal symbol.</p>
+      <p>This alternative pattern grammar and semantics only changes the syntax and semantics of BMP patterns. The following grammar extensions include productions parameterized with the [UnicodeFlag] parameter. However, none of these extensions change the syntax of Unicode patterns recognized when parsing with the [UnicodeFlag] parameter present on the goal symbol.</p>
       <h2>Syntax</h2>
       <emu-grammar type="definition">
         Term[U, N] ::
```
jmdyck, line 41921:
```suggestion
        Term[UnicodeFlag, N] ::
```
There are 12 other occurrences of "U" in this `<emu-grammar>` that need to be converted.

spec.html
```diff
@@ -41992,17 +41992,17 @@ <h2>Syntax</h2>
...
+          CharacterClassEscape[?UnicodeFlag]
           CharacterEscape[?U, ?N]
 
         ClassControlLetter ::
           DecimalDigit
           `_`
       </emu-grammar>
       <emu-note>
-        <p>When the same left hand sides occurs with both [+U] and [\~U] guards it is to control the disambiguation priority.</p>
+        <p>When the same left hand sides occurs with both [+UnicodeFlag] and [\~U] guards it is to control the disambiguation priority.</p>
```
jmdyck, line 42005:
```suggestion
        <p>When the same left hand sides occurs with both [+UnicodeFlag] and [\~UnicodeFlag] guards it is to control the disambiguation priority.</p>
```

jmdyck (changes requested):

spec.html
mathiasbynens: Good catch; done!

spec.html
mathiasbynens: Your suggestion also made me notice "left hand sides" → "left-hand sides" which I've now fixed as well. Thanks!

jmdyck (commented):
Looks fine now.

michaelficarra:
I'd rather see us have a consistent editorial direction here, either going the verbose route for all grammar flags or the terse route for all grammar flags. We can discuss the tradeoffs of each in the editor call, but right now I don't think we should do this as a 1-off change.

mathiasbynens:
> We can discuss the tradeoffs of each in the editor call, but right now I don't think we should do this as a 1-off change.

Just to be clear, I agree with this! As I said:

> If the editors are open to this change, I’d be happy to work on a patch that renames `N` as well (perhaps to `WithNamedCaptureGroups`).

The reason I proposed doing it as two separate patches is because this one would (editorially) affect the RegExp set notation proposal whereas the `N` change wouldn’t. But I’d be happy to update this patch to change both at once if that's preferred.

michaelficarra:
We talked about this in the editor call. Longer term, we would like to link all of the grammar flags to a place that has descriptions. But for now, changing `U` to something more descriptive is fine. We just ask that flag names not include `Flag` in the name (even though we understand that here it is referring to the RegExp flags).

mathiasbynens:
> We talked about this in the editor call. Longer term, we would like to link all of the grammar flags to a place that has descriptions. But for now, changing `U` to something more descriptive is fine. We just ask that flag names not include `Flag` in the name (even though we understand that here it is referring to the RegExp flags).

Any suggested new name for `UnicodeFlag`? Perhaps `UnicodeMode`? It would help to know the exact rationale for avoiding "Flag" (because maybe "Mode" is a no-go for the same reason). For context, I'm trying to think of a name that's unique within the spec document, to improve searchability — so I’d rather not go with just `Unicode`.

michaelficarra:
`UnicodeMode` sounds good to me. Two reasons not to use "flag" in the name:

1. A reader is probably more likely to think the word refers to it being a grammar flag than a RegExp flag, and we certainly aren't going to add `Flag` to the end of every grammar flag.
2. The grammar flags should be referring to the mode that the grammar is in or the thing that is being controlled, not the cause for entering this mode. For example, the `Await` flag is called `Await` because it controls the handling of the `await` contextual keyword. It is not called `Async` even though it is introduced by the `async` keyword.

mathiasbynens:
I’ve gone ahead and updated the patch to use `UnicodeMode`. PTAL

> * A reader is probably more likely to think the word refers to it being a grammar flag than a RegExp flag, and we certainly aren't going to add `Flag` to the end of every grammar flag.

That's interesting. I never thought of these as "grammar flags". The spec refers to them as [grammar _parameters_](https://tc39.es/ecma262/#sec-notational-conventions).

michaelficarra (approved):

markusicu:
I suggest changing the PR title to say "UnicodeMode"

mathiasbynens:
Is anything blocking this? I'd be happy to resolve the conflicts if that moves things along.

michaelficarra:
@mathiasbynens That won't be necessary. We just need @bakkot or @syg to find some time to do a review.

bakkot (approved):
This LGTM, sorry for the delay. I've also rebased, so it should be good to land.

ljharb:
@bakkot it looks like the rebase ended up with the wrong commit - d03e65a instead of b025795ad9cd115847c584957a1d141c25f934b3.

bakkot:
Whoops, fixed.

Issue #2412, opened 2021-05-24
Incorrect `!` assertion in AsyncGeneratorAwaitReturn 


bakkot:
[AsyncGenerator.prototype.return](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgenerator-prototype-return) step 8.b., and [AsyncGeneratorDrainQueue](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgeneratordrainqueue) step 5.c.ii, are `Perform ! AsyncGeneratorAwaitReturn(generator).`

But [AsyncGeneratorAwaitReturn](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgeneratorawaitreturn) can throw: step 6 is `Let promise be ? PromiseResolve(%Promise%, completion.[[Value]]).`. And that `?` is correct: [PromiseResolve](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise-resolve) 2.a is `a. Let xConstructor be ? Get(x, "constructor").`, where `x` is a user-supplied value.

Concretely, this is observable in, for example, the following program:

```js
let gen = (async function*f(){ })();
let brokenPromise = Promise.resolve(42);
Object.defineProperty(brokenPromise, 'constructor', { get: function(){ console.log('called'); throw new Error('oh no'); } });
gen.return(brokenPromise);
```

Tracing through the execution of the above snippet:

- On line 1 we create an AsyncGenerator object and [set its state to `suspendedStart`
](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgeneratorstart)
- On line 4 we do [`AsyncGenerator.prototype.return`](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgenerator-prototype-return),
  - which [enqueues](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgeneratorenqueue)  a `return` completion for the generator, and then
  - since the state is `suspendedStart`, performs `! AsyncGeneratorAwaitReturn`, which
  - since the completion is a return completion, does `Let promise be ? PromiseResolve(%Promise%, completion.[[Value]]).`

I'm honestly not sure what the intent is in this case. In this simple example you could in principle throw when you call `return`, but it gets more complicated if you call `.return` while the generator is `awaiting`, as in

```js
let gen = (async function*f(){ await 0; })();
gen.next();
[...]

Issue #2409, opened 2021-05-17
Clarify Running Execution context and the impact of push/pop and suspend/resume


codehag:
In preparing top level await, something I thought was resolved kept nagging me and I have been trying to get further input. 

Specifically, in the top level await specification, we are suspending the running context for both the sync and async modules. In the case of async, that suspended module is still at the top of the stack:

Module suspension: https://github.com/tc39/proposal-top-level-await/blob/main/spec.html#L1208
Reference to running context in the asyncBlockStart: https://github.com/tc39/proposal-top-level-await/blob/main/spec.html#L52

The question is, if the top module on the stack is _suspended_ is it still the running context? If it is still the running context, we should clarify this here: https://tc39.es/ecma262/#running-execution-context -- my expectation that, when we suspend, that the running context will be the next one on the stack. I am also wondering if we really need this step, or if pushing a new context for the module is enough in the sync case.

On the other hand, if a _suspended_ module is **not** the running context when it is at the top of the context stack, then we have a bug in the spec code, as we will end up asserting that we are pointing to the wrong one here: https://github.com/tc39/proposal-top-level-await/blob/main/spec.html#L67

In any case, a clarification of push/pop and suspend/resume would help.

jmdyck:
> The question is, if the top module on the stack is _suspended_ is it still the running context?

The running execution context (REC) is just whatever context is top-of-stack. "Suspend" isn't defined, but it's pretty clear that it doesn't manipulate the stack (since that always has to be done separately). So I'd say theoretically yes, if you Suspend the top-of-stack, it's still the REC. However, 'in practice', it only stays that way for a step or two before it's no longer top-of-stack, during which time there aren't any references to the REC, so the answer currently doesn't make a difference normatively.

> In any case, a clarification of push/pop and suspend/resume would help, especially if suspending/resuming is in terms of push/pop. Does suspending mean, save the context, remove it from the stack, and return it when resuming?

Nope. Stack manipulation always happens explicitly (I think). "Suspend" (whatever it means) is separate, though it typically precedes a Push or Pop. (The cases where it doesn't might be bugs, it's hard to say.)

My guess is that the intent was that a Suspend would only affect a context's 'code evaluation state', but it's a grey area.

> This seems to be hinted at by:
> 
> > [...] Transition of the running execution context status among execution contexts usually occurs in stack-like last-in/first-out manner. However, some ECMAScript features require non-LIFO transitions of the running execution context.
> 
[...]

Issue #2400, opened 2021-05-03
Editorial: Notes that tell you where a `Return` step returns to


jmdyck:
Normally, when an operation executes a `Return` step, it transfers control back to the operation that invoked it. In fact, this is so normal that the spec doesn't even bother saying it.

So it's pretty weird that when the operation [GeneratorYield](https://tc39.es/ecma262/#sec-generatoryield) executes the `Return` at step 9, it *doesn't* return to its caller.

What makes it even weirder is that the only thing in the spec that *tells* you that it doesn't return to its caller is a Note (step 10), but Notes are supposed to be non-normative. If you remove the operation's Notes, there's nothing to tell you that step 9 *isn't* a garden-variety `Return`.

(As for the `Return` at 8.a, it's harder to say what the normal interpretation would be, because it's less clear what would constitute its "caller" if/when it executes, though I'd say that the GeneratorResume that will have just resumed it would have a better claim than whatever originally called GeneratorYield. So again, the Note at 8.b is necessary to convey the intended semantics of the `Return` step.)

The spec has 2 other places where a Note tells you where a `Return` returns to:
- [`Await`](https://tc39.es/ecma262/#await)
- [`AsyncGeneratorYield`](https://tc39.es/ecma262/#sec-asyncgeneratoryield)

Given that Notes should be non-normative, what are we to make of these cases?

----

For brevity, ORT = "operation return-target", i.e. where a `Return`-step in an abstract operation will return to.

In order for the `Return` steps to behave as the Notes indicate, the assumed mental model appears to be this:

If algorithm _A_ invokes operation _B_, and _B_ has a step of the form:
```
Remove _context_ from the execution context stack
 and restore the execution context that is at the top of the execution context stack
 as the running execution context.
```
then when _B_ executes that step, this causes _B_ to modify its ORT. Specifically, _B_'s ORT changes from:
- _A_, the algorithm that invoked _B_
to:
- _R_, the operation that had "most previously" (which I read as "most recently") resumed evaluation of _context_. I believe this is also the operation that was executing when the (newly) running execution context was last "running".

(E.g., when _B_ is `GeneratorYield`, _A_ might be a definition of the `Evaluation` SDO for a _YieldExpression_, and _R_ is `GeneratorResume`.)

Roughly speaking, the assumption appears to be that the ORT is an attribute of each execution context, so when you change the running execution context, you change where the current operation will return to.

(1)
[...]

PR #2413, opened 2021-05-24
Editorial: Clarify the AsyncGenerator state machine

bakkot:
AsyncGenerators are complicated, but at the root there's a relatively straightforward state machine. I find that it's difficult to see the shape of that machine with the current spec text, which routes every operation through [AsyncGeneratorResumeNext](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgeneratorresumenext).

I've rewritten all the logic here. The main reason I prefer this text is that each AO does one thing and returns, instead of kicking off a chain of self-recursive AOs which you have to follow for quite a while, keeping track of the state, before you figure out what's actually going to happen. For example, I think this PR makes it much more obvious that AsyncGeneratorYield might not return control to the rest of the program, and under what circumstances that happens. As a bonus, it makes AsyncGeneratorYield follow the pattern described in [this comment](https://github.com/tc39/ecma262/issues/2400#issuecomment-832335872), hopefully providing a path towards [clarifying execution context juggling](https://github.com/tc39/ecma262/issues/2409).

I don't believe there's any normative changes ~except for possibly, in certain cases, to the realm used for creating iterator result objects~. (Edit: I've tweaked things to ensure it preserves the original, fairly strange behavior. I think we may want to change that behavior, but not as part of this PR.)

This PR has an incorrect `!` assertion in callers of AsyncGeneratorAwaitReturn, corresponding to the same assertion in callers of AsyncGeneratorResumeNext described in https://github.com/tc39/ecma262/issues/2412. I'll fix this PR up if that issue is resolved first. Otherwise it just leaves things in the same state.

Rather than trying to confirm that the logic is the same by comparing the two texts, I strongly suggest you read [this gist](https://gist.github.com/bakkot/82186e27a2e78c6e507c56dd49d8a197) where I've described the state machine, and confirm that both the original spec text and the text in this PR correspond to that state machine.

cc @jmdyck, I know you've spent a long time puzzling over this as well.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -39858,8 +39858,21 @@ THH:mm:ss.sss
         <h1>AsyncGenerator.prototype.next ( _value_ )</h1>
         <emu-alg>
           1. Let _generator_ be the *this* value.
+          1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+          1. Let _result_ be AsyncGeneratorValidate(_generator_, ~empty~).
+          1. IfAbruptRejectPromise(_result_, _promiseCapability_).
+          1. Let _state_ be _generator_.[[AsyncGeneratorState]].
+          1. If _state_ is ~completed~, then
+            1. Let _iteratorResult_ be ! CreateIterResultObject(*undefined*, *true*).
+            1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, &laquo; _iteratorResult_ &raquo;).
+            1. Return _promiseCapability_.[[Promise]].
           1. Let _completion_ be NormalCompletion(_value_).
-          1. Return ! AsyncGeneratorEnqueue(_generator_, _completion_, ~empty~).
+          1. Perform ! AsyncGeneratorEnqueue(_generator_, _completion_, _promiseCapability_).
+          1. If _state_ is either ~suspendedStart~ or ~suspendedYield~, then
+            1. Perform ! AsyncGeneratorResume(_generator_, _completion_).
+          1. Else,
+            1. Assert: _state_ is either ~executing~ or ~awaiting-return~.
+          1. Return _promiseCapability_.[[Promise]].
         </emu-alg>
       </emu-clause>
 
@@ -39867,8 +39880,20 @@ THH:mm:ss.sss
         <h1>AsyncGenerator.prototype.return ( _value_ )</h1>
         <emu-alg>
           1. Let _generator_ be the *this* value.
+          1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+          1. Let _result_ be AsyncGeneratorValidate(_generator_, ~empty~).
+          1. IfAbruptRejectPromise(_result_, _promiseCapability_).
           1. Let _completion_ be Completion { [[Type]]: ~return~, [[Value]]: _value_, [[Target]]: ~empty~ }.
-          1. Return ! AsyncGeneratorEnqueue(_generator_, _completion_, ~empty~).
+          1. Perform ! AsyncGeneratorEnqueue(_generator_, _completion_, _promiseCapability_).
+          1. Let _state_ be _generator_.[[AsyncGeneratorState]].
+          1. If _state_ is either ~suspendedStart~ or ~completed~, then
+            1. Set _generator_.[[AsyncGeneratorState]] to ~awaiting-return~.
+            1. Perform ! AsyncGeneratorAwaitReturn(_generator_).
+          1. Else if _state_ is ~suspendedYield~, then
+            1. Perform ! AsyncGeneratorResume(_generator_, _completion_).
+          1. Else,
+            1. Assert: _state_ is either ~executing~ or ~awaiting-return~.
+          1. Return _promiseCapability_.[[Promise]].
         </emu-alg>
       </emu-clause>
 
@@ -39876,8 +39901,23 @@ THH:mm:ss.sss
         <h1>AsyncGenerator.prototype.throw ( _exception_ )</h1>
         <emu-alg>
           1. Let _generator_ be the *this* value.
+          1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+          1. Let _result_ be AsyncGeneratorValidate(_generator_, ~empty~).
+          1. IfAbruptRejectPromise(_result_, _promiseCapability_).
+          1. Let _state_ be _generator_.[[AsyncGeneratorState]].
+          1. If _state_ is ~suspendedStart~, then
+            1. Set _generator_.[[AsyncGeneratorState]] to ~completed~.
+            1. Set _state_ to ~completed~.
+          1. If _state_ is ~completed~, then
+            1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _exception_ &raquo;).
+            1. Return _promiseCapability_.[[Promise]].
           1. Let _completion_ be ThrowCompletion(_exception_).
-          1. Return ! AsyncGeneratorEnqueue(_generator_, _completion_, ~empty~).
+          1. Perform ! AsyncGeneratorEnqueue(_generator_, _completion_, _promiseCapability_).
+          1. If _state_ is ~suspendedYield~, then
+            1. Perform ! AsyncGeneratorResume(_generator_, _completion_).
+          1. Else,
+            1. Assert: _state_ is either ~executing~ or ~awaiting-return~.
+          1. Return _promiseCapability_.[[Promise]].
         </emu-alg>
       </emu-clause>
 
@@ -39908,7 +39948,7 @@ THH:mm:ss.sss
           </tr>
           <tr>
             <td>[[AsyncGeneratorQueue]]</td>
-            <td>A List of AsyncGeneratorRequest records which represent requests to resume the async generator.</td>
+            <td>A List of AsyncGeneratorRequest records which represent requests to resume the async generator. Except during state transitions, it is nonempty if and only if [[AsyncGeneratorState]] is either ~executing~ or ~awaiting-return~.</td>
           </tr>
           <tr>
             <td>[[GeneratorBrand]]</td>
@@ -39923,7 +39963,7 @@ THH:mm:ss.sss
       <h1>AsyncGenerator Abstract Operations</h1>
       <emu-clause id="sec-asyncgeneratorrequest-records">
         <h1>AsyncGeneratorRequest Records</h1>
-        <p>The AsyncGeneratorRequest is a Record value used to store information about how an async generator should be resumed and contains capabilities for fulfilling or rejecting the corresponding promise.</p>
+        <p>An <dfn>AsyncGeneratorRequest</dfn> is a Record value used to store information about how an async generator should be resumed and contains capabilities for fulfilling or rejecting the corresponding promise.</p>
         <p>They have the following fields:</p>
         <emu-table caption="AsyncGeneratorRequest Record Fields">
           <table>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -39974,26 +39974,66 @@ <h1>AsyncGenerator.prototype.constructor</h1>
...
         <emu-alg>
           1. Let _generator_ be the *this* value.
+          1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+          1. Let _result_ be AsyncGeneratorValidate(_generator_, ~empty~).
+          1. IfAbruptRejectPromise(_result_, _promiseCapability_).
+          1. Let _state_ be _generator_.[[AsyncGeneratorState]].
+          1. If _state_ is ~completed~, then
+            1. Let _iteratorResult_ be ! CreateIterResultObject(*undefined*, *true*).
+            1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, &laquo; _iteratorResult_ &raquo;).
+            1. Return _promiseCapability_.
```
jmdyck, line 39984: A function can't return a spec Record.

spec.html
```diff
@@ -40105,76 +40149,43 @@ <h1>AsyncGeneratorValidate ( _generator_, _generatorBrand_ )</h1>
         </emu-alg>
       </emu-clause>
 
-      <emu-clause id="sec-asyncgeneratorresolve" aoid="AsyncGeneratorResolve">
-        <h1>AsyncGeneratorResolve ( _generator_, _value_, _done_ )</h1>
-        <p>The abstract operation AsyncGeneratorResolve takes arguments _generator_, _value_, and _done_ (a Boolean). It performs the following steps when called:</p>
+      <emu-clause id="sec-asyncgeneratorenqueue" aoid="AsyncGeneratorEnqueue">
+        <h1>AsyncGeneratorEnqueue ( _generator_, _completion_, _promiseCapability_ )</h1>
+        <p>The abstract operation AsyncGeneratorEnqueue takes arguments _generator_ (an Async Generator), _completion_ (a Completion Record), and _promiseCapability_ (a PromiseCapability Record). It performs the following steps when called:</p>
```
jmdyck, line 40154:
```suggestion
        <p>The abstract operation AsyncGeneratorEnqueue takes arguments _generator_ (an AsyncGenerator), _completion_ (a Completion Record), and _promiseCapability_ (a PromiseCapability Record). It performs the following steps when called:</p>
```

spec.html
```diff
@@ -40105,76 +40149,43 @@ <h1>AsyncGeneratorValidate ( _generator_, _generatorBrand_ )</h1>
...
-          1. Let _queue_ be _generator_.[[AsyncGeneratorQueue]].
-          1. Assert: _queue_ is not an empty List.
-          1. Let _next_ be the first element of _queue_.
-          1. Remove the first element from _queue_.
-          1. Let _promiseCapability_ be _next_.[[Capability]].
-          1. Let _iteratorResult_ be ! CreateIterResultObject(_value_, _done_).
-          1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, &laquo; _iteratorResult_ &raquo;).
-          1. Perform ! AsyncGeneratorResumeNext(_generator_).
-          1. Return *undefined*.
+          1. Let _realm_ be the current Realm record.
```
jmdyck, line 40156:
```suggestion
          1. Let _realm_ be the current Realm Record.
```

spec.html
```diff
@@ -40105,76 +40149,43 @@ <h1>AsyncGeneratorValidate ( _generator_, _generatorBrand_ )</h1>
...
+          1. Let _request_ be AsyncGeneratorRequest { [[Completion]]: _completion_, [[Capability]]: _promiseCapability_, [[Realm]]: _realm_ }.
+          1. Append _request_ to the end of _generator_.[[AsyncGeneratorQueue]].
         </emu-alg>
       </emu-clause>
 
-      <emu-clause id="sec-asyncgeneratorreject" aoid="AsyncGeneratorReject">
-        <h1>AsyncGeneratorReject ( _generator_, _exception_ )</h1>
-        <p>The abstract operation AsyncGeneratorReject takes arguments _generator_ and _exception_. It performs the following steps when called:</p>
+      <emu-clause id="sec-ascyngeneratorcompletestep" aoid="AsyncGeneratorCompleteStep">
```
jmdyck, line 40162:
```suggestion
      <emu-clause id="sec-asyncgeneratorcompletestep" aoid="AsyncGeneratorCompleteStep">
```

spec.html
```diff
@@ -40105,76 +40149,43 @@ <h1>AsyncGeneratorValidate ( _generator_, _generatorBrand_ )</h1>
...
+        <h1>AsyncGeneratorCompleteStep ( _generator_, _completion_, _done_ )</h1>
+        <p>The abstract operation AsyncGeneratorCompleteStep takes arguments _generator_ (an AsyncGenerator) and _completion_ (a Completion Record). It performs the following steps when called:</p>
```
jmdyck, line 40164:
```suggestion
        <p>The abstract operation AsyncGeneratorCompleteStep takes arguments _generator_ (an AsyncGenerator), _completion_ (a Completion Record), and _done_ (a Boolean). It performs the following steps when called:</p>
```

spec.html
```diff
@@ -40244,17 +40218,99 @@ <h1>AsyncGeneratorYield ( _value_ )</h1>
...
-          1. Set the code evaluation state of _genContext_ such that when evaluation is resumed with a Completion _resumptionValue_ the following steps will be performed:
-            1. If _resumptionValue_.[[Type]] is not ~return~, return Completion(_resumptionValue_).
-            1. Let _awaited_ be Await(_resumptionValue_.[[Value]]).
-            1. If _awaited_.[[Type]] is ~throw~, return Completion(_awaited_).
-            1. Assert: _awaited_.[[Type]] is ~normal~.
-            1. Return Completion { [[Type]]: ~return~, [[Value]]: _awaited_.[[Value]], [[Target]]: ~empty~ }.
-            1. NOTE: When one of the above steps returns, it returns to the evaluation of the |YieldExpression| production that originally called this abstract operation.
-          1. Return ! AsyncGeneratorResolve(_generator_, _value_, *false*).
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _genContext_.
+          1. Let _completion_ be NormalCompletion(value).
```
jmdyck, line 40221:
```suggestion
          1. Let _completion_ be NormalCompletion(_value_).
```

spec.html
```diff
@@ -40244,17 +40218,99 @@ <h1>AsyncGeneratorYield ( _value_ )</h1>
...
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-asyncgeneratorawaitreturn" aoid="AsyncGeneratorAwaitReturn">
+        <h1>AsyncGeneratorAwaitReturn ( _generator_ )</h1>
+        <p>The abstract operation AsyncGeneratorAwaitReturn takes argument _generator_ (an AsyncGenerator). It performs the following steps when called:</p>
+        <emu-alg>
+          1. Let _queue_ be _generator_.[[AsyncGeneratorQueue]].
+          1. Assert: _queue_ is not empty.
+          1. Let _next_ the first element of _queue_.
```
jmdyck, line 40246:
```suggestion
          1. Let _next_ be the first element of _queue_.
```

spec.html
```diff
@@ -40244,17 +40218,99 @@ <h1>AsyncGeneratorYield ( _value_ )</h1>
...
+            1. Let _completion_ be _next_.[[Completion]].
+            1. If _completion_.[[Type]] is ~return~, then
+              1. Set _generator_.[[AsyncGeneratorState]] to ~awaiting-return~.
+              1. Perform ! AsyncGeneratorAwaitReturn(_generator_).
+              1. Set _done_ to *true*.
+            1. Else,
+              1. If _completion_.[[Type]] is ~normal~, then
+                1. Set _completion_ to NormalCompletion(*undefined*).
+              1. Perform ! AsyncGeneratorCompleteStep(_generator_, _completion_, *true*).
+              1. If _queue_ is empty, set *done* to *true*.
```
jmdyck, line 40313:
```suggestion
              1. If _queue_ is empty, set _done_ to *true*.
```

jmdyck (changes requested):

jmdyck:
Also, this PR undefines the following ids:
- sec-asyncgeneratorreject
- sec-asyncgeneratorresolve
- sec-asyncgeneratorresumenext

(I don't know if there's an appropriate place to make them oldids.)

spec.html
bakkot: Fixed.

bakkot:
> Also, this PR undefines the following ids:

Yeah, I don't think there's a reasonable place to put them. They don't really correspond to any of the new operations - the whole point of the PR is that those previous operations were not a clean way of organizing the logic here, after all.

spec.html
```diff
@@ -40244,17 +40215,102 @@ <h1>AsyncGeneratorYield ( _value_ )</h1>
...
-            1. Let _awaited_ be Await(_resumptionValue_.[[Value]]).
-            1. If _awaited_.[[Type]] is ~throw~, return Completion(_awaited_).
-            1. Assert: _awaited_.[[Type]] is ~normal~.
-            1. Return Completion { [[Type]]: ~return~, [[Value]]: _awaited_.[[Value]], [[Target]]: ~empty~ }.
-            1. NOTE: When one of the above steps returns, it returns to the evaluation of the |YieldExpression| production that originally called this abstract operation.
-          1. Return ! AsyncGeneratorResolve(_generator_, _value_, *false*).
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _genContext_.
+          1. Let _completion_ be NormalCompletion(_value_).
+          1. Assert: There are at least two items on the execution context stack.
+          1. Let _previousContext_ be the execution context immediately beneath the running execution context.
```
jmdyck, line 40220:
```suggestion
          1. Assert: The execution context stack has at least two elements.
          1. Let _previousContext_ be the second to top element of the execution context stack.
```
(precedent in [19.2.1 eval(x)](https://tc39.es/ecma262/#sec-eval-x) and [20.2.1.1.1 CreateDynamicFunction](https://tc39.es/ecma262/#sec-createdynamicfunction))

syg (approved):
@michaelficarra and I are splitting up reviewing the current text for correspondence with [the gist](https://gist.github.com/bakkot/82186e27a2e78c6e507c56dd49d8a197) and reviewing the PR text for correspondence.

This lgtm is for correspondence with the current text.

The state machine in the gist reads correct to me. My understanding is as follows.

The current spec uses AsyncGeneratorResumeNext as a message pump that ~~blocks~~ suspends to wait for new messages. That is, it doesn't just resume next and continues execution, it generally pumps until it runs out of messages. This pump loop is awkwardly captured by self-recursive calls in the following states:

- From `executing`, when a `yield` is evaluated, if the queue is not empty, there is a synchronous transition `executing` -> `suspendedYield` -> `executing`.
- From `completed`, the queue is pumped and each promise is resolved with undefined until a return completion, in which case it goes into `awaiting-return`.

Put a third way, the current spec puts the focus on the message pump machinery, which is more confusing than putting the focus on the state machine machinery.

spec.html
```diff
@@ -40244,17 +40215,102 @@ <h1>AsyncGeneratorYield ( _value_ )</h1>
...
+          1. Set _onFulfilled_.[[Generator]] to _generator_.
+          1. Let _stepsRejected_ be the algorithm steps defined in <emu-xref href="#async-generator-await-return-rejected-functions" title></emu-xref>.
+          1. Let _lengthRejected_ be the number of non-optional parameters of the function definition in <emu-xref href="#async-generator-await-return-rejected-functions" title></emu-xref>.
+          1. Let _onRejected_ be ! CreateBuiltinFunction(_stepsRejected_, _lengthRejected_, *""*, &laquo; [[Generator]] &raquo;).
+          1. Set _onRejected_.[[Generator]] to _generator_.
+          1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
+        </emu-alg>
+
+        <emu-clause id="async-generator-await-return-fulfilled-functions" oldids="async-generator-resume-next-return-processor-fulfilled">
+          <h1>AsyncGeneratorAwaitReturn Fulfilled Functions</h1>
```
michaelficarra, line 40262: I'd prefer to finish up #2109 and then make these abstract closures.

spec.html
```diff
@@ -40244,17 +40215,102 @@ <h1>AsyncGeneratorYield ( _value_ )</h1>
...
+          <p>An AsyncGeneratorAwaitReturn fulfilled function is an anonymous built-in function that is used as part of the AsyncGeneratorAwaitReturn specification device to unwrap promises passed in to the <emu-xref href="#sec-asyncgenerator-prototype-return" title></emu-xref> method. Each AsyncGeneratorAwaitReturn fulfilled function function has a [[Generator]] internal slot.</p>
```
michaelficarra, line 40263:
```suggestion
          <p>An AsyncGeneratorAwaitReturn fulfilled function is an anonymous built-in function that is used as part of the AsyncGeneratorAwaitReturn specification device to unwrap promises passed in to the <emu-xref href="#sec-asyncgenerator-prototype-return" title></emu-xref> method. Each AsyncGeneratorAwaitReturn fulfilled function has a [[Generator]] internal slot.</p>
```

michaelficarra (approved):
LGTM with respect to correspondence to the state machine.

PR #2416, opened 2021-05-26
Editorial: `meta` should be listed alongside `target`

### LINKED ISSUE #2298: `meta` should be listed alongside `target` in 12.6.2
The spec lists various categories of identifier names, the last of which is:

> * Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where Identifier is not allowed: `as`, `async`, `from`, `get`, `of`, `set`, and `target`.

Since `target` is listed here, I think `meta` should be too.

himsngh:
Fixes #2298

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13870,7 +13870,7 @@
           <p>Those that are contextually disallowed as identifiers, in strict mode code: `let`, `static`, `implements`, `interface`, `package`, `private`, `protected`, and `public`;</p>
         </li>
         <li>
-          <p>Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where |Identifier| is not allowed: `as`, `async`, `from`, `get`, `of`, `set`, and `target`.</p>
+          <p>Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where |Identifier| is not allowed: `as`, `async`, `from`, `get`, `meta`, `of`, `set`, and `target`.</p>
         </li>
       </ul>
       <p>The term <dfn>conditional keyword</dfn>, or <dfn>contextual keyword</dfn>, is sometimes used to refer to the keywords that fall in the last three categories, and thus can be used as identifiers in some contexts and as keywords in others.</p>


spec.html
```diff
@@ -13870,7 +13870,7 @@ <h1>Keywords and Reserved Words</h1>
           <p>Those that are contextually disallowed as identifiers, in strict mode code: `let`, `static`, `implements`, `interface`, `package`, `private`, `protected`, and `public`;</p>
         </li>
         <li>
-          <p>Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where |Identifier| is not allowed: `as`, `async`, `from`, `get`, `of`, `set`, and `target`.</p>
+          <p>Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where |Identifier| is not allowed: `as`, `async`, `from`, `get`, `of`, `set`, `meta`, and `target`.</p>
```
nicolo-ribaudo, line 13873: They are in alphabetical order:

```suggestion
          <p>Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where |Identifier| is not allowed: `as`, `async`, `from`, `get`, `meta`, `of`, `set`, and `target`.</p>
```

.gitignore
```diff
@@ -1,6 +1,7 @@
 node_modules/
 out/
 test*.js
+.idea/
```
nicolo-ribaudo, line 4: Ignores specific to your environment should be defined only on your local machine. You can add it to [`.git/info/excludes`](https://luisdalmolin.dev/blog/ignoring-files-in-git-without-gitignore/#:~:text=To%20ignore%20untracked%20files%2C%20you,tracking%20any%20(untracked)%20file.) in your local clone of this repository.

nicolo-ribaudo (approved):

bakkot (approved):

michaelficarra (approved):
WFM. Any way we can avoid forgetting to modify this list in the future when we add more contextual keywords?

PR #2428, opened 2021-06-06
Editorial: fix note about var decls in eval in parameter initializers

bakkot:
As pointed out [here](https://stackoverflow.com/questions/67855714/eval-declaration-instantiation-when-calling-context-is-evaluating-formal-param), this note in `PerformEval` no longer reflects the specified behavior, which changed in #1046.


loganfsmyth:
The SO question also mentions https://262.ecma-international.org/#sec-functiondeclarationinstantiation which has a note

> Note 3
>
> Parameter Initializers may contain direct eval expressions. Any top level declarations of such evals are only visible to the eval code (10.2). The creation of the environment for such declarations is described in 14.1.22.

which also looks incorrect.

bakkot:
Good catch, updated.

michaelficarra (approved):

PR #2429, opened 2021-06-06
Editorial: do not use "Return" to transfer control between execution contexts

### LINKED ISSUE #2400: Editorial: Notes that tell you where a `Return` step returns to

bakkot:
Fixes #2400.

Depends on #2413. Also I'm not 100% sure this all makes sense, so marking as a draft for now.

TODO: ensure the `Let` steps are on their own line (if coherent).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4048,7 +4048,7 @@
             1. Let _prevContext_ be the running execution context.
             1. Suspend _prevContext_.
             1. Push _asyncContext_ onto the execution context stack; _asyncContext_ is now the running execution context.
-            1. Resume the suspended evaluation of _asyncContext_ using NormalCompletion(_value_) as the result of the operation that suspended it.
+            1. Resume the suspended evaluation of _asyncContext_ with NormalCompletion(_value_).
             1. Assert: When we reach this step, _asyncContext_ has already been removed from the execution context stack and _prevContext_ is the currently running execution context.
             1. Return *undefined*.
           1. Let _onFulfilled_ be ! CreateBuiltinFunction(_fulfilledClosure_, 1, *""*, &laquo; &raquo;).
@@ -4056,15 +4056,17 @@
             1. Let _prevContext_ be the running execution context.
             1. Suspend _prevContext_.
             1. Push _asyncContext_ onto the execution context stack; _asyncContext_ is now the running execution context.
-            1. Resume the suspended evaluation of _asyncContext_ using ThrowCompletion(_reason_) as the result of the operation that suspended it.
+            1. Resume the suspended evaluation of _asyncContext_ with ThrowCompletion(_reason_).
             1. Assert: When we reach this step, _asyncContext_ has already been removed from the execution context stack and _prevContext_ is the currently running execution context.
             1. Return *undefined*.
           1. Let _onRejected_ be ! CreateBuiltinFunction(_rejectedClosure_, 1, *""*, &laquo; &raquo;).
           1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-          1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
-          1. Return.
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
+          1. Let _outerContext_ be the running execution context.
+          1. NOTE: _outerContext_ is the execution context for the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
+          1. Resume _outerContext_.
+          1. Assert: If control reaches here, _asyncContext_ is once again the running execution context.
+          1 Let _completion_ be the Completion Record with which _asyncContext_ was resumed.
         </emu-alg>
 
         <p>where all aliases in the above steps, with the exception of _completion_, are ephemeral and visible only in the steps pertaining to Await.</p>
@@ -44624,13 +44626,20 @@ THH:mm:ss.sss
             1. Assert: If we return here, the generator either threw an exception or performed either an implicit or explicit return.
             1. Remove _genContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
             1. Set _generator_.[[GeneratorState]] to ~completed~.
-            1. Once a generator enters the ~completed~ state it never leaves it and its associated execution context is never resumed. Any execution state associated with _generator_ can be discarded at this point.
-            1. If _result_.[[Type]] is ~normal~, let _resultValue_ be *undefined*.
-            1. Else if _result_.[[Type]] is ~return~, let _resultValue_ be _result_.[[Value]].
+            1. NOTE: once a generator enters the ~completed~ state it never leaves it and its associated execution context is never resumed. Any execution state associated with _generator_ can be discarded at this point.
+            1. If _result_.[[Type]] is ~normal~, then
+              1. Let _resultValue_ be CreateIterResultObject(*undefined*, *true*).
+              1. Let _resultCompletion_ be NormalCompletion(_resultValue_).
+            1. Else if _result_.[[Type]] is ~return~, then
+              1. Let _resultValue_ be CreateIterResultObject(_result_.[[Value]], *true*).
+              1. Let _resultCompletion_ be NormalCompletion(_resultValue_).
             1. Else,
               1. Assert: _result_.[[Type]] is ~throw~.
-              1. Return Completion(_result_).
-            1. Return CreateIterResultObject(_resultValue_, *true*).
+              1. Let _resultCompletion_ be _result_.
+            1. Let _outerContext_ be the running execution context.
+            1. NOTE: _outerContext_ is the execution context for the evaluation of the operation that had most previously resumed evaluation of _genContext_.
+            1. Resume _outerContext_ passing _resultCompletion_.
+            1. Assert: Control does not reach here.
           1. Set _generator_.[[GeneratorContext]] to _genContext_.
           1. Set _generator_.[[GeneratorState]] to ~suspendedStart~.
           1. Return NormalCompletion(*undefined*).
@@ -44676,7 +44685,7 @@ THH:mm:ss.sss
           1. Suspend _methodContext_.
           1. Set _generator_.[[GeneratorState]] to ~executing~.
           1. Push _genContext_ onto the execution context stack; _genContext_ is now the running execution context.
-          1. Resume the suspended evaluation of _genContext_ using NormalCompletion(_value_) as the result of the operation that suspended it. Let _result_ be the value returned by the resumed computation.
+          1. Resume the suspended evaluation of _genContext_ with NormalCompletion(_value_). Let _result_ be the Completion Record returned by the resumed computation.
           1. Assert: When we return here, _genContext_ has already been removed from the execution context stack and _methodContext_ is the currently running execution context.
           1. Return Completion(_result_).
         </emu-alg>
@@ -44708,7 +44717,7 @@ THH:mm:ss.sss
           1. Suspend _methodContext_.
           1. Set _generator_.[[GeneratorState]] to ~executing~.
           1. Push _genContext_ onto the execution context stack; _genContext_ is now the running execution context.
-          1. Resume the suspended evaluation of _genContext_ using _abruptCompletion_ as the result of the operation that suspended it. Let _result_ be the completion record returned by the resumed computation.
+          1. Resume the suspended evaluation of _genContext_ with _abruptCompletion_. Let _result_ be the Completion Record returned by the resumed computation.
           1. Assert: When we return here, _genContext_ has already been removed from the execution context stack and _methodContext_ is the currently running execution context.
           1. Return Completion(_result_).
         </emu-alg>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -3572,9 +3572,11 @@ <h1>Await</h1>
           1. Set _onRejected_.[[AsyncContext]] to _asyncContext_.
           1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-          1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
-          1. Return.
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
+          1. Let _outerContext_ be the running execution context.
```
brabalan, line 3575: Could you name `_outerContext` in the previous step (the one where `_asyncContext` is removed from the stack) and thus skip this step?

spec.html
```diff
@@ -39913,11 +39922,13 @@ <h1>GeneratorYield ( _iterNextObj_ )</h1>
           1. Assert: GetGeneratorKind() is ~sync~.
           1. Set _generator_.[[GeneratorState]] to ~suspendedYield~.
           1. Remove _genContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-          1. Set the code evaluation state of _genContext_ such that when evaluation is resumed with a Completion _resumptionValue_ the following steps will be performed:
-            1. Return _resumptionValue_.
-            1. NOTE: This returns to the evaluation of the |YieldExpression| that originally called this abstract operation.
-          1. Return NormalCompletion(_iterNextObj_).
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _genContext_.
+          1. Let _outerContext_ be the running execution context.
```
brabalan, line 39925: Same comment as above, `_outerContext_` could be named in the previous step.

spec.html
```diff
@@ -40229,11 +40241,13 @@ <h1>AsyncGeneratorYield ( _value_ )</h1>
           1. Else,
             1. Set _generator_.[[AsyncGeneratorState]] to ~suspendedYield~.
             1. Remove _genContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-            1. Set the code evaluation state of _genContext_ such that when evaluation is resumed with a Completion _resumptionValue_ the following steps will be performed:
-              1. Return AsyncGeneratorUnwrapYieldResumption(_resumptionValue_).
-              1. NOTE: When the above step returns, it returns to the evaluation of the |YieldExpression| production that originally called this abstract operation.
-            1. Return *undefined*.
-            1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _genContext_.
+            1. Let _outerContext_ be the running execution context.
```
brabalan, line 40244: Same as above.

spec.html
```diff
@@ -3572,9 +3572,11 @@ <h1>Await</h1>
...
           1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-          1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
-          1. Return.
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
+          1. Let _outerContext_ be the running execution context.
+          1. NOTE: _outerContext_ is the execution context for the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
+          1. Resume _outerContext_.
+          1. Assert: If control reaches here, _asyncContext_ is once again the running execution context.
+          1 Let _completion_ be the Completion Record with which _asyncContext_ was resumed.
```
jmdyck, line 3579:
```suggestion
          1. Let _completion_ be the Completion Record with which _asyncContext_ was resumed.
```

spec.html
```diff
@@ -3572,9 +3572,11 @@ <h1>Await</h1>
           1. Set _onRejected_.[[AsyncContext]] to _asyncContext_.
           1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-          1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
-          1. Return.
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
+          1. Let _outerContext_ be the running execution context.
+          1. NOTE: _outerContext_ is the execution context for the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
+          1. Resume _outerContext_.
+          1. Assert: If control reaches here, _asyncContext_ is once again the running execution context.
```
jmdyck, line 3578:
```suggestion
          1. Assert: If control reaches here, then _asyncContext_ is once again the running execution context.
```

spec.html
```diff
@@ -39913,11 +39922,13 @@ <h1>GeneratorYield ( _iterNextObj_ )</h1>
...
           1. Remove _genContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-          1. Set the code evaluation state of _genContext_ such that when evaluation is resumed with a Completion _resumptionValue_ the following steps will be performed:
-            1. Return _resumptionValue_.
-            1. NOTE: This returns to the evaluation of the |YieldExpression| that originally called this abstract operation.
-          1. Return NormalCompletion(_iterNextObj_).
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _genContext_.
+          1. Let _outerContext_ be the running execution context.
+          1. NOTE: _outerContext_ is the execution context for the evaluation of the operation that had most previously resumed evaluation of _genContext_.
+          1. Resume _outerContext_ passing NormalCompletion(_iterNextObj_).
+          1. Assert: If control reaches here, _genContext_ is once again the running execution context.
```
jmdyck, line 39928:
```suggestion
          1. Assert: If control reaches here, then _genContext_ is once again the running execution context.
```

spec.html
```diff
@@ -40229,11 +40241,13 @@ <h1>AsyncGeneratorYield ( _value_ )</h1>
...
             1. Remove _genContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-            1. Set the code evaluation state of _genContext_ such that when evaluation is resumed with a Completion _resumptionValue_ the following steps will be performed:
-              1. Return AsyncGeneratorUnwrapYieldResumption(_resumptionValue_).
-              1. NOTE: When the above step returns, it returns to the evaluation of the |YieldExpression| production that originally called this abstract operation.
-            1. Return *undefined*.
-            1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _genContext_.
+            1. Let _outerContext_ be the running execution context.
+            1. NOTE: _outerContext_ is the execution context for the evaluation of the operation that had most previously resumed evaluation of _genContext_.
+            1. Resume _outerContext_.
+            1. Assert: If control reaches here, _genContext_ is once again the running execution context.
```
jmdyck, line 40247:
```suggestion
            1. Assert: If control reaches here, then _genContext_ is once again the running execution context.
```

jmdyck (changes requested):

spec.html
```diff
@@ -39820,13 +39822,20 @@ <h1>GeneratorStart ( _generator_, _generatorBody_ )</h1>
             1. Assert: If we return here, the generator either threw an exception or performed either an implicit or explicit return.
             1. Remove _genContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
             1. Set _generator_.[[GeneratorState]] to ~completed~.
-            1. Once a generator enters the ~completed~ state it never leaves it and its associated execution context is never resumed. Any execution state associated with _generator_ can be discarded at this point.
-            1. If _result_.[[Type]] is ~normal~, let _resultValue_ be *undefined*.
-            1. Else if _result_.[[Type]] is ~return~, let _resultValue_ be _result_.[[Value]].
+            1. NOTE: once a generator enters the ~completed~ state it never leaves it and its associated execution context is never resumed. Any execution state associated with _generator_ can be discarded at this point.
```
jmdyck, line 39825:
```suggestion
            1. NOTE: Once a generator enters the ~completed~ state it never leaves it and its associated execution context is never resumed. Any execution state associated with _generator_ can be discarded at this point.
```

spec.html
```diff
@@ -4048,23 +4048,25 @@ <h1>Await</h1>
...
             1. Return *undefined*.
           1. Let _onRejected_ be ! CreateBuiltinFunction(_rejectedClosure_, 1, *""*, &laquo; &raquo;).
           1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-          1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
-          1. Return.
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
+          1. Let _outerContext_ be the running execution context.
+          1. NOTE: _outerContext_ is the execution context for the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
```
jmdyck, line 4066: It's unclear what "most previously" is supposed to mean.

If it means "the longest ago", then that's wrong, because the execution context that first resumed evaluation of _asyncContext_ isn't even necessarily on the stack, let alone in that position on the stack.

And if it means "most recently", then that got a problem too. It's true that, in the status quo, the most recent `Resume the suspended evaluation of _asyncContext_` step must have come from the execution context that's below it on the stack. However, this PR introduces "resume" steps that go in the 'other direction', so if _asyncContext_'s function itself calls an Async/GeneratorFunction, then it may have been resumed several times since _outerContext_ resumed it.

So unless you want to introduce terminology that distinguishes the two 'directions' of resumes, I think it might be hard to describe how _outerContext_ relates to _asyncContext_ (other than being directly below it on the stack, of course).

spec.html
```diff
@@ -4048,23 +4048,25 @@ <h1>Await</h1>
...
+            1. Resume the suspended evaluation of _asyncContext_ with ThrowCompletion(_reason_).
             1. Assert: When we reach this step, _asyncContext_ has already been removed from the execution context stack and _prevContext_ is the currently running execution context.
             1. Return *undefined*.
           1. Let _onRejected_ be ! CreateBuiltinFunction(_rejectedClosure_, 1, *""*, &laquo; &raquo;).
           1. Perform ! PerformPromiseThen(_promise_, _onFulfilled_, _onRejected_).
           1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
-          1. Set the code evaluation state of _asyncContext_ such that when evaluation is resumed with a Completion _completion_, the following steps of the algorithm that invoked Await will be performed, with _completion_ available.
-          1. Return.
-          1. NOTE: This returns to the evaluation of the operation that had most previously resumed evaluation of _asyncContext_.
+          1. Let _outerContext_ be the running execution context.
```
jmdyck, line 4065: Currently, the spec only uses "outer" when talking about environments, which is appropriate because we think of outer and inner scopes. But for these execution contexts, outer/inner doesn't seem like the right metaphor. Elsewhere, the spec calls them "caller" + "callee". (Which makes sense, though the fact that they only differ by one letter makes them a bit hard to distinguish visually.)

bakkot:
Closing in favor of https://github.com/tc39/ecma262/pull/2665 etc.

PR #2433, opened 2021-06-11
Normative: restrict abstract op. completion type

jugglinmike:
Specify two acceptable completion types for implementations of
HostEnsureCanCompileStrings.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -24623,7 +24623,7 @@
       <emu-clause id="sec-hostensurecancompilestrings" aoid="HostEnsureCanCompileStrings">
         <h1>HostEnsureCanCompileStrings ( _callerRealm_, _calleeRealm_ )</h1>
         <p>The host-defined abstract operation HostEnsureCanCompileStrings takes arguments _callerRealm_ (a Realm Record) and _calleeRealm_ (a Realm Record). It allows host environments to block certain ECMAScript functions which allow developers to compile strings into ECMAScript code.</p>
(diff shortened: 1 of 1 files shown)


jugglinmike:
I have not enabled GitHub's "Allow Edits" feature because the feature is not available for this pull request (perhaps because the source repository does not belong to my personal GitHub account).

ljharb:
@jugglinmike unfortunately, when you make a PR from a fork that you don't own, github has a bug where we can't force push to it - it'd be great if you could add me to that repo, so when it's time to merge this I can rebase it and adjust the commit message as needed.

ljharb:
Does this need consensus?

jugglinmike:
Works for me. GitHub says:

> ljharb has been added as a collaborator with write permissions on the repository.

syg:
> Does this need consensus?

I'd argue no as it was more of an underspecification: I'd be very surprised if anyone understood the current consensus to mean that this host hook can return `break`, `continue`, or `return` completions.

Though I do think it's worthwhile to seek consensus on imposing the restriction that _no_ host hooks can result in transfer of control, i.e. no hooks can return a `break`, `continue`, or `return` abrupt completions. Out of those `return` is the only one that might be useful. Even in this case, `~handled~` could be modeled by a return completion, but I think that's harder to read.

syg:
We discussed this on the editor call. The plan here is:

1. Require all host hook AOs to be only able to return normal completions or throw completions.
2. Require the same of all host functions.

We will go ahead and make the change for 1, though it's normative in a narrow sense.

We'll bring 2 to committee for consensus, though it should be uncontroversial.

syg:
#2442 has landed, which supersedes this.

Issue #925, opened 2017-05-30
Inconsistent spacing around binary operators


domenic:
Mostly there seems to be space between binary operators and their operands, e.g. `Repeat, while q ≠ s`, `If e = p`, etc. But sometimes there is no spacing, e.g. the latter half of `If e = p, let q be q+1`, or the former half of `If q+r > s`.

Obviously very minor, but I think cleaning this up to add spaces everywhere would be good.

dantesolis:
hey guys, could I work on this issue, is this a good issue for a `first-time contributor` or should I take something else?  Thanks

littledan:
@dantesolis Go for it! I think an editorial patch like this would be great for a first-time contributor. You can do this by editing the spec.html file in this repository and making a PR on GitHub. Feel free to ask questions here or on the #tc39 channel on Freenode IRC.

ljharb:
@dantesolis are you still hoping to take this on?

jmdyck:
If not, I'll do it.

littledan:
@jmdyck Go for it. I also ran into this in the mathematical values change.

PR #2435, opened 2021-06-14
Editorial: extract StringNumericValue from MV and add/use RoundStringMVResult helper

michaelficarra:


bakkot:
Note that this races with #1554.

spec.html
```diff
@@ -4490,51 +4490,59 @@ <h2>Syntax</h2>
...
+          </emu-alg>
+          <emu-grammar>StringNumericLiteral ::: StrWhiteSpace? StrNumericLiteral StrWhiteSpace?</emu-grammar>
+          <emu-alg>
+            1. Return 𝔽(MV of |StrNumericLiteral|).
+          </emu-alg>
+          <emu-grammar>StrDecimalLiteral ::: `-` StrUnsignedDecimalLiteral</emu-grammar>
+          <emu-alg>
+            1. Let _a_ be MV of |StrUnsignedDecimalLiteral|.
+            1. If _a_ is 0, return *-0*<sub>𝔽</sub>.
+            1. Return RoundStringMVResult(-1 &times; _a_).
```
michaelficarra, line 4507: We could also do this:

```suggestion
            1. Return RoundStringMVResult(0 - _a_).
```

Or this:

```suggestion
            1. Return RoundStringMVResult(-_a_).
```

ljharb: second seems most straightforward to me

spec.html
```diff
@@ -4490,51 +4490,59 @@ <h2>Syntax</h2>
...
+          <p>The conversion of a String to a Number value is similar overall to the determination of the Number value for a numeric literal (see <emu-xref href="#sec-literals-numeric-literals"></emu-xref>), but some of the details are different, so the process for converting a String numeric literal to a value of Number type is given here. Additional definitions for MV are provided in <emu-xref href="#sec-static-semantics-mv"></emu-xref>.</p>
+          <emu-grammar>StringNumericLiteral ::: [empty]</emu-grammar>
+          <emu-grammar>StringNumericLiteral ::: StrWhiteSpace</emu-grammar>
+          <emu-alg>
+            1. Return *+0*<sub>𝔽</sub>.
+          </emu-alg>
+          <emu-grammar>StringNumericLiteral ::: StrWhiteSpace? StrNumericLiteral StrWhiteSpace?</emu-grammar>
+          <emu-alg>
+            1. Return 𝔽(MV of |StrNumericLiteral|).
+          </emu-alg>
```
jmdyck, line 4502: You should probably have a rule for `StrNumericLiteral ::: NonDecimalIntegerLiteral`, where you apply 𝔽 to the MV of NonDecimalIntegerLiteral.

spec.html
```diff
@@ -4490,51 +4490,59 @@ <h2>Syntax</h2>
...
+            1. If |ExponentPart| is present, let _e_ be MV of |ExponentPart|. Otherwise, let _e_ be 0.
+            1. Let _n_ be the number of code points in the second |DecimalDigits|.
+            1. Return RoundStringMVResult((_a_ + (_b_ &times; 10<sup>-_n_</sup>)) &times; 10<sup>_e_</sup>).
+          </emu-alg>
+          <emu-grammar>StrUnsignedDecimalLiteral ::: `.` DecimalDigits ExponentPart?</emu-grammar>
+          <emu-alg>
+            1. Let _a_ be MV of |DecimalDigits|.
+            1. If |ExponentPart| is present, let _e_ be MV of |ExponentPart|. Otherwise, let _e_ be 0.
+            1. Let _n_ be the number of code points in |DecimalDigits|.
+            1. Return RoundStringMVResult(_a_ &times; 10<sup>_e_-_n_</sup>).
```
jmdyck, line 4526:
```suggestion
            1. Return RoundStringMVResult(_a_ &times; 10<sup>_e_ - _n_</sup>).
```

spec.html
```diff
@@ -4490,51 +4490,59 @@ <h2>Syntax</h2>
...
+          </emu-alg>
+          <emu-grammar>StrUnsignedDecimalLiteral ::: DecimalDigits ExponentPart?</emu-grammar>
+          <emu-alg>
+            1. Let _a_ be MV of |DecimalDigits|.
+            1. If |ExponentPart| is present, let _e_ be MV of |ExponentPart|. Otherwise, let _e_ be 0.
+            1. Return RoundStringMVResult(_a_ &times; 10<sup>_e_</sup>).
+          </emu-alg>
```
jmdyck, line 4533: The algorithm for built-in function `parseFloat` invokes `MV` on a `StrDecimalLiteral`, and it's expecting the result to be a mathematical value, but the rules above define it to be a Number.

spec.html
```diff
@@ -4490,51 +4490,59 @@ <h2>Syntax</h2>
...
+        </emu-clause>
+
+        <emu-clause id="sec-roundstringmvresult" aoid="RoundStringMVResult">
+          <h1>RoundStringMVResult ( _n_ )</h1>
+          <p>The abstract operation RoundStringMVResult takes argument _n_ (a mathematical value). It converts _n_ to a Number in an implementation-defined manner. For the purposes of this abstract operation, a digit is significant if it is not zero or there is a non-zero digit to its left and there is a non-zero digit to its right. It performs the following steps when called:</p>
+          <emu-alg>
+            1. If the decimal representation of _n_ has fewer than 20 significant digits in its decimal representation, return 𝔽(_n_).
```
jmdyck, line 4540:
```suggestion
            1. If the decimal representation of _n_ has fewer than 20 significant digits, return 𝔽(_n_).
```

spec.html
```diff
@@ -4490,51 +4490,59 @@ <h2>Syntax</h2>
...
+            1. Let _option1_ be the mathematical value denoted by the result of replacing each significant digit in the decimal representation of _n_ after the 20th with a 0 digit.
+            1. Let _option2_ be the mathematical value denoted by the result of replacing each significant digit in the decimal representation of _n_ after the 20th with a 0 digit and then incrementing it at the 20th position (with carrying as necessary).
```
jmdyck, line 4542: It's odd that this operation can 'invoke' "the mathematical value denoted by <a sequence of digits>" when MV itself has to explicitly define the rules by which that value is computed. (The current spec uses MV at the corresponding points.)

jmdyck (changes requested):

spec.html
```diff
@@ -4490,51 +4490,59 @@ <h2>Syntax</h2>
 
         <emu-clause id="sec-runtime-semantics-mv-s">
           <h1>Runtime Semantics: MV</h1>
```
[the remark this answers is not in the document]
michaelficarra: I am not concerned about the naming consistency from your first point. I agree that this operation should be renamed, but that was not the point of this PR and we can address that in a follow-up.

spec.html
michaelficarra: This is again the thing we were talking past each other about in #1554. See the definition of "the mathematical value of" in [5.2.5 Mathematical Operations](https://tc39.es/ecma262/#sec-mathematical-operations). While the kind of usage here is arguably still not described (and probably should be), notice how there is no mention of the MV AO in that entire section. MV is just another AO and you should avoid associating it with the spec's internal conversions between numeric types.

spec.html
```diff
@@ -4490,51 +4490,59 @@ <h2>Syntax</h2>
...
-          </ul>
+          <p>The conversion of a String to a Number value is similar overall to the determination of the Number value for a numeric literal (see <emu-xref href="#sec-literals-numeric-literals"></emu-xref>), but some of the details are different, so the process for converting a String numeric literal to a value of Number type is given here. Additional definitions for MV are provided in <emu-xref href="#sec-static-semantics-mv"></emu-xref>.</p>
+          <emu-grammar>StringNumericLiteral ::: [empty]</emu-grammar>
+          <emu-grammar>StringNumericLiteral ::: StrWhiteSpace</emu-grammar>
+          <emu-alg>
+            1. Return *+0*<sub>𝔽</sub>.
+          </emu-alg>
+          <emu-grammar>StringNumericLiteral ::: StrWhiteSpace? StrNumericLiteral StrWhiteSpace?</emu-grammar>
+          <emu-alg>
+            1. Return 𝔽(MV of |StrNumericLiteral|).
```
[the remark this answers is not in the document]
michaelficarra: Good catch, I missed that chaining. It looks like this'll force me to separate this operation from MV after all.

spec.html
jmdyck: (ref issue #925)

michaelficarra:
@jmdyck Thanks for the review. I've made another major refactoring to split this AO from MV. Please review again.

spec.html
jmdyck: > This is again the thing we were talking past each other about in #1554.

(I don't think so. At least, it wasn't what *I* was talking about in #1554.)

> See the definition of "the mathematical value of" in [5.2.5 Mathematical Operations](https://tc39.es/ecma262/#sec-mathematical-operations). While the kind of usage here is arguably still not described (and probably should be), notice how there is no mention of the MV AO in that entire section. MV is just another AO

"The mathematical value of x" is how the spec converts a Number or BigInt value to a mathematical value, but that's not relevant to these steps because neither involves a Number or BigInt value. 

Or are you drawing an analogy? I.e., in the same way that the definition of "the mathematical value of x" doesn't refer to MV, so too the definition of RoundStringMVResult shouldn't refer to MV? I don't see how that analogy is valid.

Do you think the current spec is incorrect to invoke MV at the corresponding points in 7.1.4.1.1 and 12.8.3.2?

> and you should avoid associating it with the spec's internal conversions between numeric types.

I'm not.

michaelficarra: I just don't understand at all what you're talking about here. Do you agree with me that the AO named MV and the prose "the mathematical value of" are not related in any way? Is the fact that the name `MV` is composed of the first two letters of "mathematical" and "value" causing you to think that it is being used as an initialism for "mathematical value" instead of to refer to the AO with that name?

michaelficarra: We talked about this a bit in the editor call today. @jmdyck does the most recent commit (e61d6dc) address your concerns?

spec.html
```diff
@@ -4488,53 +4488,71 @@ <h2>Syntax</h2>
...
-          <ul>
-            <li>
-              it is not `0`; or
-            </li>
-            <li>
-              there is a non-zero digit to its left and there is a non-zero digit, not in the |ExponentPart|, to its right.
-            </li>
-          </ul>
+        <emu-clause id="sec-runtime-semantics-stringnumericvalue" aoid="StringNumericValue" oldids="sec-runtime-semantics-mv-s">
```
jmdyck, line 4491:
```suggestion
        <emu-clause id="sec-runtime-semantics-stringnumericvalue" type="sdo" aoid="StringNumericValue" oldids="sec-runtime-semantics-mv-s">
```

spec.html
```diff
@@ -4488,53 +4488,71 @@ <h2>Syntax</h2>
...
+          <emu-alg>
+            1. Return StringNumericValue of |StrNumericLiteral|.
+          </emu-alg>
+          <emu-grammar>StrNumericLiteral ::: NonDecimalIntegerLiteral</emu-grammar>
+          <emu-alg>
+            1. Return RoundStringMVResult(MV of |NonDecimalIntegerLiteral|).
+          </emu-alg>
+          <emu-grammar>
+            StrDecimalLiteral ::: StrUnsignedDecimalLiteral
+            StrDecimalLiteral ::: `+` StrUnsignedDecimalLiteral
```
jmdyck, line 4508: The two productions should be separated by an empty line.

Alternatively, you could drop this grammar+alg pair and let the chain rule handle it.

spec.html
jmdyck: If the editors prefer the superscript without spaces, there are about 18 other superscripts *with* spaces that you might wish to convert.

spec.html
```diff
@@ -4488,53 +4488,71 @@ <h2>Syntax</h2>
...
+          <emu-grammar>StrUnsignedDecimalLiteral ::: DecimalDigits ExponentPart?</emu-grammar>
+          <emu-alg>
+            1. Let _a_ be MV of |DecimalDigits|.
+            1. If |ExponentPart| is present, let _e_ be MV of |ExponentPart|. Otherwise, let _e_ be 0.
+            1. Return RoundStringMVResult(_a_ &times; 10<sup>_e_</sup>).
+          </emu-alg>
+        </emu-clause>
+
+        <emu-clause id="sec-roundstringmvresult" aoid="RoundStringMVResult">
+          <h1>RoundStringMVResult ( _n_ )</h1>
```
jmdyck, line 4547: Since this operation is also used for NumericLiterals, where there isn't a String involved, you might want to remove "String" from the name.

spec.html
```diff
@@ -4488,53 +4488,71 @@ <h2>Syntax</h2>
...
+          <p>The abstract operation RoundStringMVResult takes argument _n_ (a mathematical value). It converts _n_ to a Number in an implementation-defined manner. For the purposes of this abstract operation, a digit is significant if it is not zero or there is a non-zero digit to its left and there is a non-zero digit to its right. For the purposes of this abstract operation, "the mathematical value denoted by" a representation of a mathematical value is the inverse of "the decimal representation of" a mathematical value. It performs the following steps when called:</p>
+          <emu-alg>
+            1. If the decimal representation of _n_ has fewer than 20 significant digits, return 𝔽(_n_).
```
jmdyck, line 4550:
```suggestion
            1. If the decimal representation of _n_ has 20 or fewer significant digits, return 𝔽(_n_).
```
(or "fewer than 21", I suppose.) As is, if the decimal rep has exactly 20 sig digits, then _option2_ increments the 20th position, which is not currently allowed.

spec.html
```diff
@@ -4488,53 +4488,71 @@ <h2>Syntax</h2>
...
+          <emu-alg>
+            1. Return *+0*<sub>𝔽</sub>.
+          </emu-alg>
+          <emu-grammar>StringNumericLiteral ::: StrWhiteSpace? StrNumericLiteral StrWhiteSpace?</emu-grammar>
+          <emu-alg>
+            1. Return StringNumericValue of |StrNumericLiteral|.
+          </emu-alg>
+          <emu-grammar>StrNumericLiteral ::: NonDecimalIntegerLiteral</emu-grammar>
+          <emu-alg>
+            1. Return RoundStringMVResult(MV of |NonDecimalIntegerLiteral|).
```
jmdyck, line 4504:
```suggestion
            1. Return 𝔽(MV of |NonDecimalIntegerLiteral|).
```
In the current spec, the 20-digit rounding only applies to *decimal* literals. If you apply it to non-decimal literals, then there are cases where the spec would allow two valid outcomes where it currently only allows one, which would be a normative change.

spec.html
```diff
@@ -14233,22 +14251,12 @@ <h1>Static Semantics: MV</h1>
         <h1>Static Semantics: NumericValue</h1>
         <emu-grammar>NumericLiteral :: DecimalLiteral</emu-grammar>
         <emu-alg>
-          1. Return the Number value that results from rounding the MV of |DecimalLiteral| as described below.
+          1. Return RoundStringMVResult(MV of |DecimalLiteral|).
         </emu-alg>
         <emu-grammar>NumericLiteral :: NonDecimalIntegerLiteral</emu-grammar>
         <emu-alg>
-          1. Return the Number value that results from rounding the MV of |NonDecimalIntegerLiteral| as described below.
+          1. Return RoundStringMVResult(MV of |NonDecimalIntegerLiteral|).
```
jmdyck, line 14258:
```suggestion
          1. Return 𝔽(MV of |NonDecimalIntegerLiteral|).
```
As above, 20-digit rounding doesn't apply to non-decimal literals.

spec.html
```diff
@@ -4488,53 +4488,71 @@ <h2>Syntax</h2>
...
+          <emu-alg>
+            1. If the decimal representation of _n_ has fewer than 20 significant digits, return 𝔽(_n_).
+            1. Let _option1_ be the mathematical value denoted by the result of replacing each significant digit in the decimal representation of _n_ after the 20th with a 0 digit.
+            1. Let _option2_ be the mathematical value denoted by the result of replacing each significant digit in the decimal representation of _n_ after the 20th with a 0 digit and then incrementing it at the 20th position (with carrying as necessary).
+            1. Let _chosen_ be an implementation-defined choice of either _option1_ or _option2_.
+            1. Return 𝔽(_chosen_).
+          </emu-alg>
```
jmdyck, line 4555: Before the call to RoundStringMVResult, we convert a decimal representation to a mathematical value, and then inside the call, we convert that to a decimal representation, tweak that, then convert that to a mathematical value, and convert *that* to a Number. So it's:
- decimal rep -> math value -> decimal rep -> math value -> Number

I'm not saying it's wrong, it just seems circuitous.

spec.html
jmdyck: > Is the fact that the name `MV` is composed of the first two letters of "mathematical" and "value" causing you to think that it is being used as an initialism for "mathematical value" instead of to refer to the AO with that name?

Nope.

> Do you agree with me that the AO named MV and the prose "the mathematical value of" are not related in any way?

Sure:
- MV is an AO that maps literals to mathematical values.
- "the mathematical value of" is a special phrase that 'converts' Numbers and BigInts to mathematical values.
 
And neither is defined in terms of the other.

However, what I'm talking about here is the new phrase "the mathematical value *denoted by*", which maps 'representations' to mathematical values, and thus is similar to (some of) what MV currently does. I'm saying it's odd  that RoundStringMVResult can just say "the mathematical value denoted by x" when elsewhere we explicitly do the arithmetic to accomplish the same result. It's an inconsistency in what the spec assumes as its 'base level'.

> We talked about this a bit in the editor call today. @jmdyck does the most recent commit address your concerns?

(This adds a sentence to RoundStringMVResult defining the new phrase.) That's an improvement, I guess, but no, I don't think it addresses the point I raised.

- One way to address it would be, as in the current spec, to use MV (rather than the new phrase) to get the mathematical values of the two 20-digit options. (Though that's somewhat complicated by RoundStringMVResult working with decimal representations rather than literals.)

- Another way would be to keep the new phrase and use it elsewhere instead of explicit arithmetic. (Though I don't know how well that would work out.)

But if the editors understand my point and are okay with the inconsistency, well, it's not like it'd be the only one in the spec.

jmdyck:
(Hm, that's weird. I meant to submit a reply and a review separately, but GitHub seems to have merged them. I think it still makes sense.)

spec.html
bakkot: I think it's reasonable to assume that readers understand how to derive a mathematical value from its decimal representation, as is done in this PR, but since MV (the "elsewhere") is working with parse nodes - which are not normally imbued with mathematical values as a matter of background knowledge, unlike decimal representations - it warrants a more explicit definition there. (Besides which, MV has a number of other cases it needs to define explicitly anyway, so it might as well be explicit in this case.)

That is to say: the operation "take the mathematical value denoted by this decimal representation" is a reasonable thing to assume readers are familiar with. The operation "take the mathematical value denoted by this parse node" is less so, even when the parse node happens to represent a string which is the decimal representation of a number, since the parse nodes are an artifact defined by this specification.

So I don't really regard this as an inconsistency. At any rate, I'm fine with it.

jmdyck: Okay, thanks.

spec.html
jmdyck: Oh, something else: In the current spec, "the decimal representation of x" is only applied to integers (and only positive integers, I think), whereas this applies it to non-integers (and negatives, and zero). I don't think this creates any definitional problems, but I thought it shouldn't go un-noted.

spec.html
michaelficarra: @jmdyck Thanks, that was much clearer. I think I understand your concerns now. My opinion on the matter matches @bakkot's.

spec.html
michaelficarra: Thanks, I'll look into it.

spec.html
michaelficarra: Are you sure? It seems to apply in this case. Are you saying that it was a spec bug before?

jmdyck: NonDecimalIntegerLiteral's MV *is* subject to "rounding ... as described below" but that amounts to just the standard "the Number value for x" (i.e., 𝔽). The 20-digit 2-option rounding is only invoked if the numeric literal is a DecimalLiteral.

michaelficarra: Oh I see, I missed this part:

> unless the literal is a |DecimalLiteral| and

spec.html
michaelficarra: The first input (the input to MV) is a parse node, not a (typical) decimal representation of a mathematical value. I think this is fine, but I'd be open to alternative suggestions.

spec.html
syg: Let's keep spaces for this PR and decide the rendering afterwards.

spec.html
```diff
@@ -14349,7 +14350,7 @@ <h2>Syntax</h2>
 
       <emu-clause id="sec-static-semantics-sv" oldids="sec-string-literals-static-semantics-stringvalue" type="sdo" aoid="SV">
         <h1>Static Semantics: SV</h1>
-        <p>A string literal stands for a value of the String type. The String value (SV) of the literal is described in terms of String values contributed by the various parts of the string literal. As part of this process, some Unicode code points within the string literal are interpreted as having a mathematical value (MV), as described below or in <emu-xref href="#sec-literals-numeric-literals"></emu-xref>.</p>
+        <p>A string literal stands for a value of the String type. The SV of the literal is described in terms of String values contributed by the various parts of the string literal. As part of this process, some Unicode code points within the string literal are interpreted as having a mathematical value, as described below or in <emu-xref href="#sec-literals-numeric-literals"></emu-xref>.</p>
```
syg, line 14353: Here with SV and below with TV and TRV, how come you removed the expansion of the acronyms? I feel like they read a bit clearer with the expansion.

syg (approved):
lgtm, though I'd like to hear the reasoning for not expanding SV, TV, and TRV in prose.

spec.html
```diff
@@ -4488,53 +4488,64 @@ <h2>Syntax</h2>
...
+          <emu-grammar>StrUnsignedDecimalLiteral ::: `Infinity`</emu-grammar>
+          <emu-alg>
+            1. Return *+&infin;*<sub>𝔽</sub>.
+          </emu-alg>
+          <emu-grammar>StrUnsignedDecimalLiteral ::: DecimalDigits `.` DecimalDigits? ExponentPart?</emu-grammar>
+          <emu-alg>
+            1. Let _a_ be MV of the first |DecimalDigits|.
+            1. If the second |DecimalDigits| is present, let _b_ be MV of the second |DecimalDigits|. Otherwise, let _b_ be 0.
+            1. If |ExponentPart| is present, let _e_ be MV of |ExponentPart|. Otherwise, let _e_ be 0.
+            1. Let _n_ be the number of code points in the second |DecimalDigits|.
```
jmdyck, line 4521: If `the second |DecimalDigits|` is absent, it's unclear if we can talk about the number of code points in it. (The spec says basically nothing about absent Parse Nodes.) On the other hand, it'd be reasonable to take the number of code points as zero.

I'm just noting this in case the editors want to mull it over.

spec.html
```diff
@@ -4488,53 +4488,64 @@ <h2>Syntax</h2>
...
+            1. Return RoundMVResult((_a_ + (_b_ &times; 10<sup>-_n_</sup>)) &times; 10<sup>_e_</sup>).
+          </emu-alg>
+          <emu-grammar>StrUnsignedDecimalLiteral ::: `.` DecimalDigits ExponentPart?</emu-grammar>
+          <emu-alg>
+            1. Let _a_ be MV of |DecimalDigits|.
+            1. If |ExponentPart| is present, let _e_ be MV of |ExponentPart|. Otherwise, let _e_ be 0.
+            1. Let _n_ be the number of code points in |DecimalDigits|.
+            1. Return RoundMVResult(_a_ &times; 10<sup>_e_ - _n_</sup>).
```
jmdyck, line 4529: It might be better to rename `_a_` to `_b_` in this algorithm, to echo the corresponding alias in the previous algorithm.

spec.html
jmdyck: I think there's a fairly strong tradition that _n_, as a variable name, denotes an integer, so it might be better to better to rename `_n_` to `_x_`, say.

jmdyck (commented):
I believe this PR is now semantics-preserving.

jmdyck:
(Hrmph. GitHub took my comment about `_n_` as continuing a resolved conversation, rather than starting a new one. I don't see any way to fix it after the fact.)

spec.html
```diff
@@ -24750,11 +24751,7 @@ <h1>parseFloat ( _string_ )</h1>
         1. Let _trimmedString_ be ! TrimString(_inputString_, ~start~).
         1. If neither _trimmedString_ nor any prefix of _trimmedString_ satisfies the syntax of a |StrDecimalLiteral| (see <emu-xref href="#sec-tonumber-applied-to-the-string-type"></emu-xref>), return *NaN*.
         1. Let _numberString_ be the longest prefix of _trimmedString_, which might be _trimmedString_ itself, that satisfies the syntax of a |StrDecimalLiteral|.
-        1. Let _mathFloat_ be MV of _numberString_.
-        1. If _mathFloat_ = 0, then
-          1. If the first code unit of _trimmedString_ is the code unit 0x002D (HYPHEN-MINUS), return *-0*<sub>𝔽</sub>.
-          1. Return *+0*<sub>𝔽</sub>.
-        1. Return 𝔽(_mathFloat_).
+        1. Return StringNumericValue of _numberString_.
```
bakkot, line 24754: I don't think this is coherent, though I suppose it wasn't before either: StringNumericValue is an SDO and hence is defined over parse nodes, not over strings.

It's repairable by doing

```suggestion
        1. Let _parsedNumber_ be ParseText(! StringToCodePoints(_numberString_), |StrDecimalLiteral|).
        1. Assert: _parsedNumber_ is a Parse Node.
        1. Return StringNumericValue of _parsedNumber_.
```

spec.html
bakkot: I would prefer to be more explicit, as in

```suggestion
            1. If the second |DecimalDigits| is present, let _n_ be the number of code points in the second |DecimalDigits|. Otherwise, let _n_ be 0.
```

Also I'd move it up a line, at which point it could be combined with what would then be the previous line, as in

```
1. If the second |DecimalDigits| is present, then
  1. Let _b_ be MV of the second |DecimalDigits|.
  1. Let _n_ be the number of code points in the second |DecimalDigits|.
1. Else,
  1. Let _b_ be 0.
  1. Let _n_ be 0.
```

bakkot (approved):
This LGTM other than the outstanding comments.

spec.html
```diff
@@ -14659,7 +14664,7 @@ <h2>Syntax</h2>
...
         <h1>Static Semantics: TV and TRV</h1>
-        <p>A template literal component is interpreted as a sequence of Unicode code points. The Template Value (TV) of a literal component is described in terms of String values (SV, <emu-xref href="#sec-literals-string-literals"></emu-xref>) contributed by the various parts of the template literal component. As part of this process, some Unicode code points within the template component are interpreted as having a mathematical value (MV, <emu-xref href="#sec-literals-numeric-literals"></emu-xref>). In determining a TV, escape sequences are replaced by the UTF-16 code unit(s) of the Unicode code point represented by the escape sequence. The Template Raw Value (TRV) is similar to a Template Value with the difference that in TRVs escape sequences are interpreted literally.</p>
+        <p>A template literal component is interpreted by TV and TRV as a value of the String type. In TV, escape sequences are replaced by the UTF-16 code unit(s) of the Unicode code point represented by the escape sequence. TRV is similar to TV with the difference being that in TRV, escape sequences are interpreted as they appear in the literal.</p>
```
bakkot, line 14667: I am [still not convinced](https://github.com/tc39/ecma262/pull/2435#discussion_r657526179) it makes sense to drop the expansions of TV and TRV, at least not in this PR.

michaelficarra: I was just in the middle of splitting it out, as we had discussed at the last editor call. See #2451.

michaelficarra:
This is ready for review again.

ljharb:
@michaelficarra should i wait to merge this until there's a stamp after your latest comment, or is it good to go now?

bakkot:
@ljharb It's good to go.

PR #2439, opened 2021-06-20
Editorial: use abstract closures for anonymous functions in more places

### LINKED ISSUE #933: Better closures, or promise-handling, in the spec

### LINKED ISSUE #1077: Missing definitions for "<thing> equivalent to a function that <action>"
Added in #1073

### LINKED ISSUE #1894: use abstract closures for CreateBuiltInFunction

bakkot:
Followup to #2109. Fixes #1894, fixes #933, fixes #1077 (in passing, in the "less handwaving" commit).

There's still six places where I haven't adopted the new style: two ([x](https://tc39.es/ecma262/multipage/#sec-promise-reject-functions), [x](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise-resolve-functions)) in [CreateResolvingFunctions](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-createresolvingfunctions) (used mainly in the Promise constructor), [one](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise.all-resolve-element-functions) in [PerformPromiseAll](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-performpromiseall), two ([x](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise.allsettled-resolve-element-functions), [x](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise.allsettled-reject-element-functions)) in [PerformPromiseAllSettled](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-performpromiseallsettled), and [one](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise.any-reject-element-functions) in [PerformPromiseAny](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-performpromiseany).

These remaining cases generally seemed nontrivial enough that I didn't think it worth inlining them. I could, I suppose, replace each with an abstract operation which returns the appropriate abstract closure (or built-in function), but that doesn't seem like it would be much of an improvement.

This is mostly a mechanical transformation. Where there was non-mutable state provided by setting internal slots, I have instead had the abstract closure capture the relevant aliases. I've left mutable state as internal slots, since abstract closures can't write to captured aliases.

One thing of note is that the Default Constructor Functions took a rest parameter, which isn't something we have established syntax for with abstract closures; I decided to mostly mimic what we do in [the function constructor](https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-function-p1-p2-pn-body) and refer to "the List of arguments that was passed to this function by [[Call]] or [[Construct]]" explicitly, rather than introducing syntax for rest parameters in Abstract Closures.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38339,17 +38339,6 @@ THH:mm:ss.sss
             1. Return ! AsyncFromSyncIteratorContinuation(_result_, _promiseCapability_).
           </emu-alg>
         </emu-clause>
-
-        <emu-clause id="sec-async-from-sync-iterator-value-unwrap-functions">
-          <h1>Async-from-Sync Iterator Value Unwrap Functions</h1>
-          <p>An async-from-sync iterator value unwrap function is an anonymous built-in function that is used by AsyncFromSyncIteratorContinuation when processing the *"value"* property of an <i>IteratorResult</i> object, in order to wait for its value if it is a promise and re-package the result in a new "unwrapped" <i>IteratorResult</i> object. Each async-from-sync iterator value unwrap function has a [[Done]] internal slot.</p>
-          <p>When an async-from-sync iterator value unwrap function is called with argument _value_, the following steps are taken:</p>
-
-          <emu-alg>
-            1. Let _F_ be the active function object.
-            1. Return ! CreateIterResultObject(_value_, _F_.[[Done]]).
-          </emu-alg>
-        </emu-clause>
       </emu-clause>
 
       <emu-clause id="sec-properties-of-async-from-sync-iterator-instances">
@@ -38381,7 +38370,7 @@ THH:mm:ss.sss
         </emu-table>
       </emu-clause>
 
-      <emu-clause id="sec-asyncfromsynciteratorcontinuation" aoid="AsyncFromSyncIteratorContinuation">
+      <emu-clause id="sec-asyncfromsynciteratorcontinuation" oldids="sec-async-from-sync-iterator-value-unwrap-functions" aoid="AsyncFromSyncIteratorContinuation">
         <h1>AsyncFromSyncIteratorContinuation ( _result_, _promiseCapability_ )</h1>
         <p>The abstract operation AsyncFromSyncIteratorContinuation takes arguments _result_ and _promiseCapability_ (a PromiseCapability Record). It performs the following steps when called:</p>
 
@@ -38392,10 +38381,10 @@ THH:mm:ss.sss
           1. IfAbruptRejectPromise(_value_, _promiseCapability_).
           1. Let _valueWrapper_ be PromiseResolve(%Promise%, _value_).
           1. IfAbruptRejectPromise(_valueWrapper_, _promiseCapability_).
-          1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-async-from-sync-iterator-value-unwrap-functions" title></emu-xref>.
-          1. Let _length_ be the number of non-optional parameters of the function definition in <emu-xref href="#sec-async-from-sync-iterator-value-unwrap-functions" title></emu-xref>.
-          1. Let _onFulfilled_ be ! CreateBuiltinFunction(_steps_, _length_, *""*, &laquo; [[Done]] &raquo;).
-          1. Set _onFulfilled_.[[Done]] to _done_.
+          1. Let _unwrap_ be a new Abstract Closure with parameters (_value_) that captures _done_ and performs the following steps when called:
+            1. Return ! CreateIterResultObject(_value_, _done_).
+          1. Let _onFulfilled_ be ! CreateBuiltinFunction(_unwrap_, 1, *""*, &laquo; &raquo;).
+          1. NOTE: _onFulfilled_ is used when processing the *"value"* property of an IteratorResult object in order to wait for its value if it is a promise and re-package the result in a new "unwrapped" IteratorResult object.
           1. Perform ! PerformPromiseThen(_valueWrapper_, _onFulfilled_, *undefined*, _promiseCapability_).
           1. Return _promiseCapability_.[[Promise]].
         </emu-alg>
--- a/spec.html
+++ b/spec.html
@@ -38619,17 +38619,20 @@ THH:mm:ss.sss
         </emu-alg>
       </emu-clause>
 
-      <emu-clause id="sec-newpromisecapability" aoid="NewPromiseCapability">
+      <emu-clause id="sec-newpromisecapability" oldids="sec-getcapabilitiesexecutor-functions" aoid="NewPromiseCapability">
         <h1>NewPromiseCapability ( _C_ )</h1>
         <p>The abstract operation NewPromiseCapability takes argument _C_. It attempts to use _C_ as a constructor in the fashion of the built-in Promise constructor to create a Promise object and extract its `resolve` and `reject` functions. The Promise object plus the `resolve` and `reject` functions are used to initialize a new PromiseCapability Record. It performs the following steps when called:</p>
         <emu-alg>
           1. If IsConstructor(_C_) is *false*, throw a *TypeError* exception.
           1. NOTE: _C_ is assumed to be a constructor function that supports the parameter conventions of the Promise constructor (see <emu-xref href="#sec-promise-executor"></emu-xref>).
           1. Let _promiseCapability_ be the PromiseCapability Record { [[Promise]]: *undefined*, [[Resolve]]: *undefined*, [[Reject]]: *undefined* }.
-          1. Let _steps_ be the algorithm steps defined in <emu-xref href="#sec-getcapabilitiesexecutor-functions" title></emu-xref>.
-          1. Let _length_ be the number of non-optional parameters of the function definition in <emu-xref href="#sec-getcapabilitiesexecutor-functions" title></emu-xref>.
-          1. Let _executor_ be ! CreateBuiltinFunction(_steps_, _length_, *""*, &laquo; [[Capability]] &raquo;).
-          1. Set _executor_.[[Capability]] to _promiseCapability_.
+          1. Let _executorClosure_ be a new Abstract Closure with parameters (_resolve_, _reject_) that captures _promiseCapability_ and performs the following steps when called:
+            1. If _promiseCapability_.[[Resolve]] is not *undefined*, throw a *TypeError* exception.
+            1. If _promiseCapability_.[[Reject]] is not *undefined*, throw a *TypeError* exception.
+            1. Set _promiseCapability_.[[Resolve]] to _resolve_.
+            1. Set _promiseCapability_.[[Reject]] to _reject_.
+            1. Return *undefined*.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -39308,54 +39224,24 @@ <h1>Promise.prototype.finally ( _onFinally_ )</h1>
...
+              1. Return ? Invoke(_promise_, *"then"*, &laquo; _valueThunk_ &raquo;).
+            1. Let _thenFinally_ be ! CreateBuiltinFunction(_thenFinallyClosure_, 1, *""*, &laquo; &raquo;).
+            1. Let _catchFinallyClosure_ be a new Abstract Closure with parameters (_reason_) that captures _onFinally_ and _C_ and performs the following steps when called:
+              1. Let _result_ be ? Call(_onFinally_, *undefined*).
+              1. Let _promise_ be ? PromiseResolve(_C_, _result_).
+              1. Let _throwReason_ be a new Abstract Closure with no parameters that captures _reason_ and performs the following steps when called:
+                1. Return ThrowCompletion(_reason_).
+              1. Let _thrower_ be ! CreateBuiltinFunction(_throwReason_, 0, *""*, &laquo; &raquo;).
+              1. Return ? Invoke(_promise_, *"then"*, &laquo; _thrower_ &raquo;).
+            1. Let _catchFinally_ be ! CreateBuiltinFunction(_catchFinallyClosure_, 1, *""*, &laquo; &raquo;).
```
ljharb, line 39242: fwiw, confirmed these exact changes pass in https://npmjs.com/promise.prototype.finally

ljharb (commented):
LGTM

spec.html
```diff
@@ -3562,14 +3562,22 @@ <h1>Await</h1>
...
+            1. Return *undefined*.
+          1. Let _onFulfilled_ be ! CreateBuiltinFunction(_fulfilledClosure_, 1, *""*, &laquo; &raquo;).
+          1. Let _rejectedClosure_ be a new Abstract Closure with parameters (_reason_) that captures _asyncContext_ and performs the following steps when called:
+            1. Let _prevContext_ be the running execution context.
+            1. Suspend _prevContext_.
+            1. Push _asyncContext_ onto the execution context stack; _asyncContext_ is now the running execution context.
+            1. Resume the suspended evaluation of _asyncContext_ using ThrowCompletion(_reason_) as the result of the operation that suspended it.
+            1. Assert: When we reach this step, _asyncContext_ has already been removed from the execution context stack and _prevContext_ is the currently running execution context.
+            1. Return *undefined*.
+          1. Let _onRejected_ be ! CreateBuiltinFunction(_rejectedClosure_, 1, *""*, &laquo; [[AsyncContext]] &raquo;).
```
michaelficarra, line 3580:
```suggestion
          1. Let _onRejected_ be ! CreateBuiltinFunction(_rejectedClosure_, 1, *""*, &laquo; &raquo;).
```

spec.html
```diff
@@ -21514,8 +21462,19 @@ <h1>Runtime Semantics: ClassDefinitionEvaluation</h1>
...
+            1. Let _F_ be the active function object.
+            1. If _F_.[[ConstructorKind]] is ~derived~, then
+              1. NOTE: This branch behaves similarly to `constructor(...args) { super(...args); }`. The most notable distinction is that while the aforementioned ECMAScript source text observably calls the @@iterator method on `%Array.prototype%`, this function does not.
+              1. Let _func_ be ! _F_.[[GetPrototypeOf]]().
+              1. If IsConstructor(_func_) is *false*, throw a *TypeError* exception.
+              1. Return ? Construct(_func_, _args_, NewTarget).
+            1. Else,
+              1. NOTE: This branch behaves similarly to `constructor() {}`.
+              1. Return ? OrdinaryCreateFromConstructor(NewTarget, *"%Object.prototype%"*).
+          1. Let _F_ be ! CreateBuiltinFunction(_defaultConstructor_, 0, _className_, &laquo; [[ConstructorKind]], [[SourceText]] &raquo;, the current Realm Record, _constructorParent_).
```
michaelficarra, line 21477:
```suggestion
          1. Let _F_ be ! CreateBuiltinFunction(_defaultConstructor_, 0, _className_, &laquo; [[ConstructorKind]] &raquo;, the current Realm Record, _constructorParent_).
```

bakkot: This PR doesn't interact with the `[[SourceText]]` slot at all. It should remain. (It's used to [store the source of the class itself](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#sec-runtime-semantics-bindingclassdeclarationevaluation).)

michaelficarra (approved):
LGTM otherwise.

spec.html
```diff
@@ -40586,27 +40444,12 @@ <h1>Properties of the Proxy Constructor</h1>
...
-          1. Return _result_.
-        </emu-alg>
-
-        <emu-clause id="sec-proxy-revocation-functions">
-          <h1>Proxy Revocation Functions</h1>
-          <p>A Proxy revocation function is an anonymous built-in function that has the ability to invalidate a specific Proxy object.</p>
-          <p>Each Proxy revocation function has a [[RevocableProxy]] internal slot.</p>
-          <p>When a Proxy revocation function is called, the following steps are taken:</p>
-          <emu-alg>
+          1. Let _revokerClosure_ be a new Abstract Closure with no parameters that captures nothing and performs the following steps when called:
```
syg, line 40452: Why can't this capture _p_? The closure doesn't need to change _p_, only mutates its slots.

syg (approved):
lgtm modulo the Proxy revoke comment.

spec.html
bakkot: It does: it sets _p_ to `null` once the proxy has been revoked, so it can bail early if called twice, in `d. Set F.[[RevocableProxy]] to null.`

I suppose since this operation is idempotent we could just do away with the early bailing, but I was trying to stick with the existing logic as closely as possible.

syg: Oh wow, this terrible GH review UI! When I expanded the unchanged lines to compare, it skipped the line that sets it to null! Notice the missing lines in the attached screenshot.
Screen Shot 2021-06-23 at 6 54 16 PM

syg: Anyways, then everything lgtm.

bakkot: Yup, for large documents that button silently fails in a way almost indistinguishable from success. And it has done for literally years with no sign they intend to fix it are even aware it's an issue. /shrug

spec.html
bakkot: Done.

PR #2441, opened 2021-06-23
Editorial: disambiguate `with` statement's status

ddbeck:
Currently, the spec sends mixed signals to programmers about the `with` statement; this PR attempts to clear up the ambiguity by adding text that explicitly discourages the use of the `with` statement in new ECMAScript code.

This is an attempt to solve a specific problem: developers and technical writers are unclear about the status of the `with` statement. See [this StackOverflow question](https://stackoverflow.com/questions/48810592/is-the-javascript-keyword-with-really-deprecated) or [this discussion in MDN Web Docs's browser-compat-data project](https://github.com/mdn/browser-compat-data/issues/10490) to get a taste. This makes it difficult to document on MDN Web Docs and elsewhere.

By forbidding `with` in strict mode code and cutting it off from many contemporary language features, the specification seems to imply that programmers shouldn't use `with`, but this is less explicit than other disfavored parts of the language (e.g., [the note for the `getYear` method](https://tc39.es/ecma262/#sec-date.prototype.getyear) or [the introduction to Annex B](https://tc39.es/ecma262/#sec-additional-ecmascript-features-for-web-browsers)). Because there's no single way that the spec indicates non-preferred features, developers and technical writers have to make do with close readings and interpretations.

Ideally, the specification could resolve the question definitively. This PR's fix to the problem is text that speaks directly to programmers, discouraging them from using the `with` statement. This is my attempt to put into words what I believe the existing text implies to programmers, but I'm open to making revisions which more accurately captures the posture toward `with` statement.

Thanks for considering this change!

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -21933,8 +21933,12 @@
     </emu-clause>
   </emu-clause>
 
-  <emu-clause id="sec-with-statement">
+  <emu-clause id="sec-with-statement" legacy>
     <h1>The `with` Statement</h1>
+    <emu-note>
+      <p>Use of the Legacy `with` statement is discouraged in new ECMAScript code. Consider alternatives that are permitted in both strict mode code and non-strict code, such as <emu-xref href="#sec-destructuring-assignment">destructuring assignment</emu-xref>.</p>
+    </emu-note>
+
     <h2>Syntax</h2>
     <emu-grammar type="definition">
       WithStatement[Yield, Await, Return] :


rolivo3:
The message must be very clear if this characteristic is going to be deprecated or not. One thing is for it to be discouraged and the other more traumatic for it to be deprecated.

If you plan to deprecate "with statement" you should consider the huge number of sites that use this feature and are going to be broken. On the other hand, the credibility that the specification has when a proposal reaches stage 4 of the process for most developers (I include myself) we consider it stable, it is traumatic and painful to think that a feature that has reached stage 4 is eliminated. Give some thought to the developers.

michaelficarra:
#2125 will be defining the term "legacy" for this purpose. However, applying the term "legacy" to a feature requires committee consensus, so this will have to be marked normative, not editorial.

ljharb:
It seems like this might be better marked as "Legacy", once #2125 lands.

ljharb:
@michaelficarra i don't think it's normative just because it requires consensus.

WebReflection:
There are no substitutes for the `with` statement, and the spec has `Symbol.unscopables` to deal with those cases too.

Discouraging with reasons looks like a good way forward, but "*deprecating*" it makes little sense.

I hope the `obj.{ ... };` syntax, one day, will be back, otherwise we are all in `Function('...')` hands to bring back what `with` can do, as nothing else can.

ljharb:
deprecation (in a context where basically nothing can be removed) is the same as discouragement.

ddbeck:
Thanks for the comments here. I wasn't previously aware of #2125. "Legacy" seems like a great way to handle this sort of thing. When #2125 is merged, I'll be happy to update this PR to use the `legacy` attribute, though I'm not certain if my note should be retained or not.

> applying the term "legacy" to a feature requires committee consensus

I'm new to contributing to the specification and the contributing docs don't seem to cover a case like this. What, if anything, should I do make it easier for the committee to consider this for consensus?

michaelficarra:
@ddbeck The typical process would be to find a champion, which is a representative of a TC39 member organisation who will present your topic to committee during one of the regular meetings. If you update the language here to use the term "legacy", as defined in #2125, I will present this topic as a needs-consensus PR for you at the next meeting. The case for `with` is not as obvious as the already agreed-upon applications of "legacy", but I do personally think it merits the term.

ddbeck:
@michaelficarra Excellent, thank you. I'd be delighted if you and the rest of the committee to take this up at the next meeting. I've updated the PR to use the term "legacy" and to set the legacy attribute for the clause (though this shouldn't really do anything until the other PR is merged). As soon as the other PR is merged, I'll rebase on top of that, too. Thank you again!

ddbeck:
Since #2125 merged last week, I rebased my changes against the default branch and confirmed that this builds properly with the `legacy` attribute (since, as a first-time contributor, the GitHub workflows don't run).

I'll be happy to make any additional changes requested, as this proposal works its way through the process. Thanks again!

michaelficarra:
@ddbeck This was presented to committee today (by @ljharb) and achieved consensus. 🎉

spec.html
```diff
@@ -21828,8 +21828,12 @@ <h1>Runtime Semantics: Evaluation</h1>
     </emu-clause>
   </emu-clause>
 
-  <emu-clause id="sec-with-statement">
+  <emu-clause id="sec-with-statement" legacy>
     <h1>The `with` Statement</h1>
+    <emu-note>
+      <p>It is recommended that the legacy `with` statement is not used in new ECMAScript code. Consider alternatives, such as <emu-xref href="#sec-destructuring-assignment" title></emu-xref>, which are permitted in both strict mode code and non-strict code.</p>
```
michaelficarra, line 21834:
```suggestion
      <p>It is recommended that the Legacy `with` statement is not used in new ECMAScript code. Consider alternatives, such as <emu-xref href="#sec-destructuring-assignment" title></emu-xref>, which are permitted in both strict mode code and non-strict code.</p>
```

We need to use a capital L for Legacy now for it to auto-link.

michaelficarra:
```suggestion
      <p>It is recommended that the legacy `with` statement is not used in new ECMAScript code. Consider alternatives, such as <emu-xref href="#sec-destructuring-assignment" title></emu-xref>, which is permitted in both strict mode code and non-strict code.</p>
```

ddbeck: @michaelficarra Thank you for the suggestions and letting me know the good news!

I've fixed the auto-link Legacy and removed the unneeded `title`. 👍

Instead of directly applying the suggested grammar change, I slightly reworded the sentence to avoid ambiguity in numerical agreement (i.e., "Consider alternatives … which are…" is both right and wrong, depending on whether we read the "such as" clause as subordinate). The sentence now reads:

> Consider alternatives that are permitted in both strict mode code and non-strict code, such as…"

spec.html
```diff
@@ -21855,8 +21855,12 @@ <h1>Runtime Semantics: Evaluation</h1>
     </emu-clause>
   </emu-clause>
 
-  <emu-clause id="sec-with-statement">
+  <emu-clause id="sec-with-statement" legacy>
     <h1>The `with` Statement</h1>
+    <emu-note>
+      <p>It is recommended that the Legacy `with` statement is not used in new ECMAScript code. Consider alternatives that are permitted in both strict mode code and non-strict code, such as <emu-xref href="#sec-destructuring-assignment"></emu-xref>.</p>
```
sideshowbarker, line 21861:
```suggestion
      <p>Use of the Legacy `with` statement is discouraged in new ECMAScript code. Consider alternatives that are permitted in both strict mode code and non-strict code, such as <emu-xref href="#sec-destructuring-assignment"></emu-xref>.</p>            
```

…because:

* I think *discouraged* is more forceful than “not recommended”.
* I think it would be good for the spec to use consistent language for features that we want to warn developers away from. And I in general believe for that the term “discouraged” is an apt choice, for reasons outlined in https://github.com/mdn/content/discussions/5549#discussioncomment-814689

sideshowbarker: OK, only after writing the above did it dawn on me that _Legacy_ is the term y’all have already landed on to flag these kinds of cases. I anyway still prefer getting “discouraged” in there rather than just “not recommended”. But I guess as long as y’all have already arrived at _Legacy_ as the flag to consistently use for these cases, I don’t feel that strongly about the ancillary wording

ddbeck: I'd be happy to strengthen the wording here. For what it's worth, I borrowed the "It is recommended…" style from elsewhere in the spec. Left wholly to my own devices, I'd probably go with something more active (and assertive), such as:

> Do not use the Legacy `with` statement in new ECMAScript code. Consider alternatives….

michaelficarra (approved):

spec.html
michaelficarra: In the editor call today, we decided to adopt @sideshowbarker's suggested wording. I pushed a commit that updates it.

bakkot:
We should put `destructuring assignment` inside of the xref element, so that it's not just a section number.

(Github isn't letting me comment directly on the PR for some reason.)

michaelficarra:
@bakkot done

bakkot (approved):

ljharb:
@ddbeck would you mind signing the [IPR form](https://tc39.es/agreements/contributor/)? I think you may have been employed at Mozilla at the time of this PR, but it'd help some of our automation if you filled it out anyways :-)

ddbeck:
@ljharb signed! ✅

PR #2442, opened 2021-06-24
Normative: Constrain all host hooks to return either normal or throw completions

syg:
This supersedes #2433.

This PR also makes the prose around requirements for host hook
implementations consistent to the template "An implementation of
HostHook must conform to the following requirements:".

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -169,7 +169,10 @@
     <p>An <dfn id="implementation-defined">implementation-defined</dfn> facility is one that defers its definition to an external source without further qualification. This specification does not make any recommendations for particular behaviours, and conforming implementations are free to choose any behaviour within the constraints put forth by this specification.</p>
     <p>An <dfn id="implementation-approximated">implementation-approximated</dfn> facility is one that defers its definition to an external source while recommending an ideal behaviour. While conforming implementations are free to choose any behaviour within the constraints put forth by this specification, they are encouraged to strive to approximate the ideal. Some mathematical operations, such as <emu-xref href="#sec-math.exp"><code>Math.exp</code></emu-xref>, are implementation-approximated.</p>
     <p>A <dfn id="host">host</dfn> is an external source that further defines facilities listed in Annex <emu-xref href="#sec-host-layering-points"></emu-xref> but does not further define other implementation-defined or implementation-approximated facilities. In informal use, a host refers to the set of all implementations, such as the set of all web browsers, that interface with this specification in the same way via Annex <emu-xref href="#sec-host-layering-points"></emu-xref>. A host is often an external specification, such as WHATWG HTML (<a href="https://html.spec.whatwg.org/">https://html.spec.whatwg.org/</a>). In other words, facilities that are host-defined are often further defined in external specifications.</p>
-    <p>A <dfn id="host-hook">host hook</dfn> is an abstract operation that is defined in whole or in part by an external source. All host hooks must be listed in Annex <emu-xref href="#sec-host-layering-points"></emu-xref>.</p>
+    <p>A <dfn id="host-hook">host hook</dfn> is an abstract operation that is defined in whole or in part by an external source. All host hooks must be listed in Annex <emu-xref href="#sec-host-layering-points"></emu-xref>. A host hook must conform to at least the following requirements:</p>
+    <ul>
+      <li>It must return either a normal completion or a throw completion.</li>
+    </ul>
     <p>A <dfn id="host-defined">host-defined</dfn> facility is one that defers its definition to an external source without further qualification and is listed in Annex <emu-xref href="#sec-host-layering-points"></emu-xref>. Implementations that are not hosts may also provide definitions for host-defined facilities.</p>
     <p>A <dfn id="host-environment">host environment</dfn> is a particular choice of definition for all host-defined facilities. A host environment typically includes objects or functions which allow obtaining input and providing output as host-defined properties of the global object.</p>
     <p>This specification follows the editorial convention of always using the most specific term. For example, if a facility is host-defined, it should not be referred to as implementation-defined.</p>
@@ -3835,7 +3838,15 @@
           </tbody>
         </table>
       </emu-table>
-      <p>The term &ldquo;<dfn>abrupt completion</dfn>&rdquo; refers to any completion with a [[Type]] value other than ~normal~.</p>
+      <p>The following shorthand terms are sometimes used to refer to completions.</p>
+      <ul>
+        <li><dfn>normal completion</dfn> refers to any completion with a [[Type]] value of ~normal~.</li>
+        <li><dfn>break completion</dfn> refers to any completion with a [[Type]] value of ~break~.</li>
+        <li><dfn>continue completion</dfn> refers to any completion with a [[Type]] value of ~continue~.</li>
+        <li><dfn>return completion</dfn> refers to any completion with a [[Type]] value of ~return~.</li>
+        <li><dfn>throw completion</dfn> refers to any completion with a [[Type]] value of ~throw~.</li>
+        <li><dfn>abrupt completion</dfn> refers to any completion with a [[Type]] value other than ~normal~.</li>
+      </ul>
       <p>Callable objects that are defined in this specification only return a normal completion or a throw completion. Returning any other kind of completion is considered an editorial error.</p>
       <p>Implementation-defined callable objects must return either a normal completion or a throw completion.</p>
 
@@ -11662,10 +11673,9 @@
       </h1>
       <dl class="header">
       </dl>
-      <p>The implementation of HostMakeJobCallback must conform to the following requirements:</p>
+      <p>An implementation of HostMakeJobCallback must conform to the following requirements:</p>
       <ul>
-        <li>It must always complete normally (i.e., not return an abrupt completion).</li>
-        <li>It must always return a JobCallback Record whose [[Callback]] field is _callback_.</li>
+        <li>It must complete normally with a JobCallback Record whose [[Callback]] field is _callback_.</li>
       </ul>
       <p>The default implementation of HostMakeJobCallback performs the following steps when called:</p>
       <emu-alg>
@@ -11688,9 +11698,9 @@
       </h1>
       <dl class="header">
       </dl>
-      <p>The implementation of HostCallJobCallback must conform to the following requirements:</p>
+      <p>An implementation of HostCallJobCallback must conform to the following requirements:</p>
       <ul>
-        <li>It must always perform and return the result of Call(_jobCallback_.[[Callback]], _V_, _argumentsList_).</li>
+        <li>It must perform and return the result of Call(_jobCallback_.[[Callback]], _V_, _argumentsList_).</li>
       </ul>
       <emu-note>
         <p>This requirement means that hosts cannot change the [[Call]] behaviour of function objects defined in this specification.</p>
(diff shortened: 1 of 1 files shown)


syg:
cc @jugglinmike

spec.html
```diff
@@ -24617,7 +24619,11 @@ <h1>PerformEval ( _x_, _callerRealm_, _strictCaller_, _direct_ )</h1>
       <emu-clause id="sec-hostensurecancompilestrings" aoid="HostEnsureCanCompileStrings">
         <h1>HostEnsureCanCompileStrings ( _callerRealm_, _calleeRealm_ )</h1>
         <p>The host-defined abstract operation HostEnsureCanCompileStrings takes arguments _callerRealm_ (a Realm Record) and _calleeRealm_ (a Realm Record). It allows host environments to block certain ECMAScript functions which allow developers to compile strings into ECMAScript code.</p>
-        <p>An implementation of HostEnsureCanCompileStrings may complete normally or abruptly. Any abrupt completions will be propagated to its callers. The default implementation of HostEnsureCanCompileStrings is to unconditionally return an empty normal completion.</p>
+        <p>An implementation of HostEnsureCanCompileStrings must conform to the following requirements:</p>
+        <ul>
+          <li>It must propagate throw completions to its callers.</li>
+        </ul>
+        <p>The default implementation of HostEnsureCanCompileStrings is to unconditionally return NormalCompletion(~empty~).</p>
```
jugglinmike, line 24626: The word "unconditionally" isn't used consistently for the return values modified in this patch. I believe it's included merely for emphasis, but I wonder if consistency would be valuable.

spec.html
```diff
@@ -16638,9 +16640,9 @@ <h1>HostFinalizeImportMeta ( _importMeta_, _moduleRecord_ )</h1>
 
           <p>Most hosts will be able to simply define HostGetImportMetaProperties, and leave HostFinalizeImportMeta with its default behaviour. However, HostFinalizeImportMeta provides an "escape hatch" for hosts which need to directly manipulate the object before it is exposed to ECMAScript code.</p>
 
-          <p>The implementation of HostFinalizeImportMeta must conform to the following requirements:</p>
+          <p>An implementation of HostFinalizeImportMeta must conform to the following requirements:</p>
           <ul>
-            <li>It must always complete normally (i.e., not return an abrupt completion).</li>
+            <li>It must complete normally (i.e., not return an abrupt completion).</li>
```
jugglinmike, line 16645: I like this change probably more than I should.

spec.html
```diff
@@ -10543,7 +10545,8 @@ <h1>Host Hooks</h1>
 
       <emu-clause id="sec-host-cleanup-finalization-registry" aoid="HostEnqueueFinalizationRegistryCleanupJob">
         <h1>HostEnqueueFinalizationRegistryCleanupJob ( _finalizationRegistry_ )</h1>
-        <p>The abstract operation HostEnqueueFinalizationRegistryCleanupJob takes argument _finalizationRegistry_ (a FinalizationRegistry). HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that is expected to call CleanupFinalizationRegistry(_finalizationRegistry_) at some point in the future, if possible. The host's responsibility is to make this call at a time which does not interrupt synchronous ECMAScript code execution.</p>
+        <p>The abstract operation HostEnqueueFinalizationRegistryCleanupJob takes argument _finalizationRegistry_ (a FinalizationRegistry). HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that is expected to call CleanupFinalizationRegistry(_finalizationRegistry_) at some point in the future, if possible.</p>
+        <p>An implementation of HostEnqueueFinalizationRegistryCleanupJob must conform to the requirements in <emu-xref href="#sec-jobs"></emu-xref>.</p>
```
jugglinmike, line 10549: > The host's responsibility is to make this call at a time which does not interrupt synchronous ECMAScript code execution.

At first glance, I thought the new requirement on host hooks obviated this statement. On second thought, it doesn't seem like it applies to the behavior of an abstract operation invoked after the host hook has returned. Is there some other justification for removing the statement?

spec.html
```diff
@@ -16621,15 +16624,14 @@ <h1>Runtime Semantics: Evaluation</h1>
...
           <ul>
-            <li>It must return a List, whose values are all Records with two fields, [[Key]] and [[Value]].</li>
+            <li>It must return a normal completion of a List whose values are all Records with two fields, [[Key]] and [[Value]].</li>
             <li>Each such Record's [[Key]] field must be a property key, i.e., IsPropertyKey must return *true* when applied to it.</li>
             <li>Each such Record's [[Value]] field must be an ECMAScript value.</li>
-            <li>It must always complete normally (i.e., not return an abrupt completion).</li>
           </ul>
 
-          <p>The default implementation of HostGetImportMetaProperties is to return a new empty List.</p>
+          <p>The default implementation of HostGetImportMetaProperties is to return a normal completion of a new empty List.</p>
```
jugglinmike, line 16634: There doesn't appear to be precedence for the construction "a normal completion of __x__". There are some uses of "a normal completion with a value of __x__", though "NormalCompletion(x)" is more prevalent and more precise. What do you think about using one of those two?

spec.html
syg: Good idea, consistency would be valuable. I don't think we need to include "unconditionally". Any preference here?

spec.html
syg: The intention here was the "must not interrupt synchronous execution" is more precisely captured by the requirements in `#sec-jobs`, specifically the stuff around waiting for the execution context stack to be empty. No job queued by a job-queuing host defined AO in ecma262 is allowed to interrupt synchronous execution.

syg: Hm, a strict reading of the `#sec-jobs` restriction only applies to Job Abstract Closures, which HostEnqueueFinalizationRegistryCleanupJob explicitly spells out. I'll spell that out so it's clear that it applies.

spec.html
syg: I'd like to use the NormalCompletion(x) form, but thought NormalCompletion(a new empty List) would look weird. But now that I've typed that out it's probably fine.

ljharb: What about `NormalCompletion(« »)`?

syg: > What about NormalCompletion(« »)

Great idea, will use that.

spec.html
jugglinmike: Omitting it is my preference, too

spec.html
```diff
@@ -10543,7 +10545,14 @@ <h1>Host Hooks</h1>
...
+        <p>The abstract operation HostEnqueueFinalizationRegistryCleanupJob takes argument _finalizationRegistry_ (a FinalizationRegistry). Let _cleanupJob_ be a new Job Abstract Closure with no parameters that captures _finalizationRegistry_ and performs the following steps when called:</p>
+        <emu-alg>
+          1. Let _cleanupResult_ be CleanupFinalizationRegistry(_finalizationRegistry_).
+          1. If _cleanupResult_ is an abrupt completion, perform any host-defined steps for reporting the error.
+          1. Return NormalCompletion(~empty~).
+        </emu-alg>
+        <p>HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that schedules to be performed at some future time, if possible.</p>
```
jugglinmike, line 10554:
```suggestion
        <p>HostEnqueueFinalizationRegistryCleanupJob is an implementation-defined abstract operation that schedules _cleanupJob_ to be performed at some future time, if possible.</p>
```

syg: Oops, thanks for catching this.

jugglinmike (approved):

spec.html
```diff
@@ -170,7 +170,10 @@ <h1>Hosts and Implementations</h1>
...
-    <p>A <dfn id="host-hook">host hook</dfn> is an abstract operation that is defined in whole or in part by an external source. All host hooks must be listed in Annex <emu-xref href="#sec-host-layering-points"></emu-xref>.</p>
+    <p>A <dfn id="host-hook">host hook</dfn> is an abstract operation that is defined in whole or in part by an external source. All host hooks must be listed in Annex <emu-xref href="#sec-host-layering-points"></emu-xref>. A host hook must conform to at least the following requirements:</p>
+    <ul>
+      <li>It must return either a normal completion or a throw completion.</li>
```
michaelficarra, line 175: Although it's currently used twice, I don't think we've defined the term "throw completion". We should probably address those existing uses similarly to my suggestion below:

```suggestion
      <li>It must return a completion with a [[Type]] value of ~normal~ or ~throw~.</li>
```

This is the same form used in #2448. Perhaps we want to define a term "callable completion" or something that can be used in all of these cases?

spec.html
```diff
@@ -23779,10 +23787,10 @@ <h1>HostResolveImportedModule ( _referencingScriptOrModule_, _specifier_ )</h1>
           <p>there will be no active script or module at the time the <emu-xref href="#sec-import-calls">`import()`</emu-xref> expression runs. More generally, this can happen in any situation where the host pushes execution contexts with *null* ScriptOrModule components onto the execution context stack.</p>
         </emu-note>
 
-        <p>The implementation of HostResolveImportedModule must conform to the following requirements:</p>
+        <p>An implementation of HostResolveImportedModule must conform to the following requirements:</p>
         <ul>
           <li>
-            The normal return value must be an instance of a concrete subclass of Module Record.
+            If it completes normally, it must return an instance of a concrete subclass of Module Record.
```
michaelficarra, line 23793:
```suggestion
            If it completes normally, the [[Value]] slot of the completion must contain an instance of a concrete subclass of Module Record.
```

spec.html
syg: I'd prefer not to unless we start referring to this disjunction in more places. Otherwise it's just another thing to look up when the full definition isn't very long.

syg: I'll define a generic shorthand, like "foo completion means a completion with a [[Type]] value of ~foo~".

spec.html
syg: Done.

spec.html
```diff
@@ -171,7 +171,7 @@ <h1>Hosts and Implementations</h1>
...
     <p>A <dfn id="host-hook">host hook</dfn> is an abstract operation that is defined in whole or in part by an external source. All host hooks must be listed in Annex <emu-xref href="#sec-host-layering-points"></emu-xref>. A host hook must conform to at least the following requirements:</p>
     <ul>
-      <li>It must return either a normal completion or a throw completion.</li>
+      <li>It must return a completion with a [[Type]] value of ~normal~ or ~throw~.</li>
```
michaelficarra, line 174: Now that you've defined the shorthands, you don't need to make this change.

syg: I made it to be consistent with #2448 as you pointed out. Any preference on going with this or staying with "normal or throw completion" for both?

michaelficarra: "normal completion or throw completion" for both seems best to me. It means your addition of the definitions of those terms will race. Maybe just split it out and we get that in now?

syg: Let's just sequence this one before the other -- meeting's in 2 days and we should get consensus for the other one by then.

bakkot (approved):

michaelficarra (approved):

ljharb:
@syg this will need a rebase before it's mergeable

syg:
Rebase against structured headers.

michaelficarra (approved):

PR #2451, opened 2021-07-07
Editorial: don't imply that SV, TV, and TRV are shorthands

michaelficarra:
This was split out from #2435. I took issue with the way the usage of parentheses here implied that we were defining terms that were interchangeable (for example, wherever you use TV you may replace it with Template Value, and vice versa). In the process of removing these, I believe I also simplified the descriptions of the SDOs to get more to the point with less clutter.


syg (approved):
lgtm, thanks!

bakkot (approved):

PR #2457, opened 2021-07-15
Editorial: Document parameter type for Date Abstract Operations

gibson042:
A bigger refactoring of the Date AOs is still in order, but this will help comprehensibility in the meantime.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -31809,8 +31809,8 @@
       <emu-clause id="sec-local-time-zone-adjustment" type="implementation-defined abstract operation">
         <h1>
           LocalTZA (
-            _t_: unknown,
-            _isUTC_: unknown,
+            _t_: a Number,
+            _isUTC_: a Boolean,
           )
         </h1>
         <dl class="header">
@@ -31833,7 +31833,7 @@
       <emu-clause id="sec-localtime" type="abstract operation">
         <h1>
           LocalTime (
-            _t_: unknown,
+            _t_: a time value,
           )
         </h1>
         <dl class="header">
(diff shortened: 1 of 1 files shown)


bakkot (approved):

WeslleyFratini (approved):

michaelficarra:
While we're at it, the `tv` parameter of `TimeString`, `DateString`, `TimeZoneString`, and `ToDateString` is a `Number`.

spec.html
```diff
@@ -30604,12 +30604,12 @@ <h1>
...
         <h1>
           UTC (
-            _t_: unknown,
+            _t_: a Number,
           )
         </h1>
         <dl class="header">
           <dt>description</dt>
-          <dd>It converts _t_ from local time to UTC.</dd>
+          <dd>It converts _t_ from local time to a UTC time value.</dd>
```
michaelficarra, line 30612: I would prefer this wording to be more similar to the wording above:

> It converts time value _t_ from UTC to local time.

michaelficarra: What does "from local time" add here?

gibson042: It communicates the semantics of the value, albeit poorly (as noted in the PR description, a bigger refactoring of the Date AOs is still in order). But specifically, _t_ is a Number representing a local date and time of day equal to the UTC date and time of day that correspond with (mis)interpreting it as a time value and ignoring the bounds that normally apply.

spec.html
```diff
@@ -30585,12 +30585,12 @@ <h1>
...
         <h1>
           LocalTime (
-            _t_: unknown,
+            _t_: a Number,
           )
         </h1>
         <dl class="header">
           <dt>description</dt>
-          <dd>It converts _t_ from UTC to local time.</dd>
+          <dd>It converts time value _t_ from UTC to local time.</dd>
```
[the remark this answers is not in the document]
bakkot: I went ahead and pushed this change.

michaelficarra (approved):

PR #2459, opened 2021-07-16
Editorial: Set [[Construct]] for default constructor in ClassDefiniti…

h2oche:
In [15.7.12 Runtime Semantics: ClassDefinitionEvaluation](https://tc39.es/ecma262/#sec-runtime-semantics-classdefinitionevaluation), when constructor is not specified in class, `F` use default constructor created in step 14. However, in this case, there is no explicit step which set `[[Construct]]` internal property of built-in function object `F`. As a result, when evaluating a new expression with class using default constructor, current spec throws `TypeError`. Here is JS code which seems to be problematic in current spec:

```js
class Rectangle {
a() { return 1; }
}

// 13.3.5.1.1 EvaluateNew throws TypeError exception in step 7
// since [[Construct]] internal property is not set
var rec = new Rectangle; 
```

Setting `[[Construct]]` for built-in function objects in [10.2.5. MakeConstructor](https://tc39.es/ecma262/#sec-makeconstructor) seems to be a plausible fix.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13175,6 +13175,8 @@
           1. Assert: IsConstructor(_F_) is *false*.
           1. Assert: _F_ is an extensible object that does not have a *"prototype"* own property.
           1. Set _F_.[[Construct]] to the definition specified in <emu-xref href="#sec-ecmascript-function-objects-construct-argumentslist-newtarget"></emu-xref>.
+        1. Else,
+          1. Set _F_.[[Construct]] to the definition specified in <emu-xref href="#sec-built-in-function-objects-construct-argumentslist-newtarget"></emu-xref>.
         1. Set _F_.[[ConstructorKind]] to ~base~.
         1. If _writablePrototype_ is not present, set _writablePrototype_ to *true*.
         1. If _prototype_ is not present, then


ljharb:
If it's an ecmascript function object, it already has [[Construct]] applied: https://tc39.es/ecma262/#sec-ecmascript-function-objects-construct-argumentslist-newtarget

h2oche:
@ljharb I think so. However, it seems that `F` is not ecmascript function object since it is created by `CreateBuiltinFunction` with abstract closure in step 14.b.

bakkot (approved):

bakkot:
Thanks, I believe this was an oversight in #2216.

spec.html
```diff
@@ -13122,6 +13122,7 @@ <h1>
           1. Assert: IsConstructor(_F_) is *false*.
           1. Assert: _F_ is an extensible object that does not have a *"prototype"* own property.
           1. Set _F_.[[Construct]] to the definition specified in <emu-xref href="#sec-ecmascript-function-objects-construct-argumentslist-newtarget"></emu-xref>.
+        1. Else, set _F_.[[Construct]] to the definition specified in <emu-xref href="#sec-built-in-function-objects-construct-argumentslist-newtarget"></emu-xref>.
```
jmdyck, line 13125:
```suggestion
        1. Else,
          1. Set _F_.[[Construct]] to the definition specified in <emu-xref href="#sec-built-in-function-objects-construct-argumentslist-newtarget"></emu-xref>.
```

jmdyck, line 13125:
```suggestion
        1. Else,
          1. Set _F_.[[Construct]] to the definition specified in <emu-xref href="#sec-built-in-function-objects-construct-argumentslist-newtarget"></emu-xref>.
```
(@ljharb [suggested](https://github.com/tc39/ecma262/pull/2144#discussion_r473334681) that if one arm of an `If` is multi-line, all should be.)

h2oche: @jmdyck
Thanks, I added your suggestion.

PR #2460, opened 2021-07-16
Editorial: Validate array length using `SameValueZero` in `ArraySetLength` and `Array`

pacokwon:
I believe that the `&ne;` (≠) notation should be used when comparing array lengths in [10.4.2.4 ArraySetLength](https://tc39.es/ecma262/#sec-arraysetlength)'s step `5`.

According to [test262/test/built-ins/Object/defineProperties/15.2.3.7-6-a-127.js](https://github.com/tc39/test262/blob/master/test/built-ins/Object/defineProperties/15.2.3.7-6-a-127.js), a `RangeError` must not be thrown, meaning that `newLen` and `numberLen`, which are `0` and `-0.0` in this case, should be "the same value". 

From what I've seen in #1963 and [one of the discussions](https://github.com/tc39/ecma262/pull/2007/files#r501329621) when the phrase "the same value" was first introduced, the likes of `=`, `≠` seem to be used for mathematical equality, which is what I think is a more suitable notation to express the comparison.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13728,7 +13728,7 @@
           1. Let _newLenDesc_ be a copy of _Desc_.
           1. [id="step-arraysetlength-newlen"] Let _newLen_ be ? ToUint32(_Desc_.[[Value]]).
           1. [id="step-arraysetlength-numberlen"] Let _numberLen_ be ? ToNumber(_Desc_.[[Value]]).
-          1. If _newLen_ is not the same value as _numberLen_, throw a *RangeError* exception.
+          1. If SameValueZero(_newLen_, _numberLen_) is *false*, throw a *RangeError* exception.
           1. Set _newLenDesc_.[[Value]] to _newLen_.
           1. Let _oldLenDesc_ be OrdinaryGetOwnProperty(_A_, *"length"*).
           1. Assert: ! IsDataDescriptor(_oldLenDesc_) is *true*.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -13713,7 +13713,7 @@ <h1>
           1. Let _newLenDesc_ be a copy of _Desc_.
           1. [id="step-arraysetlength-newlen"] Let _newLen_ be ? ToUint32(_Desc_.[[Value]]).
           1. [id="step-arraysetlength-numberlen"] Let _numberLen_ be ? ToNumber(_Desc_.[[Value]]).
-          1. If _newLen_ is not the same value as _numberLen_, throw a *RangeError* exception.
+          1. If _newLen_ &ne; _numberLen_, throw a *RangeError* exception.
```
bakkot, line 13716: Thanks for the patch! I agree this is incorrect currently. However, I would prefer that we avoid the use of `=` (or `≠`) to compare Number values, particularly when one of them might be `-0`, because [it's ambiguous](https://github.com/tc39/ecma262/pull/2007/#issuecomment-637070282) in cases like this. ("is the same value as" is unambiguous, but here it's incorrect.)

Instead, we can use a comparison operation that makes it explicit that the floats `-0` and `0` should be considered to equivalent, as in

```suggestion
          1. If SameValueZero(_newLen_, _numberLen_) is *false*, throw a *RangeError* exception.
```

pacokwon: I understand, and I agree with your suggestion! Thank you.

pacokwon:
A quick question - [23.1.1.1](https://tc39.es/ecma262/#sec-array)'s step `5.d.ii` seems like it has a similar problem, but I haven't found any tests to refer to for the actual semantics. Would this step be considered incorrect as well?

bakkot:
Since `new Array(-0)` succeeds on all engines, and since in ES6 it was specified with the same wording as was used in ArraySetLength, I think we can reasonably conclude that that step is currently incorrect in the same way, yes. Would you update this PR to cover that case as well?

pacokwon:
Of course. I've applied the changes and updated the PR, let me know what you think!

bakkot (approved):
Looks good to me. Once another editor approves we'll land this. Thanks for the contribution!

michaelficarra (approved):

PR #2468, opened 2021-07-20
Editorial: Cleanup ValidateAndApplyPropertyDescriptor and own prop operations

### LINKED ISSUE #2466: ValidateAndApplyPropertyDescriptor description should state that P can be undefined
https://tc39.es/ecma262/#sec-validateandapplypropertydescriptor says:

> The abstract operation ValidateAndApplyPropertyDescriptor takes arguments O (an Object or undefined), P (a property key), ...

but IsCompatiblePropertyDescriptor calls it with P = undefined, so I think the above should say "P (a property key or undefined)".

gibson042:
Fixes #2466

Commits can be reviewed sequentially, but the [build preview](https://ci.tc39.es/preview/tc39/ecma262/pull/2468) is probably best:
* [merged Property Attributes table](https://ci.tc39.es/preview/tc39/ecma262/pull/2468#sec-property-attributes)
* [indirect Record field references](https://ci.tc39.es/preview/tc39/ecma262/pull/2468#sec-list-and-record-specification-type) ([[\<_fieldName_>]])
* [ValidateAndApplyPropertyDescriptor](https://ci.tc39.es/preview/tc39/ecma262/pull/2468#sec-validateandapplypropertydescriptor)
* [Proxy [[GetOwnProperty]] note clarifications](https://ci.tc39.es/preview/tc39/ecma262/pull/2468#sec-proxy-object-internal-methods-and-internal-slots-getownproperty-p)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2593,26 +2593,32 @@
 
       <emu-clause id="sec-property-attributes">
         <h1>Property Attributes</h1>
-        <p>Attributes are used in this specification to define and explain the state of Object properties. A data property associates a key value with the attributes listed in <emu-xref href="#table-data-property-attributes"></emu-xref>.</p>
-        <emu-table id="table-data-property-attributes" caption="Attributes of a Data Property" oldids="table-2">
+        <!--
+           TODO: merge this with sec-property-descriptor-specification-type?
+          cf. https://github.com/tc39/ecma262/issues/1067
+        -->
+        <p>Attributes are used in this specification to define and explain the state of Object properties as described in <emu-xref href="#table-object-property-attributes"></emu-xref>. Unless specified explicitly, the initial value of each attribute is its Default Value.</p>
+        <emu-table id="table-object-property-attributes" caption="Attributes of an Object property" oldids="table-2,table-3,table-4,table-data-property-attributes,table-accessor-property-attributes,table-default-attribute-values">
           <table>
             <tr>
-              <th>
-                Attribute Name
-              </th>
-              <th>
-                Value Domain
-              </th>
-              <th>
-                Description
-              </th>
+              <th>Attribute Name</th>
+              <th>Types of property for which it is present</th>
+              <th>Value Domain</th>
+              <th>Default Value</th>
+              <th>Description</th>
             </tr>
             <tr>
               <td>
                 [[Value]]
               </td>
               <td>
-                Any ECMAScript language type
+                data property
+              </td>
+              <td>
+                an ECMAScript language value
+              </td>
+              <td>
+                *undefined*
               </td>
               <td>
                 The value retrieved by a get access of the property.
@@ -2623,7 +2629,13 @@
                 [[Writable]]
               </td>
               <td>
-                Boolean
+                data property
+              </td>
+              <td>
+                a Boolean
+              </td>
+              <td>
+                *false*
               </td>
               <td>
                 If *false*, attempts by ECMAScript code to change the property's [[Value]] attribute using [[Set]] will not succeed.
@@ -2631,48 +2643,16 @@
             </tr>
             <tr>
               <td>
-                [[Enumerable]]
-              </td>
-              <td>
-                Boolean
-              </td>
-              <td>
-                If *true*, the property will be enumerated by a for-in enumeration (see <emu-xref href="#sec-for-in-and-for-of-statements"></emu-xref>). Otherwise, the property is said to be non-enumerable.
-              </td>
-            </tr>
-            <tr>
-              <td>
-                [[Configurable]]
-              </td>
-              <td>
-                Boolean
-              </td>
-              <td>
-                If *false*, attempts to delete the property, change the property to be an accessor property, or change its attributes (other than [[Value]], or changing [[Writable]] to *false*) will fail.
-              </td>
-            </tr>
-          </table>
-        </emu-table>
-        <p>An accessor property associates a key value with the attributes listed in <emu-xref href="#table-accessor-property-attributes"></emu-xref>.</p>
-        <emu-table id="table-accessor-property-attributes" caption="Attributes of an Accessor Property" oldids="table-3">
-          <table>
-            <tr>
-              <th>
-                Attribute Name
-              </th>
-              <th>
-                Value Domain
-              </th>
-              <th>
-                Description
-              </th>
-            </tr>
-            <tr>
-              <td>
                 [[Get]]
               </td>
               <td>
-                Object | Undefined
+                accessor property
+              </td>
+              <td>
+                an Object or *undefined*
+              </td>
+              <td>
+                *undefined*
               </td>
               <td>
                 If the value is an Object it must be a function object. The function's [[Call]] internal method (<emu-xref href="#table-additional-essential-internal-methods-of-function-objects"></emu-xref>) is called with an empty arguments list to retrieve the property value each time a get access of the property is performed.
@@ -2683,7 +2663,13 @@
                 [[Set]]
               </td>
               <td>
-                Object | Undefined
+                accessor property
+              </td>
+              <td>
+                an Object or *undefined*
+              </td>
+              <td>
+                *undefined*
               </td>
               <td>
                 If the value is an Object it must be a function object. The function's [[Call]] internal method (<emu-xref href="#table-additional-essential-internal-methods-of-function-objects"></emu-xref>) is called with an arguments list containing the assigned value as its sole argument each time a set access of the property is performed. The effect of a property's [[Set]] internal method may, but is not required to, have an effect on the value returned by subsequent calls to the property's [[Get]] internal method.
@@ -2694,10 +2680,16 @@
                 [[Enumerable]]
               </td>
               <td>
-                Boolean
+                data property or accessor property
               </td>
               <td>
-                If *true*, the property is to be enumerated by a for-in enumeration (see <emu-xref href="#sec-for-in-and-for-of-statements"></emu-xref>). Otherwise, the property is said to be non-enumerable.
+                a Boolean
+              </td>
+              <td>
+                *false*
+              </td>
+              <td>
+                If *true*, the property will be enumerated by a for-in enumeration (see <emu-xref href="#sec-for-in-and-for-of-statements"></emu-xref>). Otherwise, the property is said to be non-enumerable.
               </td>
             </tr>
             <tr>
@@ -2705,71 +2697,16 @@
                 [[Configurable]]
               </td>
               <td>
-                Boolean
+                data property or accessor property
               </td>
               <td>
-                If *false*, attempts to delete the property, change the property to be a data property, or change its attributes will fail.
-              </td>
-            </tr>
-          </table>
-        </emu-table>
-        <p>If the initial values of a property's attributes are not explicitly specified by this specification, the default value defined in <emu-xref href="#table-default-attribute-values"></emu-xref> is used.</p>
-        <emu-table id="table-default-attribute-values" caption="Default Attribute Values" oldids="table-4">
-          <table>
-            <tr>
-              <th>
-                Attribute Name
-              </th>
-              <th>
-                Default Value
-              </th>
-            </tr>
-            <tr>
-              <td>
-                [[Value]]
-              </td>
-              <td>
-                *undefined*
-              </td>
-            </tr>
-            <tr>
-              <td>
-                [[Get]]
-              </td>
-              <td>
-                *undefined*
-              </td>
-            </tr>
-            <tr>
-              <td>
-                [[Set]]
-              </td>
-              <td>
-                *undefined*
-              </td>
-            </tr>
-            <tr>
-              <td>
-                [[Writable]]
+                a Boolean
               </td>
               <td>
                 *false*
               </td>
-            </tr>
-            <tr>
               <td>
-                [[Enumerable]]
-              </td>
-              <td>
-                *false*
-              </td>
-            </tr>
-            <tr>
-              <td>
-                [[Configurable]]
-              </td>
-              <td>
-                *false*
+                If *false*, attempts to delete the property, change it from a data property to an accessor property or from an accessor property to a data property, or make any changes to its attributes (other than replacing an existing [[Value]] or setting [[Writable]] to *false*) will fail.
               </td>
             </tr>
           </table>
@@ -4333,8 +4270,9 @@
     <emu-clause id="sec-property-descriptor-specification-type">
       <h1>The Property Descriptor Specification Type</h1>
       <p>The <dfn variants="Property Descriptors">Property Descriptor</dfn> type is used to explain the manipulation and reification of Object property attributes. Values of the Property Descriptor type are Records. Each field's name is an attribute name and its value is a corresponding attribute value as specified in <emu-xref href="#sec-property-attributes"></emu-xref>. In addition, any field may be present or absent. The schema name used within this specification to tag literal descriptions of Property Descriptor records is &ldquo;PropertyDescriptor&rdquo;.</p>
-      <p>Property Descriptor values may be further classified as data Property Descriptors and accessor Property Descriptors based upon the existence or use of certain fields. A data Property Descriptor is one that includes any fields named either [[Value]] or [[Writable]]. An accessor Property Descriptor is one that includes any fields named either [[Get]] or [[Set]]. Any Property Descriptor may have fields named [[Enumerable]] and [[Configurable]]. A Property Descriptor value may not be both a data Property Descriptor and an accessor Property Descriptor; however, it may be neither. A generic Property Descriptor is a Property Descriptor value that is neither a data Property Descriptor nor an accessor Property Descriptor. A fully populated Property Descriptor is one that is either an accessor Property Descriptor or a data Property Descriptor and that has all of the fields that correspond to the property attributes defined in either <emu-xref href="#table-data-property-attributes"></emu-xref> or <emu-xref href="#table-accessor-property-attributes"></emu-xref>.</p>
+      <p>Property Descriptor values may be further classified as data Property Descriptors and accessor Property Descriptors based upon the existence or use of certain fields. A data Property Descriptor is one that includes any fields named either [[Value]] or [[Writable]]. An accessor Property Descriptor is one that includes any fields named either [[Get]] or [[Set]]. Any Property Descriptor may have fields named [[Enumerable]] and [[Configurable]]. A Property Descriptor value may not be both a data Property Descriptor and an accessor Property Descriptor; however, it may be neither (in which case it is a generic Property Descriptor). A <dfn>fully populated Property Descriptor</dfn> is one that is either an accessor Property Descriptor or a data Property Descriptor and that has all of the corresponding fields defined in <emu-xref href="#table-object-property-attributes"></emu-xref>.</p>
       <p>The following abstract operations are used in this specification to operate upon Property Descriptor values:</p>
+      <!-- TODO: replace Boolean-returning operations with PropertyDescriptorType → ~generic~ | ~data~ | ~accessor~? -->
 
       <emu-clause id="sec-isaccessordescriptor" type="abstract operation">
         <h1>
@@ -12653,7 +12591,7 @@
         <dl class="header">
         </dl>
         <emu-alg>
-          1. Return ValidateAndApplyPropertyDescriptor(*undefined*, *undefined*, _Extensible_, _Desc_, _Current_).
+          1. Return ValidateAndApplyPropertyDescriptor(*undefined*, *""*, _Extensible_, _Desc_, _Current_).
         </emu-alg>
       </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -3742,7 +3674,7 @@ <h1>The List and Record Specification Types</h1>
...
-      <p>The <dfn>Record</dfn> type is used to describe data aggregations within the algorithms of this specification. A Record type value consists of one or more named fields. The value of each field is either an ECMAScript value or an abstract value represented by a name associated with the Record type. Field names are always enclosed in double brackets, for example [[Value]].</p>
+      <p>The <dfn>Record</dfn> type is used to describe data aggregations within the algorithms of this specification. A Record type value consists of one or more named fields. The value of each field is either an ECMAScript value or an abstract value represented by a name associated with the Record type. Field names are always enclosed in double brackets, for example [[Value]]. When the name of a field is identified by a named alias (<emu-xref href="#sec-algorithm-conventions"></emu-xref>), the alias name is wrapped in angle brackets inside the double brackets, for example [[&lt;_fieldName_&gt;]].</p>
```
bakkot, line 3677: Is this a thing we actually do?

gibson042: It currently exists as an extension in [ECMA-402](https://tc39.es/ecma402/#conventions), and I'm pulling it upstream.

ljharb:
oof, this AO is a mess :-/ i'll try to update es-abstract with these changes and see if it still passes my tests.

spec.html
```diff
@@ -12308,50 +12241,35 @@ <h1>
...
-              1. If _Desc_.[[Set]] is present and SameValue(_Desc_.[[Set]], _current_.[[Set]]) is *false*, return *false*.
-              1. If _Desc_.[[Get]] is present and SameValue(_Desc_.[[Get]], _current_.[[Get]]) is *false*, return *false*.
-              1. Return *true*.
+            1. Let _unsettableFields_ be &laquo; [[Configurable]], [[Enumerable]], [[Get]], [[Set]] &raquo;.
+            1. If ! IsAccessorDescriptor(_current_) is *true*, append [[Value]] and [[Writable]] to _unsettableFields_.
+            1. If _current_.[[Writable]] is present and is *false*, append [[Value]] and [[Writable]] to _unsettableFields_.
+            1. For each field name _field_ of _unsettableFields_, do
+              1. If _Desc_.[[&lt;_field_&gt;]] is present and either _current_.[[&lt;_field_&gt;]] is not present or ! SameValue(_Desc_.[[&lt;_field_&gt;]], _current_.[[&lt;_field_&gt;]]) is *false*, return *false*.
           1. If _O_ is not *undefined*, then
+            1. If ! IsDataDescriptor(_current_) is *true* and ! IsAccessorDescriptor(_Desc_) is *true*, then
```
ljharb, line 12269: how can both of these be true?

spec.html
```diff
@@ -12308,50 +12241,35 @@ <h1>
...
+              1. Replace the property named _P_ of object _O_ with an accessor property having [[Configurable]] and [[Enumerable]] attributes as described by _current_ and each other attribute set to its <emu-xref href="#table-object-property-attributes">default value</emu-xref>.
+            1. Else if ! IsAccessorDescriptor(_current_) is *true* and ! IsDataDescriptor(_Desc_) is *true*, then
+              1. Replace the property named _P_ of object _O_ with a data property having [[Configurable]] and [[Enumerable]] attributes as described by _current_ and each other attribute set to its <emu-xref href="#table-object-property-attributes">default value</emu-xref>.
             1. For each field of _Desc_ that is present, set the corresponding attribute of the property named _P_ of object _O_ to the value of the field.
```
ljharb, line 12273: could these perhaps be rewritten to use `! DefineOwnPropertyOrThrow`, since we know O is configurable?

spec.html
```diff
@@ -12308,50 +12241,35 @@ <h1>
...
-            1. Assert: ! IsAccessorDescriptor(_current_) and ! IsAccessorDescriptor(_Desc_) are both *true*.
-            1. If _current_.[[Configurable]] is *false*, then
-              1. If _Desc_.[[Set]] is present and SameValue(_Desc_.[[Set]], _current_.[[Set]]) is *false*, return *false*.
-              1. If _Desc_.[[Get]] is present and SameValue(_Desc_.[[Get]], _current_.[[Get]]) is *false*, return *false*.
-              1. Return *true*.
+            1. Let _unsettableFields_ be &laquo; [[Configurable]], [[Enumerable]], [[Get]], [[Set]] &raquo;.
+            1. If ! IsAccessorDescriptor(_current_) is *true*, append [[Value]] and [[Writable]] to _unsettableFields_.
+            1. If _current_.[[Writable]] is present and is *false*, append [[Value]] and [[Writable]] to _unsettableFields_.
+            1. For each field name _field_ of _unsettableFields_, do
+              1. If _Desc_.[[&lt;_field_&gt;]] is present and either _current_.[[&lt;_field_&gt;]] is not present or ! SameValue(_Desc_.[[&lt;_field_&gt;]], _current_.[[&lt;_field_&gt;]]) is *false*, return *false*.
```
ljharb, line 12267: this seems like it could be shortened to something that said "if current and desc are not the same, and !(current.writable  && !Desc.writable), then return false", without using this new dynamic slot thing?

spec.html
gibson042: I don't think so, because we're logically one or two layers down from that, setting field-like attributes on the Record-like data structure that represents "a property of an object"... note that ValidateAndApplyPropertyDescriptor is invoked _by_ (and is basically the implementation of) OrdinaryDefineOwnProperty.

spec.html
gibson042: _current_ and _Desc_ are allowed to differ, e.g. `Object.defineProperty(Object.defineProperty({}, "x", pd1={value: 1, writable: true}), "x", pd2={writable: false})`. This is specifically checking if specific **fields** of _Desc_ are attempting to introduce invalid changes.

spec.html
ljharb: fair point, thanks

spec.html
ljharb: there's only 5 possible field names; is the dynamic loop that much of an improvement over listing each one?

gibson042: That's a matter of opinion. The big change I like here is consolidation of "attempt to set unsettable attribute" into a single test rather than having that logic dispersed. Testing each attribute individually while preserving that consolidation would look something like the below, which seems worse to me:
```
1. If _Desc_.[[Configurable]] is present and ! SameValue(_Desc_.[[Configurable]], _current_.[[Configurable]]) is *false*, return *false*.
1. If _Desc_.[[Enumerable]] is present and ! SameValue(_Desc_.[[Enumerable]], _current_.[[Enumerable]]) is *false*, return *false*.
1. If _Desc_.[[Get]] is present and either _current_.[[Get]] is not present or ! SameValue(_Desc_.[[Get]], _current_.[[Get]]) is *false*, return *false*.
1. If _Desc_.[[Set]] is present and either _current_.[[Set]] is not present or ! SameValue(_Desc_.[[Set]], _current_.[[Set]]) is *false*, return *false*.
1. If _unsettableFields_ contains [[Value]] and _Desc_.[[Value]] is present and either _current_.[[Value]] is not present or ! SameValue(_Desc_.[[Value]], _current_.[[Value]]) is *false*, return *false*.
1. If _unsettableFields_ contains [[Writable]] and _Desc_.[[Writable]] is present and either _current_.[[Writable]] is not present or ! SameValue(_Desc_.[[Writable]], _current_.[[Writable]]) is *false*, return *false*.
```

ljharb: i'd expect unconditional checks for configurable/enumerable, and then two indented branches - one for "data" and another for "accessor", one of which that checked value/writable and the other get/set. the indirection doesn't seem as readable to me.

ljharb:
Currently I'm getting two test failures - still not sure if there's an issue with my library or with this PR. The tests failing are about calling ArraySetLength while making the length nonwritable (which no longer changes writable to false), and another ArraySetLength that fails to change the value. @gibson042 if you're up for a quick pairing session sometime we can look over it and make sure I'm the one who's got the bug :-D

spec.html
bakkot: I also find this loop kind of hard to read. I'd prefer to be more explicit, as @ljharb suggests, I think:

```
1. If _current_.[[Configurable]] is *false*, then
  1. If ! SameValue(! IsDataDescriptor(_current_), ! IsDataDescriptor(_Desc_)) is *false*, return *false*.
  1. If _Desc_.[[Configurable]] is present and _Desc_.[[Configurable]] is *true*, return *false*.
  1. If _Desc_.[[Enumerable]] is present and ! SameValue(_Desc_.[[Enumerable]], _current_.[[Enumerable]]) is *false*, return *false*.
  1. If ! IsDataDescriptor(_current_) is *true*, then
    1. If _current_.[[Writable]] is *false*, then
      1. If _Desc_.[[Value]] is present and ! SameValue(_Desc_.[[Value]], _current_.[[Value]]) is *false*, return *false*.
      1. If _Desc_.[[Writable]] is present and _Desc_.[[Writable]] is *true*, return *false*.
  1. Else,
    1. Assert: ! IsAccessorDescriptor(_current_) and ! IsAccessorDescriptor(_Desc_) are both *true*.
    1. If _Desc_.[[Get]] is present and ! SameValue(_Desc_.[[Get]], _current_.[[Get]]) is *false*, return *false*.
    1. If _Desc_.[[Set]] is present and ! SameValue(_Desc_.[[Set]], _current_.[[Set]]) is *false*, return *false*.
```

(We should also change to description of `_current_` to be "a fully populated Property Descriptor", since it is and that makes it clear why we don't need presence checks for `_current_`.)

gibson042: > `1. If _Desc_.[[Configurable]] is present and _Desc_.[[Configurable]] is *true*, return *false*.`

Am I alone in preferring `SameValue` over explicit `*true*`/`*false*` that requires me to scan back up and find how that relates to the covering conditions? This block is logically a collection of checks verifying the preservation of attributes that must not change, which I find to be obscured by the deep structure and divergent logic across attributes.

gibson042:
Proof that behavior is the same in all cases: https://jsfiddle.net/028yqu6d/
But I'm happy to pair up as well.

spec.html
bakkot: All the attributes in my example use SameValue except for [[Configurable]] and [[Writable]], and for both of those I find the condition to be most clearly stated as "if false, must not become true", rather than (as is the case for the other attributes) "must not change".

spec.html
```diff
@@ -12308,50 +12241,35 @@ <h1>
...
-              1. Return *true*.
-          1. Else,
-            1. Assert: ! IsAccessorDescriptor(_current_) and ! IsAccessorDescriptor(_Desc_) are both *true*.
-            1. If _current_.[[Configurable]] is *false*, then
-              1. If _Desc_.[[Set]] is present and SameValue(_Desc_.[[Set]], _current_.[[Set]]) is *false*, return *false*.
-              1. If _Desc_.[[Get]] is present and SameValue(_Desc_.[[Get]], _current_.[[Get]]) is *false*, return *false*.
-              1. Return *true*.
+            1. Let _unsettableFields_ be &laquo; [[Configurable]], [[Enumerable]], [[Get]], [[Set]] &raquo;.
+            1. If ! IsAccessorDescriptor(_current_) is *true*, append [[Value]] and [[Writable]] to _unsettableFields_.
+            1. If _current_.[[Writable]] is present and is *false*, append [[Value]] and [[Writable]] to _unsettableFields_.
```
ljharb, line 12265:
```suggestion
            1. Else if _current_.[[Writable]] is *false*, append [[Value]] and [[Writable]] to _unsettableFields_.
```

spec.html
```diff
@@ -12308,50 +12241,35 @@ <h1>
...
             1. Else,
-              1. Assert: ! IsAccessorDescriptor(_Desc_) is *true*.
-              1. If _O_ is not *undefined*, create an own accessor property named _P_ of object _O_ whose [[Get]], [[Set]], [[Enumerable]], and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its <emu-xref href="#table-default-attribute-values">default value</emu-xref>.
+              1. Create an own data property named _P_ of object _O_ whose [[Value]], [[Writable]], [[Enumerable]], and [[Configurable]] attribute values are described by _Desc_. If the value of an attribute field of _Desc_ is absent, the attribute of the newly created property is set to its <emu-xref href="#table-object-property-attributes">default value</emu-xref>.
             1. Return *true*.
           1. If every field in _Desc_ is absent, return *true*.
```
ljharb, line 12261:
```suggestion
          1. Assert: _current_ is a completed Property Descriptor.
          1. If every field in _Desc_ is absent, return *true*.
```
(altho we'll have to define what a completed property descriptor is, which is what CompletePropertyDescriptor returns)

ljharb (commented):
We paired offline - this looks great and passes my tests!

spec.html
bakkot: Is that the same as a "[fully populated Property Descriptor](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-property-descriptor-specification-type)"? 

> A fully populated Property Descriptor is one that is either an accessor Property Descriptor or a data Property Descriptor and that has all of the fields that correspond to the property attributes defined in either [Table 3](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#table-data-property-attributes) or [Table 4](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#table-accessor-property-attributes).

ljharb: aha, that's the phrase i was looking for :-) i'll update my suggestion.

gibson042: Done.

michaelficarra:
I strongly prefer not introducing a dynamic record field lookup without very compelling motivation. Enumerating the possible field accesses, as suggested, is fine.

gibson042:
OK, the people have spoken so I pulled out the record field name indirection.

spec.html
```diff
@@ -12308,50 +12241,40 @@ <h1>
...
-              1. Return *true*.
+              1. If _Desc_.[[Value]] is present and ! SameValue(_Desc_.[[Value]], _current_.[[Value]]) is *false*, return *false*.
           1. If _O_ is not *undefined*, then
+            1. If ! IsDataDescriptor(_current_) is *true* and ! IsAccessorDescriptor(_Desc_) is *true*, then
+              1. Replace the property named _P_ of object _O_ with an accessor property having [[Configurable]] and [[Enumerable]] attributes as described by _current_ and each other attribute set to its <emu-xref href="#table-object-property-attributes">default value</emu-xref>.
+            1. Else if ! IsAccessorDescriptor(_current_) is *true* and ! IsDataDescriptor(_Desc_) is *true*, then
+              1. Replace the property named _P_ of object _O_ with a data property having [[Configurable]] and [[Enumerable]] attributes as described by _current_ and each other attribute set to its <emu-xref href="#table-object-property-attributes">default value</emu-xref>.
             1. For each field of _Desc_ that is present, set the corresponding attribute of the property named _P_ of object _O_ to the value of the field.
```
ljharb, line 12278: do we prefer this line as-is, or instead, inlining the specific fields into the if/else branch above?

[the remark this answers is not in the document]
ljharb: I'll update this suggestion, per @syg's request, to inline the configurable/enumerable stuff in the if and else if, and will update the PR.

ljharb:
I've rebased and updated this PR; PTAL.

spec.html
```diff
@@ -12677,51 +12615,46 @@ <h1>
...
+              1. If _Desc_ has a [[Configurable]] field, let _configurable_ be Desc.[[Configurable]], else let _configurable_ be _current_.[[Configurable]].
+              1. If _Desc_ has a [[Enumerable]] field, let _enumerable_ be Desc.[[Enumerable]], else let _enumerable_ be _current_.[[Enumerable]].
+              1. Replace the property named _P_ of object _O_ with an accessor property having [[Configurable]] and [[Enumerable]] attributes to _configurable_ and _enumerable_, respectively, and each other attribute set to its corresponding value in _Desc_ if present, otherwise to its <emu-xref href="#table-object-property-attributes">default value</emu-xref>.
+            1. Else if ! IsAccessorDescriptor(_current_) is *true* and ! IsDataDescriptor(_Desc_) is *true*, then
+              1. If _Desc_ has a [[Configurable]] field, let _configurable_ be Desc.[[Configurable]], else let _configurable_ be _current_.[[Configurable]].
+              1. If _Desc_ has a [[Enumerable]] field, let _enumerable_ be Desc.[[Enumerable]], else let _enumerable_ be _current_.[[Enumerable]].
```
gibson042, line 12654: I'd still prefer to avoid this duplication, but it works and is an improvement over the current spec.

gibson042 (commented):
:+1:

spec.html
```diff
@@ -4346,8 +4283,9 @@ <h1>
...
+      <p>Property Descriptor values may be further classified as data Property Descriptors and accessor Property Descriptors based upon the existence or use of certain fields. A data Property Descriptor is one that includes any fields named either [[Value]] or [[Writable]]. An accessor Property Descriptor is one that includes any fields named either [[Get]] or [[Set]]. Any Property Descriptor may have fields named [[Enumerable]] and [[Configurable]]. A Property Descriptor value may not be both a data Property Descriptor and an accessor Property Descriptor; however, it may be neither. A generic Property Descriptor is a Property Descriptor value that is neither a data Property Descriptor nor an accessor Property Descriptor. A <dfn>fully populated Property Descriptor</dfn> is one that is either an accessor Property Descriptor or a data Property Descriptor and that has all of the corresponding fields defined in <emu-xref href="#table-object-property-attributes"></emu-xref>.</p>
```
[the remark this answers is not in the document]
ljharb: nice, good cleanup

spec.html
```diff
@@ -2593,183 +2593,120 @@ <h1>The Object Type</h1>
...
+                accessor property
               </td>
               <td>
-                If the value is an Object it must be a function object. The function's [[Call]] internal method (<emu-xref href="#table-additional-essential-internal-methods-of-function-objects"></emu-xref>) is called with an empty arguments list to retrieve the property value each time a get access of the property is performed.
-              </td>
-            </tr>
-            <tr>
-              <td>
-                [[Set]]
+                Object | Undefined
```
michaelficarra, line 2652: These will conflict with https://github.com/tc39/ecma262/pull/2602 depending on which order they land, FYI.

spec.html
```diff
@@ -2593,183 +2593,120 @@ <h1>The Object Type</h1>
...
-              </td>
               <td>
-                *undefined*
+                [[Enumerable]]
               </td>
-            </tr>
-            <tr>
               <td>
-                [[Get]]
+                All
```
michaelficarra, line 2683: I think "data property or accessor property" is clearer.

michaelficarra: Although we may also have to add "or generic property"? Still, probably an improvement.

ljharb: "generic" can only apply to a Property Descriptor; a property must only be a data or an accessor property.

spec.html
ljharb: Rebasing this didn't seem to show a conflict. What did you have in mind?

spec.html
```diff
@@ -12664,51 +12602,46 @@ <h1>
...
-            1. Assert: ! IsAccessorDescriptor(_current_) and ! IsAccessorDescriptor(_Desc_) are both *true*.
-            1. If _current_.[[Configurable]] is *false*, then
-              1. If _Desc_.[[Set]] is present and SameValue(_Desc_.[[Set]], _current_.[[Set]]) is *false*, return *false*.
-              1. If _Desc_.[[Get]] is present and SameValue(_Desc_.[[Get]], _current_.[[Get]]) is *false*, return *false*.
-              1. Return *true*.
+              1. If _Desc_.[[Value]] is present and ! SameValue(_Desc_.[[Value]], _current_.[[Value]]) is *false*, return *false*.
           1. If _O_ is not *undefined*, then
-            1. For each field of _Desc_ that is present, set the corresponding attribute of the property named _P_ of object _O_ to the value of the field.
+            1. If ! IsDataDescriptor(_current_) is *true* and ! IsAccessorDescriptor(_Desc_) is *true*, then
+              1. If _Desc_ has a [[Configurable]] field, let _configurable_ be Desc.[[Configurable]], else let _configurable_ be _current_.[[Configurable]].
```
syg, line 12636: This PR mixes the "If _alias_ has a [[Field]] field" language and "If _alias_.[[Field]] is present". Could we pick one? I don't really have any preference for one over the other.

syg (approved):
looks mostly good to me modulo the consistent language comment below.

spec.html
ljharb: As discussed, I'll change the entire spec in a followup to the former form.

spec.html
ljharb: done, as discussed

michaelficarra (approved):

PR #2469, opened 2021-07-21
Markup: prefer `&notin;` over `&lt;!`

bakkot:
The inverse of #2327. I find this easier to understand.


syg (approved):
Strong agree from me, `&notin;` is much more readable.

PR #2471, opened 2021-07-24
Editorial: Improve the clarity and consistency of Number.prototype.toPrecision

gibson042:
We should strive to make algorithm steps maximally deterministic, and in particular should avoid interdependent assignment where possible.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -29827,11 +29827,12 @@
               1. Set _e_ to _e_ + 1.
               1. Set _n_ to 10<sup>_p_ - 1</sup>.
             1. Let _m_ be the String value consisting of the digits of the decimal representation of _n_ (in order, with no leading zeroes).
+            1. Assert: the length of _m_ is _p_.
             1. If _e_ &lt; -6 or _e_ &ge; _p_, then
               1. Assert: _e_ &ne; 0.
               1. If _p_ &ne; 1, then
-                1. Let _a_ be the first code unit of _m_.
-                1. Let _b_ be the other _p_ - 1 code units of _m_.
+                1. Let _a_ be the code unit at index 0 within _m_.
+                1. Let _b_ be the substring of _m_ from 1.
                 1. Set _m_ to the string-concatenation of _a_, *"."*, and _b_.
               1. If _e_ &gt; 0, then
                 1. Let _c_ be the code unit 0x002B (PLUS SIGN).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -29821,13 +29821,18 @@ <h1>Number.prototype.toPrecision ( _precision_ )</h1>
...
             1. If _e_ &lt; -6 or _e_ &ge; _p_, then
               1. Assert: _e_ &ne; 0.
               1. If _p_ &ne; 1, then
-                1. Let _a_ be the first code unit of _m_.
-                1. Let _b_ be the other _p_ - 1 code units of _m_.
+                1. Let _a_ be the code unit at index 0 within _m_.
+                1. Let _b_ be the substring of _m_ from 1.
                 1. Set _m_ to the string-concatenation of _a_, *"."*, and _b_.
               1. If _e_ &gt; 0, then
                 1. Let _c_ be the code unit 0x002B (PLUS SIGN).
```
gibson042, line 29838: I'm also thinking about improving the names of these aliases and replacing the code point references with inline literals.
```suggestion
                1. Let _sign_ be *"+"*.
```

bakkot:
This seems like an improvement to my eyes, although I'm not sure about the `n < 10^p` -> `n ≤ 10^p` change - what was the motivation for that?

Also, I note that there's a very similar text in `Number.prototype.toExponential`; presumably they should match.

gibson042:
> This seems like an improvement to my eyes, although I'm not sure about the `n < 10^p` -> `n ≤ 10^p` change - what was the motivation for that?

This is actually the crux of the change, and it relates to cases where rounding increases magnitude. For example, consider x = 995 with p = 2. The old algorithm would require non-constructive discovery of (e=2, n=99) and (e=3, n=10), respectively corresponding to approximations 99e1 (990) and 10e2 (1000), and then tiebreaking to the latter because it is larger. The proposed algorithm instead anchors e=2 to discover n=99 and n=100 (the latter only allowed by looseness of "≤"), performs the same tiebreaking logic to settle on n=100, and then ends up at the same place by normalizing 100e1 to 10e2.

> Also, I note that there's a very similar text in `Number.prototype.toExponential`; presumably they should match.

Yes, and also in ECMA-402 (this change was actually prompted by https://github.com/tc39/ecma402/pull/572), and arguably in Number::toString as well. I'd like to explore the possibility of having them all defer to a new operation for finding the relevant integers, but haven't yet determined what shape it would have.

PR #2477, opened 2021-08-02
Editorial: Fix alias use in IteratorBindingInitialization

gibson042:
It's cleaner and (IMO) more comprehensible to have a single Let.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -9253,16 +9253,16 @@
       <emu-alg>
         1. Let _bindingId_ be StringValue of |BindingIdentifier|.
         1. Let _lhs_ be ? ResolveBinding(_bindingId_, _environment_).
+        1. Let _v_ be *undefined*.
         1. If _iteratorRecord_.[[Done]] is *false*, then
           1. Let _next_ be IteratorStep(_iteratorRecord_).
           1. If _next_ is an abrupt completion, set _iteratorRecord_.[[Done]] to *true*.
           1. ReturnIfAbrupt(_next_).
           1. If _next_ is *false*, set _iteratorRecord_.[[Done]] to *true*.
           1. Else,
-            1. Let _v_ be IteratorValue(_next_).
+            1. Set _v_ to IteratorValue(_next_).
             1. If _v_ is an abrupt completion, set _iteratorRecord_.[[Done]] to *true*.
             1. ReturnIfAbrupt(_v_).
-        1. If _iteratorRecord_.[[Done]] is *true*, let _v_ be *undefined*.
         1. If |Initializer| is present and _v_ is *undefined*, then
           1. If IsAnonymousFunctionDefinition(|Initializer|) is *true*, then
             1. Set _v_ to the result of performing NamedEvaluation for |Initializer| with argument _bindingId_.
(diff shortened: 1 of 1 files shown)


WeslleyFratini (approved):

WeslleyFratini (approved):

spec.html
```diff
@@ -9253,16 +9253,16 @@ <h1>Runtime Semantics: IteratorBindingInitialization</h1>
...
           1. Let _next_ be IteratorStep(_iteratorRecord_).
           1. If _next_ is an abrupt completion, set _iteratorRecord_.[[Done]] to *true*.
           1. ReturnIfAbrupt(_next_).
           1. If _next_ is *false*, set _iteratorRecord_.[[Done]] to *true*.
           1. Else,
-            1. Let _v_ be IteratorValue(_next_).
+            1. Set _v_ to IteratorValue(_next_).
             1. If _v_ is an abrupt completion, set _iteratorRecord_.[[Done]] to *true*.
             1. ReturnIfAbrupt(_v_).
-        1. If _iteratorRecord_.[[Done]] is *true*, let _v_ be *undefined*.
```
michaelficarra, line 9265: I believe we still need

```
1. If _iteratorRecord_.[[Done]] is *true*, set _v_ to *undefined*.
```

because above we do

```
1. If _next_ is an abrupt completion, set _iteratorRecord_.[[Done]] to *true*.
```

gibson042: I don't think so, because that is immediately followed by `ReturnIfAbrupt(_v_)`.

michaelficarra: Wow, that is exceptionally weird. Can we replace those kinds of things with a `Return _v_.[[Value]]`?

bakkot: [ReturnIfAbrupt](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-returnifabrupt) returns the whole completion record, so not quite, but we could replace

```
1. If _next_ is an abrupt completion, set _iteratorRecord_.[[Done]] to *true*.
1. ReturnIfAbrupt(_next_).
```
with
```
1. If _next_ is an abrupt completion, then
  1. Set _iteratorRecord_.[[Done]] to *true*.
  1. Return _next_.
```
This happens in ~two dozen places.

gibson042: ReturnIfAbrupt also has completion value unwrapping behavior that would be needed if it were replaced, e.g.
```
1. If _next_ is an abrupt completion, then
  1. Set _iteratorRecord_.[[Done]] to *true*.
  1. Return _next_.
1. Set _next_ to _next_.[[Value]].
```
Or alternatively,
* Introduce a new abstract operation, e.g.:
  ```
  1. Set _next_ to ? UnwrapOrMarkIteratorDone(_next_, _iteratorRecord_).
  ```
* refactor to move the [[Done]]-setting logic elsewhere (although that would be a potentially bigger change that I'd rather not pull in here).

michaelficarra: Ah, that's right. Then I guess it's probably not worth making the change.

michaelficarra (approved):

syg (approved):

gwinamberly (commented):
I need to have the latest version of Java
Script for API AND ORDERING

PR #2482, opened 2021-08-05
Normative: `continue` labels should not pass through blocks

### LINKED ISSUE #2481: ContainsUndefinedContinueTarget needs a BlockStatement implementation
**Description:** The Static Semantics function `ContainsUndefinedContinueTarget` needs (at least) one more not-default implementation.

This function takes two arguments, an _iteration_set_, which contains labels a continue statement may use, and _label_set_, a list of labels which might get added to the _iteration_set_, if they apply to iteration statements.
Specified implementations except for _LabelledStatement_ and _BreakableStatement_ always reset the _label_set_ parameter to the empty list (as it's clear their productions are _not_ iteration statements).
The _default_ implementation, however, does _not_ clear the label set.

Consider: `lbl : { for ( ;; ) continue lbl ; }`

This is an undefined continue target. The _BlockStatement_ labelled by `lbl` is not an iteration statement. Web reality produces a syntax error here.

The production tree looks like

```
    LabelledStatement : LabelIdentifier : LabelledItem
    └── LabelledItem : Statement
        └── Statement : BlockStatement
            └── BlockStatement : Block
                └── Block : { StatementList }
                    └── StatementList : StatementListItem
                        └── StatementListItem : Statement
                            └── Statement : BreakableStatement
                                └── BreakableStatement : IterationStatement
```

_ContainsUndefinedContinueTarget_ for the first production adds `lbl` to the label set. 
All the remaining productions prior to the _BreakableStatement_ use the default implementation.
The _BreakableStatement_ moves the label set (containing `lbl`) into the iteration set and thus makes this label seem like a _valid_ target.

(If we actually tried to execute this code, the loop executor functions would actually treat that continue statement like a break statement, and the seemingly infinite loop would actually terminate at the first iteration.)

Fix: 
  The clearest fix would seem to be to add an implementation for the _Statement_ : _BlockStatement_ production:
  1. Return _ContainsUndefinedContinueTarget_ of _BlockStatement_ with arguments _iterationSet_ and « ».

This error seems to have been in every ECMAScript spec since the static semantics system was added in ES 6. (In particular, this issue was not due to the recent refactoring of syntax directed operations.)

bakkot:
Fixes #2481. Without this, the spec permits `lbl : { for ( ;; ) continue lbl ; }`, contrary to all implementations I have on hand.

There's a few different productions which would work here - adding a definition for any of `BlockStatement : Block`, `Block : { StatementList }`, `StatementList : StatementListItem`, or `StatementListItem : Statement` would work, in addition to `Statement : BlockStatement` as chosen in this PR. I think `Statement : BlockStatement` is clearest, but it doesn't really matter.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -8475,6 +8475,10 @@
       <emu-alg>
         1. Return *false*.
       </emu-alg>
+      <emu-grammar>Statement : BlockStatement</emu-grammar>
+      <emu-alg>
+        1. Return ContainsUndefinedContinueTarget of |BlockStatement| with arguments _iterationSet_ and &laquo; &raquo;.
+      </emu-alg>
       <emu-grammar>BreakableStatement : IterationStatement</emu-grammar>
       <emu-alg>
         1. Let _newIterationSet_ be the list-concatenation of _iterationSet_ and _labelSet_.


devsnek:
chakracore scares me

bakkot:
I got consensus for the normative change implied by this PR at TC39 today.

jugglinmike:
Test under review here: https://github.com/tc39/test262/pull/3180

michaelficarra (approved):

PR #2483, opened 2021-08-05
Editorial: move assertions about AO parameter types into AO header metadata

michaelficarra:


spec.html
```diff
@@ -13887,15 +13798,13 @@ <h1>
       <emu-clause id="sec-stringgetownproperty" type="abstract operation">
         <h1>
           StringGetOwnProperty (
-            _S_: unknown,
-            _P_: unknown,
+            _S_: an Object that has a [[StringData]] internal slot,
+            _P_: a property key,
```
bakkot, line 13802: we should maybe `dfn` this, though I guess that's can happen outside of this PR.

spec.html
```diff
@@ -6878,7 +6817,7 @@ <h1>
       <h1>
         AsyncIteratorClose (
           _iteratorRecord_: unknown,
```
bakkot, line 6819: We should also define `Iterator Record` somewhere, so that we can refer to it here. (Again, not necessarily in this PR.)

spec.html
```diff
@@ -6906,15 +6844,14 @@ <h1>
...
         CreateIterResultObject (
           _value_: unknown,
-          _done_: unknown,
+          _done_: a Boolean,
         )
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It creates an object that supports the IteratorResult interface.</dd>
+        <dd>It creates an object that conforms to the <i>IteratorResult</i> interface.</dd>
```
bakkot, line 6852: Unrelated: we should dfn `IteratorResult interface`, and all the [other interface names](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-common-iteration-interfaces), and stop italicizing them.

spec.html
```diff
@@ -10927,7 +10863,7 @@ <h1>Module Environment Records</h1>
           <h1>
             GetBindingValue (
               _N_: a String,
-              _S_: a Boolean,
+              _S_: *true*,
```
bakkot, line 10866: Ehhh, I don't like this change. This is an implementation of an abstract method, and the abstract method takes a boolean here, not `*true*`. I'd prefer to keep the assert.

spec.html
```diff
@@ -13476,7 +13393,7 @@ <h1>
     <emu-clause id="sec-createbuiltinfunction" type="abstract operation">
       <h1>
         CreateBuiltinFunction (
-          _behaviour_: unknown,
+          _behaviour_: an Abstract Closure, a set of algorithm steps, or some other definition of a function's behaviour provided in this specification,
```
bakkot, line 13396: This preamble now reads

> The abstract operation CreateBuiltinFunction takes arguments behaviour (an Abstract Closure, a set of algorithm steps, or some other definition of a function's behaviour provided in this specification), length (a non-negative integer or +∞), name (a property key), and additionalInternalSlotsList (a List of names of internal slots) and optional arguments realm (a Realm Record), prototype (an Object or null), and prefix (a String).

I... guess that's ok? But it's getting pretty hard to read.

spec.html
```diff
@@ -24405,13 +24282,12 @@ <h1>Tail Position Calls</h1>
     <emu-clause id="sec-isintailposition" type="abstract operation">
       <h1>
         Static Semantics: IsInTailPosition (
-          _call_: unknown,
+          _call_: a |CallExpression| Parse Node, an |OptionalChain| Parse Node, or a |MemberExpression| Parse Node,
```
bakkot, line 24285: I'm not convinced the extra precision over just `a Parse Node` is valuable here.

spec.html
```diff
@@ -24894,7 +24770,7 @@ <h1>Script Records</h1>
     <emu-clause id="sec-parse-script" type="abstract operation">
       <h1>
         ParseScript (
-          _sourceText_: unknown,
+          _sourceText_: ECMAScript source text,
```
bakkot, line 24773: There's an existing use of "an ECMAScript source text" in Note 1 under [The Syntactic Grammar](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-syntactic-grammar). If you're changing this to omit "an", you should change that one too, to match.

spec.html
```diff
@@ -26406,7 +26277,7 @@ <h1>
             _referencingScriptOrModule_: unknown,
             _specifier_: unknown,
             _promiseCapability_: a PromiseCapability Record,
-            _completion_: unknown,
+            _completion_: a normal completion whose [[Value]] is *undefined*,
```
bakkot, line 26280: Nope; the assertion below was conditional. You can make this `a Completion Record`, but you do need to retain the assertion below.

spec.html
```diff
@@ -34631,7 +34486,7 @@ <h1>
               1. Let _patternCharacters_ be a List whose elements are the code unit elements of _P_.
             1. Let _parseResult_ be ParsePattern(_patternText_, _u_).
             1. If _parseResult_ is a non-empty List of *SyntaxError* objects, throw a *SyntaxError* exception.
-            1. Assert: _parseResult_ is a Parse Node for |Pattern|.
+            1. Assert: _parseResult_ is a |Pattern|.
```
bakkot, line 34489:
```suggestion
            1. Assert: _parseResult_ is a |Pattern| Parse Node.
```

right?

spec.html
```diff
@@ -37841,7 +37683,7 @@ <h1>Abstract Operations for TypedArray Objects</h1>
       <emu-clause id="typedarray-species-create" type="abstract operation">
         <h1>
           TypedArraySpeciesCreate (
-            _exemplar_: unknown,
+            _exemplar_: an Object that has [[TypedArrayName]] and [[ContentType]] internal slots,
```
bakkot, line 37686:
```suggestion
            _exemplar_: a TypedArray,
```

Though, we actually have a definition for `Integer-Indexed exotic object`, unlike `TypedArray`, so possibly we should be using the former everywhere?

spec.html
```diff
@@ -38107,7 +37942,7 @@ <h1>
         <emu-clause id="sec-allocatetypedarraybuffer" type="abstract operation">
           <h1>
             AllocateTypedArrayBuffer (
-              _O_: a TypedArray object,
+              _O_: a TypedArray that has a [[ViewedArrayBuffer]] internal slot,
```
bakkot, line 37945:
```suggestion
              _O_: a TypedArray,
```

All TypedArrays have [[ViewedArrayBuffer]].

spec.html
```diff
@@ -38231,7 +38065,7 @@ <h1>
         </dl>
         <emu-alg>
           1. If IsCallable(_adder_) is *false*, throw a *TypeError* exception.
-          1. Assert: _iterable_ is present, and is neither *undefined* nor *null*.
+          1. Assert: _iterable_ is neither *undefined* nor *null*.
```
bakkot, line 38068: Any reason not to just declare its type to be `an ECMAScript language value, but not *undefined* or *null*`?

spec.html
```diff
@@ -39139,13 +38971,12 @@ <h1>
       <emu-clause id="sec-isdetachedbuffer" type="abstract operation">
         <h1>
           IsDetachedBuffer (
-            _arrayBuffer_: unknown,
+            _arrayBuffer_: an Object that has an [[ArrayBufferData]] internal slot,
```
bakkot, line 38974:
```suggestion
            _arrayBuffer_: an ArrayBuffer or a SharedArrayBuffer,
```

spec.html
```diff
@@ -39154,14 +38985,13 @@ <h1>
       <emu-clause id="sec-detacharraybuffer" type="abstract operation">
         <h1>
           DetachArrayBuffer (
-            _arrayBuffer_: unknown,
+            _arrayBuffer_: an Object that has [[ArrayBufferData]], [[ArrayBufferByteLength]], and [[ArrayBufferDetachKey]] internal slots,
```
bakkot, line 38988:
```suggestion
            _arrayBuffer_: an ArrayBuffer,
```

spec.html
```diff
@@ -38026,17 +37865,15 @@ <h1>
         <emu-clause id="sec-initializetypedarrayfromarraybuffer" type="abstract operation" oldids="sec-typedarray-buffer-byteoffset-length">
           <h1>
             InitializeTypedArrayFromArrayBuffer (
-              _O_: a TypedArray object,
-              _buffer_: an ArrayBuffer object,
+              _O_: a TypedArray,
+              _buffer_: an ArrayBuffer,
```
bakkot, line 37869: This is actually either an ArrayBuffer or a SharedArrayBuffer.

spec.html
```diff
@@ -39177,7 +39007,7 @@ <h1>
       <emu-clause id="sec-clonearraybuffer" type="abstract operation">
         <h1>
           CloneArrayBuffer (
-            _srcBuffer_: an ArrayBuffer object,
+            _srcBuffer_: an ArrayBuffer,
```
bakkot, line 39010:
```suggestion
            _srcBuffer_: an ArrayBuffer or a SharedArrayBuffer,
```

spec.html
```diff
@@ -39589,15 +39418,14 @@ <h1>
       <emu-clause id="sec-issharedarraybuffer" type="abstract operation">
         <h1>
           IsSharedArrayBuffer (
-            _obj_: unknown,
+            _obj_: an Object that has an [[ArrayBufferData]] internal slot,
```
bakkot, line 39421:
```suggestion
            _obj_: an ArrayBuffer or a SharedArrayBuffer,
```

spec.html
```diff
@@ -40168,14 +39996,13 @@ <h1>
       <emu-clause id="sec-validateatomicaccess" type="abstract operation">
         <h1>
           ValidateAtomicAccess (
-            _typedArray_: unknown,
+            _typedArray_: an Object that has a [[ViewedArrayBuffer]] internal slot,
```
bakkot, line 39999:
```suggestion
            _typedArray_: a TypedArray,
```
"has a [[ViewedArrayBuffer]] internal slot" is actually equivalent to "is a TypedArray or a DataView", but the algorithm assumes the existence of the `[[ArrayLength]]` slot, which rules out DataViews.

spec.html
```diff
@@ -41704,7 +41528,7 @@ <h1>The %AsyncFromSyncIteratorPrototype% Object</h1>
           <h1>%AsyncFromSyncIteratorPrototype%.next ( [ _value_ ] )</h1>
           <emu-alg>
             1. Let _O_ be the *this* value.
-            1. Assert: Type(_O_) is Object and _O_ has a [[SyncIteratorRecord]] internal slot.
+            1. Assert: Type(_O_) is Object with a [[SyncIteratorRecord]] internal slot.
```
bakkot, line 41531: This change makes no grammatical sense. Ditto the other occurrences.

spec.html
michaelficarra: I almost did that as part of this PR but backed it out. I can send a follow-up.

spec.html
michaelficarra: Yes but this concrete variant is only ever passed `true`. If we wanted all variants of this method to have the same signature, we should have a centralised description of it.

spec.html
michaelficarra: I think it reads fine. It's the number of parameters that's the issue, not the description of their types.

spec.html
bakkot: We do, in [table 17](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#table-abstract-methods-of-environment-records).

spec.html
michaelficarra: Sure. I think there's other usages like this though. It'd be nice to normalise them.

spec.html
michaelficarra: I wasn't sure if that assertion was conditional upon the IsCallable check above it.

spec.html
michaelficarra: Oops, I thought I fixed these before opening the PR.

spec.html
michaelficarra: I mean a centralised signature, with all the metadata.

bakkot: That seems arguably useful, but I don't see what bearing it has on this question. Overloads generally have the same signature as the method they're overloading. This should as well. Please put it back.

spec.html
bakkot: It is not.

spec.html
ljharb: We should definitely dfn that :-)

spec.html
ljharb: Do these have definitions at all?

bakkot: If we're being precise, "an ArrayBuffer" means "a value which was created by AllocateArrayBuffer", and "a SharedArrayBuffer" means "a value which was created by AllocateSharedArrayBuffer". (Similarly, "a TypedArray" means "a value which was created by AllocateTypedArray", "an Array" means "a value which was created by ArrayCreate", etc.)

Equivalently, ArrayBuffers are those objects with [[ArrayBufferData]], [[ArrayBufferByteLength]], and [[ArrayBufferDetachKey]] slots, and SharedArrayBuffers are those objects with [[ArrayBufferData]] and [[ArrayBufferByteLength]] slots but not a [[ArrayBufferDetachKey]] slot.

ljharb: Yes, but where in the spec is that defined explicitly? Do we use “a TypedArray” and similar anywhere else?

bakkot: > Yes, but where in the spec is that defined explicitly?

[AllocateArrayBuffer](https://tc39.es/ecma262/multipage/structured-data.html#sec-allocatearraybuffer) says "It is used to create an ArrayBuffer object".

> Do we use “a TypedArray” and similar anywhere else?

There are number of existing uses of "a TypedArray object" and "an ArrayBuffer object" as parameter types, yes.

jmdyck (changes requested):

spec.html
ljharb: ok, so we're using the phrasing "a(n) X object" in this PR, then?

spec.html
michaelficarra: I don't think minimising churn in this case is the top priority. We should do whatever reads clearest. Maybe it's worth discussing in the editor call.

spec.html
bakkot: > ok, so we're using the phrasing "a(n) X object" in this PR, then?

You can see for yourself that it uses "an X" (including changing existing usages). I agree that that change is basically orthogonal, but I don't mind including it here.

ljharb: It seems like it would be useful to make that change in a separate PR - in either order. Lots of people look at changes to the spec PR by PR, so readability/understandability isn't solely determined by the final output - and this is already a massive PR.

spec.html
jmdyck: Discussion sounds good.

spec.html
```diff
@@ -15329,7 +15213,7 @@ <h1>[[OwnPropertyKeys]] ( )</h1>
         1. If _trapResult_ contains any duplicate entries, throw a *TypeError* exception.
         1. Let _extensibleTarget_ be ? IsExtensible(_target_).
         1. Let _targetKeys_ be ? _target_.[[OwnPropertyKeys]]().
-        1. Assert: _targetKeys_ is a List whose elements are only String and Symbol values.
+        1. Assert: _targetKeys_ is a List whose elements are Strings or Symbols.
```
jmdyck, line 15216: I think it's possible to read this as meaning "whose elements are all Strings or all Symbols", which we don't intend. Here are some alternatives that don't have this problem:
- a List of elements, each of which is a String or a Symbol
- a List, each of whose elements is a String or a Symbol
- a List in which each element is a String or a Symbol
- a List whose elements are each a String or a Symbol
- a List whose every element is a String or a Symbol

Of these, I think I prefer the 3rd.

ljharb: I definitely don't think "all one or all the other" is even remotely implied or possible to infer, but any of the first three also seems fine.

michaelficarra: I also don't read the implication that @jmdyck does, but I am open to rewordings. Unfortunately, I don't think any of the suggestions are an improvement. In fact, I interpret the latter two suggestions to actually have the problem raised by @jmdyck.

I think something like "a List in which each element may independently be either a String or a Symbol" may be a bit clearer, but also seems unnecessarily verbose.

bakkot: We could define "property key" to be "a value of type String or Symbol" and then say "a List of property keys".

arhadthedev: @bakkot, there is already such a definition in <https://tc39.es/ecma262/#sec-object-type>, it just needs a `<dfn>` wrapping:

> Properties are identified using key values. A property key value is either an ECMAScript String value or a Symbol value. ...

bakkot: Sorry, yes, by "define" I meant "<dfn>".

spec.html
bakkot: In particular, when an AO takes exactly one kind of parse node, it's helpful to know that, but when it takes more than one it's a lot less so.

spec.html
michaelficarra: Editor call: We should normalise these 3 forms to "a |Foo| Parse Node".

michaelficarra:
@bakkot All comments addressed. Please review again.

bakkot (approved):
LGTM, but #2489 should maybe land first.

jmdyck (changes requested):
I have a bunch of similar changes for other operations, so I'll put them in a PR against this PR's branch.

michaelficarra:
@jmdyck Thanks. Merged.

bakkot:
This needs a rebase now that #2489 landed.

michaelficarra:
Rebased.

PR #2489, opened 2021-08-15
Editorial: "an Array", not "an Array object" [etc]

bakkot:
https://github.com/tc39/ecma262/pull/2483 proposes to change a few occurrences of "TypedArray object" to "TypedArray". (There were some usages of the latter form already.) That seems better, so we might as well do that everywhere it makes sense to do so.

I've changed most uses of "an X object" to be "an X", with the following exceptions:

- when referring to an exotic object in a context which emphasizes its exotic nature, I've left it as "an X exotic object"
- "Set object" remains as-is, to distinguish from the built-in type of the same name (I've also added a note calling attention to this)
- "Error object" and variants remain as-is, to make it clear it's talking about a value rather than an exceptional condition
- the various boxed primitive types remain "a Boolean object" (etc), to distinguish from the primitives
- "RegExp object" remains as-is to distinguish it from RegExp literals (this one I could go either way)
- "Proxy object" remains as-is because it read a little better to me (this one I could go either way)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -3319,7 +3319,7 @@
               <td>
               </td>
               <td>
-                The constructor of generator objects (<emu-xref href="#sec-generatorfunction-constructor"></emu-xref>)
+                The constructor of Generators (<emu-xref href="#sec-generatorfunction-constructor"></emu-xref>)
               </td>
             </tr>
             <tr>
@@ -3772,7 +3772,7 @@
 
     <emu-clause id="sec-set-and-relation-specification-type">
       <h1>The Set and Relation Specification Types</h1>
-      <p>The <em>Set</em> type is used to explain a collection of unordered elements for use in the memory model. Values of the Set type are simple collections of elements, where no element appears more than once. Elements may be added to and removed from Sets. Sets may be unioned, intersected, or subtracted from each other.</p>
+      <p>The <em>Set</em> type is used to explain a collection of unordered elements for use in the memory model. It is distinct from the ECMAScript collection type of the same name. To disambiguate, instances of the ECMAScript collection are consistently referred to as "Set objects" within this specification. Values of the Set type are simple collections of elements, where no element appears more than once. Elements may be added to and removed from Sets. Sets may be unioned, intersected, or subtracted from each other.</p>
       <p>The <dfn>Relation</dfn> type is used to explain constraints on Sets. Values of the Relation type are Sets of ordered pairs of values from its value domain. For example, a Relation on events is a set of ordered pairs of events. For a Relation _R_ and two values _a_ and _b_ in the value domain of _R_, _a_ _R_ _b_ is shorthand for saying the ordered pair (_a_, _b_) is a member of _R_. A Relation is least with respect to some conditions when it is the smallest Relation that satisfies those conditions.</p>
       <p>A <dfn variants="strict partial orders">strict partial order</dfn> is a Relation value _R_ that satisfies the following.</p>
       <ul>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -3762,7 +3762,7 @@ <h1>The List and Record Specification Types</h1>
 
     <emu-clause id="sec-set-and-relation-specification-type">
       <h1>The Set and Relation Specification Types</h1>
-      <p>The <dfn>Set</dfn> type is used to explain a collection of unordered elements for use in the memory model. Values of the Set type are simple collections of elements, where no element appears more than once. Elements may be added to and removed from Sets. Sets may be unioned, intersected, or subtracted from each other.</p>
+      <p>The <dfn>Set</dfn> type is used to explain a collection of unordered elements for use in the memory model. It is distinct from the ECMAScript collection type of the same name, instances of which this specification refers to as "Set objects". Values of the Set type are simple collections of elements, where no element appears more than once. Elements may be added to and removed from Sets. Sets may be unioned, intersected, or subtracted from each other.</p>
```
michaelficarra, line 3765:
```suggestion
      <p>The <dfn>Set</dfn> type is used to explain a collection of unordered elements for use in the memory model. It is distinct from the ECMAScript collection type of the same name. To disambiguate, instances of the ECMAScript collection are consistently referred to as "Set objects" within this specification. Values of the Set type are simple collections of elements, where no element appears more than once. Elements may be added to and removed from Sets. Sets may be unioned, intersected, or subtracted from each other.</p>
```

michaelficarra (approved):
I don't love "Date object" to "Date", but it's fine.

spec.html
```diff
@@ -11945,7 +11945,7 @@ <h1>Objectives</h1>
 
       <p>This specification does not make any guarantees that any object will be garbage collected. Objects which are not live may be released after long periods of time, or never at all. For this reason, this specification uses the term "may" when describing behaviour triggered by garbage collection.</p>
 
-      <p>The semantics of WeakRef and FinalizationRegistry objects is based on two operations which happen at particular points in time:</p>
+      <p>The semantics of WeakRefs and FinalizationRegistrys is based on two operations which happen at particular points in time:</p>
```
ljharb, line 11948: oof, `FinalizationRegistrys`

bakkot: Ehhh I will just fix that.

ljharb: lol that's also awkward if FInalizationRegistry is a dfn tho - making this "FinalizationRegistry objects" to avoid this problem was an explicit editorial review item for the weakref proposal :-p

bakkot: Neither `FinalizationRegistrys` nor `FinalizationRegistries` will auto-link currently, so we might as well go with the better one.

ljharb:
```suggestion
      <p>The semantics of a WeakRef and a FinalizationRegistry are based on two operations which happen at particular points in time:</p>
```
?

bakkot: That reads worse to me. It's the semantics of _all_ WeakRefs.

To be clear, you're just concerned about the automatic linking, right? I guess this can be the impetus for me to finally fix https://github.com/tc39/ecmarkup/issues/157.

ljharb: I'm concerned about both automatic linking and also natural pluralization; i suppose if ecmarkup could detect `y → ies` then it's not a big deal, but it makes it awkward to speak about out loud.

bakkot: > i suppose if ecmarkup could detect `y → ies` then it's not a big deal

I was thinking I'd just allow `dfn` to specify pluralizations explicitly.

> it makes it awkward to speak about out

Sorry, what becomes awkward?

ljharb: `ys` obviously, when spoken aloud, nobody will come up with the spelling; the `ies` one just seems weird to me even though also obviously, the pluralization of `registry` is `registries`, it's weird when it's a camelcased thing.

not a strong opinion, it's just weird :-)

michaelficarra: I think FinalizationRegistrys is fine.

bakkot: Rebased on https://github.com/tc39/ecma262/pull/2492, so now the plural `FinalizationRegistrys ` should link.

bakkot:
Ugh, hold off merging for a sec, I just noticed a typo

PR #2492, opened 2021-08-19
Markup: auto-link plural forms of terms

bakkot:
I [just added](https://github.com/tc39/ecmarkup/pull/346) support in ecmarkup for `dfn` tags to define other forms of the terms they define. This PR bumps ecmarkup to that version and makes use of the new attribute as appropriate for plural forms of defined nouns and noun phrases.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -187,13 +187,13 @@
     <p>An <em id="implementation">implementation</em> is an external source that further defines facilities enumerated in Annex <emu-xref href="#sec-host-layering-points"></emu-xref> or those that are marked as implementation-defined or implementation-approximated. In informal use, an implementation refers to a concrete artefact, such as a particular web browser.</p>
     <p>An <dfn id="implementation-defined">implementation-defined</dfn> facility is one that defers its definition to an external source without further qualification. This specification does not make any recommendations for particular behaviours, and conforming implementations are free to choose any behaviour within the constraints put forth by this specification.</p>
     <p>An <dfn id="implementation-approximated">implementation-approximated</dfn> facility is one that defers its definition to an external source while recommending an ideal behaviour. While conforming implementations are free to choose any behaviour within the constraints put forth by this specification, they are encouraged to strive to approximate the ideal. Some mathematical operations, such as <emu-xref href="#sec-math.exp"><code>Math.exp</code></emu-xref>, are implementation-approximated.</p>
-    <p>A <dfn id="host">host</dfn> is an external source that further defines facilities listed in Annex <emu-xref href="#sec-host-layering-points"></emu-xref> but does not further define other implementation-defined or implementation-approximated facilities. In informal use, a host refers to the set of all implementations, such as the set of all web browsers, that interface with this specification in the same way via Annex <emu-xref href="#sec-host-layering-points"></emu-xref>. A host is often an external specification, such as WHATWG HTML (<a href="https://html.spec.whatwg.org/">https://html.spec.whatwg.org/</a>). In other words, facilities that are host-defined are often further defined in external specifications.</p>
-    <p>A <dfn id="host-hook">host hook</dfn> is an abstract operation that is defined in whole or in part by an external source. All host hooks must be listed in Annex <emu-xref href="#sec-host-layering-points"></emu-xref>. A host hook must conform to at least the following requirements:</p>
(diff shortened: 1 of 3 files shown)


ljharb:
Lots of boilerplate from not auto-handling the "s" form :-/

bakkot:
Less than you might think - there's a lot of `dfn`s which aren't nouns and which therefore should not have `-s` forms, so we'd need at least a marker to indicate that a particular term is a noun and should be pluralized. Given that, also including the plural isn't much additional overhead.

spec.html
```diff
@@ -2406,14 +2406,14 @@ <h1>The Object Type</h1>
...
-      <p>An <dfn id="integer-index">integer index</dfn> is a String-valued property key that is a canonical numeric String (see <emu-xref href="#sec-canonicalnumericindexstring"></emu-xref>) and whose numeric value is either *+0*<sub>𝔽</sub> or a positive integral Number &le; 𝔽(2<sup>53</sup> - 1). An <dfn id="array-index">array index</dfn> is an integer index whose numeric value _i_ is in the range <emu-eqn>*+0*<sub>𝔽</sub> &le; _i_ &lt; 𝔽(2<sup>32</sup> - 1)</emu-eqn>.</p>
+      <p>An <dfn id="integer-index" variants="integer indexes">integer index</dfn> is a String-valued property key that is a canonical numeric String (see <emu-xref href="#sec-canonicalnumericindexstring"></emu-xref>) and whose numeric value is either *+0*<sub>𝔽</sub> or a positive integral Number &le; 𝔽(2<sup>53</sup> - 1). An <dfn id="array-index" variants="array indexes">array index</dfn> is an integer index whose numeric value _i_ is in the range <emu-eqn>*+0*<sub>𝔽</sub> &le; _i_ &lt; 𝔽(2<sup>32</sup> - 1)</emu-eqn>.</p>
```
michaelficarra, line 2416: We should settle on one of "indexes" or "indices". There's 2 occurrences of "indexes" (1 of each of these plurals) currently and 11 of "indices".

gibson042: [ECMA-402](https://tc39.es/ecma402/) also has a single occurrence of "indices" and none of "indexes", and the [Segmenter proposal](https://tc39.es/proposal-intl-segmenter/) has two of "indices" and none of "indexes".

bakkot: I'm happy to change to be consistently "indices" everywhere.

spec.html
```diff
@@ -13662,7 +13662,7 @@ <h1>Array Exotic Objects</h1>
         <p>A String property name _P_ is an <em>array index</em> if and only if ToString(ToUint32(_P_)) equals _P_ and ToUint32(_P_) is not the same value as 𝔽(<emu-eqn>2<sup>32</sup> - 1</emu-eqn>).</p>
       </emu-note>
 
-      <p>An object is an <dfn id="array-exotic-object">Array exotic object</dfn> (or simply, an Array object) if its [[DefineOwnProperty]] internal method uses the following implementation, and its other essential internal methods use the definitions found in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>. These methods are installed in ArrayCreate.</p>
+      <p>An object is an <dfn id="array-exotic-object" variants="Array exotic object">Array exotic object</dfn> (or simply, an Array object) if its [[DefineOwnProperty]] internal method uses the following implementation, and its other essential internal methods use the definitions found in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>. These methods are installed in ArrayCreate.</p>
```
michaelficarra, line 13665:
```suggestion
      <p>An object is an <dfn id="array-exotic-object" variants="Array exotic objects">Array exotic object</dfn> (or simply, an Array object) if its [[DefineOwnProperty]] internal method uses the following implementation, and its other essential internal methods use the definitions found in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>. These methods are installed in ArrayCreate.</p>
```

spec.html
```diff
@@ -42019,7 +42019,7 @@ <h1>FinalizationRegistry Objects</h1>
 
     <emu-clause id="sec-finalization-registry-constructor">
       <h1>The FinalizationRegistry Constructor</h1>
-      <p>The <dfn>FinalizationRegistry</dfn> constructor:</p>
+      <p>The <dfn variants="FinalizationRegistries">FinalizationRegistry</dfn> constructor:</p>
```
michaelficarra, line 42022: I'm still kind of partial to `FinalizationRegistrys`.

spec.html
michaelficarra: This is a double dfn. Shouldn't the linter complain?

michaelficarra (approved):
LGTM otherwise. I didn't review for any new unintended links, but I think we can just address those as we find them.

spec.html
michaelficarra: Let's just address it here. I don't think it's worth another issue/PR.

spec.html
bakkot: There is not currently a lint rule for definitions which occur multiple times, though yeah there probably should be.

PR #2493, opened 2021-08-20
Normative: Add relative indexing method at()

syg:
Adds Array#at, String#at, and %TypedArray%#at.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -948,7 +948,7 @@
       </ul>
 
       <p>In the language of this specification, numerical values are distinguished among different numeric kinds using subscript suffixes. The subscript <sub>𝔽</sub> refers to Numbers, and the subscript <sub>ℤ</sub> refers to BigInts. Numeric values without a subscript suffix refer to mathematical values.</p>
-      <p>Numeric operators such as +, &times;, =, and &ge; refer to those operations as determined by the type of the operands. When applied to mathematical values, the operators refer to the usual mathematical operations. When applied to Numbers, the operators refer to the relevant operations within IEEE 754-2019. When applied to BigInts, the operators refer to the usual mathematical operations applied to the mathematical value of the BigInt.</p>
+      <p>Numeric operators such as +, &times;, =, and &ge; refer to those operations as determined by the type of the operands. When applied to mathematical values, the operators refer to the usual mathematical operations. When applied to extended mathematical values, the operators refer to the usual mathematical operations over the extended real numbers; indeterminate forms are not defined and their use in this specification should be considered an editorial error. When applied to Numbers, the operators refer to the relevant operations within IEEE 754-2019. When applied to BigInts, the operators refer to the usual mathematical operations applied to the mathematical value of the BigInt.</p>
       <p>In general, when this specification refers to a numerical value, such as in the phrase, "the length of _y_" or "the integer represented by the four hexadecimal digits ...", without explicitly specifying a numeric kind, the phrase refers to a mathematical value. Phrases which refer to a Number or a BigInt value are explicitly annotated as such; for example, "the Number value for the number of code points in &hellip;" or "the BigInt value for &hellip;".</p>
       <p>Numeric operators applied to mixed-type operands (such as a Number and a mathematical value) are not defined and should be considered an editorial error in this specification.</p>
       <p>This specification denotes most numeric values in base 10; it also uses numeric values of the form 0x followed by digits 0-9 or A-F as base-16 values.</p>
@@ -13933,7 +13933,7 @@
           1. Assert: Type(_str_) is String.
           1. Let _len_ be the length of _str_.
           1. If ℝ(_index_) &lt; 0 or _len_ &le; ℝ(_index_), return *undefined*.
-          1. Let _resultStr_ be the String value of length 1, containing one code unit from _str_, specifically the code unit at index ℝ(_index_).
+          1. Let _resultStr_ be the substring of _str_ from ℝ(_index_) to ℝ(_index_) + 1.
           1. Return the PropertyDescriptor { [[Value]]: _resultStr_, [[Writable]]: *false*, [[Enumerable]]: *true*, [[Configurable]]: *false* }.
         </emu-alg>
       </emu-clause>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -32903,6 +32903,22 @@ <h1>Properties of the String Prototype Object</h1>
...
+          1. Let _O_ be ? RequireObjectCoercible(*this* value).
+          1. Let _S_ be ? ToString(_O_).
+          1. Let _len_ be the length of _S_.
+          1. Let _relativeIndex_ be ? ToIntegerOrInfinity(_index_).
+          1. If _relativeIndex_ &ge; 0, then
+            1. Let _k_ be _relativeIndex_.
+          1. Else,
+            1. Let _k_ be _len_ + _relativeIndex_.
+          1. If _k_ &lt; 0 or _k_ &ge; _len_, return *undefined*.
+          1. Return the String value of length 1, containing the code unit at index _k_ in _S_.
```
ljharb, line 32918: i wonder if both this and charAt could/should use `StringGetOwnProperty`?

syg: I don't think that adds any more clarity than this sentence, since none of the checking steps in StringGetOwnProperty are needed. But that points out that we should make the language to make a 1-length string consistent across all these uses, I thought it was just the two methods.

ljharb: Perhaps a simpler AO that this, charAt, and StringGetOwnProperty could call? I'm lightly hoping for something that doesn't rest on prose.

michaelficarra: Nothing wrong with using a consistent form in prose instead of an AO.

ljharb: It certainly works, but it's less clear to me that way.

syg: In this case it bottoms out at prose no matter what. StringGetOwnProperty bottoms out on prose. Unless you want to fully model strings as having some internal List of code units, which is not a project I think would add much right now.

ljharb: Fair point. It would certainly minimize the number of places that prose exists tho :-p yalls call either way, ofc.

syg:
This has test:

https://github.com/tc39/test262/tree/main/test/built-ins/Array/prototype/at
https://github.com/tc39/test262/tree/main/test/built-ins/String/prototype/at
https://github.com/tc39/test262/tree/main/test/built-ins/TypedArray/prototype/at

spec.html
```diff
@@ -36261,6 +36277,21 @@ <h1>Properties of the Array Prototype Object</h1>
...
+      <emu-clause id="sec-array.prototype.at">
+        <h1>Array.prototype.at ( _index_ )</h1>
+        <emu-alg>
+          1. Let _O_ be ? ToObject(*this* value).
+          1. Let _len_ be ? LengthOfArrayLike(_O_).
+          1. Let _relativeIndex_ be ? ToIntegerOrInfinity(_index_).
+          1. If _relativeIndex_ &ge; 0, then
+            1. Let _k_ be _relativeIndex_.
+          1. Else,
+            1. Let _k_ be _len_ + _relativeIndex_.
```
bakkot, line 36289: This can end up performing arithmetic on an infinity (specifically negative infinity), which we generally avoid so we don't have to define it. Can we guard against it here (and in the other methods), for example by adding

```
1. If _relativeIndex_ is not finite, return *undefined*.
```
after the `ToIntegerOrInfinity` step?

(Or we could define in [the relevant section](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-mathematical-operations) that the sum of an infinity and any finite value is the infinity, I guess.)

michaelficarra: I kind of prefer just permitting this arithmetic and not having to remember to avoid it everywhere.

syg: I'm fine with adding the step, but I do wonder why not define it. What are the concrete risks?

bakkot: Defining it seems fine, if we want to do that. I can make a PR or you can include it in this one.

zloirock:
@syg maybe you could review PRs in the proposal repo that waits for that for a half year, like https://github.com/tc39/proposal-relative-indexing-method/pull/44?

spec.html
jmdyck: You could use `the substring of _S_ from _k_ to _k_ + 1`. (And similarly in StringGetOwnProperty and charAt.)

syg: I like that suggestion, @jmdyck, given its prevalence in other string algorithms.

spec.html
syg: I've added a sentence that refers to the usual definition of arithmetic on extended reals. We could be more surgical here and only define e.g. + and - between a finite MV and an infinity value, if you think this is too handwavy, but I didn't see the need to explicitly give the definitions since we don't try to be formal with math underpinnings to begin with.

jmdyck: > When applied to extended mathematical values, the operators refer to the usual mathematical operations over the extended real numbers;

The spec doesn't define what "extended real numbers" are.

> indeterminate forms are undefined and their use in this specification should be considered an editorial error.

(You should maybe change "undefined" to "not defined", so that nobody thinks you mean `*undefined*`.)

Personally, I think I'd prefer to be told which operations involving infinities are allowed in the spec.

bakkot: > The spec doesn't define what "extended real numbers" are.

True, but it doesn't define "real numbers" either.

ljharb: so, to confirm, this means `-Infinity + finiteNumber === -Infinity`?

syg: Yeah, I think here's where we need to make an editorial decision for the level of mathematical background we can assume. We can, for example, reproduce these definitions [from Wikipedia](https://en.wikipedia.org/wiki/Extended_real_number_line#Arithmetic_operations) in the spec, along with definitions of &ge; etc. But I'm not convinced that level of formality is needed. I think it's reasonable to reference a searchable mathematical phrase, like "extended real numbers" to make the information easily searchable for unfamiliar folks.

syg: > so, to confirm, this means -Infinity + finiteNumber === -Infinity?

Yeah

jmdyck: I think a Note with an explicit link (to Wikipedia, or whatever) would help.

syg: I'll bring this up for discussion at the next editor call.

michaelficarra (approved):

PR #2500, opened 2021-08-26
Editorial: tweak note in FunctionDeclarationInstantiation

bakkot:
I noticed in [this thread](https://es.discourse.group/t/why-are-parameter-bindings-of-non-strict-functions-instantiated-in-a-new-environment-if-default-value-initialisers-exist/941/12) that [FunctionDeclarationInstantiation](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-functiondeclarationinstantiation) step 19.a is misleading. It reads

> 19. If strict is true or if hasParameterExpressions is false, then
>   a. NOTE: Only a single Environment Record is needed for the parameters and top-level vars.

The NOTE is identical to that in step 27.a, from which it was presumably copied:

> 27. If hasParameterExpressions is false, then
>    a. NOTE: Only a single Environment Record is needed for the parameters and top-level vars.

But the conditions are different, and in a relevant way. Step 19 actually has nothing to do with top-level `var` declarations; rather, it's about where to put `var` declarations arising from sloppy-mode direct calls to `eval` in parameter expressions. (Step 27/28 is what governs where to put `var` declarations from the function body.)

So the NOTE in step 19 is misleading. I've changed it to talk about `eval` instead.

This was introduced in #1046. See also #623 for discussion of whether to keep these branches at all - they're not actually observable to user code (that is, taking the "else" branch in both cases is equivalent to the current state). I think we should probably pursue that eventually, but in the mean time the notes should at least be correct.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13369,7 +13369,7 @@
           1. If *"arguments"* is an element of _functionNames_ or if *"arguments"* is an element of _lexicalNames_, then
             1. Set _argumentsObjectNeeded_ to *false*.
         1. If _strict_ is *true* or if _hasParameterExpressions_ is *false*, then
-          1. NOTE: Only a single Environment Record is needed for the parameters and top-level vars.
+          1. NOTE: Only a single Environment Record is needed for the parameters, since calls to `eval` in strict mode code cannot create new bindings which are visible outside of the `eval`.
           1. Let _env_ be the LexicalEnvironment of _calleeContext_.
         1. Else,
           1. NOTE: A separate Environment Record is needed to ensure that bindings created by direct eval calls in the formal parameter list are outside the environment where parameters are declared.


meena-kaizad:
@bakkot not sure if you'd want to include this for completeness, but step 19 is also needed to preserve backward compatibility with ES5 (see [comment](http://dmitrysoshnikov.com/ecmascript/chapter-5-functions/) by Dmitry Soshnikov on 26/08). Parameter defaults didn’t exist back then, so we'd have to use manual defaults like:

```javascript
function foo(x, y) {
  if (typeof y == 'undefined') {
    y = function() { x = 2; };
  }
  var x = 3;
  y(); // is `x` shared?
  console.log(x); // yes! 2
}
 
foo();
```

bakkot:
@meena-kaizad In the presence of parameter expressions step 27/28 would create a separate scope anyway, so step 19 doesn't really have any bearing on the backwards compatibility concern.

But even if step 19 were eliminated and step 20 made mandatory, _and_ step 27 were eliminated and step 28 made mandatory, backwards compatibility would still be maintained by 28.e. Your example doesn't get at this because it doesn't attempt to read from `x` before it's assigned to in the body (remember, `var` declarations are hoisted to the top of the containing scope), but you can observe it with

```js
function foo(x) {
  console.log(x);
  var x = 2;
}
foo(1);
```
which prints `1` rather than `undefined`, as it would in the absence of 28.e. This happens whether or not you have parameter expressions:

```js
function foo(x, y = 'an expression') {
  console.log(x);
  var x = 2;
}
foo(1);
```
also prints `1`.

You can actually observe the fact that there's two scopes with

```js
function foo(x, probeX = () => x) {
  console.log(x);
  var x = 2;
  console.log(x);
  console.log(probeX());
}
 
foo(1);
```
which prints `1`, `2`, `1` - but of course you can't observe this without parameter expressions, so it's a valid optimization to collapse it into a single scope if there are no parameter expressions. But it's _just_ an optimization; the backwards compatible behavior would be maintained either way by 28.e.

meena-kaizad:
> But it's _just_ an optimization; the backwards compatible behavior would be maintained either way by 28.e.

Oh right, I forgot about 27/28, Thanks so much for clearing that up @bakkot.

syg (approved):

PR #2504, opened 2021-08-29
Editorial: make the order of parameters for class field AOs more consistent

bakkot:
Various of the abstract operations introduced in #1668 took their parameters in the order (key, object) or (key, object, value), rather than the prevailing convention of (object, key) or (object, key, value) respectively (as in [Get](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-get-o-p), [Set](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-set-o-p-v-throw), etc). (Thanks to @devsnek for [pointing this out](https://matrixlogs.bakkot.com/TC39_General/2021-07-17#L10).)

This PR swaps the order of the parameters for those methods to match the prior convention.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6595,8 +6595,8 @@
     <emu-clause id="sec-privateelementfind" type="abstract operation">
       <h1>
         PrivateElementFind (
-          _P_: a Private Name,
           _O_: an Object,
+          _P_: a Private Name,
         )
       </h1>
       <dl class="header">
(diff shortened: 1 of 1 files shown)


liannemarie:
> Various of the abstract operations introduced in #1668 took their parameters in the order (key, object) or (key, object, value), rather than the prevailing convention of (object, key) or (object, key, value) respectively (as in [Get](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-get-o-p), [Set](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-set-o-p-v-throw), etc). (Thanks to @devsnek for [pointing this out](https://matrixlogs.bakkot.com/TC39_General/2021-07-17#L10).)
> 
> This PR swaps the order of the parameters for those methods to match the prior convention.

This is a hacked acct. Please remove illegal access

michaelficarra (approved):

PR #2505, opened 2021-08-29
Normative: add missing AsyncGenerator cases to ContainsArguments

bakkot:
ContainsArguments, added in #1668, is used to forbid class field initializers which refer to the outer `arguments` binding. It's not intended to descend into non-arrow functions, since those introduce their own binding for `arguments`, which is fine. Unfortunately it was accidentally missing the `AsyncGenerator` cases, meaning that, as written, `(class { x = async function*(){ arguments } })` was an Early Error, unlike `(class { x = function*(){ arguments } })` or `(class { x = async function(){ arguments } })`.

Engine262 actually implements the specified behavior, but no other engine I tested did.

This is technically normative, because the current semantics are coherent, but I think we need not seek consensus for it - it was definitely just an oversight.

The first commit is just a reorganization of the SDO to make it easier to read. I suggest reviewing (and landing) the commits separately.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -23967,11 +23967,26 @@
       </emu-alg>
 
       <emu-grammar>
-        FunctionDeclaration : `function` BindingIdentifier `(` FormalParameters `)` `{` FunctionBody `}`
+        FunctionDeclaration :
+          `function` BindingIdentifier `(` FormalParameters `)` `{` FunctionBody `}`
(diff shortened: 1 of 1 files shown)


bakkot:
I disagree with the `needs tests` label, since this is just a bugfix. @syg thoughts?

liannemarie:
Unauth access. Please remove access

michaelficarra (approved):

PR #2510, opened 2021-09-11
Editorial: Make the behavior of SameValue and SameValueZero a little more clear

gibson042:
Every time I encounter a reference to one of the two operations, there's a momentary pause while I remember with differentiates +0 from -0. This should shorten that time by a bit, and make the text slightly more approachable for everyone (especially since `::sameValue` and `::sameValueZero` are not currently links).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1433,6 +1433,7 @@
               _T_::sameValue(x,&nbsp;y)
             </td>
             <td>
+              `Object.is(x, y)`
             </td>
             <td>
               Object internal methods,
@@ -1448,11 +1449,12 @@
               _T_::sameValueZero(x,&nbsp;y)
             </td>
             <td>
+              `[x].includes(y)`
             </td>
             <td>
               Array, Map, and Set methods,
               via <emu-xref href="#sec-samevaluezero" title></emu-xref>,
-              to test value equality ignoring differences among members of the zero cohort (i.e., *-0*<sub>𝔽</sub> and *+0*<sub>𝔽</sub>)
+              to test value equality ignoring differences between Numbers in the zero cohort (i.e., *-0*<sub>𝔽</sub> and *+0*<sub>𝔽</sub>)
             </td>
             <td>
               Boolean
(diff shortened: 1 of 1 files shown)


bakkot (approved):
I love it.

spec.html
```diff
@@ -1448,11 +1449,12 @@ <h1>Numeric Types</h1>
               _T_::sameValueZero(x,&nbsp;y)
             </td>
             <td>
+              `[x].includes(y)`
```
ljharb, line 1452: this seems much less clear than `x === y` - which is recursive, but so is `.includes`
```suggestion
              `x === y`
```

bakkot: `x === y` does not use `sameValueZero` - `===` considers `NaN` as being unequal to itself. `[NaN].includes(NaN)` holds, though, and so is appropriate here.

ljharb: ah, right, forgot about NaN.

This doesn't link to `Array.prototype.includes` though, just like `T::sameValue` etc don't link - so i'm not sure how much clearer this is if i still have to find-in-page to understand the difference?

bakkot: A reader might remember how the JS operations work but not the spec AOs.

Besides, these were the only two empty cells in this column - might as well fill them out.

ljharb: fair enough

spec.html
```diff
@@ -5896,7 +5898,7 @@ <h1>
         1. Return ! SameValueNonNumeric(_x_, _y_).
       </emu-alg>
       <emu-note>
-        <p>SameValueZero differs from SameValue only in its treatment of *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub>.</p>
+        <p>SameValueZero differs from SameValue only by treating *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub> as equivalent.</p>
```
michaelficarra, line 5901:
```suggestion
        <p>SameValueZero differs from SameValue only in that it treats *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub> as equivalent.</p>
```

spec.html
```diff
@@ -5887,7 +5889,7 @@ <h1>
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It returns a completion record whose [[Type]] is ~normal~ and whose [[Value]] is a Boolean.</dd>
+        <dd>It returns a completion record whose [[Type]] is ~normal~ and whose [[Value]] is a Boolean indicating whether or not the two arguments are the same value or are both zero-valued Numbers.</dd>
```
michaelficarra, line 5892:
```suggestion
        <dd>It returns a completion record whose [[Type]] is ~normal~ and whose [[Value]] is a Boolean indicating whether or not the two arguments are the same value or are both either *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>.</dd>
```

michaelficarra (approved):
LGTM otherwise

gibson042:
Accepted suggestions from @michaelficarra with tweaks.

PR #2512, opened 2021-09-12
Markup: have SDOs use structured headers

bakkot:
~Depends on https://github.com/tc39/ecmarkup/pull/358 and will not build until that's upstreamed. I've uploaded a rendered version built with that PR [here](https://bakkot.github.io/ecma262-previews/sdo-structured-header/multipage/).~

This now includes ecmarkup 9.2.0, which introduces support for this.

In a followup, now that the format accommodates it, I'll start adding descriptions and parameter types.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4957,9 +4957,13 @@
           </emu-alg>
         </emu-clause>
 
-        <emu-clause id="sec-runtime-semantics-stringnumericvalue" type="sdo" aoid="StringNumericValue" oldids="sec-runtime-semantics-mv-s">
+        <emu-clause id="sec-runtime-semantics-stringnumericvalue" type="sdo" oldids="sec-runtime-semantics-mv-s">
           <h1>Runtime Semantics: StringNumericValue</h1>
-          <p>The conversion of a |StringNumericLiteral| to a Number value is similar overall to the determination of the NumericValue of a |NumericLiteral| (see <emu-xref href="#sec-literals-numeric-literals"></emu-xref>), but some of the details are different.</p>
+          <dl class="header">
+          </dl>
+          <emu-note>
+            <p>The conversion of a |StringNumericLiteral| to a Number value is similar overall to the determination of the NumericValue of a |NumericLiteral| (see <emu-xref href="#sec-literals-numeric-literals"></emu-xref>), but some of the details are different.</p>
+          </emu-note>
           <emu-grammar>StringNumericLiteral ::: StrWhiteSpace?</emu-grammar>
           <emu-alg>
             1. Return *+0*<sub>𝔽</sub>.
@@ -6980,8 +6984,10 @@
   <emu-clause id="sec-syntax-directed-operations-scope-analysis">
     <h1>Scope Analysis</h1>
 
-    <emu-clause id="sec-static-semantics-boundnames" oldids="sec-identifiers-static-semantics-boundnames,sec-let-and-const-declarations-static-semantics-boundnames,sec-variable-statement-static-semantics-boundnames,sec-destructuring-binding-patterns-static-semantics-boundnames,sec-for-in-and-for-of-statements-static-semantics-boundnames,sec-function-definitions-static-semantics-boundnames,sec-arrow-function-definitions-static-semantics-boundnames,sec-generator-function-definitions-static-semantics-boundnames,sec-async-generator-function-definitions-static-semantics-boundnames,sec-class-definitions-static-semantics-boundnames,sec-async-function-definitions-static-semantics-BoundNames,sec-async-arrow-function-definitions-static-semantics-BoundNames,sec-imports-static-semantics-boundnames,sec-exports-static-semantics-boundnames" type="sdo" aoid="BoundNames">
+    <emu-clause id="sec-static-semantics-boundnames" oldids="sec-identifiers-static-semantics-boundnames,sec-let-and-const-declarations-static-semantics-boundnames,sec-variable-statement-static-semantics-boundnames,sec-destructuring-binding-patterns-static-semantics-boundnames,sec-for-in-and-for-of-statements-static-semantics-boundnames,sec-function-definitions-static-semantics-boundnames,sec-arrow-function-definitions-static-semantics-boundnames,sec-generator-function-definitions-static-semantics-boundnames,sec-async-generator-function-definitions-static-semantics-boundnames,sec-class-definitions-static-semantics-boundnames,sec-async-function-definitions-static-semantics-BoundNames,sec-async-arrow-function-definitions-static-semantics-BoundNames,sec-imports-static-semantics-boundnames,sec-exports-static-semantics-boundnames" type="sdo">
       <h1>Static Semantics: BoundNames</h1>
+      <dl class="header">
+      </dl>
       <emu-note>
         <p>*"\*default\*"* is used within this specification as a synthetic name for hoistable anonymous functions that are defined using export declarations.</p>
       </emu-note>
@@ -7229,8 +7235,10 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-static-semantics-declarationpart" type="sdo" aoid="DeclarationPart">
+    <emu-clause id="sec-static-semantics-declarationpart" type="sdo">
       <h1>Static Semantics: DeclarationPart</h1>
+      <dl class="header">
+      </dl>
       <emu-grammar>HoistableDeclaration : FunctionDeclaration</emu-grammar>
       <emu-alg>
         1. Return |FunctionDeclaration|.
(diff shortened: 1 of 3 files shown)


spec.html
```diff
@@ -8118,9 +8146,10 @@ <h1>Static Semantics: TopLevelVarScopedDeclarations</h1>
...
-      <h1>Static Semantics: ContainsDuplicateLabels</h1>
-      <p>With parameter _labelSet_.</p>
+    <emu-clause id="sec-static-semantics-containsduplicatelabels" oldids="sec-statement-semantics-static-semantics-containsduplicatelabels,sec-block-static-semantics-containsduplicatelabels,sec-if-statement-static-semantics-containsduplicatelabels,sec-do-while-statement-static-semantics-containsduplicatelabels,sec-while-statement-static-semantics-containsduplicatelabels,sec-for-statement-static-semantics-containsduplicatelabels,sec-for-in-and-for-of-statements-static-semantics-containsduplicatelabels,sec-with-statement-static-semantics-containsduplicatelabels,sec-switch-statement-static-semantics-containsduplicatelabels,sec-labelled-statements-static-semantics-containsduplicatelabels,sec-try-statement-static-semantics-containsduplicatelabels,sec-function-definitions-static-semantics-containsduplicatelabels,sec-module-semantics-static-semantics-containsduplicatelabels" type="sdo">
+      <h1>Static Semantics: ContainsDuplicateLabels ( _labelSet_ )</h1>
```
jmdyck, line 8150: I notice that some SDOs (e.g. EvaluateBody) have a 545-like header, with one parameter per line. But some (like this one), although they replace the "With parameter" paragraph with a parameter list in the `<h1>`, don't put it in 545 form. Is this form just transitional, draft-only?

bakkot: Ah, yeah, I didn't bother making them multiline when there were no types or optional parameters. But since the intent is to eventually list types for everything, I suppose we might as well make them multiline now.

spec.html
```diff
@@ -16649,9 +16744,12 @@ <h1>Static Semantics: Early Errors</h1>
...
+      <emu-clause id="sec-static-semantics-sv" oldids="sec-string-literals-static-semantics-stringvalue,sec-additional-syntax-string-literals-static-semantics" type="sdo">
         <h1>Static Semantics: SV</h1>
-        <p>A string literal stands for a value of the String type. SV produces String values for string literals through recursive application on the various parts of the string literal. As part of this process, some Unicode code points within the string literal are interpreted as having a mathematical value, as described below or in <emu-xref href="#sec-literals-numeric-literals"></emu-xref>.</p>
+        <dl class="header">
+          <dt>description</dt>
+          <dd><p>A string literal stands for a value of the String type. SV produces String values for string literals through recursive application on the various parts of the string literal. As part of this process, some Unicode code points within the string literal are interpreted as having a mathematical value, as described below or in <emu-xref href="#sec-literals-numeric-literals"></emu-xref>.</p></dd>
```
michaelficarra, line 16751:
```suggestion
          <dd>A string literal stands for a value of the String type. SV produces String values for string literals through recursive application on the various parts of the string literal. As part of this process, some Unicode code points within the string literal are interpreted as having a mathematical value, as described below or in <emu-xref href="#sec-literals-numeric-literals"></emu-xref>.</dd>
```

We don't wrap these in `<p>` elsewhere.

bakkot: That's because these descriptions are not normally an entire paragraph of text. Wrapping in `<p>` causes this part of the description to be set off in its own paragraph, with the "The syntax-directed operation X takes Y." and "It is defined piecewise over the following productions" parts as whole paragraphs before and after the description. That read better to me, which is why I've put this in `<p>`.

michaelficarra: It seems weird that it's typically a text node which is inserted as a child of the summary `<p>`, but when it's a `<p>`, it becomes a sibling.

bakkot: Does it? It seems like the obvious behavior to me.

michaelficarra (approved):
LGTM otherwise.

spec.html
bakkot: Discussed offline; @michaelficarra said he was OK with this.

spec.html
```diff
@@ -6980,8 +6984,12 @@ <h1>Syntax-Directed Operations</h1>
...
+    <emu-clause id="sec-static-semantics-boundnames" oldids="sec-identifiers-static-semantics-boundnames,sec-let-and-const-declarations-static-semantics-boundnames,sec-variable-statement-static-semantics-boundnames,sec-destructuring-binding-patterns-static-semantics-boundnames,sec-for-in-and-for-of-statements-static-semantics-boundnames,sec-function-definitions-static-semantics-boundnames,sec-arrow-function-definitions-static-semantics-boundnames,sec-generator-function-definitions-static-semantics-boundnames,sec-async-generator-function-definitions-static-semantics-boundnames,sec-class-definitions-static-semantics-boundnames,sec-async-function-definitions-static-semantics-BoundNames,sec-async-arrow-function-definitions-static-semantics-BoundNames,sec-imports-static-semantics-boundnames,sec-exports-static-semantics-boundnames" type="sdo">
       <h1>Static Semantics: BoundNames</h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd></dd>
```
jmdyck, line 6991: `<dd></dd>` is odd. Could just drop the `<dt><dd>` pair.

spec.html
jmdyck: [This is a new "conversation", but GitHub doesn't recognize that.]
```suggestion
          <dd>
            <p>A string literal stands for a value of the String type. SV produces String values for string literals through recursive application on the various parts of the string literal. As part of this process, some Unicode code points within the string literal are interpreted as having a mathematical value, as described below or in <emu-xref href="#sec-literals-numeric-literals"></emu-xref>.</p>
          </dd>
```
The only other example of `<p>` within `<dd>` is at `ResolveExport`, which formats it this way. (Of course, it has to, because there are 2 such `<p>` elements. Still, given the difference in rendering, I think it makes sense to format the 1 `<p>` case more like the 2 `<p>` case than the 0 `<p>` case.)

spec.html
```diff
@@ -22949,19 +23214,30 @@ <h1>Static Semantics: ConciseBodyContainsUseStrict</h1>
...
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-runtime-semantics-instantiatearrowfunctionexpression" type="sdo" aoid="InstantiateArrowFunctionExpression">
-      <h1>Runtime Semantics: InstantiateArrowFunctionExpression</h1>
-      <p>With optional parameter _name_.</p>
+    <emu-clause id="sec-runtime-semantics-instantiatearrowfunctionexpression" type="sdo">
+      <h1>
+        Runtime Semantics: InstantiateArrowFunctionExpression (
+          optional _name_: unknown
```
jmdyck, line 23236:
```suggestion
          optional _name_: unknown,
```

spec.html
```diff
@@ -23538,9 +23859,12 @@ <h1>Runtime Semantics: InstantiateAsyncGeneratorFunctionObject</h1>
...
 
-    <emu-clause id="sec-runtime-semantics-instantiateasyncgeneratorfunctionexpression" type="sdo" aoid="InstantiateAsyncGeneratorFunctionExpression">
-      <h1>Runtime Semantics: InstantiateAsyncGeneratorFunctionExpression</h1>
-      <p>With optional parameter _name_.</p>
+    <emu-clause id="sec-runtime-semantics-instantiateasyncgeneratorfunctionexpression" type="sdo">
+      <h1>
+        Runtime Semantics: InstantiateAsyncGeneratorFunctionExpression (
+          optional _name_: unknown,
+        )
+      </h1>
```
jmdyck, line 23867: Insert `<dl>` element.

spec.html
```diff
@@ -24548,9 +24941,12 @@ <h1>Runtime Semantics: EvaluateAsyncConciseBody</h1>
...
 
-    <emu-clause id="sec-runtime-semantics-instantiateasyncarrowfunctionexpression" type="sdo" aoid="InstantiateAsyncArrowFunctionExpression">
-      <h1>Runtime Semantics: InstantiateAsyncArrowFunctionExpression</h1>
-      <p>With optional parameter _name_.</p>
+    <emu-clause id="sec-runtime-semantics-instantiateasyncarrowfunctionexpression" type="sdo">
+      <h1>
+        Runtime Semantics: InstantiateAsyncArrowFunctionExpression (
+          optional _name_: unknown,
+        )
+      </h1>
```
jmdyck, line 24949: Insert `<dl>` element.

jmdyck (changes requested):

jmdyck (approved):

swanprincess386:
> ~Depends on [tc39/ecmarkup#358](https://github.com/tc39/ecmarkup/pull/358) and will not build until that's upstreamed. I've uploaded a rendered version built with that PR [here](https://bakkot.github.io/ecma262-previews/sdo-structured-header/multipage/).~
> 
> This now includes ecmarkup 9.2.0, which introduces support for this.
> 
> In a followup, now that the format accommodates it, I'll start adding descriptions and parameter types.

PR #2523, opened 2021-09-17
Normative: require Number::exponentiate to return a Number value

michaelficarra:
This is normative but I believe it was always what was intended, so we shouldn't need consensus on it.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1622,7 +1622,7 @@
               1. If abs(ℝ(_base_)) &lt; 1, return *+&infin;*<sub>𝔽</sub>.
             1. Assert: _exponent_ is finite and is neither *+0*<sub>𝔽</sub> nor *-0*<sub>𝔽</sub>.
             1. If _base_ &lt; *+0*<sub>𝔽</sub> and _exponent_ is not an integral Number, return *NaN*.
-            1. Return an implementation-approximated value representing the result of raising ℝ(_base_) to the ℝ(_exponent_) power.
+            1. Return an implementation-approximated Number value representing the result of raising ℝ(_base_) to the ℝ(_exponent_) power.
           </emu-alg>
           <emu-note>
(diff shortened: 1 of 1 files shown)


bakkot:
Every math function also returns "an implementation-approximated value" rather than "an implementation-approximated Number value". I think it would be better for "implementation-approximated" to imply Number. That's how I understood it, certainly.

michaelficarra:
I don't think it's clear that "implementation-approximated value" implies that it's a Number value. I'll add the other cases to this PR.

spec.html
```diff
@@ -28182,7 +28182,7 @@ <h1>parseInt ( _string_, _radix_ )</h1>
...
         1. If _Z_ is empty, return *NaN*.
-        1. Let _mathInt_ be the integer value that is represented by _Z_ in radix-_R_ notation, using the letters <b>A</b>-<b>Z</b> and <b>a</b>-<b>z</b> for digits with values 10 through 35. (However, if _R_ is 10 and _Z_ contains more than 20 significant digits, every significant digit after the 20th may be replaced by a 0 digit, at the option of the implementation; and if _R_ is not 2, 4, 8, 10, 16, or 32, then _mathInt_ may be an implementation-approximated value representing the integer value that is represented by _Z_ in radix-_R_ notation.)
+        1. Let _mathInt_ be the integer value that is represented by _Z_ in radix-_R_ notation, using the letters <b>A</b>-<b>Z</b> and <b>a</b>-<b>z</b> for digits with values 10 through 35. (However, if _R_ is 10 and _Z_ contains more than 20 significant digits, every significant digit after the 20th may be replaced by a 0 digit, at the option of the implementation; and if _R_ is not 2, 4, 8, 10, 16, or 32, then _mathInt_ may be an implementation-approximated integer representing the integer value denoted by _Z_ in radix-_R_ notation.)
```
michaelficarra, line 28185: On this line, I've also adopted the "denoted by" wording we use in `RoundMVResult`.

ljharb:
Will there ever be implementation-approximated non-Number values?

michaelficarra:
@ljharb There is one in this PR.

ljharb:
ah, an integer MV. and you think the boilerplate of "Number" is useful versus having it be the default, given its frequency?

michaelficarra:
Yeah, I wouldn't want everything described as "implementation-approximated" to also be implied to be a Number value, even if they all were.

PR #2528, opened 2021-09-21
Editorial: inline infallible CreateDataPropertyOrThrow calls

michaelficarra:
There's no reason to call `CreateDataPropertyOrThrow` when it cannot fail (marked with `!`). Instead, call `CreateDataProperty` directly.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4410,17 +4410,17 @@
           1. Let _obj_ be ! OrdinaryObjectCreate(%Object.prototype%).
           1. Assert: _obj_ is an extensible ordinary object with no own properties.
           1. If _Desc_ has a [[Value]] field, then
-            1. Perform ! CreateDataPropertyOrThrow(_obj_, *"value"*, _Desc_.[[Value]]).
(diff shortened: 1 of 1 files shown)


bakkot (changes requested):
This no longer indicates to the reader that the property is guaranteed to be created, since `CreateDataProperty` can return false and there is nothing to indicate to the reader that it does not in these cases. I'm opposed.

ljharb:
Indeed, `CreateDataProperty` can return false. This caused many spec bugs in the past when authors didn't realize that this needed to be explicitly checked, which was why they were all explicitly converted to `! CreateDataPropertyOrThrow`.

In this case, your change would require an assertion that the result was `true`, which is currently lacking.

michaelficarra:
Result from editor call today is that we should make a new AO for infallible `CreateDataProperty` and use that instead to make it more clear, but we'll wait to do that until after completion record reform is complete.

ljharb:
@michaelficarra should this be left open until then? or closed?

michaelficarra:
let's leave it open and do the work here

PR #2521, opened 2021-09-16
Meta: update static default branch name to a dynamic one

ljharb:
This only leaves the word "master" hardcoded in build scripts, and contributor instructions prose.

PR #2532, opened 2021-09-23
Meta: Keep old years in gh-pages

### LINKED ISSUE #2529: need `/$YEAR/` URLs
Whenever we update https://tc39.es/ecma262/, we should also ensure that the following URLs exist and work:
 - https://tc39.es/ecma262/2021/
 - https://tc39.es/ecma262/2020/
 - https://tc39.es/ecma262/2019/
 - https://tc39.es/ecma262/2018/
 - https://tc39.es/ecma262/2017/
 - https://tc39.es/ecma262/2016/
 - https://tc39.es/ecma262/2015/ (if possible, but maybe not since this was never generated by ecmarkup)
 
 It's fine if past years are not updated once the edition is cut, since ecmarkup evolves and the spec snapshot does not.

bakkot:
I updated the [gh-pages branch](https://github.com/tc39/ecma262/tree/gh-pages) with es2016-es2021 subdirectories (including some minor tweaks, which I pushed to the corresponding branches).

This updates the auto-deploy script to preserve those subdirectories.

It also fixes it - #2521 changed the branch name for the workflow to `$default-branch`, which [was a good idea](https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/), but apparently that syntax [only works in workflow templates](https://github.com/actions/starter-workflows/pull/590#issuecomment-672360634). So we haven't been updating the rendered spec since that PR landed.

Fixes #2529.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,9 +1,9 @@
-name: 'ecma-262'
+name: 'ecma-262 deploy'
 
 on:
   push:
     branches:
-      - $default-branch
+      - master
 
 jobs:
   deploy:
--- a/scripts/auto-deploy.sh
+++ b/scripts/auto-deploy.sh
@@ -8,12 +8,14 @@ declare -r GH_USER_NAME="Bot"
 declare -r COMMIT_MESSAGE="Update gh-pages"
 
 
-cd "$(dirname "$BASH_SOURCE")"/../out
+cd "$(dirname "$BASH_SOURCE")"/..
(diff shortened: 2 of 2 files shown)


ljharb:
whoops, good catch.

bakkot:
Oh, I should mention that I used subdirectories named `es20XX`, rather than just `20XX`. I could switch to just `20XX` if we'd prefer (or if that's what we used to do).

ljharb:
The old links always used just the year, so to unbreak the old URLs we'd need a redirect at least anyways, so I'd prefer the simplicity of just always only doing the year.

bakkot:
Works for me, done.

michaelficarra (approved):

PR #2536, opened 2021-09-28
Editorial: `String.prototype.lastIndexOf`: use explicit algorithm steps

ljharb:
See also, https://github.com/tc39/ecma262/pull/2258

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33428,9 +33428,12 @@ THH:mm:ss.sss
           1. If _numPos_ is *NaN*, let _pos_ be +&infin;; otherwise, let _pos_ be ! ToIntegerOrInfinity(_numPos_).
           1. Let _len_ be the length of _S_.
           1. Let _start_ be the result of clamping _pos_ between 0 and _len_.
+          1. If _searchStr_ is the empty String, return 𝔽(_start_).
           1. Let _searchLen_ be the length of _searchStr_.
-          1. Let _k_ be the largest possible non-negative integer not larger than _start_ such that _k_ + _searchLen_ &le; _len_, and for all non-negative integers _j_ such that _j_ &lt; _searchLen_, the code unit at index _k_ + _j_ within _S_ is the same as the code unit at index _j_ within _searchStr_; but if there is no such integer, let _k_ be -1.
-          1. Return 𝔽(_k_).
+          1. For each non-negative integer _i_ starting with _start_ such that _i_ &le; _len_ - _searchLen_, in descending order, do
+            1. Let _candidate_ be the substring of _string_ from _i_ to _i_ + _searchLen_.
+            1. If _candidate_ is the same sequence of code units as _searchStr_, return 𝔽(_i_).
+          1. Return *-1*<sub>𝔽</sub>.
         </emu-alg>
         <emu-note>
           <p>The `lastIndexOf` function is intentionally generic; it does not require that its *this* value be a String object. Therefore, it can be transferred to other kinds of objects for use as a method.</p>


spec.html
```diff
@@ -33394,9 +33394,12 @@ <h1>String.prototype.lastIndexOf ( _searchString_ [ , _position_ ] )</h1>
...
           1. Let _len_ be the length of _S_.
           1. Let _start_ be the result of clamping _pos_ between 0 and _len_.
+          1. If _searchStr_ is the empty String and _start_ &le; _len_, return 𝔽(_start_)_.
           1. Let _searchLen_ be the length of _searchStr_.
-          1. Let _k_ be the largest possible non-negative integer not larger than _start_ such that _k_ + _searchLen_ &le; _len_, and for all non-negative integers _j_ such that _j_ &lt; _searchLen_, the code unit at index _k_ + _j_ within _S_ is the same as the code unit at index _j_ within _searchStr_; but if there is no such integer, let _k_ be -1.
-          1. Return 𝔽(_k_).
+          1. For each integer _i_ starting with _start_ such that _i_ &le; _len_ - _searchLen_, in descending order, do
+            1. Let _candidate_ be the substring of _string_ from _i_ to _i_ + _searchLen_.
+            1. If _candidate_ is the same sequence of code units as _searchStr_, return 𝔽(_i_).
+          1. Return 𝔽(-1).
```
bakkot, line 33402:
```suggestion
          1. Return *-1*<sub>𝔽</sub>.
```

ljharb: aha, thanks.

spec.html
```diff
@@ -33428,9 +33428,12 @@ <h1>String.prototype.lastIndexOf ( _searchString_ [ , _position_ ] )</h1>
           1. If _numPos_ is *NaN*, let _pos_ be +&infin;; otherwise, let _pos_ be ! ToIntegerOrInfinity(_numPos_).
           1. Let _len_ be the length of _S_.
           1. Let _start_ be the result of clamping _pos_ between 0 and _len_.
+          1. If _searchStr_ is the empty String and _start_ &le; _len_, return 𝔽(_start_)_.
```
bakkot, line 33431: `_start_ &le; _len_` always holds, because we've constrained it to be so on the previous line. So just

```suggestion
          1. If _searchStr_ is the empty String, return 𝔽(_start_)_.
```

(This line isn't actually necessary, nor the similar line in `StringIndexOf`, but it seems fine to keep it.)

spec.html
```diff
@@ -33428,9 +33428,12 @@ <h1>String.prototype.lastIndexOf ( _searchString_ [ , _position_ ] )</h1>
           1. If _numPos_ is *NaN*, let _pos_ be +&infin;; otherwise, let _pos_ be ! ToIntegerOrInfinity(_numPos_).
           1. Let _len_ be the length of _S_.
           1. Let _start_ be the result of clamping _pos_ between 0 and _len_.
+          1. If _searchStr_ is the empty String, return 𝔽(_start_).
           1. Let _searchLen_ be the length of _searchStr_.
-          1. Let _k_ be the largest possible non-negative integer not larger than _start_ such that _k_ + _searchLen_ &le; _len_, and for all non-negative integers _j_ such that _j_ &lt; _searchLen_, the code unit at index _k_ + _j_ within _S_ is the same as the code unit at index _j_ within _searchStr_; but if there is no such integer, let _k_ be -1.
-          1. Return 𝔽(_k_).
+          1. For each integer _i_ starting with _start_ such that _i_ &le; _len_ - _searchLen_, in descending order, do
```
michaelficarra, line 33433:
```suggestion
          1. For each non-negative integer _i_ starting with _start_ such that _i_ &le; _len_ - _searchLen_, in descending order, do
```

or

```suggestion
          1. For each integer _i_ starting with _start_ such that 0 &le; _i_ &le; _len_ - _searchLen_, in descending order, do
```

ljharb: went with "and" instead of a semicolon, lmk your thoughts

michaelficarra: Oh, I didn't mean to suggest both. I was okay with either, but both is unnecessary. And my latter suggestion used chained comparison, not a semicolon.

ljharb: oh hmm. i'll stick with the first suggestion then.

michaelficarra (approved):

spec.html
```diff
@@ -33428,9 +33428,12 @@ <h1>String.prototype.lastIndexOf ( _searchString_ [ , _position_ ] )</h1>
...
           1. Let _len_ be the length of _S_.
           1. Let _start_ be the result of clamping _pos_ between 0 and _len_.
+          1. If _searchStr_ is the empty String, return 𝔽(_start_).
           1. Let _searchLen_ be the length of _searchStr_.
-          1. Let _k_ be the largest possible non-negative integer not larger than _start_ such that _k_ + _searchLen_ &le; _len_, and for all non-negative integers _j_ such that _j_ &lt; _searchLen_, the code unit at index _k_ + _j_ within _S_ is the same as the code unit at index _j_ within _searchStr_; but if there is no such integer, let _k_ be -1.
-          1. Return 𝔽(_k_).
+          1. For each non-negative integer _i_ starting with _start_ such that _i_ &le; _len_ - _searchLen_, in descending order, do
+            1. Let _candidate_ be the substring of _string_ from _i_ to _i_ + _searchLen_.
+            1. If _candidate_ is the same sequence of code units as _searchStr_, return 𝔽(_i_).
+            1. Return *-1*<sub>𝔽</sub>.
```
bakkot, line 33436:
```suggestion
          1. Return *-1*<sub>𝔽</sub>.
```

ljharb: oof, thanks. moving too quickly.

bakkot (approved):

PR #2546, opened 2021-10-07
Editorial: refer to numeric methods as we do other AOs

michaelficarra:
This leaves the `::` in the names for now to make the diff a little easier. We can rename the AOs in a follow-up if we think it's worth it.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1235,13 +1235,13 @@
 
     <emu-clause id="sec-numeric-types">
       <h1>Numeric Types</h1>
-      <p>ECMAScript has two built-in numeric types: Number and BigInt. In this specification, every numeric type _T_ contains a multiplicative identity value denoted _T_::unit. The specification types also have the following abstract operations, likewise denoted _T_::<i>op</i> for a given operation with specification name <i>op</i>. All argument types are _T_. The "Result" column shows the return type, along with an indication if it is possible for some invocations of the operation to return an abrupt completion.</p>
+      <p>ECMAScript has two built-in numeric types: Number and BigInt. The following abstract operations are defined over these numeric types. The "Result" column shows the return type, along with an indication if it is possible for some invocations of the operation to return an abrupt completion.</p>
       <emu-table id="table-numeric-type-ops" caption="Numeric Type Operations">
         <table>
           <tbody>
           <tr>
             <th>
-              Invocation Synopsis
+              Operation
             </th>
             <th>
               Example source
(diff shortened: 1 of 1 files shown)


michaelficarra:
@bakkot Comments addressed.

jmdyck:
> I think having these automatically link is well worth the slight overhead in amount of prose.

Is that the rationale for this PR? The initial comment didn't really say.

jmdyck:
> The original motivation for the PR was to more precisely capture the return types for these operations - to talk sensibly about the return type of e.g. `T::unsignedRightShift` you have to say it can return either a normal completion of a Number or a throw completion of a TypeError, which obscures what's actually happening - namely, that `Number::unsignedRightShift` always returns a Number and never throws, and `BigInt::unsignedRightShift` always throws.

Okay, but `Number::uRS` and `BigInt::uRS` are defined separately, so (with tc39/ecmarkup#363) we can precisely convey each one's return type. Wouldn't that be enough to address the original motivation?

> This matters more with #2547, which is where the issue came up.

That one says "not ready for review", so I'm trying not to look at it.

> However, separately, I think having these automatically link is enough motivation on its own.

Yup, could be.

bakkot:
> Okay, but `Number::uRS` and `BigInt::uRS` are defined separately, so (with tc39/ecmarkup#363) we can precisely convey each one's return type.

Yes, but you have to know which you're calling for that to be useful at the call site. Otherwise you end up with exactly the same lack of precision - you don't know if your value is a completion record or not. Hence this PR.

jmdyck:
> > Okay, but `Number::uRS` and `BigInt::uRS` are defined separately, so (with [tc39/ecmarkup#363](https://github.com/tc39/ecmarkup/pull/363)) we can precisely convey each one's return type.
> 
> Yes, but you have to know which you're calling for that to be useful at the call site. Otherwise you end up with exactly the same lack of precision - you don't know if your value is a completion record or not.

Right! But if you don't know which you're calling in the status quo, how do you think this PR improves that?

E.g., if the status quo has `Type(x)::uRS(x,y)` and this PR splits that into (roughly)
 `if x is a Number then Number::uRS(x,y) else if x is a BigInt then BigInt::uRS(x,y)`
then sure, you know that the type of `Number::uRS(x,y)` is Number and the type of `BigInt::uRS(x,y)` is throw(TypeError), but the type of the whole expression is still exactly the same as for `Type(x)::uRS(x,y)`.

ljharb:
Sure, but then you don't have a mixed return type at an individual callsite.

michaelficarra:
In particular, mixed between a completion record and a non completion record.

bakkot:
@jmdyck Oh, right. The plan is for #2547 to modify `ApplyStringOrNumericBinaryOperator` to distinguish between the throwing and non-throwing AOs, [like so](https://ci.tc39.es/preview/tc39/ecma262/sha/ba9380a842cc2b8901105bfc5a9276a569e10292/multipage/ecmascript-language-expressions.html#sec-applystringornumericbinaryoperator), but I agree it doesn't affect any callsites in that way right now.

Issue #2486, opened 2021-08-12
<dfn> more terms used to describe types of values


michaelficarra:
In #2483, we identified some terms that could use a `<dfn>`.

* ["property key"](https://github.com/tc39/ecma262/pull/2483#discussion_r683111221)
* ["Iterator Record"](https://github.com/tc39/ecma262/pull/2483#discussion_r683112158)
* ["IteratorResult"](https://github.com/tc39/ecma262/pull/2483#discussion_r683112677) and [other interface names](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-common-iteration-interfaces)

We should remove italics from existing usages and use these terms wherever applicable.

Also, [in RegExp](https://tc39.es/ecma262/multipage/text-processing.html#sec-notation):
- CharSet
- State
- MatchResult
- Continuation
- Matcher

michaelficarra:
Also see #2712.

michaelficarra:
This appears to be done after #3459.

PR #2547, opened 2021-10-08
Editorial: remove implicit wrapping/unwrapping of completion records

### LINKED ISSUE #253: Notate abstract ops that can and cannot throw

### LINKED ISSUE #486: !? usage

### LINKED ISSUE #496: abstract operations don't always return Completion Records

### LINKED ISSUE #497: eliminating implicit conversions to/from Completion Records

### LINKED ISSUE #1572: ! shorthand for completion record [[Value]] extraction is unnecessary

### LINKED ISSUE #1796: Sorting out completion records

### LINKED ISSUE #2645: bug in ClassDefinitionEvaluation
Step 31.b.ii is

```
ii. Let result be ? Call(elementRecord.[[BodyFunction]], F).
```

That `?` is stray - on the next line we check for abrupt completions so we can fix up the PrivateEnvironment. The `?` prevents us from getting there, so the PrivateEnvironment ends up wrong.

Fix is to just remove the `?`. We'll do that as part of #2547.

michaelficarra:
Ready for review. Based on #2546. Fixes #1796. Fixes #497. Fixes #496. Closes #1916. Fixes #253. Fixes #1572. Closes #486. Closes #2612. Fixes #2645.

1. [x] remove the provision for implicit wrapping in normal completion
1. [x] remove the provision for treating normal completions as their [[Value]]
1. [x] annotate return types
1. [x] make sure every AO has a return type specified; figure out what to do with any remaining explicit "unknown" returns
1. [x] see if we can replace any `~empty~` returns with `~unused~`
1. [x] make sure all internal methods with the same name have matching return types
1. [x] make sure all concrete methods with the same name have matching return types
1. [x] call sites for AOs that return completion records should use `?`/`!`/`Completion()`; call sites for AOs that do not return completion records should not use `!`/`?`/`Completion()`
1. [x] make sure that each AO that doesn't return a completion record also doesn't contain a step that uses `ReturnIfAbrupt`/`?`/`throw`
1. [x] make sure that each AO that does return a completion record might plausibly return an abrupt completion, i.e. it uses `?` or `throw` or `ReturnIfAbrupt` or returns a completion explicitly
1. [x] make sure that all AOs that return `~unused~` are always invoked using "Perform". if it's a completion record that normally contains `~unused~`, make sure only its `[[Type]]` is relied upon.
1. [x] If an AO is only ever invoked using "Perform", it should return `~unused~` or a completion record that normally contains `~unused~`
1. [x] find any AOs that return completion records but are always called using `!` and make them not return completion records
1. [x] add notational convention for wrapping of returns in `NormalCompletion` in AOs that return completion records; except `Completion` or `ThrowCompletion` or `NormalCompletion`
1. [x] replace `return Completion(X)` with  `return ? X` in AOs that return a completion record
1. [x] make sure that all uses of `Completion()` are on the call of an AO that returns a completion, field accesses, or some other opaque source
1. [x] go through each AO that returns a completion record with unknown normal type and see if we can further refine it
1. [x] make sure we're not returning `Completion(...)` or `Completion Record { ... }` or `NormalCompletion(...)` or `ThrowCompletion(...)` in algorithms that don't return completion records
1. [x] check as many of the above as possible in ecmarkup and bump ecmarkup
1. [x] go through effect suppressions and make sure they are still valid and necessary
1. [x] go through issues mentioned in #2548 and make sure they are addressed here
1. [ ] add visual indicator to AO headers when the AO returns a completion record
1. [x] explain what "a Completion Record normally containing an X" means; maybe think of better phrasing?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "postbuild-spec": "git remote rm travis-origin",
     "build-head": "npm run build-only -- --lint-spec --strict",
     "prebuild-only": "npm run clean && mkdir out && cp -R img out",
-    "build-only": "ecmarkup --no-ecma-262-biblio --verbose spec.html --multipage out",
+    "build-only": "ecmarkup --verbose spec.html --multipage out",
     "build": "npm run build-head",
     "build-for-pdf": "npm run build-head -- --old-toc",
     "build-travis": "npm run build-head && npm run build-spec",
@@ -27,7 +27,7 @@
   "license": "SEE LICENSE IN https://tc39.es/ecma262/#sec-copyright-and-software-license",
   "homepage": "https://tc39.es/ecma262/",
   "dependencies": {
-    "ecmarkup": "^9.8.0"
+    "ecmarkup": "^10.0.2"
   },
   "devDependencies": {
     "glob": "^7.1.6",
--- a/spec.html
+++ b/spec.html
@@ -855,39 +855,32 @@
 
     <emu-clause id="sec-runtime-semantics">
       <h1>Runtime Semantics</h1>
-      <p>Algorithms which specify semantics that must be called at runtime are called <dfn>runtime semantics</dfn>. Runtime semantics are defined by abstract operations or syntax-directed operations. Such algorithms always return a completion record.</p>
+      <p>Algorithms which specify semantics that must be called at runtime are called <dfn>runtime semantics</dfn>. Runtime semantics are defined by abstract operations or syntax-directed operations.</p>
 
-      <emu-clause id="sec-implicit-completion-values">
-        <h1>Implicit Completion Values</h1>
-        <p>The algorithms of this specification often implicitly return Completion Records whose [[Type]] is ~normal~. Unless it is otherwise obvious from the context, an algorithm statement that returns a value that is not a Completion Record, such as:</p>
-        <emu-alg>
-          1. Return *"Infinity"*.
-        </emu-alg>
-        <p>means the same thing as:</p>
-        <emu-alg>
-          1. Return NormalCompletion(*"Infinity"*).
-        </emu-alg>
-        <p>However, if the value expression of a &ldquo;return&rdquo; statement is a Completion Record construction literal, the resulting Completion Record is returned. If the value expression is a call to an abstract operation, the &ldquo;return&rdquo; statement simply returns the Completion Record produced by the abstract operation.</p>
-        <p>The abstract operation Completion(_completionRecord_) is used to emphasize that a previously computed Completion Record is being returned. The Completion abstract operation takes a single argument, _completionRecord_, and performs the following steps:</p>
+      <emu-clause id="sec-completion-ao" type="abstract operation">
+        <h1>
+          Completion (
+            _completionRecord_: a Completion Record,
+          ): a Completion Record
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It is used to emphasize that a Completion Record is being returned.</dd>
+        </dl>
         <emu-alg>
           1. Assert: _completionRecord_ is a Completion Record.
-          1. Return _completionRecord_ as the Completion Record of this abstract operation.
+          1. Return _completionRecord_.
         </emu-alg>
-        <p>A &ldquo;return&rdquo; statement without a value in an algorithm step means the same thing as:</p>
-        <emu-alg>
-          1. Return NormalCompletion(*undefined*).
-        </emu-alg>
-        <p>Any reference to a Completion Record value that is in a context that does not explicitly require a complete Completion Record value is equivalent to an explicit reference to the [[Value]] field of the Completion Record value unless the Completion Record is an abrupt completion.</p>
       </emu-clause>
 
       <emu-clause id="sec-throw-an-exception">
         <h1>Throw an Exception</h1>
         <p>Algorithms steps that say to throw an exception, such as</p>
-        <emu-alg>
+        <emu-alg example>
           1. Throw a *TypeError* exception.
         </emu-alg>
         <p>mean the same things as:</p>
-        <emu-alg>
+        <emu-alg example>
           1. Return ThrowCompletion(a newly created *TypeError* object).
         </emu-alg>
       </emu-clause>
@@ -895,32 +888,32 @@
       <emu-clause id="sec-returnifabrupt" aoid="ReturnIfAbrupt">
         <h1>ReturnIfAbrupt</h1>
         <p>Algorithms steps that say or are otherwise equivalent to:</p>
-        <emu-alg>
+        <emu-alg example>
           1. ReturnIfAbrupt(_argument_).
         </emu-alg>
         <p>mean the same thing as:</p>
-        <emu-alg>
-          1. If _argument_ is an abrupt completion, return _argument_.
+        <emu-alg example>
+          1. If _argument_ is an abrupt completion, return Completion(_argument_).
           1. Else if _argument_ is a Completion Record, set _argument_ to _argument_.[[Value]].
         </emu-alg>
         <p>Algorithms steps that say or are otherwise equivalent to:</p>
-        <emu-alg>
+        <emu-alg example>
           1. ReturnIfAbrupt(AbstractOperation()).
         </emu-alg>
         <p>mean the same thing as:</p>
-        <emu-alg>
+        <emu-alg example>
           1. Let _hygienicTemp_ be AbstractOperation().
-          1. If _hygienicTemp_ is an abrupt completion, return _hygienicTemp_.
+          1. If _hygienicTemp_ is an abrupt completion, return Completion(_hygienicTemp_).
           1. Else if _hygienicTemp_ is a Completion Record, set _hygienicTemp_ to _hygienicTemp_.[[Value]].
         </emu-alg>
         <p>Where _hygienicTemp_ is ephemeral and visible only in the steps pertaining to ReturnIfAbrupt.</p>
         <p>Algorithms steps that say or are otherwise equivalent to:</p>
-        <emu-alg>
+        <emu-alg example>
           1. Let _result_ be AbstractOperation(ReturnIfAbrupt(_argument_)).
         </emu-alg>
         <p>mean the same thing as:</p>
-        <emu-alg>
-          1. If _argument_ is an abrupt completion, return _argument_.
+        <emu-alg example>
+          1. If _argument_ is an abrupt completion, return Completion(_argument_).
           1. If _argument_ is a Completion Record, set _argument_ to _argument_.[[Value]].
           1. Let _result_ be AbstractOperation(_argument_).
         </emu-alg>
@@ -957,6 +950,42 @@
         <p>Syntax-directed operations for runtime semantics make use of this shorthand by placing `!` or `?` before the invocation of the operation:</p>
         <emu-alg example>
           1. Perform ! SyntaxDirectedOperation of |NonTerminal|.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-implicit-normal-completion" oldids="sec-implicit-completion-values">
+        <h1>Implicit Normal Completion</h1>
+        <p>In algorithms within abstract operations which are declared to return a Completion Record, within the Evaluation syntax-directed operation, and within all built-in functions, the returned value is first passed to NormalCompletion, and the result is used instead. This rule does not apply within the Completion algorithm or when the value being returned is clearly marked as a Completion Record in that step; these cases are:</p>
+        <ul>
+          <li>when the result of applying Completion, NormalCompletion, or ThrowCompletion is directly returned</li>
+          <li>when the result of constructing a Completion Record is directly returned</li>
+          <li>when directly returning with the phrase "the result of evaluating"</li>
+        </ul>
+        <p>It is an editorial error if a Completion Record is returned from such an abstract operation through any other means. For example, within these abstract operations,</p>
+        <emu-alg example>
+          1. Return *true*.
+        </emu-alg>
+        <p>means the same things as any of</p>
+        <emu-alg example>
+          1. Return NormalCompletion(*true*).
+        </emu-alg>
+        <p>or</p>
+        <emu-alg example>
+          1. Let _completion_ be NormalCompletion(*true*).
+          1. Return Completion(_completion_).
+        </emu-alg>
+        <p>or</p>
+        <emu-alg example>
+          1. Return Completion Record { [[Type]]: ~normal~, [[Value]]: *true*, [[Target]]: ~empty~ }.
+        </emu-alg>
+        <p>Note that, through the ReturnIfAbrupt expansion, the following example is allowed, as within the expanded steps, the result of applying Completion is returned directly in the abrupt case and the implicit NormalCompletion application occurs after unwrapping in the normal case.</p>
+        <emu-alg example>
+          1. Return ? _completion_.
+        </emu-alg>
+        <p>The following example would be an editorial error because a Completion Record is being returned without being annotated in that step.</p>
+        <emu-alg example>
+          1. Let _completion_ be NormalCompletion(*true*).
+          1. Return _completion_.
         </emu-alg>
       </emu-clause>
     </emu-clause>
@@ -1056,7 +1085,7 @@
             _string_: a String,
             _searchValue_: a String,
             _fromIndex_: a non-negative integer,
-          )
+          ): an integer
         </h1>
         <dl class="header">
         </dl>
@@ -1335,7 +1364,7 @@
               BigInt::exponentiate
             </td>
             <td>
-              BigInt; may throw *RangeError*
+              either a normal completion containing a BigInt or an abrupt completion
             </td>
           </tr>
 
@@ -1381,7 +1410,7 @@
               BigInt::divide
             </td>
             <td>
-              BigInt; may throw *RangeError*
+              either a normal completion containing a BigInt or an abrupt completion
             </td>
           </tr>
 
@@ -1404,7 +1433,7 @@
               BigInt::remainder
             </td>
             <td>
-              BigInt; may throw *RangeError*
+              either a normal completion containing a BigInt or an abrupt completion
             </td>
           </tr>
 
@@ -1527,7 +1556,7 @@
               BigInt::unsignedRightShift
             </td>
             <td>
-              always throws a *TypeError*
+              a throw completion
             </td>
           </tr>
 
@@ -1746,7 +1775,7 @@
           <h1>
             Number::unaryMinus (
               _x_: a Number,
-            )
+            ): a Number
           </h1>
           <dl class="header">
           </dl>
@@ -1760,7 +1789,7 @@
           <h1>
             Number::bitwiseNOT (
               _x_: a Number,
-            )
+            ): an integral Number
           </h1>
           <dl class="header">
           </dl>
@@ -1775,7 +1804,7 @@
             Number::exponentiate (
               _base_: a Number,
               _exponent_: a Number,
-            )
+            ): a Number
           </h1>
           <dl class="header">
             <dt>description</dt>
@@ -1822,7 +1851,7 @@
             Number::multiply (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): a Number
           </h1>
           <dl class="header">
             <dt>description</dt>
@@ -1850,7 +1879,7 @@
             Number::divide (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): a Number
           </h1>
           <dl class="header">
             <dt>description</dt>
@@ -1883,7 +1912,7 @@
             Number::remainder (
               _n_: a Number,
               _d_: a Number,
-            )
+            ): a Number
           </h1>
           <dl class="header">
             <dt>description</dt>
@@ -1911,7 +1940,7 @@
             Number::add (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): a Number
           </h1>
           <dl class="header">
             <dt>description</dt>
@@ -1937,7 +1966,7 @@
             Number::subtract (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): a Number
           </h1>
           <dl class="header">
             <dt>description</dt>
@@ -1956,7 +1985,7 @@
             Number::leftShift (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): an integral Number
           </h1>
           <dl class="header">
           </dl>
@@ -1973,7 +2002,7 @@
             Number::signedRightShift (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): an integral Number
           </h1>
           <dl class="header">
           </dl>
@@ -1990,7 +2019,7 @@
             Number::unsignedRightShift (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): an integral Number
           </h1>
           <dl class="header">
           </dl>
@@ -2007,7 +2036,7 @@
             Number::lessThan (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): a Boolean or *undefined*
           </h1>
           <dl class="header">
           </dl>
@@ -2031,7 +2060,7 @@
             Number::equal (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): a Boolean
           </h1>
           <dl class="header">
           </dl>
@@ -2050,7 +2079,7 @@
             Number::sameValue (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): a Boolean
           </h1>
           <dl class="header">
           </dl>
@@ -2068,7 +2097,7 @@
             Number::sameValueZero (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): a Boolean
           </h1>
           <dl class="header">
           </dl>
@@ -2087,7 +2116,7 @@
               _op_: `&amp;`, `^`, or `|`,
               _x_: a Number,
               _y_: a Number,
-            )
+            ): an integral Number
           </h1>
           <dl class="header">
           </dl>
@@ -2108,7 +2137,7 @@
             Number::bitwiseAND (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): an integral Number
           </h1>
           <dl class="header">
           </dl>
@@ -2122,7 +2151,7 @@
             Number::bitwiseXOR (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): an integral Number
           </h1>
           <dl class="header">
           </dl>
@@ -2136,7 +2165,7 @@
             Number::bitwiseOR (
               _x_: a Number,
               _y_: a Number,
-            )
+            ): an integral Number
           </h1>
           <dl class="header">
           </dl>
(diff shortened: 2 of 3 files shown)


arhadthedev:
The PR introduces a term <code>a Completion Record normally containing [a *Type*]</code>. I think it would be useful to explicitly define it in <https://tc39.es/ecma262/#sec-completion-record-specification-type> via a normal completion (probably with no `<dfn>` to avoid visual clutter). Otherwise it may seem that "normally" means "sometimes it can be violated and another type may appear".

spec.html
```diff
@@ -845,29 +845,22 @@ <h1>Syntax-Directed Operations</h1>
...
-        <p>The abstract operation Completion(_completionRecord_) is used to emphasize that a previously computed Completion Record is being returned. The Completion abstract operation takes a single argument, _completionRecord_, and performs the following steps:</p>
+      <emu-clause id="sec-completion-ao" type="abstract operation">
+        <h1>
+          Completion (
+            _completionRecord_: a Completion Record,
+          ): a Completion Record
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It is used to emphasize that a Completion Record is being returned.</dd>
```
arhadthedev, line 858: In the light of explicit emphasis via a return type, does `Completion()` give any benefit or it may be removed altogether?

bakkot: You don't have the return type sitting in front of you at the call site, only at the definition site, so yes, it's still helpful.

spec.html
```diff
@@ -6285,15 +6285,16 @@ <h1>
           _O_: an Object,
           _P_: a property key,
           _V_: an ECMAScript language value,
-        ): a Completion Record normally containing a Boolean
+        ): a Completion Record normally containing ~unused~
```
ljharb, line 6288: "normally containing" implies to me that sometimes it will contain something else. is this trying to convey that, or that it's a Completion of [[Type]] normal, containing `~unused~`?

bakkot: @ljharb the second thing; see [comment above](https://github.com/tc39/ecma262/pull/2547#issuecomment-940356606).

ljharb: thanks; the comment sums it up - altho i think i'd rather find another term than just settle for defining it.

spec.html
```diff
@@ -867,7 +867,7 @@ <h1>
           <dt>description</dt>
           <dd>It is used to emphasize that a Completion Record is being returned.</dd>
         </dl>
-        <emu-alg>
+        <emu-alg example>
```
ljharb, line 870: how is this an example?

michaelficarra: oops, thanks for catching that

jmdyck:
(You should probably also update the PR description where it says "Not ready for review".)

jmdyck:
> replace `return Completion(X)` with `return ? X` in AOs that return a completion record

Why?

michaelficarra:
@jmdyck No reason, we just thought it was stylistically neater and semantically it's equivalent.

bakkot:
In particular, we already have lots of `Return ? Foo()`, which is equivalent to `Return Completion(Foo())`; we might as well pick one of those two forms and stick to it. And the former is less noisy (and, to my mind, clearer).

jmdyck:
> In particular, we already have lots of `Return ? Foo()`, which is equivalent to `Return Completion(Foo())`; we might as well pick one of those two forms and stick to it.

Okay. Given those choices, I'd vote for `Return Completion(Foo())`.

> And the former is less noisy (and, to my mind, clearer).

I might agree that `Return ? X` is less noisy than `Return Completion(X)`, but
- not much, and
- if noise is a concern, you could invent a less noisy notation for `Completion(X)`.

As for "clearer", I strongly disagree. To me,
- `Return Completion(X)` says: `X` is a CR, which is returned.
- `Return ? X` says: `X` is a CR. If it's an abrupt CR, then it's [implicitly] returned (via the semantics of `?`).  Otherwise (i.e., if it's a normal CR), the CR's `[[Value]]` is extracted and becomes the argument of the `Return`, at which point that value is implicitly passed to `NormalCompletion()`, and the resulting normal CR is [explicitly] returned.

Also, in a lot of cases of `Return ? _cr_`, the `_cr_` is never a normal CR, so the abrupt 'path' is always taken and the `Return` is never actually executed, which feels odd. But there's no corresponding problem for `Return Completion(_cr_)`.

bakkot:
It's true that if you think about what the spec looks like after applying the rewrite rules, that thing ends up being a fairly awkward sequence of steps.

But most readers are not thinking about what the spec looks like after applying rewrite rules, and they already encounter `? AO()` all the time. `Return ? AO()` looks like `Let _a_ be ? AO()` and requires no further thought. Using a different construct than the usual one here obscures that fundamental similarity.

I stand by "clearer" here.

jmdyck:
> It's true that if you think about what the spec looks like after applying the rewrite rules, that thing ends up being a fairly awkward sequence of steps. But most readers are not thinking about what the spec looks like after applying rewrite rules,

I'm not thinking about what the spec looks like after applying rewrite rules, I'm thinking about the semantics of `?`.

> and they already encounter `? AO()` all the time.

I don't understand: are you saying they're used to seeing it, so they don't think about what the `?` means? Or they do, but in a simplified way?

> `Return ? AO()` looks like `Let _a_ be ? AO()` and requires no further thought. Using a different construct than the usual one here obscures that fundamental similarity.

By the same logic, `Return Completion(X)` looks like `Let _a_ be Completion(X)` and requires no further thought; using a different construct would obscure that fundamental similarity.

(Personally, I don't think the putative obscuring happens in either direction.)

spec.html
```diff
@@ -18005,28 +18051,27 @@ <h2>Syntax</h2>
...
           1. Let _len_ be _nextIndex_ + 1.
           1. Perform ? Set(_array_, *"length"*, 𝔽(_len_), *true*).
-          1. NOTE: The above Set throws if _len_ exceeds 2<sup>32</sup>-1.
+          1. NOTE: The above step throws if _len_ exceeds 2<sup>32</sup>-1.
           1. Return _len_.
         </emu-alg>
         <emu-grammar>Elision : Elision `,`</emu-grammar>
         <emu-alg>
-          1. Return the result of performing ArrayAccumulation of |Elision| with arguments _array_ and _nextIndex_ + 1.
+          1. Return ? ArrayAccumulation of |Elision| with arguments _array_ and _nextIndex_ + 1.
```
```suggestion
          1. Return ? ArrayAccumulation of |Elision| with arguments _array_ and (_nextIndex_ + 1).
```
Otherwise, there's an ambiguity: an alternative parse would have `? ArrayAccumulation of |Elision| with arguments _array_ and _nextIndex_` as the left-hand operand of the `+`.

spec.html
```diff
@@ -19158,7 +19191,7 @@ <h1>Runtime Semantics: ArgumentListEvaluation</h1>
           1. Let _firstSub_ be ? GetValue(_firstSubRef_).
           1. Let _restSub_ be ? SubstitutionEvaluation of |TemplateSpans|.
           1. Assert: _restSub_ is a possibly empty List.
-          1. Return the list-concatenation of &laquo; _firstSub_ &raquo; and _restSub_.
+          1. Return NormalComplation(the list-concatenation of &laquo; _firstSub_ &raquo; and _restSub_).
```
```suggestion
          1. Return NormalCompletion(the list-concatenation of &laquo; _firstSub_ &raquo; and _restSub_).
```

spec.html
```diff
@@ -21095,29 +21129,31 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+          1. Return ~empty~.
         </emu-alg>
         <emu-note>
           <p>A static semantics rule ensures that this form of |LexicalBinding| never occurs in a `const` declaration.</p>
         </emu-note>
         <emu-grammar>LexicalBinding : BindingIdentifier Initializer</emu-grammar>
         <emu-alg>
           1. Let _bindingId_ be StringValue of |BindingIdentifier|.
-          1. Let _lhs_ be ResolveBinding(_bindingId_).
+          1. <!-- TODO: Figure out if ResolveBinding can actually throw here. --> Let _lhs_ be Completion(ResolveBinding(_bindingId_)).
```
jmdyck: Delete or resolve TODO comment?

spec.html
```diff
@@ -29431,7 +29475,7 @@ <h1>
             1. For each element _nextKey_ of _keys_, do
               1. If Type(_nextKey_) is Symbol and _type_ is ~symbol~ or Type(_nextKey_) is String and _type_ is ~string~, then
                 1. Append _nextKey_ as the last element of _nameList_.
-            1. Return CreateArrayFromList(_nameList_).
+            1. Return _nameList.
```
```suggestion
            1. Return _nameList_.
```

spec.html
```diff
@@ -41524,9 +41559,9 @@ <h1>
...
-          1. Perform LeaveCriticalSection(_WL_) and suspend _W_ for up to _timeout_ milliseconds, performing the combined operation in such a way that a notification that arrives after the critical section is exited but before the suspension takes effect is not lost. _W_ can notify either because the timeout expired or because it was notified explicitly by another agent calling NotifyWaiter(_WL_, _W_), and not for any other reasons at all.
+          1. Perform LeaveCriticalSection(_WL_) and suspend _W_ for up to _timeout_ milliseconds, performing the combined operation in such a way that a notification that arrives after the critical section is exited but before the suspension takes effect is not lost. _W_ can notify either because the timeout expired or because it was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, and not for any other reasons at all.
           1. Perform EnterCriticalSection(_WL_).
-          1. If _W_ was notified explicitly by another agent calling NotifyWaiter(_WL_, _W_), return *true*.
+          1. If _W_ was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, return *true*.
```
jmdyck: Not sure why these were changed to SDO-style "with arguments". NotifyWaiter is invoked with a parenthesized list of args, so why wouldn't these references to hypothetical invocations be formatted that way too?

jmdyck (changes requested):

bakkot:
> I don't understand: are you saying they're used to seeing it, so they don't think about what the `?` means? Or they do, but in a simplified way?

It's not so much "simplified" as "abstracted": `X()` might be abrupt, in which case `Return ? X()` returns abruptly, and it might be normal, in which case it returns normally. If you think about this in terms of the concrete Record values which represent those two branches then you might find this strange, since it is after all going to have the same behavior in both branches, yes. But the whole point of `?` is that it abstracts away the concrete values in most cases, so that you don't actually have to think about them to follow the flow of the algorithm. 

> By the same logic, `Return Completion(X)` looks like `Let _a_ be Completion(X)` and requires no further thought; using a different construct would obscure that fundamental similarity.

Yes, and that's the point. `Let _a_ be Completion(X)` is an unusual thing to encounter and indicates something strange is going on: it means we're manipulating Completion values explicitly, usually because we're doing something with Promises or iterators which requires additional steps rather than just propagating abrupt completions. So you _should_ find `Let _a_ be Completion(X)` to be unusual; typical algorithm steps should _not_ be similar to it, even though we are in fact almost always manipulating Completion values. The point of abstractions like `?` is to background details like that.

By contrast, `Return ? X()` and `Let _a_ be ? X()` are the same kind of thing as each other: not working with a Completion directly, just propagating abrupt completions the same way we usually do.

package.json
```diff
@@ -27,7 +27,7 @@
   "license": "SEE LICENSE IN https://tc39.es/ecma262/#sec-copyright-and-software-license",
   "homepage": "https://tc39.es/ecma262/",
   "dependencies": {
-    "ecmarkup": "^9.8.0"
+    "ecmarkup": "^9.8.1"
```
ljharb, line 30: since this is a patch version, the changes in this PR shouldn't be needed to land the ecmarkup upgrade, right?

bakkot: The patch fixes a bug which this PR would exercise, so yes, it is necessary.

spec.html
```diff
@@ -1335,7 +1364,7 @@ <h1>Numeric Types</h1>
               BigInt::exponentiate
             </td>
             <td>
-              BigInt; may throw *RangeError*
+              a Completion Record normally containing a BigInt
```
```suggestion
              a Completion Record (Abrupt, or Normal containing a BigInt)
```
what about this? the form also works for Return/Continue/Break

package.json
michaelficarra: The question was whether we could do the ecmarkup upgrade first. This patch bump we could do, but the upcoming ecmarkup bump will be major, and should be included in this PR. So no point in doing this patch bump first.

jmdyck:
When I try to "Load diff" for #2547, I get "Oops, something went wrong."
So instead, I've submitted my suggested changes (old ones and new ones) in [a PR against GH-1796 branch](https://github.com/tc39/ecma262/pull/2653).

jmdyck:
Oh, also, I noticed that you have 4 occurrences like:
```a Record with [[A]] and [[B]] fields```
and 3 occurrences like:
```a Record with fields [[A]] (a Foo) and [[B]] (a Bar)```

The latter is more informative, so presumably preferable. Did you want to convert the former?

File: spec.html (hunk 8 above)
Comment (michaelficarra) [reply spec.html]: Two reasons:

1. We didn't want these references to be treated as calls by ecmarkup because (if they were) they're not used appropriately for the types.
2. There's actually ambiguity in the previous phrasing. Are we referring to the result of the call? This makes it clear that we are referring to `NotifyWaiter` itself.

File: spec.html (hunk 6 above)
Comment (michaelficarra) [reply spec.html:21142]: Deleted, but I think we should still try to figure this out.

File: spec.html (hunk 10 above)
Comment (bakkot) [reply spec.html:1367]: If everyone else prefers this form I'll defer to that. Let's get a vote from @syg?

Comment (michaelficarra) [reply spec.html:1367]: I think we should just wait until we put all options on the table at the editor call.

Comment (jmdyck) [reply spec.html:1367]: (It's a bit wordy, but on the other hand, you're clearly not going for conciseness, else you could say something like `Throw | Normal(BigInt)`.)

File: spec.html (hunk 8 above)
Comment (jmdyck) [reply spec.html]: Okay, make sense.

**Comment by jmdyck** [comment]
I've submitted a PR for a third batch of changes: #2657

Also, I noticed some other things but wasn't sure how to resolve them:

HasBinding:
Its return is "a Completion Record normally containing a Boolean", and yet almost all invocations of HasBinding() treat it like its return is just "a Boolean". (i.e., no `!`, `?`, or `Completion()`)

AsyncGeneratorUnwrapYieldResumption:
can't it return a normal completion?

There's an ambiguity:
`a List of Strings or *null*`
could mean:
`a List of (Strings or *null*)`
or:
`(a List of Strings) or *null*`

**Comment by bakkot** [comment]
> HasBinding

Oh, good catch. We went through all the concrete methods to make sure their declared types were consistent but we forgot to check the uses afterwards, I think.

I believe the sites which currently are missing `!` or `?` or `Completion` should all be `!`; that method only throws for object and global environment records, which I think are all screened out for the use sites in question. I'll go over it with @michaelficarra to check and update though.

> AsyncGeneratorUnwrapYieldResumption: can't it return a normal completion?

Oh, from the first step? Yup, I think you're right.

> There's an ambiguity:

Yeah, English has this problem. Probably best to add "either" to disambiguate, I guess.

**Comment by jmdyck** [comment]
(Diff still isn't working, so I'll just put this here.)
For `GetExportedNames`, in the return type,
> `either a List of Strings or *null*`

should be:
> `a List of either Strings or *null*`

**Comment by bakkot** [comment]
Ha, great catch, fixed. I think we read the ambiguity wrong while disambiguating, which is pretty good evidence for the need for `either`.

**Review by ljharb — COMMENTED** [review]
there's changes pending after this; i'll review those ne

File: spec.html [hunk 11]
```diff
@@ -7071,14 +7107,14 @@ <h1>
         CreateIterResultObject (
           _value_: unknown,
           _done_: a Boolean,
-        )
+        ): an Object that conforms to the <i>IteratorResult</i> interface
```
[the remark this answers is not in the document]
michaelficarra: no https://github.com/tc39/ecma262/issues/2486

spec.html
```diff
@@ -19008,9 +19051,7 @@ <h1>
           1. If Type(_func_) is not Object, throw a *TypeError* exception.
           1. If IsCallable(_func_) is *false*, throw a *TypeError* exception.
           1. If _tailPosition_ is *true*, perform PrepareForTailCall().
-          1. Let _result_ be Call(_func_, _thisValue_, _argList_).
-          1. Assert: If _result_ is not an abrupt completion, then Type(_result_) is an ECMAScript language type.
```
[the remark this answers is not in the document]
jmdyck: > if completions can only contain language values

This PR has lots of examples of normal completions containing non-language values (e.g., a List, a Record). However, in the "Completion Record Fields" table, `[[Value]]` still says "an ECMAScript language value or ~empty~", so that should probably be fixed.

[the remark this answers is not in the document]
michaelficarra: Completion Records can contain things that are not ECMAScript language values. `~unused~` is a very common one. This assertion is removed because it's redundant with the type signature.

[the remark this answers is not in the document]
ljharb: ah, because `Call` can only return a normal completion when it contains an ES language value, got it. thanks!

spec.html
```diff
@@ -855,72 +855,65 @@ <h1>Syntax-Directed Operations</h1>
...
+          Completion (
+            _completionRecord_: a Completion Record,
+          ): a Completion Record
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It is used to emphasize that a Completion Record is being returned.</dd>
+        </dl>
         <emu-alg>
           1. Assert: _completionRecord_ is a Completion Record.
```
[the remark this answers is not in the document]
ljharb: i guess so, but i thought the main benefit of the structured param types was to remove the explicit assertion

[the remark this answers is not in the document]
bakkot: In general, yes, because in general the assertion is incidental to the operation of the AO. Here the assertion is the whole point of the AO.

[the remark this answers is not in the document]
ljharb: Fair enough!

syg:
My review since I can't get GitHub to load the diff at all.

Without listing every location, the "either a normal completion containing either an X or a Y or an abrupt completion" doesn't read great. IMO this is because of the nested "either", so I recommend for those to be changed to:

2-form: either a normal completion containing an X or Y, or an abrupt completion

2+-form: either a normal completion containing an X, Y, or Z, or an abrupt completion

<hr>

L964: `<p>The Evaluation syntax-directed operation and all built-in functions return a Completion Record. It is an editorial error if a Completion Record is returned from such an abstract operation through any other means. For example, within these abstract operations,</p>`

I find the "The Evaluation [...]" sentence out of place and also makes "such an abstract operation" confusing in the following sentence. While it is true that Evaluation and built-ins always return completions, this section is about the _implicit_ normal completions. Let's move it up to L958 as:

In algorithms within abstract operations which are declared to return a Completion Record, **within the Evaluation syntax-directed operation, and within all built-in functions**, the returned value is first passed to NormalCompletion, and the result is used instead. This rule does not apply within the Completion algorithm or when the value being returned is clearly marked as a Completion Record in that **same** step; these cases are:

<hr>

L985: Please add a short explanation for why the example is an editorial error.

<hr>

L9493: Missing `?`, should be `1. Return ? _result_.`
L21688: Missing `?`, should be `1. Return ? _bodyResult_.`

<hr>

L38842: Nice, suppressing UC here.

<hr>

L41573: Why was the `with arguments _WL_ and _W_` change made?
L41575: Ditto

michaelficarra:
@syg

> L41573: Why was the `with arguments _WL_ and _W_` change made?

See https://github.com/tc39/ecma262/pull/2547#discussion_r799945817

syg:
> See https://github.com/tc39/ecma262/pull/2547#discussion_r799945817

Ah okay, makes sense.

bakkot:
@ljharb @syg I believe we've addressed all your comments. The first commit is just the rebase and the subsequent ones are changes we made after your reviews; PTAL.

spec.html
michaelficarra: @ljharb This is another little thing that can be pulled into a separate PR if you like.

bakkot: It would need to land first, though.

ljharb: #2667

syg (approved):
Wording lgtm, trusting correctness to existing checks.

spec.html
```diff
@@ -24765,15 +24809,15 @@ <h1>
...
           1. Let _F_ be _constructorInfo_.[[Closure]].
-          1. Perform ! MakeClassConstructor(_F_).
-          1. Perform ! SetFunctionName(_F_, _className_).
-        1. Perform ! MakeConstructor(_F_, *false*, _proto_).
+          1. Perform MakeClassConstructor(_F_).
+          1. Perform SetFunctionName(_F_, _className_).
+        1. Perform MakeConstructor(_F_, *false*, _proto_).
         1. If |ClassHeritage?| is present, set _F_.[[ConstructorKind]] to ~derived~.
-        1. Perform ! CreateMethodProperty(_proto_, *"constructor"*, _F_).
+        1. Perform CreateMethodPropertyInfallibly(_proto_, *"constructor"*, _F_).
```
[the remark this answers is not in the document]
michaelficarra: Already done.

spec.html
```diff
@@ -29461,23 +29512,23 @@ <h1>Object.getOwnPropertyDescriptors ( _O_ )</h1>
         <h1>Object.getOwnPropertyNames ( _O_ )</h1>
         <p>When the `getOwnPropertyNames` function is called, the following steps are taken:</p>
         <emu-alg>
-          1. Return ? GetOwnPropertyKeys(_O_, ~string~).
+          1. Return CreateArrayFromList(? GetOwnPropertyKeys(_O_, ~string~)).
```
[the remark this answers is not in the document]
michaelficarra: Yeah just a small refactor. See my comment over there.

michaelficarra:
@ljharb Addressed. Thanks for the review.

PR #2548, opened 2021-10-08
Editorial: Can-call-user-code annotation

syg:
This PR adds an annotation to call sites of AOs and SDOs that can call user code, possibly transitively.

Ecmarkup starts with a set of "roots" that can call user code, and propagates the effect to all reachable AO and SDO invocations.

For spec authors:
- Roots and false negatives can be manually tagged with `<emu-meta effects="user-code">root cause</emu-meta>`.
- Most invocations of internal methods (e.g. [[Get]]) are roots.
- The phrase "the result of evaluating" is automatically considered a root.
- False positives can be manually suppressed with `<emu-meta suppress-effects="user-code">suppressed</emu-meta>`.
- Steps can be manually fenced with `1. [fence-effects="user-code"] Step`, which fences any child step effects from propagating.
- Some spec constructs in algorithm steps are like nested functions and should fence effects from propagating. Abstract Closures are automatic fences.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4047,7 +4047,7 @@
             1. Let _prevContext_ be the running execution context.
             1. Suspend _prevContext_.
             1. Push _asyncContext_ onto the execution context stack; _asyncContext_ is now the running execution context.
-            1. Resume the suspended evaluation of _asyncContext_ using NormalCompletion(_value_) as the result of the operation that suspended it.
+            1. <emu-meta effects="user-code">Resume the suspended evaluation of _asyncContext_</emu-meta> using NormalCompletion(_value_) as the result of the operation that suspended it.
             1. Assert: When we reach this step, _asyncContext_ has already been removed from the execution context stack and _prevContext_ is the currently running execution context.
             1. Return *undefined*.
           1. Let _onFulfilled_ be ! CreateBuiltinFunction(_fulfilledClosure_, 1, *""*, &laquo; &raquo;).
@@ -4055,7 +4055,7 @@
             1. Let _prevContext_ be the running execution context.
             1. Suspend _prevContext_.
             1. Push _asyncContext_ onto the execution context stack; _asyncContext_ is now the running execution context.
-            1. Resume the suspended evaluation of _asyncContext_ using ThrowCompletion(_reason_) as the result of the operation that suspended it.
+            1. <emu-meta effects="user-code">Resume the suspended evaluation of _asyncContext_</emu-meta> using ThrowCompletion(_reason_) as the result of the operation that suspended it.
             1. Assert: When we reach this step, _asyncContext_ has already been removed from the execution context stack and _prevContext_ is the currently running execution context.
             1. Return *undefined*.
           1. Let _onRejected_ be ! CreateBuiltinFunction(_rejectedClosure_, 1, *""*, &laquo; &raquo;).
@@ -4246,11 +4246,11 @@
             1. [id="step-getvalue-toobject"] Let _baseObj_ be ? ToObject(_V_.[[Base]]).
             1. If IsPrivateReference(_V_) is *true*, then
               1. Return ? PrivateGet(_baseObj_, _V_.[[ReferencedName]]).
-            1. Return ? _baseObj_.[[Get]](_V_.[[ReferencedName]], GetThisValue(_V_)).
+            1. Return ? <emu-meta effects="user-code">_baseObj_.[[Get]]</emu-meta>(_V_.[[ReferencedName]], GetThisValue(_V_)).
           1. Else,
             1. Let _base_ be _V_.[[Base]].
             1. Assert: _base_ is an Environment Record.
-            1. Return ? _base_.GetBindingValue(_V_.[[ReferencedName]], _V_.[[Strict]]) (see <emu-xref href="#sec-environment-records"></emu-xref>).
+            1. Return ? <emu-meta effects="user-code">_base_.GetBindingValue</emu-meta>(_V_.[[ReferencedName]], _V_.[[Strict]]) (see <emu-xref href="#sec-environment-records"></emu-xref>).
         </emu-alg>
         <emu-note>
           <p>The object that may be created in step <emu-xref href="#step-getvalue-toobject"></emu-xref> is not accessible outside of the above abstract operation and the ordinary object [[Get]] internal method. An implementation might choose to avoid the actual creation of the object.</p>
@@ -4278,13 +4278,13 @@
             1. [id="step-putvalue-toobject"] Let _baseObj_ be ? ToObject(_V_.[[Base]]).
             1. If IsPrivateReference(_V_) is *true*, then
               1. Return ? PrivateSet(_baseObj_, _V_.[[ReferencedName]], _W_).
-            1. Let _succeeded_ be ? _baseObj_.[[Set]](_V_.[[ReferencedName]], _W_, GetThisValue(_V_)).
+            1. Let _succeeded_ be ? <emu-meta effects="user-code">_baseObj_.[[Set]]</emu-meta>(_V_.[[ReferencedName]], _W_, GetThisValue(_V_)).
             1. If _succeeded_ is *false* and _V_.[[Strict]] is *true*, throw a *TypeError* exception.
             1. Return.
           1. Else,
             1. Let _base_ be _V_.[[Base]].
             1. Assert: _base_ is an Environment Record.
-            1. Return ? _base_.SetMutableBinding(_V_.[[ReferencedName]], _W_, _V_.[[Strict]]) (see <emu-xref href="#sec-environment-records"></emu-xref>).
+            1. Return ? <emu-meta effects="user-code">_base_.SetMutableBinding</emu-meta>(_V_.[[ReferencedName]], _W_, _V_.[[Strict]]) (see <emu-xref href="#sec-environment-records"></emu-xref>).
         </emu-alg>
         <emu-note>
           <p>The object that may be created in step <emu-xref href="#step-putvalue-toobject"></emu-xref> is not accessible outside of the above abstract operation and the ordinary object [[Set]] internal method. An implementation might choose to avoid the actual creation of that object.</p>
@@ -5968,7 +5968,7 @@
         <dd>It returns a completion record which, if its [[Type]] is ~normal~, has a [[Value]] which is a Boolean. It is used to determine whether additional properties can be added to _O_.</dd>
       </dl>
       <emu-alg>
-        1. Return ? _O_.[[IsExtensible]]().
+        1. Return ? <emu-meta effects="user-code">_O_.[[IsExtensible]]</emu-meta>().
       </emu-alg>
     </emu-clause>
 
@@ -6297,7 +6297,7 @@
         <dd>It is used to retrieve the value of a specific property of an object.</dd>
       </dl>
       <emu-alg>
-        1. Return ? _O_.[[Get]](_P_, _O_).
+        1. Return ? <emu-meta effects="user-code">_O_.[[Get]]</emu-meta>(_P_, _O_).
       </emu-alg>
     </emu-clause>
 
@@ -6314,7 +6314,7 @@
       </dl>
       <emu-alg>
         1. Let _O_ be ? ToObject(_V_).
-        1. Return ? _O_.[[Get]](_P_, _V_).
+        1. Return ? <emu-meta effects="user-code">_O_.[[Get]]</emu-meta>(_P_, _V_).
       </emu-alg>
     </emu-clause>
 
@@ -6332,7 +6332,7 @@
         <dd>It is used to set the value of a specific property of an object. _V_ is the new value for the property.</dd>
       </dl>
       <emu-alg>
-        1. Let _success_ be ? _O_.[[Set]](_P_, _V_, _O_).
+        1. Let _success_ be ? <emu-meta effects="user-code">_O_.[[Set]]</emu-meta>(_P_, _V_, _O_).
         1. If _success_ is *false* and _Throw_ is *true*, throw a *TypeError* exception.
         1. Return _success_.
       </emu-alg>
@@ -6352,7 +6352,7 @@
       </dl>
       <emu-alg>
         1. Let _newDesc_ be the PropertyDescriptor { [[Value]]: _V_, [[Writable]]: *true*, [[Enumerable]]: *true*, [[Configurable]]: *true* }.
-        1. Return ? _O_.[[DefineOwnProperty]](_P_, _newDesc_).
+        1. Return ? <emu-meta effects="user-code">_O_.[[DefineOwnProperty]]</emu-meta>(_P_, _newDesc_).
       </emu-alg>
       <emu-note>
         <p>This abstract operation creates a property whose attributes are set to the same defaults used for properties created by the ECMAScript language assignment operator. Normally, the property will not already exist. If it does exist and is not configurable or if _O_ is not extensible, [[DefineOwnProperty]] will return *false*.</p>
@@ -6373,7 +6373,7 @@
       </dl>
       <emu-alg>
         1. Let _newDesc_ be the PropertyDescriptor { [[Value]]: _V_, [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.
-        1. Return ? _O_.[[DefineOwnProperty]](_P_, _newDesc_).
+        1. Return ? <emu-meta effects="user-code">_O_.[[DefineOwnProperty]]</emu-meta>(_P_, _newDesc_).
       </emu-alg>
       <emu-note>
         <p>This abstract operation creates a property whose attributes are set to the same defaults used for built-in methods and methods defined using class declaration syntax. Normally, the property will not already exist. If it does exist and is not configurable or if _O_ is not extensible, [[DefineOwnProperty]] will return *false*.</p>
@@ -6436,7 +6436,7 @@
         <dd>It is used to call the [[DefineOwnProperty]] internal method of an object in a manner that will throw a *TypeError* exception if the requested property update cannot be performed.</dd>
       </dl>
       <emu-alg>
-        1. Let _success_ be ? _O_.[[DefineOwnProperty]](_P_, _desc_).
+        1. Let _success_ be ? <emu-meta effects="user-code">_O_.[[DefineOwnProperty]]</emu-meta>(_P_, _desc_).
         1. If _success_ is *false*, throw a *TypeError* exception.
         1. Return _success_.
       </emu-alg>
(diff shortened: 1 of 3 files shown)


ljharb:
A false positive is https://ci.tc39.es/preview/tc39/ecma262/sha/a5ba1ae3a48e8cffb1e6051bc97dceace591af20/#sec-createregexpstringiterator.

iow, technically if you called that AO with a user-provided regex, the annotations are correct; but it's only ever called with a spec-fiction-regex, so they're not. there's probably a similar case with async iteration.

bakkot:
As a concrete example of a place the analysis fails for abstract closures, `%TypedArray%.prototype.entries` has as its third step `Return CreateArrayIterator(_O_, ~key+value~).`, which does not _directly_ invoke user code; rather, it returns an abstract closure which does so.

---

Another thing is `Yield`, introduced in https://github.com/tc39/ecma262/pull/2045, which we use both for a literal `yield` expression and to specify built-in iterators like `CreateArrayIterator`. `Yield` just defers to `GeneratorYield` or `AsyncGeneratorYield` depending on which kind of generator we're in. It's is only marked as can-call-user-code because in the async case it calls `Await`, which calls `PromiseResolve`, which `[[Get]]`s the user-controlled `.constructor` of the passed value. In the sync case, it does not call user code (although it does pause execution, such that user code can be called in the mean time).

Other than the `yield` in ES code, I'm pretty sure we always know at the callsite whether we're in a sync or an async generator. (Even for the user code one, we can figure it out by calling `GetGeneratorKind()`.) In fact, because we don't currently have any async iterators in the spec, for the built-in iterators it's always sync. I think we should probably just call `GeneratorYield` directly, but failing that, we should mark all the `Yield`s in built-in iterators as not calling user code.

syg:
> Another thing is Yield, introduced in #2045, which we use both for a literal yield expression and to specify built-in iterators like CreateArrayIterator. Yield just defers to GeneratorYield or AsyncGeneratorYield depending on which kind of generator we're in. It's is only marked as can-call-user-code because in the async case it calls Await, which calls PromiseResolve, which [[Get]]s the user-controlled .constructor of the passed value. In the sync case, it does not call user code (although it does pause execution, such that user code can be called in the mean time).

> Other than the yield in ES code, I'm pretty sure we always know at the callsite whether we're in a sync or an async generator. (Even for the user code one, we can figure it out by calling GetGeneratorKind().) In fact, because we don't currently have any async iterators in the spec, for the built-in iterators it's always sync. I think we should probably just call GeneratorYield directly, but failing that, we should mark all the Yields in built-in iterators as not calling user code (since they're all sync).

So by my reading GeneratorYield() can also never return abrupt completions, yeah?

bakkot:
> So by my reading GeneratorYield() can also never return abrupt completions, yeah?

From the perspective of the consumer of `GeneratorYield`, that's not correct: the completion record which `GeneratorYield` returns can be a throw or return completion. This happens when calling [`.throw`](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-generator.prototype.throw) or [`.return`](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-generator.prototype.return), which will explicitly construct and pass in such completions using the "resume ... using" language in [`GeneratorResumeAbrupt`](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-generatorresumeabrupt) step 10, which is basically a jump to [`GeneratorYield`](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-generatoryield) step 7.a.

("Return" is kind of a funny word here, because of the execution context stack manipulation - the "Return" in `GeneratorYield` step 8 isn't an actual return. I have a PR fixing this in https://github.com/tc39/ecma262/pull/2429, but it needs rebasing. Still, you might find this easier to follow by reading [the version of `GeneratorYield` in that PR](https://ci.tc39.es/preview/tc39/ecma262/sha/177b3eb7d4103744b44e4b9283a2506d034c0ba8/multipage/control-abstraction-objects.html#sec-generatoryield).)

syg:
Ah, of course, when it "returns" upon abnormal resumption.

syg:
The Abstract Closure stuff won't be correct in the preview until https://github.com/tc39/ecmarkup/pull/382 is merged and bumped here.

bakkot:
> The Abstract Closure stuff won't be correct in the preview until tc39/ecmarkup#382 is merged and bumped here.

Done.

bakkot:
Below lists are not comprehensive.

False negatives:
- all calls to the Evaluate and ExecuteModule concrete methods for cyclic module records
- every use of the phrase "the result of evaluating", except for the one in JSON.parse - these are the _actual_ roots, even though they're mostly intermediated through internal methods (this phrase probably deserves explicit support in ecmarkup)
- most callsites of the HasBinding, CreateMutableBinding, InitializeBinding, SetMutableBinding, GetBindingValue, and DeleteBinding concrete methods of environment records (which are user-code-y because of `with(proxy)`) - these don't propagate for the same reason the `[[`-style internal methods don't propagate
- calls to IsExtensible - I don't know what's up with that; the IsExtensible method itself is correctly marked as containing user code in its body
- calls to [[GetPrototypeOf]], [[PreventExtensions]], and [[OwnPropertyKeys]] - I'm guessing your regex for internal methods just missed these 

False positives:
- the eight instances of `Perform DefinePropertyOrThrow` without a `!` or `?` (which should just get a `!`; presumably that will be fixed by #2547)
- the first step in RegExpCreate (which should be `!`; it can't throw here because it's using the built-in RegExp constructor)
- the call to RegExpCreate in [the evaluation semantics for regexp literals](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-regular-expression-literals-runtime-semantics-evaluation) (RegExpCreate only calls user code if one of the arguments is an object with a trapped ToString)
- AllocateTypedArrayBuffer step 5 (it's `?` because CreateByteDataBlock can throw, but it doesn't call user code; it uses the built-in ArrayBuffer constructor) 
- SortCompare steps 7 and 9 - IsLessThan can only call user code when called with objects, not strings

Other issues:
- the call to ToString in CreateMappedArgumentsObject step 19.b.ii.3 is marked as UC, but it's supposed to be associated with the surrounding `_map_.[[DefineProperty]]` call (which is explicitly marked as user code). This is presumably a bug in ecmarkup.
- the call to GetThisValue in GetValue step 4.c has the same issue and should have the same fix (probably there's others)

syg:
> every use of the phrase "the result of evaluating", except for the one in JSON.parse - these are the actual roots, even though they're mostly intermediated through internal methods (this phrase probably deserves explicit support in ecmarkup)

My preference is to add the `<dt>effects</dt><dd>user-code</dd>` to the AOs that contain that phrasing than to special case it for now, because all the roots now are manually marked like that. There's no special handling of an AO that has an `<emu-meta effects="user-code">...</emu-meta>` somewhere, and I'm kind of worried about having a pass that looks through the contents of every AO, until it's no longer maintainable to manually mark all the roots.

bakkot:
> You're not talking about just marking a surrounding AO that has it as a root, but actually rendering the UC for the phrase "the result of evaluating |Foo|".

Right.

> But still it does the extra pass that I'm not thrilled about...

Ehh... you can just stick `.replace(/the result of evaluating ([a-zA-Z_|0-9]+)/g, '<emu-meta effects="user-code">$1</emu-meta>` at the end of [this line](https://github.com/tc39/ecmarkup/blob/90438c13e442a2303830bd75d28a663e0e1fe12e/src/Algorithm.ts#L60) and call it good, I think.

We could do it property, probably in ecmarkdown, but that's... more work.

syg:
> - all calls to the Evaluate and ExecuteModule concrete methods for cyclic module records

Done

> - every use of the phrase "the result of evaluating", except for the one in JSON.parse - these are the actual roots, even though they're mostly intermediated through internal methods (this phrase probably deserves explicit support in ecmarkup)
most callsites of the HasBinding, CreateMutableBinding, InitializeBinding, SetMutableBinding, GetBindingValue, and DeleteBinding concrete methods of environment records (which are user-code-y because of with(proxy)) - these don't propagate for the same reason the [[-style internal methods don't propagate

Done in https://github.com/tc39/ecmarkup/pull/385

> calls to IsExtensible - I don't know what's up with that; the IsExtensible method itself is correctly marked as containing user code in its body

Done in https://github.com/tc39/ecmarkup/pull/385

> calls to [[GetPrototypeOf]], [[PreventExtensions]], and [[OwnPropertyKeys]] - I'm guessing your regex for internal methods just missed these

Done

> False positives:

> - the eight instances of Perform DefinePropertyOrThrow without a ! or ? (which should just get a !; presumably that will be fixed by Editorial: remove implicit wrapping/unwrapping of completion records #2547)

I opted to leave these to be automatically fixed later when they get a `!`.

> - the first step in RegExpCreate (which should be !; it can't throw here because it's using the built-in RegExp constructor)
the call to RegExpCreate in the evaluation semantics for regexp literals (RegExpCreate only calls user code if one of the arguments is an object with a trapped ToString)

Done

> - AllocateTypedArrayBuffer step 5 (it's ? because CreateByteDataBlock can throw, but it doesn't call user code; it uses the built-in ArrayBuffer constructor)

Done

> - SortCompare steps 7 and 9 - IsLessThan can only call user code when called with objects, not strings

Done

> Other issues:

> - the call to ToString in CreateMappedArgumentsObject step 19.b.ii.3 is marked as UC, but it's supposed to be associated with the surrounding _map_.[[DefineProperty]] call (which is explicitly marked as user code). This is presumably a bug in ecmarkup.

Done in https://github.com/tc39/ecmarkup/pull/385

> - the call to GetThisValue in GetValue step 4.c has the same issue and should have the same fix (probably there's others)

Done in https://github.com/tc39/ecmarkup/pull/385

bakkot:
Few more things:

- bac10dd can get reverted given https://github.com/tc39/ecmarkup/pull/385/commits/80959f08caff5f09f69bfac318848a140de8ab28
- apparently we invoke some SDOs as `Foo for |Bar|` instead of `Foo of |Bar|` - e.g. in FunctionDeclarationInstantiation step 25.a./26.a, we `Perform ? IteratorBindingInitialization for _formals_`. That should get marked as user code, but doesn't, presumably because ecmarkup doesn't know you can invoke SDOs with "for" (I didn't either - we only use this form about 50 times total, compared to ~1000 uses with "of").
- despite appearances, GeneratorStart doesn't actually execute user code; it just sets up code to be executed later (the first time `.next` is invoked). Not sure how best to deal with that - put `no-user-code` in the `dl` header, maybe? Or special case "set the code evaluation state" like we did for "a new Abstract Closure"?
- on the other hand, GeneratorResume effectively _does_ execute user code, by explicitly pushing and resuming an execution context which can contain user code. Ditto GeneratorResumeAbrupt.
- Re: GeneratorResume, `%StringIteratorPrototype%.next`, `%MapIteratorPrototype%.next`, and `%SetIteratorPrototype%.next` don't invoke user code despite calling GeneratorResume, because the execution contexts which they're resuming do not ever invoke user code. On the other hand, `%ArrayIteratorPrototype%.next` and `%RegExpStringIteratorPrototype%.next` _do_ invoke user code.

Didn't spot anything else wrong in some random spot-checks, except for the aforementioned DefinePropertyOrThrow issues (plus CreateMappedArgumentsObject step 19.b.ii.3, which should really be a `! DefinePropertyOrThrow`).

syg:
> apparently we invoke some SDOs as Foo for |Bar| instead of Foo of |Bar| - e.g. in FunctionDeclarationInstantiation step 25.a./26.a, we Perform ? IteratorBindingInitialization for _formals_. That should get marked as user code, but doesn't, presumably because ecmarkup doesn't know you can invoke SDOs with "for" (I didn't either - we only use this form about 50 times total, compared to ~1000 uses with "of").

I'll make a separate PR that changes these to "of" instead.

syg:
Note that 9a55c12 depends on https://github.com/tc39/ecmarkup/pull/388.

spec.html
```diff
@@ -44553,7 +44553,7 @@ <h1>
           1. Let _methodContext_ be the running execution context.
           1. Suspend _methodContext_.
           1. Set _generator_.[[GeneratorState]] to ~executing~.
-          1. Push _genContext_ onto the execution context stack; _genContext_ is now the running execution context.
+          1. <emu-meta effects="user-code">Push _genContext_ onto the execution context stack</emu-meta>; _genContext_ is now the running execution context.
```
bakkot, line 44556: Nit: I'd put this one the following step. The "Resume" is where the actual transfer of control happens.

Ditto below.

syg: Done.

bakkot:
You missed `GeneratorReturnAbrupt`, I think.

spec.html
```diff
@@ -14109,7 +14109,7 @@ <h1>
           1. Let _succeeded_ be ! OrdinaryDefineOwnProperty(_A_, *"length"*, _newLenDesc_).
           1. If _succeeded_ is *false*, return *false*.
           1. For each own property key _P_ of _A_ that is an array index, whose numeric value is greater than or equal to _newLen_, in descending numeric index order, do
-            1. Let _deleteSucceeded_ be ! _A_.[[Delete]](_P_).
+            1. Let _deleteSucceeded_ be ! <emu-meta effects="user-code">_A_.[[Delete]]</emu-meta>(_P_).
```
bakkot, line 14112: Because we know this is an Array exotic object, we know that this does not actually invoke user code. It's just the ordinary delete.

The first two uses of OrdinaryDefineOwnProperty in this method also do not invoke user code. I'm pretty sure they could be prefixed with `!`, so as with the above-mentioned unprefixed DefinePropertyOrThrow uses, #2547 will maybe fix them.

bakkot (commented):
Found a few false positives in the internal methods for various exotic objects.

- In the [internal methods for Arguments exotic objects](https://tc39.es/ecma262/multipage/#sec-arguments-exotic-objects-defineownproperty-p-desc), none of the things which operate on [[ParameterMap]] (usually spelled `_map_`) can invoke user code, since [[ParameterMap]] is an entirely spec-fictional object. Nor can the OrdinaryDelete call in the [[Delete]] section, since a user can't trap that call.  (The Get and Set calls on the arguments object itself can still invoke user code, since a user can add getters/setters to the arguments object.)

- In the [`[[Delete]]` semantics for Integer-Indexed exotic objects](https://tc39.es/ecma262/multipage/#sec-integer-indexed-exotic-objects-delete-p), the call to OrdinaryDelete can't call user code, because the only step in OrdinaryDelete which calls user code is the invocation of `[[GetOwnProperty]]`, and we know the `[[GetOwnProperty]]` semantics for Integer-Indexed exotic objects doesn't call user code. (In fact it also doesn't throw, so I think the `? OrdinaryDelete` here could become `! OrdinaryDelete`.)

- None of the internal methods of [module namespace exotic objects](https://tc39.es/ecma262/multipage/#sec-module-namespace-exotic-objects-setprototypeof-v) can invoke use code.

Otherwise LGTM. I'm sure there's remaining false positives and potentially false negatives, but I didn't see any except for those already mentioned when spot-checking stuff.

A surprisingly small diff, in the end, and would be smaller still if [ecmarkup understood concrete and/or internal methods](https://github.com/tc39/ecmarkup/issues/391).

syg:
> In the internal methods for Arguments exotic objects, none of the things which operate on [[ParameterMap]] ...

I feel like this is also one of those things better fixed by correct prefixes of `!` to those calls than `<emu-meta>` overrides slate in this PR.

Also interested in @michaelficarra's thoughts whether we should be more eager with manual suppressions for cases that could be getting fixed `!`.

> In the [[Delete]] semantics for Integer-Indexed exotic objects, the call to OrdinaryDelete can't call user code ...

This is just a 1 liner so I've changed this to a `! OrdinaryDelete`. I agree it can't call a Proxy trap since the receiver is guaranteed to be an IIEO.

> None of the internal methods of module namespace exotic objects can invoke use code.

I'd also like these to be automatically fixed by switching to `!` prefixing.

bakkot:
> I feel like this is also one of those things better fixed by correct prefixes of `!` to those calls than `<emu-meta>` overrides slate in this PR.

Fine with me.

> > None of the internal methods of module namespace exotic objects can invoke use code.
>
> I'd also like these to be automatically fixed by switching to `!` prefixing.

Well, in a couple cases cases ([`[[DefineOwnProperty]]`](https://ci.tc39.es/preview/tc39/ecma262/sha/ce8d4b36c89494241a619d211ea74af4bf3a996b/#sec-module-namespace-exotic-objects-defineownproperty-p-desc) step 2, e.g.) these methods are invoking other internal methods, and those sites are explicitly marked as calling user code by this PR. So those cases wouldn't be _automatically_ fixed.

And in a couple cases the operation can actually throw - calls to `[[Get]]` can throw if the module isn't linked (though I don't remember how that can happen...) or via the HostResolveImportedModule hook.

syg:
> Well, in a couple cases cases ([[DefineOwnProperty]] step 2, e.g.) these methods are invoking other internal methods, and those sites are explicitly marked as calling user code by this PR. So those cases wouldn't be automatically fixed.

Good point, done.

spec.html
```diff
@@ -34274,7 +34274,7 @@ <h1>String.prototype [ @@iterator ] ( )</h1>
               1. Let _nextIndex_ be _position_ + _cp_.[[CodeUnitCount]].
               1. Let _resultString_ be the substring of _s_ from _position_ to _nextIndex_.
               1. Set _position_ to _nextIndex_.
-              1. Perform ? Yield(_resultString_).
+              1. Perform ? GeneratorYield(_resultString_).
```
michaelficarra, line 34277: Don't these changes have to look like this?

```suggestion
              1. Perform ? GeneratorYield(! CreateIterResultObject(_resultString_, false)).
```

michaelficarra (approved):
LGTM otherwise. Only missing suppressions I saw while reviewing the rendering will be suppressed by `!` in the future. All the manual suppressions and annotations in this PR look correct. I didn't check for missing suppressions/annotations. I also didn't check whether the suppressions/annotations were necessary; I assume ecmarkup would complain for any it could determine through propagation.

bakkot:
> I also didn't check whether the suppressions/annotations were necessary; I assume ecmarkup would complain for any it could determine through propagation.

That's not currently implemented, but I've added it to the backlog for the linter. Since there wouldn't be any difference in the rendered output, I don't think landing that check needs to block this PR. (I don't think there are currently any redundant ones, though.)

spec.html
syg: Good catch, done.

bakkot (approved):
LGTM. Still some known false positives but that's fine.

syg:
@ljharb When you merge this, please squash. There's no value in keeping the separate commits for this.

domenic:
I may be blind, but is this supposed to be visible in the output? I cannot seem to find any differences in the live spec, although if I inspect element I do see various `<emu-meta>`s.

bakkot:
@domenic Hit `u` to toggle the annotation (you can also hit `?` pop up a modal listing such shortcuts).

PR #2559, opened 2021-10-28
Editorial: Remove <SP> and <NBSP> as the ones already included by <USP>

arhadthedev:
- [Table 39](https://tc39.es/ecma262/#table-white-space-code-points) explicitly lists \<SP> and \<NBSP> to be later used as a right-hand side of `WhiteSpace`. However, they are already implicitly included there by neighbour \<USP> because of their `Zs` category and need no repetition (see <https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt>).

  *This repetition is minor, but deduplication need to be done manually or by a generator while creation of a parser. A simple removal of explicit \<SP> and \<NBSP> allows to avoid such a step.*

  Also I added a note explaining where space and NBSP characters have gone:

  > U+0020 (SPACE) and U+00A0 (NO-BREAK SPACE) code points are part of \<USP>.

   but I can remove it if a reader would not care about where the grammar introduces spaces.

- In addition, with another commit I removed a line that just repeats the grammar (unconditional inclusion of \<USP\> as a right-hand side of `WhiteSpace`) just before the relevant declaration:

  > ECMAScript implementations must recognize as *WhiteSpace* code points listed in the “Space_Separator” (“Zs”) category.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -16204,28 +16204,6 @@
         </tr>
         <tr>
           <td>
-            `U+0020`
-          </td>
-          <td>
-            SPACE
-          </td>
-          <td>
-            &lt;SP&gt;
-          </td>
-        </tr>
-        <tr>
-          <td>
-            `U+00A0`
-          </td>
-          <td>
-            NO-BREAK SPACE
-          </td>
-          <td>
-            &lt;NBSP&gt;
-          </td>
-        </tr>
-        <tr>
-          <td>
             `U+FEFF`
           </td>
           <td>
@@ -16237,10 +16215,10 @@
         </tr>
         <tr>
           <td>
-            Other category &ldquo;Zs&rdquo;
+            Category &ldquo;Zs&rdquo;
           </td>
           <td>
-            Any other Unicode &ldquo;Space_Separator&rdquo; code point
+            Any Unicode &ldquo;Space_Separator&rdquo; code point
           </td>
           <td>
             &lt;USP&gt;
@@ -16249,7 +16227,9 @@
         </tbody>
       </table>
     </emu-table>
-    <p>ECMAScript implementations must recognize as |WhiteSpace| code points listed in the &ldquo;Space_Separator&rdquo; (&ldquo;Zs&rdquo;) category.</p>
+    <emu-note>
+      <p>U+0020 (SPACE) and U+00A0 (NO-BREAK SPACE) code points are part of &lt;USP&gt;.</p>
+    </emu-note>
     <emu-note>
       <p>Other than for the code points listed in <emu-xref href="#table-white-space-code-points"></emu-xref>, ECMAScript |WhiteSpace| intentionally excludes all code points that have the Unicode &ldquo;White_Space&rdquo; property but which are not classified in category &ldquo;Space_Separator&rdquo; (&ldquo;Zs&rdquo;).</p>
     </emu-note>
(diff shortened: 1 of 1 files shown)


mathiasbynens (approved):
Indeed, `assert(/^\p{Space_Separator}+$/u.test('\x20\xA0'))`.

jmdyck:
The row for `<USP>` says `*Other* category “Zs”` and `Any *other* Unicode “Space_Separator” code point`. I interpret that as "any Zs/Space_Separator code points *other than ones that already appeared in this table*", so there isn't *technically* a repetition. However, things would be slightly simpler with this PR, and I can't see a drawback. (E.g., there's no other reference to `<SP>` or `<NBSP>`.)

In the table, you should probably change "Other category" to "Category". (Or maybe "Unicode category".)

jmdyck:
> that's strange, it looks correct (if I entered it correctly for bash) - both traditional and non-breaking spaces pass as `White_Space` for hosts that support unicode property escapes

What's strange? U+0020 and U+00A0 both have the Unicode "White_Space" property.

But I'm not sure why @mathiasbynens brought this up. As the existing NOTE says, having the "White_Space" property is not relevant to being ECMAScript *WhiteSpace*. (E.g., U+000A, U+000D, U+2028 and U+2029 all have the "White_Space" property, but they're not in Zs and not in table 39, so they're not ES *WhiteSpace*. However, they are all *LineTerminator*s.)

arhadthedev:
@jmdyck 

> In the table, you should probably change "Other category" to "Category". (Or maybe "Unicode category".)

Done. However, I left "Unicode" out for now to avoid repetition of the next cell.

> What's strange?

Mathias mentioned an assertion, so I suggested that the pull request somehow breaks it.

gibson042:
I personally find the explicit mention of SPACE (and to a lesser extent, NO-BREAK SPACE) to be useful even though it is not necessary.

arhadthedev:
@gibson042, I think there is no semantic difference between, for example, NO-BREAK SPACE and NARROW NO-BREAK SPACE, or SPACE and IDEOGRAPHIC SPACE to explicitly mention the former leaving the latter behind a reference to Space_Separator. Such separation was convenient in a transition period between single-octet code pages and Unicode when people thought in terms of Windows-1252 treating everything else as something extra. However, now it has no sense.

Moreover, all space characters are equal from the Unicode point of view (being just various kinds of ligature-breaking spacing), so the specification should treat them equally as well. Rarity cannot be a reason here because they all are an inherent and equally valid feature of Unicode.

mathiasbynens:
> But I'm not sure why @mathiasbynens brought this up. As the existing NOTE says, having the "White_Space" property is not relevant to being ECMAScript _WhiteSpace_. (E.g., U+000A, U+000D, U+2028 and U+2029 all have the "White_Space" property, but they're not in Zs and not singled out in table 39, so they're not ES _WhiteSpace_. However, they are all ES _LineTerminator_s.)

You’re right, I meant `/^\p{Space_Separator}+$/u.test('\x20\xA0')`. Retroactively corrected my comment to reduce confusion. Anyway, I approved the PR as I posted my initial comment, so I also don’t understand what’s “strange” — the proposed patch looks good to me!

arhadthedev:
@mathiasbynens It looks like I misread a purpose behind your first comment, my apologizes for that.

gibson042:
There is no semantic difference, but there is a practical difference in that U+0020 SPACE constitutes the overwhelming majority of whitespace in ECMAScript source and is inarguably the type specimen _of_ Space_Separator. I'm not strongly opposed, though, just sharing a data point.

syg (approved):
Seems fine to me. I don't think any signal is lost by moving the space and nbsp inclusion into a note from the grammar.

michaelficarra (approved):

PR #2562, opened 2021-10-30
Editorial: Fix false claim in notes for find/findIndex

bakkot:
This text was presumably copied from `map` or a similar older method, but these two more modern methods do an unconditional [[Get]] rather than gating on [[HasProperty]].


bakkot:
Ah dang beaten by five minutes. https://github.com/tc39/ecma262/pull/2561

PR #2564, opened 2021-11-02
Markup: auto-format

bakkot:
Automatic formatting of the markup using [a ~WIP~ ecmarkup formatter](https://github.com/tc39/ecmarkup/pull/367).

You can see that it mostly uses the same style we already do, i.e. does not change anything, with a few exceptions:
- useless `<tbody>` tags are dropped
- when referring to an optional nonterminal in prose, it's written `|Nonterminal?|` rather than `|Nonterminal_opt|`. We use both styles currently, but the `_opt` form more frequently; I decided to go with the less frequent style anyway because it matches how it's spelled in the grammar (which does not support the `_opt` suffix in source)
- when there's an `emu-grammar` which references the same LHS twice, those are collapsed into a single multiline production
- when there's an `emu-grammar` with some items expanded, all of the items are expanded

(The latter two have effects on the rendering. That's intentional.)

I recommend not manually reviewing the first commit; that just fiddles with the whitespace for a few tables so that the later commits are more readable.

~Marked as a draft until the formatter is ready and included in this PR, both as an npm script and enforced as a github action.~

Formatter is ready and included in this PR, both as an npm script and enforced as a github action.

### DIFF
--- a/.github/workflows/enforce-format.yml
+++ b/.github/workflows/enforce-format.yml
@@ -1,0 +1,17 @@
+name: 'enforce format'
+
+on: [pull_request]
+
+jobs:
+  build:
+    name: 'enforce format'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ljharb/actions/node/install@56f291b2a0d9e3bce72634356ae949b1e054b9f8
+        name: 'nvm install lts/* && npm ci'
+        with:
+          node-version: lts/*
+          use-npm-ci: true
+      - run: 'if ! $(npm bin)/emu-format --check spec.html; then echo "You need to run `npm run format`"; exit 1; fi'
--- a/spec.html
+++ b/spec.html
@@ -16520,6 +16520,7 @@
       <p>The term <dfn variants="conditional keywords">conditional keyword</dfn>, or <dfn variants="contextual keywords">contextual keyword</dfn>, is sometimes used to refer to the keywords that fall in the last three categories, and thus can be used as identifiers in some contexts and as keywords in others.</p>
       <h2>Syntax</h2>
       <emu-grammar type="definition">
+        // emu-format ignore
         ReservedWord :: one of
           `await`
           `break`
@@ -16561,6 +16562,7 @@
       OptionalChainingPunctuator ::
         `?.` [lookahead &notin; DecimalDigit]
 
+      // emu-format ignore
       OtherPunctuator :: one of
         `{` `(` `)` `[` `]`
         `.` `...` `;` `,`
@@ -16717,6 +16719,7 @@
           HexDigits[?Sep] HexDigit
           [+Sep] HexDigits[+Sep] NumericLiteralSeparator HexDigit
 
+        // emu-format ignore
         HexDigit :: one of
           `0` `1` `2` `3` `4` `5` `6` `7` `8` `9` `a` `b` `c` `d` `e` `f` `A` `B` `C` `D` `E` `F`
       </emu-grammar>
@@ -20385,6 +20388,7 @@
         LeftHandSideExpression[?Yield, ?Await] `||=` AssignmentExpression[?In, ?Yield, ?Await]
         LeftHandSideExpression[?Yield, ?Await] `??=` AssignmentExpression[?In, ?Yield, ?Await]
 
+      // emu-format ignore
       AssignmentOperator : one of
         `*=` `/=` `%=` `+=` `-=` `&lt;&lt;=` `&gt;&gt;=` `&gt;&gt;&gt;=` `&amp;=` `^=` `|=` `**=`
     </emu-grammar>
@@ -20450,6 +20454,7 @@
         1. Let _assignmentOpText_ be the source text matched by |AssignmentOperator|.
         1. Let _opText_ be the sequence of Unicode code points associated with _assignmentOpText_ in the following table:
           <figure>
+            <!-- emu-format ignore -->
             <table class="lightweight-table">
               <tr><th> _assignmentOpText_ </th><th> _opText_       </th></tr>
               <tr><td> `**=`              </td><td> `**`           </td></tr>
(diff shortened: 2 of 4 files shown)


ljharb:
Could we make it keep the tbody tags? The browser adds them anyways, and it's nicer imo when the source HTML(ish) matches the output as closely as possible.

bakkot:
We could, but it's just boilerplate with no function, and an extra level of indentation. I'd strongly prefer not to.

ljharb:
The way we've always written the spec, there isn't an extra level of indentation.

bakkot:
Well, roughly a third of the tables in the spec _do_ have that extra level of indentation, so I think that special case is not obvious to contributors. I'm happier not special casing `<tbody>` so that level of indentation is always equivalent to level of nesting (with the exception of the top-level `html`, `head`, and `body` elements.)

ljharb:
So would the formatter keep tbody when there's a thead or tfoot?

The issue then is that there's diff churn; you have to change a lot more of the document when adding the first or removing the last header/footer.

bakkot:
> So would the formatter keep tbody when there's a thead or tfoot?

No.

ljharb:
I'm confused; when there's a thead or tfoot doesn't there have to be a tbody? or does the browser add that as well.

bakkot:
From https://html.spec.whatwg.org/dev/tables.html#the-tbody-element:

> A tbody element's start tag can be omitted if the first thing inside the tbody element is a tr element, and if the element is not immediately preceded by a tbody, thead, or tfoot element whose end tag has been omitted. (It can't be omitted if the element is empty.)
> A tbody element's end tag can be omitted if the tbody element is immediately followed by a tbody or tfoot element, or if there is no more content in the parent element.

arhadthedev (changes requested):
I think all `emu-format ignore` comments can be replaced with formatter heuristics based on reasons of ignoring:

bakkot:
It's true that I could add more specialized heuristics to the formatter, but I prefer to use comments for cases like these.

syg (approved):
lgtm

+1 from me for removing `<tbody>` tags.

spec.html
```diff
@@ -64,7 +61,7 @@
 <div id="metadata-block">
   <h1>About this Specification</h1>
   <p>The document at <a href="https://tc39.es/ecma262/">https://tc39.es/ecma262/</a> is the most accurate and up-to-date ECMAScript specification. It contains the content of the most recent yearly snapshot plus any <a href="https://github.com/tc39/proposals/blob/HEAD/finished-proposals.md">finished proposals</a> (those that have reached Stage&nbsp;4 in the <a href="https://tc39.es/process-document/">proposal process</a> and thus are implemented in several implementations and will be in the next practical revision) since that snapshot was taken.</p>
-  <p>This document is available as <a href="">a single page</a> and as <a href="multipage/">multiple pages</a>.</p>
+  <p>This document is available as <a href>a single page</a> and as <a href="multipage/">multiple pages</a>.</p>
```
ljharb, line 64: i don't think these are the same - isn't `<a href>` equivalent to `<a href="href">`?

spec.html
```diff
@@ -109,18 +107,12 @@ <h1>Introduction</h1>
...
-  <p>
-    Brian Terlson<br>
-    ECMA-262, Project Editor, 7<sup>th</sup> through 10<sup>th</sup> Editions
-  </p>
-  <p>
-    Jordan Harband<br>
-    ECMA-262, Project Editor, 10<sup>th</sup> through 12<sup>th</sup> Editions
-  </p>
+  <p>Allen Wirfs-Brock<br>
+    ECMA-262, Project Editor, 6<sup>th</sup> Edition</p>
```
ljharb, line 111: oof, this is really gross - can we not make the `<p>` indentation match the `</p>`?

spec.html
bakkot: > isn't `<a href>` equivalent to `<a href="href">`?

No.

spec.html
bakkot: > oof, this is really gross - can we not make the `<p>` indentation match the `</p>`?

I'm not sure what you're suggesting; can you write it out?

ljharb: Basically the way it was before the change. `<p>`, a newline, the indented content, a newline, and `</p>`, just like any OTBS style would do with braces.

bakkot: Every other paragraph in the spec has the text on the same line as the `<p>`. I am reluctant to relax that rule and I don't really want to special-case these.

bakkot: Yes they do; there's something like a dozen cases that have newlines inside of the `<p>` and also have the text on the same line as the tags.

I'm OK adding a rule which says that if there are multiple lines in the `<p>` then it goes on a different line than the opening tag, but that's _not_ the rule we currently follow.

bakkot: Pushed a commit using the proposed rule. It will fail CI unless/until https://github.com/tc39/ecmarkup/pull/369 is included in this PR, but you can see what it looks like, at least.

.github/workflows/enforce-format.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ljharb/actions/node/install@56f291b2a0d9e3bce72634356ae949b1e054b9f8
+        name: 'nvm install lts/* && npm ci'
+        with:
+          node-version: lts/*
+          use-npm-ci: true
+      - run: 'if ! $(npm bin)/emu-format --check spec.html; then echo "You need to run ''npm run format''"; exit 1; fi'
```
ljharb, line 17:
```suggestion
      - run: 'if ! $(npm bin)/emu-format --check spec.html; then echo 'You need to run `npm run format`'; exit 1; fi'
```

spec.html
ljharb: thanks, that looks SO much better.

PR #2566, opened 2021-11-04
Editorial: mostly stop using strings for special values

bakkot:
Replaces/ closes #2155.

There are four special strings used in the import/export machinery, one of which is used in three different ways:
- `"*"`, used in the [[ImportName]] field of an ExportEntry to mark an `export * from 'mod'`; this has been renamed to `~all-but-default~`
- `"*"`, used in the [[ImportName]] field of an ExportEntry to mark an `export * as foo from 'mod'` export; this has been renamed to `~all~`
- `"*"`, used in the [[ImportName]] field of an ImportEntry to mark an `import * as foo from 'mod'` import; this has been renamed to `~namespace-object~`
- `"*namespace*"`, used in the [[BindingName]] field of a ResolvedBinding to mark a `export * as foo from 'mod'`; this has been renamed to `~namespace~`
- `"ambiguous"`, used in the return value for `ResolveExport` when resolving a name exported by multiple `export * from 'mod'` exports; this has been renamed to `~ambiguous~`.
- `"*default*"`, used for default exports which do not have any other name; this has been left alone, and the note about it expanded.

The goal is to unblock https://github.com/tc39/ecma262/pull/2154 by ensuring that we never use a special string as a placeholder in a context where that PR would allow arbitrary well-formed strings to appear.

@michaelficarra and I spent a couple hours staring at all of this stuff, most of which was fixing our own confusion about how this worked, to wit:

We originally thought that `"*default*"` was such a string, but it turns out not to be - it is only used as a name strictly internal to a module, to provide the linkage between the `"default"` export and the local binding by using the same machinery as named exports.

And internally, for named exports, that linkage uses string names looked up in the actual scope object for the module. So the `"*default*"` string is used in a context which does actually require a string, and thus can't be trivially replaced by a spec enum. We considered modifying the machinery for default exports to work in a different way, but that seemed like it would be more complexity for little benefit, given that it does _not_ actually conflict with #2154, so we've left it along. I did expand the note about `"*default*"` to hopefully make this clearer to future readers.

On the other hand, `"*"` _was_ actually being used in a position which would conflict with #2154, as you can see from the [[ImportName]] column of the `import * as ns from "mod";` row in [this table](https://tc39.es/ecma262/#table-import-forms-mapping-to-importentry-records). So was `"*namespace*"`, as you can sort of see from step 9/10 of [this algorithm](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-module-namespace-exotic-objects-get-p-receiver). (Edit: actually I guess `"*namespace*"` was ok, on thinking further - step 9 happens after the public -> local name translation has already occurred.)

So this PR does still block #2154, even though we didn't actually end up needing to deal with `"*default*"`.

---

Just to summarize the relevant machinery here, mostly for my own benefit:

An `import` creates an ImportEntry Record whose [[ImportName]] is the exported name and whose [[LocalName]] is strictly internal to the module. [[ImportName]] has/had a special case of `"*"` which indicated a request for the namespace object rather than any particular name.

An `export ... from 'foo'` creates an ExportEntry Record whose [[ExportName]] is the name visible to importers of this module and whose [[ImportName]] is the name exported by the upstream module ('foo'), and has/had the same special case for `"*"`.

An `export ...` without a `from` creates an ExportEntry Record whose [[ExportName]] is the name visible to importers of this module and whose [[LocalName]] is the name used within this module. [[LocalName]] has a special case of `"*default*"` for anonymous default exports, which is actually a name used on the module's Environment object alongside other top-level variables.

Resolution hooks up these entries to create a ResolvedBinding Record which holds the original module defining the export, and whose [[BindingName]] is the name within that module's environment. It has/had a special case of `"*namespace*"` which indicated a request for the namespace object rather than any particular name. The [[BindingName]] is derived from an ExportEntry's [[LocalName]], except when the request is for a namespace object. As such, it can be `"*default*"`, which will get looked up within the module's environment like any other name.

Also, if resolution is ambiguous, the string `"ambiguous"` is used instead of a ResolvedBinding. It's not used in a position where names occur at all.

So, the [[ImportName]] / [[ExportName]] fields hold the actual linker-exposed names, which are the things relaxed by #2154; the other names are strictly internal to a module. Of the special strings, only `"*"` was used in the linker-exposed position (in [[ImportName]]) and so only `"*"` actually needed to change. But we might as well get rid of the others while we're at it.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -14941,7 +14941,7 @@
           1. Assert: _binding_ is a ResolvedBinding Record.
           1. Let _targetModule_ be _binding_.[[Module]].
           1. Assert: _targetModule_ is not *undefined*.
-          1. If _binding_.[[BindingName]] is *"\*namespace\*"*, then
+          1. If _binding_.[[BindingName]] is ~namespace~, then
             1. Return ? GetModuleNamespace(_targetModule_).
           1. Let _targetEnv_ be _targetModule_.[[Environment]].
           1. If _targetEnv_ is ~empty~, throw a *ReferenceError* exception.
@@ -26140,7 +26140,7 @@
                 ResolveExport(_exportName_ [, _resolveSet_])
               </td>
               <td>
-                <p>Return the binding of a name exported by this module. Bindings are represented by a <dfn id="resolvedbinding-record" variants="ResolvedBinding Records">ResolvedBinding Record</dfn>, of the form { [[Module]]: Module Record, [[BindingName]]: String }. If the export is a Module Namespace Object without a direct binding in any module, [[BindingName]] will be set to *"\*namespace\*"*. Return *null* if the name cannot be resolved, or *"ambiguous"* if multiple bindings were found.</p>
+                <p>Return the binding of a name exported by this module. Bindings are represented by a <dfn id="resolvedbinding-record" variants="ResolvedBinding Records">ResolvedBinding Record</dfn>, of the form { [[Module]]: Module Record, [[BindingName]]: String | ~namespace~ }. If the export is a Module Namespace Object without a direct binding in any module, [[BindingName]] will be set to ~namespace~. Return *null* if the name cannot be resolved, or *"ambiguous"* if multiple bindings were found.</p>
                 <p>Each time this operation is called with a specific _exportName_, _resolveSet_ pair as arguments it must return the same result if it completes normally.</p>
               </td>
             </tr>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -28366,7 +28367,7 @@ <h1>Static Semantics: ExportEntries</h1>
           1. Return a List whose sole element is _entry_.
         </emu-alg>
         <emu-note>
-          <p>*"\*default\*"* is used within this specification as a synthetic name for anonymous default export values.</p>
+          <p>*"\*default\*"* is used within this specification as a synthetic name for anonymous default export values. See <emu-xref href="#note-star-default-star">this note</emu-xref> for more details.</p>
```
michaelficarra, line 28370:
```suggestion
          <p>*"\*default\*"* is used within this specification as a synthetic name for anonymous default export values. See <emu-xref href="#note-star-default-star"></emu-xref> for more details.</p>
```

I don't like links that say things like "this" or "here".

bakkot: That note does not have a name or section number, so there's nothing which would be generated there. We need to put _some_ text there.

arhadthedev: Maybe `this note` → `a note on BoundNames`?

ljharb:
awesome, very happy to see #2155 replaced.

spec.html
```diff
@@ -27671,13 +27672,13 @@ <h1>
             1. For each ExportEntry Record _e_ of _module_.[[StarExportEntries]], do
               1. Let _importedModule_ be ? HostResolveImportedModule(_module_, _e_.[[ModuleRequest]]).
               1. Let _resolution_ be ? _importedModule_.ResolveExport(_exportName_, _resolveSet_).
-              1. If _resolution_ is *"ambiguous"*, return *"ambiguous"*.
+              1. If _resolution_ is ~ambiguous~ return ~ambiguous~.
```
syg, line 27675:
```suggestion
              1. If _resolution_ is ~ambiguous~, return ~ambiguous~.
```

spec.html
syg: I prefer the original text of "this note".

syg (approved):
Thanks for writing the explanations, made reviewing this a lot easier.

PR #2588, opened 2021-11-25
Editorial: Use abstract closures in `Promise` static methods

ExE-Boss:
**Refs:** <https://github.com/tc39/ecma262/pull/2439>

---

I left <code>[CreateResolvingFunctions]</code> alone because the [Promise Resolve Functions] algorithm steps are rather long compared to `CreateResolvingFunctions` that has been converted to use Abstract Closures, and I didn’t want to only convert it halfway.

[CreateResolvingFunctions]: https://tc39.es/ecma262/#sec-createresolvingfunctions
[Promise Resolve Functions]: https://tc39.es/ecma262/#sec-promise-resolve-functions

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -43650,7 +43650,7 @@ THH:mm:ss.sss
           </emu-alg>
         </emu-clause>
 
-        <emu-clause id="sec-performpromiseall" type="abstract operation">
+        <emu-clause id="sec-performpromiseall" oldids="sec-promise.all-resolve-element-functions" type="abstract operation">
           <h1>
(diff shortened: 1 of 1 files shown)


bakkot:
Thanks for the PR. I'm not totally convinced this is an improvement, which is why I did not convert these as part of #2439. We can see what the other editors think, though.

syg:
Thanks for the PR, @ExE-Boss. We discussed this on the editor call, and none of us thought it was a strict improvement. So erring on the side of less churn, we're going to keep the current forms for now.

PR #2590, opened 2021-11-26
Make `npm run build-*` working on Windows

arhadthedev:
Currently, `npm run build-only` and `npm run build-snapshot` do not work on Windows (`rm` and `cp` not found). The reason is invocation of  scripts `pre-*`→`clean`→`rm` and `prebuild-only`→`cp`.

This commit fixes it by using [command line tools listed in bcoe/awesome-cross-platform-nodejs](https://github.com/bcoe/awesome-cross-platform-nodejs#filesystem):

- rimraf to replace `rm -rf` (2.2.0 introduced a command line utility that will never break its `rimraf <dirname>` basic functionality so no upper version limit)
- cpy-cli to replace `cp -R` (any version since it never broke and will never break its `cpy <src> <dst>` use case).


Jack-Works (approved):

package.json
```diff
@@ -27,7 +27,9 @@
   "license": "SEE LICENSE IN https://tc39.es/ecma262/#sec-copyright-and-software-license",
   "homepage": "https://tc39.es/ecma262/",
   "dependencies": {
-    "ecmarkup": "9.6.3"
+    "ecmarkup": "9.6.3",
+    "cpy-cli": "*",
```
ljharb, line 31: * is dangerous and should never be used in a package.json, ever anywhere.

package.json
```diff
@@ -27,7 +27,9 @@
...
+    "rimraf": ">=2.2.0"
```
ljharb, line 32: >= is dangerous for the same reason.

ljharb (dismissed):

package.json
arhadthedev: Fixed. I understand that I got too optimistic about forward compatibility even for features that seem basic.

package.json
```diff
@@ -27,7 +27,9 @@
   "license": "SEE LICENSE IN https://tc39.es/ecma262/#sec-copyright-and-software-license",
   "homepage": "https://tc39.es/ecma262/",
   "dependencies": {
-    "ecmarkup": "9.6.3"
+    "ecmarkup": "9.6.3",
+    "cpy-cli": "^3.1.1",
```
bakkot, line 31: Installing this package adds 49 transitive dependencies, which seems excessive for such a trivial utility. While I'm not opposed to using a dependency for this, I would prefer something with fewer transitive dependencies, ideally zero.

arhadthedev: I found and applied another package ([cpr](https://www.npmjs.com/package/cpr)) with almost no dependencies:

- graceful-fs (has no dependencies)
- minimist (has no dependencies)
- mkdirp (has no dependencies)
- rimraf (already a dependency of ecma262 itself)

package.json
```diff
@@ -6,17 +6,17 @@
   "scripts": {
     "ipr-check": "node scripts/check-form tc39/ecma262 \"$TRAVIS_COMMIT\"",
     "prebuild-spec": "git remote rm travis-origin 2>/dev/null ; git remote add travis-origin \"git@github.com:${TRAVIS_REPO_SLUG:-tc39/ecma262}.git\" && git fetch --quiet travis-origin",
-    "build-spec": "YEAR=2021 && git checkout --quiet \"es${YEAR}\" && mkdir -p \"out/${YEAR}\" && cp -R img \"out/${YEAR}\" && ecmarkup --verbose spec.html \"out/${YEAR}/index.html\" --css \"out/${YEAR}/ecmarkup.css\" --js \"out/${YEAR}/ecmarkup.js\" && git checkout --quiet travis-origin/test-travis",
+    "build-spec": "YEAR=2021 && git checkout --quiet \"es${YEAR}\" && mkdir -p \"out/${YEAR}\" && cpr img \"out/${YEAR}/img\" && ecmarkup --verbose spec.html \"out/${YEAR}/index.html\" --css \"out/${YEAR}/ecmarkup.css\" --js \"out/${YEAR}/ecmarkup.js\" && git checkout --quiet travis-origin/test-travis",
     "postbuild-spec": "git remote rm travis-origin",
     "build-head": "npm run build-only -- --lint-spec --strict",
-    "prebuild-only": "npm run clean && mkdir out && cp -R img out",
+    "prebuild-only": "npm run clean && mkdir out && cpr img out/img/",
```
ljharb, line 12:
```suggestion
    "prebuild-only": "npm run clean && mkdir out && node -pe \"fs.copyFileSync('img', 'out/img/')\"",
```
maybe this doesn't need a dependency at all?

package.json
```diff
@@ -6,17 +6,17 @@
...
     "build-only": "ecmarkup --no-ecma-262-biblio --verbose spec.html --multipage out",
     "build": "npm run build-head",
     "build-for-pdf": "npm run build-head -- --old-toc",
     "build-travis": "npm run build-head && npm run build-spec",
     "prebuild-snapshot": "npm run clean",
     "build-snapshot": "npm run build-head && node scripts/insert_snapshot_warning.js",
-    "clean": "rm -rf out",
+    "clean": "rimraf out",
```
ljharb, line 19:
```suggestion
    "clean": "node -pe \"fs.rmdirSync('out', { recursive: true })\"",
```

ljharb (commented):
just a thought; haven't tested this on the command line

package.json
arhadthedev: Unfortunately, `fs.copyFile` cannot copy directories so a one-liner is impossible here. I also tried `fs.symlinkSync` but it requires admin privileges on Windows (while hard links and junction points do not).

In addition I would like to note that I intentionally left `build-spec` unported. I see no sence to introduce another package (`mkdirp` for `mkdir -p`) for a script used by CI and useless for a repo user.

package.json
arhadthedev: I like the idea of bare and direct invocation with no wrappers. For `rmdirSync`, Node 16.1.0 emits a deprecation, so I used:

```json
    "clean": "node -e \"fs.rmSync('out', { recursive: true })\"",
```

However, I do not know whether we need to omit a package already installed by `cpr` (if no native way of directory copying is found of course).

<details>
<summary>Indeed, here is a patch just for the case</summary>

```diff
diff --git a/package.json b/package.json
index fc86dd2..5b7327e 100644
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "build-travis": "npm run build-head && npm run build-spec",
     "prebuild-snapshot": "npm run clean",
     "build-snapshot": "npm run build-head && node scripts/insert_snapshot_warning.js",
-    "clean": "rimraf out",
+    "clean": "node -e \"fs.rmSync('out', { recursive: true })\"",
     "format": "emu-format --write spec.html",
     "test": "exit 0",
     "watch": "npm run build-only -- --lint-spec --watch",
@@ -28,8 +28,7 @@
   "homepage": "https://tc39.es/ecma262/",
   "dependencies": {
     "cpr": "^3.0.1",
-    "ecmarkup": "9.6.3",
-    "rimraf": "^3.0.2"
+    "ecmarkup": "9.6.3"
   },
   "devDependencies": {
     "glob": "^7.1.6",
diff --git a/package-lock.json b/package-lock.json
index cdacd7f..d8afee9 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,7 @@
       "license": "SEE LICENSE IN https://tc39.es/ecma262/#sec-copyright-and-software-license",
       "dependencies": {
         "cpr": "^3.0.1",
-        "ecmarkup": "9.6.3",
-        "rimraf": "^3.0.2"
+        "ecmarkup": "9.6.3"
       },
       "devDependencies": {
         "glob": "^7.1.6",
@@ -549,17 +548,6 @@
         "cpr": "bin/cpr"
       }
     },
-    "node_modules/cpr/node_modules/rimraf": {
-      "version": "2.7.1",
-      "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-2.7.1.tgz",
-      "integrity": "sha512-uWjbaKIK3T1OSVptzX7Nl6PvQ3qAGtKEtVRjRuazjfL3Bx5eI409VZSqgND+4UNnmzLVdPj9FqFJNPqBZFve4w==",
-      "dependencies": {
-        "glob": "^7.1.3"
-      },
-      "bin": {
-        "rimraf": "bin.js"
-      }
-    },
     "node_modules/cross-spawn": {
       "version": "7.0.3",
       "resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-7.0.3.tgz",
@@ -1288,6 +1276,20 @@
         "node": "^10.12.0 || >=12.0.0"
       }
     },
+    "node_modules/flat-cache/node_modules/rimraf": {
+      "version": "3.0.2",
+      "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-3.0.2.tgz",
+      "integrity": "sha512-JZkJMZkAGFFPP2YqXZXPbMlMBgsxzE8ILs4lMIX/2o0L9UBw9O/Y3o6wFw/i9YLapcUJWwqbi3kdxIPdC62TIA==",
+      "dependencies": {
+        "glob": "^7.1.3"
+      },
+      "bin": {
+        "rimraf": "bin.js"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/isaacs"
+      }
+    },
     "node_modules/flatted": {
       "version": "3.2.4",
       "resolved": "https://registry.npmjs.org/flatted/-/flatted-3.2.4.tgz",
@@ -2189,17 +2191,14 @@
       }
     },
     "node_modules/rimraf": {
-      "version": "3.0.2",
-      "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-3.0.2.tgz",
-      "integrity": "sha512-JZkJMZkAGFFPP2YqXZXPbMlMBgsxzE8ILs4lMIX/2o0L9UBw9O/Y3o6wFw/i9YLapcUJWwqbi3kdxIPdC62TIA==",
+      "version": "2.7.1",
+      "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-2.7.1.tgz",
+      "integrity": "sha512-uWjbaKIK3T1OSVptzX7Nl6PvQ3qAGtKEtVRjRuazjfL3Bx5eI409VZSqgND+4UNnmzLVdPj9FqFJNPqBZFve4w==",
       "dependencies": {
         "glob": "^7.1.3"
       },
       "bin": {
         "rimraf": "bin.js"
-      },
-      "funding": {
-        "url": "https://github.com/sponsors/isaacs"
       }
     },
     "node_modules/run-parallel": {
@@ -3201,16 +3200,6 @@
         "minimist": "^1.2.0",
         "mkdirp": "~0.5.1",
         "rimraf": "^2.5.4"
-      },
-      "dependencies": {
-        "rimraf": {
-          "version": "2.7.1",
-          "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-2.7.1.tgz",
-          "integrity": "sha512-uWjbaKIK3T1OSVptzX7Nl6PvQ3qAGtKEtVRjRuazjfL3Bx5eI409VZSqgND+4UNnmzLVdPj9FqFJNPqBZFve4w==",
-          "requires": {
-            "glob": "^7.1.3"
-          }
-        }
       }
     },
     "cross-spawn": {
@@ -3786,6 +3775,16 @@
       "requires": {
         "flatted": "^3.1.0",
         "rimraf": "^3.0.2"
+      },
+      "dependencies": {
+        "rimraf": {
+          "version": "3.0.2",
+          "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-3.0.2.tgz",
+          "integrity": "sha512-JZkJMZkAGFFPP2YqXZXPbMlMBgsxzE8ILs4lMIX/2o0L9UBw9O/Y3o6wFw/i9YLapcUJWwqbi3kdxIPdC62TIA==",
+          "requires": {
+            "glob": "^7.1.3"
+          }
+        }
       }
     },
     "flatted": {
@@ -4460,9 +4459,9 @@
       "integrity": "sha512-U9nH88a3fc/ekCF1l0/UP1IosiuIjyTh7hBvXVMHYgVcfGvt897Xguj2UOLDeI5BG2m7/uwyaLVT6fbtCwTyzw=="
     },
     "rimraf": {
-      "version": "3.0.2",
-      "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-3.0.2.tgz",
-      "integrity": "sha512-JZkJMZkAGFFPP2YqXZXPbMlMBgsxzE8ILs4lMIX/2o0L9UBw9O/Y3o6wFw/i9YLapcUJWwqbi3kdxIPdC62TIA==",
+      "version": "2.7.1",
+      "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-2.7.1.tgz",
+      "integrity": "sha512-uWjbaKIK3T1OSVptzX7Nl6PvQ3qAGtKEtVRjRuazjfL3Bx5eI409VZSqgND+4UNnmzLVdPj9FqFJNPqBZFve4w==",
       "requires": {
         "glob": "^7.1.3"
       }
```

</details>

package.json
arhadthedev: I found that `fs.symlinkSync` totally ignores its `type='junction'` argument on non-Windows systems so it can be used without checks and restrictions. 

Thanks to it, this PR finally got rid of third party dependencies.

package.json
arhadthedev: Update: `cpr` is no more, so I used your proposal with extra `force: true` to not abort when `out` does not exists yet.

arhadthedev:
@ljharb @bakkot I replaced both rimraf and cpr with `node -e`.

Also, I pushed an empty commit to update a link to ci.tc39.es (old `https://ci.tc39.es/preview/tc39/ecma262/sha/.../` does not work anymore). However, it seems to be out of service for a while.

ljharb:
hm, thanks for the heads up; i'll look into why it's not working.

ljharb:
Seems like today's [AWS outage](https://www.theverge.com/2021/12/7/22822332/amazon-server-aws-down-disney-plus-ring-outage?scrolla=5eb6d68b7fedc32c19ef33b4) may be the issue; i'll rebase the PR for you once that's resolved.

arhadthedev:
How can I help to make this PR into the main?

arhadthedev:
@ljharb Is there any blocker preventing a merge? The PR looks done and ready.

ljharb:
No blocker, editors just need to review.

michaelficarra:
We shouldn't merge this unless Windows support is going to be enforced by CI somehow. I'm also not really interested in trying to maintain Windows support going forward if it's going to make us do strange things like this. Anyone using Windows can run a free Linux distro in a VM and build ecma262 just fine.

ljharb:
Windows also has WSL2 now.

arhadthedev:
> We shouldn't merge this unless Windows support is going to be enforced by CI somehow.

I agree with the CI use case. However, the document can be built locally on any platform supported by Node.js.

I ran into the issue because I develop an implementation of an exact draft revision of ecma262 and need to somehow refer it from its in-source-file commented head. Initially I planned to include the revision as a submodule, refer to the submodule, and let maintainers locally build it for themselves.

Nonetheless, if there is a way to refer an online CI-prebuilt revision associated with a commit in `main` (instead of a PR ID that is a GitHub-only entity so not a permanent distinguisher), it would be better.

> Anyone using Windows can run a free Linux distro in a VM and build ecma262 just fine.

The program is planned for distribution so a requirement to install VirtualBox/QEMU/WSL/VMware Workstation Player (or even MSYS2) for a single dependency looks like an overkill and requires admin privileges that a user can be administratively deprived of.

> Windows also has WSL2 now.

Unfortunately, it requires Hyper-V available for Windows Pro and Enterprise Editions only. I hope that WSL1 is sufficient, however it requires separate installation like, for example, any third party VM.

package.json
ljharb: we very much want to copy here, and not make a symlink, so i don't think this is viable

arhadthedev:
I don't know whether *Files changed* survives branch removal so just in case the changes are the following:

```diff
-    "prebuild-only": "npm run clean && mkdir out && cp -R img out",
+    "prebuild-only": "npm run clean && mkdir out && node -e \"fs.symlinkSync('../img', 'out/img', 'junction')\"",
```

```diff
-    "clean": "rm -rf out",
+    "clean": "node -e \"fs.rmSync('out', { recursive: true, force: true })\"",
```

PR #2594, opened 2021-12-01
Editorial: Use `<code>` for values from other specifications

gibson042:
From original discussion at https://github.com/tc39/ecma402/pull/624#discussion_r754010696 .

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -34070,7 +34070,7 @@ THH:mm:ss.sss
           1. Let _L_ be ! CodePointsToString(_lowerText_).
           1. Return _L_.
         </emu-alg>
-        <p>The result must be derived according to the locale-insensitive case mappings in the Unicode Character Database (this explicitly includes not only the UnicodeData.txt file, but also all locale-insensitive mappings in the SpecialCasings.txt file that accompanies it).</p>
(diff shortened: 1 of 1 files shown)


ryzokuken (approved):

bakkot:
`~` is used in 262 for _values_ which get passed around and manipulated by algorithm steps, which these aren't. An explicit `<code>` tag seems like the right thing to me.

gibson042:
@bakkot is there editor consensus on that? Whether or not external-specification file names/processor instructions/etc. would count as "[Values which are internal to the [ECMA-262] specification](https://tc39.es/ecma262/#sec-value-notation)" is not obvious to me and `<code>` looks worse in my opinion, but I have no problem switching.

bakkot:
@gibson042 There is consensus among the ecma-262 editors that `~` should only be used for internal values, and that the names of files in external specifications (and similar) are not values in that sense, yes.

We didn't discuss the rendering explicitly; if you have an alternative rendering you'd prefer, feel free to suggest it.

gibson042:
That's good enough for me; updated accordingly.

michaelficarra:
@gibson042 Can you update this PR to format the referenced files with links as in https://github.com/tc39/ecma262/pull/2515#discussion_r774921012?

gibson042:
@michaelficarra done

michaelficarra (approved):

bakkot (approved):

PR #2597, opened 2021-12-10
Editorial: Use "SDO of |Foo|" form for all SDO invocations

syg:
This makes the invocation form consistent for the user-code effect propagation in #2548.


bakkot:
This misses a use in the second case under NamedEvaluation. Otherwise LGTM.

syg:
> This misses a use in the second case under NamedEvaluation. Otherwise LGTM.

Done

bakkot (approved):

michaelficarra (approved):

PR #2606, opened 2022-01-03
 Editorial: Define SortCompare as an Abstract Closure

nicolo-ribaudo:
_This PR is extracted from a PR to the "Change Array by Copy" proposal (https://github.com/tc39/proposal-change-array-by-copy/pull/69). Thanks @ljharb for pointing out that we can bring this improvement independently from that proposal._

The `SortCompare` AO implicitly receives `_comparefn_` from `Array.prototype.sort`, in a way that looks like a closure (but I don't think that it is, since `SortCompare` is an AO and not an Abstract Closure defined in the `Array.prototype.sort` steps). This PR passes `_comparefn_` explicitly as a parameter.

I also did the same thing for `TypedArraySortCompare`, that implicitly receives both `_comparefn_` and `_buffer_`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37526,7 +37526,7 @@ THH:mm:ss.sss
         </emu-note>
       </emu-clause>
 
-      <emu-clause id="sec-array.prototype.sort">
+      <emu-clause id="sec-array.prototype.sort" oldids="sec-sortcompare">
         <h1>Array.prototype.sort ( _comparefn_ )</h1>
         <p>The elements of this array are sorted. The sort must be stable (that is, elements that compare equal must remain in their original order). If _comparefn_ is not *undefined*, it should be a function that accepts two arguments _x_ and _y_ and returns a negative Number if _x_ &lt; _y_, a positive Number if _x_ &gt; _y_, or a zero otherwise.</p>
         <p>When the `sort` method is called, the following steps are taken:</p>
@@ -37544,7 +37544,7 @@ THH:mm:ss.sss
               1. Append _kValue_ to _items_.
             1. Set _k_ to _k_ + 1.
           1. Let _itemCount_ be the number of elements in _items_.
-          1. [id="step-array-sort-sortcompare"] <a name="sec-sortcompare"></a> Let _sortCompare_ be a new Abstract Closure with parameters (_x_, _y_) that captures _comparefn_ and performs the following steps when called:
+          1. [id="step-array-sort-sortcompare"] Let _sortCompare_ be a new Abstract Closure with parameters (_x_, _y_) that captures _comparefn_ and performs the following steps when called:
             1. If _x_ and _y_ are both *undefined*, return *+0*<sub>𝔽</sub>.
             1. If _x_ is *undefined*, return *1*<sub>𝔽</sub>.
             1. If _y_ is *undefined*, return *-1*<sub>𝔽</sub>.
--- a/spec.html
+++ b/spec.html
@@ -38925,7 +38925,7 @@ THH:mm:ss.sss
         <p>This function is not generic. The *this* value must be an object with a [[TypedArrayName]] internal slot.</p>
       </emu-clause>
 
-      <emu-clause id="sec-%typedarray%.prototype.sort">
+      <emu-clause id="sec-%typedarray%.prototype.sort" oldids="sec-typedarraysortcompare">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -39012,27 +39010,41 @@ <h1>%TypedArray%.prototype.sort ( _comparefn_ )</h1>
...
+              _buffer_: an ArrayBuffer or a SharedArrayBuffer,
+              _x_: a Number or BigInt,
+              _y_: a Number or BigInt,
+            )
+          </h1>
+          <dl class="header">
+            <dt>description</dt>
+            <dd>It performs a numeric comparison rather than the string comparison used in <emu-xref href="#sec-array.prototype.sort"></emu-xref>.</dd>
+          </dl>
+          <emu-alg>
```
nicolo-ribaudo, line 39028: All the next lines are just indented.

ljharb:
This helps progress on #1884.

spec.html
```diff
@@ -39012,27 +39010,41 @@ <h1>%TypedArray%.prototype.sort ( _comparefn_ )</h1>
...
-        </emu-note>
+        <p>Instead of calling SortCompare with arguments &laquo; _comparefn_, _a_, _b_ &raquo;, it calls TypedArraySortCompare with arguments &laquo; _comparefn_, _buffer_, _a_, _b_ &raquo;.</p>
+
+        <emu-clause id="sec-typedarraysortcompare" type="abstract operation">
+          <h1>
+            TypedArraySortCompare (
+              _comparefn_: *undefined* or a function object,
+              _buffer_: an ArrayBuffer or a SharedArrayBuffer,
+              _x_: a Number or BigInt,
+              _y_: a Number or BigInt,
```
michaelficarra, line 39021:
```suggestion
              _x_: a Number or a BigInt,
              _y_: a Number or a BigInt,
```

michaelficarra (approved):

bakkot:
This doesn't quite work as written because the spec talks about whether SortCompare is a consistent comparison function, which is a function of two parameters rather than 3.

The approach @jmdyck takes to resolving this in #2305 (which is blocked on, that PR has unintended [normative consequences](https://github.com/tc39/ecma262/pull/2305#discussion_r696054039)) is to make an abstract closure which captures the relevant parameters, so that we can talk about whether that abstract closure is a consistent comparison function without needing to change that definition. That seems like a good approach to take here while we work on #2305.

spec.html
```diff
@@ -37544,7 +37544,22 @@ <h1>Array.prototype.sort ( _comparefn_ )</h1>
               1. Append _kValue_ to _items_.
             1. Set _k_ to _k_ + 1.
           1. Let _itemCount_ be the number of elements in _items_.
-          1. [id="step-array-sort"] Sort _items_ using an implementation-defined sequence of calls to SortCompare. If any such call returns an abrupt completion, stop before performing any further calls to SortCompare or steps in this algorithm and return that completion.
+          1. [id="step-array-sort-sortcompare"] <a name="sec-sortcompare"></a> Let _sortCompare_ be a new Abstract Closure with parameters (_x_, _y_) that captures _comparefn_ and performs the following steps when called:
```
nicolo-ribaudo, line 37547: `<a name="sec-sortcompare"></a>` is so that existing links to the spec with `#sec-sortcompare` continue working:
- Is that a goal we have?
- Is there a better way to do it?

acutmore: I've seen other sections with an `oldids` attribute, though not sure if that would work in this case.

acutmore:
Do we think we could split the difference and still retain the bulk of the steps as a seperate AO, reducing the steps in the abstract closure? That would make it easier to share with `toSorted` in https://github.com/tc39/proposal-change-array-by-copy.

Alternatively that could be done as part of the change-array-by-copy proposal itself.

bakkot:
I'm going to close this in favor of https://github.com/tc39/ecma262/pull/2305, now that that PR has been updated to be strictly editorial. They're basically the same, except the older #2305 also refactors the definition of `TypedArray.prototype.sort` to be fully specified like other built-ins rather than with weird prose. I'm very sorry about the duplicated work. Please feel free to raise any comments there.

@acutmore: Yes, you could define a new AO (say `CompareAsStrings`) which is just the current body of the SortCompare abstract closure, and replace the current body of SortCompare with `Return ? CompareAsStrings(_x_, _y_, _comparefn_)`.

PR #2609, opened 2022-01-06
Editorial: Improve function creation text

gibson042:


bakkot (approved):

spec.html
```diff
@@ -13967,7 +13967,7 @@ <h1>
         </dl>
         <emu-alg>
           1. Let _proto_ be ? _targetFunction_.[[GetPrototypeOf]]().
-          1. Let _internalSlotsList_ be the internal slots listed in <emu-xref href="#table-internal-slots-of-bound-function-exotic-objects"></emu-xref>, plus [[Prototype]] and [[Extensible]].
+          1. Let _internalSlotsList_ be the list-concatenation of &laquo; [[Prototype]], [[Extensible]] &raquo; and the internal slots listed in <emu-xref href="#table-internal-slots-of-bound-function-exotic-objects"></emu-xref>.
```
ljharb, line 13970: the order might not matter, but this does invert the ordering (previously, [[Prototype]] and [[Extensible]] were always last)

gibson042: Agreed. I don't think order matters, but if it does then this should be an improvement because it brings BoundFunctionCreate into alignment with OrdinaryFunctionCreate (and with all other object creation AFAICT—every other use of MakeBasicObject either starts with [[Prototype]] and [[Extensible]] or does not include them at all).

michaelficarra (approved):

syg (approved):

PR #2612, opened 2022-01-08
Editorial: Remove ! from IsCallable invocation, be consistent with others.

zhangenming:
NOW, I'm just wondering why other methods(forEach/map...) don't have this.
Should it be consistent ?
THANKS...

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37074,7 +37074,7 @@ THH:mm:ss.sss
           <dl class="header">
           </dl>
           <emu-alg>
-            1. Assert: If _mapperFunction_ is present, then ! IsCallable(_mapperFunction_) is *true*, _thisArg_ is present, and _depth_ is 1.
+            1. Assert: If _mapperFunction_ is present, then IsCallable(_mapperFunction_) is *true*, _thisArg_ is present, and _depth_ is 1.
             1. Let _targetIndex_ be _start_.
             1. Let _sourceIndex_ be *+0*<sub>𝔽</sub>.
             1. Repeat, while ℝ(_sourceIndex_) &lt; _sourceLen_,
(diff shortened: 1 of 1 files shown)


jmdyck:
Hi. Normally, the opening comment in a PR should have a bit more explanation. For instance, in this case, it would be good to point out that `IsCallable` can't return an abrupt completion, so it's unnecessary to put `!` before an invocation of `IsCallable`.

Also the title "fix flatMap" is somewhat misleading, because `flatMap` isn't 'broken': although the `!` is unnecessary, it isn't causing `flatMap` to have incoherent or incorrect semantics. My suggestion for title would be "Remove ! from IsCallable invocation", as it gives a better idea of the nature of the PR's change.

Note that the spec's only other occurrence of `! IsCallable` is 34 lines up, in `FlattenIntoArray`, so you might want to include that in this PR.

zhangenming:
Thank you for your patient instruction : )

I'm just wondering why other methods(`forEach`/`map`...) don't have this
I thought it was a mistake
or Is there any other explanation
thank you

jmdyck:
> I thought the `! ` means js Logical NOT.

Ah! That's why it's good to explain your thinking. The `!` and `?` prefixes are defined in [5.2.3.4 ReturnIfAbrupt Shorthands](https://tc39.es/ecma262/#sec-returnifabrupt-shorthands).

claudepache:
About `!`: see also #568

ljharb:
Currently, every AO call whose return value isn't an introspected Completion Record has a `?` or a `!`. However, the editors have avoided adding them to predicates (like `IsCallable`), and #1796 will alter the conventions around Completion Records.

There are no plans I'm aware of to change the `!` and `?` notation.

Issue #2502, opened 2021-08-26
ToLength suffers from off-by-one errors


gibson042:
(found when working through semantics for #2501)

[ToLength](https://tc39.es/ecma262/#sec-tolength) is described as returning a value "suitable for use as the length of an array-like object", but its algorithm clamps that to 2<sup>53</sup> - 1, which is the maximum possible value of an [_integer index_](https://tc39.es/ecma262/#integer-index) and therefore one **less** than the maximum possible length of an array-like object. And some uses of it correspond with the description ([LengthOfArrayLike](https://tc39.es/ecma262/#sec-lengthofarraylike) and [StringPad](https://tc39.es/ecma262/#sec-stringpad)), while others correspond with the algorithm ([ToIndex](https://tc39.es/ecma262/#sec-toindex) and RegExp retrieval of `lastIndex`).

ToIndex itself seems to have a similar problem, with [_TypedArray_](https://tc39.es/ecma262/#sec-typedarray) and [ArrayBuffer](https://tc39.es/ecma262/#sec-arraybuffer-length) and [SharedArrayBuffer](https://tc39.es/ecma262/#sec-sharedarraybuffer-length) and [InitializeTypedArrayFromArrayBuffer](https://tc39.es/ecma262/#sec-initializetypedarrayfromarraybuffer) using its output as a _length_.

I believe this concern does not yet manifest in practice due to the extremely large values in question, but it should still be addressed.

liannemarie:
This is a hacked acct. Please send me all details to liannemadrigal1@gmail.com

claudepache:
Be careful that, in several cases, it is perfectly reasonable to have an “index” that is equal to the “length” of the corresponding string/array. This is the case for RegExp#lastIndex (where it means that the end of the string has been attained), or of Array#slice. (In fact, in those two examples, the “index” is better viewed as a position _between_ two consecutive items, where 0 means “just before the first item” and _length_ means “just after the last item”.)

gibson042:
[...]

PR #2613, opened 2022-01-08
Editorial: Define both integer index and array index using inclusive intervals

gibson042:
I think #2502 stems in part from inconsistency in the definition of _integer index_ with an inclusive upper bound of 2\*\*53 - 1 vs. _array index_ with an **exclusive** upper bound of 2\*\*32 - 1, obscuring the fact that only the latter excludes the integer immediately preceding the relevant power of two (because array _length_ rather than maximum index is constrained to uint32). This is an attempt to clear that up by defining both terms as analogous inclusive intervals (with clearly different upper bounds), along with some miscellaneous related improvements in separate commits.

Alternatively, it might make sense to use analogous half-open intervals, e.g. **+0<sub>𝔽</sub> ≤ _i_ < 𝔽(2<sup>53</sup>)** vs. **+0<sub>𝔽</sub> ≤ _a_ < 𝔽(2<sup>32</sup> - 1)**.


spec.html
```diff
@@ -2676,8 +2676,10 @@ <h1>The <dfn variants="is an Object,is not an Object">Object Type</dfn></h1>
...
+      <p>Properties are identified using key values. A <dfn variants="property keys">property key</dfn> value is either an ECMAScript String value or a Symbol value. All String and Symbol values, including the empty String, are valid as property keys. A <dfn id="property-name">property name</dfn> is a property key that is a String value. An <dfn id="integer-index" variants="integer indices">integer index</dfn> is a property name _n_ that is a canonical numeric string other than *"-0"* such that CanonicalNumericIndexString(_n_) returns a non-negative integral Number in the inclusive interval from *+0*<sub>𝔽</sub> to 𝔽(2<sup>53</sup> - 1). An <dfn id="array-index" variants="array indices">array index</dfn> is an integer index whose numeric value is in the inclusive interval from *+0*<sub>𝔽</sub> to 𝔽(2<sup>32</sup> - 2).</p>
```
michaelficarra, line 2679: I've reworded this using a more modern style where we omit words that do not add anything.

```suggestion
      <p>The properties of an object are uniquely identified using property keys. A <dfn variants="property keys">property key</dfn> is either a String or a Symbol. All Strings and Symbols, including the empty String, are valid as property keys. A <dfn id="property-name">property name</dfn> is a property key that is a String. An <dfn id="integer-index" variants="integer indices">integer index</dfn> is a property name _n_ such that CanonicalNumericIndexString(_n_) returns an integral Number in the inclusive interval from *+0*<sub>𝔽</sub> to 𝔽(2<sup>53</sup> - 1). An <dfn id="array-index" variants="array indices">array index</dfn> is an integer index _n_ such that CanonicalNumericIndexString(_n_) returns an integral Number in the inclusive interval from *+0*<sub>𝔽</sub> to 𝔽(2<sup>32</sup> - 2).</p>
```

gibson042: @michaelficarra Thanks, that's much better. And you want it to be part of this PR?

michaelficarra: Yep.

gibson042: Done.

michaelficarra (approved):

spec.html
```diff
@@ -2676,8 +2676,10 @@ <h1>The <dfn variants="is an Object,is not an Object">Object Type</dfn></h1>
...
+      <p>The properties of an object are uniquely identified using property keys. A <dfn variants="property keys">property key</dfn> is either a String or a Symbol. All Strings and Symbols, including the empty String, are valid as property keys. A <dfn id="property-name">property name</dfn> is a property key that is a String. An <dfn id="integer-index" variants="integer indices">integer index</dfn> is a property name _n_ such that CanonicalNumericIndexString(_n_) returns an integral Number in the inclusive interval from *+0*<sub>𝔽</sub> to 𝔽(2<sup>53</sup> - 1). An <dfn id="array-index" variants="array indices">array index</dfn> is an integer index _n_ such that CanonicalNumericIndexString(_n_) returns an integral Number in the inclusive interval from *+0*<sub>𝔽</sub> to 𝔽(2<sup>32</sup> - 2).</p>
```
syg, line 2679: Why combine 2 paragraphs into a single paragraph? I have slightly preference for keeping the 2 paragraphs.

syg (approved):
lgtm % question

spec.html
gibson042: Basically because there seems to be one topic: _property key_ and its subcategories. A paragraph boundary between defining _property name_ as a string-valued property key and _integer index_ as a property name canonically expressing a number in [+0, 2**53 - 1] strikes me as just a bit too artificial:

> The properties of an object are uniquely identified using property keys. A _property key_ is either a String or a Symbol. All Strings and Symbols, including the empty String, are valid as property keys. A _property name_ is a property key that is a String.
> 
> An _integer index_ is a property name <tt>n</tt> such that CanonicalNumericIndexString(<tt>n</tt>) returns an integral Number in the inclusive interval from **+0**<sub>𝔽</sub> to 𝔽(2<sup>53</sup> - 1). An _array index_ is an integer index <tt>n</tt> such that CanonicalNumericIndexString(<tt>n</tt>) returns an integral Number in the inclusive interval from **+0**<sub>𝔽</sub> to 𝔽(2<sup>32</sup> - 2).

But I don't feel too strongly about it.

syg: I'd prefer to keep the 2 paragraphs then.

gibson042: Done.

syg (approved):
Still lgtm, thanks!

PR #2619, opened 2022-01-12
Markup: add `sdo` types to `emu-clause` in missing parts

p51lee:
I believe that all syntax-directed operations (SDOs) should have `type` attribute with `"sdo"` in their `emu-clause` HTML elements. So, I searched all the missing parts and added `type` attributes with `"sdo"` to them.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -17862,7 +17862,7 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-identifiers-runtime-semantics-evaluation">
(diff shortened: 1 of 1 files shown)


michaelficarra:
These are all just the Evaluation SDO, right?

bakkot:
I did originally [intentionally leave these out](https://github.com/tc39/ecma262/pull/2271#issuecomment-759929848) - Evaluation is different from almost all other SDOs in that it is defined throughout the spec rather than in a single place - but looking again I guess there's no reason to do so.

bakkot (approved):

michaelficarra (approved):

PR #2620, opened 2022-01-12
Editorial: consistently test whether a field is present

ljharb:


spec.html
```diff
@@ -14226,7 +14226,7 @@ <h1>
           1. Let _isMapped_ be HasOwnProperty(_map_, _P_).
           1. Let _newArgDesc_ be _Desc_.
           1. If _isMapped_ is *true* and IsDataDescriptor(_Desc_) is *true*, then
-            1. If _Desc_.[[Value]] is not present and _Desc_.[[Writable]] is present and its value is *false*, then
+            1. If _Desc_ does not have a [[Value]] field, and _Desc_ has a [[Writable]] field, and _Desc_.[[Writable]] value is *false*, then
```
michaelficarra, line 14229:
```suggestion
            1. If _Desc_ does not have a [[Value]] field, and _Desc_ has a [[Writable]] field, and _Desc_.[[Writable]] is *false*, then
```

michaelficarra: Maybe we also don't need all these commas, I'm not sure.

michaelficarra (approved):
LGTM modulo comments.

spec.html
ljharb: whoops, thanks

syg (approved):
lgtm, thanks!

spec.html
jmdyck: When we want the conjunction of 3 or more conditions, the syntax is about evenly split between
`If A and B and C, ...`
vs
`If A, B, and C, ...`

So for this line, you can either delete two commas to get the first form, or delete the first "and" to get the second form. (In this case, I think the first form is slightly more readable.)

PR #2622, opened 2022-01-13
Editorial: Use consistent phrasing for parameters that are Number or BigInt

gibson042:
Before this change, NumericToRawBytes is the odd one out (it uses "a BigInt or a Number" while SameValueNonNumeric and SetValueInBuffer and GetModifySetValueInBuffer all use "a Number or a BigInt").

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -40845,7 +40845,7 @@ THH:mm:ss.sss
         <h1>
           NumericToRawBytes (
             _type_: a TypedArray element type,
-            _value_: a BigInt or a Number,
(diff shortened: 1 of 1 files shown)


bakkot (approved):

michaelficarra:
I introduce a few of these in https://github.com/tc39/ecma262/pull/2547. Can we hold off on merging this until we have an ecmarkup lint rule enforcing it so I don't just re-introduce these?

syg (approved):

PR #2647, opened 2022-02-02
Editorial: Suppress user-code on CloneArrayBuffer in TA.p.set

syg:


bakkot (approved):
Let's see: `CloneArrayBuffer` only invokes user code via `AllocateArrayBuffer`, specifically by triggering getters (or `[[Get]]` proxy traps) for the `.prototype` property of the fourth parameter to `CloneArrayBuffer`. Here that property is the intrinsic `%ArrayBuffer%`, whose `.prototype` is a non-configurable data property.

So, agreed.

ljharb:
Can’t any operation always fail to have enough memory? Or is this a special case because of typed arrays having an explicit check for that?

bakkot:
This is a special case because there is an explicit check (in [CreateByteDataBlock](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-createbytedatablock)) for having enough memory to allocate a new ArrayBuffer.

PR #2648, opened 2022-02-02
Editorial: NormalCompletion/ThrowCompletion create a Completion Record

ljharb:
Currently, "return Completion" links to the Completion() AO, which isn't what's happening here.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4037,7 +4037,7 @@
         <dl class="header">
         </dl>
         <emu-alg>
-          1. Return Completion { [[Type]]: ~normal~, [[Value]]: _value_, [[Target]]: ~empty~ }.
+          1. Return Completion Record { [[Type]]: ~normal~, [[Value]]: _value_, [[Target]]: ~empty~ }.
         </emu-alg>
       </emu-clause>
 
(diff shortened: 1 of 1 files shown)


michaelficarra (approved):

ljharb:
(updated to fix the remaining instances)

bakkot (approved):

syg (approved):

jmdyck:
> Currently, "return Completion" links to the Completion() AO, [...]

Hm. I've got a copy of the Jan 20 version of the rendered spec in my browser, and in occurrences of `Return Completion {...}`, the "Completion" links to [6.2.3 The Completion Record Specification Type](https://tc39.es/ecma262/#sec-completion-record-specification-type), not to the `Completion` AO in [5.2.3.1 Implicit Completion Values](https://tc39.es/ecma262/#_ref_1111). (In fact, nothing links to the `Completion` AO, not even `Completion(...)` calls!)

Looking at [the diff of the rendered specs](https://github.com/tc39/ecma262/commit/44a1f6855d972858f099b941adb8868def364304), there wasn't a change to the `<a>` element surrounding "Completion [Record]", there was only a change in the generated `<emu-xref>` element. Presumably browsers pay attention to the `<a>` element, not the `<emu-xref>` element.

I'm fine with the changes of this PR, it's just the quoted sentence that seems off.

ljharb:
hmm, when i clicked in the deploy preview on "Completion" i could have sworn it took me to the AO, which is how I noticed the change. Either way, this change seems like an improvement :-)

PR #2671, opened 2022-02-23
Editorial: Omit coma in a related description list

SebastianMaciel:
A simple edit removing a comma where normally is not necessary.


arhadthedev:
The specification uses [Oxford commas](https://en.wikipedia.org/wiki/Serial_comma) so its placement here is intended.

ljharb:
Oxford commas are very much necessary :-)

SebastianMaciel:
I understand, it's ok to use that convention here. Should I close this PR then? Thanks!

PR #2672, opened 2022-02-23
Editorial: `Object.getOwnProperty{Names,Symbols}`: refactor where the List is converted

ljharb:
Extracted this from #2547.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -29461,7 +29461,7 @@
         <h1>Object.getOwnPropertyNames ( _O_ )</h1>
         <p>When the `getOwnPropertyNames` function is called, the following steps are taken:</p>
         <emu-alg>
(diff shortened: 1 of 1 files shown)


ljharb:
ahhh sorry, didn't read enough of the diff first. in main, the AO returns an array; you changed it in #2547 so it returns a List.

There's no bug here, but perhaps it'd be nice to do the refactor sooner?

ljharb:
I've updated it to be editorial, but now maybe it's not worth landing separately; feel free to close if you prefer.

michaelficarra:
I don't think it's as motivated before #2547. We could drop it from there and have this go in after, but I think it's fine for it to just go in with that PR.

ljharb:
Sure, totally fair.

Issue #2321, opened 2021-02-21
When generating the spec, we should output an 'emca262biblio.json' for tooling consumption.


rbuckton:
The current spec build script does not output the latest version of `ecma262biblio.json`, which would be useful for tooling consumption:

https://github.com/tc39/ecma262/blob/c328b0aded262172094e1d9a44be216fc67b0632/package.json#L13

Suggested Change:

```diff
--- "build-only": "ecmarkup --no-ecma-262-biblio --verbose spec.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js", 
+++ "build-only": "ecmarkup --no-ecma-262-biblio --verbose spec.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js --biblio out/ecma262biblio.json", 
```

This would allow tools to fetch `https://tc39.es/ecma262/ecma262biblio.json` to ensure links in proposals are generated against the latest version of the spec.

rbuckton:
Or, as is mentioned in tc39/ecmarkup#251, an external process that updates and publishes ecma262biblio.json so that other tools can use it as well.

bakkot:
Yeah, we should address this at some point. I prefer publishing it as an npm package so that proposals can pin the version they were using at the time, for reproducible builds.

rbuckton:
The problem with that is the same issue I just ran into prompting me to file the issue, pinning a version doesn't work for link references unless you have a published *pinned* version of that spec to refer to as well. I just ran into this with the change to rename `FunctionCreate`. If I was using a pinned version of the biblio, I still wouldn't be up to date with the spec.

In my opinion, if we were to pin anything it would only be for published editions of the spec according to the ECMA. We'd always want to also be able to reference the *live* version of the spec so that links work correctly. The only use I might have in having a version history of ecma262biblio.json files on NPM would be to find *when* a change occurred, i.e., reporting an error like "You previously referenced `FunctionCreate`, but that doesn't exist anymore. Maybe you meant `OrdinaryFunctionCreate`?".

We don't explicitly call out algorithm names syntactically in the spec, we just use string matching. As a result, a rename like `FunctionCreate` can't easily be caught (since it goes from being a valid link to being just PascalCased text without warning).

bakkot:
[...]

PR #2675, opened 2022-02-25
Meta: publish biblio to npm

bakkot:
Fixes https://github.com/tc39/ecmarkup/issues/251. Fixes https://github.com/tc39/ecma262/issues/2321.

I haven't tested this, but it will probably work.

~Also, we need to pick a name for it. `ecma262-biblio` seems like an obvious choice. We could choose to scope it under the `@tc39` namespace, but to my understanding we can have it be controlled by that organization even if we place it at the top-level scope.~ Edit: went with `@tc39/ecma262-biblio`.

As for version numbers, ~right now I'm just using the number of seconds since 1970 as the patch version~ edit: went with number of commits on the branch, with an explicit callout in the readme that the package does not follow semver (since the biblio can change at any time, which can break your build if you depend on it for ecmarkup and build using `--strict`).

### DIFF
--- a/biblio/README.md
+++ b/biblio/README.md
@@ -1,0 +1,9 @@
+# ECMA-262 Bibliography
+
+This package, available on npm as `@tc39/ecma262-biblio`, contains a machine-readable representation of the terms, clauses, grammar, and abstract operations defined in ECMA-262. This will primarily be of use to people working with the specification itself.
+
+If added as a dependency to a project using ecmarkup, you can load it by passing `--load-biblio @tc39/ecma262-biblio`.
+
+It is automatically updated whenever ECMA-262 is. It is inherently unstable: editorial changes to the specification may add, remove, or modify the biblio, which may break your build (for example, if using ecmarkup with `--lint-spec --strict`). As such, **the usual semver guarantees do not hold**. You should pin a precise version of this package.
+
+Major version bumps may be used for breaking changes to the format of the biblio itself. Minor version bumps may be used for non-breaking additions to the biblio format.
--- a/biblio/package.json
+++ b/biblio/package.json
@@ -1,0 +1,18 @@
+{
+  "name": "@tc39/ecma262-biblio",
+  "version": "VERSIONED-DURING-PUBLISH",
+  "description": "Machine-readable representation of the internals of the ecma-262 spec",
+  "keywords": [
+    "ecmascript"
+  ],
+  "author": "TC39",
+  "main": "biblio.json",
+  "exports": {
+    ".": "./biblio.json",
+    "./package.json": "./package.json"
+  },
+  "files": [
+    "biblio.json"
+  ],
+  "license": "SEE LICENSE IN LICENSE.md"
+}
--- a/scripts/publish-biblio.sh
+++ b/scripts/publish-biblio.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+
+$(npm bin)/ecmarkup --verbose spec.html --write-biblio biblio/biblio.json /dev/null
+
+cp LICENSE.md biblio/
+
+cd biblio
+
+COMMIT_COUNT=$(git rev-list --count HEAD)
+npm version --no-git-tag-version "1.0.${COMMIT_COUNT}"
+
+SHORT_COMMIT=$(git rev-parse --short HEAD)
+LONG_COMMIT=$(git rev-parse --verify HEAD)
+echo "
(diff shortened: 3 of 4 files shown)


biblio/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "TBD",
+  "version": "VERSIONED-DURING-PUBLISH",
```
ljharb, line 3: maybe it should have `"private": true`, and have the publish step remove it? to prevent accidents

scripts/publish-biblio.sh
```diff
@@ -0,0 +1,20 @@
...
+
+TS=$(date +%s)
+npm version --no-git-tag-version "1.0.${TS}"
+
+SHORT_COMMIT=$(git rev-parse --short HEAD)
+LONG_COMMIT=$(git rev-parse --verify HEAD)
+echo "
+This version was built from commit [${SHORT_COMMIT}](https://github.com/tc39/ecma262/tree/${LONG_COMMIT})." >> README.md
+
+npm publish
```
ljharb, line 20: one thought is that this won't be idempotent - it'll either fail on a rerun, or needlessly publish a new one that might not actually be later - so rerunning an old one would cause lots of havoc.

it seems better to find some way for it to be deterministic but also monotonically increasing.

ljharb (commented):
It seems nicer to scope it regardless

biblio/package.json
bakkot: Will npm actually let you publish with a version string of `VERSIONED-DURING-PUBLISH`?

ljharb: ha, no, true enough, it wouldn't.

actually an invalid version string might break npm in a number of ways; this should probably be `0.0.0`

scripts/publish-biblio.sh
bakkot: Oh, good point. Switched to the timestamp of the commit itself.

biblio/package.json
bakkot: This version string is good enough to run `npm version`, which is all it needs to do.

scripts/publish-biblio.sh
ljharb: Normally that wouldn't be any better, but since the policy in this repo is to use `git commit --amend --date=now` before landing everything, we do in fact have this guarantee.

bakkot: Squash-and-merge and rebase-and-merge give the same guarantee, I'm pretty sure; you have to go pretty far out of your way to end up with non-monotonic commit times.

ljharb: True as well.

biblio/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "description": "Machine-readable representation of the internals of the ecma-262 spec",
+  "keywords": [
+    "ecmascript"
+  ],
+  "author": "TC39",
+  "main": "biblio.json",
```
ljharb, line 9:
```suggestion
  "main": "biblio.json",
  "exports": {
    ".": "./biblio.json",
    "./package.json": "./package.json"
  },
```
since it's a potential breaking change to add later :-)

scripts/publish-biblio.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+
+$(npm bin)/ecmarkup --verbose spec.html --write-biblio biblio/biblio.json /dev/null
+
+cp LICENSE.md biblio
```
syg, line 8:
```suggestion
cp LICENSE.md biblio/
```

biblio/README.md
```diff
@@ -0,0 +1,9 @@
+# ECMA-262 Bibliography
+
+This package, available on npm as NAME TBD, contains a machine-readable representation of the terms, clauses, grammar, and abstract operations defined in ECMA-262. This will primarily be of use to people working with the specification itself.
```
syg, line 3: Don't forget to fix all the NAME TBD before landing.

syg (approved):

ljharb:
I added https://docs.npmjs.com/cli/v8/configuring-npm/package-json#publishconfig to the biblio's package.json, to ensure the first publish, at least, works (scoped packages require the initial publish to specify "public")

bakkot:
Is the `npm publish --access public` not sufficient for that?

ljharb:
It is, but the script doesn't have that. Would you prefer I put that in the script versus using the publishConfig?

bakkot:
> It is, but the script doesn't have that.

Pretty sure it does?

ljharb:
oh, lol, nice work :-) i guess you added it when adding the scope, and i missed it. i'll remove the publishConfig and get this landed.

PR #2678, opened 2022-03-02
Editorial: Move class and method related abstract operations to more relevant section

pzuraq:
Moves the following abstract operations to a more relevant section:

- MakeConstructor
- MakeClassConstructor
- MakeMethod
- DefineMethodProperty

These abstract operations are used primarily in the *ECMAScript
Language: Functions and Classes* section, despite being defined in
*10.2 ECMAScript Function Objects*. I found myself jumping back and
forth frequently while working on the Decorators branch, and it seemed
like it would make more sense for them to be located near to where they
are used.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13518,89 +13518,6 @@
       </emu-clause>
     </emu-clause>
 
-    <emu-clause id="sec-makeconstructor" type="abstract operation">
-      <h1>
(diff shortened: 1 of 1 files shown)


bakkot:
Editors discussed this briefly and were only really in favor of moving DefineMethodProperty.

PR #2679, opened 2022-03-03
Meta: fix biblio publish job

bakkot:
bah humbug


ljharb:
My alternative is https://github.com/ljharb/ecma262/commit/db8db4c78fff278d1c2ebaea8b048b21c7cf7558, which uses the same actions setup the rest of the jobs are using.

PR #2687, opened 2022-03-10
Editorial: fix ambiguity in Annex B definition of typeof semantics

### LINKED ISSUE #2685: editorial issue causing ambiguity in typeof support for [[IsHTMLDDA]]
Section [B.3.6.3 Changes to the typeof Operator](https://tc39.es/ecma262/#sec-IsHTMLDDA-internal-slot-typeof) adds an entry to [Table 41](https://tc39.es/ecma262/#table-typeof-operator-results) which causes the left column to be ambiguous for some objects. Every language value should fall into exactly one row of that table. Instead of modifying the table to add the row, we should modify the algorithm, adding a step that early exits for objects with an `[[IsHTMLDDA]]` slot.

michaelficarra:
Fixes #2685.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -19702,6 +19702,7 @@
           1. If _val_ is a Reference Record, then
             1. If IsUnresolvableReference(_val_) is *true*, return *"undefined"*.
           1. Set _val_ to ? GetValue(_val_).
+          1. [id="step-typeof-web-compat-insertion-point"] NOTE: This step is replaced in section <emu-xref href="#sec-IsHTMLDDA-internal-slot-typeof"></emu-xref>.
           1. Return a String according to <emu-xref href="#table-typeof-operator-results"></emu-xref>.
         </emu-alg>
         <emu-table id="table-typeof-operator-results" caption="typeof Operator Results" oldids="table-35">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -47640,30 +47641,10 @@ <h1>Changes to IsLooselyEqual</h1>
...
-              <td>
-                Object (has an [[IsHTMLDDA]] internal slot)
-              </td>
-              <td>
-                *"undefined"*
-              </td>
-            </tr>
-          </table>
-        </emu-table>
+        <p>The following steps replace step <emu-xref href="#step-typeof-web-compat-insertion-point"></emu-xref> of <emu-xref href="#sec-typeof-operator-runtime-semantics-evaluation">the evaluation semantics for `typeof`</emu-xref>:</p>
```
bakkot, line 47644:
```suggestion
        <p>The following step replaces step <emu-xref href="#step-typeof-web-compat-insertion-point"></emu-xref> of <emu-xref href="#sec-typeof-operator-runtime-semantics-evaluation">the evaluation semantics for `typeof`</emu-xref>:</p>
```

michaelficarra: We use the plural form when referring to single-step algorithms all the time: https://tc39.es/ecma262/#sec-date.prototype.valueof

I prefer the consistency of phrasing. It's 1 or more steps.

syg: > I prefer the consistency of phrasing. It's 1 or more steps.

I think in that case it sounds fine because of the "or more". In this case though, I prefer @bakkot's suggestion.

syg (approved):

PR #2688, opened 2022-03-10
Editorial: work around ecmarkup limititation with replacement step numbering

michaelficarra:
Ecmarkup currently doesn't number steps nicely when replacing one step with multiple. See https://github.com/tc39/ecmarkup/issues/418. This uses the workaround described in that issue. Additionally, I changed the phrasing in [B.3.6.2](https://tc39.es/ecma262/#sec-IsHTMLDDA-internal-slot-aec) to match the phrasing used to replace steps elsewhere in Annex B.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -47541,25 +47541,26 @@ THH:mm:ss.sss
         <h1>Changes to GlobalDeclarationInstantiation</h1>
         <p>During GlobalDeclarationInstantiation the following steps are performed in place of step <emu-xref href="#step-globaldeclarationinstantiation-web-compat-insertion-point"></emu-xref>:</p>
         <emu-alg replaces-step="step-globaldeclarationinstantiation-web-compat-insertion-point">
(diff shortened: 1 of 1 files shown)


bakkot (approved):

spec.html
```diff
@@ -47499,11 +47500,12 @@ <h1>Changes to BlockDeclarationInstantiation</h1>
         </emu-alg>
         <p>During BlockDeclarationInstantiation the following steps are performed in place of step <emu-xref href="#step-blockdeclarationinstantiation-initializebinding"></emu-xref>:</p>
```
michaelficarra, line 47501: We could probably generate this preceding paragraph for any `emu-alg[replace-step]` nodes to ensure consistency and remove duplication in the source doc.

bakkot: Eh, I don't think it comes up enough to be worth it.

syg (approved):

PR #2693, opened 2022-03-16
Editorial: Add ToZeroPaddedDecimalString abstract operation.

Ms2ger:
This allows the call sites to be much less verbose.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33851,7 +33851,7 @@ THH:mm:ss.sss
           <h1>
             ToZeroPaddedDecimalString (
               _n_: an integer,
-              _minLength_: an integer,
+              _minLength_: a non-negative integer,
             ): a String
           </h1>
           <dl class="header">
--- a/spec.html
+++ b/spec.html
@@ -33858,7 +33858,7 @@ THH:mm:ss.sss
           </dl>
           <emu-alg>
             1. Let _S_ be the String representation of _n_, formatted as a decimal number.
-            1. Return ! StringPad(_year_, 𝔽(_minLength_), *"0"*, ~start~).
+            1. Return ! StringPad(_S_, 𝔽(_minLength_), *"0"*, ~start~).
           </emu-alg>
         </emu-clause>
       </emu-clause>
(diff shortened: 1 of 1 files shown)


ljharb (commented):
I like this!

ljharb:
@Ms2ger unfortunately, this PR is from igalia's fork instead of your own, and github has a bug where repo owners can't force push to PRs in that case. can you add me to igalia's fork, at least until this PR is closed or merged?

michaelficarra (approved):

spec.html
```diff
@@ -33848,6 +33846,21 @@ <h1>
...
+          <h1>
+            ToZeroPaddedDecimalString (
+              _n_: an integer,
+              _minLength_: an integer,
+            ): a String
+          </h1>
+          <dl class="header">
+          </dl>
+          <emu-alg>
+            1. Let _S_ be the String representation of _n_, formatted as a decimal number.
```
bakkot, line 33860:
```suggestion
            1. Let _year_ be the String representation of _n_, formatted as a decimal number.
```

spec.html
```diff
@@ -33848,6 +33846,21 @@ <h1>
...
+            1. Return ! StringPad(_year_, 𝔽(_minLength_), *"0"*, ~start~).
```
ljharb, line 33861: or perhaps
```suggestion
            1. Return ! StringPad(_S_, 𝔽(_minLength_), *"0"*, ~start~).
```

spec.html
michaelficarra: Let's do the other way around, please.

spec.html
```diff
@@ -33848,6 +33846,21 @@ <h1>
             <p>The argument _fillString_ defaults to *" "* (the String value consisting of the code unit 0x0020 SPACE).</p>
           </emu-note>
         </emu-clause>
+
+        <emu-clause id="sec-tozeropaddeddecimalstring" type="abstract operation">
+          <h1>
+            ToZeroPaddedDecimalString (
+              _n_: an integer,
+              _minLength_: an integer,
```
bakkot, line 33854:
```suggestion
              _minLength_: a non-negative integer,
```

Ms2ger:
Moved to #2699

ljharb:
It would have been preferable to avoid creating  new permanent PR ref, but i guess that unblocks it too.

PR #2697, opened 2022-03-16
Editorial: `StringPad`: refactor to be completionless

ljharb:
The observable order of exceptions should be maintained, and now StringPad doesn't need to return a completion record.

Whichever lands first, this or #2693 (probably that one), I'll be happy to update the other.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33101,7 +33101,7 @@ THH:mm:ss.sss
             1. Let _yv_ be YearFromTime(_tv_).
             1. If _yv_ &ge; *+0*<sub>𝔽</sub>, let _yearSign_ be the empty String; otherwise, let _yearSign_ be *"-"*.
             1. Let _year_ be the String representation of abs(ℝ(_yv_)), formatted as a decimal number.
-            1. Let _paddedYear_ be ! StringPad(_year_, *4*<sub>𝔽</sub>, *"0"*, ~start~).
+            1. Let _paddedYear_ be StringPad(_year_, 4, *"0"*, ~start~).
             1. Return the string-concatenation of _weekday_, the code unit 0x0020 (SPACE), _month_, the code unit 0x0020 (SPACE), _day_, the code unit 0x0020 (SPACE), _yearSign_, and _paddedYear_.
           </emu-alg>
           <emu-table id="sec-todatestring-day-names" caption="Names of days of the week">
@@ -33346,7 +33346,7 @@ THH:mm:ss.sss
           1. Let _yv_ be YearFromTime(_tv_).
           1. If _yv_ &ge; *+0*<sub>𝔽</sub>, let _yearSign_ be the empty String; otherwise, let _yearSign_ be *"-"*.
           1. Let _year_ be the String representation of abs(ℝ(_yv_)), formatted as a decimal number.
-          1. Let _paddedYear_ be ! StringPad(_year_, *4*<sub>𝔽</sub>, *"0"*, ~start~).
+          1. Let _paddedYear_ be StringPad(_year_, 4, *"0"*, ~start~).
           1. Return the string-concatenation of _weekday_, *","*, the code unit 0x0020 (SPACE), _day_, the code unit 0x0020 (SPACE), _month_, the code unit 0x0020 (SPACE), _yearSign_, _paddedYear_, the code unit 0x0020 (SPACE), and TimeString(_tv_).
         </emu-alg>
       </emu-clause>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -33815,29 +33820,32 @@ <h1>String.prototype.padStart ( _maxLength_ [ , _fillString_ ] )</h1>
...
           </dl>
           <emu-alg>
-            1. Let _S_ be ? ToString(_O_).
-            1. Let _intMaxLength_ be ℝ(? ToLength(_maxLength_)).
             1. Let _stringLength_ be the length of _S_.
-            1. If _intMaxLength_ &le; _stringLength_, return _S_.
-            1. If _fillString_ is *undefined*, let _filler_ be the String value consisting solely of the code unit 0x0020 (SPACE).
-            1. Else, let _filler_ be ? ToString(_fillString_).
+            1. If _maxLength_ &le; _stringLength_, return _S_.
+            1. If _fillString_ is ~empty~, let _filler_ be the String value consisting solely of the code unit 0x0020 (SPACE).
```
bakkot, line 33845: I'd inline this step into `padStart`/`padEnd` as well, I think.

spec.html
```diff
@@ -33815,29 +33820,32 @@ <h1>String.prototype.padStart ( _maxLength_ [ , _fillString_ ] )</h1>
         <p>When the `padStart` method is called, the following steps are taken:</p>
         <emu-alg>
           1. Let _O_ be ? RequireObjectCoercible(*this* value).
-          1. Return ? StringPad(_O_, _maxLength_, _fillString_, ~start~).
+          1. Let _S_ be ? ToString(_O_).
+          1. Let _maxLen_ be ℝ(? ToLength(_maxLength_)).
+          1. Let _fillStr_ be ? ToString(_fillString_).
+          1. If _fillStr_ is *undefined*, let _filler_ be ~empty~.
```
bakkot, line 33826: Gotta do the `undefined` check before the `ToString` call.

spec.html
```diff
@@ -33815,29 +33820,32 @@ <h1>String.prototype.padStart ( _maxLength_ [ , _fillString_ ] )</h1>
...
         </emu-alg>
 
         <emu-clause id="sec-stringpad" type="abstract operation">
           <h1>
             StringPad (
-              _O_: an ECMAScript language value,
-              _maxLength_: an ECMAScript language value,
-              _fillString_: an ECMAScript language value,
+              _S_: a String
+              _maxLength_: a non-negative integer that is less than 2 ** 53
```
bakkot, line 33835:
```suggestion
              _maxLength_: a non-negative integer that is less than 2<sup>53</sup>
```

spec.html
ljharb: tbh it feelsbadman to have all these steps repeated in both padEnd and padStart. what about another non-pure AO that they can both share, that calls into the pure StringPad?

spec.html
```diff
@@ -33814,29 +33819,30 @@ <h1>String.prototype.padStart ( _maxLength_ [ , _fillString_ ] )</h1>
         <p>When the `padStart` method is called, the following steps are taken:</p>
         <emu-alg>
           1. Let _O_ be ? RequireObjectCoercible(*this* value).
-          1. Return ? StringPad(_O_, _maxLength_, _fillString_, ~start~).
+          1. Let _S_ be ? ToString(_O_).
+          1. Let _maxLen_ be ℝ(? ToLength(_maxLength_)).
+          1. If _fillString_ is *undefined*, let _filler_ be ~empty~.
+          1. Else, let _filler_ be ? ToString(_fillString_).
+          1. If _filler_ is ~empty~, let _filler_ be the String value consisting solely of the code unit 0x0020 (SPACE).
```
bakkot, line 33826: No need for this to be a separate step; you can just do `If _fillString_ is *undefined*, let _filler_ be the String value consisting solely of the code unit 0x0020 (SPACE)` above.

ljharb: oh right, duh, thanks

spec.html
bakkot: Ehhhh I think in this case I would lean towards keeping the duplication, personally. I am a lot less inclined to factor out the "type-coerce / provide defaults for all the arguments" steps than I am most logic.

spec.html
```diff
@@ -33814,29 +33818,29 @@ <h1>String.prototype.padStart ( _maxLength_ [ , _fillString_ ] )</h1>
...
           <dl class="header">
           </dl>
           <emu-alg>
-            1. Let _S_ be ? ToString(_O_).
-            1. Let _intMaxLength_ be ℝ(? ToLength(_maxLength_)).
             1. Let _stringLength_ be the length of _S_.
-            1. If _intMaxLength_ &le; _stringLength_, return _S_.
-            1. If _fillString_ is *undefined*, let _filler_ be the String value consisting solely of the code unit 0x0020 (SPACE).
-            1. Else, let _filler_ be ? ToString(_fillString_).
+            1. If _maxLength_ &le; _stringLength_, return _S_.
```
bakkot, line 33841: Oh no, I just noticed this is normative, because the length check happens _before_ coercing: `'aaaa'.padStart(0, { toString(){ throw 'no touching'; } })` does not throw, but would with this change.

That makes this refactoring a lot less appealing...

ljharb: ah, makes sense

PR #2700, opened 2022-03-18
Editorial: Handled -0 case in 'Number::multiply'

h2oche:
In [Number::multiply](https://tc39.es/ecma262/#sec-numeric-types-number-multiply), it seems that it evaluates `(-0) * (+0)` to `+0` which is different to [previous specification](http://262.ecma-international.org/11.0/#sec-numeric-types-number-multiply). Currently, it fails following test262 tests according to our implementation [ESMeta](https://github.com/es-meta/esmeta):
- test/language/expressions/multiplication/S11.5.1_A4_T2.js

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1867,6 +1867,12 @@
               1. If _x_ is *+0*<sub>𝔽</sub> or _x_ is *-0*<sub>𝔽</sub>, return *NaN*.
               1. If _x_ &gt; *+0*<sub>𝔽</sub>, return _y_.
               1. Return -_y_.
+            1. If _x_ is *-0*<sub>𝔽</sub>, then
+              1. If _y_ is *-0*<sub>𝔽</sub> or _y_ &lt; *-0*<sub>𝔽</sub>, return *+0*<sub>𝔽</sub>.
+              1. Else, return *-0*<sub>𝔽</sub>.
+            1. If _y_ is *-0*<sub>𝔽</sub>, then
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -1867,6 +1867,12 @@ <h1>
               1. If _x_ is *+0*<sub>𝔽</sub> or _x_ is *-0*<sub>𝔽</sub>, return *NaN*.
               1. If _x_ &gt; *+0*<sub>𝔽</sub>, return _y_.
               1. Return -_y_.
+            1. If _x_ is *-0*<sub>𝔽</sub>, then
+              1. If _y_ &lt; *+0*<sub>𝔽</sub>, return *+0*<sub>𝔽</sub>.
```
claudepache, line 1871: I don’t know what “< +0<sub>F</sub>” means: Is -0<sub>F</sub> equal to, or less than, or both, +0<sub>F</sub>? For clarity, you should say  “_y_ is -0<sub>F</sub> or _y_ < -0<sub>F</sub>”.

h2oche: I mean the Number value with negative sign. I agree with your opinion, but if I say `y < -0_F`, ecmarkup emits an error message follow:
```
Error: spec.html:1871:59: comparisons against floating-point zero should use positive zero
```

bakkot: I'll just change ecmarkup to allow this: https://github.com/tc39/ecma262/pull/2708

I've pushed a commit adopting that here. This branch will need to be rebased on the above PR (after it lands) for it to pass CI.

michaelficarra (approved):

syg (approved):

PR #2705, opened 2022-03-22
Editorial: Improve consistency of property and record field names

gibson042:
* `"…" property` → `*"…"* property`
* ``` `…` property``` → `*"…"* property`
* `[[…]] property` → `[[…]] field`


michaelficarra (approved):
Not totally sure about the `SuperProperty` change, but the rest LGTM.

spec.html
```diff
@@ -10669,7 +10669,7 @@ <h1>Function Environment Records</h1>
                 GetSuperBase()
               </td>
               <td>
-                Return the object that is the base for `super` property accesses bound in this Environment Record. The value *undefined* indicates that `super` property accesses will produce runtime errors.
+                Return the object that is the base for |SuperProperty| accesses bound in this Environment Record. The value *undefined* indicates that such accesses will produce runtime errors.
```
bakkot, line 10672: I'm fine with the `such` change, but I do prefer the original `` `super` property`` phrasing.

spec.html
```diff
@@ -13177,7 +13177,7 @@ <h1>ECMAScript Function Objects</h1>
             an Object
           </td>
           <td>
-            If the function uses `super`, this is the object whose [[GetPrototypeOf]] provides the object where `super` property lookups begin.
+            If the function uses `super`, this is the object whose [[GetPrototypeOf]] provides the object where |SuperProperty| lookups begin.
```
bakkot, line 13180: Ditto above.

syg (approved):
lgtm % @bakkot and @michaelficarra's comment about super.

bakkot (approved):
LGTM other than the `super` things.

spec.html
gibson042: Restored, but it still reads a little awkwardly to me. What would @tc39/ecma262-editors think about adding parenthetical clarification?
```suggestion
                Return the object that is the base for `super` property accesses bound in this Environment Record (i.e., instances of |SuperProperty|). The value *undefined* indicates that evaluation of such accesses will produce runtime errors.
```

bakkot: The parenthetical doesn't really seem helpful to me personally.

PR #2711, opened 2022-03-24
Editorial: Improve consistency of referring to String values

gibson042:
This leaves a few cases of `the String value *"…"*` where it seems important to highlight that the value is a string (and most consistently in sections defining the initial value of a `@@toStringTag` property), but otherwise updates to just `*"…"*`. There is also some minor cleanup of operation descriptions in the neighborhood of those changes, for CanonicalNumericIndexString and Atomics.wait.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2248,10 +2248,10 @@
             <dd>It represents _x_ as a String using a positional numeral system with radix _radix_. The digits used in the representation of a number using radix _r_ are taken from the first _r_ code units of *"0123456789abcdefghijklmnopqrstuvwxyz"* in order. The representation of numbers with magnitude greater than or equal to *1*<sub>𝔽</sub> never includes leading zeroes.</dd>
           </dl>
           <emu-alg>
-            1. If _x_ is *NaN*, return the String *"NaN"*.
-            1. If _x_ is *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, return the String *"0"*.
+            1. If _x_ is *NaN*, return *"NaN"*.
+            1. If _x_ is *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, return *"0"*.
             1. If _x_ &lt; *-0*<sub>𝔽</sub>, return the string-concatenation of *"-"* and Number::toString(-_x_, _radix_).
-            1. If _x_ is *+&infin;*<sub>𝔽</sub>, return the String *"Infinity"*.
+            1. If _x_ is *+&infin;*<sub>𝔽</sub>, return *"Infinity"*.
             1. [id="step-number-tostring-intermediate-values"] Let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, _radix_<sup>_k_ - 1</sup> &le; _s_ &lt; _radix_<sup>_k_</sup>, 𝔽(_s_ &times; _radix_<sup>_n_ - _k_</sup>) is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the representation of _s_ using radix _radix_, that _s_ is not divisible by _radix_, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
             1. If _radix_ &ne; 10 or _n_ is in the inclusive interval from -5 to 21, then
               1. If _n_ &ge; _k_, then
@@ -2657,7 +2657,7 @@
             <dd>It represents _x_ as a String using a positional numeral system with radix _radix_. The digits used in the representation of a BigInt using radix _r_ are taken from the first _r_ code units of *"0123456789abcdefghijklmnopqrstuvwxyz"* in order. The representation of BigInts other than *0*<sub>ℤ</sub> never includes leading zeroes.</dd>
           </dl>
           <emu-alg>
-            1. If _x_ &lt; *0*<sub>ℤ</sub>, return the string-concatenation of the String *"-"* and BigInt::toString(-_x_, _radix_).
+            1. If _x_ &lt; *0*<sub>ℤ</sub>, return the string-concatenation of *"-"* and BigInt::toString(-_x_, _radix_).
             1. Return the String value consisting of the representation of _x_ using radix _radix_.
           </emu-alg>
         </emu-clause>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -5740,7 +5740,7 @@ <h1>
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It returns _argument_ converted to a Number value if it is a String representation of a Number that would be produced by ToString, or the string *"-0"*. Otherwise, it returns *undefined*.</dd>
+        <dd>If _argument_ is either *"-0"* or exactly matches the result of ToString(_n_) for corresponding Number value _n_, it returns the respective Number value. Otherwise, it returns *undefined*.</dd>
```
bakkot, line 5743:
```suggestion
        <dd>If _argument_ is either *"-0"* or exactly matches the result of ToString(_n_) for some Number value _n_, it returns the respective Number value. Otherwise, it returns *undefined*.</dd>
```

spec.html
```diff
@@ -7240,7 +7240,7 @@ <h1>Static Semantics: BoundNames ( ): a List of Strings</h1>
...
+        <p>*"\*default\*"* is used within this specification as a synthetic name for a module's default export when it does not have another name. An entry in the module's [[Environment]] is created with that name and holds the corresponding value, and resolving the export named *"default"* by calling <emu-xref href="#sec-resolveexport" title></emu-xref> for the module will return a ResolvedBinding Record whose [[BindingName]] is *"\*default\*"*, which will then resolve in the module's [[Environment]] to the above-mentioned value. This is done only for ease of specification, so that anonymous default exports can be resolved like any other export. This *"\*default\*"* value is never accessible to ECMAScript code or to the module linking algorithm.</p>
```
bakkot, line 7243:
```suggestion
        <p>*"\*default\*"* is used within this specification as a synthetic name for a module's default export when it does not have another name. An entry in the module's [[Environment]] is created with that name and holds the corresponding value, and resolving the export named *"default"* by calling <emu-xref href="#sec-resolveexport" title></emu-xref> for the module will return a ResolvedBinding Record whose [[BindingName]] is *"\*default\*"*, which will then resolve in the module's [[Environment]] to the above-mentioned value. This is done only for ease of specification, so that anonymous default exports can be resolved like any other export. This *"\*default\*"* string is never accessible to ECMAScript code or to the module linking algorithm.</p>
```
I prefer making it explicit this is referring to the string because the string is also used to bind a value in the environment, and hence `*"\*default\*"* value` could be taken to refer to the value bound by that string in the environment rather than to the string itself.

spec.html
```diff
@@ -42032,7 +42032,7 @@ <h1>Atomics.sub ( _typedArray_, _index_, _value_ )</h1>
 
     <emu-clause id="sec-atomics.wait">
       <h1>Atomics.wait ( _typedArray_, _index_, _value_, _timeout_ )</h1>
-      <p>`Atomics.wait` puts the calling agent in a wait queue and puts it to sleep until it is notified or the sleep times out. The following steps are taken:</p>
+      <p>Puts the calling agent in a wait queue and suspends it until notified or the wait times out, returning a String differentiating those cases. The following steps are taken:</p>
```
bakkot, line 42035:
```suggestion
      <p>`Atomics.wait` puts the calling agent in a wait queue and suspends it until notified or the wait times out, returning a String differentiating those cases. The following steps are taken:</p>
```
We don't normally use sentence fragments in this position, I don't think? We're inconsistent in the terminology here, but I don't think we have that particular inconsistency, at least not often.

Also I prefer "until it is notified" rather than "until notified", since it makes explicit which thing is being notified.

bakkot (approved):
LGTM other than comments.

spec.html
gibson042: Sentence fragments here are not <em>un</em>common (there are lots of examples like `<p>Returns …` and `<p>Performs …` and `<p>Produces …`, cf. [Number.prototype.toLocaleString](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-number.prototype.tolocalestring)/[BigInt.prototype.toLocaleString](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-bigint.prototype.tolocalestring) and [Math](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-function-properties-of-the-math-object) and [String.prototype.matchAll](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.matchall) and [String.prototype.split](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.split) and [RegExp.prototype.exec ( string )](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp.prototype.exec) and [Array.prototype.concat](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.concat) and [%TypedArray%.prototype.subarray](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.subarray), with honorable mentions for [Bitwise Shift Operators](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-bitwise-shift-operators) and [String.prototype.charCodeAt](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.charcodeat)/[String.prototype.codePointAt](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.codepointat) [in which it is part of a note]).

But I actually share your grammatical preference, and don't remember my motivation here so I have accepted your suggestions. And in the long run, I expect this to be resolved by introducing more structure so the phrasing will become almost moot.

jmdyck: > Sentence fragments here are not _un_common (there are lots of examples like `<p>Returns …` and `<p>Performs …` and `<p>Produces …`, 

(Commit c6ca25fc of PR #2592 changes these by inserting "This function" or "This method".)

gibson042: Thanks @jmdyck, I fully support the efforts of #2592. @bakkot Should I pull forward and merge its changes here?
```suggestion
      <p>This function puts the calling agent in a wait queue and suspends it until it is notified or the wait times out, returning a String differentiating those cases.</p>
      <p>It performs the following steps when called:</p>
```

bakkot: > Should I pull forward and merge its changes here?

Ehhh that seems pretty unrelated. I'd prefer not to try to tweak wording piecemeal; better for each PR to do its own thing.

michaelficarra (approved):

spec.html
```diff
@@ -42032,7 +42032,7 @@ <h1>Atomics.sub ( _typedArray_, _index_, _value_ )</h1>
 
     <emu-clause id="sec-atomics.wait">
       <h1>Atomics.wait ( _typedArray_, _index_, _value_, _timeout_ )</h1>
-      <p>`Atomics.wait` puts the calling agent in a wait queue and puts it to sleep until it is notified or the sleep times out. The following steps are taken:</p>
+      <p>`Atomics.wait` puts the calling agent in a wait queue and suspends it until it is notified or the wait times out, returning a String differentiating those cases. The following steps are taken:</p>
```
syg, line 42035: While not grammatically wrong, I don't think, I'd prefer "... and suspends it until notified or until the wait times out", i.e. duplicating the "until", which I think reads better.

syg (approved):

michaelficarra:
Rebased and accepted @syg's suggestion. Should be ready to merge now.

PR #2712, opened 2022-03-24
Editorial: Align the algorithm steps and description of CanonicalNumericIndexString

gibson042:
* Use a "return number [if canonical], otherwise return `undefined`" step sequence in correspondence with the description.
* Make "canonical numeric string" a linkable \<dfn> term.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2635,7 +2635,7 @@
         </li>
       </ul>
       <p>Properties are identified using key values. A <dfn variants="property keys">property key</dfn> value is either an ECMAScript String value or a Symbol value. All String and Symbol values, including the empty String, are valid as property keys. A <dfn id="property-name">property name</dfn> is a property key that is a String value.</p>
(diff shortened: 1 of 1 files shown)


bakkot (approved):

spec.html
```diff
@@ -5745,10 +5745,10 @@ <h1>
       <emu-alg>
         1. If _argument_ is *"-0"*, return *-0*<sub>𝔽</sub>.
         1. Let _n_ be ! ToNumber(_argument_).
-        1. If SameValue(! ToString(_n_), _argument_) is *false*, return *undefined*.
-        1. Return _n_.
+        1. If SameValue(! ToString(_n_), _argument_) is *true*, return _n_.
+        1. Return *undefined*.
       </emu-alg>
-      <p>A <em>canonical numeric string</em> is any String value for which the CanonicalNumericIndexString abstract operation does not return *undefined*.</p>
+      <p>A <dfn>canonical numeric string</dfn> is any String value for which the CanonicalNumericIndexString abstract operation does not return *undefined*.</p>
```
bakkot, line 5751:
```suggestion
      <p>A <dfn variants="canonical numeric strings">canonical numeric string</dfn> is any String value for which the CanonicalNumericIndexString abstract operation does not return *undefined*.</p>
```

as long as we're dfn'ing things.

gibson042: Done.

PR #2714, opened 2022-03-27
Editorial: MakeMatchIndicesIndexPairArray: avoid creating a property on undefined

ljharb:
This is currently a _potential_ spec bug, not an actual one. This AO is not called such that this would be an error, but there are no assertions to prevent it from being one.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -36322,6 +36322,7 @@ THH:mm:ss.sss
                 1. Let _matchIndexPair_ be *undefined*.
               1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(𝔽(_i_)), _matchIndexPair_).
               1. If _i_ &gt; 0 and _groupNames_[_i_ - 1] is not *undefined*, then
+                1. Assert: _groups_ is not *undefined*.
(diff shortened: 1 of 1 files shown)


bakkot:
I'd prefer to add an assertion, rather than a guard.

ljharb:
How would you word it? That groupNames only contains strings when hasGroups is true?

spec.html
```diff
@@ -36321,7 +36321,7 @@ <h1>
               1. Else,
                 1. Let _matchIndexPair_ be *undefined*.
               1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(𝔽(_i_)), _matchIndexPair_).
-              1. If _i_ &gt; 0 and _groupNames_[_i_ - 1] is not *undefined*, then
+              1. If _hasGroups_ is *true*, _i_ &gt; 0, and _groupNames_[_i_ - 1] is not *undefined*, then
```
bakkot, line 36324:
```suggestion
              1. If _i_ &gt; 0, and _groupNames_[_i_ - 1] is not *undefined*, then
                1. Assert: _groups_ is not *undefined*.
```

bakkot (approved):

PR #2715, opened 2022-03-27
Editorial: `ToZeroPaddedDecimalString`: this AO is nonsensical on a negative integer

ljharb:
`ToZeroPaddedDecimalString(-1, 3)` would produce `0-1`, for example, prior to this change.


bakkot (approved):
Sure.

PR #2719, opened 2022-03-31
Normative: remove Symbol.species from InitializeTypedArrayFromTypedArray

### LINKED ISSUE #2677: Normative: Remove @@species from InitializeTypedArrayFromTypedArray
When creating a `TypedArray` (the _target_) from another `TypedArray` (the _source_) in [InitializeTypedArrayFromTypedArray](https://tc39.es/ecma262/#sec-initializetypedarrayfromtypedarray), _target_ gets a newly allocated `ArrayBuffer` whose prototype is the _source_'s buffer's species constructor's prototype, but that species constructor is never called. Also this only happens if the buffer is not a `SharedArrayBuffer`.

Pretty funny!

Anyway not sure what purpose it serves to only hook up the species **prototype** but never call the species constructor. AFAICT, in all other uses of [SpeciesConstructor](https://tc39.es/ecma262/#sec-speciesconstructor), we call [Construct](https://tc39.es/ecma262/#sec-construct) on it.

Alas, major engines agree on this comedy:

```javascript
let construction_count = 0;

class GrossBuffer extends ArrayBuffer {
  constructor() {
    super(...arguments);
    if (construction_count++ == 1) throw "bad";
  }
  get isGross() { return true; };
  static get [Symbol.species]() { print("barf"); return GrossBuffer; }
}

let gross_buf = new GrossBuffer(1024);
let gross_ta = new Uint8Array(gross_buf);
let mystery_ta = new Int8Array(gross_ta);
print(mystery_ta.buffer.__proto__ === GrossBuffer.prototype);
// true
print(mystery_ta.buffer.isGross);
// true
```

```
#### jsc
barf
true
true

#### sm
barf
true
true

#### v8
barf
true
true

#### xs
barf
true
true
```

Still, I want to see we can remove this one weird use.

/cc @marjakh who initially discovered this.

syg:
Closes #2677

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -39079,8 +39079,7 @@ THH:mm:ss.sss
             1. Else, let _same_ be SameValue(_srcBuffer_, _targetBuffer_).
             1. If _same_ is *true*, then
               1. Let _srcByteLength_ be _source_.[[ByteLength]].
-              1. Set _srcBuffer_ to ? <emu-meta suppress-effects="user-code">CloneArrayBuffer(_srcBuffer_, _srcByteOffset_, _srcByteLength_, %ArrayBuffer%)</emu-meta>.
-              1. NOTE: %ArrayBuffer% is used to clone _srcBuffer_ because is it known to not have any observable side-effects.
+              1. Set _srcBuffer_ to ? CloneArrayBuffer(_srcBuffer_, _srcByteOffset_, _srcByteLength_).
               1. Let _srcByteIndex_ be 0.
             1. Else, let _srcByteIndex_ be _srcByteOffset_.
             1. Let _targetByteIndex_ be _targetOffset_ &times; _targetElementSize_ + _targetByteOffset_.
@@ -39494,15 +39493,10 @@ THH:mm:ss.sss
             1. Let _srcByteOffset_ be _srcArray_.[[ByteOffset]].
             1. Let _elementLength_ be _srcArray_.[[ArrayLength]].
             1. Let _byteLength_ be _elementSize_ &times; _elementLength_.
-            1. If IsSharedArrayBuffer(_srcData_) is *false*, then
-              1. Let _bufferConstructor_ be ? SpeciesConstructor(_srcData_, %ArrayBuffer%).
-            1. Else,
-              1. Let _bufferConstructor_ be %ArrayBuffer%.
             1. If _elementType_ is the same as _srcType_, then
-              1. Let _data_ be ? CloneArrayBuffer(_srcData_, _srcByteOffset_, _byteLength_, _bufferConstructor_).
+              1. Let _data_ be ? CloneArrayBuffer(_srcData_, _srcByteOffset_, _byteLength_).
             1. Else,
-              1. Let _data_ be ? AllocateArrayBuffer(_bufferConstructor_, _byteLength_).
-              1. If IsDetachedBuffer(_srcData_) is *true*, throw a *TypeError* exception.
+              1. Let _data_ be ? <emu-meta suppress-effects="user-code">AllocateArrayBuffer(%ArrayBuffer%, _byteLength_)</emu-meta>.
               1. If _srcArray_.[[ContentType]] &ne; _O_.[[ContentType]], throw a *TypeError* exception.
               1. Let _srcByteIndex_ be _srcByteOffset_.
               1. Let _targetByteIndex_ be 0.
@@ -40670,7 +40664,6 @@ THH:mm:ss.sss
             _srcBuffer_: an ArrayBuffer or a SharedArrayBuffer,
             _srcByteOffset_: a non-negative integer,
             _srcLength_: a non-negative integer,
-            _cloneConstructor_: a constructor,
           ): either a normal completion containing an ArrayBuffer or an abrupt completion
         </h1>
         <dl class="header">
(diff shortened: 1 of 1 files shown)


syg:
Test262 PR: https://github.com/tc39/test262/pull/3460

anba:
`CloneArrayBuffer` is now always called with `%ArrayBuffer%`, so we can also:
1. Remove the `cloneConstructor` argument from `CloneArrayBuffer`.
2. Change the `IsDetachedBuffer` check in `CloneArrayBuffer` to an assertion.

Furthermore I think we should move this line in `InitializeTypedArrayFromTypedArray`:
```
If srcArray.[[ContentType]] ≠ O.[[ContentType]], throw a TypeError exception.
```

So that it appears before the `AllocateArrayBuffer` call (or possibly even before the outer if-else block). That way an incompatible typed array type always throws a TypeError. Currently a TypeError **or** RangeError can be thrown, as seen below. (Technically speaking, for implementations it should already be possible to perform the typed array type check before the allocation, because the RangeError clause in `CreateByteDataBlock` is loose enough, so that it's not required to throw a RangeError if the returned Data Block isn't used. Kind of: "If a tree falls in a forest and no one is around to hear it, does it make a sound?")

SpiderMonkey and V8 show this behaviour:
```js
js> void new BigInt64Array(new Int8Array(4*1024*1024*1024))  
typein:1:6 RangeError: invalid array length
Stack:
  @typein:1:6
js> void new BigInt64Array(new Int8Array(1*1024*1024*1024)) 
typein:2:6 TypeError: Int8Array elements are incompatible with BigInt64Array
Stack:
  @typein:2:6
````

**Comment by syg** [comment]
Thanks for the suggestions, @anba. 1 and 2 are definitely in scope for this PR and I'll make the changes.

The suggestion for moving `If srcArray.[[ContentType]] ≠ O.[[ContentType]], throw a TypeError exception.` seems better as a separate follow up issue, mind filing it separately?

File: spec.html [hunk 1]
```diff
@@ -39078,7 +39078,7 @@ <h1>
             1. Else, let _same_ be SameValue(_srcBuffer_, _targetBuffer_).
             1. If _same_ is *true*, then
               1. Let _srcByteLength_ be _source_.[[ByteLength]].
-              1. Set _srcBuffer_ to ? <emu-meta suppress-effects="user-code">CloneArrayBuffer(_srcBuffer_, _srcByteOffset_, _srcByteLength_, %ArrayBuffer%)</emu-meta>.
+              1. Set _srcBuffer_ to ? <emu-meta suppress-effects="user-code">CloneArrayBuffer(_srcBuffer_, _srcByteOffset_, _srcByteLength_)</emu-meta>.
               1. NOTE: %ArrayBuffer% is used to clone _srcBuffer_ because is it known to not have any observable side-effects.
```
anba, line 39082: This line should also be removed.

syg: Done.

spec.html
```diff
@@ -40678,16 +40673,15 @@ <h1>
...
         </h1>
         <dl class="header">
           <dt>description</dt>
           <dd>It creates a new ArrayBuffer whose data is a copy of _srcBuffer_'s data over the range starting at _srcByteOffset_ and continuing for _srcLength_ bytes.</dd>
         </dl>
         <emu-alg>
-          1. Let _targetBuffer_ be ? AllocateArrayBuffer(_cloneConstructor_, _srcLength_).
-          1. If IsDetachedBuffer(_srcBuffer_) is *true*, throw a *TypeError* exception.
+          1. Let _targetBuffer_ be ? <emu-meta suppress-effects="user-code">AllocateArrayBuffer(%ArrayBuffer%, _srcLength_)</emu-meta>.
+          1. Assert: IsDetachedBuffer(_surcBuffer_) is *false*.
```
bakkot, line 40684:
```suggestion
          1. Assert: IsDetachedBuffer(_srcBuffer_) is *false*.
```
Also should probably go on the previous line, since it's now an assertion about an argument. The fact that it isn't detached by the previous line is implied by the absence of user code on that line, I would think.

spec.html
```diff
@@ -39502,15 +39501,11 @@ <h1>
...
-            1. Else,
-              1. Let _bufferConstructor_ be %ArrayBuffer%.
             1. If _elementType_ is the same as _srcType_, then
-              1. Let _data_ be ? CloneArrayBuffer(_srcData_, _srcByteOffset_, _byteLength_, _bufferConstructor_).
+              1. Let _data_ be ? CloneArrayBuffer(_srcData_, _srcByteOffset_, _byteLength_).
             1. Else,
-              1. Let _data_ be ? AllocateArrayBuffer(_bufferConstructor_, _byteLength_).
-              1. If IsDetachedBuffer(_srcData_) is *true*, throw a *TypeError* exception.
+              1. Let _data_ be ? <emu-meta suppress-effects="user-code">AllocateArrayBuffer(%ArrayBuffer%, _byteLength_)</emu-meta>.
+              1. Assert: IsDetachedBuffer(_srcData_) is *false*.
```
bakkot, line 39508: I don't think this assertion really adds anything; there's a detachedness check above and no user code runs between then and here. Also the only place it's going to be consumed is as an argument to `GetValueFromBuffer`, which has this as an assertion itself.

bakkot (approved):
LGTM other than comments.

spec.html
syg: Fixed typo and moved.

spec.html
syg: Removed.

PR #2721, opened 2022-04-03
Normative: allow duplicate named capture groups

bakkot:
This allows you to have a regex like

```js
/(?<year>[0-9]{4})-[0-9]{2}|[0-9]{2}-(?<year>[0-9]{4})/
```
where a capturing group name is re-used across alternatives. It continues to be illegal to re-use a name within the same alternative.

~~As currently specified, it also enforces that named backreferences correspond to capturing groups in the same alternative, which would make the following (currently legal) program illegal:~~

```js
/(?<a>x)|\k<a>/
```

~~There is no reason to write this because the `\k` can never refer to anything, meaning it will always match the empty string. For this reason I think it should have been illegal in the first place. But if we want to preserve that behavior, it's easy enough to specify.~~

EDIT: updated so that the above remains legal, per plenary.

(I have a [proposal repo](https://github.com/bakkot/proposal-duplicate-named-capturing-groups) for this, but figured it might as well be a PR.)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -35986,7 +35986,7 @@ THH:mm:ss.sss
             It is a Syntax Error if CountLeftCapturingParensWithin(|Pattern|) ≥ 2<sup>32</sup> - 1.
           </li>
           <li>
-            It is a Syntax Error if |Pattern| contains two or more |GroupSpecifier|s for which the CapturingGroupName of |GroupSpecifier| is the same.
+            It is a Syntax Error if |Pattern| contains two distinct |GroupSpecifier|s _x_ and _y_ such that the CapturingGroupName of _x_ is the CapturingGroupName of _y_ and such that MightBothParticipate(_x_, _y_) is *true*.
           </li>
         </ul>
         <emu-grammar>QuantifierPrefix :: `{` DecimalDigits `,` DecimalDigits `}`</emu-grammar>
@@ -36129,6 +36129,22 @@ THH:mm:ss.sss
           1. Assert: _node_ is an instance of a production in <emu-xref href="#sec-patterns">the RegExp Pattern grammar</emu-xref>.
           1. Let _pattern_ be the |Pattern| containing _node_.
           1. Return the number of <emu-grammar>Atom :: `(` GroupSpecifier? Disjunction `)`</emu-grammar> Parse Nodes contained within _pattern_ that either occur before _node_ or contain _node_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-mightbothparticipate" type="abstract operation">
+        <h1>
+          Static Semantics: MightBothParticipate (
+            _x_: a Parse Node,
+            _y_: a Parse Node,
+          ): a Boolean
+        </h1>
+        <dl class="header">
+        </dl>
+        <emu-alg>
+          1. Assert: _x_ and _y_ have the same enclosing |Pattern|.
+          1. If the enclosing |Pattern| contains a <emu-grammar>Disjunction :: Alternative `|` Disjunction</emu-grammar> Parse Node such that either _x_ is contained within the |Alternative| and _y_ is contained within the derived |Disjunction|, or _x_ is contained within the derived |Disjunction| and _y_ is contained within the |Alternative|, return *false*.
+          1. Return *true*.
         </emu-alg>
       </emu-clause>
 
@@ -37189,7 +37205,7 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _n_ be the CapturingGroupNumber of |DecimalEscape|.
           1. Assert: _n_ ≤ _rer_.[[CapturingGroupsCount]].
-          1. Return BackreferenceMatcher(_rer_, _n_, _direction_).
+          1. Return BackreferenceMatcher(_rer_, « _n_ », _direction_).
         </emu-alg>
         <emu-note>
           <p>An escape sequence of the form `\\` followed by a non-zero decimal number _n_ matches the result of the _n_<sup>th</sup> set of capturing parentheses (<emu-xref href="#sec-pattern-notation"></emu-xref>). It is an error if the regular expression has fewer than _n_ capturing parentheses. If the regular expression has _n_ or more capturing parentheses but the _n_<sup>th</sup> one is *undefined* because it has not captured anything, then the backreference always succeeds.</p>
@@ -37225,10 +37241,11 @@ THH:mm:ss.sss
         <emu-grammar>AtomEscape :: `k` GroupName</emu-grammar>
         <emu-alg>
           1. Let _matchingGroupSpecifiers_ be GroupSpecifiersThatMatch(|GroupName|).
-          1. Assert: _matchingGroupSpecifiers_ contains a single |GroupSpecifier|.
-          1. Let _groupSpecifier_ be the sole element of _matchingGroupSpecifiers_.
-          1. Let _parenIndex_ be CountLeftCapturingParensBefore(_groupSpecifier_).
-          1. Return BackreferenceMatcher(_rer_, _parenIndex_, _direction_).
+          1. Let _parenIndices_ be a new empty List.
+          1. For each |GroupSpecifier| _groupSpecifier_ of _matchingGroupSpecifiers_, do
+            1. Let _parenIndex_ be CountLeftCapturingParensBefore(_groupSpecifier_).
+            1. Append _parenIndex_ to _parenIndices_.
+          1. Return BackreferenceMatcher(_rer_, _parenIndices_, _direction_).
         </emu-alg>
 
         <emu-clause id="sec-runtime-semantics-charactersetmatcher-abstract-operation" type="abstract operation">
@@ -37271,20 +37288,23 @@ THH:mm:ss.sss
           <h1>
             BackreferenceMatcher (
               _rer_: a RegExp Record,
-              _n_: a positive integer,
+              _ns_: a List of positive integers,
               _direction_: ~forward~ or ~backward~,
             ): a Matcher
           </h1>
           <dl class="header">
           </dl>
           <emu-alg>
-            1. Assert: _n_ ≥ 1.
-            1. Return a new Matcher with parameters (_x_, _c_) that captures _rer_, _n_, and _direction_ and performs the following steps when called:
+            1. Return a new Matcher with parameters (_x_, _c_) that captures _rer_, _ns_, and _direction_ and performs the following steps when called:
               1. Assert: _x_ is a MatchState.
               1. Assert: _c_ is a MatcherContinuation.
               1. Let _Input_ be _x_.[[Input]].
               1. Let _cap_ be _x_.[[Captures]].
-              1. Let _r_ be _cap_[_n_].
+              1. Let _r_ be *undefined*.
+              1. For each integer _n_ of _ns_, do
+                1. If _cap_[_n_] is not *undefined*, then
+                  1. Assert: _r_ is *undefined*.
+                  1. Set _r_ to _cap_[_n_].
               1. If _r_ is *undefined*, return _c_(_x_).
               1. Let _e_ be _x_.[[EndIndex]].
               1. Let _rs_ be _r_.[[StartIndex]].
@@ -38523,6 +38543,7 @@ THH:mm:ss.sss
             1. Let _groups_ be *undefined*.
             1. Let _hasGroups_ be *false*.
           1. Perform ! CreateDataPropertyOrThrow(_A_, *"groups"*, _groups_).
+          1. Let _matchedGroupNames_ be a new empty List.
           1. For each integer _i_ such that 1 ≤ _i_ ≤ _n_, in ascending order, do
             1. Let _captureI_ be _i_<sup>th</sup> element of _r_.[[Captures]].
             1. If _captureI_ is *undefined*, then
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -34644,7 +34644,7 @@ <h1>Static Semantics: Early Errors</h1>
             It is a Syntax Error if _NcapturingParens_ &ge; 2<sup>32</sup> - 1.
           </li>
           <li>
-            It is a Syntax Error if |Pattern| contains multiple |GroupSpecifier|s whose enclosed |RegExpIdentifierName|s have the same CapturingGroupName.
+            It is an Syntax Error if |Pattern| contains a pair of |GroupSpecifier|s (_x_, _y_) whose enclosed RegExpIdentifierNames have the same CapturingGroupName and such that CanBothParticipate(_x_, _y_) is *true*.
```
ljharb, line 34647: an syntax? O.o
```suggestion
            It is a Syntax Error if |Pattern| contains a pair of |GroupSpecifier|s (_x_, _y_) whose enclosed RegExpIdentifierNames have the same CapturingGroupName and such that CanBothParticipate(_x_, _y_) is *true*.
```

bakkot: lol whoops, fixed.

spec.html
```diff
@@ -35566,7 +35582,7 @@ <h1>
         </emu-alg>
         <emu-grammar>AtomEscape :: `k` GroupName</emu-grammar>
         <emu-alg>
-          1. Search the enclosing |Pattern| for an instance of a |GroupSpecifier| containing a |RegExpIdentifierName| which has a CapturingGroupName equal to the CapturingGroupName of the |RegExpIdentifierName| contained in |GroupName|.
+          1. Search the enclosing |Pattern| for an instance of a |GroupSpecifier| containing a |RegExpIdentifierName| which has a CapturingGroupName equal to the CapturingGroupName of the |RegExpIdentifierName| contained in |GroupName|, and such that CanBothParticipate(|AtomEscape|, |GroupSpecifier|) is *true*.
           1. Assert: A unique such |GroupSpecifier| is found.
```
bakkot, line 35586: This is actually wrong. Consider:

```js
/(?:(?<a>x)|(?<a>y))\k<a>/
```

This is legal as specified (and should be) - the duplicated name occurs in different alternative - but both of the `a` groups can participate alongside the backreference. So we no longer know statically precisely which group a backreference refers to.

This _should_ work - at most one of those two can have matched, so the backreference should be to whichever of the two did match - but these semantics don't get us there. Making this work will be more complicated.

spec.html
```diff
@@ -36262,8 +36282,16 @@ <h1>
               1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(𝔽(_i_)), _capturedValue_).
               1. If the _i_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
                 1. Let _s_ be the CapturingGroupName of that |GroupName|.
-                1. Perform ! CreateDataPropertyOrThrow(_groups_, _s_, _capturedValue_).
-                1. Append _s_ to _groupNames_.
+                1. Let _isMatchedElsewhere_ be *false*.
+                1. For each integer _j_ such that _j_ &ne; _i_, _j_ &ge; 1, and _j_ &le; _n_, do
+                  1. If the _j_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
+                    1. Let _sj_ be the CapturingGroupName of that |GroupName|.
+                    1. If _sj_ equals _s_ and the _j_<sup>th</sup> capture of _R_ is not *undefined*, set _isMatchedElsewhere_ to *true*.
```
gibson042, line 36289: The _j_<sup>th</sup> capture of _R_ is a static concept; this dynamic state is in _r_.
```suggestion
                    1. If _sj_ equals _s_ and the _j_<sup>th</sup> element of _r_'s _captures_ List is not *undefined*, set _isMatchedElsewhere_ to *true*.
```
Also, is "equals" valid for comparing String values? I'm not clear on the preferred approach, and have tended to use patterns like `SameValue(_s_, _sj_) is *true*` recently.

spec.html
```diff
@@ -36262,8 +36282,16 @@ <h1>
               1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(𝔽(_i_)), _capturedValue_).
               1. If the _i_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
                 1. Let _s_ be the CapturingGroupName of that |GroupName|.
-                1. Perform ! CreateDataPropertyOrThrow(_groups_, _s_, _capturedValue_).
-                1. Append _s_ to _groupNames_.
+                1. Let _isMatchedElsewhere_ be *false*.
+                1. For each integer _j_ such that _j_ &ne; _i_, _j_ &ge; 1, and _j_ &le; _n_, do
+                  1. If the _j_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
```
gibson042, line 36287: Would this be better expressed more procedurally?
```suggestion
                1. For each integer _j_ such that _j_ &ge; 1 and _j_ &le; _n_, in ascending order, do
                  1. If _j_ &ne; _i_ and the _j_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
```

spec.html
```diff
@@ -36262,8 +36282,16 @@ <h1>
...
+                    1. Let _sj_ be the CapturingGroupName of that |GroupName|.
```
gibson042, line 36288: I would not object to refactoring these aliases, e.g. _s_ → _groupName_ and _sj_ → _otherName_.

spec.html
```diff
@@ -36262,8 +36282,16 @@ <h1>
...
+                    1. If _sj_ equals _s_ and the _j_<sup>th</sup> capture of _R_ is not *undefined*, set _isMatchedElsewhere_ to *true*.
+                1. If _isMatchedElsewhere_ is *false*, then
+                  1. Perform ! CreateDataPropertyOrThrow(_groups_, _s_, _capturedValue_).
+                  1. Append _s_ to _groupNames_.
+                1. Else,
+                  1. Append *undefined* to _groupNames_.
               1. Else,
                 1. Append *undefined* to _groupNames_.
```
gibson042, line 36296: Would it be worthwhile to collapse these Else cases?
```
              1. Let _groupName_ be *undefined*.
              1. If the _i_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
                1. Set _groupName_ to the CapturingGroupName of that |GroupName|.
                1. For each integer _j_ such that …, do
                  1. If the _j_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
                    1. …
                    1. If …, set _groupName_ to *undefined*.
                1. If _groupName_ is not *undefined*, then
                  1. Perform ! CreateDataPropertyOrThrow(_groups_, _groupName_, _capturedValue_).
              1. Append _groupName_ to _groupNames_.
```

gibson042 (approved):
The only materially necessary change I see is `_j_<sup>th</sup> capture of _R_` → `_j_<sup>th</sup> element of _r_'s _captures_ List`, but I also left some suggestions that you can adopt or ignore at your discretion.

spec.html
```diff
@@ -36262,8 +36282,17 @@ <h1>
               1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(𝔽(_i_)), _capturedValue_).
               1. If the _i_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
                 1. Let _s_ be the CapturingGroupName of that |GroupName|.
-                1. Perform ! CreateDataPropertyOrThrow(_groups_, _s_, _capturedValue_).
-                1. Append _s_ to _groupNames_.
+                1. Let _isMatchedElsewhere_ be *false*.
+                1. For each integer _j_ such that _j_ &ne; _i_, _j_ &ge; 1, and _j_ &le; _n_, do
+                  1. If the _j_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
+                    1. Let _sj_ be the CapturingGroupName of that |GroupName|.
+                    1. If _sj_ equals _s_ and the _j_<sup>th</sup> element of _R_'s _captures_ List is not *undefined*, set _isMatchedElsewhere_ to *true*.
```
gibson042, line 36289:
```suggestion
                    1. If _sj_ equals _s_ and the _j_<sup>th</sup> element of _r_'s _captures_ List is not *undefined*, set _isMatchedElsewhere_ to *true*.
```

bakkot: Sight, thanks, fixed.

spec.html
```diff
@@ -36262,8 +36282,17 @@ <h1>
               1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(𝔽(_i_)), _capturedValue_).
               1. If the _i_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
                 1. Let _s_ be the CapturingGroupName of that |GroupName|.
-                1. Perform ! CreateDataPropertyOrThrow(_groups_, _s_, _capturedValue_).
-                1. Append _s_ to _groupNames_.
+                1. Let _isMatchedEarlier_ be *false*.
+                1. For each integer _j_ such that _j_ &ge; 1 and _j_ &lt; _i_, do
```
gibson042, line 36286: Another possible simplification:
 ```patch
             1. Perform ! CreateDataPropertyOrThrow(_A_, *"groups"*, _groups_).
+            1. Let _matchedGroupNames_ be a new empty List.
             1. For each integer _i_ such that _i_ &ge; 1 and _i_ &le; _n_, in ascending order, do
               1. Let _captureI_ be _i_<sup>th</sup> element of _r_'s _captures_ List.
               …
               1. If the _i_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
                 1. Let _s_ be the CapturingGroupName of that |GroupName|.
-                1. Let _isMatchedEarlier_ be *false*.
-                1. For each integer _j_ such that _j_ &ge; 1 and _j_ &lt; _i_, do
-                  1. If the _j_<sup>th</sup> capture of _R_ was defined with a |GroupName|, then
-                    1. Let _sj_ be the CapturingGroupName of that |GroupName|.
-                    1. If SameValue(_sj_, _s_) is *true* and the _j_<sup>th</sup> element of _r_'s _captures_ List is not *undefined*, set _isMatchedEarlier_ to *true*.
-                1. If _isMatchedEarlier_ is *false*, then
+                1. If _matchedGroupNames_ contains _s_, then
+                  1. Set _s_ to *undefined*.
+                1. Else,
+                  1. If _capturedValue_ is not *undefined*, append _s_ to _matchedGroupNames_.
                   1. NOTE: If there are multiple groups named _s_, _groups_ may already have an _s_ property at this point. However, because _groups_ is an ordinary object whose properties are all writable data properties, the call to CreateDataPropertyOrThrow is nevertheless guaranteed to succeed.
                   1. Perform ! CreateDataPropertyOrThrow(_groups_, _s_, _capturedValue_).
-                  1. Append _s_ to _groupNames_.
-                1. Else,
-                  1. Append *undefined* to _groupNames_.
+                1. Append _s_ to _groupNames_.
               1. Else,
                 1. Append *undefined* to _groupNames_.
 ```

bakkot: Oh, nice. Pushed a [slight variation](https://github.com/tc39/ecma262/commit/61e7436a9e2dafdc79b1cc24f4504e4c52ef8bb3) which doesn't involve mutating _s_.

spec.html
```diff
@@ -34758,6 +34758,22 @@ <h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-patterns-static-semantics-can-both-participate" type="abstract operation">
+        <h1>
+          Static Semantics: CanBothParticipate (
```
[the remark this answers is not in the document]
msaboff: I agree with @gibson042, the sense of the `CanBothParticipate` AO is confusing.  It returns _true_ when _x_ and _y_ are in the same _Alternative_ and therefore we should throw a `Syntax Error`.  Can we rename `CanBothParticipate` to `CannotBothParticipate`?

syg:
Happy do defer the editorial review, I'm told there've been enough eyes on this.

spec.html
```diff
@@ -35715,7 +35715,7 @@ <h1>Static Semantics: Early Errors</h1>
             It is a Syntax Error if CountLeftCapturingParensWithin(|Pattern|) ≥ 2<sup>32</sup> - 1.
           </li>
           <li>
-            It is a Syntax Error if |Pattern| contains two or more |GroupSpecifier|s for which CapturingGroupName of |GroupSpecifier| is the same.
+            It is a Syntax Error if |Pattern| contains two distinct |GroupSpecifier|s _x_ and _y_ for which CapturingGroupName(_x_) is the same as CapturingGroupName(_y_) and such that CanBothParticipate(_x_, _y_) is *true*.
```
jmdyck, line 35718: 1) PR #2877 says to avoid "is the same as", use "is" for comparing Strings.

2) It's rare (though not unheard of) to invoke an SDO with parenthesis notation. On the other hand, using the normal notation would put `CapturingGroupName of _y_` on the RHS of `is`, which I don't think we *ever* do.

3) `... for which A and such that B` is odd. Seems like either the two arms should agree re "for which" vs "such that", or else just use one that covers both arms: `... such that A and B`

```suggestion
            It is a Syntax Error if |Pattern| contains two distinct |GroupSpecifier|s _x_ and _y_ such that CapturingGroupName(_x_) is CapturingGroupName(_y_) and CanBothParticipate(_x_, _y_) is *true*.
```

spec.html
```diff
@@ -35861,6 +35861,22 @@ <h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="sec-patterns-static-semantics-can-both-participate" type="abstract operation">
```
jmdyck, line 35864:
```suggestion
      <emu-clause id="sec-canbothparticipate" type="abstract operation">
```

jmdyck (dismissed):

michaelficarra (approved):
LGTM other than @jmdyck's nits.

spec.html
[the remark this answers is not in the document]
michaelficarra: It's more like `MightBothParticipate`. As in, both could be components of a single match.

spec.html
bakkot: Changed to

> It is a Syntax Error if |Pattern| contains two distinct |GroupSpecifier|s `_x_` and `_y_` such that the CapturingGroupName of `_x_` is the CapturingGroupName of `_y_` and such that CanBothParticipate(`_x_`, `_y_`) is `*true*`.

bakkot:
Comments addressed.

spec.html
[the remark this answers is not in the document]
bakkot: Renamed to `MightBothParticipate`.

Issue #1673, opened 2019-08-20
bug? re named groups in Annex B


jmdyck:
In the [main body](https://tc39.es/ecma262/#sec-patterns), we have (in part):
> Atom[U, N] :: ( GroupSpecifier[?U] Disjunction[?U, ?N] )

[Annex B](https://tc39.es/ecma262/#sec-regular-expressions-patterns) doesn't change that production, but it does restrict its use (in `Term`) to `[+U]`. For `[~U]`, the corresponding production is
> ExtendedAtom[N] :: ( Disjunction[~U, ?N] )

That is, if a web browser (following Annex B) is parsing a regex with `[~U]`, it is supposed to reject named groups.

I'm not sure if this is a bug, but it does seem odd to me.

- A *non*-browser is required to accept named groups when parsing with `[~U]`, and I thought Annex B was only supposed to *add* features, not subtract.

- The Annex B production for `AtomEscape` allows named *backreferences* under `[~U, +N]`, but what's the point if `[~U]` doesn't allow named groups in the first place?

This goes back to the introduction of named capture groups in 95ec0c6e (PR #1027).

The final state of the [original proposal](https://github.com/tc39/proposal-regexp-named-groups) doesn't mention Annex B, so it's unclear what #1027 was supposed to do re Annex B.

@mathiasbynens and/or @littledan, please comment.

PR #2729, opened 2022-04-09
Normative: allow named capture groups in Annex B non-u regex grammar

bakkot:
Fixes https://github.com/tc39/ecma262/issues/1673.

Capturing group names were introduced in https://github.com/tc39/ecma262/issues/1673, in both `u` and non-`u` regexes, but the alternate grammar for regexes in Annex B (i.e., the actual grammar for regexes) replaces in non-`u` regexes the nonterminal (`Atom`) which derives `GroupSpecifier`, and that PR neglected to update the new nonterminal (`ExtendedAtom`). This PR corrects that oversight.

I've marked this as a normative change rather than editorial because the current spec is, technically, coherent. However, the current spec does not match committee intent or web reality, so in my view we can land this without asking for consensus.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -46979,7 +46979,7 @@ THH:mm:ss.sss
           `\` AtomEscape[~UnicodeMode, ?N]
           `\` [lookahead == `c`]
           CharacterClass[~UnicodeMode]
-          `(` Disjunction[~UnicodeMode, ?N] `)`
+          `(` GroupSpecifier[~UnicodeMode] Disjunction[~UnicodeMode, ?N] `)`
           `(` `?` `:` Disjunction[~UnicodeMode, ?N] `)`
           InvalidBracedQuantifier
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -46953,7 +46953,7 @@ <h2>Syntax</h2>
           `\` AtomEscape[~UnicodeMode, ?N]
           `\` [lookahead == `c`]
           CharacterClass[~UnicodeMode]
-          `(` Disjunction[~UnicodeMode, ?N] `)`
+          `(` GroupSpecifier Disjunction[~UnicodeMode, ?N] `)`
```
jmdyck, line 46956:
```suggestion
          `(` GroupSpecifier[~UnicodeMode] Disjunction[~UnicodeMode, ?N] `)`
```

syg (approved):

michaelficarra (approved):

PR #2730, opened 2022-04-11
Editorial: Improve consistency of mathematical conventions

gibson042:
* ~~Escape `>` as `&gt;`.~~
* Make clear that the symbol used for mathematical subtraction is `-` (U+002D HYPHEN-MINUS) rather than e.g. `–` (U+2013 EN DASH) or `−` (U+2212 MINUS SIGN).

Ref https://github.com/tc39/ecma402/pull/672

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1010,7 +1010,7 @@
       </ul>
 
       <p>In the language of this specification, numerical values are distinguished among different numeric kinds using subscript suffixes. The subscript <sub>𝔽</sub> refers to Numbers, and the subscript <sub>ℤ</sub> refers to BigInts. Numeric values without a subscript suffix refer to mathematical values.</p>
-      <p>Numeric operators such as +, &times;, =, and &ge; refer to those operations as determined by the type of the operands. When applied to mathematical values, the operators refer to the usual mathematical operations. When applied to extended mathematical values, the operators refer to the usual mathematical operations over the extended real numbers; indeterminate forms are not defined and their use in this specification should be considered an editorial error. When applied to Numbers, the operators refer to the relevant operations within IEEE 754-2019. When applied to BigInts, the operators refer to the usual mathematical operations applied to the mathematical value of the BigInt.</p>
+      <p>Numeric operators such as +, -, &times;, =, and &ge; refer to those operations as determined by the type of the operands. When applied to mathematical values, the operators refer to the usual mathematical operations. When applied to extended mathematical values, the operators refer to the usual mathematical operations over the extended real numbers; indeterminate forms are not defined and their use in this specification should be considered an editorial error. When applied to Numbers, the operators refer to the relevant operations within IEEE 754-2019. When applied to BigInts, the operators refer to the usual mathematical operations applied to the mathematical value of the BigInt.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -17401,7 +17401,7 @@ <h2>Syntax</h2>
           `u` `{` CodePoint [lookahead &notin; HexDigit] [lookahead != `}`]
 
         NotCodePoint ::
-          HexDigits[~Sep] [> but only if MV of |HexDigits| > 0x10FFFF]
+          HexDigits[~Sep] [> but only if MV of |HexDigits| &gt; 0x10FFFF]
```
ljharb, line 17404: this seems like something ecmarkup should be able to do - also there's another `>` earlier in the line, would that need it too?

gibson042: > this seems like something ecmarkup should be able to do

It seems to already be handled for output, but this corrects the existing _source_ inconsistency w.r.t. `[> but only if … &le; …]` as observable three lines down. Extension to linting and/or formatting would make sense.

> also there's another `>` earlier in the line, would that need it too?

No, because literal `[> …]` is [grammarkdown prose assertion syntax](https://rbuckton.github.io/grammarkdown/usage/syntax.html#assertions) (and is [converted to `<emu-gmod>…</emu-gmod>` on output](https://github.com/rbuckton/grammarkdown/blob/061b7b1c0e39fc9462add8a64e3920a3d5212921/src/emitter/ecmarkup.ts#L349)).

bakkot: Grammarkdown accepts `[&gt;`: https://github.com/rbuckton/grammarkdown/issues/60

But it also accepts `>`, in both places. I kind of prefer it to be `>` but I recognize we're inconsistent.

gibson042: Should I revert changes to this line then? emu-format isn't happy with the `&gt;` anyway.

gibson042: I just did it. The main goal here is to indicate use of `-` rather than `–` or `−`.

michaelficarra:
Editors don't think this is an improvement. This list is not meant to be exhaustive. The closer we make this list to exhaustive, the more likely it is that readers will come to the incorrect conclusion that it is. So an obvious omission like this one helps, if anything.

michaelficarra:
@gibson042 I don't think it's useful to try to explore that unless you think there's a usage in the spec of an arithmetic operator that you think might be unclear to some reader.

gibson042:
@michaelficarra It's about consistency rather than confusion. Subtraction can plausibly be represented by either "-" or "−", multiplication by "*" or "×", division by "/" or "÷", etc. I don't think we want a free mixture, but there's no documentation of which option is correct other than "look around", and that clearly didn't work for ECMA-402 as evidenced by https://github.com/tc39/ecma402/pull/672/commits/a50498d89d11a9103ad6b9aab4146a1cee7d2b89 .

michaelficarra:
There are many places where we are consistent without documenting our editorial convention within the spec. This section is not meant to document an editorial convention but to explain a general notational convention.

gibson042:
How would @tc39/ecma262-editors feel about providing an explicit list of numeric operators? It looks to me like there are only 10:
* \+
* \-
* &times;
* /
* =
* &ne;
* &lt;
* &le;
* &gt;
* &ge;

PR #2731, opened 2022-04-11
Refactor the "enforce format" job to output literal backticks rather than command output

gibson042:
cf. https://github.com/tc39/ecma262/runs/5979841913?check_suite_focus=true
```
Run if ! $(npm bin)/emu-format --check spec.html; then echo "You need to run `npm run format`"; exit 1; fi
Need formatting:
spec.html
You need to run 
> ecma262@1.0.0 format
> emu-format --write spec.html
```


ljharb (approved):
woof, good catch

bakkot (approved):

PR #2737, opened 2022-04-14
Editorial: Make the steps of IsLessThan for String values more explicit

gibson042:
This should simplify reading and also help with KAIST-like efforts.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6129,12 +6129,14 @@
           1. Let _py_ be ? ToPrimitive(_y_, ~number~).
           1. Let _px_ be ? ToPrimitive(_x_, ~number~).
         1. [id="step-arc-string-check"] If Type(_px_) is String and Type(_py_) is String, then
-          1. If IsStringPrefix(_py_, _px_) is *true*, return *false*.
-          1. If IsStringPrefix(_px_, _py_) is *true*, return *true*.
-          1. Let _k_ be the smallest non-negative integer such that the code unit at index _k_ within _px_ is different from the code unit at index _k_ within _py_. (There must be such a _k_, for neither String is a prefix of the other.)
-          1. Let _m_ be the integer that is the numeric value of the code unit at index _k_ within _px_.
-          1. Let _n_ be the integer that is the numeric value of the code unit at index _k_ within _py_.
-          1. If _m_ &lt; _n_, return *true*. Otherwise, return *false*.
+          1. Let _lx_ be the length of _px_.
+          1. Let _ly_ be the length of _py_.
+          1. For each integer _i_ starting with 0 such that _i_ &lt; min(_lx_, _ly_), in ascending order, do
+            1. Let _cx_ be the integer that is the numeric value of the code unit at index _i_ within _px_.
+            1. Let _cy_ be the integer that is the numeric value of the code unit at index _i_ within _py_.
+            1. If _cx_ &lt; _cy_, return *true*.
+            1. If _cx_ &gt; _cy_, return *false*.
+          1. If _lx_ &lt; _ly_, return *true*. Otherwise, return *false*.
         1. Else,
           1. If Type(_px_) is BigInt and Type(_py_) is String, then
             1. Let _ny_ be StringToBigInt(_py_).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -6126,12 +6126,14 @@ <h1>
...
         1. [id="step-arc-string-check"] If Type(_px_) is String and Type(_py_) is String, then
-          1. If IsStringPrefix(_py_, _px_) is *true*, return *false*.
-          1. If IsStringPrefix(_px_, _py_) is *true*, return *true*.
-          1. Let _k_ be the smallest non-negative integer such that the code unit at index _k_ within _px_ is different from the code unit at index _k_ within _py_. (There must be such a _k_, for neither String is a prefix of the other.)
-          1. Let _m_ be the integer that is the numeric value of the code unit at index _k_ within _px_.
-          1. Let _n_ be the integer that is the numeric value of the code unit at index _k_ within _py_.
-          1. If _m_ &lt; _n_, return *true*. Otherwise, return *false*.
+          1. Let _lx_ be the length of _px_.
+          1. Let _ly_ be the length of _py_.
+          1. For each integer _i_ starting with 0 such that _i_ &lt; min(_lx_, _ly_), in ascending order, do
```
ljharb, line 6131: this seems like the more common "let i = 0; repeat while i < min(_lx_, _ly_)" and then with "increment i by 1" as the last step?

gibson042: That's a fair point. What about the other uses of this minority pattern?
```sh
awk '
  /<emu-clause/ { clause=$0; h1=""; }
  /<h1/ { h1=$0 }
  /1. For each.*starting with/ { printf "\n%s\n%s\n%s\n", clause, h1, $0 }
' spec.html
```

1. [StringIndexOf](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-stringindexof)
   > For each integer _i_ starting with _fromIndex_ such that _i_ &le; _len_ - _searchLen_, in ascending order, do

2. [String Exotic Objects [[OwnPropertyKeys]]](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-string-exotic-objects-ownpropertykeys)
   > For each integer _i_ starting with 0 such that _i_ &lt; _len_, in ascending order, do

3. [Integer-Indexed Exotic Objects [[OwnPropertyKeys]]](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-integer-indexed-exotic-objects-ownpropertykeys)
   > For each integer _i_ starting with 0 such that _i_ &lt; _O_.[[ArrayLength]], in ascending order, do

4. [String.prototype.lastIndexOf](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.lastindexof)
   > For each non-negative integer _i_ starting with _start_ such that _i_ &le; _len_ - _searchLen_, in descending order, do

5. [MakeMatchIndicesIndexPairArray](https://tc39.es/ecma262/multipage/text-processing.html#sec-makematchindicesindexpairarray)
   > For each integer _i_ starting with 0 such that _i_ &lt; _n_, in ascending order, do

ljharb: the "such that" stuff reads awkwardly to me, we should see what the editors think, but i'd prefer to change them all to the more standard repeat loops.

bakkot: Personally I'm happier with range-based loops, at least in this case.

spec.html
```diff
@@ -6160,7 +6162,7 @@ <h1>
...
+        <p>The comparison of Strings uses a simple lexicographic ordering on sequences of UTF-16 code unit values. There is no attempt to use the more complex, semantically oriented definitions of character or string equality and collating order defined in the Unicode specification. Therefore String values that are canonically equal according to the Unicode Standard but not in the same normalization form could test as unequal. Also note that ordering by <em>code unit</em> differs from ordering by <em>code point</em> for Strings containing surrogate pairs.</p>
```
bakkot, line 6165:
```suggestion
        <p>The comparison of Strings uses a simple lexicographic ordering on sequences of UTF-16 code unit values. There is no attempt to use the more complex, semantically oriented definitions of character or string equality and collating order defined in the Unicode specification. Therefore String values that are canonically equal according to the Unicode Standard but not in the same normalization form could test as unequal. Also note that lexicographic ordering by <em>code unit</em> differs from ordering by <em>code point</em> for Strings containing surrogate pairs.</p>
```

bakkot (approved):
I am approximately neutral on this change but am OK with it.

spec.html
gibson042: Accepted.

syg (approved):
lgtm. Please remove the definition of IsStringPrefix as well since these seem to be the only uses.

gibson042:
Nice catch on IsStringPrefix! Removed.

ljharb:
RIP IsStringPrefix, ES2018-2022.

Issue #1426, opened 2019-01-20
Possible bug in GetSubstitution or test262


targos:
I think I found a mismatch between the spec and some test262 tests.

In the [table for Replacement Text Symbol Substitutions](https://tc39.github.io/ecma262/#table-45), the 5th row says "$n where n is one of 1 2 3 4 5 6 7 8 9 and $n is not followed by a decimal digit".

In the following test: https://github.com/tc39/test262/blob/master/test/built-ins/String/prototype/replace/S15.5.4.11_A3_T1.js
The substition string is `"$1115"`. This string does not match that row of the table (`$1` is followed by the decimal digit `1`), but the test result expects that it does (`$1` is replaced by the first capture).

Which is correct? Spect text or test?

anba:
Probably an oversight in #853, which changed it from `If nn is 00 or nn>m, the result is implementation-defined.` to `If nn is 00 or nn > m, no replacement is done.` while we actually want to fallback to `n mod 10` first and then to `no-replacement`?

claudepache:
BTW, the wording ”no replacement is done” introduced in #853 is obscure (I first understood as: `"foo".replace(/foo/, "x$42")` will return `"foo"`, and not `"x$42"`). Now, I understand that the expression meant in fact: ”use the literal string `$nn`”.

targos:
> Probably an oversight in #853, which changed it from `If nn is 00 or nn>m, the result is implementation-defined.` to `If nn is 00 or nn > m, no replacement is done.` while we actually want to fallback to `n mod 10` first and then to `no-replacement`?

That seems right. I tested Chrome, Firefox and Edge. All behave like this:

![image](https://user-images.githubusercontent.com/2352663/51709152-fbafec00-2025-11e9-8e2b-f37a570230a8.png)

Should I try to do a PR to fix the spec?

littledan:
A PR for the fix would be great. Thank you for your good work here in catching my error!

ljharb:
I ran that code in eshost and got this:
<details><summary>code</summary>

```js
print('ABCDEFGHIJKL'.replace(/(A)(B)(C)(D)(E)(F)(G)(H)(I)(J)(K)(L)/, '$12'));
print('ABCDEFGHIJKL'.replace(/(A)(B)(C)(D)(E)(F)(G)(H)(I)(J)(K)/, '$12'));
```
</details>

<details><summary>results</summary>

```sh
#### Chakra
L
A2L

#### JavaScriptCore
L
[...]

PR #2739, opened 2022-04-14
Editorial: Simplify GetSubstitution

gibson042:
* Remove an unnecessary "for each" step
* Combine out-of-bounds index branches
* Make aliases more clear
* Replace "number of code units in [string]" with "length of"

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33991,24 +33991,20 @@ THH:mm:ss.sss
                 1. Let _refReplacement_ be the substring of _str_ from min(_tailPos_, _stringLength_).
                 1. NOTE: _tailPos_ can exceed _stringLength_ only if this abstract operation was invoked by a call to the intrinsic @@replace method of %RegExp.prototype% on an object whose *"exec"* property is not the intrinsic %RegExp.prototype.exec%.
               1. Else if _templateRemainder_ starts with *"$"* followed by 1 or more decimal digits, then
-                1. Let _found_ be *false*.
-                1. For each integer _d_ of &laquo; 2, 1 &raquo;, do
-                  1. If _found_ is *false* and _templateRemainder_ starts with *"$"* followed by _d_ or more decimal digits, then
-                    1. Set _found_ to *true*.
-                    1. Let _ref_ be the substring of _templateRemainder_ from 0 to 1 + _d_.
-                    1. Let _digits_ be the substring of _templateRemainder_ from 1 to 1 + _d_.
-                    1. Let _index_ be ℝ(StringToNumber(_digits_)).
-                    1. Assert: 0 &le; _index_ &le; 99.
-                    1. If _index_ = 0, then
-                      1. Let _refReplacement_ be _ref_.
-                    1. Else if _index_ &le; the number of elements in _captures_, then
-                      1. Let _capture_ be _captures_[_index_ - 1].
-                      1. If _capture_ is *undefined*, then
-                        1. Let _refReplacement_ be the empty String.
-                      1. Else,
-                        1. Let _refReplacement_ be _capture_.
-                    1. Else,
-                      1. Let _refReplacement_ be _ref_.
+                1. If _templateRemainder_ starts with *"$"* followed by 2 or more decimal digits, let _digitCount_ be 2. Otherwise, let _digitCount_ be 1.
+                1. Let _ref_ be the substring of _templateRemainder_ from 0 to 1 + _digitCount_.
+                1. Let _digits_ be the substring of _templateRemainder_ from 1 to 1 + _digitCount_.
+                1. Let _index_ be ℝ(StringToNumber(_digits_)).
+                1. Assert: 0 &le; _index_ &le; 99.
+                1. Let _captureLen_ be the number of elements in _captures_.
+                1. If 1 &le; _index_ &le; _captureLen_, then
+                  1. Let _capture_ be _captures_[_index_ - 1].
+                  1. If _capture_ is *undefined*, then
+                    1. Let _refReplacement_ be the empty String.
+                  1. Else,
+                    1. Let _refReplacement_ be _capture_.
+                1. Else,
+                  1. Let _refReplacement_ be _ref_.
               1. Else if _templateRemainder_ starts with *"$&lt;"*, then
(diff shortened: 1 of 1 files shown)


jmdyck:
I think these changes are semantics-preserving. The problem is that the "for each" step that the first commit removes is there to allow an easy fix for Issue #1426. See https://github.com/tc39/ecma262/pull/2484#issuecomment-920918576. (It looks like I just forgot to make the follow-up PR.)

(I note that you also have PR #1732 to fix #1426, but it's based the pre-2484 spec, where GetSubstitution was mostly defined by a table. I don't *think* the fix there would translate well to current spec, or to the result of this PR, but I'm not sure.)

gibson042:
Right, if this simplification is accepted then I'm expecting the fix for #1426 to be explicit, e.g.
```patch
                 1. If _templateRemainder_ starts with *"$"* followed by 2 or more decimal digits, let _digitCount_ be 2. Otherwise, let _digitCount_ be 1.
                 1. Let _ref_ be the substring of _templateRemainder_ from 0 to 1 + _digitCount_.
                 1. Let _digits_ be the substring of _templateRemainder_ from 1 to 1 + _digitCount_.
                 1. Let _index_ be ℝ(StringToNumber(_digits_)).
                 1. Assert: 0 &le; _index_ &le; 99.
-                1. If 0 &lt; _index_ &le; the number of elements in _captures_, then
+                1. Let _captureLen_ be the number of elements in _captures_.
+                1. If _digitCount_ &gt; 1 and _index_ &gt; _captureLen_, then
+                  1. Set _ref_ to the substring of _ref_ from 0 to _digitCount_.
+                  1. Set _index_ to floor(_index_ / 10).
+                1. If 0 &lt; _index_ &le; _captureLen_, then
                   1. Let _capture_ be _captures_[_index_ - 1].
                   1. If _capture_ is *undefined*, then
                     1. Let _refReplacement_ be the empty String.
                   1. Else,
                     1. Let _refReplacement_ be _capture_.
                 1. Else,
                   1. Let _refReplacement_ be _ref_.
```

jmdyck:
Yeah, I think that'll work.

spec.html
```diff
@@ -34032,7 +34028,7 @@ <h1>
               1. Else,
                 1. Let _ref_ be the substring of _templateRemainder_ from 0 to 1.
                 1. Let _refReplacement_ be _ref_.
-              1. Let _refLength_ be the number of code units in _ref_.
+              1. Let _refLength_ be the length of _ref_.
```
gibson042, line 34031: We could also eliminate the redundancy of the several «if _templateRemainder_ starts with *"$…"*, then let _ref_ be *"$…"*» steps:
```
1. If _templateRemainder_ starts with *"$$"*, then
  1. Let _advanceBy_ be 2.
  1. Let _replacement_ be *"$"*.
1. Else if _templateRemainder_ starts with *"$`"*, then
  1. Let _advanceBy_ be 2.
  1. Let _replacement_ be the substring of _str_ from 0 to _position_.
1. Else if _templateRemainder_ starts with *"$&amp;"*, then
  1. Let _advanceBy_ be 2.
  1. Let _replacement_ be _matched_.
1. Else if _templateRemainder_ starts with *"$'"* (0x0024 (DOLLAR SIGN) followed by 0x0027 (APOSTROPHE)), then
  1. Let _advanceBy_ be 2.
  …
…
1. Set _templateRemainder_ to the substring of _templateRemainder_ from _advanceBy_.
1. Set _result_ to the string-concatenation of _result_ and _replacement_.
```

spec.html
```diff
@@ -33975,7 +33975,7 @@ <h1>
             <dd>For the purposes of this abstract operation, a <em>decimal digit</em> is a code unit in the range 0x0030 (DIGIT ZERO) to 0x0039 (DIGIT NINE) inclusive.</dd>
           </dl>
           <emu-alg>
-            1. Let _stringLength_ be the number of code units in _str_.
+            1. Let _stringLength_ be the length of _str_.
```
bakkot, line 33978: There's a handful of other uses of "the number of code units in". If we're decided that we prefer "length" (which I personally do), we should fix those as well, in a separate PR.

bakkot (approved):

spec.html
gibson042: #2746

gibson042:
@bakkot Moved the «Replace "number of code units in [string]" with "length of"» changes into #2746.

michaelficarra:
@gibson042 Wanna rebase now that #2746 is merged?

gibson042:
@michaelficarra Rebased.

spec.html
```diff
@@ -33991,24 +33991,20 @@ <h1>
...
-                    1. Else if _index_ &le; the number of elements in _captures_, then
-                      1. Let _capture_ be _captures_[_index_ - 1].
-                      1. If _capture_ is *undefined*, then
-                        1. Let _refReplacement_ be the empty String.
-                      1. Else,
-                        1. Let _refReplacement_ be _capture_.
-                    1. Else,
-                      1. Let _refReplacement_ be _ref_.
+                1. If _templateRemainder_ starts with *"$"* followed by 2 or more decimal digits, let _digitCount_ be 2. Otherwise, let _digitCount_ be 1.
+                1. Let _ref_ be the substring of _templateRemainder_ from 0 to 1 + _digitCount_.
```
michaelficarra, line 33995: Can we define an alias for `1 + _digitCount_` so it reads more clearly?

gibson042: Yes, but the obvious name for that is _refLength_, which already has a Let step after this If...Else if...Else series. Should I adopt the suggestion of https://github.com/tc39/ecma262/pull/2739#discussion_r850845182 , using length rather than substring in every case?

michaelficarra: That WFM. @tc39/ecma262-editors?

spec.html
bakkot: There's a few different `Let refReplacement be ref.` steps, and it seems like it would be a shame to lose those.

spec.html
syg: No alias sgtm.

syg (approved):

PR #2746, opened 2022-04-19
Editorial: Use consistent phrasing for string length

gibson042:
As suggested at https://github.com/tc39/ecma262/pull/2739/files#r852477711


michaelficarra (approved):
From [6.1.4 The String Type](https://tc39.es/ecma262/#sec-ecmascript-language-types-string-type),

> The length of a String is the number of elements (i.e., 16-bit values) within it.

So this LGTM.

bakkot (approved):
I would prefer to include the remaining cases, which are currently in https://github.com/tc39/ecma262/pull/2739, so that each PR does one thing. Otherwise LGTM.

gibson042:
@bakkot Done.

PR #2749, opened 2022-04-20
Elision opt

xp44mm:
1. \`,\` Elision? -> Elision (14p)
2. append `</html>`


xp44mm:
> it would be necessary to modify the semantics as well.

I agree with you.

PR #2755, opened 2022-04-27
Editorial: Add missing preposition

gibson042:
Ref 58057150a5fe6dd2e08f7c83d0ba28e8165a9490

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33767,7 +33767,7 @@ THH:mm:ss.sss
             // &#x1E0D;&#x0307; LATIN SMALL LETTER D WITH DOT BELOW + COMBINING DOT ABOVE
             "\u1E0B\u0323".localeCompare("\u1E0D\u0307")
 
-            // &#x1100;&#x1161; HANGUL CHOSEONG KIYEOK + HANGUL JUNGSEONG A
+            // &#x1100;&#x1161; HANGUL CHOSEONG KIYEOK + HANGUL JUNGSEONG A vs.
             // &#xAC00; HANGUL SYLLABLE GA
             "\u1100\u1161".localeCompare("\uAC00")
           </code></pre>


spec.html
```diff
@@ -33767,7 +33767,7 @@ <h1>String.prototype.localeCompare ( _that_ [ , _reserved1_ [ , _reserved2_ ] ]
             // &#x1E0D;&#x0307; LATIN SMALL LETTER D WITH DOT BELOW + COMBINING DOT ABOVE
             "\u1E0B\u0323".localeCompare("\u1E0D\u0307")
 
-            // &#x1100;&#x1161; HANGUL CHOSEONG KIYEOK + HANGUL JUNGSEONG A
+            // &#x1100;&#x1161; HANGUL CHOSEONG KIYEOK + HANGUL JUNGSEONG A vs.
             // &#xAC00; HANGUL SYLLABLE GA
```
ljharb, line 33771:
```suggestion
            // &#x1100;&#x1161; HANGUL CHOSEONG KIYEOK + HANGUL JUNGSEONG A vs. &#xAC00; HANGUL SYLLABLE GA
```
any reason to limit the line length?

jmdyck: All the other examples in this code chunk have the two descriptions on separate lines.

gibson042: > any reason to limit the line length?

Yes, so the code block renders correctly as a PDF without content being hidden by horizontal scrollbars.

ljharb: Seems like something print CSS and soft wrapping would handle.

michaelficarra (approved):

spec.html
ljharb: Then we should fix that holistically instead of doing it ad hoc.

gibson042: I'm not opposed, but this PR is literally just adding four missing characters. But even with such formatting, for https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.localecompare specifically I would still prefer consecutive single-line comments for vertical alignment between the respective glyphs and code point decompositions.

ljharb: That the first of these lines has two glyphs on it means the alignment isn’t achieved tho.

gibson042: No, those two _code points_ (U+1100, U+1161); constitute one _glyph_ ("가"): https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.localecompare

bakkot (approved):

PR #2759, opened 2022-05-01
Editorial: Correct notes in Create{Method,NonEnumerableData}Property 

linusg:
Details in the commit messages.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6325,11 +6325,11 @@
       <emu-alg>
         1. Assert: _O_ is an ordinary, extensible object with no non-configurable properties.
         1. Let _newDesc_ be the PropertyDescriptor { [[Value]]: _V_, [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.
-        1. Perform ! _O_.[[DefineOwnProperty]](_P_, _newDesc_).
+        1. Perform ! DefinePropertyOrThrow(_O_, _P_, _newDesc_).
         1. Return ~unused~.
       </emu-alg>
       <emu-note>
-        <p>This abstract operation creates a property whose attributes are set to the same defaults used for built-in methods and methods defined using class declaration syntax. Normally, the property will not already exist. If it does exist and is not configurable or if _O_ is not extensible, [[DefineOwnProperty]] will return *false*.</p>
+        <p>This abstract operation creates a property whose attributes are set to the same defaults used for built-in methods and methods defined using class declaration syntax. Normally, the property will not already exist. If it does exist, DefinePropertyOrThrow is guaranteed to complete normally.</p>
       </emu-note>
     </emu-clause>
 
@@ -6374,7 +6374,7 @@
         1. Return ~unused~.
       </emu-alg>
       <emu-note>
-        <p>This abstract operation creates a property whose attributes are set to the same defaults used for properties created by the ECMAScript language assignment operator except it is not enumerable. Normally, the property will not already exist. If it does exist and is not configurable or if _O_ is not extensible, [[DefineOwnProperty]] will return *false* causing this operation to throw a *TypeError* exception.</p>
+        <p>This abstract operation creates a property whose attributes are set to the same defaults used for properties created by the ECMAScript language assignment operator except it is not enumerable. Normally, the property will not already exist. If it does exist, DefinePropertyOrThrow is guaranteed to complete normally.</p>
       </emu-note>
     </emu-clause>
 


arhadthedev:
Should the order of the AOs be returned back (`CreateDataProperty` then `CreateNonEnumerableDataProperty`)? The original looks alphabetically sorted.

linusg:
The current order is

```
7.3.5 CreateDataProperty ( O, P, V )
7.3.6 CreateMethodProperty ( O, P, V )
7.3.7 CreateDataPropertyOrThrow ( O, P, V )
7.3.8 CreateNonEnumerableDataPropertyOrThrow ( O, P, V )
7.3.9 DefinePropertyOrThrow ( O, P, desc )
7.3.10 DeletePropertyOrThrow ( O, P )
```

...i.e. not alphabetical. My intention was to keep "...OrThrow" last.

linusg:
The current order on `main` is

```
7.3.5 CreateDataProperty ( O, P, V )
7.3.6 CreateMethodProperty ( O, P, V )
7.3.7 CreateDataPropertyOrThrow ( O, P, V )
7.3.8 CreateNonEnumerableDataPropertyOrThrow ( O, P, V )
7.3.9 DefinePropertyOrThrow ( O, P, desc )
7.3.10 DeletePropertyOrThrow ( O, P )
```

...i.e. not alphabetical. My intention was to keep "...OrThrow" last.

spec.html
```diff
@@ -6348,52 +6348,58 @@ <h1>
...
       <dl class="header">
         <dt>description</dt>
-        <dd>It is used to create a new own property of an object. It throws a *TypeError* exception if the requested property update cannot be performed.</dd>
+        <dd>It is used to create a new non-enumerable own property of an ordinary object.</dd>
       </dl>
       <emu-alg>
-        1. Let _success_ be ? CreateDataProperty(_O_, _P_, _V_).
-        1. If _success_ is *false*, throw a *TypeError* exception.
-        1. Return _success_.
+        1. Assert: _O_ is an ordinary, extensible object with no non-configurable properties.
```
ljharb, line 6371: How can this assertion be made? This spec isn’t the only thing that calls these AOs.

arhadthedev:
@linusg Now I see, thank you. I didn't look through the whole section initially.

spec.html
```diff
@@ -6367,12 +6370,15 @@ <h1>
       <emu-alg>
         1. Assert: _O_ is an ordinary, extensible object with no non-configurable properties.
         1. Let _newDesc_ be the PropertyDescriptor { [[Value]]: _V_, [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.
-        1. Perform ! DefinePropertyOrThrow(_O_, _P_, _newDesc_).
+        1. Perform ! _O_.[[DefineOwnProperty]](_P_, _newDesc_).
```
ljharb, line 6373: I find this reduces clarity; calling an AO a better than calling an internal method.

bakkot:
Thanks for the PR. I don't think the renaming is an improvement; the "or throw" is intentional, for consistency with `CreateDataPropertyOrThrow`. We do need to clean these up (cf https://github.com/tc39/ecma262/pull/2528) but I don't think we want to make that particular change.

The refactoring to call `[[DefineOwnProperty]]` directly also does not seem like an improvement improvement, since it now ignores the return value instead of checking it. `! DefinePropertyOrThrow()` is the pattern we use throughout the specification when defining a non-accessor property in a way where we know that will succeed.

Second commit seems fine, if you want to reduce this PR to just that.

spec.html
linusg: I didn't make that change, it only shows as new from the re-ordering. This is already in the current spec.

linusg:
> the "or throw" is intentional, for consistency with `CreateDataPropertyOrThrow`.

`CreateDataPropertyOrThrow` actually *can* throw. `CreateNonEnumerableDataPropertyOrThrow` *can't* as written, so the name is super misleading. It's functionally equivalent to `CreateMethodProperty`, which isn't named `CreateMethodPropertyOrThrow` either. I don't think consistency is applicable there?

> The refactoring to call [[DefineOwnProperty]] directly also does not seem like an improvement, since it now ignores the return value instead of checking it.

That however was purely for consistency, it's the same way `CreateMethodProperty` works. Would you accept a change to update that to call the `DefinePropertyOrThrow` AO instead of directly the internal method?

> ! DefinePropertyOrThrow() is the pattern we use throughout the specification when defining a non-accessor property in a way where we know that will succeed.

I know, and it still does that (`! O.[[DefineOwnProperty]](...)`). The only difference is less indirection.

> Second commit seems fine, if you want to reduce this PR to just that.

Happy to do that, but just want to confirm that there's no misunderstanding here. Keeping the current name seems like a rather unfortunate decision, and the PR you mentioned doesn't seem to cover that either.

spec.html
linusg: Thanks, I'll keep that in mind :+1:

bakkot:
>  I don't think consistency is applicable there?

I agree there's already inconsistency, but I'm not a fan of this particular fix for the inconsistency.

> Would you accept a change to update that to call the `DefinePropertyOrThrow` AO instead of directly the internal method?

Sure, that sounds good.

> I know, and it still does that (`! O.[[DefineOwnProperty]](...)`). The only difference is less indirection.

No, `! DefinePropertyOrThrow(O, ...)` is meaningfully different than `! O.[[DefineOwnProperty]](...)`: the former contains an assertion that creation of the property succeeds, whereas the latter does not. (`[[DefineOwnProperty]]` returns `*false*` on failure rather than throwing.)

> Keeping the current name seems like a rather unfortunate decision, and the PR you mentioned doesn't seem to cover that either.

The PR I mentioned isn't what we actually want to do, I just linked it to point to where (some) discussion has happened previously. Editors are agreed we need to rename and clean up these AOs but have not yet actually figured out what we want that to look like.

linusg:
Thanks for the feedback all, updated!

bakkot (approved):

spec.html
```diff
@@ -6344,11 +6344,11 @@ <h1>
       <emu-alg>
         1. Assert: _O_ is an ordinary, extensible object with no non-configurable properties.
         1. Let _newDesc_ be the PropertyDescriptor { [[Value]]: _V_, [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.
-        1. Perform ! _O_.[[DefineOwnProperty]](_P_, _newDesc_).
+        1. Perform ! DefinePropertyOrThrow(_O_, _P_, _newDesc_).
         1. Return ~unused~.
       </emu-alg>
       <emu-note>
-        <p>This abstract operation creates a property whose attributes are set to the same defaults used for built-in methods and methods defined using class declaration syntax. Normally, the property will not already exist. If it does exist and is not configurable or if _O_ is not extensible, [[DefineOwnProperty]] will return *false*.</p>
+        <p>This abstract operation creates a property whose attributes are set to the same defaults used for built-in methods and methods defined using class declaration syntax. Normally, the property will not already exist. If it does exist, [[DefineOwnProperty]] is guaranteed to complete normally and return *true* within this specification.</p>
```
michaelficarra, line 6351:
```suggestion
        <p>This abstract operation creates a property whose attributes are set to the same defaults used for built-in methods and methods defined using class declaration syntax. Normally, the property will not already exist. If it does exist, DefinePropertyOrThrow is guaranteed to complete normally.</p>
```

spec.html
```diff
@@ -6375,7 +6375,7 @@ <h1>
         1. Return ~unused~.
       </emu-alg>
       <emu-note>
-        <p>This abstract operation creates a property whose attributes are set to the same defaults used for properties created by the ECMAScript language assignment operator except it is not enumerable. Normally, the property will not already exist. If it does exist and is not configurable or if _O_ is not extensible, [[DefineOwnProperty]] will return *false* causing this operation to throw a *TypeError* exception.</p>
+        <p>This abstract operation creates a property whose attributes are set to the same defaults used for properties created by the ECMAScript language assignment operator except it is not enumerable. Normally, the property will not already exist. If it does exist, [[DefineOwnProperty]] is guaranteed to complete normally and return *true* within this specification.</p>
```
michaelficarra, line 6378: Same here, please.

```suggestion
        <p>This abstract operation creates a property whose attributes are set to the same defaults used for properties created by the ECMAScript language assignment operator except it is not enumerable. Normally, the property will not already exist. If it does exist, DefinePropertyOrThrow is guaranteed to complete normally.</p>
```

linusg: oops, fixed!

michaelficarra (approved):

linusg:
Who do I ping to click the merge button? :^)

syg (approved):

PR #2764, opened 2022-05-02
Editorial: Mark concrete method calls of Environment Records with ?/!

linusg:


syg (approved):
This seems correct to me. Huh, I wonder why these got omitted in the big completion reform PR.

bakkot (approved):

bakkot:
> This seems correct to me. Huh, I wonder why these got omitted in the big completion reform PR.

Concrete records are trickier to typecheck, basically, so we did them by hand. Some of them we initially had as never-throwing but later changed, and I guess we didn't manage to fix all of the callsites when we did that.

PR #2765, opened 2022-05-02
Editorial: Fix descriptor kind check in ValidateAndApplyPropertyDescriptor

### LINKED ISSUE #2761: Invalid assumption in ValidateAndApplyPropertyDescriptor
Assuming `ValidateAndApplyPropertyDescriptor` is called with the following property descriptors for `Desc` and `current`:

```
Desc = { [[Configurable]]: false }
current = { [[Get]]: <some function>, [[Set]]: undefined, [[Enumerable]]: true, [[Configurable]]: false }
```

as it would happen in the second `freeze` call in this piece of code:

```js
Object.freeze(Object.freeze({ get foo() {} }));
```

It'll then perform the following steps:

```
   [...]
✅ 5. If current.[[Configurable]] is false, then
❌    a. If Desc has a [[Configurable]] field and Desc.[[Configurable]] is true, return false.
❌    b. If Desc has an [[Enumerable]] field and SameValue(Desc.[[Enumerable]], current.[[Enumerable]]) is false, return false.
❌    c. If IsGenericDescriptor(Desc) is false and SameValue(IsAccessorDescriptor(Desc), IsAccessorDescriptor(current)) is false, return false.
❌    d. If IsAccessorDescriptor(Desc) is true, then
         [...]
💥    e. Else if current.[[Writable]] is false, then
```

- `Desc` neither has `[[Configurable]]` set to true, nor has it an `[[Enumerable]]` field (steps a and b)
- `SameValue(IsAccessorDescriptor(Desc), IsAccessorDescriptor(current))` is false but `IsGenericDescriptor(Desc)` is true (step c)
- `IsAccessorDescriptor(Desc)` is false (step d)
- Step e is reached without ever confirming that `current` is a data property descriptor, and accessing the `[[Writable]]` field is invalid

This is likely a regression in https://github.com/tc39/ecma262/commit/b9efa97, it worked with the old algorithm steps. (cc @gibson042)

linusg:
Closes #2761.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12721,7 +12721,7 @@
             1. If _Desc_ has a [[Configurable]] field and _Desc_.[[Configurable]] is *true*, return *false*.
             1. If _Desc_ has an [[Enumerable]] field and SameValue(_Desc_.[[Enumerable]], _current_.[[Enumerable]]) is *false*, return *false*.
             1. If IsGenericDescriptor(_Desc_) is *false* and SameValue(IsAccessorDescriptor(_Desc_), IsAccessorDescriptor(_current_)) is *false*, return *false*.
-            1. If IsAccessorDescriptor(_Desc_) is *true*, then
+            1. If IsAccessorDescriptor(_current_) is *true*, then
               1. If _Desc_ has a [[Get]] field and SameValue(_Desc_.[[Get]], _current_.[[Get]]) is *false*, return *false*.
               1. If _Desc_ has a [[Set]] field and SameValue(_Desc_.[[Set]], _current_.[[Set]]) is *false*, return *false*.
             1. Else if _current_.[[Writable]] is *false*, then


gibson042 (approved):

ljharb:
Confirming this is purely editorial in es-abstract.

bakkot (approved):

michaelficarra:
Should this be marked as normative?

bakkot:
> Should this be marked as normative?

The change which introduced this bug wasn't, so it would be kind of weird for this one to be.

ljharb:
@linusg would you please sign the [IPR form](https://tc39.es/agreements/contributor/)?

linusg:
@ljharb done! Sorry, I thought I already had, but that was a separate one for test262.

syg (approved):

bakkot:
@ljharb At the editor call today Michael agreed with calling this editorial; was that the only thing you wanted to discuss or was there something else?

ljharb:
@bakkot nope, it was just a review ping :-) i'll land shortly.

PR #2768, opened 2022-05-04
Editorial: Declare iteration aliases as close to iterations as practical

gibson042:
As observed at https://github.com/tc39/ecma262/pull/2766#discussion_r865008959 , it is very common for an iteration alias (i.e., a value that changes during iteration and effects its conclusion) to be declared immediately before its corresponding iteration. This updates a few deviations to conform more strongly with that convention.


michaelficarra (approved):
Sure.

bakkot (approved):

PR #2772, opened 2022-05-09
Remove Job from Promise Resolve Functions

### LINKED ISSUE #2770: Direct return from async function takes too many ticks
**Description:**

In #1250, we reduced the number of ticks that awaiting a native promise value (from 3 ticks to 1 tick). However, returning from an async function can still take 3 ticks (somehow, I don't really understand it).

```js
async function returnDirectPrimitive() {
  return 1;
}

async function returnDirectPromisePrimitive() {
  return Promise.resolve(1);
}

async function returnAwaitPrimitive() {
  return await 1;
}

async function returnAwaitPromisePrimitive() {
  return await Promise.resolve(1);
}

async function test(fn) {
  let resolved = false;
  let count = 0;
  const done = fn().then(() => {
    resolved = true;
  });

  while (!resolved) {
    await 0;
    count++;
  }

  print(`${fn.name} took ${count} ticks`);
}

async function tests() {
  await test(returnDirectPrimitive);
  await test(returnAwaitPrimitive);

  await test(returnDirectPromisePrimitive);
  await test(returnAwaitPromisePrimitive);
}

tests();
```

**eshost Output:**

```
#### engine262, Moddable XS, QuickJS, SpiderMonkey, V8
returnDirectPrimitive took 1 ticks
returnAwaitPrimitive took 2 ticks
returnDirectPromisePrimitive took 3 ticks
returnAwaitPromisePrimitive took 2 ticks

#### Hermes
returnDirectPrimitive took 1 ticks
returnAwaitPrimitive took 2 ticks
returnDirectPromisePrimitive took 1 ticks
returnAwaitPromisePrimitive took 2 ticks

#### JavaScriptCore
returnDirectPrimitive took 1 ticks
returnAwaitPrimitive took 2 ticks
returnDirectPromisePrimitive took 3 ticks
returnAwaitPromisePrimitive took 3 ticks
```

(I believe that Hermes and JSC are incorrect here, and the other browsers are correct.)

✅ The "returnDirectPrimitive" branch, we should only be waiting 1 tick for the async function to assimilate the returned primitive.

✅ The "returnAwaitPrimitive" branch, we should wait 1 tick to extract the value out of the implicitly created promise of `1`, then 1 tick for the async function to assimilate the extracted value.

❌ The "returnDirectPromisePrimitive" branch. There's disagreement here between impls, and I don't think any of the them are the correct behavior. My thinking is that this should be just like "returnAwaitPrimitive", where we wait 1 tick to extract the value from the already resolved promise, and 1 tick for the async function to assimilate the extracted value. But somehow, it seems we should be waiting 3 ticks?! Is this creating a promise, wrapping it with another promise, then assimilating that promise (similar to #1250)? If so, we should fix this to take 2 ticks.

✅ Finally we have the "returnAwaitPromisePrimitive", and this reenforces my belief that "returnDirectPromisePrimitive" is incorrect. In this branch, we wait 1 tick to extract the value from already resolved promise, and 1 tick to assimilate the extracted value.

jridgewell:
This aligns the behavior of handling thenables with the Promises A+ spec, which says when the resolution value is a thenable, you're supposed to synchronously call `thenable.then(onRes, onRej)` ([Step 2.3.3.3][call-then]).

Given the example code:

```javascript
new Promise(resolve => {
  resolve(thenable)
});
```

The current behavior requires 2 ticks for the outer promise to fully resolve. One tick is created by the Promise Resolving Functions (and is removed in this PR), and one tick is created by the wrapping of the `onRes`/`onRej` callbacks passed to [`Promise.p.then`][then]. This made it noticeably slower to resolve with a thenable than to invoke the thenable's `.then` directly: `thenable.then(onRes, onRej)` only requires a single tick (for the wrapped `onRes`/`onRej`).

~~With this change, we could revert #1250 without slowing down `Await`.~~ Partially, I didn't see the 2nd promise allocation. This PR makes the pre-#1250 run in 2 ticks instead of 3. We still want 1 tick.

Fixes #2770.

[call-then]: https://promisesaplus.com/#point-56
[then]: https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-performpromisethen

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -43600,7 +43600,7 @@ THH:mm:ss.sss
             1. Let _resolvingFunctions_ be CreateResolvingFunctions(_promise_).
             1. Let _thenCallResult_ be Completion(Call(_then_, _resolution_, &laquo; _resolvingFunctions_.[[Resolve]], _resolvingFunctions_.[[Reject]] &raquo;)).
             1. If _thenCallResult_ is an abrupt completion, then
-              1. Call(_resolvingFunctions_.[[Reject]], *undefined*, &laquo; _thenCallResult_.[[Value]] &raquo;).
+              1. Perform ! Call(_resolvingFunctions_.[[Reject]], *undefined*, &laquo; _thenCallResult_.[[Value]] &raquo;).
             1. Return *undefined*.
           </emu-alg>
           <p>The *"length"* property of a promise resolve function is *1*<sub>𝔽</sub>.</p>
--- a/spec.html
+++ b/spec.html
@@ -12126,7 +12126,7 @@
       </ul>
 
       <emu-note>
-        <p>The _realm_ for Jobs returned by NewPromiseResolveThenableJob is usually the result of calling GetFunctionRealm on the _then_ function object. The _realm_ for Jobs returned by NewPromiseReactionJob is usually the result of calling GetFunctionRealm on the handler if the handler is not *undefined*. If the handler is *undefined*, _realm_ is *null*. For both kinds of Jobs, when GetFunctionRealm completes abnormally (i.e. called on a revoked Proxy), _realm_ is the current Realm at the time of the GetFunctionRealm call. When the _realm_ is *null*, no user ECMAScript code will be evaluated and no new ECMAScript objects (e.g. Error objects) will be created. The WHATWG HTML specification (<a href="https://html.spec.whatwg.org/">https://html.spec.whatwg.org/</a>), for example, uses _realm_ to check for the ability to run script and for the <a href="https://html.spec.whatwg.org/#entry">entry</a> concept.</p>
+        <p>The _realm_ for Jobs returned by NewPromiseReactionJob is usually the result of calling GetFunctionRealm on the handler if the handler is not *undefined*. If the handler is *undefined*, _realm_ is *null*. When GetFunctionRealm completes abnormally (i.e. called on a revoked Proxy), _realm_ is the current Realm at the time of the GetFunctionRealm call. When the _realm_ is *null*, no user ECMAScript code will be evaluated and no new ECMAScript objects (e.g. Error objects) will be created. The WHATWG HTML specification (<a href="https://html.spec.whatwg.org/">https://html.spec.whatwg.org/</a>), for example, uses _realm_ to check for the ability to run script and for the <a href="https://html.spec.whatwg.org/#entry">entry</a> concept.</p>
       </emu-note>
     </emu-clause>
   </emu-clause>
@@ -43597,9 +43597,10 @@ THH:mm:ss.sss
             1. If IsCallable(_thenAction_) is *false*, then
               1. Perform FulfillPromise(_promise_, _resolution_).
               1. Return *undefined*.
-            1. Let _thenJobCallback_ be HostMakeJobCallback(_thenAction_).
-            1. Let _job_ be NewPromiseResolveThenableJob(_promise_, _resolution_, _thenJobCallback_).
-            1. Perform HostEnqueuePromiseJob(_job_.[[Job]], _job_.[[Realm]]).
+            1. Let _resolvingFunctions_ be CreateResolvingFunctions(_promise_).
+            1. Let _thenCallResult_ be Completion(Call(_then_, _resolution_, &laquo; _resolvingFunctions_.[[Resolve]], _resolvingFunctions_.[[Reject]] &raquo;)).
+            1. If _thenCallResult_ is an abrupt completion, then
+              1. Call(_resolvingFunctions_.[[Reject]], *undefined*, &laquo; _thenCallResult_.[[Value]] &raquo;).
             1. Return *undefined*.
           </emu-alg>
           <p>The *"length"* property of a promise resolve function is *1*<sub>𝔽</sub>.</p>
@@ -43792,34 +43793,6 @@ THH:mm:ss.sss
             1. NOTE: _handlerRealm_ is never *null* unless the handler is *undefined*. When the handler is a revoked Proxy and no ECMAScript code runs, _handlerRealm_ is used to create error objects.
           1. Return the Record { [[Job]]: _job_, [[Realm]]: _handlerRealm_ }.
         </emu-alg>
-      </emu-clause>
-
-      <emu-clause id="sec-newpromiseresolvethenablejob" type="abstract operation" oldids="sec-promiseresolvethenablejob">
-        <h1>
-          NewPromiseResolveThenableJob (
-            _promiseToResolve_: unknown,
-            _thenable_: unknown,
-            _then_: unknown,
-          ): a Record with fields [[Job]] (a Job Abstract Closure) and [[Realm]] (a Realm Record)
-        </h1>
-        <dl class="header">
-        </dl>
-        <emu-alg>
-          1. Let _job_ be a new Job Abstract Closure with no parameters that captures _promiseToResolve_, _thenable_, and _then_ and performs the following steps when called:
-            1. Let _resolvingFunctions_ be CreateResolvingFunctions(_promiseToResolve_).
-            1. Let _thenCallResult_ be Completion(HostCallJobCallback(_then_, _thenable_, &laquo; _resolvingFunctions_.[[Resolve]], _resolvingFunctions_.[[Reject]] &raquo;)).
-            1. If _thenCallResult_ is an abrupt completion, then
-              1. Return ? Call(_resolvingFunctions_.[[Reject]], *undefined*, &laquo; _thenCallResult_.[[Value]] &raquo;).
-            1. Return ? _thenCallResult_.
-          1. Let _getThenRealmResult_ be Completion(GetFunctionRealm(_then_.[[Callback]])).
-          1. If _getThenRealmResult_ is a normal completion, let _thenRealm_ be _getThenRealmResult_.[[Value]].
-          1. Else, let _thenRealm_ be the current Realm Record.
-          1. NOTE: _thenRealm_ is never *null*. When _then_.[[Callback]] is a revoked Proxy and no code runs, _thenRealm_ is used to create error objects.
-          1. Return the Record { [[Job]]: _job_, [[Realm]]: _thenRealm_ }.
-        </emu-alg>
-        <emu-note>
-          <p>This Job uses the supplied thenable and its `then` method to resolve the given promise. This process must take place as a Job to ensure that the evaluation of the `then` method occurs after evaluation of any surrounding code has completed.</p>
-        </emu-note>
       </emu-clause>
     </emu-clause>
 


ljharb:
What are the observable changes as a result of this PR, besides "number of ticks greater than zero"?

ljharb:
Also, it's worth investigating how this would impact websites deployed with es6-shim, core-js, and other Promise shims - if browsers start having different behavior here, it might cause the shims to kick in when they otherwise wouldn't.

mhofman:
Also very curious about the web compat aspect of this change. I don't doubt some code out there now relies on `then` jobs being in a new tick.

spec.html
```diff
@@ -43597,9 +43597,10 @@ <h1>Promise Resolve Functions</h1>
             1. If IsCallable(_thenAction_) is *false*, then
               1. Perform FulfillPromise(_promise_, _resolution_).
               1. Return *undefined*.
-            1. Let _thenJobCallback_ be HostMakeJobCallback(_thenAction_).
-            1. Let _job_ be NewPromiseResolveThenableJob(_promise_, _resolution_, _thenJobCallback_).
-            1. Perform HostEnqueuePromiseJob(_job_.[[Job]], _job_.[[Realm]]).
+            1. Let _resolvingFunctions_ be CreateResolvingFunctions(_promise_).
+            1. Let _thenCallResult_ be Completion(Call(_then_, _resolution_, &laquo; _resolvingFunctions_.[[Resolve]], _resolvingFunctions_.[[Reject]] &raquo;)).
+            1. If _thenCallResult_ is an abrupt completion, then
+              1. ! Call(_resolvingFunctions_.[[Reject]], *undefined*, &laquo; _thenCallResult_.[[Value]] &raquo;).
```
ljharb, line 43603:
```suggestion
              1. Perform ! Call(_resolvingFunctions_.[[Reject]], *undefined*, &laquo; _thenCallResult_.[[Value]] &raquo;).
```

jridgewell:
> What are the observable changes as a result of this PR, besides "number of ticks greater than zero"?

**Case 1:** Observable Change

```js
new Promise(res => {
  let sync = false;

  resolve({
    then(onRes, onRej) {
      sync = true;
      onRes();
    }
  });

  console.assert(sync === true);
})
````

Before, `sync === false`. Now, it'll be `true`. There are no other observable changes. If there's were an abrupt completion, the outer promise would behave the same.

- - -

**Case 2:** Observable Change

```js
const parent = Promise.resolve();

parent.then(() => {
  let sync = false;
  parent.then(() => {
    console.assert(sync === true);
  });

  return {
    then(onRes) {
      sync = true;
      onRes();
    }
  };
});
````

Same deal, now `sync === true`. There are no observable changes to the outer promise.

There are no other orderings of `parent.then()` which would be observable. Because you have to `reutrn thenable`, it's not possible to observe that its `then` will be sync called after you, well, returned the thenable. If you added more `parent.then(() => sync)` calls in any other location, you could only observe `sync === false`. It is only `parent.then`s called in that exact spot before we return the thenable that you can observe the `then` is sync called.

File: spec.html (hunk 1 above)
Comment (jridgewell) [reply spec.html:43603]: I keep getting "This diff has recently been updated. Refresh and try again.", except refreshing doesn't help. I applied locally and pushed.

**Comment by jridgewell** [comment]
> Also, it's worth investigating how this would impact websites deployed with es6-shim, core-js, and other Promise shims…

Both es6-shim and core-js behave with the current semantics. I do not know if they actually test for this behavior, but it'd be ill advised to overwrite the native Promise impl if they did, given that Promise is used all over, and native APIs/async functions will never return the userland impl.

I imagine most other Promises impls are using A+ semantics, because it actually has a portable test suite and isn't written in spec language.

**Comment by mhofman** [comment]
For what it's worth, this change would break my "hack" to [detect promise adoption/redirect](https://gist.github.com/mhofman/331a8b0dddbd7928a12327706faa89c4) through node's `async_hooks`. I heavily relied on the fact that the `then` would be called on a clean stack to differentiate from other async hooks triggers. Definitely not a blocker as I've been meaning to request a better hook for this in v8 (and have thoughts on a less privileged hook).

**Comment by raulsebastianmihaila** [comment]
Not only is the proposed change a breaking change but it is also counter-intuitive. `then` in my mind means 'in a future tick' so the function shouldn't be called synchronously.

**Comment by nicolo-ribaudo** [comment]
@raulsebastianmihaila As far as I understand, with this change the function called to then is still called asynchronously. The difference is `.then` itself is called synchronously (i.e. it behave like `thenable.then(fn)` instead of `Promise.resolve(thenable).then(fn)`).

**Comment by raulsebastianmihaila** [comment]
> @raulsebastianmihaila As far as I understand, with this change the function called to then is still called asynchronously. The difference is only that `.then` itself is called synchronously (i.e. it behave like `thenable.then(fn)` instead of `Promise.resolve().then(() => thenable.then(fn))`).

This thread starts by saying
> you're supposed to synchronously call thenable.then(onRes, onRej)

**Comment by bakkot** [comment]
Yes. My point is that as a _producer_ of thenables you should not design on the assumption that `then` is only called asychronously, because this is not true; rather, your `then` should always schedule work to be done asynchronously. And if producers of thenables adhere to this principle, then consumers of thenables will not need to know whether `await` invokes `then` itself synchronously or asychronously.

**Comment by raulsebastianmihaila** [comment]
Yes, and my point is that fortunately you don't have to care about that because the promises behavior is consistent with regard to how it handles userland code.

**Comment by bakkot** [comment]
It does not make changes relevant to producing. Producers already have to be aware that `then` may be invoked synchronously or asynchronously. This PR does not change that.

**Comment by raulsebastianmihaila** [comment]
> Producers already have to be aware that then may be invoked synchronously or asynchronously.

Unfortunately I can not agree with that. The reason being that producers can rely on how promises work and this change will break such code. Take the small example I gave previously for executing code in a future tick. Even if there is universal agreement with your view on how thenables should be written in general, there are certainly exceptions where such snippets are used in practice as workarounds in codebases that are not well taken care of. I have seen it in real code (way before `queueMicrotask` was invented, given promises became standard 7 years ago). It's not that one would disagree with you on what the best practices should be for creating thenables. It's just that sometimes the creation of thenables is not the main focus, but instead they are created as part of a workaround or something like that which relies on the standard behavior of promises.

**Comment by raulsebastianmihaila** [comment]
This change does complicate the mental model because it introduces a new rule to follow. But I think this is less relevant nowadays than the other things I was discussing (like this being a breaking change, which is precisely what I started with).

**Comment by bakkot** [comment]
It does not introduce a new rule to follow.

**Comment by jridgewell** [comment]
This discussion is getting off track, please take it to matrix.

**Comment by bakkot** [comment]
Dug up where this change was introduced: https://github.com/domenic/promises-unwrapping/issues/105

The concern was specifically that `Promise.resolve(foo)` could break invariants in the surrounding code. That's already out the window because of the observable `Get(x, "constructor")`, but I guess the idea was that this wouldn't matter in practice?

If we don't want to outright revert the above PR, we could instead add a fast-path in Promise Resolve Functions which says that if `thenAction` is the built-in `Promise.prototype.then` then it's invoked immediately rather than on the next tick (with custom exception handling). That avoids the problem in the linked issue and also keeps the old behavior for custom `thenables` which @raulsebastianmihaila is worried about. And it doesn't invoke any user code synchronously which was not already invoked synchronously, because the real `Promise.prototype.then` never invokes any user code synchronously - it would just happen after a single tick instead of after two.

The cost is that this is a place thenables take an extra tick vs userland promises, but since that's already true for `await`, I think that's fine.

**Comment by ljharb** [comment]
That seems like a pretty reasonable/safe/best-of-both-worlds compromise.

**Comment by jridgewell** [comment]
> For what it's worth, this change would partially break my "hack" to [detect promise adoption/redirect](https://gist.github.com/mhofman/331a8b0dddbd7928a12327706faa89c4) through node's `async_hooks`.
> — https://github.com/tc39/ecma262/pull/2772#issuecomment-1121836146

I think is already broken for async/await since #1250. Doing `await nativePromiseWithMonkeyedThen` does not call the monkey-patched `then`, it uses the internal slots to access data directly.

- - -

> The concern was specifically that `Promise.resolve(foo)` could break invariants in the surrounding code. That's already out the window because of the synchronous `Get(x, "constructor")` in [PromiseResolve](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise-resolve) and `Get(resolution, "then")` in [Promise Resolve Functions](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise-resolve-functions), but I guess the idea was that this wouldn't matter in practice?
> — https://github.com/tc39/ecma262/pull/2772#issuecomment-1122816364


I found a few more comments about untrusted code in the repo:
1. https://github.com/domenic/promises-unwrapping/issues/105#issuecomment-42752199 (the linked issue)
2. https://github.com/domenic/promises-unwrapping/issues/42#issuecomment-25933838
3. https://github.com/domenic/promises-unwrapping/issues/65#issuecomment-27442347

In particular, it seems to boil down to:

```js
const cast = Promise.resolve(untrustedThenable);
cast.then(untrustedFulfilled, untrustedRejected);
```

The desire is that at no point during this tick should untrusted code be run. But as https://github.com/tc39/ecma262/pull/2772#issuecomment-1122816364 points out, that's been broken for a while because the `untrustedThenable`'s `.constructor` and `.then` are both accessed during this tick. It seems to me that invoking the `then` wouldn't be that bad.

Note that during the time these comments were made, the `new Promise(executor)` [wasn't supposed to invoke the `executor`](https://github.com/promises-aplus/constructor-spec/issues/17) during this tick either, but [that was changed too](https://github.com/promises-aplus/constructor-spec/issues/17#issuecomment-13344256). The security guarantees softened to accommodate a more consistent API. We could decided that they should soften again, given they're already soft and it'll improve speed in both promise and async/await codebases.

Also, the `untrustedFulfilled` and `untrustedRejected` will still wait till the next tick to be invoked with the proposed changes. Casting an untrusted (and possibly badly behaving) thenable into a native promise will guarantee that `cast.then` calls its callbacks after a tick, so we don't have any zalgo issues.

- - -

PS. `Promise.resolve` has never been a good security guarantee. The following has been exploitable since ES2015:

```js
const p = Promise.resolve();
p.then = () => {
  alert('Gotcha!');
};

const cast = Promise.resolve(p);
p.then(); // Gotcha!
```

mhofman:
> I think is already broken for async/await since #1250. Doing `await nativePromiseWithMonkeyedThen` does not call the monkey-patched `then`, it uses the internal slots to access data directly.

Thankfully #1250 does not impact my use case since I only care about promises which are adopted by a user reachable promise, like as returned by an async function.

> That seems like a pretty reasonable/safe/best-of-both-worlds compromise.

Agreed. That compromise also wouldn't break my fringe scenario, but more importantly, the more or less legitimate assumptions that custom `then` execute on a new tick / clean stack.

> The desire is that at no point during this tick should untrusted code be run. But as [#2772 (comment)](https://github.com/tc39/ecma262/pull/2772#issuecomment-1122816364) points out, that's been broken for a while because the `untrustedThenable`'s `.constructor` and `.then` are both accessed during this tick. It seems to me that invoking the `then` wouldn't be that bad.

Correct, it's already broken, but only in the case of evil promises using get traps. I believe the argument here is that it's a lot easier to write code that assumes the `then` is called in a new tick. It also would be a much harder to observe change if we did `SameValue(thenValue, %Promise.prototype.then%)`, while reaping most of the benefits.

> PS. `Promise.resolve` has never been a good security guarantee. The following has been exploitable since ES2015:

Right, and that's why I would really like if we could expose a clean [`IsPromise`](https://tc39.es/ecma262/#sec-ispromise) to userland as that would make it possible to be defensive against evil `get constructor` and `get then` shenanigans.

bakkot:
Another alternative fast-path would be to use the same test as `Promise.resolve` and `await`: before reading `resolution.then`, first test `IsPromise(resolution) && resolution.constructor === Promise`, and if that passes then invoke the built-in `PerformPromiseThen` machinery on `resolution` directly, bypassing any check of `.then`.

This would have the advantage of skipping two observable lookups (of `then` and `Symbol.species`) in the common case of a native Promise. The `constructor` lookup is already present for native Promises because [`Promise.prototype.then`](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise.prototype.then) calls [SpeciesConstructor](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-speciesconstructor) which does a lookup of `constructor`, so that lookup just gets moved around. It does incur an additional lookup (of `constructor`) for Promise subclasses.

jridgewell:
> before reading `resolution.then`, first test `IsPromise(resolution) && resolution.constructor === Promise`, and if that passes then invoke the built-in `PerformPromiseThen` machinery on `resolution` directly, bypassing any check of `.then`.

Wouldn't this immediately call `PerformPromiseThen` with the now-verified Promise?

```
8. If Type(resolution) is not Object, then
  a. Perform FulfillPromise(promise, resolution).
  b. Return undefined.
9. If IsPromise(x) is true, then
  a. Let xConstructor be ? Get(x, "constructor").
  b. If SameValue(xConstructor, %Promise%) is true, then
    i. Let resolvingFunctions be CreateResolvingFunctions(promiseToResolve).
    ii. Perform PerformPromiseThen(x, resolvingFunctions.[[Resolve]], resolvingFunctions.[[Reject]]).
    iii. Return undefined.
10. Let then be Completion(Get(resolution, "then")).
11. If then is an abrupt completion, then
  a. Perform RejectPromise(promise, then.[[Value]]).
  b. Return undefined.
12. Let thenAction be then.[[Value]].
12. If IsCallable(thenAction) is false, then
  a. Perform FulfillPromise(promise, resolution).
  b. Return undefined.
13. Let thenJobCallback be HostMakeJobCallback(thenAction).
14. Let job be NewPromiseResolveThenableJob(promise, resolution, thenJobCallback).
15. Perform HostEnqueuePromiseJob(job.[[Job]], job.[[Realm]]).
16. Return undefined.

```

bakkot:
> Wouldn't this immediately call `PerformPromiseThen` with the now-verified Promise?

That was the idea, yes. (Same as my [earlier suggestion](https://github.com/tc39/ecma262/pull/2772#issuecomment-1122816364), just skipping some intermediate steps.)

`PerformPromiseThen` doesn't invoke any user code synchronously, so that seems fine.

zloirock:
`core-js` implementation use the actual semantic. The feature detection does not contain this behavior.

However, sometimes I see issues about the order of promises resolution in similar cases, so I'm pretty sure that it will break something in the Web.

mhofman:
@zloirock could you clarify "the"/"this"?

Does core-js follow the A+ rules (synchronous call) or the currently specified ES Promises (delayed call).

If taking the hybrid approach of only delaying non native promises and fast pathing native promise with no custom then, the only impact should be on programs that are particularly sensitive to the number of ticks.

zloirock:
> could you clarify "the"/"this"?

I mean [`core-js` feature detection of `Promise`](https://github.com/zloirock/core-js/blob/v3.22.5/packages/core-js/internals/promise-constructor-detection.js).

> Does core-js follow the A+ rules (synchronous call) or the currently specified ES Promises (delayed call).

Sure, [delayed call](https://github.com/zloirock/core-js/blob/v3.22.5/packages/core-js/modules/es.promise.constructor.js#L175), like in the actual spec.

jridgewell:
For interested parties, I've created https://github.com/tc39/proposal-faster-promise-adoption to track the proposal.

PR #2773, opened 2022-05-10
Editorial: Improve the numeric string grammar summary

gibson042:
* Imply semantics for `:::` grammars (cf. https://github.com/tc39/proposal-temporal/issues/2190 ).
* Make the numeric string grammar summary more similar to other grammars.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -533,7 +533,7 @@
     <emu-clause id="sec-numeric-string-grammar">
       <h1>The Numeric String Grammar</h1>
       <p>Another grammar is used for translating Strings into numeric values. This grammar is similar to the part of the lexical grammar having to do with numeric literals and has as its terminal symbols |SourceCharacter|. This grammar appears in <emu-xref href="#sec-tonumber-applied-to-the-string-type"></emu-xref>.</p>
-      <p>Productions of the numeric string grammar are distinguished by having three colons &ldquo;<b>:::</b>&rdquo; as punctuation.</p>
+      <p>Productions of the numeric string grammar are distinguished by having three colons &ldquo;<b>:::</b>&rdquo; as punctuation, and are never used for parsing source text.</p>
     </emu-clause>
 
     <emu-clause id="sec-syntactic-grammar">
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -533,7 +533,7 @@ <h1>The Lexical and RegExp Grammars</h1>
     <emu-clause id="sec-numeric-string-grammar">
       <h1>The Numeric String Grammar</h1>
       <p>Another grammar is used for translating Strings into numeric values. This grammar is similar to the part of the lexical grammar having to do with numeric literals and has as its terminal symbols |SourceCharacter|. This grammar appears in <emu-xref href="#sec-tonumber-applied-to-the-string-type"></emu-xref>.</p>
-      <p>Productions of the numeric string grammar are distinguished by having three colons &ldquo;<b>:::</b>&rdquo; as punctuation.</p>
+      <p>Productions of the numeric string grammar are distinguished by having three colons &ldquo;<b>:::</b>&rdquo; as punctuation, and are never used for parsing source text.</p>
```
jmdyck, line 536: I think this addition might fit better in the preceding paragraph, since they both talk about what the grammar is used for.

gibson042: Perhaps, but I'm trying to associate `:::` with "never used for parsing source text" (as the key differentiator from `::` grammars, which _can_ be used for parsing source text).

jmdyck: (Another possibility is that it's not that useful a distinction and we should just use two colons for all lexical productions.)

gibson042: Strongly agree. This stemmed from a discussion on Matrix: https://matrix.to/#/!WgJwmjBNZEXhJnXHXw:matrix.org/$6jQ6AA-u_dncsX1u8q68yzAqEfl_yTYbhvt7mVBJyz8
> but yeah, `:::` is for parsings strings rather than source text seems consistent with the current usages

bakkot (approved):
Works for me.

michaelficarra (approved):

syg (approved):

PR #2776, opened 2022-05-16
Editorial: Remove use of "must" from liveness note

syg:
Came up in https://es.discourse.group/t/note-in-liveness-section-appears-to-be-normative/1345,
where the use of "must" can be misconstrued as additional normative
requirements.


syg:
cc @acutmore

acutmore:
Thanks!

bakkot (approved):

PR #2777, opened 2022-05-17
Normative: Permit Symbols as WeakMap keys

### LINKED ISSUE #1194: Symbols can't be used as WeakMap keys

acutmore:
## [rendered preview here](https://arai-a.github.io/ecma262-compare/?pr=2777)

This PR represents the Stage 3 specification text for [Symbols as WeakMap keys](https://github.com/tc39/proposal-symbols-as-weakmap-keys).

The proposed spec has been written as a PR against the full spec due to the high number of small tweaks to descriptions and notes in sections such as liveness.

A separate minimal proposal specification with only the core changes is available at http://tc39.es/proposal-symbols-as-weakmap-keys

Closes #1194

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12012,8 +12012,8 @@
         </tr>
         <tr>
           <td>[[KeptAlive]]</td>
-          <td>a List of Objects</td>
-          <td>Initially a new empty List, representing the list of objects to be kept alive until the end of the current Job</td>
+          <td>a List of either Objects or Symbols</td>
+          <td>Initially a new empty List, representing the list of objects and/or symbols to be kept alive until the end of the current Job</td>
         </tr>
       </table>
     </emu-table>
@@ -12128,22 +12128,22 @@
   </emu-clause>
 
   <emu-clause id="sec-weakref-processing-model">
-    <h1>Processing Model of WeakRef and FinalizationRegistry Objects</h1>
+    <h1>Processing Model of WeakRef and FinalizationRegistry Targets</h1>
 
     <emu-clause id="sec-weakref-invariants">
       <h1>Objectives</h1>
 
-      <p>This specification does not make any guarantees that any object will be garbage collected. Objects which are not live may be released after long periods of time, or never at all. For this reason, this specification uses the term "may" when describing behaviour triggered by garbage collection.</p>
+      <p>This specification does not make any guarantees that any object or symbol will be garbage collected. Objects or symbols which are not live may be released after long periods of time, or never at all. For this reason, this specification uses the term "may" when describing behaviour triggered by garbage collection.</p>
 
       <p>The semantics of WeakRefs and FinalizationRegistrys is based on two operations which happen at particular points in time:</p>
 
       <ul>
         <li>
-          When `WeakRef.prototype.deref` is called, the referent (if *undefined* is not returned) is kept alive so that subsequent, synchronous accesses also return the object. This list is reset when synchronous work is done using the ClearKeptObjects abstract operation.
+          When `WeakRef.prototype.deref` is called, the referent (if *undefined* is not returned) is kept alive so that subsequent, synchronous accesses also return the same value. This list is reset when synchronous work is done using the ClearKeptObjects abstract operation.
         </li>
 
         <li>
-          When an object which is registered with a FinalizationRegistry becomes unreachable, a call of the FinalizationRegistry's cleanup callback may eventually be made, after synchronous ECMAScript execution completes. The FinalizationRegistry cleanup is performed with the CleanupFinalizationRegistry abstract operation.
+          When an object or symbol which is registered with a FinalizationRegistry becomes unreachable, a call of the FinalizationRegistry's cleanup callback may eventually be made, after synchronous ECMAScript execution completes. The FinalizationRegistry cleanup is performed with the CleanupFinalizationRegistry abstract operation.
         </li>
       </ul>
 
@@ -12155,33 +12155,33 @@
     <emu-clause id="sec-liveness">
       <h1>Liveness</h1>
 
-      <p>For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref of a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
+      <p>For some set of objects and/or symbols _S_ a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref of a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
 
       <emu-note>
-        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep an object alive. Two, that cycles in liveness does not imply that an object is live. To be concrete, if determining _obj_'s liveness depends on determining the liveness of another WeakRef referent, _obj2_, _obj2_'s liveness cannot assume _obj_'s liveness, which would be circular reasoning.
+        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep its referent alive. Two, that cycles in liveness does not imply that a value is live. To be concrete, if determining _v_'s liveness depends on determining the liveness of a WeakRef referent, _r_, _r_'s liveness cannot assume _v_'s liveness, which would be circular reasoning.
       </emu-note>
       <emu-note>
-        WeakRef-obliviousness is defined on sets of objects instead of individual objects to account for cycles. If it were defined on individual objects, then an object in a cycle will be considered live even though its Object value is only observed via WeakRefs of other objects in the cycle.
+        WeakRef-obliviousness is defined on sets of objects or symbols instead of individual values to account for cycles. If it were defined on individual values, then a WeakRef referent in a cycle will be considered live even though its identity is only observed via other WeakRef referents in the cycle.
       </emu-note>
       <emu-note>
-        Colloquially, we say that an individual object is live if every set of objects containing it is live.
+        Colloquially, we say that an individual object or symbol is live if every set containing it is live.
       </emu-note>
 
-      <p>At any point during evaluation, a set of objects _S_ is considered <dfn>live</dfn> if either of the following conditions is met:</p>
+      <p>At any point during evaluation, a set of objects and/or symbols _S_ is considered <dfn>live</dfn> if either of the following conditions is met:</p>
 
       <ul>
         <li>
           Any element in _S_ is included in any agent's [[KeptAlive]] List.
         </li>
         <li>
-          There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes the Object value of any object in _S_.
+          There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes the identity of any value in _S_.
         </li>
       </ul>
       <emu-note>
-        The second condition above intends to capture the intuition that an object is live if its identity is observable via non-WeakRef means. An object's identity may be observed by observing a strict equality comparison between objects or observing the object being used as key in a Map.
+        The second condition above intends to capture the intuition that a value is live if its identity is observable via non-WeakRef means. A value's identity may be observed by observing a strict equality comparison or observing the value being used as key in a Map.
       </emu-note>
       <emu-note>
-        <p>Presence of an object in a field, an internal slot, or a property does not imply that the object is live. For example if the object in question is never passed back to the program, then it cannot be observed.</p>
+        <p>Presence of an object or a symbol in a field, an internal slot, or a property does not imply that the value is live. For example if the value in question is never passed back to the program, then it cannot be observed.</p>
 
         <p>This is the case for keys in a WeakMap, members of a WeakSet, as well as the [[WeakRefTarget]] and [[UnregisterToken]] fields of a FinalizationRegistry Cell record.</p>
 
@@ -12195,20 +12195,20 @@
     <emu-clause id="sec-weakref-execution">
       <h1>Execution</h1>
 
-      <p>At any time, if a set of objects _S_ is not live, an ECMAScript implementation may perform the following steps atomically:</p>
+      <p>At any time, if a set of objects and/or symbols _S_ is not live, an ECMAScript implementation may perform the following steps atomically:</p>
 
       <emu-alg>
-        1. For each element _obj_ of _S_, do
-          1. For each WeakRef _ref_ such that _ref_.[[WeakRefTarget]] is _obj_, do
+        1. For each element _value_ of _S_, do
+          1. For each WeakRef _ref_ such that _ref_.[[WeakRefTarget]] is _value_, do
             1. Set _ref_.[[WeakRefTarget]] to ~empty~.
-          1. For each FinalizationRegistry _fg_ such that _fg_.[[Cells]] contains a Record _cell_ such that _cell_.[[WeakRefTarget]] is _obj_, do
+          1. For each FinalizationRegistry _fg_ such that _fg_.[[Cells]] contains a Record _cell_ such that _cell_.[[WeakRefTarget]] is _value_, do
             1. Set _cell_.[[WeakRefTarget]] to ~empty~.
             1. Optionally, perform HostEnqueueFinalizationRegistryCleanupJob(_fg_).
-          1. For each WeakMap _map_ such that _map_.[[WeakMapData]] contains a Record _r_ such that _r_.[[Key]] is _obj_, do
+          1. For each WeakMap _map_ such that _map_.[[WeakMapData]] contains a Record _r_ such that _r_.[[Key]] is _value_, do
             1. Set _r_.[[Key]] to ~empty~.
             1. Set _r_.[[Value]] to ~empty~.
-          1. For each WeakSet _set_ such that _set_.[[WeakSetData]] contains _obj_, do
-            1. Replace the element of _set_.[[WeakSetData]] whose value is _obj_ with an element whose value is ~empty~.
+          1. For each WeakSet _set_ such that _set_.[[WeakSetData]] contains _value_, do
+            1. Replace the element of _set_.[[WeakSetData]] whose value is _value_ with an element whose value is ~empty~.
       </emu-alg>
 
       <emu-note>
@@ -12221,8 +12221,8 @@
         <p>Because calling HostEnqueueFinalizationRegistryCleanupJob is optional, registered objects in a FinalizationRegistry do not necessarily hold that FinalizationRegistry live. Implementations may omit FinalizationRegistry callbacks for any reason, e.g., if the FinalizationRegistry itself becomes dead, or if the application is shutting down.</p>
       </emu-note>
       <emu-note>
-        <p>Implementations are not obligated to empty WeakRefs for maximal sets of non-live objects.</p>
-        <p>If an implementation chooses a non-live set _S_ in which to empty WeakRefs, this definition requires that it empties WeakRefs for all objects in _S_ simultaneously. In other words, it is not conformant for an implementation to empty a WeakRef pointing to an object _obj_ without emptying out other WeakRefs that, if not emptied, could result in an execution that observes the Object value of _obj_.</p>
+        <p>Implementations are not obligated to empty WeakRefs for maximal sets of non-live objects or symbols.</p>
+        <p>If an implementation chooses a non-live set _S_ in which to empty WeakRefs, this definition requires that it empties WeakRefs for all values in _S_ simultaneously. In other words, it is not conformant for an implementation to empty a WeakRef pointing to a value _v_ without emptying out other WeakRefs that, if not emptied, could result in an execution that observes the value of _v_.</p>
       </emu-note>
     </emu-clause>
 
@@ -12264,18 +12264,18 @@
   <emu-clause id="sec-addtokeptobjects" type="abstract operation">
     <h1>
       AddToKeptObjects (
-        _object_: an Object,
+        _value_: an Object or a Symbol,
       ): ~unused~
     </h1>
     <dl class="header">
     </dl>
     <emu-alg>
       1. Let _agentRecord_ be the surrounding agent's Agent Record.
-      1. Append _object_ to _agentRecord_.[[KeptAlive]].
+      1. Append _value_ to _agentRecord_.[[KeptAlive]].
       1. Return ~unused~.
     </emu-alg>
     <emu-note>
-      When the abstract operation AddToKeptObjects is called with a target object reference, it adds the target to a list that will point strongly at the target until ClearKeptObjects is called.
+      When the abstract operation AddToKeptObjects is called with a target object or symbol, it adds the target to a list that will point strongly at the target until ClearKeptObjects is called.
     </emu-note>
   </emu-clause>
 
@@ -12296,6 +12296,26 @@
         1. Perform ? HostCallJobCallback(_callback_, *undefined*, « _cell_.[[HeldValue]] »).
       1. Return ~unused~.
     </emu-alg>
+  </emu-clause>
+
+  <emu-clause id="sec-canbeheldweakly" type="abstract operation">
+    <h1>
+      CanBeHeldWeakly (
+        _v_: an ECMAScript language value,
+      ): a Boolean
+    </h1>
+    <dl class="header">
+      <dt>description</dt>
+      <dd>It returns *true* if and only if _v_ is suitable for use as a weak reference. Only values that are suitable for use as a weak reference may be a key of a WeakMap, an element of a WeakSet, the target of a WeakRef, or one of the targets of a FinalizationRegistry.</dd>
+    </dl>
+    <emu-alg>
+      1. If _v_ is an Object, return *true*.
+      1. If _v_ is a Symbol and KeyForSymbol(_v_) is *undefined*, return *true*.
+      1. Return *false*.
+    </emu-alg>
+    <emu-note>
+      <p>A language value without <emu-xref href="#sec-identity">language identity</emu-xref> can be manifested without prior reference and is unsuitable for use as a weak reference. A Symbol value produced by <emu-xref href="#sec-symbol.for">Symbol.for</emu-xref>, unlike other Symbol values, does not have language identity and is unsuitable for use as a weak reference. <emu-xref href="#sec-well-known-symbols">Well-known symbols</emu-xref> are likely to never be collected, but are nonetheless treated as suitable for use as a weak reference because they are limited in number and therefore manageable by a variety of implementation approaches. However, any value associated to a well-known symbol in a live WeakMap is unlikely to be collected and could “leak” memory resources in implementations.</p>
+    </emu-note>
   </emu-clause>
 </emu-clause>
 
@@ -30263,7 +30283,7 @@
           1. Append the Record { [[Key]]: _stringKey_, [[Symbol]]: _newSymbol_ } to the GlobalSymbolRegistry List.
           1. Return _newSymbol_.
         </emu-alg>
-        <p>The GlobalSymbolRegistry is a List that is globally available. It is shared by all realms. Prior to the evaluation of any ECMAScript code, it is initialized as a new empty List. Elements of the GlobalSymbolRegistry are Records with the structure defined in <emu-xref href="#table-globalsymbolregistry-record-fields"></emu-xref>.</p>
+        <p>The GlobalSymbolRegistry is an append-only List that is globally available. It is shared by all realms. Prior to the evaluation of any ECMAScript code, it is initialized as a new empty List. Elements of the GlobalSymbolRegistry are Records with the structure defined in <emu-xref href="#table-globalsymbolregistry-record-fields"></emu-xref>.</p>
         <emu-table id="table-globalsymbolregistry-record-fields" caption="GlobalSymbolRegistry Record Fields" oldids="table-44">
           <table>
             <tr>
@@ -30326,10 +30346,7 @@
         <p>This function performs the following steps when called:</p>
         <emu-alg>
           1. If _sym_ is not a Symbol, throw a *TypeError* exception.
-          1. For each element _e_ of the GlobalSymbolRegistry List (see <emu-xref href="#sec-symbol.for"></emu-xref>), do
-            1. If SameValue(_e_.[[Symbol]], _sym_) is *true*, return _e_.[[Key]].
-          1. Assert: GlobalSymbolRegistry does not currently contain an entry for _sym_.
-          1. Return *undefined*.
+          1. Return KeyForSymbol(_sym_).
         </emu-alg>
       </emu-clause>
 
@@ -30485,6 +30502,28 @@
     <emu-clause id="sec-properties-of-symbol-instances">
       <h1>Properties of Symbol Instances</h1>
       <p>Symbol instances are ordinary objects that inherit properties from the Symbol prototype object. Symbol instances have a [[SymbolData]] internal slot. The [[SymbolData]] internal slot is the Symbol value represented by this Symbol object.</p>
+    </emu-clause>
+
+    <emu-clause id="sec-abstract-operations-for-symbols">
+      <h1>Abstract Operations for Symbols</h1>
+
+      <emu-clause id="sec-keyforsymbol" type="abstract operation">
+        <h1>
+          KeyForSymbol (
+            _sym_: a Symbol,
+          ): a String or *undefined*
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>If _sym_ is in the GlobalSymbolRegistry (see <emu-xref href="#sec-symbol.for"></emu-xref>) the String used to register _sym_ will be returned.</dd>
+        </dl>
+        <emu-alg>
+          1. For each element _e_ of the GlobalSymbolRegistry List, do
+            1. If SameValue(_e_.[[Symbol]], _sym_) is *true*, return _e_.[[Key]].
+          1. Assert: GlobalSymbolRegistry does not currently contain an entry for _sym_.
+          1. Return *undefined*.
+        </emu-alg>
+      </emu-clause>
     </emu-clause>
   </emu-clause>
 
@@ -40910,11 +40949,11 @@ THH:mm:ss.sss
 
   <emu-clause id="sec-weakmap-objects">
     <h1>WeakMap Objects</h1>
-    <p>WeakMaps are collections of key/value pairs where the keys are objects and values may be arbitrary ECMAScript language values. A WeakMap may be queried to see if it contains a key/value pair with a specific key, but no mechanism is provided for enumerating the objects it holds as keys. In certain conditions, objects which are not live are removed as WeakMap keys, as described in <emu-xref href="#sec-weakref-execution"></emu-xref>.</p>
+    <p>WeakMaps are collections of key/value pairs where the keys are objects and/or symbols and values may be arbitrary ECMAScript language values. A WeakMap may be queried to see if it contains a key/value pair with a specific key, but no mechanism is provided for enumerating the values it holds as keys. In certain conditions, values which are not live are removed as WeakMap keys, as described in <emu-xref href="#sec-weakref-execution"></emu-xref>.</p>
     <p>An implementation may impose an arbitrarily determined latency between the time a key/value pair of a WeakMap becomes inaccessible and the time when the key/value pair is removed from the WeakMap. If this latency was observable to ECMAScript program, it would be a source of indeterminacy that could impact program execution. For that reason, an ECMAScript implementation must not provide any means to observe a key of a WeakMap that does not require the observer to present the observed key.</p>
     <p>WeakMaps must be implemented using either hash tables or other mechanisms that, on average, provide access times that are sublinear on the number of key/value pairs in the collection. The data structure used in this specification is only intended to describe the required observable semantics of WeakMaps. It is not intended to be a viable implementation model.</p>
     <emu-note>
-      <p>WeakMap and WeakSets are intended to provide mechanisms for dynamically associating state with an object in a manner that does not “leak” memory resources if, in the absence of the WeakMap or WeakSet, the object otherwise became inaccessible and subject to resource reclamation by the implementation's garbage collection mechanisms. This characteristic can be achieved by using an inverted per-object mapping of weak map instances to keys. Alternatively each weak map may internally store its key to value mappings but this approach requires coordination between the WeakMap or WeakSet implementation and the garbage collector. The following references describe mechanism that may be useful to implementations of WeakMap and WeakSets:</p>
+      <p>WeakMap and WeakSet are intended to provide mechanisms for dynamically associating state with an object or symbol in a manner that does not “leak” memory resources if, in the absence of the WeakMap or WeakSet instance, the object or symbol otherwise became inaccessible and subject to resource reclamation by the implementation's garbage collection mechanisms. This characteristic can be achieved by using an inverted per-object/symbol mapping of WeakMap or WeakSet instances to keys. Alternatively, each WeakMap or WeakSet instance may internally store its key and value data, but this approach requires coordination between the WeakMap or WeakSet implementation and the garbage collector. The following references describe mechanism that may be useful to implementations of WeakMap and WeakSet:</p>
       <p>Barry Hayes. 1997. Ephemerons: a new finalization mechanism. In <i>Proceedings of the 12th ACM SIGPLAN conference on Object-oriented programming, systems, languages, and applications (OOPSLA '97)</i>, A. Michael Berman (Ed.). ACM, New York, NY, USA, 176-183, <a href="http://doi.acm.org/10.1145/263698.263733">http://doi.acm.org/10.1145/263698.263733</a>.</p>
       <p>Alexandra Barros, Roberto Ierusalimschy, Eliminating Cycles in Weak Tables. Journal of Universal Computer Science - J.UCS, vol. 14, no. 21, pp. 3481-3497, 2008, <a href="http://www.jucs.org/jucs_14_21/eliminating_cycles_in_weak">http://www.jucs.org/jucs_14_21/eliminating_cycles_in_weak</a></p>
     </emu-note>
@@ -40984,7 +41023,7 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _M_ be the *this* value.
           1. Perform ? RequireInternalSlot(_M_, [[WeakMapData]]).
-          1. If _key_ is not an Object, return *false*.
+          1. If CanBeHeldWeakly(_key_) is *false*, return *false*.
           1. For each Record { [[Key]], [[Value]] } _p_ of _M_.[[WeakMapData]], do
             1. If _p_.[[Key]] is not ~empty~ and SameValue(_p_.[[Key]], _key_) is *true*, then
               1. Set _p_.[[Key]] to ~empty~.
@@ -41003,7 +41042,7 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _M_ be the *this* value.
           1. Perform ? RequireInternalSlot(_M_, [[WeakMapData]]).
-          1. If _key_ is not an Object, return *undefined*.
+          1. If CanBeHeldWeakly(_key_) is *false*, return *undefined*.
           1. For each Record { [[Key]], [[Value]] } _p_ of _M_.[[WeakMapData]], do
             1. If _p_.[[Key]] is not ~empty~ and SameValue(_p_.[[Key]], _key_) is *true*, return _p_.[[Value]].
           1. Return *undefined*.
@@ -41016,7 +41055,7 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _M_ be the *this* value.
           1. Perform ? RequireInternalSlot(_M_, [[WeakMapData]]).
-          1. If _key_ is not an Object, return *false*.
+          1. If CanBeHeldWeakly(_key_) is *false*, return *false*.
           1. For each Record { [[Key]], [[Value]] } _p_ of _M_.[[WeakMapData]], do
             1. If _p_.[[Key]] is not ~empty~ and SameValue(_p_.[[Key]], _key_) is *true*, return *true*.
           1. Return *false*.
@@ -41029,7 +41068,7 @@ THH:mm:ss.sss
         <emu-alg>
           1. Let _M_ be the *this* value.
           1. Perform ? RequireInternalSlot(_M_, [[WeakMapData]]).
-          1. If _key_ is not an Object, throw a *TypeError* exception.
+          1. If CanBeHeldWeakly(_key_) is *false*, throw a *TypeError* exception.
           1. For each Record { [[Key]], [[Value]] } _p_ of _M_.[[WeakMapData]], do
             1. If _p_.[[Key]] is not ~empty~ and SameValue(_p_.[[Key]], _key_) is *true*, then
               1. Set _p_.[[Value]] to _value_.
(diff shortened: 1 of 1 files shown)


ljharb:
@acutmore i'll take a look at it when i have a moment.

Jack-Works (changes requested):

spec.html
```diff
@@ -12491,6 +12491,29 @@ <h1>
...
+      1. If Type(_v_) is Symbol, then
+        1. For each element _e_ of the GlobalSymbolRegistry List (see <emu-xref href="#sec-symbol.for"></emu-xref>), do
+          1. If SameValue(_e_.[[Symbol]], _v_) is *true*, return *false*.
+        1. Return *true*.
+      1. Return *false*.
+    </emu-alg>
+    <emu-note type="editor">
+      <p>This abstract operation determines if _v_ is considered to have an identity suitable for a weak reference. i.e. Is _v_ a valid candidate key of a WeakMap, or entry in a WeakSet, or target of a WeakRef or FinalizationRegistry.</p>
+      <p>'Registered symbols' are not considered to have a suitable identity because their identity is similar to strings, and implementations may attempt to collect them when they are unreachable.</p>
+      <p>'Well-Known symbols' <emu-xref href="#sec-well-known-symbols"></emu-xref> are considered to have a suitable identity because there are a small finite number of them and implementations do not attempt to unobservably collect them.</p>
```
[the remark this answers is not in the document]
acutmore: Hi @Jack-Works , thanks for reviewing the proposal.

The champion group have been discussing this at the record&tuple monthly meeting. The reason we landed on only disallowing registered symbols is that: engines like spidermonkey do use an internal WeakMap to store registered symbols but we did not find evidence of an engine applying this ‘optimisation’ to well known symbols. As there are a small finite number of well known symbols it is arguably more efficient to allocate them once and share these references across all realms (as engines currently do) than to try and collect and re-materialize them.

The optimisation has greater benefit for registered symbols as there is effectively no limit to how many a program could create.

In your example you would also need `delete Symbol.prototype.constructor`. The examples of code demonstrating how well known symbols could be made unreachable do not seem to be realistic use cases for the majority of programs, we find this to be further evidence that engines would be unlikely to invest in applying this 'optimisation'.

I hope that helps clarify. If not please do raise an issue https://github.com/tc39/proposal-symbols-as-weakmap-keys/issues so we can discuss it there. Thanks!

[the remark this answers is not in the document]
Jack-Works: Yeah, I agree to make well-known symbols inaccessible is a very edge case.

I just want to use this example to show the difference in how we categorize well-known symbols. I think well-known symbols are more like registered symbols than normal symbols, if my example doesn't convince you, I don't have stronger objections.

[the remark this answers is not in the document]
hax: I assume the spec is not only a spec of current known implementations, but for all implementors/programmers, so the problem here might be, current draft text only say "implementations do not attempt to unobservably collect well-known symbols" in the editor note, but if we allow well-known symbols, I think the text should clearly say "well-known symbols must never be collected observably" (or it will break SES requirements if i understand correctly). 

On the other side, I also very dislike allowing well-known symbols, because if they must never be collected, and seems very contradict the "CanBeHeldWeakly" semantic and the purpose of using weakmap/weakset/weakref in the first place, and I can't imagine any valid use case. Allowing them just add confusion and footgun to developers.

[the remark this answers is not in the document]
acutmore: > so the problem here might be, current draft text only say "implementations do not attempt to unobservably collect well-known symbols" in the editor note, but if we allow well-known symbols, I think the text should clearly say "well-known symbols must never be collected observably"

Good point. I agree that I worded this note poorly. It should not reference current state of engines. I will re-word.

> On the other side, I also very dislike allowing well-known symbols, because if they must never be collected, it seems very contradict the "CanBeHeldWeakly" semantic 

One reason the champion group is ok with this is because for a single realm application this is no different to `Object.prototype` - while it can be held weakly, this does not imply that it is useful to do so. Hope that helps explain our reasoning.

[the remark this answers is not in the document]
Jack-Works: image

```js
delete globalThis.Symbol
// No access to @@asyncIterator now!
// (despite Host APIs)
const proxy = new Proxy({}, { get(...args) { console.log(...args) } })
for await (const x of proxy) {}

```

In this example, there is no multiple Realm, but I still made it (I get `@@asyncIterator` from the void).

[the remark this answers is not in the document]
acutmore: Yes - that example shows why `Symbol.iterator` and `Symbol.asyncIterator` would still be considered 'live'. The definition of liveness:

> There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes any value in _S_.

So if there is a way to 'make' the well-know-symbol then it is still 'live' and cannot be removed from a WeakMap

acutmore:
> @acutmore i'll take a look at it when i have a moment.

The generated preview seems to have sorted itself out now - thanks anyway!

spec.html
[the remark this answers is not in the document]
acutmore: Some of the reasons we are treating them as different from registered symbols are: it is easier to _detect_ a registered symbol*  and there are a small finite number of them.  Originally the proposal allowed all symbols. After many years of discussing this detail we feel that only blocking registered symbols is the best compromise all things considered.

The use cases for symbols as WeakMap keys are for more _advanced_ patterns that libraries can utilize to provide their higher-level features. This should also help increase the chance that the users of this proposal are aware of the footgun.

<details>
<summary>*</summary>

```js
const {keyFor} = Symbol;
const isRegisteredSymbol = s => typeof s === 'symbol'  && keyFor(s) !== undefined;

https://www.npmjs.com/package/is-well-known-symbol
https://gist.github.com/acutmore/0a5bf0f35785edfe3c17e1f6d05e8921
```

</details>

spec.html
```diff
@@ -12349,33 +12349,33 @@ <h1>Objectives</h1>
...
           Any element in _S_ is included in any agent's [[KeptAlive]] List.
         </li>
         <li>
-          There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes the Object value of any object in _S_.
+          There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes any value in _S_.
         </li>
       </ul>
       <emu-note>
-        The second condition above intends to capture the intuition that an object is live if its identity is observable via non-WeakRef means. An object's identity may be observed by observing a strict equality comparison between objects or observing the object being used as key in a Map.
+        The second condition above intends to capture the intuition that a referent is live if its identity is observable via non-WeakRef means. A referent's identity may be observed by observing a strict equality comparison or observing the value being used as key in a Map.
```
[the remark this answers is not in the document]
acutmore: I'm more than happy to have this change, though curious what made you think to add it? In my mind's pseudo-mark-sweep-gc of the spec, there is always at least one live realm during execution and the AOs and intrinsic functions are like a _static roots_, so because they reference the well-known symbols then the symbols are reachable from the GC - even if not reachable from userland.

@rricard and I did briefly talk about if we should add a concrete addition to the definition of liveness that says well-known symbols are always considered live. What do you think?

spec.html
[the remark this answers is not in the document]
hax: > for a single realm application this is no different to `Object.prototype` - while it can be held weakly, this does not imply that it is useful to do so. Hope that helps explain our reasoning.

I think we need to consider the issue in the scope of multiple realms because the constraints are actually highly cross-realm related. So `Object.prototype` is not same as well-known symbols. Though the `Object.prototype` from the same realm of `WeakMap/WeakSet/WeakRef` is unlikely to be GCed in current realm lift time, the `Object.prototype` from other realm (via iframe) are just normal objects and should be GCed and could be observed.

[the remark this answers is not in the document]
hax: > The important boundary to draw isn’t “cross-realm symbols must be disallowed”, it’s “an infinite number of cross-realm symbols must not be weakly held”.

I don't know why "infinite" matters here. As I understand, the hard constraint is avoiding side-channel which cross-realm observable GC could create. So even well-known symbols are finite, they can't have observable GC, and as current implementations, they are not GCed at all.

Generally, any values which can be instantiated from the void should either never be GCed (like null, true, false, numbers) or could be GCed but non-observable (like bigints, strings, registered symbols).  In both cases they can't be held weakly.

Jack's example explain that well-known symbols could also be instantiated from the void. So make it "can held weakly" just introduce an exceptional case.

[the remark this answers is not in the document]
acutmore: Hi @hax - well-known symbols are never instantiated by the program - they exist outside of realms. When a realm is created the global `Symbol` object is created and has it's properties defined, but the value that `Symbol.iterator` references is not constructed.

The design revolves around:

- WeakMap does not attempt to block invalid objects that can't be collected
  - you can put a weakMap inside itself: `((wm) => wm.set(wm, true))(new WeakMap)`
  - The root-realm's `Object.prototype` will never be collected, and is allowed
- There are implementation complexities when allowing registered-symbols as weakMap keys
  - SpiderMonkey implement registered symbols using an internal weakMap. Caution would need to be taken in weakmap/weakset/weakref/finalizationregistry to ensure this optimization isn't observable.
- Delegates have raised concern that having an API that treats unique and well-known symbols differently adds more complexity to users, more than helps to avoid bugs.
- If well-known symbols do not throw an exception, this does not prevent devtools from issuing a warning.

Having different delegates with different opinions on which categories of symbol should be allowed is one of the reasons this subject has [been talked about for at least 4 years](https://github.com/tc39/ecma262/issues/1194#issue-323332905). The champions of this proposal feel that a design that only blocks registered symbols is the best compromise once all things have been taken into account.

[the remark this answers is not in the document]
ljharb: > I don't know why "infinite" matters here. 

Because that is the primary cause of objection from multiple delegates about allowing registered symbols - that the number of them is uncountable.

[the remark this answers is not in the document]
hax: Still don't get the reason why "uncountable" matters. I guess it means u could check finite set of values to avoid leak, but u can't check "uncountable" values??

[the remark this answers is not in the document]
hax: > Delegates have raised concern that having an API that treats unique and well-known symbols differently adds more complexity to users, more than helps to avoid bugs.

Treats well-known symbols and registered symbols differently also have similar problems, and even worse, because in most aspects ("undeniable", GC behavior), well-known symbols are much close to registered symbols.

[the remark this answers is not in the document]
ljharb: Yes, that's exactly right.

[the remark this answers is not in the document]
ljharb: Well-known symbols aren't part of the symbol registry. They *are* different, in a way that well-known and normal symbols aren't different.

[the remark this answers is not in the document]
hax: > WeakMap does not attempt to block invalid objects that can't be collected

I didn't talk about the occasional uncollectable cases (for example, unwritable and unconfigurable properties on `globalThis` will never be collected, `wm.set(wm)` and `Object.prototype` are all such cases), but the inherent non-GC/unobservable-GC cases.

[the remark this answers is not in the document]
acutmore: > > WeakMap does not attempt to block invalid objects that can't be collected
> 
> I didn't talk about the occasional uncollectable cases (for example, unwritable and unconfigurable properties on `globalThis` will never be collected, `wm.set(wm)` and `Object.prototype` are also such cases), but the inherent non-GC/unobservable-GC cases.

I was not clear. I did not mean to suggest that well-known symbols are the same as these objects. What I mean is that the existing API does not attempt to throw a fatal exception for values that are likely to be programmer errors.

The main reason for throwing an exception on well-known symbols wouldn’t be because they don’t have observable GC but because we think that this level of protection is required to prevent users of the API making a mistake. The use cases for this API do not suggest that attempting to use well-known symbols as weak map keys will be a common mistake - so it is perhaps simpler to allow them rather than special case them.

spec.html
```diff
@@ -12370,12 +12370,15 @@ <h1>Liveness</h1>
         <li>
           There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes any value in _S_.
         </li>
+        <li>
+          Any element in _S_ is a well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>).
```
ljharb, line 12374:
```suggestion
          Any element in _S_ is a well-known symbol (<emu-xref href="#sec-well-known-symbols"></emu-xref>).
```

spec.html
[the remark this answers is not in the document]
acutmore: PR updated to make the liveness of well-known symbols explicit.

spec.html
```diff
@@ -12206,8 +12206,8 @@ <h1>Agents</h1>
         </tr>
         <tr>
           <td>[[KeptAlive]]</td>
-          <td>a List of Objects</td>
-          <td>Initially a new empty List, representing the list of objects to be kept alive until the end of the current Job</td>
+          <td>a List of ECMAScript language values</td>
+          <td>Initially a new empty List, representing the list of objects and symbols to be kept alive until the end of the current Job</td>
```
ljharb, line 12210:
```suggestion
          <td>Initially a new empty List, representing the list of objects and/or symbols to be kept alive until the end of the current Job</td>
```

spec.html
```diff
@@ -40468,7 +40494,7 @@ <h1>WeakMap.prototype.set ( _key_, _value_ )</h1>
           1. Let _M_ be the *this* value.
           1. Perform ? RequireInternalSlot(_M_, [[WeakMapData]]).
           1. Let _entries_ be the List that is _M_.[[WeakMapData]].
-          1. If Type(_key_) is not Object, throw a *TypeError* exception.
+          1. If CanBeHeldWeakly(_key_) is *false*, throw a *TypeError* exception.
```
ljharb, line 40497: personally i'd find it nice if the refactor of "type is object" to "canbeheldweakly" was done in a separate commit (or PR) - ie, that one first, and the "allow symbols" second. I think it would make the changes much easier to review.

not sure what @tc39/ecma262-editors think about that?

ljharb (commented):
Before this lands, I think 402 will need to be updated so the IntlFallbackSymbol is appended to the well-known symbols table (such that the symbol is automatically allowed to be held weakly)

Looks good overall though!

brad4d (approved):

spec.html
acutmore: During stage 2 I've been prioritizing keeping the git history intact, so hopefully easier for an individual to see what has changed since the last time they reviewed.

If this gets to stage 3 we could definitely look at splitting apart the editorial and normative changes into separate PRs as you suggest.

acutmore:
> Before this lands, I think 402 will need to be updated so the IntlFallbackSymbol is appended to the well-known symbols table (such that the symbol is automatically allowed to be held weakly)
> 
> Looks good overall though!

The AO for determining if a symbol can be held weakly is done by exclusion, so IntlFallbackSymbol will be allowed implicitly. It would only be disallowed if an implementation had incorrectly implemented it using `Symbol.for`, which does not appear to be the case for the implementations I've checked.

https://github.com/v8/v8/blob/5fe0aa3bc79c0a9d3ad546b79211f07105f09585/src/init/heap-symbols.h#L457

https://searchfox.org/mozilla-central/rev/3419858c997f422e3e70020a46baae7f0ec6dacc/js/src/builtin/intl/CommonFunctions.js#735

https://github.com/WebKit/WebKit/blob/84cfeb18efa90b3f67962f66c701f753633c618a/Source/JavaScriptCore/builtins/BuiltinNames.cpp#L44

ljharb:
@acutmore thats a good callout, but a cross-real, non-registry symbol still belongs in that table, and imo it’s an editorial error in 402 that it does not.

mhofman:
> @acutmore thats a good callout, but a cross-realm, non-registry symbol still belongs in that table, and imo it’s an editorial error in 402 that it does not.

I don't believe it's an error. The fallback symbol is not meant to be cross-realm, 402 says explicitly it isn't. However there is no impact[^*] to this proposal since in both cases it could be held weakly.

[^*]: In practice, making `Intl` unreachable in the current realm would potentially make the symbol to itself become unreachable, which could be observed through `WeakRef` / `FinalizationRegistry` if implementations did the work to collect these intrinsics.

spec.html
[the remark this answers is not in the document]
mhofman: Can we move or duplicate the editor's note about well-known symbol under this section.

It's not immediately clear why well-known symbols are not covered by the second bullet point. While some well-known symbols are syntactically undeniable, technically an execution where the deniable well-known symbols were removed from the `Symbol` global and where no new realm could be created wouldn't allow their observation.

spec.html
```diff
@@ -12491,6 +12494,29 @@ <h1>
...
+    <emu-alg>
+      1. If Type(_v_) is Object, return *true*.
+      1. If Type(_v_) is Symbol, then
+        1. For each element _e_ of the GlobalSymbolRegistry List (see <emu-xref href="#sec-symbol.for"></emu-xref>), do
+          1. If SameValue(_e_.[[Symbol]], _v_) is *true*, return *false*.
+        1. Return *true*.
+      1. Return *false*.
+    </emu-alg>
+    <emu-note type="editor">
+      <p>Symbols in the GlobalSymbolRegistry (<emu-xref href="#sec-symbol.for"></emu-xref>) are not considered to have a suitable identity because they are permanently live. Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are also permanently live, but are nonetheless treated as suitable for weak reference because they are limited to elements of a fixed size list and therefore manageable by a variety of implementation approaches. However, using a well-known symbol as a key in a WeakMap is likely to &ldquo;leak&rdquo; memory resources in many implementations.</p>
```
mhofman, line 12517: According to the 3rd bullet point under the definition of liveness, it's not that well-known symbols are "likely" to leak, it's that they will leak as specified.

mhofman (commented):
This all looks good to me, minus a confusing bit about well-known symbols liveness.

Given we increased the usage of the word "referent", I'm wondering if we should have a definition somewhere that would be automatically linked to explaining it's any object or symbol value used as a WeakRef or FinalizationRegistry target.

I wish we had a more generic word to use that also applied to WeakMap keys / WeakSet values to avoid this "object/symbol" dance. Unfortunately @gibson042 "language value" future-proofing is only safe to do where the context already restricts these values in the first place.

spec.html
acutmore: I think the term "leak" is vague enough that "likely" is justified. The value in the weakMap would be held onto until either the key is removed or until the weakMap itself is collected - and there are a small number of well-known symbols. So the leak is 'bound' in that regard. Compared to a leak in an app where a dialog is open and closed repeatedly and the memory can keep increasing until OOM; that style of leak feels different to the type of leak that could be described by using a well-known symbol as a weakMap key.

spec.html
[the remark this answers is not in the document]
acutmore: FWIW well-known symbols can't be removed from the global `Symbol`, they are non-writable and non-configurable. To deny access the whole `Symbol` object needs to be denied and any other operation that could expose the symbol.

```js
// first run before all other code:
delete Symbol.prototype.constructor;
delete gbloalThis.Symbol;
delete Array.prototype.concat;
delete globalThis.ShadowRealm; // and all other ways to construct a new realm
// No longer an execution that could observe the identity of `Symbol.isConcatSpreadable`
```

There is the question of how much flexibility to give engines here. In the above situation should they be allowed to GC `@@isConcatSpreadable`. The idea behind the 3rd bullet point (saying that well-known symbols are always considered live) is to deny that ability to ensure consistency across engines and avoid ambiguity in the spec. Though I can appreciate if engines would rather have the freedom.

cc: @syg @codehag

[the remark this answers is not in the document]
acutmore: Thanks @littledan , that's a good point. If we start to specifically mention well-known symbol here it starts to raise the question why other thing are/are-not also specified.

I've backed out the normative change. I'll re-add later as an editors note.

spec.html
mhofman: Could we rephrase to something like: 
> However any values associated to a well-known symbol through a WeakMap will be kept live.

mhofman: Since well-known symbols are no longer explicitly live maybe the following wording is more appropriate.

```suggestion
      <p>Symbols in the GlobalSymbolRegistry (<emu-xref href="#sec-symbol.for"></emu-xref>) are not considered to have a suitable identity because they are permanently live. Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are likely to never be collected, but are nonetheless treated as suitable for weak reference because they are limited to elements of a fixed size list and therefore manageable by a variety of implementation approaches. However, any values associated to a well-known symbol in a live WeakMap is unlikely to be collected and will &ldquo;leak&rdquo; memory resources in many implementations.</p>
```

acutmore:
> To capture @erights's comment during the presentation, we would like to have well-known symbols be normatively described as an invariant along the lines of:
> 
> > The string properties of the global `%Symbol%` that have a value with a symbol type.
> 
> Let me know if you want me to open a dedicated issue for this.

Opening an issue on the ecma262 repo sounds like the best place to capture that.

spec.html
acutmore: Given the current definition of liveness, well-known symbols technically are not permanently live. It is possible to be in a state where the well-known symbols are not in the KeptAlive list and there does not exist a  future hypothetical WeakRef-oblivious execution that could observe the value of the symbol.

```js
// first run before all other code:
delete Symbol.prototype.constructor;
delete gbloalThis.Symbol;
delete Array.prototype.concat;
delete globalThis.ShadowRealm; // and all other ways to construct a new realm
delete globalThis.Proxy; // not needed but for good measure
// No longer an execution that could observe the identity of `Symbol.isConcatSpreadable`
```

That said I do not ever seeing engines putting in the work to try and collect them, it would be too small a benefit. I'll update the notes to make it clear that they are most likely always live, while engines are technically free to collect them if they so chose.

spec.html
```diff
@@ -12491,6 +12491,29 @@ <h1>
...
+    <emu-alg>
+      1. If Type(_v_) is Object, return *true*.
+      1. If Type(_v_) is Symbol, then
+        1. For each element _e_ of the GlobalSymbolRegistry List (see <emu-xref href="#sec-symbol.for"></emu-xref>), do
+          1. If SameValue(_e_.[[Symbol]], _v_) is *true*, return *false*.
+        1. Return *true*.
+      1. Return *false*.
+    </emu-alg>
+    <emu-note type="editor">
+      <p>Symbols in the GlobalSymbolRegistry (<emu-xref href="#sec-symbol.for"></emu-xref>) are not considered to have a suitable identity because they are permanently live. Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are likely to never be collected, but are nonetheless treated as suitable for weak reference because they are limited to elements of a fixed size list and therefore manageable by a variety of implementation approaches. However, any values associated to a well-known symbol in a live WeakMap is unlikely to be collected and could &ldquo;leak&rdquo; memory resources in many implementations.</p>
```
gibson042, line 12514: I'm sorry to keep harping on this, but I think it's important. What sense of "permanently live" applies to symbols in the GlobalSymbolRegistry but not also to well-known symbols? Members of the former set can _technically_ be collected if they are disconnected **and** any means of creating a new realm is removed, and so can most (but not all!) members of the latter set—and it's the aspect relating to new realms that is subtle and therefore motivating special mention here.

```suggestion
      <p>Symbols in the GlobalSymbolRegistry (<emu-xref href="#sec-symbol.for"></emu-xref>) are not considered to have a suitable identity because they are effectively permanently live (in the sense that their identity can still be observed even after no references to them remain). Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are also effectively permanently live, but are nonetheless treated as suitable for weak reference because they are limited to elements of a fixed size list and therefore manageable by a variety of implementation approaches. However, any values associated to a well-known symbol in a live WeakMap is unlikely to be collected and could &ldquo;leak&rdquo; memory resources in many implementations.</p>
```

acutmore: > I'm sorry to keep harping on this, but I think it's important. What sense of "permanently live" applies to symbols in the GlobalSymbolRegistry but not also to well-known symbols

No worries! It's good to get this text right. I think a key thing the text is not capturing is that registered-symbols are designed so that engines are free to implement the GlobalSymbolRegistry as a weakMap.

```js
// no other code has run
let s = Symbol.for('hello');
s = null;
// engines can now collect s from the table too, as nothing would observe that happening
s = Symbol.for('hello'); // this can actually be a fresh new symbol, no one would know
```

(This same optimisation technically applies to the well-known ones, but is less appetising to implement as there are a fixed number of them. Rather than one for each possible string value.)

The description of a WeakMap says that it doesn't extend the lifetime of it's key. But that wouldn't be true for registered symbols where the engine uses a weak GlobalSymbolTable. If Userland tries to hold it weakly, the engine can't also invisibly do it. So instead whenever Userland tries to hold it weakly the engine must ensure it is actually held strongly. That is the core important detail of why registered symbols are not allowed, but well-known ones are. That gives engines the most flexibility.

gibson042: Right, but those are implementation choices rather than normative specification constraints. From the perspective of ECMAScript code running in a conforming implementation, well-known and registered symbols have equivalent special behavior with respect to resurrection, and this proposal makes that special behavior relevant in a way that it was not previously (as the last paragraph of your comment describes).

mhofman: Let's consider the following hypothetical execution:
- Create a registered symbol with `Symbol.for`
- Remove global `Symbol` and `ShadowRealm` (and other realm creating host APIs), and drop all existing reference to them and `Symbol.for` (if any).
- Create a WeakRef targeting earlier registered symbol (assuming it wasn't prevented by the spec) and drop reference to the symbol.

At this point the WeakRef would be allowed by the spec to be cleared because there is no `WeakRef`-oblivious execution where the identity of the symbol can be observed.

However, the important part here is that the GlobalSymbolRegistry is reified as `Symbol.for()` which I think makes it more tangible to programs and appears as a clear way of "resuscitating" symbols, where well-known symbols being plain properties feels less coordinated. While, both are subject to leak if used as WeakMap key, and technically both could be eligible for collection (minus some syntax undeniable well-known symbols), in my mind it's ok for only registered symbols to be explicitly prevented from usage a WMK because of this difference in tangibility.

gibson042: I'm not talking about whether it is _acceptable_ for registered vs. well-known symbols to be supported as weakly-held entities, I'm just asking for explicit acknowledgement in the spec text that both have the same reference-independent liveness behavior that requires special care with respect to such support.

gibson042: I would say both. Implementers **must** be aware of the special behavior in order to conform with the specification, and practitioners **should** be aware of it before using well-known symbols in weak contexts (particularly as WeakMap keys).

ptomato:
This proposal has test262 tests now.

ljharb:
@acutmore it may be time to rebase this :-)

spec.html
```diff
@@ -12012,8 +12012,8 @@ <h1>Agents</h1>
         </tr>
         <tr>
           <td>[[KeptAlive]]</td>
-          <td>a List of Objects</td>
-          <td>Initially a new empty List, representing the list of objects to be kept alive until the end of the current Job</td>
+          <td>a List of ECMAScript language values</td>
```
ljharb, line 12015:
```suggestion
          <td>a List of either Objects or Symbols</td>
```

spec.html
```diff
@@ -12297,6 +12297,29 @@ <h1>
...
+    <dl class="header">
+      <dt>description</dt>
+      <dd>It returns *true* if and only if _v_ is considered to have an identity suitable for a weak reference. i.e. Is _v_ a valid candidate key of a WeakMap, or entry in a WeakSet, or target of a WeakRef or FinalizationRegistry.</dd>
+    </dl>
+    <emu-alg>
+      1. If Type(_v_) is an Object, return *true*.
+      1. If Type(_v_) is a Symbol, then
+        1. For each element _e_ of the GlobalSymbolRegistry List (see <emu-xref href="#sec-symbol.for"></emu-xref>), do
+          1. If SameValue(_e_.[[Symbol]], _v_) is *true*, return *false*.
+        1. Return *true*.
```
ljharb, line 12316: it may be time to make an AO for this, since the loop happens in two places? (and hopefully soon three, if symbol predicates advances)

spec.html
```diff
@@ -12155,33 +12155,33 @@ <h1>Objectives</h1>
...
-        WeakRef-obliviousness is defined on sets of objects instead of individual objects to account for cycles. If it were defined on individual objects, then an object in a cycle will be considered live even though its Object value is only observed via WeakRefs of other objects in the cycle.
+        WeakRef-obliviousness is defined on sets of objects or symbols instead of individual referents to account for cycles. If it were defined on individual values, then a referent in a cycle will be considered live even though its value is only observed via WeakRefs of other referents in the cycle.
       </emu-note>
       <emu-note>
-        Colloquially, we say that an individual object is live if every set of objects containing it is live.
+        Colloquially, we say that an individual object or symbol is live if every set of objects containing it is live.
       </emu-note>
 
-      <p>At any point during evaluation, a set of objects _S_ is considered <dfn>live</dfn> if either of the following conditions is met:</p>
+      <p>At any point during evaluation, a set of objects and/or symbols, _S_, is considered <dfn>live</dfn> if either of the following conditions is met:</p>
```
michaelficarra, line 12170: I wouldn't use these commas.

```suggestion
      <p>At any point during evaluation, a set of objects and/or symbols _S_ is considered <dfn>live</dfn> if either of the following conditions is met:</p>
```

Same for above on line 12158.

spec.html
```diff
@@ -12155,33 +12155,33 @@ <h1>Objectives</h1>
     <emu-clause id="sec-liveness">
       <h1>Liveness</h1>
 
-      <p>For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref of a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
+      <p>For some set of objects and/or symbols, _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref of a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
```
michaelficarra, line 12158:
```suggestion
      <p>For some set of objects and/or symbols _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref of a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
```

I decided to leave a suggestion to clarify what I meant.

spec.html
```diff
@@ -12297,6 +12297,29 @@ <h1>
...
+      CanBeHeldWeakly (
+        _v_: unknown,
+      ): a Boolean
+    </h1>
+    <dl class="header">
+      <dt>description</dt>
+      <dd>It returns *true* if and only if _v_ is considered to have an identity suitable for a weak reference. i.e. Is _v_ a valid candidate key of a WeakMap, or entry in a WeakSet, or target of a WeakRef or FinalizationRegistry.</dd>
+    </dl>
+    <emu-alg>
+      1. If Type(_v_) is an Object, return *true*.
```
michaelficarra, line 12312:
```suggestion
      1. If _v_ is an Object, return *true*.
```

spec.html
```diff
@@ -12297,6 +12297,29 @@ <h1>
...
+      1. If Type(_v_) is a Symbol, then
```
michaelficarra, line 12313:
```suggestion
      1. If _v_ is a Symbol, then
```

spec.html
```diff
@@ -12297,6 +12297,29 @@ <h1>
...
+        1. For each element _e_ of the GlobalSymbolRegistry List (see <emu-xref href="#sec-symbol.for"></emu-xref>), do
+          1. If SameValue(_e_.[[Symbol]], _v_) is *true*, return *false*.
+        1. Return *true*.
+      1. Return *false*.
+    </emu-alg>
+    <emu-note type="editor">
+      <p>Symbols in the GlobalSymbolRegistry (<emu-xref href="#sec-symbol.for"></emu-xref>) are not considered to have a suitable identity because they are permanently live. Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are likely to never be collected, but are nonetheless treated as suitable for weak reference because they are limited to elements of a fixed size list and therefore manageable by a variety of implementation approaches. However, any values associated to a well-known symbol in a live WeakMap is unlikely to be collected and could “leak” memory resources in many implementations.</p>
```
michaelficarra, line 12320: I don't like the phrase "not considered to have a suitable identity". Can't we just say "are not considered suitable for use as a WeakMap key"?

spec.html
```diff
@@ -12297,6 +12297,29 @@ <h1>
       1. Return ~unused~.
     </emu-alg>
   </emu-clause>
+
+  <emu-clause id="sec-canbeheldweakly-abstract-operation" type="abstract operation">
```
jmdyck, line 12301:
```suggestion
  <emu-clause id="sec-canbeheldweakly" type="abstract operation">
```

spec.html
```diff
@@ -12297,6 +12297,29 @@ <h1>
...
+    <h1>
+      CanBeHeldWeakly (
+        _v_: unknown,
```
jmdyck, line 12304:
```suggestion
        _v_: an ECMAScript language value,
```

jmdyck (changes requested):

spec.html
```diff
@@ -12155,33 +12155,33 @@ <h1>Objectives</h1>
...
+      <p>For some set of objects and/or symbols _S_ a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref of a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
 
       <emu-note>
-        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep an object alive. Two, that cycles in liveness does not imply that an object is live. To be concrete, if determining _obj_'s liveness depends on determining the liveness of another WeakRef referent, _obj2_, _obj2_'s liveness cannot assume _obj_'s liveness, which would be circular reasoning.
+        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep a referent alive. Two, that cycles in liveness does not imply that a referent is live. To be concrete, if determining _ref_'s liveness depends on determining the liveness of another WeakRef referent, _ref2_, _ref2_'s liveness cannot assume _ref_'s liveness, which would be circular reasoning.
```
syg, line 12161:
```suggestion
        WeakRef-obliviousness, together with liveness, capture two notions. One, that a WeakRef itself does not keep its referent alive. Two, that cycles in liveness does not imply that a value is live. To be concrete, if determining _ref_'s liveness depends on determining the liveness of another WeakRef referent, _ref2_, _ref2_'s liveness cannot assume _ref_'s liveness, which would be circular reasoning.
```

I find "a referent" reads oddly, since it makes me wonder "whose referent?" when used with an indefinite article.

Here and below, when using the indefinite article (that is, when not talking about a particular WeakRef's referent), I propose we either use "value" as you do elsewhere, or "WeakRef referent" explicitly. "WeakRef target" would probably be even clearer.

spec.html
```diff
@@ -12155,33 +12155,33 @@ <h1>Objectives</h1>
...
       </emu-note>
       <emu-note>
-        WeakRef-obliviousness is defined on sets of objects instead of individual objects to account for cycles. If it were defined on individual objects, then an object in a cycle will be considered live even though its Object value is only observed via WeakRefs of other objects in the cycle.
+        WeakRef-obliviousness is defined on sets of objects or symbols instead of individual referents to account for cycles. If it were defined on individual values, then a referent in a cycle will be considered live even though its value is only observed via WeakRefs of other referents in the cycle.
```
syg, line 12164:
```suggestion
        WeakRef-obliviousness is defined on sets of objects or symbols instead of individual referents to account for cycles. If it were defined on individual values, then a referent in a cycle will be considered live even though its identity is only observed via WeakRefs of other referents in the cycle.
```

I guess we can use identity now that we've defined it?

spec.html
```diff
@@ -12297,6 +12297,26 @@ <h1>
...
+      <dt>description</dt>
+      <dd>It returns *true* if and only if _v_ is considered to have an identity suitable for a weak reference. i.e. Is _v_ a valid candidate key of a WeakMap, or entry in a WeakSet, or target of a WeakRef or FinalizationRegistry.</dd>
+    </dl>
+    <emu-alg>
+      1. If _v_ is an Object, return *true*.
+      1. If _v_ is a Symbol and KeyForSymbol(_v_) is *undefined*, return *true*.
+      1. Return *false*.
+    </emu-alg>
+    <emu-note type="editor">
+      <p>Symbols in the GlobalSymbolRegistry (<emu-xref href="#sec-symbol.for"></emu-xref>) are not considered suitable for use as a WeakMap key because they are permanently live. Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are likely to never be collected, but are nonetheless treated as suitable for weak reference because they are limited to elements of a fixed size list and therefore manageable by a variety of implementation approaches. However, any values associated to a well-known symbol in a live WeakMap is unlikely to be collected and could “leak” memory resources in many implementations.</p>
```
syg, line 12317: "Permanently live" doesn't quite convey the reason. Strictly speaking, since Symbol.for symbols aren't usable as weak targets, they don't participate in the definition of liveness anyway. Now that we've defined identity, I'm inclined to just say it's because they don't have identity.

spec.html
bakkot: I would say something like "are not considered suitable for use in WeakMaps or WeakRefs because such a symbol can be manifested without already having a reference to it", or something to that effect, rather than referring to liveness, collection, or identity.

spec.html
acutmore: Updated, and also updated the identity section to not say that all symbols have identity.

spec.html
```diff
@@ -30326,10 +30346,7 @@ <h1>Symbol.keyFor ( _sym_ )</h1>
         <p>This function performs the following steps when called:</p>
         <emu-alg>
           1. If _sym_ is not a Symbol, throw a *TypeError* exception.
-          1. For each element _e_ of the GlobalSymbolRegistry List (see <emu-xref href="#sec-symbol.for"></emu-xref>), do
-            1. If SameValue(_e_.[[Symbol]], _sym_) is *true*, return _e_.[[Key]].
-          1. Assert: GlobalSymbolRegistry does not currently contain an entry for _sym_.
-          1. Return *undefined*.
+          1. return KeyForSymbol(_sym_).
```
jmdyck, line 30349:
```suggestion
          1. Return KeyForSymbol(_sym_).
```

jmdyck (changes requested):

spec.html
```diff
@@ -12014,8 +12014,8 @@ <h1>Agents</h1>
         </tr>
         <tr>
           <td>[[KeptAlive]]</td>
-          <td>a List of Objects</td>
-          <td>Initially a new empty List, representing the list of objects to be kept alive until the end of the current Job</td>
+          <td>a List of either Objects or Symbols</td>
```
acutmore, line 12017: A conversation with ChatGPT:


> Me: In the phrase: "A bag of either apples or oranges" does this mean that the bag can contain either only applies or only oranges? Or does it mean that the bag can contain a mix of both apples and oragnes?
>
> AI: The phrase "A bag of either apples or oranges" means that the bag can contain either only apples or only oranges, not a mix of both.
>
> Me: are you sure?
>
> AI: It is possible that the phrase "A bag of either apples or oranges" is being used to indicate that the bag could contain a mix of both apples and oranges, however, without more context it is impossible to know for certain. The phrase "A bag of either apples or oranges" is often used to indicate that the bag can only contain one type of fruit, either apples or oranges.

Would any of the following be clearer?

 - `a List of ECMAScript Language values`
 - `a List of Objects and/or Symbols`
 - `a List of Objects and Symbols`
 
 CC: @jmdyck

jmdyck: > > The phrase "A bag of either apples or oranges" is often used to indicate that the bag can only contain one type of fruit, either apples or oranges.

Yeah, that's probably fair. 
 
> Would any of the following be clearer?
> 
>     * `a List of ECMAScript Language values`

Correct, but imprecise.
 
>     * `a List of Objects and/or Symbols`

It's a possibility.

>     * `a List of Objects and Symbols`

In the apples+oranges analogy, "a bag of apples and oranges" suggests it has some of each kind of fruit. So it wouldn't be clear that (e.g.) a List of only objects would satisfy the description.

---
The status quo isn't entirely consistent on how to describe such lists:

4 times `a List of either X or Y`:
- 16.2.1.6.2 (return type): `a List of either Strings or *null*`
- 22.2.7.8 (type of _indices_): `a List of either Match Records or *undefined*`
- 22.2.7.8 (type of _groupNames_): `a List of either Strings or *undefined*`
- 29.5.4 (type of _Ws_): `a List of either WriteSharedMemory or ReadModifyWriteSharedMemory events`

1 time `a List of X or Y` (but see below re ambiguity):
- 29.5.5 (assertion on _Ws_): `a List of WriteSharedMemory or ReadModifyWriteSharedMemory events ...`

2 times `a List ... whose elements ...`
- 22.1.3.18.1 (type of _captures_): `a possibly empty List, each of whose elements is a String or *undefined*`
- 29.6.2 (condition): `a List ... whose elements are WriteSharedMemory or ReadModifyWriteSharedMemory events ...`

2 times break it out as a separate sentence (which isn't much of an option in the above contexts):
- 6.1.7.3 ([[OwnPropertyKeys]]): `The Type of each element of the returned List is either String or Symbol.`
- 10.5.11 (Note re [[OwnPropertyKeys]]): `The Type of each result List element is either String or Symbol.`

So my advice is to use `a List of either Objects or Symbols` for this PR, and then raise an issue/PR to establish an editorial convention on this question.

----

Note that, in general, `a List of X or Y` can be ambiguous.
E.g., `a List of Strings or *undefined*` could mean:
- `a List of (Strings or *undefined*)`
or
- `(a List of Strings) or *undefined*`

The spec uses 'either' to distinguish these two meanings:
- `a List of either Strings or *undefined*` (the type of _groupNames_ noted above)
vs
- `either a List of Strings or *undefined*` (the type of [[PropertyList]] in 25.5.2.1)

acutmore:
This proposal is now stage 4: https://github.com/tc39/proposals/commit/d2a5cf370e4a9dc8b0249ad8a8386cd1280db815

spec.html
```diff
@@ -12157,33 +12157,33 @@ <h1>Objectives</h1>
...
       </emu-note>
       <emu-note>
-        WeakRef-obliviousness is defined on sets of objects instead of individual objects to account for cycles. If it were defined on individual objects, then an object in a cycle will be considered live even though its Object value is only observed via WeakRefs of other objects in the cycle.
+        WeakRef-obliviousness is defined on sets of objects or symbols instead of individual values to account for cycles. If it were defined on individual values, then a WeakRef referent in a cycle will be considered live even though its identity is only observed via other WeakRef referents in the cycle.
       </emu-note>
       <emu-note>
-        Colloquially, we say that an individual object is live if every set of objects containing it is live.
+        Colloquially, we say that an individual object or symbol is live if every set of objects containing it is live.
```
bakkot, line 12169:
```suggestion
        Colloquially, we say that an individual object or symbol is live if every set containing it is live.
```

I'm also fine being more verbose, as in

```suggestion
        Colloquially, we say that an individual object or symbol is live if every set of objects and/or symbols containing it is live.
```
but it seems like overkill; the universe of values to which go in these sets ought to be understood at this point.

spec.html
```diff
@@ -12157,33 +12157,33 @@ <h1>Objectives</h1>
...
-      <p>At any point during evaluation, a set of objects _S_ is considered <dfn>live</dfn> if either of the following conditions is met:</p>
+      <p>At any point during evaluation, a set of objects and/or symbols _S_ is considered <dfn>live</dfn> if either of the following conditions is met:</p>
 
       <ul>
         <li>
           Any element in _S_ is included in any agent's [[KeptAlive]] List.
         </li>
         <li>
-          There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes the Object value of any object in _S_.
+          There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes any value in _S_.
```
bakkot, line 12179:
```suggestion
          There exists a valid future hypothetical WeakRef-oblivious execution with respect to _S_ that observes the identity of any value in _S_.
```

spec.html
```diff
@@ -12299,6 +12299,26 @@ <h1>
...
+    <dl class="header">
+      <dt>description</dt>
+      <dd>It returns *true* if and only if _v_ is considered to have an identity suitable for a weak reference. i.e. Is _v_ a valid candidate key of a WeakMap, or entry in a WeakSet, or target of a WeakRef or FinalizationRegistry.</dd>
+    </dl>
+    <emu-alg>
+      1. If _v_ is an Object, return *true*.
+      1. If _v_ is a Symbol and KeyForSymbol(_v_) is *undefined*, return *true*.
+      1. Return *false*.
+    </emu-alg>
+    <emu-note type="editor">
```
jmdyck, line 12318: I'm guessing there shouldn't be editor's notes at Stage 4. Convert to a regular note, or drop it?

jmdyck (changes requested):
Looks okay to me otherwise.

spec.html
```diff
@@ -12299,6 +12299,26 @@ <h1>
...
+
+  <emu-clause id="sec-canbeheldweakly" type="abstract operation">
+    <h1>
+      CanBeHeldWeakly (
+        _v_: an ECMAScript language value,
+      ): a Boolean
+    </h1>
+    <dl class="header">
+      <dt>description</dt>
+      <dd>It returns *true* if and only if _v_ is considered to have an identity suitable for a weak reference. i.e. Is _v_ a valid candidate key of a WeakMap, or entry in a WeakSet, or target of a WeakRef or FinalizationRegistry.</dd>
```
jmdyck, line 12311: Syntax seems a bit rough around "i.e. Is". Maybe something like:
```suggestion
      <dd>It returns *true* if and only if _v_ is considered to have an identity suitable for a weak reference. This determines whether _v_ can be a key of a WeakMap, or entry in a WeakSet, or target of a WeakRef or FinalizationRegistry.</dd>
```

bakkot (approved):

jmdyck:
@acutmore: This PR has merge conflicts (from PR #2947, it looks like). If you resolve those, I can take another look.

spec.html
```diff
@@ -12296,6 +12296,26 @@ <h1>
...
+    </dl>
+    <emu-alg>
+      1. If _v_ is an Object, return *true*.
+      1. If _v_ is a Symbol and KeyForSymbol(_v_) is *undefined*, return *true*.
+      1. Return *false*.
+    </emu-alg>
+    <emu-note>
+      <p>Symbols in the GlobalSymbolRegistry (<emu-xref href="#sec-symbol.for"></emu-xref>) are not considered suitable for use as a WeakMap key because they can be described by the String used to register them, and therefore do not have <emu-xref href="#sec-identity">identity</emu-xref>. Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are likely to never be collected, but are nonetheless treated as suitable for weak reference because they are limited to elements of a fixed size list and therefore manageable by a variety of implementation approaches. However, any values associated to a well-known symbol in a live WeakMap is unlikely to be collected and could “leak” memory resources in many implementations.</p>
```
jmdyck, line 12316: The phrase "they are limited to elements of a fixed size list" seems wordier than necessary (unless I'm missing the point). Maybe:
```suggestion
      <p>Symbols in the GlobalSymbolRegistry (<emu-xref href="#sec-symbol.for"></emu-xref>) are not considered suitable for use as a WeakMap key because they can be described by the String used to register them, and therefore do not have <emu-xref href="#sec-identity">identity</emu-xref>. Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are likely to never be collected, but are nonetheless treated as suitable for weak reference because they are limited in number and therefore manageable by a variety of implementation approaches. However, any values associated to a well-known symbol in a live WeakMap is unlikely to be collected and could “leak” memory resources in many implementations.</p>
```

jmdyck (commented):
Looks okay to me.

michaelficarra (commented):
I've opened https://github.com/acutmore/ecma262/pull/3 with some changes we'll discuss in the 262 editor call tomorrow, but otherwise LGTM.

syg:
@michaelficarra and I have chatted about our disagreement of the use of "identity" for the concept of "can be used as a weak target", and we're agreed on stratifying the concept of identity into "specification identity" and "language identity", with "language identity" being the concept that this proposal should use.

This PR should block on https://github.com/tc39/ecma262/pull/3027 which does the stratification.

syg (approved):

spec.html
```diff
@@ -12297,6 +12297,26 @@ <h1>
...
+    </dl>
+    <emu-alg>
+      1. If _v_ is an Object, return *true*.
+      1. If _v_ is a Symbol and KeyForSymbol(_v_) is *undefined*, return *true*.
+      1. Return *false*.
+    </emu-alg>
+    <emu-note>
+      <p>Language values without <emu-xref href="#sec-identity">language identity</emu-xref> can be manifested without prior reference and are unsuitable for use as weak references. Symbols produced by <emu-xref href="#sec-symbol.for">Symbol.for</emu-xref>, unlike other Symbol values, do not have language identity and are unsuitable for use as weak references. Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are likely to never be collected, but are nonetheless treated as suitable for weak reference because they are limited in number and therefore manageable by a variety of implementation approaches. However, any values associated to a well-known symbol in a live WeakMap is unlikely to be collected and could “leak” memory resources in implementations.</p>
```
michaelficarra, line 12317: A few minor nits.

```suggestion
      <p>Language values without <emu-xref href="#sec-identity">language identity</emu-xref> can be manifested without prior reference and are unsuitable for use as a weak reference. Symbols produced by <emu-xref href="#sec-symbol.for">Symbol.for</emu-xref>, unlike other Symbol values, do not have language identity and are unsuitable for use as a weak reference. Well-known symbols (<emu-xref href="#sec-well-known-symbols"></emu-xref>) are likely to never be collected, but are nonetheless treated as suitable for use as a weak reference because they are limited in number and therefore manageable by a variety of implementation approaches. However, any value associated to a well-known symbol in a live WeakMap is unlikely to be collected and could leak memory resources in implementations.</p>
```

syg: "Values are unsuitable for use as a weak reference" sounds weird to me. Why do you prefer that over plural in both cases, as it is now?

syg: Anyway I don't have much of a preference for plural or singular, but the "value" part and "as weak reference" part should agree.

michaelficarra (approved):

PR #2780, opened 2022-05-18
Editorial: Replace "legal" with more precise language

gibson042:
Also includes minor cleanup of the "Identifier Names" section, which included some inaccuracies (e.g., implication that every [|UnicodeEscapeSequence|](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-UnicodeEscapeSequence) includes a |CodePoint| and that components of source text are code units rather than code points) and some clunky repetitions.


syg:
@gibson042 WDYT about saying "conformant" instead of having ad-hoc phrases?

gibson042:
@syg I support that in general, but honestly I think it would be more awkward for the four specific changes of this PR.

spec.html
```diff
@@ -46496,7 +46496,7 @@ <h1>Shared Memory Guidelines</h1>
         <li>Non-lock-free atomics compile to a spinlock acquire, a full fence, a series of non-atomic load and store instructions, a full fence, and a spinlock release.</li>
       </ul>
       <p>That mapping is correct so long as an atomic operation on an address range does not race with a non-atomic write or with an atomic operation of different size. However, that is all we need: the memory model effectively demotes the atomic operations involved in a race to non-atomic status. On the other hand, the naive mapping is quite strong: it allows atomic operations to be used as sequentially consistent fences, which the memory model does not actually guarantee.</p>
-      <p>A number of local improvements to those basic patterns are also intended to be legal:</p>
+      <p>A number of local improvements to those basic patterns are also intended to be available:</p>
```
syg, line 46499: Ditto here.

spec.html
```diff
@@ -12406,7 +12406,7 @@ <h1>Execution</h1>
       </emu-alg>
 
       <emu-note>
-        <p>Together with the definition of liveness, this clause prescribes legal optimizations that an implementation may apply regarding WeakRefs.</p>
+        <p>Together with the definition of liveness, this clause prescribes optimizations that an implementation may apply regarding WeakRefs.</p>
```
syg, line 12409: When describing optimization prescription or proscription, "legal" is pretty widely used and understood vernacular. Can you expand on why you think it's not precise?

syg: I don't understand the inclusion concerns point, but I'm also fine with other single word descriptors like "conformant" or "valid" or "allowed" when describing optimizations. I do think a word like "available" actually makes it less precise.

ljharb: I don't agree that "legal" strictly refers to matters of law; games have legal/illegal moves, for example.

gibson042: > games have legal/illegal moves, for example

That is precisely the kind of **loose** linguistic extension that I referred to above; it does not change the **strict** "_of or pertaining to the law_" denotation.

Regardless, I have now changed the text of the note in [Shared Memory Guidelines](https://tc39.es/ecma262/multipage/memory-model.html#sec-shared-memory-guidelines) to use "allowed" rather than "legal" for better alignment with notes in e.g. [WeakRef/FinalizationRegistry Execution](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-weakref-execution) ("_optimizations are thus **allowed** to observably empty WeakRefs_") and [Sequentially Consistent Atomics](https://tc39.es/ecma262/multipage/memory-model.html#sec-memory-order) ("_[events] that are not constrained by happens-before or synchronizes-with are **allowed** to occur anywhere in the order_").

spec.html
```diff
@@ -46496,7 +46496,7 @@ <h1>Shared Memory Guidelines</h1>
         <li>Non-lock-free atomics compile to a spinlock acquire, a full fence, a series of non-atomic load and store instructions, a full fence, and a spinlock release.</li>
       </ul>
       <p>That mapping is correct so long as an atomic operation on an address range does not race with a non-atomic write or with an atomic operation of different size. However, that is all we need: the memory model effectively demotes the atomic operations involved in a race to non-atomic status. On the other hand, the naive mapping is quite strong: it allows atomic operations to be used as sequentially consistent fences, which the memory model does not actually guarantee.</p>
-      <p>A number of local improvements to those basic patterns are also intended to be legal:</p>
+      <p>Local improvements to those basic patterns are also allowed and expected, subject to the constraints of the memory model. For example:</p>
```
syg, line 46499:
```suggestion
      <p>Local improvements to those basic patterns are also allowed, subject to the constraints of the memory model. For example:</p>
```

syg (approved):

spec.html
gibson042: Done.

michaelficarra (approved):

bakkot (approved):
I'm neutral on the first change, fine with the others. I don't want to have this considered precedent to avoid the use of the word "legal", though, if there happen to be other cases which arise where it seems like the most suitable word - I don't agree that it refers strictly to matters of law.

PR #2781, opened 2022-05-20
Normative: Use DefaultTimeZone to get the local time zone for Date

ptomato:
Currently, it is technically possible for the following two operations to
result in different local times and different time zones:

```js
> date = new Date()
> date.toTimeString()
'13:43:47 GMT-0700 (Pacific Daylight Time)'
> new Intl.DateTimeFormat(undefined, {timeStyle: 'full'}).format(date)
'1:43:47 p.m. Pacific Daylight Time'
```

This is because Date gets its notion of the current system time zone from
the operation LocalTZA, while Intl.DateTimeFormat uses the operation
DefaultTimeZone from ECMA-402 to determine the current system time zone.

This is a normative change because there was previously no requirement for
these to be linked. However, they *should* definitely be linked, and no
implementation that includes ECMA-402 that I'm aware of has a discrepancy
between the two. This patch adds this requirement.

This is accomplished by removing LocalTZA (which was actually two
operations rolled into one, anyway) and redefining the LocalTime and UTC
operations in terms of DefaultTimeZone, which moves into ECMA-262, as well
as the two new operations GetNamedTimeZoneOffsetNanoseconds and
GetNamedTimeZoneEpochNanoseconds.

It essentially pushes the "implementation-definedness" of LocalTZA into
GetNamedTimeZoneOffsetNanoseconds, GetNamedTimeZoneEpochNanoseconds, and
DefaultTimeZone.

Although it is a normative change because it adds the requirement for Date
and Intl.DateTimeFormat to use the same default time zone, it should have
no other effects. Despite specifying what LocalTZA used to do in terms of
different operations, the observable effects are the same.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1063,7 +1063,8 @@
       <p>The notation &ldquo;<emu-eqn id="eqn-modulo" aoid="modulo">_x_ modulo _y_</emu-eqn>&rdquo; (_y_ must be finite and non-zero) computes a value _k_ of the same sign as _y_ (or zero) such that <emu-eqn>abs(_k_) &lt; abs(_y_) and _x_ - _k_ = _q_ &times; _y_</emu-eqn> for some integer _q_.</p>
       <p>The phrase "the result of <dfn id="clamping">clamping</dfn> _x_ between _lower_ and _upper_" (where _x_ is an extended mathematical value and _lower_ and _upper_ are mathematical values such that _lower_ &le; _upper_) produces _lower_ if _x_ &lt; _lower_, produces _upper_ if _x_ &gt; _upper_, and otherwise produces _x_.</p>
       <p>The mathematical function <emu-eqn id="eqn-floor" aoid="floor">floor(_x_)</emu-eqn> produces the largest integer (closest to +&infin;) that is not larger than _x_.</p>
-      <p>Mathematical functions min, max, abs, and floor are not defined for Numbers and BigInts, and any usage of those methods that have non-mathematical value arguments would be an editorial error in this specification.</p>
+      <p>The mathematical function <emu-eqn id="eqn-truncate" aoid="truncate">truncate(_x_)</emu-eqn> removes the fractional part of _x_ by rounding towards zero, producing <emu-eqn>-floor(-_x_)</emu-eqn> if _x_ &lt; 0 and otherwise producing <emu-eqn>floor(_x_)</emu-eqn>.</p>
+      <p>Mathematical functions min, max, abs, floor, and truncate are not defined for Numbers and BigInts, and any usage of those methods that have non-mathematical value arguments would be an editorial error in this specification.</p>
       <emu-note>
         <p><emu-eqn>floor(_x_) = _x_ - (_x_ modulo 1)</emu-eqn>.</p>
       </emu-note>
@@ -32015,7 +32016,7 @@
 
       <emu-clause id="sec-time-values-and-time-range">
         <h1>Time Values and Time Range</h1>
-        <p>Time measurement in ECMAScript is analogous to time measurement in POSIX, in particular sharing definition in terms of the proleptic Gregorian calendar, an epoch of midnight at the beginning of 1 January 1970 UTC, and an accounting of every day as comprising exactly 86,400 seconds (each of which is 1000 milliseconds long).</p>
+        <p>Time measurement in ECMAScript is analogous to time measurement in POSIX, in particular sharing definition in terms of the proleptic Gregorian calendar, an <dfn id="epoch">epoch</dfn> of midnight at the beginning of 1 January 1970 UTC, and an accounting of every day as comprising exactly 86,400 seconds (each of which is 1000 milliseconds long).</p>
         <p>An ECMAScript <dfn variants="time values">time value</dfn> is a Number, either a finite integral Number representing an instant in time to millisecond precision or *NaN* representing no specific instant. A time value that is a multiple of <emu-eqn>24 &times; 60 &times; 60 &times; 1000 = 86,400,000</emu-eqn> (i.e., is equal to 86,400,000 &times; _d_ for some integer _d_) represents the instant at the start of the UTC day that follows the epoch by _d_ whole UTC days (preceding the epoch for negative _d_). Every other finite time value _t_ is defined relative to the greatest preceding time value _s_ that is such a multiple, and represents the instant that occurs within the same UTC day as _s_ but follows it by (_t_ - _s_) milliseconds.</p>
         <p>Time values do not account for UTC leap seconds&mdash;there are no time values representing instants within positive leap seconds, and there are time values representing instants removed from the UTC timeline by negative leap seconds. However, the definition of time values nonetheless yields piecewise alignment with UTC, with discontinuities only at leap second boundaries and zero difference outside of leap seconds.</p>
         <p>A Number can exactly represent all integers from -9,007,199,254,740,992 to 9,007,199,254,740,992 (<emu-xref href="#sec-number.min_safe_integer"></emu-xref> and <emu-xref href="#sec-number.max_safe_integer"></emu-xref>). A time value supports a slightly smaller range of -8,640,000,000,000,000 to 8,640,000,000,000,000 milliseconds. This yields a supported time value range of exactly -100,000,000 days to 100,000,000 days relative to midnight at the beginning of 1 January 1970 UTC.</p>
@@ -32109,43 +32110,141 @@
         <p>A weekday value of *+0*<sub>𝔽</sub> specifies Sunday; *1*<sub>𝔽</sub> specifies Monday; *2*<sub>𝔽</sub> specifies Tuesday; *3*<sub>𝔽</sub> specifies Wednesday; *4*<sub>𝔽</sub> specifies Thursday; *5*<sub>𝔽</sub> specifies Friday; and *6*<sub>𝔽</sub> specifies Saturday. Note that <emu-eqn>WeekDay(*+0*<sub>𝔽</sub>) = *4*<sub>𝔽</sub></emu-eqn>, corresponding to Thursday, 1 January 1970.</p>
       </emu-clause>
 
-      <emu-clause id="sec-local-time-zone-adjustment" type="implementation-defined abstract operation">
+      <emu-clause id="sec-getutcepochnanoseconds" type="abstract operation">
         <h1>
-          LocalTZA (
-            _t_: a Number,
-            _isUTC_: a Boolean,
-          ): an integral Number
+          GetUTCEpochNanoseconds (
+            _year_: an integer,
+            _month_: an integer in the inclusive interval from 1 to 12,
+            _day_: an integer in the inclusive interval from 1 to 31,
+            _hour_: an integer in the inclusive interval from 0 to 23,
+            _minute_: an integer in the inclusive interval from 0 to 59,
+            _second_: an integer in the inclusive interval from 0 to 59,
+            _millisecond_: an integer in the inclusive interval from 0 to 999,
+            _microsecond_: an integer in the inclusive interval from 0 to 999,
+            _nanosecond_: an integer in the inclusive interval from 0 to 999,
+          ): a BigInt
         </h1>
         <dl class="header">
           <dt>description</dt>
-          <dd>Its return value represents the local time zone adjustment, or offset, in milliseconds. The local political rules for standard time and daylight saving time in effect at _t_ should be used to determine the result in the way specified in this section.</dd>
+          <dd>The returned value represents a number of nanoseconds since the epoch that corresponds to the given ISO 8601 calendar date and wall-clock time in UTC.</dd>
         </dl>
-        <p>When _isUTC_ is true, <emu-eqn>LocalTZA( _t_<sub>UTC</sub>, true )</emu-eqn> should return the offset of the local time zone from UTC measured in milliseconds at time represented by time value <emu-eqn>_t_<sub>UTC</sub></emu-eqn>. When the result is added to <emu-eqn>_t_<sub>UTC</sub></emu-eqn>, it should yield the corresponding Number <emu-eqn>_t_<sub>local</sub></emu-eqn>.</p>
-        <p>When _isUTC_ is false, <emu-eqn>LocalTZA( _t_<sub>local</sub>, false )</emu-eqn> should return the offset of the local time zone from UTC measured in milliseconds at local time represented by Number <emu-eqn>_t_<sub>local</sub></emu-eqn>. When the result is subtracted from <emu-eqn>_t_<sub>local</sub></emu-eqn>, it should yield the corresponding time value <emu-eqn>_t_<sub>UTC</sub></emu-eqn>.</p>
-        <p>Input _t_ is nominally a time value but may be any Number value. This can occur when _isUTC_ is false and _t_<sub>local</sub> represents a time value that is already offset outside of the time value range at the range boundaries. The algorithm must not limit _t_<sub>local</sub> to the time value range, so that such inputs are supported.</p>
-        <p>When <emu-eqn>_t_<sub>local</sub></emu-eqn> represents local time repeating multiple times at a negative time zone transition (e.g. when the daylight saving time ends or the time zone offset is decreased due to a time zone rule change) or skipped local time at a positive time zone transitions (e.g. when the daylight saving time starts or the time zone offset is increased due to a time zone rule change), <emu-eqn>_t_<sub>local</sub></emu-eqn> must be interpreted using the time zone offset before the transition.</p>
-        <p>If an implementation does not support a conversion described above or if political rules for time _t_ are not available within the implementation, the result must be *+0*<sub>𝔽</sub>.</p>
+        <emu-alg>
+          1. Let _date_ be MakeDay(𝔽(_year_), 𝔽(_month_ - 1), 𝔽(_day_)).
+          1. Let _time_ be MakeTime(𝔽(_hour_), 𝔽(_minute_), 𝔽(_second_), 𝔽(_millisecond_)).
+          1. Let _ms_ be MakeDate(_date_, _time_).
+          1. Assert: _ms_ is an integral Number.
+          1. Return ℤ(ℝ(_ms_) &times; 10<sup>6</sup> + _microsecond_ &times; 10<sup>3</sup> + _nanosecond_).
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-getnamedtimezoneepochnanoseconds" type="implementation-defined abstract operation">
+        <h1>
+          GetNamedTimeZoneEpochNanoseconds (
+            _timeZoneIdentifier_: a String,
+            _year_: an integer,
+            _month_: an integer in the inclusive interval from 1 to 12,
+            _day_: an integer in the inclusive interval from 1 to 31,
+            _hour_: an integer in the inclusive interval from 0 to 23,
+            _minute_: an integer in the inclusive interval from 0 to 59,
+            _second_: an integer in the inclusive interval from 0 to 59,
+            _millisecond_: an integer in the inclusive interval from 0 to 999,
+            _microsecond_: an integer in the inclusive interval from 0 to 999,
+            _nanosecond_: an integer in the inclusive interval from 0 to 999,
+          ): a List of BigInts
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>
+            Each value in the returned List represents a number of nanoseconds since the epoch that corresponds to the given ISO 8601 calendar date and wall-clock time in the named time zone identified by _timeZoneIdentifier_.
+          </dd>
+        </dl>
+        <p>
+          When the input represents a local time occurring more than once because of a negative time zone transition (e.g. when daylight saving time ends or the time zone offset is decreased due to a time zone rule change), the returned List will have more than one element and will be sorted by ascending numerical value.
+          When the input represents a local time skipped because of a positive time zone transition (e.g. when daylight saving time begins or the time zone offset is increased due to a time zone rule change), the returned List will be empty.
+          Otherwise, the returned List will have one element.
+        </p>
+        <p>The default implementation of GetNamedTimeZoneEpochNanoseconds, to be used for ECMAScript implementations that do not include local political rules for any time zones, performs the following steps when called:</p>
+        <emu-alg>
+          1. Assert: _timeZoneIdentifier_ is *"UTC"*.
+          1. Let _epochNanoseconds_ be GetUTCEpochNanoseconds(_year_, _month_, _day_, _hour_, _minute_, _second_, _millisecond_, _microsecond_, _nanosecond_).
+          1. Return &laquo; _epochNanoseconds_ &raquo;.
+        </emu-alg>
         <emu-note>
           <p>It is recommended that implementations use the time zone information of the IANA Time Zone Database <a href="https://www.iana.org/time-zones/">https://www.iana.org/time-zones/</a>.</p>
-          <p>1:30 AM on 5 November 2017 in America/New_York is repeated twice (fall backward), but it must be interpreted as 1:30 AM UTC-04 instead of 1:30 AM UTC-05. LocalTZA(TimeClip(MakeDate(MakeDay(2017, 10, 5), MakeTime(1, 30, 0, 0))), false) is <emu-eqn>-4 &times; msPerHour</emu-eqn>.</p>
-          <p>2:30 AM on 12 March 2017 in America/New_York does not exist, but it must be interpreted as 2:30 AM UTC-05 (equivalent to 3:30 AM UTC-04). LocalTZA(TimeClip(MakeDate(MakeDay(2017, 2, 12), MakeTime(2, 30, 0, 0))), false) is <emu-eqn>-5 &times; msPerHour</emu-eqn>.</p>
-          <p>Local time zone offset values may be positive <i>or</i> negative.</p>
+          <p>1:30 AM on 5 November 2017 in America/New_York is repeated twice, so GetNamedTimeZoneEpochNanoseconds(*"America/New_York"*, 2017, 11, 5, 1, 30, 0, 0, 0, 0) would return a List of length 2 in which the first element represents 05:30 UTC (corresponding with 01:30 US Eastern Daylight Time at UTC offset -04:00) and the second element represents 06:30 UTC (corresponding with 01:30 US Eastern Standard Time at UTC offset -05:00).</p>
+          <p>2:30 AM on 12 March 2017 in America/New_York does not exist, so GetNamedTimeZoneEpochNanoseconds(*"America/New_York"*, 2017, 3, 12, 2, 30, 0, 0, 0, 0) would return an empty List.</p>
+        </emu-note>
+      </emu-clause>
+
+      <emu-clause id="sec-getnamedtimezoneoffsetnanoseconds" oldids="sec-local-time-zone-adjustment" type="implementation-defined abstract operation">
+        <h1>
+          GetNamedTimeZoneOffsetNanoseconds (
+            _timeZoneIdentifier_: a String,
+            _epochNanoseconds_: a BigInt,
+          ): an integer
+        </h1>
+        <dl class="header">
+        </dl>
+        <p>The returned integer represents the offset from UTC of the named time zone identified by _timeZoneIdentifier_, at the instant corresponding with _epochNanoseconds_ relative to the epoch, both in nanoseconds.</p>
+        <p>The default implementation of GetNamedTimeZoneOffsetNanoseconds, to be used for ECMAScript implementations that do not include local political rules for any time zones, performs the following steps when called:</p>
+        <emu-alg>
+          1. Assert: _timeZoneIdentifier_ is *"UTC"*.
+          1. Return 0.
+        </emu-alg>
+        <emu-note>
+          <p>Time zone offset values may be positive or negative.</p>
+        </emu-note>
+      </emu-clause>
+
+      <emu-clause id="sec-defaulttimezone" type="implementation-defined abstract operation">
+        <h1>DefaultTimeZone ( ): a String</h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It returns a String value representing the host environment's current time zone, which is either a String representing a UTC offset for which IsTimeZoneOffsetString returns *true*, or a String identifier accepted by GetNamedTimeZoneEpochNanoseconds and GetNamedTimeZoneOffsetNanoseconds.</dd>
+        </dl>
+
+        <p>An ECMAScript implementation that includes the ECMA-402 Internationalization API must implement the DefaultTimeZone abstract operation as specified in the ECMA-402 specification.</p>
+        <p>The default implementation of DefaultTimeZone, to be used for ECMAScript implementations that do not include local political rules for any time zones, performs the following steps when called:</p>
+
+        <emu-alg>
+          1. Return *"UTC"*.
+        </emu-alg>
+
+        <emu-note>
+          <p>
+            To ensure the level of functionality that implementations commonly provide in the methods of the Date object, it is recommended that DefaultTimeZone return an IANA time zone name corresponding to the host environment's time zone setting, if such a thing exists.
+            GetNamedTimeZoneEpochNanoseconds and GetNamedTimeZoneOffsetNanoseconds must reflect the local political rules for standard time and daylight saving time in that time zone, if such rules exist.
+          </p>
+          <p>For example, if the host environment is a browser on a system where the user has chosen US Eastern Time as their time zone, DefaultTimeZone returns *"America/New_York"*.</p>
         </emu-note>
       </emu-clause>
 
       <emu-clause id="sec-localtime" type="abstract operation">
         <h1>
           LocalTime (
-            _t_: a time value,
-          ): a Number
+            _t_: a finite time value,
+          ): an integral Number
         </h1>
         <dl class="header">
           <dt>description</dt>
-          <dd>It converts _t_ from UTC to local time.</dd>
+          <dd>
+            It converts _t_ from UTC to local time.
+            The local political rules for standard time and daylight saving time in effect at _t_ should be used to determine the result in the way specified in this section.
+          </dd>
         </dl>
         <emu-alg>
-          1. Return _t_ + LocalTZA(_t_, *true*).
+          1. Let _localTimeZone_ be DefaultTimeZone().
+          1. If IsTimeZoneOffsetString(_localTimeZone_) is *true*, then
+            1. Let _offsetNs_ be ParseTimeZoneOffsetString(_localTimeZone_).
+          1. Else,
+            1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_localTimeZone_, ℤ(ℝ(_t_) &times; 10<sup>6</sup>)).
+          1. Let _offsetMs_ be truncate(_offsetNs_ / 10<sup>6</sup>).
+          1. Return _t_ + 𝔽(_offsetMs_).
         </emu-alg>
+        <p>If political rules for the local time _t_ are not available within the implementation, the result is _t_ because DefaultTimeZone returns *"UTC"* and GetNamedTimeZoneOffsetNanoseconds returns 0.</p>
+        <emu-note>
+          <p>It is recommended that implementations use the time zone information of the IANA Time Zone Database <a href="https://www.iana.org/time-zones/">https://www.iana.org/time-zones/</a>.</p>
+        </emu-note>
         <emu-note>
           <p>Two different input time values <emu-eqn>_t_<sub>UTC</sub></emu-eqn> are converted to the same local time <emu-eqn>t<sub>local</sub></emu-eqn> at a negative time zone transition when there are repeated times (e.g. the daylight saving time ends or the time zone adjustment is decreased.).</p>
           <p><emu-eqn>LocalTime(UTC(_t_<sub>local</sub>))</emu-eqn> is not necessarily always equal to <emu-eqn>_t_<sub>local</sub></emu-eqn>. Correspondingly, <emu-eqn>UTC(LocalTime(_t_<sub>UTC</sub>))</emu-eqn> is not necessarily always equal to <emu-eqn>_t_<sub>UTC</sub></emu-eqn>.</p>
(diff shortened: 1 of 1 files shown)


ptomato:
@tc39/ecma262-editors This is the PR that we discussed in the editor call yesterday, arising from https://github.com/tc39/proposal-temporal/pull/2171

I put all the new AOs together into the Date section, let me know if they need to go anywhere else. The number of new AOs is not too bad but if needed I can cut it down a bit more:
- RoundTowardsZero could instead be a mathematical function like [floor](https://tc39.es/ecma262/#eqn-floor)
- We could remove the "these are the steps if your implementation only has UTC" parts of the implementation-defined AOs (maybe their utility is questionable?) and that would allow removing GetEpochFromISOParts as well.

spec.html
```diff
@@ -32347,7 +32497,7 @@ <h1>Date Time String Format</h1>
                 `Z`
               </td>
               <td>
-                is the UTC offset representation specified as *"Z"* (for UTC with no offset) or an offset of either *"+"* or *"-"* followed by a time expression `HH:mm` (indicating local time ahead of or behind UTC, respectively)
+                is the UTC offset representation specified as *"Z"* (for UTC with no offset) or an offset of either *"+"* or *"-"* followed by a time expression `HH:mm` (indicating local time ahead of or behind UTC, respectively, which is a subset of the <emu-xref href="#sec-time-zone-offset-strings">time zone offset string format</emu-xref>.)
```
gibson042, line 32500: What is the benefit of this addition?

spec.html
```diff
@@ -32412,6 +32562,101 @@ <h1>Expanded Years</h1>
           </emu-note>
         </emu-clause>
       </emu-clause>
+
+      <emu-clause id="sec-time-zone-offset-strings">
+        <h1>Time Zone Offset String Format</h1>
+
+        <p>
+          ECMAScript defines a string interchange format for the UTC offset of a time zone.
```
gibson042, line 32570:
```suggestion
          ECMAScript defines a string interchange format for UTC offsets, derived from ISO 8601.
```

spec.html
```diff
@@ -32412,6 +32562,101 @@ <h1>Expanded Years</h1>
...
+          The format is described by the following grammar:
+        </p>
+
+        <emu-grammar type="definition">
+          ASCIISign ::: one of
```
gibson042, line 32575: For readability, ECMAScript grammars are generally presented outside-in (i.e., from the outermost nonterminal, which here would seem to be |UTCOffsetString|).

spec.html
```diff
@@ -32412,6 +32562,101 @@ <h1>Expanded Years</h1>
...
+            DecimalDigit DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit?
+
+          Fraction :::
+            DecimalSeparator FractionalPart
+
+          UTCOffsetOptionalComponents[Extended] :::
+            TimeSeparator[?Extended] MinuteSecond
+            TimeSeparator[?Extended] MinuteSecond TimeSeparator[?Extended] MinuteSecond Fraction?
+
+          UTCOffsetString :::
```
gibson042, line 32615: Suggested renames:
* |_UTCOffsetString_| to |_UTCOffset_|
* |_UTCOffsetOptionalComponents_| to something like |_HourSubcomponents_|
* |_Fraction_| to |_TemporalDecimalFraction_| (and subsume |_FractionalPart_|)
* |_DecimalSeparator_| to |_TemporalDecimalSeparator_|
* |_Sign_| to |_TemporalSign_|

gibson042 (approved):
I have some comments around naming and incorrect references to Unix (cf. https://github.com/tc39/proposal-temporal/issues/2175 ), but the bulk of this looks good.

spec.html
```diff
@@ -32086,27 +32086,127 @@ <h1>Week Day</h1>
...
+            _second_: an integer between 0 and 59 inclusive,
+            _millisecond_: an integer between 0 and 999 inclusive,
+            _microsecond_: an integer between 0 and 999 inclusive,
+            _nanosecond_: an integer between 0 and 999 inclusive,
+          ): a BigInt
         </h1>
         <dl class="header">
           <dt>description</dt>
-          <dd>Its return value represents the local time zone adjustment, or offset, in milliseconds. The local political rules for standard time and daylight saving time in effect at _t_ should be used to determine the result in the way specified in this section.</dd>
+          <dd>The returned value represents a number of nanoseconds since the Unix epoch in UTC that corresponds to the given ISO 8601 calendar date and wall-clock time in UTC.</dd>
```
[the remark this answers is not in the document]
ptomato: I'd hesitate to have a cross-reference to that whole section, because it might imply that the value here is a time value (in 𝔽). I propose making "epoch" into a `<dfn>` in that section, and having it automatically link here.

spec.html
```diff
@@ -32086,27 +32086,127 @@ <h1>Week Day</h1>
...
-          <p>2:30 AM on 12 March 2017 in America/New_York does not exist, but it must be interpreted as 2:30 AM UTC-05 (equivalent to 3:30 AM UTC-04). LocalTZA(TimeClip(MakeDate(MakeDay(2017, 2, 12), MakeTime(2, 30, 0, 0))), false) is <emu-eqn>-5 &times; msPerHour</emu-eqn>.</p>
-          <p>Local time zone offset values may be positive <i>or</i> negative.</p>
+          <p>1:30 AM on 5 November 2017 in America/New_York is repeated twice (fall backward), so GetIANATimeZoneEpochValue(*"America/New_York"*, 2017, 11, 5, 1, 30, 0, 0, 0, 0) would return a List of length 2.</p>
+          <p>2:30 AM on 12 March 2017 in America/New_York does not exist, so GetIANATimeZoneEpochValue(*"America/New_York"*, 2017, 3, 12, 2, 30, 0, 0, 0, 0) would return an empty List.</p>
+        </emu-note>
+      </emu-clause>
+
+      <emu-clause id="sec-getianatimezoneoffsetnanoseconds" type="implementation-defined abstract operation">
+        <h1>
+          GetIANATimeZoneOffsetNanoseconds (
```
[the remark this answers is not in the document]
ptomato: How about "Named" time zone, at least for the time being, since UTC offset time zones are handled in a non-implementation-defined way?

spec.html
```diff
@@ -32086,27 +32086,127 @@ <h1>Week Day</h1>
...
+          1. Let _time_ be MakeTime(𝔽(_hour_), 𝔽(_minute_), 𝔽(_second_), 𝔽(_millisecond_)).
+          1. Let _ms_ be MakeDate(_date_, _time_).
+          1. Assert: _ms_ is finite.
+          1. Return ℤ(ℝ(_ms_) &times; 10<sup>6</sup> + _microsecond_ &times; 10<sup>3</sup> + _nanosecond_).
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-getianatimezoneepochvalue" type="implementation-defined abstract operation">
+        <h1>
+          GetIANATimeZoneEpochValue (
```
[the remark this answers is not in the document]
ptomato: I've gone with GetNamedTimeZoneEpochNanoseconds (see my other comment below)

spec.html
```diff
@@ -32086,27 +32086,127 @@ <h1>Week Day</h1>
...
+          <p>2:30 AM on 12 March 2017 in America/New_York does not exist, so GetIANATimeZoneEpochValue(*"America/New_York"*, 2017, 3, 12, 2, 30, 0, 0, 0, 0) would return an empty List.</p>
+        </emu-note>
+      </emu-clause>
+
+      <emu-clause id="sec-getianatimezoneoffsetnanoseconds" type="implementation-defined abstract operation">
+        <h1>
+          GetIANATimeZoneOffsetNanoseconds (
+            _epochNanoseconds_: a BigInt,
+            _timeZoneIdentifier_: a String,
+          ): an integer
```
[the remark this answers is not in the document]
ptomato: This is a UTC offset, and the others are nanoseconds since the epoch. There's no reason in this PR specifically for nanoseconds since the epoch to be expressed as a bigint, but preserving that distinction is in the interest of aligning with Temporal, since exact time is stored as a BigInt in Temporal.Instant and Temporal.ZonedDateTime.

spec.html
ptomato: Since we have this ad-hoc format here as well as the newly added time zone offset string format below, I thought it'd be good to explain what the relationship was between them.

spec.html
ptomato: I've gone with this for now, but are you sure about having the "Temporal" name in there? It might be confusing since Temporal isn't part of the specification yet.

spec.html
gibson042: Hmm, I see. Perhaps the relationship can be expressed more strongly.
```suggestion
                is the UTC offset representation specified as *"Z"* (for UTC with no offset), or as either *"+"* or *"-"* followed by a time expression `HH:mm` (a subset of the <emu-xref href="#sec-time-zone-offset-strings">time zone offset string format</emu-xref> for indicating local time ahead of or behind UTC, respectively)
```

spec.html
gibson042: Ahead of the Temporal proposal landing, "temporal" still makes sense as a generic adjective indicating relationship to time—necessary here because date and time representations can use `,` as a decimal separator and `–` as a negative sign, neither of which are true of non-temporal (i.e., non-time-related) numeric strings in ECMAScript.

spec.html
```diff
@@ -32086,27 +32086,128 @@ <h1>Week Day</h1>
...
       </emu-clause>
 
-      <emu-clause id="sec-local-time-zone-adjustment" type="implementation-defined abstract operation">
+      <emu-clause id="sec-roundtowardszero" type="abstract operation">
         <h1>
-          LocalTZA (
-            _t_: a Number,
-            _isUTC_: a Boolean,
-          ): an integral Number
+          RoundTowardsZero (
```
gibson042, line 32091: It _is_ strange to see a general mathematical RoundTowardsZero operation defined under [Date Objects](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-date-objects); I think it would make more sense in [Mathematical Operations](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-mathematical-operations) (but I don't have an opinion on keeping it an abstract operation vs. a mathematical function such as "truncate(_x_) removes the fractional part of _x_ by rounding towards zero, producing -floor(-_x_) if _x_ &lt; 0 and otherwise producing floor(_x_)").

spec.html
```diff
@@ -32347,7 +32498,7 @@ <h1>Date Time String Format</h1>
                 `Z`
               </td>
               <td>
-                is the UTC offset representation specified as *"Z"* (for UTC with no offset) or an offset of either *"+"* or *"-"* followed by a time expression `HH:mm` (indicating local time ahead of or behind UTC, respectively)
+                is the UTC offset representation specified as *"Z"* (for UTC with no offset) or as either *"+"* or *"-"* followed by a time expression `HH:mm` (indicating local time ahead of or behind UTC, respectively, which is a subset of the <emu-xref href="#sec-time-zone-offset-strings">time zone offset string format</emu-xref> for indicating local time ahead of or behind UTC, respectively.)
```
gibson042, line 32501:
```suggestion
                is the UTC offset representation specified as *"Z"* (for UTC with no offset) or as either *"+"* or *"-"* followed by a time expression `HH:mm` (a subset of the <emu-xref href="#sec-time-zone-offset-strings">time zone offset string format</emu-xref> for indicating local time ahead of or behind UTC, respectively)
```

spec.html
```diff
@@ -32086,27 +32086,128 @@ <h1>Week Day</h1>
...
+      <emu-clause id="sec-getnamedtimezoneoffsetnanoseconds" oldids="sec-local-time-zone-adjustment" type="implementation-defined abstract operation">
+        <h1>
+          GetNamedTimeZoneOffsetNanoseconds (
+            _epochNanoseconds_: a BigInt,
+            _timeZoneIdentifier_: a String,
+          ): an integer
+        </h1>
+        <dl class="header">
+        </dl>
+        <p>The returned integer represents the offset from UTC of the IANA time zone identified by _timeZoneIdentifier_, at the instant corresponding with _epochNanoseconds_ relative to the epoch, both in nanoseconds.</p>
```
gibson042, line 32180:
```suggestion
        <p>The returned integer represents the offset from UTC of the time zone identified by _timeZoneIdentifier_, at the instant corresponding with _epochNanoseconds_ relative to the epoch, both in nanoseconds.</p>
```

spec.html
```diff
@@ -32086,27 +32086,128 @@ <h1>Week Day</h1>
...
+          <p>1:30 AM on 5 November 2017 in America/New_York is repeated twice, so GetNamedTimeZoneEpochNanoseconds(*"America/New_York"*, 2017, 11, 5, 1, 30, 0, 0, 0, 0) would return a List of length 2 in which the first element represents 05:30 UTC (corresponding with 01:30 US Eastern Daylight Time at UTC offset -04:00) and the second element represents 06:30 UTC (corresponding with 01:30 US Eastern Standard Time at UTC offset -05:00).</p>
+          <p>2:30 AM on 12 March 2017 in America/New_York does not exist, so GetNamedTimeZoneEpochNanoseconds(*"America/New_York"*, 2017, 3, 12, 2, 30, 0, 0, 0, 0) would return an empty List.</p>
+        </emu-note>
+      </emu-clause>
+
+      <emu-clause id="sec-getnamedtimezoneoffsetnanoseconds" oldids="sec-local-time-zone-adjustment" type="implementation-defined abstract operation">
+        <h1>
+          GetNamedTimeZoneOffsetNanoseconds (
+            _epochNanoseconds_: a BigInt,
+            _timeZoneIdentifier_: a String,
```
gibson042, line 32175: Would it make more sense to swap the argument order, aligning with the preceding operation?
```suggestion
            _timeZoneIdentifier_: a String,
            _epochNanoseconds_: a BigInt,
```

gibson042 (approved):
I just see a few more grammar points and stray references to IANA, and a possible improvement in operation naming.

spec.html
```diff
@@ -32086,27 +32086,128 @@ <h1>Week Day</h1>
...
+          <dd>The returned integer is the integral part of _x_, with any fractional part removed.</dd>
+        </dl>
+        <emu-alg>
+          1. Return the mathematical value that is the same sign as _x_ and whose magnitude is floor(abs(_x_)).
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-getepochnanosecondsfromisoparts" type="abstract operation">
+        <h1>
+          GetEpochNanosecondsFromISOParts (
```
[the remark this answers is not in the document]
ptomato: Sure, fair enough.

spec.html
ptomato: If we are moving it to Mathematical Operations, then making it a mathematical function seems better to me. (There are no other abstract operations in that section.)

gibson042:
This reached consensus. The promised followup details regarding DST transitions:

The requirements are described in GetNamedTimeZoneEpochNanoseconds (which returns a List of nanoseconds since epoch for a specific time zone name and local time) similar to the current text in LocalTZA: https://arai-a.github.io/ecma262-compare/?pr=2781#sec-getnamedtimezoneepochnanoseconds (emphasis mine)

> When the input represents a local time occurring more than once because of a negative time zone transition (e.g. when daylight saving time ends or the time zone offset is decreased due to a time zone rule change), the returned List will have more than one element and will be **sorted by ascending numerical value**. When the input represents a local time skipped because of a positive time zone transition (e.g. when daylight saving time begins or the time zone offset is increased due to a time zone rule change), the returned List will be empty. Otherwise, the returned List will have one element.

And specified more precisely in UTC (which interprets its input as a number representing local time and returns a single milliseconds since epoch): https://arai-a.github.io/ecma262-compare/?pr=2781#sec-utc-t (emphasis mine)

> If possibleInstants is not empty, then… Let disambiguatedInstant be possibleInstants[0]. [_i.e., the chronologically first matching local time_]
> Else [_a local time skipped at a positive time zone transition_], … Let possibleInstants be GetNamedTimeZoneEpochNanoseconds [_corresponding with_] tBefore is the largest integral Number < t for which possibleInstants is not empty (i.e., tBefore represents the last local time before the transition) … Let disambiguatedInstant be the last element of possibleInstants
> …
> t is interpreted using the time zone offset before the transition

ptomato:
I'm not sure there's really an effective test262 test possible for this; if I were to write one, it'd depend on implementation-defined date and time zone format strings. What else would be needed for merging this?

michaelficarra:
@ptomato This does not require test262 tests to be merged. Waiting on 262 editor review.

spec.html
```diff
@@ -31992,7 +31993,7 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
 
       <emu-clause id="sec-time-values-and-time-range">
         <h1>Time Values and Time Range</h1>
-        <p>Time measurement in ECMAScript is analogous to time measurement in POSIX, in particular sharing definition in terms of the proleptic Gregorian calendar, an epoch of midnight at the beginning of 1 January 1970 UTC, and an accounting of every day as comprising exactly 86,400 seconds (each of which is 1000 milliseconds long).</p>
+        <p>Time measurement in ECMAScript is analogous to time measurement in POSIX, in particular sharing definition in terms of the proleptic Gregorian calendar, an <dfn>epoch</dfn> of midnight at the beginning of 1 January 1970 UTC, and an accounting of every day as comprising exactly 86,400 seconds (each of which is 1000 milliseconds long).</p>
```
michaelficarra, line 31996:
```suggestion
        <p>Time measurement in ECMAScript is analogous to time measurement in POSIX, in particular sharing definition in terms of the proleptic Gregorian calendar, an <dfn id="epoch">epoch</dfn> of midnight at the beginning of 1 January 1970 UTC, and an accounting of every day as comprising exactly 86,400 seconds (each of which is 1000 milliseconds long).</p>
```

spec.html
```diff
@@ -32086,27 +32087,112 @@ <h1>Week Day</h1>
...
+        <dl class="header">
+        </dl>
+        <p>The returned integer represents the offset from UTC of the named time zone identified by _timeZoneIdentifier_, at the instant corresponding with _epochNanoseconds_ relative to the epoch, both in nanoseconds.</p>
+        <p>The minimum implementation of GetNamedTimeZoneOffsetNanoseconds for ECMAScript implementations that do not include local political rules for any time zones performs the following steps when called:</p>
+        <emu-alg>
+          1. Assert: _timeZoneIdentifier_ is *"UTC"*.
+          1. Return 0.
+        </emu-alg>
+        <emu-note>
+          <p>Time zone offset values may be positive <i>or</i> negative.</p>
```
michaelficarra, line 32172:
```suggestion
          <p>Time zone offset values may be positive <em>or</em> negative.</p>
```

or just

```suggestion
          <p>Time zone offset values may be positive or negative.</p>
```

since I don't think the emphasis really helps

spec.html
```diff
@@ -32086,27 +32087,112 @@ <h1>Week Day</h1>
...
+        <p>The minimum implementation of DefaultTimeZone for ECMAScript implementations that do not include local political rules for any time zones performs the following steps when called:</p>
+
+        <emu-alg>
+          1. Return *"UTC"*.
+        </emu-alg>
+
+        <emu-note>
+          <p>
+            To ensure the level of functionality that implementations commonly provide in the methods of the Date object, it is recommended that DefaultTimeZone return an IANA time zone name corresponding to the host environment's time zone setting, if such a thing exists.
+            GetNamedTimeZoneEpochNanoseconds and GetNamedTimeZoneOffsetNanoseconds must reflect the local political rules for standard time and daylight saving time in that time zone.
```
michaelficarra, line 32193:
```suggestion
            GetNamedTimeZoneEpochNanoseconds and GetNamedTimeZoneOffsetNanoseconds must reflect the local political rules for standard time and daylight saving time in that time zone if such rules exist.
```

or

```suggestion
            GetNamedTimeZoneEpochNanoseconds and GetNamedTimeZoneOffsetNanoseconds must reflect the local political rules for standard time and daylight saving time in time zones where daylight saving time is implemented.
```

spec.html
```diff
@@ -32118,11 +32204,25 @@ <h1>
...
+          1. Let _localTimeZone_ be DefaultTimeZone().
+          1. Let _offsetParseResult_ be Completion(ParseTimeZoneOffsetString(_localTimeZone_)).
+          1. If _offsetParseResult_ is a normal completion, then
+            1. Let _offsetNs_ be _offsetParseResult_.[[Value]].
+          1. Else,
+            1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_localTimeZone_, ℤ(ℝ(_t_) &times; 10<sup>6</sup>)).
+          1. Let _offsetMs_ be truncate(_offsetNs_ / 10<sup>6</sup>).
+          1. Return _t_ + 𝔽(_offsetMs_).
         </emu-alg>
+        <p>If political rules for the local time _t_ are not available within the implementation, the result is equal to _t_ because DefaultTimeZone returns *"UTC"* and GetNamedTimeZoneOffsetNanoseconds returns 0.</p>
```
michaelficarra, line 32222:
```suggestion
        <p>If political rules for the local time _t_ are not available within the implementation, the result is _t_ because DefaultTimeZone returns *"UTC"* and GetNamedTimeZoneOffsetNanoseconds returns 0.</p>
```

spec.html
```diff
@@ -32118,11 +32204,25 @@ <h1>
...
+          </dd>
         </dl>
         <emu-alg>
-          1. Return _t_ + LocalTZA(_t_, *true*).
+          1. Let _localTimeZone_ be DefaultTimeZone().
+          1. Let _offsetParseResult_ be Completion(ParseTimeZoneOffsetString(_localTimeZone_)).
+          1. If _offsetParseResult_ is a normal completion, then
+            1. Let _offsetNs_ be _offsetParseResult_.[[Value]].
+          1. Else,
+            1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_localTimeZone_, ℤ(ℝ(_t_) &times; 10<sup>6</sup>)).
```
michaelficarra, line 32218: I think we need an assertion that `_t_` is not `NaN` so that we can safely take `ℝ(_t_)`.

michaelficarra, line 32218: I don't understand why we use this fallback pattern instead of just

```suggestion
          1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_localTimeZone_, ℤ(ℝ(_t_) &times; 10<sup>6</sup>)).
```

Also, if this pattern is necessary, could we encapsulate it into its own AO? I think the swallowing of a completion record here is a bit strange and I'd rather limit it to one occurrence.

spec.html
```diff
@@ -32137,11 +32237,47 @@ <h1>
...
+            1. Let _offsetNs_ be _offsetParseResult_.[[Value]].
+          1. Else,
+            1. Let _possibleInstants_ be GetNamedTimeZoneEpochNanoseconds(_localTimeZone_, ℝ(YearFromTime(_t_)), ℝ(MonthFromTime(_t_)) + 1, ℝ(DateFromTime(_t_)), ℝ(HourFromTime(_t_)), ℝ(MinFromTime(_t_)), ℝ(SecFromTime(_t_)), ℝ(msFromTime(_t_)), 0, 0).
+            1. If _possibleInstants_ is not empty, then
+              1. Let _disambiguatedInstant_ be _possibleInstants_[0].
+            1. Else,
+              1. NOTE: _t_ represents a local time skipped at a positive time zone transition (e.g. due to daylight saving time starting or a time zone rule change increasing the UTC offset).
+              1. Let _possibleInstants_ be GetNamedTimeZoneEpochNanoseconds(_localTimeZone_, ℝ(YearFromTime(_tBefore_)), ℝ(MonthFromTime(_tBefore_)) + 1, ℝ(DateFromTime(_tBefore_)), ℝ(HourFromTime(_tBefore_)), ℝ(MinFromTime(_tBefore_)), ℝ(SecFromTime(_tBefore_)), ℝ(msFromTime(_tBefore_)), 0, 0), where _tBefore_ is the largest integral Number &lt; _t_ for which _possibleInstants_ is not empty (i.e., _tBefore_ represents the last local time before the transition).
```
michaelficarra, line 32256: I don't like how `_possibleInstants_` is shadowed here. Use a new alias.

spec.html
```diff
@@ -32137,11 +32237,47 @@ <h1>
...
+              1. Let _disambiguatedInstant_ be the last element of _possibleInstants_.
```
michaelficarra, line 32257: What's the motivation for choosing the last element here but the first element earlier?

spec.html
```diff
@@ -32137,11 +32237,47 @@ <h1>
...
-          1. Return _t_ - LocalTZA(_t_, *false*).
-        </emu-alg>
+          1. Let _localTimeZone_ be DefaultTimeZone().
+          1. Let _offsetParseResult_ be Completion(ParseTimeZoneOffsetString(_localTimeZone_)).
+          1. If _offsetParseResult_ is a normal completion, then
+            1. Let _offsetNs_ be _offsetParseResult_.[[Value]].
+          1. Else,
+            1. Let _possibleInstants_ be GetNamedTimeZoneEpochNanoseconds(_localTimeZone_, ℝ(YearFromTime(_t_)), ℝ(MonthFromTime(_t_)) + 1, ℝ(DateFromTime(_t_)), ℝ(HourFromTime(_t_)), ℝ(MinFromTime(_t_)), ℝ(SecFromTime(_t_)), ℝ(msFromTime(_t_)), 0, 0).
+            1. If _possibleInstants_ is not empty, then
+              1. Let _disambiguatedInstant_ be _possibleInstants_[0].
```
michaelficarra, line 32253: May want to move or copy the note about multiple times from below to here.

spec.html
ptomato: That should be fine, I just checked that all callers of LocalTime don't call it with NaN.

ptomato: It's because unlike the offset nanoseconds of a named time zone like `America/Sao_Paulo`, it's fully specifiable how to get the offset nanoseconds of a time zone like `-07:00` or `+05:30` or `+00:19:32.13`.

The weirdness with the completion record is because we don't have any other mechanism than ParseText for determining whether a string matches a grammar. (Previously, in Temporal, I had language like "`if _string_ satisfies the syntax of |GrammarSymbol|`" but we considered that not rigorous enough.)

Here's a proposal that I think might address your concerns:

1. If IsTimeZoneOffsetString(_localTimeZone_) is **true**, then
  - a. Let _offsetNs_ be ! ParseTimeZoneOffsetString(_localTimeZone_).
2. Else,
  - a. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_localTimeZone_, ...).

IsTimeZoneOffsetString would be a new AO that encapsulates using ParseText to determine whether the string is an offset string. That way, we don't have to use completion records at all.

ptomato: I'd also consider adding "If _t_ is **NaN**, return **NaN**" to UTC(), but that has pre-existing normative implications, so maybe should be out of scope for this PR.

(Specifically, `new Date(NaN, NaN)` ends up calling UTC(NaN) which computes NaN - LocalTZA(NaN, false), and LocalTZA did not specify what to do with NaN. Web reality seems to be that UTC(NaN) should return NaN.)

ptomato:
@michaelficarra Thanks for the review, I have kept the changes addressing your comments in a separate commit for ease of review, but I can squash them before merging if you prefer.

spec.html
michaelficarra: @ptomato I don't think you made this change.

spec.html
michaelficarra: That WFM. I would've probably gone a step further and encapsulated the rest of that conditional, but either way is fine.

spec.html
```diff
@@ -32395,6 +32529,115 @@ <h1>Expanded Years</h1>
...
+            ): an integer
+          </h1>
+          <dl class="header">
+            <dt>description</dt>
+            <dd>The return value is the UTC offset, as a number of nanoseconds, that corresponds to the String _offsetString_.</dd>
+          </dl>
+          <emu-alg>
+            1. Let _parseResult_ be ParseText(StringToCodePoints(_offsetString_), |UTCOffset|).
+            1. Assert: _parseResult_ is not a List of errors.
+            1. Let each of _parsedSign_, _parsedHours_, _parsedMinutes_, _parsedSeconds_, and _parsedFraction_ be the source text matched by the |TemporalSign|, |Hour|, first |MinuteSecond|, second |MinuteSecond|, and |TemporalDecimalFraction| Parse Node contained within _parseResult_, or an empty sequence of code points if not present.
```
michaelficarra, line 32615: While the attempt at concision here is commendable, I think this would be better as just a sequence of very similar steps introducing these aliases.

spec.html
```diff
@@ -32395,6 +32529,115 @@ <h1>Expanded Years</h1>
...
+            1. Assert: _parsedSign_ is not empty.
+            1. If _parsedSign_ is the single code point 0x002D (HYPHEN-MINUS) or 0x2212 (MINUS SIGN), then
```
michaelficarra, line 32617: Code points use the U+XXXX notation and code units use the 0xXXXX notation. You probably wanted 

```suggestion
            1. If _parsedSign_ is U+002D (HYPHEN-MINUS) or U+2212 (MINUS SIGN), then
```

spec.html
```diff
@@ -32069,43 +32070,141 @@ <h1>Week Day</h1>
...
-      <emu-clause id="sec-local-time-zone-adjustment" type="implementation-defined abstract operation">
+      <emu-clause id="sec-getutcepochnanoseconds" type="abstract operation">
         <h1>
-          LocalTZA (
-            _t_: a Number,
-            _isUTC_: a Boolean,
-          ): an integral Number
+          GetUTCEpochNanoseconds (
+            _year_: an integer,
+            _month_: an integer between 1 and 12 inclusive,
```
ljharb, line 32077: Is “inclusive” defined somewhere? as written I’m not sure if it includes 1 or 12 or both (i weakly lean towards both)

spec.html
```diff
@@ -32069,43 +32070,141 @@ <h1>Week Day</h1>
...
+          </p>
+          <p>For example, if the host environment is a browser on a system where the user has chosen US Eastern Time as their time zone, DefaultTimeZone returns *"America/New_York"*.</p>
         </emu-note>
       </emu-clause>
 
       <emu-clause id="sec-localtime" type="abstract operation">
         <h1>
           LocalTime (
-            _t_: a time value,
+            _t_: a time value, but not *NaN*,
```
ljharb, line 32185:
```suggestion
            _t_: a finite time value,
```
?

spec.html
ptomato: I realize now there were two instances of it, I had changed one but not the other.

spec.html
ptomato: I don't think it's defined, but there are many existing uses of it in ECMA-262 where it's clear that it's inclusive on both ends, e.g. [ToInt8](https://tc39.es/ecma262/#sec-toint8).

ptomato:
@michaelficarra @ljharb Thanks for the quick response, new modifications are in another separate commit.

spec.html
ljharb: Sounds good - @michaelficarra, perhaps we should define it in a separate PR?

michaelficarra: I think it's pretty intuitive and sufficiently unambiguous to our readers. I wouldn't be *against* adding a clarification of the term, but it'd probably take some convincing.

ljharb: I find it ambiguous :-) however your suggested change would make it unambiguous to me that "inclusive" applies to both ends, so that sounds great.

michaelficarra: I opened #2848. If we follow the convention I used in that PR, the phrasing here would be "an integer within the inclusive range from 1 to 12". We can wait for that to be merged before updating here, though.

ptomato:
It seems like at least the phrasing "in the inclusive interval from 1 to 12" is stable in #2848, although details are still being debated. I've updated this PR to use that phrasing and rebased it. Would it be possible to merge this, since it seems like the open discussions in #2848 are not going to change anything here? There are several things I need to do in Temporal that are waiting on this.

michaelficarra (approved):

spec.html
```diff
@@ -32069,43 +32070,141 @@ <h1>Week Day</h1>
...
-        <p>If an implementation does not support a conversion described above or if political rules for time _t_ are not available within the implementation, the result must be *+0*<sub>𝔽</sub>.</p>
+        <p>
+          When the input represents a local time occurring more than once because of a negative time zone transition (e.g. when daylight saving time ends or the time zone offset is decreased due to a time zone rule change), the returned List will have more than one element and will be sorted by ascending numerical value.
+          When the input represents a local time skipped because of a positive time zone transition (e.g. when daylight saving time begins or the time zone offset is increased due to a time zone rule change), the returned List will be empty.
+          Otherwise, the returned List will have one element.
+        </p>
+        <p>The minimum implementation of GetNamedTimeZoneEpochNanoseconds for ECMAScript implementations that do not include local political rules for any time zones performs the following steps when called:</p>
```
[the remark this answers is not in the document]
syg: I forgot to submit this comment for like a month, and now it's outdated. What was this language replaced with @ptomato?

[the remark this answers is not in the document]
ptomato: It still said the same thing, there must have been some other change that made it get marked as outdated. I've changed it to "The default implementation of ___, to be used for ECMAScript implementations that do not include local political rules for any time zones, performs the following steps when called:" in all 3 places.

ptomato:
If there's anything that I should join the editor call to discuss, that would help move this forward, please let me know - I'm available during that time today.

ljharb:
@ptomato the editor call doesn't happen during TC39 weeks, so it'd be next week.

gibson042 (approved):

bakkot (commented):
Mostly LGTM, some minor nits we can resolve when merging.

spec.html
```diff
@@ -32457,6 +32591,120 @@ <h1>Expanded Years</h1>
...
+
+        <emu-grammar type="definition">
+          UTCOffset :::
+            TemporalSign Hour
+            TemporalSign Hour HourSubcomponents[+Extended]
+            TemporalSign Hour HourSubcomponents[~Extended]
+
+          TemporalSign :::
+            ASCIISign
+            U+2212
```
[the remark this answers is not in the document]
michaelficarra: I'd rather go the other way around and just use the code point instead of introducing these aliases. Let's add grammar notation for code points.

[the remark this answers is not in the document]
bakkot: We shouldn't block this PR on adding such notation to all the tooling. You're welcome to open an issue for it and then we can address it at some point.

spec.html
```diff
@@ -32457,6 +32591,120 @@ <h1>Expanded Years</h1>
...
+          MinuteSecond :::
+            `0` DecimalDigit
+            `1` DecimalDigit
+            `2` DecimalDigit
+            `3` DecimalDigit
+            `4` DecimalDigit
+            `5` DecimalDigit
+
+          TemporalDecimalFraction :::
+            TemporalDecimalSeparator DecimalDigit DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit? DecimalDigit?
```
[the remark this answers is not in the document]
gibson042: Couldn't that also be fixed by adopting the same conventions as https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-ecmascript-language-lexical-grammar , which seems appropriate for `:::` grammars anyway?
> The source text is scanned from left to right, repeatedly taking the longest possible sequence of code points as the next input element.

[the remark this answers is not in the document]
bakkot: Yes, or by adopting the first-wins rule from the [Annex B pattern grammar](https://tc39.es/ecma262/multipage/additional-ecmascript-features-for-web-browsers.html#sec-regular-expressions-patterns), but I would prefer to keep the grammar unambiguous where it's practical to do so.

spec.html
```diff
@@ -32457,6 +32591,120 @@ <h1>Expanded Years</h1>
...
+            1. Else,
+              1. Let _parsedSeconds_ be the source text matched by the second |MinuteSecond| Parse Node contained within _parseResult_.
+              1. Let _seconds_ be the MV of CodePointsToString(_parsedSeconds_).
+            1. If _parseResult_ does not contain a |TemporalDecimalFraction| Parse Node, then
+              1. Let _nanoseconds_ be 0.
+            1. Else,
+              1. Let _parsedFraction_ be the source text matched by the |TemporalDecimalFraction| Parse Node contained within _parseResult_.
+              1. Let _fraction_ be the string-concatenation of CodePointsToString(_parsedFraction_) and the string *"000000000"*.
+              1. Let _nanosecondsString_ be the String value equal to the substring of _fraction_ from 1 to 10.
+              1. Let _nanoseconds_ be the MV of _nanosecondsString_.
```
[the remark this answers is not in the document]
ptomato: I guess we could just use the parse node that we already get from the parse tree returned from ParseText(..., |UTCOffset|)?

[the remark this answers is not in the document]
ptomato: (Except in this nanoseconds case)

ptomato:
Thanks, I've rebased this and made the changes in a separate commit so you can easily see what changed, and then squash before merging.

spec.html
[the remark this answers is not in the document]
bakkot: > I guess we could just use the parse node that we already get from the parse tree returned from ParseText(..., |UTCOffset|)?

The types line up that way, but to make it work in full you'd also have to define MV over all of those productions - right now [its definition](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-static-semantics-mv) does not know how to handle an `|Hour|` or `|MinuteSecond|` parse node. And adding those would be kind of annoying.

Now that I'm looking at this again, I think the best thing would be to use [StringToNumber](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-stringtonumber) in all cases (including the nanoseconds), and then cast to a real. This only works because you know these are small integers (i.e. less than MAX_SAFE_INTEGER), but I think that fact is obvious enough in these contexts. (Sorry I didn't suggest this originally; I hadn't noticed the other uses of MV.)

It's possible the other editors will prefer the re-parsing approach, though. I'll ask tomorrow and get this resolved. And don't worry about fixing it up; once we decide what approach to use the editors can make the last tweaks and land it.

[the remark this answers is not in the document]
syg: I also prefer using StringToNumber here with a note that it's not lossy in this case. DateTime stuff tend to get implemented line-for-line from the spec text, so using StringToNumber here would aid implementation clarity.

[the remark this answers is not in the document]
bakkot: Pushed up 1688afc59b5aa5c6757e96fc13f5f4d16ad6a3d3 switching to StringToNumber.

bakkot:
I think this is good to go now but will wait to merge until other editors confirm.

spec.html
jmdyck: You said "That should be fine", but you didn't actually add the assertion. In addition to excluding NaN, you presumably need to exclude infinities, so I think you should change the type of `_t_` from `a Number` to `a finite Number`.

spec.html
```diff
@@ -32385,6 +32519,156 @@ <h1>Expanded Years</h1>
...
+            TemporalDecimalSeparator DecimalDigit DecimalDigit DecimalDigit DecimalDigit DecimalDigit DecimalDigit DecimalDigit DecimalDigit DecimalDigit
+
+          TemporalDecimalSeparator ::: one of
+            `.` `,`
+        </emu-grammar>
+
+        <emu-clause id="sec-istimezoneoffsetstring" type="abstract operation">
+          <h1>
+            IsTimeZoneOffsetString (
+              _string_: a String,
```
jmdyck, line 32614:
```suggestion
              _offsetString_: a String,
```
because that's how the algorithm refers to it.

spec.html
```diff
@@ -32385,6 +32519,156 @@ <h1>Expanded Years</h1>
...
+            1. If _parseResult_ does not contain two |MinuteSecond| Parse Nodes, then
+              1. Let _seconds_ be 0.
+            1. Else,
+              1. Let _parsedSeconds_ be the source text matched by the second |MinuteSecond| Parse Node contained within _parseResult_.
+              1. Let _seconds_ be ℝ(StringToNumber(CodePointsToString(_parsedSeconds_))).
+            1. If _parseResult_ does not contain a |TemporalDecimalFraction| Parse Node, then
+              1. Let _nanoseconds_ be 0.
+            1. Else,
+              1. Let _parsedFraction_ be the source text matched by the |TemporalDecimalFraction| Parse Node contained within _parseResult_.
+              1. Let _fraction_ be the string-concatenation of CodePointsToString(_parsedFraction_) and the string *"000000000"*.
```
jmdyck, line 32665:
```suggestion
              1. Let _fraction_ be the string-concatenation of CodePointsToString(_parsedFraction_) and the String *"000000000"*.
```

spec.html
```diff
@@ -32385,6 +32519,156 @@ <h1>Expanded Years</h1>
...
+              1. Let _nanosecondsString_ be the String value equal to the substring of _fraction_ from 1 to 10.
```
jmdyck, line 32666:
```suggestion
              1. Let _nanosecondsString_ be the substring of _fraction_ from 1 to 10.
```

spec.html
```diff
@@ -33282,7 +33566,12 @@ <h1>
           <dl class="header">
           </dl>
           <emu-alg>
-            1. Let _offset_ be LocalTZA(_tv_, *true*).
+            1. Let _localTimeZone_ be DefaultTimeZone().
+            1. If IsTimeZoneOffsetString(_localTimeZone_) is *true*, then
+              1. Let _offsetNs_ be ParseTimeZoneOffsetString(_localTimeZone_).
+            1. Else,
+              1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_localTimeZone_, ℤ(ℝ(_t_) &times; 10<sup>6</sup>)).
```
jmdyck, line 33573:
```suggestion
              1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_localTimeZone_, ℤ(ℝ(_tv_) &times; 10<sup>6</sup>)).
```
Moreover, this formula won't work for a lot of non-NaN Numbers, so you should probably change the type of `_tv_` to `an integral Number`.

And since ToDateString calls this op, you should probably change the type of ToDateString's `_tv_` to `an integral Number or *NaN*`.

spec.html
gibson042:
```suggestion
              1. Let _fraction_ be the string-concatenation of CodePointsToString(_parsedFraction_) and *"000000000"*.
```

spec.html
ptomato: For LocalTime(), I changed the type in the structured header to "a finite time value", which is implicitly an assertion.

For UTC(), see my comment after the "that should be fine" one — I don't think it's possible to add an assertion to UTC() here without contradicting web reality, and I would like to place resolving that pre-existing discrepancy out of scope of this PR.

ptomato:
As usual, new changes are in a fixup commit.

spec.html
jmdyck: I understand your desire to leave something to another PR, but do you think that the proposed definition of UTC() *doesn't* contradict web reality?

jmdyck: I think there's an error in the spec either way: with the status quo and this PR, the error is trying to do arithmetic wiht a NaN; with a tighter type for `_t_`, the error is trying to pass a NaN to an AO that doesn't accept it.

I'm fine with whatever the editors decide.

syg (approved):

spec.html
```diff
@@ -32385,6 +32519,156 @@ <h1>Expanded Years</h1>
...
+            1. Assert: _parseResult_ contains a |TemporalSign| Parse Node.
+            1. Assert: _parseResult_ contains an |Hour| Parse Node.
+            1. Let _parsedSign_ be the source text matched by the |TemporalSign| Parse Node contained within _parseResult_.
+            1. If _parsedSign_ is the single code point U+002D (HYPHEN-MINUS) or U+2212 (MINUS SIGN), then
+              1. Let _sign_ be -1.
+            1. Else,
+              1. Let _sign_ be 1.
+            1. NOTE: Applications of StringToNumber below do not lose precision, since each of the parsed values is guaranteed to be a sufficiently short string of decimal digits.
+            1. Let _parsedHours_ be the source text matched by the |Hour| Parse Node contained within _parseResult_.
+            1. Let _hours_ be ℝ(StringToNumber(CodePointsToString(_parsedHours_))).
```
gibson042, line 32650: Nit: I think assertions make the most sense immediately before steps that assume them.
```suggestion
            1. Assert: _parseResult_ contains a |TemporalSign| Parse Node.
            1. Let _parsedSign_ be the source text matched by the |TemporalSign| Parse Node contained within _parseResult_.
            1. If _parsedSign_ is the single code point U+002D (HYPHEN-MINUS) or U+2212 (MINUS SIGN), then
              1. Let _sign_ be -1.
            1. Else,
              1. Let _sign_ be 1.
            1. NOTE: Applications of StringToNumber below do not lose precision, since each of the parsed values is guaranteed to be a sufficiently short string of decimal digits.
            1. Assert: _parseResult_ contains an |Hour| Parse Node.
            1. Let _parsedHours_ be the source text matched by the |Hour| Parse Node contained within _parseResult_.
            1. Let _hours_ be ℝ(StringToNumber(CodePointsToString(_parsedHours_))).
```

spec.html
```diff
@@ -32059,43 +32060,141 @@ <h1>Week Day</h1>
...
+      </emu-clause>
+
+      <emu-clause id="sec-defaulttimezone" type="implementation-defined abstract operation">
+        <h1>DefaultTimeZone ( ): a String</h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It returns a String value representing the host environment's current time zone, which is either a String representing a UTC offset for which IsTimeZoneOffsetString returns *true*, or a String identifier accepted by GetNamedTimeZoneEpochNanoseconds and GetNamedTimeZoneOffsetNanoseconds.</dd>
+        </dl>
+
+        <p>An ECMAScript implementation that includes the ECMA-402 Internationalization API must implement the DefaultTimeZone abstract operation as specified in the ECMA-402 specification.</p>
```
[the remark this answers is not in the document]
ptomato: Concretely, should there be any change? Should we remove the paragraph starting "An ECMAScript implementation that includes the ECMA-402 ..." in anticipation of ECMA-402 no longer providing a definition for DefaultTimeZone?

(FWIW, I think ECMA-402's "A conforming implementation must recognize [...] Zone and Link names" makes it OK to also say that the return value of DefaultTimeZone must be one of them, but I see your point as well, so I don't see any reason to be strict about that.)

ptomato:
Rebased; added the changes that Richard requested in a fixup commit.

spec.html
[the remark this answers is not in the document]
gibson042: I don't think anything needs to change here, I'm just calling attention to a downstream consequence. We can remove that text when there is no longer an override, but I think it should remain until then.

PR #2790, opened 2022-06-02
Editorial: Consistify miscellaneous phrases

h2oche:
I observed some phrases are written in inconsistent form:

- To extract the element from the list of length 1, other phrases use `the sole element of`.
- To check if the list is empty, other phrases use 'is empty' or 'is not empty'
- To reference some well-known objects, other phrases use `active function object` and `the running execution context`, respectively.
- To append an element to the end of list, other phrases use `append _element_ to _list_.`
- Removed unnecessary prefix 'the String' for denoting a string literal.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6630,7 +6630,7 @@
           1. Let _indexName_ be ! ToString(𝔽(_index_)).
           1. Let _next_ be ? Get(_obj_, _indexName_).
           1. If Type(_next_) is not an element of _elementTypes_, throw a *TypeError* exception.
-          1. Append _next_ as the last element of _list_.
+          1. Append _next_ to _list_.
           1. Set _index_ to _index_ + 1.
         1. Return _list_.
       </emu-alg>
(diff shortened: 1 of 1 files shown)


michaelficarra:
The final commit here can be dropped in favour of https://github.com/tc39/ecma262/pull/2711

michaelficarra (approved):
LGTM otherwise. This is great!

h2oche:
@michaelficarra Oh, I see. I dropped it.

spec.html
```diff
@@ -46275,7 +46275,7 @@ <h1>
         1. Let _events_ be an empty Set.
         1. For each Agent Events Record _aer_ of _execution_.[[EventsRecords]], do
           1. For each event _E_ of _aer_.[[EventList]], do
-            1. Add _E_ to _events_.
+            1. Append _E_ to _events_.
```
bakkot, line 46278: I prefer `Add` rather than `Append` for adding an element to a Set. Ditto the two cases below.

bakkot:
Looks good to me except for the three uses of "append" for updating a Set.

michaelficarra:
@bakkot @h2oche I've gone ahead and just made the requested change. Should be ready to merge now.

PR #2418, opened 2021-05-28
[Normative] Add RegExp `v` flag with set notation and properties of strings

mathiasbynens:
Proposal repo: https://github.com/tc39/proposal-regexp-set-notation

**Note:** This is a placeholder PR that allows us to set up a preview link for the spec changes in the future. For now we’ll continue to iterate on the draft spec in [the Google doc](https://docs.google.com/document/d/1Tbv3hfX9CxQtzH9r-JdxJsQZhmmDsidRUKKxg345JV0/edit).

PR #2791, opened 2022-06-02
Normative: Always check regular expression flags by "flags"

gibson042:
Discussion at https://github.com/tc39/ecma262/pull/2418#discussion_r876052820 uncovered a strange inconsistency: unlike other methods on or related to regular expressions, [RegExp.prototype[@@match]](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp.prototype-@@match) and [RegExp.prototype[@@replace]](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp.prototype-@@replace) bypass the "flags" property, instead reading directly (and only) from "global" and "unicode"—and doing so conditionally, checking the latter if and only if the former is coerced to boolean **true**. This is awkward for the introduction of v-mode regular expressions, which need to check a third flag in at least some cases and therefore require taking a position on when to get the property (which is observable).

But these are the only two methods that have such a problem... [String.prototype.matchAll](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.matchall) and [String.prototype.replaceAll](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.replaceall) and [RegExp.prototype[@@matchAll]](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp-prototype-matchall) and [RegExp.prototype[@@split]](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp.prototype-@@split) get "flags", coerce to string, and then check the result for "u"/"g"/etc. (and the built-in [RegExp.prototype.flags](https://tc39.es/ecma262/multipage/text-processing.html#sec-get-regexp.prototype.flags) itself performs an unconditional Get of each specific flag property in a stable order that is compatible with the deviant operations [i.e., "global" before "unicode"]).

I would like to fix the inconsistency and render the negotiation of https://github.com/tc39/ecma262/pull/2418#discussion_r876052820 moot by updating `@@match` and `@@replace` to align with the other methods in using "flags". This is a normative change, but one that seems likely to be web-compatible because any well-behaved regular expression analog is already required to support "flags" for the other methods, and deviation would actually require special effort—an author would need to copy the built-in `@@match` and/or `@@replace` but specifically remove or override the built-in `flags` getter to be inconsistent with independent `global`/`unicode` access. And if we _don't_ fix this now, then the problem is only going to get worse as more flags are added.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -36555,12 +36555,11 @@ THH:mm:ss.sss
           1. Let _rx_ be the *this* value.
           1. If Type(_rx_) is not Object, throw a *TypeError* exception.
           1. Let _S_ be ? ToString(_string_).
-          1. Let _global_ be ToBoolean(? Get(_rx_, *"global"*)).
-          1. If _global_ is *false*, then
+          1. Let _flags_ be ? ToString(? Get(_rx_, *"flags"*)).
+          1. If _flags_ does not contain *"g"*, then
             1. Return ? RegExpExec(_rx_, _S_).
           1. Else,
-            1. Assert: _global_ is *true*.
-            1. Let _fullUnicode_ be ToBoolean(? Get(_rx_, *"unicode"*)).
+            1. If _flags_ contains *"u"*, let _fullUnicode_ be *true*. Otherwise, let _fullUnicode_ be *false*.
             1. Perform ? Set(_rx_, *"lastIndex"*, *+0*<sub>𝔽</sub>, *true*).
             1. Let _A_ be ! ArrayCreate(0).
             1. Let _n_ be 0.
@@ -36626,9 +36625,10 @@ THH:mm:ss.sss
           1. Let _functionalReplace_ be IsCallable(_replaceValue_).
           1. If _functionalReplace_ is *false*, then
             1. Set _replaceValue_ to ? ToString(_replaceValue_).
-          1. Let _global_ be ToBoolean(? Get(_rx_, *"global"*)).
+          1. Let _flags_ be ? ToString(? Get(_rx_, *"flags"*)).
+          1. If _flags_ contains *"g"*, let _global_ be *true*. Otherwise, let _global_ be *false*.
           1. If _global_ is *true*, then
-            1. Let _fullUnicode_ be ToBoolean(? Get(_rx_, *"unicode"*)).
+            1. If _flags_ contains *"u"*, let _fullUnicode_ be *true*. Otherwise, let _fullUnicode_ be *false*.
             1. Perform ? Set(_rx_, *"lastIndex"*, *+0*<sub>𝔽</sub>, *true*).
           1. Let _results_ be a new empty List.
           1. Let _done_ be *false*.


ljharb:
This imo goes in the wrong direction - it creates *more* observable behavior, via the regex subclassing system that none of us want but it's too late to remove, and it may have the unfortunate side effect of causing websites using polyfills to break (I absolutely have tests, at least, that would break with this change, that rely on `.flags` *not* being observably read in these methods).

bakkot:
@ljharb This seems to me to reduce the amount of observable behavior? Previously `@@replace` could do two lookups (of `.global` and `.unicode`); now it only does one.

mathiasbynens:
I’d be stoked if we can get away with this (which as @gibson042 points out, actually seems plausible).

ljharb:
@gibson042 The purpose of verifying it would be that it's in the spec, and thus that's the precise behavior engines are required to have. Since no engines I'm aware of have implemented this incorrectly, i don't think I have any code that would care - but an in-progress shim I'm working on for the regex symbol methods (which isn't public yet) would need to test for this PR's change, and then it would need to patch the relevant methods in every browser prior to it, even if otherwise no patch would be needed.

gibson042:
It's rare for tests to verify that things _don't_ happen, and the set of such (non-)events is unbounded (although in this particular case two get-unicode-error.js tests introduced by @jugglinmike in https://github.com/tc39/test262/pull/352 do in fact attempt to do so).

Regardless, except for hypothetical code that that copies the built-in `@@match` and/or `@@replace` but specifically removes or overrides the built-in `flags` getter (which I posit is negligible in the same sense as other code affected by [breaking changes](https://tc39.es/ecma262/multipage/additions-and-changes-that-introduce-incompatibilities-with-prior-editions.html#sec-additions-and-changes-that-introduce-incompatibilities-with-prior-editions), especially if the alternative is renegotiating every time a relevant new flag is introduced), the changes of this PR observable by ECMAScript code would be purely additive:

**Receiver that self-reports as non-global**
(_ToBoolean(? Get(_rx_, "global")) is **false**_)
```patch
+Get "flags"
+# Built-in "flags" getter:
 Get "global"
+Get "hasIndices"
+Get "global"
+Get "ignoreCase"
+Get "multiline"
+Get "dotAll"
+Get "unicode"
+Get "sticky"
```

**Receiver that self-reports as global**
(_ToBoolean(? Get(_rx_, "global")) is **true**_)
```patch
+Get "flags"
+# Built-in "flags" getter:
 Get "global"
+Get "hasIndices"
+Get "global"
+Get "ignoreCase"
+Get "multiline"
+Get "dotAll"
 Get "unicode"
+Get "sticky"
```

The responsible discrepancy seems to date back to [ES2015](https://262.ecma-international.org/6.0/), in which [RegExp.prototype [ @@split ]](https://262.ecma-international.org/6.0/#sec-regexp.prototype-@@split) reads "flags" but [@@match](https://262.ecma-international.org/6.0/#sec-regexp.prototype-@@match) and [@@replace](https://262.ecma-international.org/6.0/#sec-regexp.prototype-@@replace) have conditional single-flag reads. The previous edition ([ES5.1](https://262.ecma-international.org/5.1/)) didn't specify a `flags` property at all, and I suspect the difference was just an oversight—split is sensitive only to new "unicode" and "sticky" flags, while match and replace gained sensitivity to the new "unicode" flag in translations of preexisting sensitivity to the "global" flag like «Let <i>global</i> be the result of calling the [[Get]] internal method of <i>rx</i> with argument <code>"global"</code>» and «If <var>searchValue</var>.global is {<b>false</b>,<b>true</b>}, then...». All three exist in support of String method signatures dating back to [ES3](https://www.ecma-international.org/wp-content/uploads/ECMA-262_3rd_edition_december_1999.pdf), and it would surprise me if any code depends upon exactly two of them _not_ reading a property that was introduced later. I think we should at least try to find out.

ljharb:
I would expect that the very inconsistency you're trying to fix here is why this specific non-happening would be tested for, but I agree it's going to be rare in existing code.

ljharb:
On a re-review of this PR and the comment thread, I'm completely convinced that this is a good change that reduces observable lookups.

It will likely have two negative effects for my packages:
1. some of my tests will break, but these are easily fixed
1. a faithful polyfill for these two methods will need to apply the shim for this purpose in browsers that ordinarily will not need it. However, none of the existing published es-shims code checks for this, so it will be unlikely to have a major impact before browsers are able to update.

imo, neither of these negative effects should obstruct consensus on this PR.

gibson042:
test262 PR: https://github.com/tc39/test262/pull/3618

michaelficarra (approved):

bakkot (approved):

PR #2793, opened 2022-06-07
Meta: bump ecmarkup to 12.1.1

bakkot:
For https://github.com/tc39/ecmarkup/pull/455, so that e.g. [NewPromiseReactionJob](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-newpromisereactionjob) will (correctly) no longer be considered to invoke user code. This is important since [we're discussing this stuff](https://github.com/tc39/proposal-faster-promise-adoption).


syg (approved):
lgtm, thanks for the fix.

PR #2795, opened 2022-06-08
Editorial: Missing unwrapping of normal completion records

h2oche:
In `Evaluation` SDOs of [13.6.6 Function Calls](https://tc39.es/ecma262/#sec-function-calls-runtime-semantics-evaluation) and [13.5.3 typeof Operator](https://tc39.es/ecma262/#sec-typeof-operator), explicit conversion from normal completion value, which is the result of evaluation, to its `[[Value]]` seem to be needed since it is used as a reference record in the condition of if step.


jmdyck:
In open PR #2744,
- commit 6 (e5eee21) handles the first of these (plus a bunch of similar cases), and
- commit 8 (c697301) handles the second.

PR #2796, opened 2022-06-13
Editorial: Extend CopyDataProperties to support excluding properties by value

gibson042:
This is another Temporal pull-forward, primarily for treating as absent any property with *undefined* value (cf. https://github.com/tc39/proposal-temporal/pull/2245).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6763,7 +6763,8 @@
         CopyDataProperties (
           _target_: an Object,
           _source_: an ECMAScript language value,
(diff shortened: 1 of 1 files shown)


michaelficarra:
This should come in with the Temporal PR.

bakkot:
Specifically, I'd prefer not to add unused code paths, especially so far from Temporal actually hitting stage 4. It's fine to stage these in separate commits in the final PR, if you'd like, but that's something to do when the time comes.

PR #2799, opened 2022-06-16
Editorial: Add an optional type restriction to IterableToList

gibson042:
Mirrors [CreateListFromArrayLike](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-createlistfromarraylike) and supports removal of [Temporal IterableToListOfType](https://tc39.es/proposal-temporal/#sec-iterabletolistoftype) and [ECMA-402 StringListFromIterable](https://tc39.es/ecma402/#sec-createstringlistfromiterable).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6398,7 +6398,7 @@
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It is used to create a List value whose elements are provided by the indexed properties of _obj_. _elementTypes_ contains the names of ECMAScript Language Types that are allowed for element values of the List that is created.</dd>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -38486,7 +38492,7 @@ <h1>%TypedArray%.from ( _source_ [ , _mapfn_ [ , _thisArg_ ] ] )</h1>
             1. Let _mapping_ be *true*.
           1. Let _usingIterator_ be ? GetMethod(_source_, @@iterator).
           1. If _usingIterator_ is not *undefined*, then
-            1. Let _values_ be ? IterableToList(_source_, _usingIterator_).
+            1. Let _values_ be ? IterableToList(_source_, ~empty~, _usingIterator_).
```
ljharb, line 38495: seems strange to add this arg in the middle. why not at the end?

PR #2802, opened 2022-06-22
Editorial: Clarify data property inheritance

### LINKED ISSUE #2798: Wording about inheriting data properties is misleading
Originally came up in https://bugs.chromium.org/p/v8/issues/detail?id=12930#c3

The wording in https://tc39.es/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots is misleading in that it says

"[Data properties](https://tc39.es/ecma262/#sec-object-type) of the [[Prototype]] object are inherited (and visible as properties of the child object) for the purposes of get access, but not for set access."

This, however, can be misinterpreted to mean that one can always shadow data properties on the prototype by setting a property lower on the chain. But that's false because of the override mistake. We should reword this.

syg:
Closes #2798


bakkot (approved):
Still some details about extensibility and so on which aren't covered, but this seems like an improvement.

spec.html
```diff
@@ -12480,7 +12480,7 @@ <h1>Ordinary and Exotic Objects Behaviours</h1>
 
   <emu-clause id="sec-ordinary-object-internal-methods-and-internal-slots">
     <h1>Ordinary Object Internal Methods and Internal Slots</h1>
-    <p>All ordinary objects have an internal slot called [[Prototype]]. The value of this internal slot is either *null* or an object and is used for implementing inheritance. Data properties of the [[Prototype]] object are inherited (and visible as properties of the child object) for the purposes of get access, but not for set access. Accessor properties are inherited for both get access and set access.</p>
+    <p>All ordinary objects have an internal slot called [[Prototype]]. The value of this internal slot is either *null* or an object and is used for implementing inheritance. Data properties of the [[Prototype]] object are inherited (and visible as properties of the child object) for get access; for set access, only writability is inherited, with assignment setting properties on the child object if the data property on the [[Prototype]] object is writable. Accessor properties are inherited for both get access and set access.</p>
```
michaelficarra, line 12483:
```suggestion
    <p>All ordinary objects have an internal slot called [[Prototype]]. The value of this internal slot is either *null* or an object and is used for implementing inheritance. For data property get access, when the property is missing from the child object, the same-named property of the [[Prototype]] object is inherited (and visible as if it was a property of the child object). For data property set access, when the property is missing from the child object, the writability of the same-named property from the [[Prototype]] object is used to determine whether a property is created on the child object or the assignment fails. Accessor properties are inherited for both get access and set access.</p>
```

Ayaz410808 (commented):
Htlm

spec.html
syg: I didn't like this rewording because leading with "For data property get access" read a bit funny to me, like you decide when you do a Get that it's a data property Get vs an accessor Get. I took another stab at rewording in a new commit.

spec.html
```diff
@@ -12480,7 +12480,7 @@ <h1>Ordinary and Exotic Objects Behaviours</h1>
...
     <h1>Ordinary Object Internal Methods and Internal Slots</h1>
-    <p>All ordinary objects have an internal slot called [[Prototype]]. The value of this internal slot is either *null* or an object and is used for implementing inheritance. Data properties of the [[Prototype]] object are inherited (and visible as properties of the child object) for the purposes of get access, but not for set access. Accessor properties are inherited for both get access and set access.</p>
+    <p>All ordinary objects have an internal slot called [[Prototype]]. The value of this internal slot is either *null* or an object and is used for implementing inheritance. Assume a property named _P_ is missing from an ordinary object _O_ but exists on its [[Prototype]] object. If it is a data property, _O_ inherits it for get accesses of _P_ on _O_, making it visible as if it were a property of _O_. If it is a data property and is writable, set accesses of _P_ on _O_ create a new same-named data property on _O_. If it is a data property and is not writable, set accesses of _P_ on _O_ fail. If it is an accessor property, the accessor is inherited for both get access and set access.</p>
```
michaelficarra, line 12483: Getting close!

```suggestion
    <p>All ordinary objects have an internal slot called [[Prototype]]. The value of this internal slot is either *null* or an object and is used for implementing inheritance. Assume a property named _P_ is missing from an ordinary object _O_ but exists on its [[Prototype]] object. If _P_ refers to a data property on the [[Prototype]] object, _O_ inherits it for get access, making it behave as if _P_ was a property of _O_. If _P_ refers to a writable data property on the [[Prototype]] object, set access of _P_ on _O_ creates a new data property named _P_ on _O_. If _P_ refers to a non-writable data property on the [[Prototype]] object, set access of _P_ on _O_ fails. If _P_ refers to an accessor property on the [[Prototype]] object, the accessor is inherited by _O_ for both get access and set access.</p>
```

syg: Adopted.

michaelficarra (approved):

Issue #2803, opened 2022-06-23
Relax HostResolveImportedModule idempotency requirement when referencingScriptOrModule is null


nicolo-ribaudo:
_HostResolveImportedModule(...) specification in ecma262: https://tc39.es/ecma262/#sec-hostresolveimportedmodule_
_HostResolveImportedModule(...) specification in HTML: https://html.spec.whatwg.org/#hostresolveimportedmodule(referencingscriptormodule,-modulerequest)_

Consider the third implementation requirement:
> - Each time this operation is called with a specific _referencingScriptOrModule_, _specifier_ pair as arguments it must return the same [Module Record](https://tc39.es/ecma262/#sec-abstract-module-records) instance if it completes normally.

Also, remember that first parameter can be `null` when, for example, the _HostResolveImportedModule_ calls comes from a dynamic import in an inline event handler on an HTML element.

Consider this example:
```html
<!-- ./index.html -->
<script>globalThis.foo = 1;</script>
<button type="button" onclick="import('./mod.js').then(console.log)">Click me</button>
<iframe src="./iframe.html"></iframe>
```
```html
<!-- ./iframe.html -->
<script>globalThis.foo = 2;</script>
<button type="button" onclick="import('./mod.js').then(console.log)">Click me</button>
```
```js
<!-- ./iframe.html -->
console.log("Hi :) I have been executed");
export const foo = globalThis.foo;
```

If you first click on the first button and then on the second one, there are two _HostResolveImportedModule(`null`, `"./mod.js"`)_ calls: according to ecma262 they should return the same module record, thus both `console.log`s will log the same namespace object and not `{ foo: 1 }`, `{ foo: 2 }`.

However, [the HTML spec](https://html.spec.whatwg.org/#hostresolveimportedmodule(referencingscriptormodule,-modulerequest)) does not respect this requirement: if _referencingScriptOrModule_ is `null`, the module resolution depends on the "[current settings object](https://html.spec.whatwg.org/multipage/webappapis.html#current-settings-object)", which is the [[HostDefined]] slot of the current Realm.

---

I think that we can fix this in two ways:
1. Change the third `HostResolveImportedModule` constraint to something like this:
    > - Each time this operation is called with a specific _referencingScriptOrModule_, _specifier_ pair as arguments, **if _referencingScriptOrModule_ is not null** it must return the same [Module Record](https://tc39.es/ecma262/#sec-abstract-module-records) instance if it completes normally.
    
[...]

PR #2806, opened 2022-06-24
Editorial: Clarify that the `HostResolveImportedModule` cache is per-Realm

nicolo-ribaudo:
This commit clarifies that `HostResolveImportedModule` is required to return the same Module Record when called two times with the same (`null`, _specifier_) arguments only if the two calls are executed with the same current Realm.

Fixes https://github.com/tc39/ecma262/issues/2803.

I don't know if this is normative or editorial, because I don't know what is the current requirement:
- (Editorial) The cache is already implicitly per-Realm, as suggested by @dominic in https://github.com/tc39/ecma262/issues/2803#issuecomment-1164675594
- (Normative) If the referencing script/module is `null` there is currently no caching requirement, as suggested by @codehag in https://github.com/tc39/ecma262/issues/2803#issuecomment-1165251478
- (Normative) If `HostResolveImportedModule` is called with the same (`null`, _specifier_) arguments it must return the same Module Record even if it is in a different Realm, which was my understanding when reading the spec.

The HTML spec already follows the updated requirement proposed by this PR.

An alternative to this PR is https://github.com/nicolo-ribaudo/ecma262/commit/0877ff734cce3c8964cb92c890d54cfaa017834a, which has the same behavior but makes the caching behavior more explicit by moving it from a host requirement to an Abstract Operation that caches values in a per-Realm List.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -27790,7 +27790,7 @@
             If a Module Record corresponding to the pair _referencingScriptOrModule_, _specifier_ does not exist or cannot be created, an exception must be thrown.
           </li>
           <li>
-            Each time this operation is called with a specific _referencingScriptOrModule_, _specifier_ pair as arguments it must return the same Module Record instance if it completes normally.
+            Each time this operation is called, if it completes normally and the (_referencingScriptOrModule_, _specifier_, current Realm Record) triple is the same, it must return the same Module Record instance.
           </li>
         </ul>
         <p>Multiple different _referencingScriptOrModule_, _specifier_ pairs may map to the same Module Record instance. The actual mapping semantic is host-defined but typically a normalization process is applied to _specifier_ as part of the mapping process. A typical normalization process would include actions such as alphabetic case folding and expansion of relative and abbreviated path specifiers.</p>


spec.html
```diff
@@ -27803,7 +27803,7 @@ <h1>
             If a Module Record corresponding to the pair _referencingScriptOrModule_, _specifier_ does not exist or cannot be created, an exception must be thrown.
           </li>
           <li>
-            Each time this operation is called with a specific _referencingScriptOrModule_, _specifier_ pair as arguments it must return the same Module Record instance if it completes normally.
+            Each time this operation is called, if it completes normally and the (_referencingScriptOrModule_, _specifier_, current Realm) triple is the same, it must return the same Module Record instance.
```
nicolo-ribaudo, line 27806:
```suggestion
            Each time this operation is called, if it completes normally and the (_referencingScriptOrModule_, _specifier_, current Realm Record) triple is the same, it must return the same Module Record instance.
```

nicolo-ribaudo:
Update: the "alternative commit" I linked (https://github.com/nicolo-ribaudo/ecma262/commit/0877ff734cce3c8964cb92c890d54cfaa017834a) is now also used in the Module Blocks proposal because it's needed to differentiate resolution of "external modules" (that go through HostResolveImportedModule) from the resolution of module blocks (https://tc39.es/proposal-js-module-blocks/#sec-resolveimportedmodule).

michaelficarra:
The editors like this PR as-is, and believe it resolves #2803. The "alternative" strategy is also fine, but it can be brought in with a feature that needs it.

@codehag does this resolve your confusion around a `null` `referencingScriptOrModule` from https://github.com/tc39/ecma262/issues/2803#issuecomment-1165289120?

michaelficarra (approved):

Issue #2786, opened 2022-05-26
Better links for "(extended) mathematical value" and "Number value"


bergus:
There are multiple editorial issues with the definitions of numerical values and conversion between them.

* In many places, the phrase "Number value" is linked to [/#number-value](https://tc39.es/ecma262/#number-value), which describes the conversion algorithm. This link should only be placed in the phrase "the Number value for". Unless followed by the word "for", it should link to the definition of the Number type instead ([/#sec-ecmascript-language-types-number-type](https://tc39.es/ecma262/#sec-ecmascript-language-types-number-type)), possibly with only "Number" being linked but not "value".
* Similar for the phrase "mathematical value" being linked to [/#mathematical-value](https://tc39.es/ecma262/#mathematical-value). The link to the conversion algorithm should only be placed in the phrase "the mathematical value of". In itself, the term should link to the definition given in the list at the start of the [*Mathematical Operations* section](https://tc39.es/ecma262/#sec-mathematical-operations), "*Mathematical values: Arbitrary real numbers, used as the default numeric type.*".
* Similar for the phrase "extended mathematical value" (only two refs), which should link to the definition not the conversion algorithm.
* In [RoundMVResult(n)](https://tc39.es/ecma262/#sec-roundmvresult), the "mathematical value denoted by" is defined, but all the links point to the mathematical value conversion algorithm again. It should probably get its own identifier, or the links should get removed altogether from occurrences of that phrase.
* Where "𝔽(x)" is used instead of the phrase "the Number value for x", it does link to [/#𝔽](https://tc39.es/ecma262/#%F0%9D%94%BD) in the algorithm conventions for mathematical conventions sections, which only says "*is defined in [6.1.6.1](https://tc39.es/ecma262/#sec-ecmascript-language-types-number-type)*"). Can we link occurrences of "𝔽" directly to the definition [/#number-value](https://tc39.es/ecma262/#number-value)?

PR #2810, opened 2022-07-05
Markup: give math conversion methods AOIDs

bakkot:
This only affects the published biblio (making these three things be `op`s instead of `term`s), which makes it clear they can be invoked. These should always been written this way.


michaelficarra (approved):

bakkot:
Apparently this breaks the links; gotta figure out why.

michaelficarra:
Related: #2786

PR #2812, opened 2022-07-06
Normative: only coerce once in BigInt constructor

bakkot:
(This was raised on [the discourse](https://es.discourse.group/t/consider-rewriting-bigint-value-only-convert-to-a-primitive-once/1391).)

Consider

```js
let first = true;
let v = {
  [Symbol.toPrimitive]() {
    console.log('called');
    if (first) {
      first = false;
      return 'string';
    } else {
      return 0n;
    }
  }
};
BigInt(v);
```

Per spec this should print `called` twice and complete normally - step 3 of [the BigInt constructor](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-bigint-constructor-number-value) will observe that the first call to `ToPrimitive` returns a value which is not a number and then step 4 will invoke [`ToBigInt`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-tobigint) on the original, pre-coercion value, which will invoke `Symbol.toPrimitive` again.

In V8, SpiderMonkey, JavaScriptCore, and XS, this code will in fact only print `called` once and then throw a `TypeError` because the initial `'string'` can't be coerced to a BigInt - i.e., in step 4 they use the post-coercion value, not the pre-coercion value. That's the behavior proposed in this PR.

GraalJS and Engine262 implement the spec faithfully.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -31249,7 +31249,7 @@
           1. If NewTarget is not *undefined*, throw a *TypeError* exception.
           1. Let _prim_ be ? ToPrimitive(_value_, ~number~).
           1. If Type(_prim_) is Number, return ? NumberToBigInt(_prim_).
-          1. Otherwise, return ? ToBigInt(_value_).
+          1. Otherwise, return ? <emu-meta suppress-effects="user-code">ToBigInt(_prim_)</emu-meta>.
         </emu-alg>
 
         <emu-clause id="sec-numbertobigint" type="abstract operation">


syg:
haha oh no

apaprocki:
@bakkot So `Date` constructor 4c is basically the same issue?
```
let first = true;
let v = {
  [Symbol.toPrimitive]() {
    console.log('called');
    if (first) {
      first = false;
      return 0;
    } else {
      return 'string';
    }
  }
};
new Date(v);
```
```
[Log] called
< Wed Dec 31 1969 19:00:00 GMT-0500 (EST)
```

bakkot:
@apaprocki The Date constructor has a similar pattern but is currently specified correctly: it does `Let v be ? ToPrimitive(value). [...] Let tv be ? ToNumber(v).`, i.e., the second line uses the post-coercion value instead of the pre-coercion value.

(That second like does want a `suppress-effects="user-code"`, though.)

apaprocki:
Ah, my brain read what you wrote (without seeing the 1-line diff) and thought this was to fix behavior to match the spec as opposed to the other way around. Thanks!

michaelficarra (approved):

bakkot:
Tests: https://github.com/tc39/test262/pull/3631

PR #2814, opened 2022-07-07
Editorial: Fix wrong assertion in 'DefineMethodProperty'

h2oche:
`DefinePropertyOrThrow` invocation in [10.2.8 DefineMethodProperty](https://tc39.es/ecma262/#sec-definemethodproperty) can throw an abrupt completion. Consider following JavaScript code:

```js
class A {
  static * ['prototype'] () {}
}
```

- `prototype` property of `A` is defined in step 16 of [15.7.14 Runtime Semantics: ClassDefinitionEvaluation](https://tc39.es/ecma262/#sec-runtime-semantics-classdefinitionevaluation) with property descriptor `{[[Configurable]]: false, ... }`
- It tries to define static generator method `prototype` of `A` in [10.2.8 DefineMethodProperty](https://tc39.es/ecma262/#sec-definemethodproperty) with property descriptor `{[[Configurable]]: true, ...}`.
- Since `ValidateAndApplyPropertyDescriptor` return `false`, it throws a`TypeError` exception.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13576,17 +13576,18 @@
           _key_: a property key or Private Name,
           _closure_: a function object,
           _enumerable_: a Boolean,
-        ): a PrivateElement or ~unused~
+        ): either a normal completion containing either a PrivateElement or ~unused~, or an abrupt completion
       </h1>
       <dl class="header">
       </dl>
       <emu-alg>
-        1. Assert: _homeObject_ is an ordinary, extensible object with no non-configurable properties.
+        1. Assert: _homeObject_ is an ordinary, extensible object.
         1. If _key_ is a Private Name, then
           1. Return PrivateElement { [[Key]]: _key_, [[Kind]]: ~method~, [[Value]]: _closure_ }.
         1. Else,
           1. Let _desc_ be the PropertyDescriptor { [[Value]]: _closure_, [[Writable]]: *true*, [[Enumerable]]: _enumerable_, [[Configurable]]: *true* }.
-          1. Perform ! DefinePropertyOrThrow(_homeObject_, _key_, _desc_).
+          1. Perform ? DefinePropertyOrThrow(_homeObject_, _key_, _desc_).
+          1. NOTE: DefinePropertyOrThrow only returns an abrupt completion when attempting to define a class static method whose _key_ is *"prototype"*.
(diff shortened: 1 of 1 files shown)


anba:
[Early errors](https://tc39.es/ecma262/#sec-class-definitions-static-semantics-early-errors) disallow this case:
> ClassElement : static MethodDefinition
> &emsp; It is a Syntax Error if PropName of MethodDefinition is "prototype".

h2oche:
Oh, I missed it. Thanks.

anba:
Yes, with a computed property name the assertion is indeed incorrect. The assertion in [DefineMethodProperty](https://tc39.es/ecma262/#sec-definemethodproperty), step 1 should also be removed:
> Assert: homeObject is an ordinary, extensible object with no non-configurable properties.

syg (approved):
Good catch. An unfortunate corner...

bakkot (approved):

spec.html
```diff
@@ -13579,17 +13579,18 @@ <h1>
...
       <emu-alg>
-        1. Assert: _homeObject_ is an ordinary, extensible object with no non-configurable properties.
+        1. Assert: _homeObject_ is an ordinary, extensible object.
         1. If _key_ is a Private Name, then
           1. Return PrivateElement { [[Key]]: _key_, [[Kind]]: ~method~, [[Value]]: _closure_ }.
         1. Else,
           1. Let _desc_ be the PropertyDescriptor { [[Value]]: _closure_, [[Writable]]: *true*, [[Enumerable]]: _enumerable_, [[Configurable]]: *true* }.
-          1. Perform ! DefinePropertyOrThrow(_homeObject_, _key_, _desc_).
+          1. Perform ? DefinePropertyOrThrow(_homeObject_, _key_, _desc_).
+          1. NOTE: DefinePropertyOrThrow only returns an abrupt completion when attempting to define a class static method whose _key_ is *"prototype"*.
```
michaelficarra, line 13593: Can we move the note up to before the DefinePropertyOrThrow step? I'd prefer to read the note first.

syg: I prefer it after. With none of us caring that much and Kevin voting in favor of doing the minimal effort thing, let's leaving it as is.

PR #2815, opened 2022-07-07
Editorial: Fix completion in GeneratorStart

h2oche:
It seems that step 4.b.ii of [27.5.3.1 GeneratorStart](https://tc39.es/ecma262/#sec-generatorstart) needs to be wrapped with `Completion` for clarity as [27.6.3.2 AsyncGeneratorStart](https://tc39.es/ecma262/#sec-asyncgeneratorstart) since later steps assume that `result` is completion.

In addition, some abstract closures seem to return non-completion value, which violates `Completion` check in the fixed step in `GeneratorStart`. I'm not sure whether [conversion rules](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-implicit-normal-completion) can be applied to those cases.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7185,7 +7185,7 @@
         1. Let _closure_ be a new Abstract Closure with no parameters that captures _list_ and performs the following steps when called:
           1. For each element _E_ of _list_, do
             1. Perform ? GeneratorYield(CreateIterResultObject(_E_, *false*)).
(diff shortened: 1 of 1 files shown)


linusg:
This has since been fixed in https://github.com/tc39/ecma262/pull/2861 and https://github.com/tc39/ecma262/pull/3601.

PR #2817, opened 2022-07-07
Editorial: IfAbrupt macros only take completion records

bakkot:
As of #2547 (edit: plus #2744 and #2842), it is always statically knowable whether a given expression at a given point is a Completion Record or not a Completion Record. So the various completion record unwrapping macros don't need to have a "what if it is not a completion record" case.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -895,8 +895,9 @@
         </emu-alg>
         <p>mean the same thing as:</p>
         <emu-alg example>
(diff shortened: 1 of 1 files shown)


michaelficarra (approved):
😍

syg (approved):

michaelficarra:
Waiting on #2842.

PR #2818, opened 2022-07-07
Normative: avoid triggering `throw` in corner case in async generators

### LINKED ISSUE #2813: `yield *` in an async iterator should not call `.throw` when the inner iterator produces a rejected promise
Consider the following horrible program:
```js
'use strict';
var console = console || { log: print };

let inner = {
  [Symbol.asyncIterator]: () => ({
    next() {
      return Promise.resolve({ done: false, value: Promise.reject('rejection') });
    },
    throw(e) {
      console.log('throw called with', e);
      throw e;
    },
    return(e) {
      console.log('return called with', e);
      return {};
    },
  }),
};

async function* outer() {
  yield* inner;
}

(async () => {
  for await (let x of outer()) {
    console.log(x);
  }
})().catch(e => {
  console.log('threw', e);
});
```

Per spec, this will print "throw called with 'rejection'". Specifically, the [evaluation semantics for `yield*`](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#sec-generator-function-definitions-runtime-semantics-evaluation) step 7.a.vi will invoke [`AsyncGeneratorYield`](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgeneratoryield) passing `Promise.reject('rejection')`, which will then be `Await`'d, which will throw, causing the `received` alias to hold a throw completion. The following machinery then acts as if `.throw` had been explicitly called on `outer`, which is the only other way that `received` can hold a throw completion - the behavior in that case being to invoke `.throw` on `inner`.

I think this possibility was simply overlooked during the spec'ing of async generators. This is the only place in the entire spec that `.throw` is called without a user explicitly invoking it at some point. (Previously I'd thought there were no such places.) And the aliases certainly imply that the algorithm is assuming that a throw completion must have come from the user, rather than from the iterator itself.

This behavior seems wrong to me. Normally when consuming an iterator the spec will either a.) determine that the iterator itself is broken (e.g. it threw), in which case it is assumed to have done its own cleanup and the spec will not make further calls to any of its methods, or b.) determine that the iterator is fine but for some reason the consumer needs to exit early, in which case the spec will call `.return` (not `.throw`) to give the iterator a chance to clean up.

---

I am inclined to treat this as in (b). Nothing in the spec prevents custom async iterators from returning promises - see the snippet in [this comment](https://github.com/tc39/proposal-array-from-async/issues/19#issuecomment-1175655515). So I think we ought to treat the `await iterResult.value` as being something the outer iterator is doing, rather than a violation of the iteration protocol, so that if this `await` happens to throw then the inner iterator should be closed with `.return`.

I could probably be persuaded to treat this as in (a), i.e. to say that this is a violation of the iteration protocol and so the spec need not clean up the inner iterator.

The current behavior seems bad, though.

---

For the snippet above, JavaScriptCore and V8 implement the spec (they print "throw called with rejection" and then "threw rejection"). SpiderMonkey, Graal, and XS only print "threw rejection" (they also don't close the inner iterator).

bakkot:
Fixes #2813.

The idea is that if you are doing `yield* m` on a manual async iterator `m` (i.e., one which is not an async generator), it is possible for the inner iterator to yield `{ next: promise, done: false }`. And if that promise rejects, it is treated as if someone had called `.throw` on the outer generator (the one doing `yield*`). That's inconsistent with the way the spec usually consumes iterators and is almost certainly a bug caused by the factoring of `AsyncGeneratorYield`. (Async generators await their result before yielding it, so it doesn't come up without a manually implemented iterator.)

There's only really three consistent options, that I can see.
1. treat this as a valid thing for a manual async iterator to do which happens to cause the outer generator to throw, in which case the inner iterator should be cleaned up before the exception propagates (which is what this PR does)
2. treat this as a protocol violation by the inner iterator, in which case neither `.throw` or `.return` should be called on the inner iterator (as if calling `.next` had thrown)
3. don't await the result from the inner iterator at all (which would mean it's possible for an async generator to yield a Promise without first unwrapping it, which is maybe surprising, but on the other hand is more consistent with treating `yield*` as being a completely transparent delegation to the inner iterator). This option would also reduce the number of microtask ticks entailed by async generators doing `yield*` in general. I've opened an alternative PR in https://github.com/tc39/ecma262/pull/2819 which implements this option.

This PR takes the first option.

I don't really like the second option, because `for await` does _not_ treat an async iterator yielding a promise as being a protocol violation - see [this snippet](https://github.com/tc39/proposal-array-from-async/issues/19#issuecomment-1175655515).

I actually kind of prefer the third option, but it has broader implications than this PR does, and so is more likely to break stuff. On the other hand, it also would improve the performance of async generators, which is nice.

---

As mentioned in the issue, this does not match any existing engine, but engines are inconsistent here, so the change in this PR is probably web compatible.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -7152,6 +7152,20 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-ifabruptcloseasynciterator" aoid="IfAbruptCloseAsyncIterator">
+      <h1>IfAbruptCloseAsyncIterator ( _value_, _iteratorRecord_ )</h1>
+      <p>IfAbruptCloseAsyncIterator is a shorthand for a sequence of algorithm steps that use an Iterator Record. An algorithm step of the form:</p>
+      <emu-alg>
+        1. IfAbruptCloseAsyncIterator(_value_, _iteratorRecord_).
+      </emu-alg>
+      <p>means the same thing as:</p>
+      <emu-alg>
+        1. Assert: _value_ is a Completion Record.
+        1. If _value_ is an abrupt completion, return ? AsyncIteratorClose(_iteratorRecord_, _value_).
+        1. Else, set _value_ to _value_.[[Value]].
+      </emu-alg>
+    </emu-clause>
+
     <emu-clause id="sec-createiterresultobject" type="abstract operation">
       <h1>
         CreateIterResultObject (
@@ -23974,7 +23988,11 @@
             1. Let _done_ be ? IteratorComplete(_innerResult_).
             1. If _done_ is *true*, then
               1. Return ? IteratorValue(_innerResult_).
-            1. If _generatorKind_ is ~async~, set _received_ to Completion(AsyncGeneratorYield(? IteratorValue(_innerResult_))).
+            1. If _generatorKind_ is ~async~, then
+              1. Set _value_ to ? IteratorValue(_innerResult_).
+              1. Let _unwrapped_ be Completion(Await(_value_)).
+              1. IfAbruptCloseAsyncIterator(_unwrapped_, _iteratorRecord_).
+              1. Set _received_ to Completion(AsyncGeneratorYield(_unwrapped_)).
             1. Else, set _received_ to Completion(GeneratorYield(_innerResult_)).
           1. Else if _received_.[[Type]] is ~throw~, then
             1. Let _throw_ be ? GetMethod(_iterator_, *"throw"*).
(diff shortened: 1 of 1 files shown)


michaelficarra:
I like the sound of option 3. Can you go into more detail about the effects of such a change?

bergus:
> I don't really like the second option, because `for await` does *not* treat an async iterator yielding a promise as being a protocol violation [and assigns the promise to the iteration variable]

That is the reason why I don't like the *first* option - if `for await` ignores promises as the `value` of an IteratorResult, then why should `yield*` wait for them? `for await` will happily continue the iteration and not clean up the iterator, so `yield*` shouldn't either.
But I don't like the second option either - not cleaning up the inner iterator, but then throwing an exception from the generator (after which the inner iterator won't be advanced any more) seems wrong.

I can even see a fourth option: Don't await the result from an inner iterator in `yield` at all, but *do* await the result `.value` in `for await`. Then, if this is a rejected promise, have it clean up the looped iterator, which in turn will make `yield*` clean up the inner iterator. So basically `yield*` working like in the third option, but having every `for await (const x of asyncIterable) { … }` work like `for await (const _x of asyncIterable) { const x = await _x; … }`.

That said, of all these I do prefer option 3.

bakkot:
> I can even see a fourth option: Don't await the result from an inner iterator in `yield*` at all, but do await the result `.value` in `for await`

That option was discussed and rejected during the original design of async iterators - it's option 2 in [these slides](https://docs.google.com/presentation/d/1U6PivKbFO0YgoFlrYB82MtXf1ofCp1xSVOODOvranBM/edit#slide=id.g223fba4116_0_144); see also discussion in [the notes](https://github.com/tc39/notes/blob/main/meetings/2017-05/may-25.md#15iva-revisiting-async-generator-yield-behavior) and [the issue tracker](https://github.com/tc39/proposal-async-iteration/issues/93).

I'd definitely prefer to avoid changing the behavior `yield` or `for await`, at this point; I'm only proposing changing the much rarer case of `yield*`.

bakkot:
Closing as we got consensus for the alternative https://github.com/tc39/ecma262/pull/2819. We may need to revive this if that PR turns out not to be web compatible, though.

PR #2821, opened 2022-07-08
Editorial: clarify spec values that have identity and its implications

### LINKED ISSUE #2405: clarify which spec values do and do not carry identity
In a recent PR, some found it valuable to call out the fact that Private Name values have identity. This got me thinking that we should be explicit (say, in 6.2 ES Specification Types) about which spec values have identity. I'm thinking that we could say that all spec values aside from a few have identity. The ones I can think of are (extended) mathematical values, primitive ECMAScript language values, sequences (as in code points for source text and code units for strings), and spec enum values. There are some types, such as Data Block, which have complex identities, and which confuse identity and equality.

Values that have identity should be referred to as "a"/"an" (an empty List) and constructed using "a new" (a new List). Values without identity should be referred to as "the" (the empty string) and not constructed, just described.

We can then audit that we're not doing things like assuming structural comparison for things that have identity.

### LINKED ISSUE #2794: use consistent phrasing for comparing string values
@gibson042 asks what our conventions should be for comparing strings in https://github.com/tc39/ecma262/pull/2721/files/#r890585298.

I think "equals" is indeed bad, but we have many options available:

* `SameValue`
* "is the same string value"
* "is the same sequence of code units"
* `SameValueNonNumeric`
* "is"

Let's pick something and make it consistent.

michaelficarra:
Opened as draft since I only have the prose for now. Once we get this part agreed on and settled, I'll audit for uses of comparison, mutation, article choice, construction, etc.

New section rendering: https://ci.tc39.es/preview/tc39/ecma262/pull/2821#sec-identity

Fixes #2405.


spec.html
```diff
@@ -1003,8 +1003,8 @@ <h1>Static Semantics</h1>
       <h1>Mathematical Operations</h1>
       <p>This specification makes reference to these kinds of numeric values:</p>
       <ul>
-        <li><em>Mathematical values</em>: Arbitrary real numbers, used as the default numeric type.</li>
-        <li><em>Extended mathematical values</em>: Mathematical values together with +&infin; and -&infin;.</li>
+        <li><dfn id="mathematical-value" variants="mathematical value,mathematical values">Mathematical values</dfn>: Arbitrary real numbers, used as the default numeric type.</li>
+        <li><dfn id="extended-mathematical-value" variants="extended mathematical value,extended mathematical values">Extended mathematical values</dfn>: Mathematical values together with +&infin; and -&infin;.</li>
```
bergus, line 1007: This is beginning to fix #2786, :+1:

spec.html
```diff
@@ -1030,7 +1030,13 @@ <h1>Mathematical Operations</h1>
     <emu-clause id="sec-value-notation">
       <h1>Value Notation</h1>
       <p>In this specification, ECMAScript language values are displayed in *bold*. Examples include *null*, *true*, or *"hello"*. These are distinguished from longer ECMAScript code sequences such as `Function.prototype.apply` or `let n = 42;`.</p>
-      <p>Values which are internal to the specification and not directly observable from ECMAScript code are indicated with a ~sans-serif~ typeface. For instance, a Completion Record's [[Type]] field takes on values like ~normal~, ~return~, or ~throw~.</p>
+    </emu-clause>
+
+    <emu-clause id="sec-identity">
+      <h1>Identity</h1>
+      <p>In this specification, the word &ldquo;is&rdquo; is used to identify two values, as in &ldquo;If _bool_ is *true*, then&rdquo;. This identification is different when applied to values which have identity than it is when applied to those without identity.</p>
```
bergus, line 1037: I would maybe suggest to use two different wordings: "*is*" for values (without identity), "*is same as*" for objects (values with identity).
Does the spec actually compare objects that often? I can only think of `SameValueNonNumeric`.

spec.html
```diff
@@ -1030,7 +1030,13 @@ <h1>Mathematical Operations</h1>
     <emu-clause id="sec-value-notation">
       <h1>Value Notation</h1>
       <p>In this specification, ECMAScript language values are displayed in *bold*. Examples include *null*, *true*, or *"hello"*. These are distinguished from longer ECMAScript code sequences such as `Function.prototype.apply` or `let n = 42;`.</p>
-      <p>Values which are internal to the specification and not directly observable from ECMAScript code are indicated with a ~sans-serif~ typeface. For instance, a Completion Record's [[Type]] field takes on values like ~normal~, ~return~, or ~throw~.</p>
+    </emu-clause>
+
+    <emu-clause id="sec-identity">
+      <h1>Identity</h1>
```
bergus, line 1036: I wonder if this whole section should be part of the *ECMAScript Data Types and Values* chapter, not the *Algorithm Contentions* one - it's describing semantics of objects, not just a notational convention.

spec.html
```diff
@@ -1030,7 +1030,13 @@ <h1>Mathematical Operations</h1>
...
...ified not only by their innate qualities but also another unobservable quality that makes each such value unique. Values with identity may be thought of as a value without identity paired with an unguessable, unchangeable, universally-unique number: the <em>identity</em>. References to existing values with identity cannot be manifest simply by describing them, as the identity itself is indescribable; instead, references to these values must be explicitly passed from one place to another. Values with identity can have their qualities modified in-place, causing all holders of the value to immediately observe the new qualities. Examples of values with identity include, but are not limited to: <emu-xref href="#sec-object-type">Objects</emu-xref>, including function objects, exotic objects, etc; any kind of Records, including Property Descriptors, PrivateElements, etc; <emu-xref href="#sec-ecmascript-language-types-symbol-type">symbols</emu-xref>; Parse Nodes; Lists; <emu-xref href="#sec-set-and-relation-specification-type">Sets</emu-xref> and Relations; Abstract Closures; Data Blocks; Private Names; execution contexts and execution context stacks; agent signifiers; and WaiterLists.</p>
```
bergus, line 1039: I'm slightly confused by the inclusion of symbols, parse nodes, private names and agent signifiers. All of these are immutable, they cannot "*have their qualities modified in-place*". Yes, they do have an "*unobservable quality that makes each such value unique*", but is that an "identity"? Specifically, they can be serialised and copied and do not need to have their references shared, as long as the "*unguessable, unchangeable, universally-unique number*" is copied with them.

I would prefer extra elaboration, possibly even a trichotomy instead of a dichotomy.

bergus: Maybe just add "mutable":
```suggestion
      <p>In contrast, <dfn variants="value with identity">values with identity</dfn> are identified not only by their innate qualities but also another unobservable quality that makes each such value unique. Values with identity may be thought of as a value without identity paired with an unguessable, unchangeable, universally-unique number: the <em>identity</em>. References to mutable values with identity cannot be manifest simply by describing them, as the identity itself is indescribable; instead, references to these values must be explicitly passed from one place to another. Mutable values with identity can have their qualities modified in-place, causing all holders of the value to immediately observe the new qualities. Examples of values with identity include, but are not limited to: <emu-xref href="#sec-object-type">Objects</emu-xref>, including function objects, exotic objects, etc; any kind of Records, including Property Descriptors, PrivateElements, etc; <emu-xref href="#sec-ecmascript-language-types-symbol-type">symbols</emu-xref>; Parse Nodes; Lists; <emu-xref href="#sec-set-and-relation-specification-type">Sets</emu-xref> and Relations; Abstract Closures; Data Blocks; Private Names; execution contexts and execution context stacks; agent signifiers; and WaiterLists.</p>
```

spec.html
michaelficarra: I made this change so that everything mentioned in the new section links somewhere useful, but I should probably split it into a separate PR since it's useful on its own.

michaelficarra:
Updated based on @syg's feedback. @tc39/ecma262-editors Please take another look.

spec.html
```diff
@@ -1030,7 +1030,13 @@ <h1>Mathematical Operations</h1>
...
+      <p><dfn variants="values without identity,value without identity">Values without identity</dfn> are identified by the set of all of their innate qualities &mdash; qualities such as the magnitude of an integer or the length of a sequence. Because of this, values without identity may be manifest anywhere simply by fully describing their qualities. It is not meaningful to modify a value that does not have identity. Examples of values without identity include, but are not limited to: <emu-xref href="#sec-ecmascript-language-types-boolean-type">Booleans</emu-xref>; mathematical values and extended mathematical values; <emu-xref href="#sec-ecmascript-language-types-number-type">Numbers</emu-xref>; <emu-xref href="#sec-ecmascript-language-types-bigint-type">BigInts</emu-xref>; *null*; *undefined*; sequences, including <emu-xref href="#sec-ecmascript-language-types-string-type">Strings</emu-xref>, ECMAScript source text, surrogate pairs, Directive Prologues, etc; UTF-16 code units; Unicode code points; enums; abstract operations, including syntax-directed operations, host hooks, etc; and ordered pairs.</p>
```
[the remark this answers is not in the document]
ljharb: other than "manifest", "reified" or "materialized"?

spec.html
```diff
@@ -1030,7 +1030,13 @@ <h1>Mathematical Operations</h1>
...
... with identity">value with identity</dfn> is unique and therefore only equivalent to itself. Values with identity are like values without identity but with an additional unguessable, unchangeable, universally-unique quality called <em>identity</em>. References to existing values with identity cannot be manifest simply by describing them, as the identity itself is indescribable; instead, references to these values must be explicitly passed from one place to another. Values with identity can have their qualities (except their identity) changed in-place, causing all holders of the value to immediately observe the new qualities. Examples of values with identity include, but are not limited to: <emu-xref href="#sec-object-type">Objects</emu-xref>, including function objects, exotic objects, etc; any kind of Records, including Property Descriptors, PrivateElements, etc; <emu-xref href="#sec-ecmascript-language-types-symbol-type">symbols</emu-xref>; Parse Nodes; Lists; <emu-xref href="#sec-set-and-relation-specification-type">Sets</emu-xref> and Relations; Abstract Closures; Data Blocks; Private Names; execution contexts and execution context stacks; agent signifiers; and WaiterLists.</p>
```
syg, line 1039:
```suggestion
      <p>In contrast, each <dfn variants="values with identity">value with identity</dfn> is unique and therefore only equal to itself. Values with identity are like values without identity but with an additional unguessable, unchangeable, universally-unique quality called <em>identity</em>. References to existing values with identity cannot be manifest simply by describing them, as the identity itself is indescribable; instead, references to these values must be explicitly passed from one place to another. Values with identity can have their qualities (except their identity) changed in-place, causing all holders of the value to immediately observe the new qualities. Examples of values with identity include, but are not limited to: <emu-xref href="#sec-object-type">Objects</emu-xref>, including function objects, exotic objects, etc; any kind of Records, including Property Descriptors, PrivateElements, etc; <emu-xref href="#sec-ecmascript-language-types-symbol-type">symbols</emu-xref>; Parse Nodes; Lists; <emu-xref href="#sec-set-and-relation-specification-type">Sets</emu-xref> and Relations; Abstract Closures; Data Blocks; Private Names; execution contexts and execution context stacks; agent signifiers; and WaiterLists.</p>
```

spec.html
bakkot: I agree with this comment. I think I'd say "Some values with identity are mutable, that is, they can have their qualities modified in place..."

michaelficarra:
Thoughts on "characteristics" as an alternative to "qualities"?

spec.html
```diff
@@ -44947,7 +44942,7 @@ <h1>prototype</h1>
         <p>Whenever an AsyncGeneratorFunction instance is created another ordinary object is also created and is the initial value of the async generator function's *"prototype"* property. The value of the prototype property is used to initialize the [[Prototype]] internal slot of a newly created AsyncGenerator when the generator function object is invoked using [[Call]].</p>
         <p>This property has the attributes { [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>
         <emu-note>
-          <p>Unlike function instances, the object that is the value of the an AsyncGeneratorFunction's *"prototype"* property does not have a *"constructor"* property whose value is the AsyncGeneratorFunction instance.</p>
+          <p>Unlike function instances, the object that is an AsyncGeneratorFunction's *"prototype"* property does not have a *"constructor"* property whose value is the AsyncGeneratorFunction instance.</p>
```
bakkot, line 44945: I would prefer to leave "the value of" here.

spec.html
```diff
@@ -44840,7 +44835,7 @@ <h1>prototype</h1>
         <p>Whenever a GeneratorFunction instance is created another ordinary object is also created and is the initial value of the generator function's *"prototype"* property. The value of the prototype property is used to initialize the [[Prototype]] internal slot of a newly created Generator when the generator function object is invoked using [[Call]].</p>
         <p>This property has the attributes { [[Writable]]: *true*, [[Enumerable]]: *false*, [[Configurable]]: *false* }.</p>
         <emu-note>
-          <p>Unlike Function instances, the object that is the value of the a GeneratorFunction's *"prototype"* property does not have a *"constructor"* property whose value is the GeneratorFunction instance.</p>
+          <p>Unlike Function instances, the object that is a GeneratorFunction's *"prototype"* property does not have a *"constructor"* property whose value is the GeneratorFunction instance.</p>
```
bakkot, line 44838: I would prefer to leave "the value of" here.

arhadthedev:
Are all values with identity guaranteed to be `Record`s? If yes, it seems we can:

1. Replace a term *values with(out) identity* with a presense/absense of a new `[[UniqueID]]` field

2. Extend abstract operations that create "objects with identity" with assignment of a value to `newObject.[[UniqueID]]`. The value is to be obtained by a call to a new implementation(or host)-defined abstract operation `GetValueUniqueId(object: Any): Any`. The operation may return an address of an underlying memory chunk, freshly created UUID, or sequentially incremented number.

3. Change the section proposed in the PR to something like this:

   > In this specification, the word *is* is used to compare two values through equality, as in “If bool is true, then”. The equality operation returns `true` if either both operands are Records and all their fields are pairwise equal, or both operands are primitive values and all their specification-defined properties are pairwise equal. Otherwise, the equality operation returns `false`.

   Probably, "is" should become a `<def>` if the specification gives it a special meaning.

In other words, I propose to keep the value identity important for comparisons only as the PR states, but move it from a vaguely defined realm into another abstract field.

bakkot:
> Are all values with identity guaranteed to be Records?

No. The spec-internal List type and the ES language Object type both have identity, and neither are Records, for example.

spec.html
```diff
@@ -5969,9 +5976,9 @@ <h1>
       </dl>
       <emu-alg>
         1. If Type(_argument_) is not Number, return *false*.
-        1. If _argument_ is *NaN*, *+&infin;*<sub>𝔽</sub>, or *-&infin;*<sub>𝔽</sub>, return *false*.
-        1. If floor(abs(ℝ(_argument_))) &ne; abs(ℝ(_argument_)), return *false*.
-        1. Return *true*.
+        1. If _argument_ is any of *NaN*, *+&infin;*<sub>𝔽</sub>, or *-&infin;*<sub>𝔽</sub>, return *false*.
+        1. If floor(abs(ℝ(_argument_))) = abs(ℝ(_argument_)), return *true*.
```
bakkot, line 5980: While I am _in general_ on board with the plan to do "if true" cases before "if false" cases, this algorithm was previously basically "if x, return false. if y, return false. if z, return false. return true" and now is "if x, return false. if y, return false, if z', return true. return false". I prefer the original.

spec.html
```diff
@@ -6089,16 +6096,11 @@ <h1>
...
-          1. If _x_ and _y_ are exactly the same sequence of code units (same length and same code units at corresponding indices), return *true*; otherwise, return *false*.
-        1. If Type(_x_) is Boolean, then
-          1. If _x_ and _y_ are both *true* or both *false*, return *true*; otherwise, return *false*.
-        1. If Type(_x_) is Symbol, then
-          1. If _x_ and _y_ are both the same Symbol value, return *true*; otherwise, return *false*.
-        1. If _x_ and _y_ are the same Object value, return *true*. Otherwise, return *false*.
+        1. If _x_ is _y_, return *true*; otherwise, return *false*.
       </emu-alg>
+      <emu-note>
+        The specifics of what it means for _x_ and _y_ to be the same are detailed in <emu-xref href="#sec-identity"></emu-xref>.
```
bakkot, line 6102: I would lightly prefer to leave the original text. We can add the note, but also having the cases doesn't hurt.

spec.html
```diff
@@ -17972,7 +17974,7 @@ <h1>Static Semantics: Early Errors</h1>
           It is a Syntax Error if the goal symbol of the syntactic grammar is |Module| and the StringValue of |IdentifierName| is *"await"*.
         </li>
         <li>
-          It is a Syntax Error if StringValue of |IdentifierName| is the same String value as the StringValue of any |ReservedWord| except for `yield` or `await`.
+          It is a Syntax Error if StringValue of |IdentifierName| is StringValue of any |ReservedWord| except for `yield` or `await`.
```
bakkot, line 17977:
```suggestion
          It is a Syntax Error if the StringValue of |IdentifierName| is the StringValue of any |ReservedWord| except for `yield` or `await`.
```

We're inconsistent about this for SDOs in general and for StringValue in particular, but in this particular case we're not exactly invoking StringValue with the argument "any |ReservedWord|", which doesn't make sense. Rather we're discussing it as an operation more abstractly. So I think at least for the second half of this sentence "the" is justified. And I'd be inclined to put it back in the first half as well.

(There's another few dozen cases of "the StringValue of", including in the previous line, and I don't think changing them ought to be in scope for this PR.)

spec.html
```diff
@@ -26277,7 +26281,7 @@ <h1>Cyclic Module Records</h1>
                 a List of Strings
               </td>
               <td>
-                A List of all the |ModuleSpecifier| strings used by the module represented by this record to request the importation of a module. The List is source text occurrence ordered.
+                A List of all the |ModuleSpecifier| strings used by the module represented by this record to request the importation of a module. The List is in source text occurrence ordered.
```
bakkot, line 26284: This is not a sentence.

spec.html
```diff
@@ -28392,13 +28396,13 @@ <h1>Static Semantics: ExportEntries ( ): a List of ExportEntry Records</h1>
...
           1. Let _localName_ be the sole element of _names_.
-          1. Return a List whose sole element is the ExportEntry Record { [[ModuleRequest]]: *null*, [[ImportName]]: *null*, [[LocalName]]: _localName_, [[ExportName]]: *"default"* }.
+          1. Return a List whose sole element is an ExportEntry Record { [[ModuleRequest]]: *null*, [[ImportName]]: *null*, [[LocalName]]: _localName_, [[ExportName]]: *"default"* }.
         </emu-alg>
         <emu-grammar>ExportDeclaration : `export` `default` ClassDeclaration</emu-grammar>
         <emu-alg>
           1. Let _names_ be BoundNames of |ClassDeclaration|.
           1. Let _localName_ be the sole element of _names_.
-          1. Return a List whose sole element is the ExportEntry Record { [[ModuleRequest]]: *null*, [[ImportName]]: *null*, [[LocalName]]: _localName_, [[ExportName]]: *"default"* }.
+          1. Return a List whose sole element is an ExportEntry Record { [[ModuleRequest]]: *null*, [[ImportName]]: *null*, [[LocalName]]: _localName_, [[ExportName]]: *"default"* }.
```
bakkot, line 28405: I am ok with this but would mildly prefer to factor the record out to an alias, so we could say `Let _entry_ be a new ExportEntry Record { ... }`

spec.html
```diff
@@ -31922,7 +31926,7 @@ <h1>Math.round ( _x_ )</h1>
         <p>It performs the following steps when called:</p>
         <emu-alg>
           1. Let _n_ be ? ToNumber(_x_).
-          1. If _n_ is *NaN*, *+&infin;*<sub>𝔽</sub>, *-&infin;*<sub>𝔽</sub>, or an integral Number, return _n_.
+          1. If _n_ is any of *NaN*, *+&infin;*<sub>𝔽</sub>, *-&infin;*<sub>𝔽</sub>, or an integral Number, return _n_.
```
bakkot, line 31929:
```suggestion
          1. If _n_ is any of *NaN*, *+&infin;*<sub>𝔽</sub>, or *-&infin;*<sub>𝔽</sub>, or _n_ is an integral Number, return _n_.
```
per above comment about it being weird to use "any of" when doing membership tests rather than testing identity.

spec.html
```diff
@@ -39344,18 +39344,15 @@ <h1>
             1. If _targetOffset_ is +&infin;, throw a *RangeError* exception.
             1. If _srcLength_ + _targetOffset_ &gt; _targetLength_, throw a *RangeError* exception.
             1. If _target_.[[ContentType]] &ne; _source_.[[ContentType]], throw a *TypeError* exception.
-            1. If both IsSharedArrayBuffer(_srcBuffer_) and IsSharedArrayBuffer(_targetBuffer_) are *true*, then
-              1. If _srcBuffer_.[[ArrayBufferData]] and _targetBuffer_.[[ArrayBufferData]] are the same Shared Data Block values, let _same_ be *true*; else let _same_ be *false*.
-            1. Else, let _same_ be SameValue(_srcBuffer_, _targetBuffer_).
-            1. If _same_ is *true*, then
+            1. If _srcBuffer_ is _targetBuffer_ or IsSharedArrayBuffer(_srcBuffer_) is *true*, IsSharedArrayBuffer(_targetBuffer_) is *true*, and _srcBuffer_.[[ArrayBufferData]] is _targetBuffer_.[[ArrayBufferData]], then
```
bakkot, line 39347: For this to have the same meaning as before, the reader must parse "if x or y, z, and w" as "if (x) or (y, z, and w)" rather than "if (x or y), z, and w". I definitely do not want to rely on the reader doing that parse.

(I didn't parse it that way at first and was in the process of writing a comment about this not being equivalent before I figured out what you'd intended.)

Also it's too much stuff to go on one line. The previous thing was fine.

spec.html
michaelficarra: I think it is.

spec.html
michaelficarra: I disagree. We can talk about it at editor call.

spec.html
michaelficarra: Agreed. This was kind of an intermediate step toward handling articles. Remember, this is still in draft!

spec.html
michaelficarra: oh I see now

```suggestion
                A List of all the |ModuleSpecifier| strings used by the module represented by this record to request the importation of a module. The List is in source text occurrence order.
```

spec.html
```diff
@@ -1033,6 +1033,13 @@ <h1>Mathematical Operations</h1>
...
...unique and therefore only equal to itself. Values with identity are like values without identity but with an additional unguessable, unchangeable, universally-unique characteristic called <em>identity</em>. References to existing values with identity cannot be manifest simply by describing them, as the identity itself is indescribable; instead, references to these values must be explicitly passed from one place to another. Some values with identity are mutable and therefore can have their characteristics (except their identity) changed in-place, causing all holders of the value to immediately observe the new characteristics. Examples of values with identity include, but are not limited to: <emu-xref href="#sec-object-type">Objects</emu-xref>, including function objects, exotic objects, etc; any kind of Records, including Property Descriptors, PrivateElements, etc; <emu-xref href="#sec-ecmascript-language-types-symbol-type">symbols</emu-xref>; Parse Nodes; Lists; <emu-xref href="#sec-set-and-relation-specification-type">Sets</emu-xref> and Relations; Abstract Closures; Data Blocks; Private Names; execution contexts and execution context stacks; agent signifiers; and WaiterLists.</p>
```
syg, line 1041:
```suggestion
      <p>In contrast, each <dfn variants="values with identity">value with identity</dfn> is unique and therefore only equal to itself. Values with identity are like values without identity but with an additional unguessable, unchangeable, universally-unique characteristic called <em>identity</em>. References to existing values with identity cannot be manifest simply by describing them, as the identity itself is indescribable; instead, references to these values must be explicitly passed from one place to another. Some values with identity are mutable and therefore can have their characteristics (except their identity) changed in-place, causing all holders of the value to observe the new characteristics. Examples of values with identity include, but are not limited to: <emu-xref href="#sec-object-type">Objects</emu-xref>, including function objects, exotic objects, etc; any kind of Records, including Property Descriptors, PrivateElements, etc; <emu-xref href="#sec-ecmascript-language-types-symbol-type">symbols</emu-xref>; Parse Nodes; Lists; <emu-xref href="#sec-set-and-relation-specification-type">Sets</emu-xref> and Relations; Abstract Closures; Data Blocks; Private Names; execution contexts and execution context stacks; agent signifiers; and WaiterLists.</p>
```

Nit: I'd remove the "immediately" since SABs do not guarantee immediate observability.

spec.html
```diff
@@ -1033,6 +1033,13 @@ <h1>Mathematical Operations</h1>
...
+    </emu-clause>
```
syg, line 1042: Maybe a 3rd paragraph that says "A value without identity is never equal to a value with identity."?

spec.html
syg: +1 to @bakkot, I think readability is hurt by this brevity.

syg (commented):
Shouldn't the lists of kinds of values that have/do not have identity be exhaustive?

syg:
@michaelficarra All of the editorial choices in https://github.com/tc39/ecma262/pull/2821#issue-1299519833 seem to be strictly speaking unrelated to nailing down a definition of identity. It'd be easier to review some of those as separate PRs.

spec.html
bathos: @syg I thought this text was “speaking at the same level” as text defining the memory model. If it is, would “immediately” still not be true? (not sure at all, mainly just wondering)

syg: @bathos I think "immediately" here is more easily interpreted as some kind of wall clock time immediate observability, and this property isn't important to defining identity here so I'd rather we drop the chance of confusion.

bakkot (approved):

syg (approved):
This is great. lgtm!

PR #2822, opened 2022-07-11
Editorial: change dfn for "mathematical value" and add "mathematical value of"

michaelficarra:
Originally part of #2821. Relates to #2786. This changes where the "(extended) mathematical value" `<dfn>` points and adds a new `<dfn>` for the operation "(extended) mathematical value of".


spec.html
```diff
@@ -1015,7 +1015,7 @@ <h1>Mathematical Operations</h1>
...
+      <p>Conversions between mathematical values and Numbers or BigInts are always explicit in this document. A conversion from a mathematical value or extended mathematical value _x_ to a Number is denoted as "the <dfn>Number value for</dfn> _x_" or <emu-eqn id="𝔽" aoid="𝔽">𝔽(_x_)</emu-eqn>, and is defined in <emu-xref href="#sec-ecmascript-language-types-number-type"></emu-xref>. A conversion from an integer _x_ to a BigInt is denoted as "the <dfn>BigInt value for</dfn> _x_" or <emu-eqn id="ℤ" aoid="ℤ">ℤ(_x_)</emu-eqn>. A conversion from a Number or BigInt _x_ to a mathematical value is denoted as "the <dfn>mathematical value of</dfn> _x_", or <emu-eqn id="ℝ" aoid="ℝ">ℝ(_x_)</emu-eqn>. The mathematical value of *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub> is the mathematical value 0. The mathematical value of non-finite values is not defined. The <dfn>extended mathematical value of</dfn> _x_ is the mathematical value of _x_ for finite values, and is +&infin; and -&infin; for *+&infin;*<sub>𝔽</sub> and *-&infin;*<sub>𝔽</sub> respectively; it is not defined for *NaN*.</p>
```
bakkot, line 1018: If you're going to `dfn` "Number value for" here, you should probably remove the `<dfn>Number value</dfn> for` under [The Number Type](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-ecmascript-language-types-number-type). Alternatively, you could remove the `dfn` from this paragraph and expand the one in that clause - after all, it's a bit weird to have the sentence with the `dfn` for X then say "X is defined in this other clause".

Also probably should give these ids, so links are useful. Just linking to the top of this clause is not as useful.

michaelficarra: @bakkot done

bakkot (approved):

PR #2824, opened 2022-07-13
Editorial: Remove incorrect setting of the AsyncEvaluation state to false

codehag:
This change removes an extraneous line that was included when this was originally merged into the spec, but did not reflect the spec that had consensus. 

The line was removed (along with others) in this pr: https://github.com/tc39/proposal-top-level-await/commit/c2375dbd6ecb0d5e9a415f6a812fc0974a2935b7#diff-181371b08d71216599b0acccbaabd03c306da6de142ea6275c2135810999805aL697

Additionally, this is only set on the fulfilled state -- and not the rejected, so this is a mistake anyway. This could be seen as editorial or normative -- in either case this is a bug.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -26584,7 +26584,6 @@
               1. Assert: _module_.[[Status]] is ~evaluating-async~.
               1. Assert: _module_.[[AsyncEvaluation]] is *true*.
               1. Assert: _module_.[[EvaluationError]] is ~empty~.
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Since the migration to [[AsyncEvaluationOrder]] we now explicitly set [[AsyncEvaluationOrder]] to DONE, and it now also happens in AsyncModuleExecutionRejected.

PR #2827, opened 2022-07-13
Editorial: `OrdinaryCreateFromConstructor`: correctly handle an omitted internalSlotsList

ljharb:
I noticed that when the list is omitted, it would be passing `undefined` to an AO that asserts that it receives a List.


bakkot (approved):
Strictly speaking it's meaningless as-is; AO parameters don't default to "undefined" - we should always be checking for the presence of an optional parameter before using it.

Anyway, fix looks good, thanks!

ljharb:
ah, fair enough. either way :-)

PR #2832, opened 2022-07-20
Meta: fix ipr-check; incorporate emeriti

ljharb:
This restores the IPR form check, and updates it to check every commit, and to also incorporate emeriti as valid contributors.


michaelficarra (approved):

.github/workflows/ipr.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on: [pull_request_target, push]
+
+jobs:
+  build:
+    name: 'check IPR form'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
```
bakkot, line 11:
```suggestion
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0
```

The default only checks out the last commit on the branch.

bakkot (approved):
Experimentally, the github API endpoint used here appears to max out at 250 commits. That's probably sufficient? But maybe warrants a comment explaining why this is fine.

This ~does need the `fetch-depth` change but otherwise~ LGTM, thanks for fixing.

Whoops, ignore the `fetch-depth` comment, that's of course not necessary because this is using the github API rather than the local checkout.

PR #2834, opened 2022-07-21
Editorial: Fix typo in CharacterValue

h2oche:
In [22.2.1.6 Static Semantics: CharacterValue](https://tc39.es/ecma262/#sec-patterns-static-semantics-character-value), it seems that `HexDigits` is typo and `Hex4Digits` is appropriate.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -34967,7 +34967,7 @@ THH:mm:ss.sss
           HexNonSurrogate :: Hex4Digits
         </emu-grammar>
         <emu-alg>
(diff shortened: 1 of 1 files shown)


jmdyck:
Looks like this goes back to PR #984, which (among other things) introduced the CharacterValue SDO to replace some regexp "evaluation" rules.

michaelficarra (approved):

bakkot (approved):

PR #2835, opened 2022-07-21
Editorial: Update ECMA-402 and -404 references

### LINKED ISSUE #2808: Obsolete normative reference to ECMA-402, ECMAScript 2015 Internationalization API Specification.
Clause 3 of both the Ecma hosted 2022 edition and the TC39 hosted 2023 draft of ECMA-262 list the 2015 edition of ECMA-402 as a normative reference.  Shouldn't that be updated to the current edition or have an "or later" added?

gibson042:
Aligns with https://tc39.es/ecma402/#normative-references

Fixes #2808

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -166,7 +166,7 @@
   </p>
   <p>
     ECMA-404, <i>The JSON Data Interchange Format</i>.<br>
-    <a href="https://ecma-international.org/publications/standards/Ecma-404.htm">https://ecma-international.org/publications/standards/Ecma-404.htm</a>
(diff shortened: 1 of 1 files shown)


michaelficarra:
Discussed this today in the 262 editors' call. We'd prefer to refer specifically to the associated 402 version instead of allowing future versions. This means that

1. For yearly releases, we add a feature to ecmarkup for both 262 and 402 to refer to the appropriate yearly release of the other
2. For draft releases, we change the name of the document to be next year with the word "draft" after the year, and refer to each other by that name

How does that sound?

gibson042:
Sure, that seems fine.

PR #2837, opened 2022-07-22
Editorial: add missing `<dfn>`s for different kinds of code

michaelficarra:
Extracted from #2820. Not sure why these were `<em>` and not `<dfn>` in the first place.


bakkot (approved):
I'd add an `id` for "ECMAScript code execution context" so the link goes to a more helpful place. Otherwise LGTM.

PR #2842, opened 2022-08-03
Editorial: GetValue and PutValue do not need to unwrap Completion Records

bakkot:
https://github.com/tc39/ecma262/pull/2744 removed almost every case where GetValue or PutValue was passed a completion record, instead explicitly unwrapping on the previous line or lines. This removes the final three cases and then updates the definitions of those algorithms so that they do not need to handle being passed Completion Records.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -22100,7 +22100,7 @@
             1. Set the running execution context's LexicalEnvironment to _newEnv_.
           1. Let _exprRef_ be Completion(Evaluation of _expr_).
           1. Set the running execution context's LexicalEnvironment to _oldEnv_.
-          1. Let _exprValue_ be ? GetValue(_exprRef_).
+          1. Let _exprValue_ be ? GetValue(? _exprRef_).
           1. If _iterationKind_ is ~enumerate~, then
             1. If _exprValue_ is *undefined* or *null*, then
               1. Return Completion Record { [[Type]]: ~break~, [[Value]]: ~empty~, [[Target]]: ~empty~ }.
@@ -24831,7 +24831,7 @@
           1. NOTE: The running execution context's PrivateEnvironment is _outerPrivateEnvironment_ when evaluating |ClassHeritage|.
           1. Let _superclassRef_ be Completion(Evaluation of |ClassHeritage|).
           1. Set the running execution context's LexicalEnvironment to _env_.
-          1. Let _superclass_ be ? GetValue(_superclassRef_).
+          1. Let _superclass_ be ? GetValue(? _superclassRef_).
           1. If _superclass_ is *null*, then
             1. Let _protoParent_ be *null*.
             1. Let _constructorParent_ be %Function.prototype%.
--- a/spec.html
+++ b/spec.html
@@ -4218,13 +4218,12 @@
       <emu-clause id="sec-getvalue" type="abstract operation">
         <h1>
           GetValue (
-            _V_: unknown,
+            _V_: a Reference Record or an ECMAScript language value,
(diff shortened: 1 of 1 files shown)


michaelficarra (approved):

spec.html
```diff
@@ -22096,17 +22100,8 @@ <h1>
...
-              1. If _lhsKind_ is either ~assignment~ or ~varBinding~, then
-                1. Assert: _lhsRef_ is a Completion Record.
-                1. If _lhsRef_ is an abrupt completion, then
-                  1. Let _status_ be _lhsRef_.
-                1. Else,
-                  1. Let _status_ be Completion(PutValue(_lhsRef_.[[Value]], _nextValue_)).
-              1. Else,
-                1. Assert: _lhsKind_ is ~lexicalBinding~.
                 1. Let _status_ be Completion(InitializeReferencedBinding(_lhsRef_, _nextValue_)).
-            1. Else,
```
jmdyck, line 22109: You could similarly distribute the `_destructuring_ is *true*` chunk over the preceding If/Else.

spec.html
```diff
@@ -4201,13 +4201,12 @@ <h1>
       <emu-clause id="sec-getvalue" type="abstract operation">
         <h1>
           GetValue (
-            _V_: unknown,
+            _V_: either a Reference Record or an ECMAScript language value,
```
jmdyck, line 4204: Maybe drop the "either". It looks like we use it in parameter-types only when there would otherwise be an ambiguity.

spec.html
```diff
@@ -4229,14 +4228,12 @@ <h1>
         <h1>
           PutValue (
             _V_: unknown,
```
jmdyck, line 4230: I think we can change `unknown` to `a Reference Record or an ECMAScript language value`.

jmdyck (commented):
`InitializeReferencedBinding` also calls `ReturnIfAbrupt` on its 2 parameters, so it might make sense to include it in this PR.

spec.html
bakkot: Done.

spec.html
bakkot: Done.

spec.html
bakkot: Done.

In fact I'm pretty sure static semantic rules will prevent calling this with an ES language value, but auditing that is tricky.

bakkot:
> `InitializeReferencedBinding` also calls `ReturnIfAbrupt` on its 2 parameters, so it might make sense to include it in this PR.

Done. In the process I marked initialization of `let x;` and `let x = 0;` as infallible (in https://github.com/tc39/ecma262/pull/2842/commits/73f2743ebb27e462bc5f922bf2aa34508e474fef), which I am _pretty_ sure is true.

michaelficarra (approved):

spec.html
```diff
@@ -22038,7 +22032,7 @@ <h1>
             1. Set the running execution context's LexicalEnvironment to _newEnv_.
           1. Let _exprRef_ be Completion(Evaluation of _expr_).
           1. Set the running execution context's LexicalEnvironment to _oldEnv_.
-          1. Let _exprValue_ be ? GetValue(_exprRef_).
+          1. Let _exprValue_ be ? GetValue(? _exprRef_).
```
syg, line 22035: Out of curiosity did ecmarkup find this or was this manually audited?

syg (approved):

spec.html
bakkot: Manually, alas. ecmarkup does not yet track types except for checking the completion-ness of callsites of AOs, which is purely local.

PR #2843, opened 2022-08-03
Editorial: move RegExpHasFlag out of the dotAll section

michaelficarra:


bakkot (approved):

michaelficarra:
Now moves into the flags getter section.

bakkot (approved):

jmdyck:
I don't think the proposed location is better than the current one. (Not that the current one is great.)

As I suggested in #2349, I think a better solution would be to create a new section (say 22.2.8 "Abstract Operations for Regular Expressions"), and move a bunch of AOs there:
- RegExpHasFlag
- all the AOs under 22.2.5.2 "RegExp.prototype.exec" (since all of them are used elsewhere than just that method)
- RegExpCreate + EscapeRegExpPattern (They're currently under "Abstract Operations for the RegExp Constructor" but they're not actually used by the RegExp Constructor.)
- or for that matter, everything under "Abstract Operations for the RegExp Constructor", since all of them are used elsewhere than just the constructor.

ljharb:
While I think that is a much better result than either main or this PR, I do think this PR is a strict improvement over main.

michaelficarra:
Still, we should probably merge this improvement in the interim.

PR #2844, opened 2022-08-04
Editorial: Revert unintentional normative change for String.prototype.substr

anba:
`"a".substr(0, Infinity)` should return `"a"`, but #2007 incorrectly changed the result to be `""`.

If we change `substr`'s algorithm to be more similar to `String.prototype.substring` and `String.prototype.slice`, we can easily fix this issue:

1. `min(intStart, size)` guarantees that `intStart` is now in `[0, size]`.
2. Clamping `intLength` guarantees it's also in `[0, size]`.

With these two changes, `intEnd = min(intStart + intLength, size)` is now in range `[intStart, size]`, so we no longer have to check for `intStart ≥ intEnd`, but instead can directly perform the `substring` operation.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -47505,10 +47505,10 @@ THH:mm:ss.sss
           1. Let _intStart_ be ? ToIntegerOrInfinity(_start_).
           1. If _intStart_ is -&infin;, set _intStart_ to 0.
           1. Else if _intStart_ &lt; 0, set _intStart_ to max(_size_ + _intStart_, 0).
+          1. Else, set _intStart_ to min(_intStart_, _size_).
           1. If _length_ is *undefined*, let _intLength_ be _size_; otherwise let _intLength_ be ? ToIntegerOrInfinity(_length_).
-          1. If _intStart_ is +&infin;, _intLength_ &le; 0, or _intLength_ is +&infin;, return the empty String.
+          1. Set _intLength_ to the result of clamping _intLength_ between 0 and _size_.
           1. Let _intEnd_ be min(_intStart_ + _intLength_, _size_).
-          1. If _intStart_ &ge; _intEnd_, return the empty String.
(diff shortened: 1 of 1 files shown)


bakkot (approved):

spec.html
```diff
@@ -47528,10 +47528,10 @@ <h1>String.prototype.substr ( _start_, _length_ )</h1>
           1. Let _intStart_ be ? ToIntegerOrInfinity(_start_).
           1. If _intStart_ is -&infin;, set _intStart_ to 0.
           1. Else if _intStart_ &lt; 0, set _intStart_ to max(_size_ + _intStart_, 0).
+          1. Else, set _intStart_ to min(_intStart_, _size_).
```
michaelficarra, line 47531: Can't we use clamping here, too?

```suggestion
          1. If _intStart_ &lt; 0, set _intStart_ to _size_ + _intStart_.
          1. Set _intStart_ to the result of clamping _intStart_ between 0 and _size_.
```

bakkot: That does arithmetic on an infinite value (-∞), which we've tried to avoid. If you restore the first line then the other changes are fine, though I don't think it's any clearer or more elegant, personally.

anba: And in addition to the `-∞` issue, the current PR matches how [String.prototype.slice](https://tc39.es/ecma262/#sec-string.prototype.slice) is specified. If we should decide to use clamping, we should also update `String.prototype.slice` accordingly.

michaelficarra: I still think the clamping is more clear, but @bakkot is right, we should leave the guard in for infinity. We can update `String.prototype.slice` in a separate PR if you want.

bakkot: @michaelficarra If you approve this as-is I will submit a following changing both; no need to drag this PR out further.

michaelficarra (approved):

syg (approved):

Issue #2181, opened 2020-09-18
use code points and code units without converting from real numbers


michaelficarra:
In many places, we use the phrase "the code point value of" a mathematical value. I propose we just use the mathematical value itself and not treat it as if some kind of conversion is happening.

michaelficarra:
@jmdyck I see you recently created a commit that goes in the opposite direction: [`c9b32a5` (#2018)](https://github.com/tc39/ecma262/pull/2018/commits/c9b32a56bdd838215d5fdafdfeac823bef5a038d). Do you have an opinion on this topic?

jmdyck:
Code points, code units, and mathematical integers are different in terms of value space, associated semantics, and appropriate use. E.g., it doesn't make sense to append a code point to a String, even if its numeric value is under 2^16. 

So I think it's useful for the spec to treat these as different kinds of things, and use specific phrasings to 'convert' between them. (And, to a large extent, that's what the spec already does.) By doing so, we can increase our confidence that algorithms are written correctly, or detect when they're written incorrectly.

It's roughly the same reason why it's useful for the spec to treat Number, BigInt, and mathematical reals as different kinds of things, and use explicit forms to convert between them.

michaelficarra:
Discussed in the editor call. We will explicitly convert between integers, code units, and code points.

PR #2848, opened 2022-08-06
Editorial: consistency in phrasing of ranges and inclusivity/exclusivity

michaelficarra:
I chose to use "within the inclusive range from A to B" or use chained comparisons where possible. For ranges that have an open end, I name each side as either exclusive or inclusive.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1027,6 +1027,14 @@
       <emu-note>
         <p><emu-eqn>floor(_x_) = _x_ - (_x_ modulo 1)</emu-eqn>.</p>
       </emu-note>
+      <p>An <dfn id="interval">interval</dfn> from lower bound _a_ to upper bound _b_ is a possibly-infinite, possibly-empty set of numeric values of the same numeric type. Each bound will be described as either inclusive or exclusive, but not both. There are four kinds of intervals, as follows:</p>
+      <ul>
+        <li>An interval from _a_ (inclusive) to _b_ (inclusive), also called an <dfn id="inclusive-interval">inclusive interval</dfn> from _a_ to _b_, includes all values _x_ of the same numeric type such that _a_ &le; _x_ &le; _b_, and no others.</li>
+        <li>An interval from _a_ (inclusive) to _b_ (exclusive) includes all values _x_ of the same numeric type such that _a_ &le; _x_ &lt; _b_, and no others.</li>
+        <li>An interval from _a_ (exclusive) to _b_ (inclusive) includes all values _x_ of the same numeric type such that _a_ &lt; _x_ &le; _b_, and no others.</li>
+        <li>An interval from _a_ (exclusive) to _b_ (exclusive) includes all values _x_ of the same numeric type such that _a_ &lt; _x_ &lt; _b_, and no others.</li>
+      </ul>
+      <p>For example, the interval from 1 (inclusive) to 2 (exclusive) consists of all mathematical values between 1 and 2, including 1 and not including 2. For the purpose of defining intervals, *-0*<sub>𝔽</sub> &lt; <!-- this comment here to avoid ecmarkup complaining about comparison to 0 --> *+0*<sub>𝔽</sub>, so, for example, an inclusive interval with a lower bound of *+0*<sub>𝔽</sub> includes *+0*<sub>𝔽</sub> but not *-0*<sub>𝔽</sub>. *NaN* is never included in an interval.</p>
     </emu-clause>
 
     <emu-clause id="sec-value-notation">
@@ -1071,7 +1079,7 @@
     <emu-clause id="sec-ecmascript-language-types-string-type">
       <h1>The String Type</h1>
       <p>The <dfn variants="is a String,is not a String">String type</dfn> is the set of all ordered sequences of zero or more 16-bit unsigned integer values (&ldquo;elements&rdquo;) up to a maximum length of 2<sup>53</sup> - 1 elements. The String type is generally used to represent textual data in a running ECMAScript program, in which case each element in the String is treated as a UTF-16 code unit value. Each element is regarded as occupying a position within the sequence. These positions are indexed with non-negative integers. The first element (if any) is at index 0, the next element (if any) at index 1, and so on. The length of a String is the number of elements (i.e., 16-bit values) within it. The empty String has length zero and therefore contains no elements.</p>
-      <p>ECMAScript operations that do not interpret String contents apply no further semantics. Operations that do interpret String values treat each element as a single UTF-16 code unit. However, ECMAScript does not restrict the value of or relationships between these code units, so operations that further interpret String contents as sequences of Unicode code points encoded in UTF-16 must account for ill-formed subsequences. Such operations apply special treatment to every code unit with a numeric value in the inclusive range 0xD800 to 0xDBFF (defined by the Unicode Standard as a <dfn id="leading-surrogate" variants="leading surrogates">leading surrogate</dfn>, or more formally as a <dfn id="high-surrogate-code-unit" variants="high-surrogate code units">high-surrogate code unit</dfn>) and every code unit with a numeric value in the inclusive range 0xDC00 to 0xDFFF (defined as a <dfn id="trailing-surrogate" variants="trailing surrogates">trailing surrogate</dfn>, or more formally as a <dfn id="low-surrogate-code-unit" variants="low-surrogate code units">low-surrogate code unit</dfn>) using the following rules:</p>
+      <p>ECMAScript operations that do not interpret String contents apply no further semantics. Operations that do interpret String values treat each element as a single UTF-16 code unit. However, ECMAScript does not restrict the value of or relationships between these code units, so operations that further interpret String contents as sequences of Unicode code points encoded in UTF-16 must account for ill-formed subsequences. Such operations apply special treatment to every code unit with a numeric value in the inclusive interval from 0xD800 to 0xDBFF (defined by the Unicode Standard as a <dfn id="leading-surrogate" variants="leading surrogates">leading surrogate</dfn>, or more formally as a <dfn id="high-surrogate-code-unit" variants="high-surrogate code units">high-surrogate code unit</dfn>) and every code unit with a numeric value in the inclusive interval from 0xDC00 to 0xDFFF (defined as a <dfn id="trailing-surrogate" variants="trailing surrogates">trailing surrogate</dfn>, or more formally as a <dfn id="low-surrogate-code-unit" variants="low-surrogate code units">low-surrogate code unit</dfn>) using the following rules:</p>
       <ul>
         <li>
           A code unit that is not a <emu-xref href="#leading-surrogate"></emu-xref> and not a <emu-xref href="#trailing-surrogate"></emu-xref> is interpreted as a code point with the same value.
@@ -1109,7 +1117,7 @@
           1. Let _len_ be the length of _string_.
           1. If _searchValue_ is the empty String and _fromIndex_ &le; _len_, return _fromIndex_.
           1. Let _searchLen_ be the length of _searchValue_.
-          1. For each integer _i_ starting with _fromIndex_ such that _i_ &le; _len_ - _searchLen_, in ascending order, do
+          1. For each integer _i_ such that _fromIndex_ &le; _i_ &le; _len_ - _searchLen_, in ascending order, do
             1. Let _candidate_ be the substring of _string_ from _i_ to _i_ + _searchLen_.
             1. If _candidate_ is the same sequence of code units as _searchValue_, return _i_.
           1. Return -1.
@@ -1765,17 +1773,17 @@
         <div class="math-display">
           _s_ &times; _m_ &times; 2<sup>_e_</sup>
         </div>
-        <p>where _s_ is 1 or -1, _m_ is an integer such that 2<sup>52</sup> &le; _m_ &lt; 2<sup>53</sup>, and _e_ is an integer such that -1074 &le; _e_ &le; 971.</p>
+        <p>where _s_ is 1 or -1, _m_ is an integer in the interval from 2<sup>52</sup> (inclusive) to 2<sup>53</sup> (exclusive), and _e_ is an integer in the inclusive interval from -1074 to 971.</p>
         <p>The remaining 9,007,199,254,740,990 (that is, <emu-eqn>2<sup>53</sup> - 2</emu-eqn>) values are denormalized, having the form</p>
         <div class="math-display">
           _s_ &times; _m_ &times; 2<sup>_e_</sup>
         </div>
-        <p>where _s_ is 1 or -1, _m_ is an integer such that 0 &lt; _m_ &lt; 2<sup>52</sup>, and _e_ is -1074.</p>
+        <p>where _s_ is 1 or -1, _m_ is an integer in the interval from 0 (exclusive) to 2<sup>52</sup> (exclusive), and _e_ is -1074.</p>
         <p>Note that all the positive and negative integers whose magnitude is no greater than 2<sup>53</sup> are representable in the Number type. The integer 0 has two representations in the Number type: *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub>.</p>
         <p>A finite number has an <em>odd significand</em> if it is non-zero and the integer _m_ used to express it (in one of the two forms shown above) is odd. Otherwise, it has an <em>even significand</em>.</p>
         <p>In this specification, the phrase &ldquo;the <dfn id="number-value-for" oldids="number-value">Number value for</dfn> _x_&rdquo; where _x_ represents an exact real mathematical quantity (which might even be an irrational number such as &pi;) means a Number value chosen in the following manner. Consider the set of all finite values of the Number type, with *-0*<sub>𝔽</sub> removed and with two additional values added to it that are not representable in the Number type, namely 2<sup>1024</sup> (which is <emu-eqn>+1 &times; 2<sup>53</sup> &times; 2<sup>971</sup></emu-eqn>) and <emu-eqn>-2<sup>1024</sup></emu-eqn> (which is <emu-eqn>-1 &times; 2<sup>53</sup> &times; 2<sup>971</sup></emu-eqn>). Choose the member of this set that is closest in value to _x_. If two values of the set are equally close, then the one with an even significand is chosen; for this purpose, the two extra values 2<sup>1024</sup> and <emu-eqn>-2<sup>1024</sup></emu-eqn> are considered to have even significands. Finally, if 2<sup>1024</sup> was chosen, replace it with *+&infin;*<sub>𝔽</sub>; if <emu-eqn>-2<sup>1024</sup></emu-eqn> was chosen, replace it with *-&infin;*<sub>𝔽</sub>; if *+0*<sub>𝔽</sub> was chosen, replace it with *-0*<sub>𝔽</sub> if and only if _x_ &lt; 0; any other chosen value is used unchanged. The result is the Number value for _x_. (This procedure corresponds exactly to the behaviour of the IEEE 754-2019 roundTiesToEven mode.)</p>
         <p>The Number value for +&infin; is *+&infin;*<sub>𝔽</sub>, and the Number value for -&infin; is *-&infin;*<sub>𝔽</sub>.</p>
-        <p>Some ECMAScript operators deal only with integers in specific ranges such as <emu-eqn>-2<sup>31</sup></emu-eqn> through <emu-eqn>2<sup>31</sup> - 1</emu-eqn>, inclusive, or in the range 0 through <emu-eqn>2<sup>16</sup> - 1</emu-eqn>, inclusive. These operators accept any value of the Number type but first convert each such value to an integer value in the expected range. See the descriptions of the numeric conversion operations in <emu-xref href="#sec-type-conversion"></emu-xref>.</p>
+        <p>Some ECMAScript operators deal only with integers in specific ranges such as the inclusive interval from <emu-eqn>-2<sup>31</sup></emu-eqn> to <emu-eqn>2<sup>31</sup> - 1</emu-eqn> or the inclusive interval from 0 to <emu-eqn>2<sup>16</sup> - 1</emu-eqn>. These operators accept any value of the Number type but first convert each such value to an integer value in the expected range. See the descriptions of the numeric conversion operations in <emu-xref href="#sec-type-conversion"></emu-xref>.</p>
 
         <emu-clause id="sec-numeric-types-number-unaryMinus" type="numeric method">
           <h1>
@@ -4472,7 +4480,7 @@
 
     <emu-clause id="sec-data-blocks">
       <h1>Data Blocks</h1>
-      <p>The <dfn variants="Data Blocks">Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A <dfn variants="byte values">byte value</dfn> is an integer value in the range 0 through 255, inclusive. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
+      <p>The <dfn variants="Data Blocks">Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A <dfn variants="byte values">byte value</dfn> is an integer in the inclusive interval from 0 to 255. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
       <p>For notational convenience within this specification, an array-like syntax can be used to access the individual bytes of a Data Block value. This notation presents a Data Block value as a 0-origined integer-indexed sequence of bytes. For example, if _db_ is a 5 byte Data Block value then _db_[2] can be used to access its 3<sup>rd</sup> byte.</p>
       <p>A data block that resides in memory that can be referenced from multiple agents concurrently is designated a <dfn variants="Shared Data Blocks">Shared Data Block</dfn>. A Shared Data Block has an identity (for the purposes of equality testing Shared Data Block values) that is <em>address-free</em>: it is tied not to the virtual addresses the block is mapped to in any process, but to the set of locations in memory that the block represents. Two data blocks are equal only if the sets of the locations they contain are equal; otherwise, they are not equal and the intersection of the sets of locations they contain is empty. Finally, Shared Data Blocks can be distinguished from Data Blocks.</p>
       <p>The semantics of Shared Data Blocks is defined using Shared Data Block events by the memory model. Abstract operations below introduce Shared Data Block events and act as the interface between evaluation semantics and the event semantics of the memory model. The events form a candidate execution, on which the memory model acts as a filter. Please consult the memory model for full semantics.</p>
@@ -5178,7 +5186,7 @@
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It converts _argument_ to one of 2<sup>32</sup> integral Number values in the range 𝔽(<emu-eqn>-2<sup>31</sup></emu-eqn>) through 𝔽(<emu-eqn>2<sup>31</sup> - 1</emu-eqn>), inclusive.</dd>
+        <dd>It converts _argument_ to one of 2<sup>32</sup> integral Number values in the inclusive interval from 𝔽(<emu-eqn>-2<sup>31</sup></emu-eqn>) to 𝔽(<emu-eqn>2<sup>31</sup> - 1</emu-eqn>).</dd>
       </dl>
       <emu-alg>
         1. Let _number_ be ? ToNumber(_argument_).
@@ -5211,7 +5219,7 @@
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It converts _argument_ to one of 2<sup>32</sup> integral Number values in the range *+0*<sub>𝔽</sub> through 𝔽(<emu-eqn>2<sup>32</sup> - 1</emu-eqn>), inclusive.</dd>
+        <dd>It converts _argument_ to one of 2<sup>32</sup> integral Number values in the inclusive interval from *+0*<sub>𝔽</sub> to 𝔽(<emu-eqn>2<sup>32</sup> - 1</emu-eqn>).</dd>
       </dl>
       <emu-alg>
         1. Let _number_ be ? ToNumber(_argument_).
@@ -5247,7 +5255,7 @@
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It converts _argument_ to one of 2<sup>16</sup> integral Number values in the range 𝔽(<emu-eqn>-2<sup>15</sup></emu-eqn>) through 𝔽(<emu-eqn>2<sup>15</sup> - 1</emu-eqn>), inclusive.</dd>
+        <dd>It converts _argument_ to one of 2<sup>16</sup> integral Number values in the inclusive interval from 𝔽(<emu-eqn>-2<sup>15</sup></emu-eqn>) to 𝔽(<emu-eqn>2<sup>15</sup> - 1</emu-eqn>).</dd>
       </dl>
       <emu-alg>
         1. Let _number_ be ? ToNumber(_argument_).
@@ -5266,7 +5274,7 @@
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It converts _argument_ to one of 2<sup>16</sup> integral Number values in the range *+0*<sub>𝔽</sub> through 𝔽(<emu-eqn>2<sup>16</sup> - 1</emu-eqn>), inclusive.</dd>
+        <dd>It converts _argument_ to one of 2<sup>16</sup> integral Number values in the inclusive interval from *+0*<sub>𝔽</sub> to 𝔽(<emu-eqn>2<sup>16</sup> - 1</emu-eqn>).</dd>
       </dl>
       <emu-alg>
         1. Let _number_ be ? ToNumber(_argument_).
@@ -5296,7 +5304,7 @@
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It converts _argument_ to one of 2<sup>8</sup> integral Number values in the range *-128*<sub>𝔽</sub> through *127*<sub>𝔽</sub>, inclusive.</dd>
+        <dd>It converts _argument_ to one of 2<sup>8</sup> integral Number values in the inclusive interval from *-128*<sub>𝔽</sub> to *127*<sub>𝔽</sub>.</dd>
       </dl>
       <emu-alg>
         1. Let _number_ be ? ToNumber(_argument_).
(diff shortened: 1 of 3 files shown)


spec.html
```diff
@@ -4482,7 +4482,7 @@ <h1>The Abstract Closure Specification Type</h1>
 
     <emu-clause id="sec-data-blocks">
       <h1>Data Blocks</h1>
-      <p>The <dfn variants="Data Blocks">Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A <dfn variants="byte values">byte value</dfn> is an integer value in the range 0 through 255, inclusive. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
+      <p>The <dfn variants="Data Blocks">Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A <dfn variants="byte values">byte value</dfn> is an integer value _b_ in the range 0 &lte; _b_ &lte; 255. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
```
ljharb, line 4485: why not 
```suggestion
      <p>The <dfn variants="Data Blocks">Data Block</dfn> specification type is used to describe a distinct and mutable sequence of byte-sized (8 bit) numeric values. A <dfn variants="byte values">byte value</dfn> is an integer value _b_ in the inclusive range 0 to 255. A Data Block value is created with a fixed number of bytes that each have the initial value 0.</p>
```
?

spec.html
```diff
@@ -35704,7 +35704,7 @@ <h1>
               1. Let _InputLength_ be the number of elements in _Input_.
               1. If _f_ &lt; 0 or _f_ &gt; _InputLength_, return ~failure~.
               1. Let _g_ be min(_e_, _f_).
-              1. If there exists an integer _i_ between 0 (inclusive) and _len_ (exclusive) such that Canonicalize(_rer_, _Input_[_rs_ + _i_]) is not the same character value as Canonicalize(_rer_, _Input_[_g_ + _i_]), return ~failure~.
+              1. If there exists an integer _i_ in the range 0 &lte; _i_ &lt; _len_ such that Canonicalize(_rer_, _Input_[_rs_ + _i_]) is not the same character value as Canonicalize(_rer_, _Input_[_g_ + _i_]), return ~failure~.
```
ljharb, line 35707:
```suggestion
              1. If there exists an integer _i_ in the inclusive range from 0 to _len_ such that Canonicalize(_rer_, _Input_[_rs_ + _i_]) is not the same character value as Canonicalize(_rer_, _Input_[_g_ + _i_]), return ~failure~.
```

spec.html
```diff
@@ -46482,7 +46482,7 @@ <h1>Valid Chosen Reads</h1>
           1. Let _readLen_ be the number of elements of _readValue_.
           1. If _chosenLen_ &ne; _readLen_, then
             1. Return *false*.
-          1. If _chosenValue_[_i_] &ne; _readValue_[_i_] for any integer value _i_ in the range 0 through _chosenLen_, exclusive, then
+          1. If _chosenValue_[_i_] &ne; _readValue_[_i_] for some integer _i_ in the range 0 &lte; _i_ &lt; _chosenLen_, then
```
ljharb, line 46485:
```suggestion
          1. If _chosenValue_[_i_] &ne; _readValue_[_i_] for some integer _i_ in the inclusive range from 0 to _chosenLen_, then
```

ljharb (commented):
This is a great improvement. Another option tho would be to use `[a, b)`-style notation throughout - that way it'd be unambiguous which sides were inclusive or exclusive, and all permutations could be represented. Any reason not to prefer that?

spec.html
michaelficarra: this range is not inclusive on the right

spec.html
michaelficarra: this range is not inclusive on the right

spec.html
ljharb: then can it use the parens form you’ve used elsewhere? (I’m sure in the rendering the difference would be easier to see than here0

michaelficarra: done

michaelficarra:
Thoughts on using the term "interval" instead of "range"? What about using (and explaining) the standard interval notation from maths? Any considerations for using this form for both floats and mathematical values?

ljharb:
I think defining and using interval notation would be much less ambiguous than prose.

michaelficarra:
Updated with far more rigour and renamed ranges to intervals.

spec.html
```diff
@@ -1024,6 +1024,7 @@ <h1>Mathematical Operations</h1>
       <emu-note>
         <p><emu-eqn>floor(_x_) = _x_ - (_x_ modulo 1)</emu-eqn>.</p>
       </emu-note>
+      <p>An <dfn>interval</dfn> from lower bound _a_ to upper bound _b_ is a possibly-infinite, possibly-empty set of numeric values of the same numeric type. A numeric value _x_ is included in the interval if it satisfies the criteria _a_ &lt; _x_ and _x_ &lt; _b_. If the lower bound is described as inclusive, _a_ &lte; _x_ is used instead of _a_ &lt; _x_. If the upper bound is described as inclusive, _x_ &lte; _b_ is used instead of _x_ &lt; _b_. An <dfn>inclusive interval</dfn> is an interval where both bounds are inclusive. *NaN* is never included in an interval. An interval with an exclusive bound of *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub> includes neither *+0*<sub>𝔽</sub> nor *-0*<sub>𝔽</sub>. An interval with an inclusive bound of *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub> includes both *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub>.</p>
```
bakkot, line 1027: We never actually use "interval" without specifying inclusive/exclusive for both ends (usually by the now-defined term "inclusive interval"); as such, I'd prefer we avoid defining what "interval" means when used without specifying the ends.

For example:

```diff
- A numeric value _x_ is included in the interval if it satisfies the criteria _a_ &lt; _x_ and _x_ &lt; _b_. If the lower bound is described as inclusive, _a_ &lte; _x_ is used instead of _a_ &lt; _x_. If the upper bound is described as inclusive, _x_ &lte; _b_ is used instead of _x_ &lt; _b_.
+ A numeric value _x_ is included in the interval if it satisfies both the lower bound &le; _x_ / a &lt; _x_ (for inclusive/exclusive lower bounds, respectively) and the upper bound _x_ &le; _b_ / x &lt; _b_ (for inclusive/exclusive upper bounds, respectively).
```

spec.html
```diff
@@ -1024,6 +1024,7 @@ <h1>Mathematical Operations</h1>
       <emu-note>
         <p><emu-eqn>floor(_x_) = _x_ - (_x_ modulo 1)</emu-eqn>.</p>
       </emu-note>
+      <p>An <dfn id="interval">interval</dfn> from lower bound _a_ to upper bound _b_ is a possibly-infinite, possibly-empty set of numeric values of the same numeric type. A numeric value _x_ is included in the interval if it satisfies the criteria _a_ &lt; _x_ and _x_ &lt; _b_. If the lower bound is described as inclusive, _a_ &le; _x_ is used instead of _a_ &lt; _x_. If the upper bound is described as inclusive, _x_ &le; _b_ is used instead of _x_ &lt; _b_. An <dfn id="inclusive-interval">inclusive interval</dfn> is an interval where both bounds are inclusive. *NaN* is never included in an interval. An interval with an exclusive bound of *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub> includes neither *+0*<sub>𝔽</sub> nor *-0*<sub>𝔽</sub>. An interval with an inclusive bound of *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub> includes both *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub>.</p>
```
bakkot, line 1027: Also

> `An interval with an inclusive bound of *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub> includes both *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub>.`

This is not consistent with how we actually use the terms. `ToUint16`, for example, never produces -0.

spec.html
michaelficarra: Eh, I still like my formulation better. "interval" is still useful in practice to link to all half-open intervals, even if there are no fully open intervals. Also, apparently unqualified "interval" implies "open interval" to mathematicians and I'd prefer to respect that intuition if there's relatively little downside.

spec.html
michaelficarra: It is true that `ToUint16` does not produce `-0` but the previous phrasing didn't guarantee that either. We can add an additional "this operation never produces `-0` or something to those sections.

michaelficarra: Alternatively, we could significantly complicate things and "order" the zeroes using the rules

* an inclusive lower bound of `-0` includes both zeroes
* an inclusive lower bound of `+0` includes `+0` but not `-0`
* an exclusive lower bound of `-0` includes `+0` but not `-0`
* an exclusive lower bound of `+0` includes neither zero
* an inclusive upper bound of `-0` includes `-0` but not `+0`
* an inclusive upper bound of `+0` includes both zeroes
* an exclusive upper bound of `-0` includes neither zero
* an exclusive upper bound of `+0` includes `-0` but not `+0`

bakkot: Or just say "for the purposes of defining intervals, -0 is considered to be strictly less than +0".

spec.html
bakkot: > Also, apparently unqualified "interval" implies "open interval" to mathematicians

No it doesn't. Different mathematicians use different customs (in analysis, for example, if you encounter an unqualified "interval" it is almost certainly meant to be closed), and most people will specify open or closed when using the term. We should always specify and should foreclose the possibility of using the term without specifying.

ljharb:
@michaelficarra was the decision not to use `(a, b]` notation?

michaelficarra:
@ljharb We did not discuss that today in the editor call, but both @bakkot and I prefer this form to the interval notation, so I'm going to stick with it.

spec.html
michaelficarra: Alright, reworded. Let me know what you think.

spec.html
```diff
@@ -1024,6 +1024,7 @@ <h1>Mathematical Operations</h1>
       <emu-note>
         <p><emu-eqn>floor(_x_) = _x_ - (_x_ modulo 1)</emu-eqn>.</p>
       </emu-note>
+      <p>An <dfn id="interval">interval</dfn> from lower bound _a_ to upper bound _b_ is a possibly-infinite, possibly-empty set of numeric values of the same numeric type. A numeric value _x_ is included in the interval if it satisfies both its <dfn>lower bound criteria</dfn> and its <dfn>upper bound criteria</dfn>. If an interval's lower bound is described as inclusive, its lower bound criteria is _a_ &le; _x_; otherwise it is _a_ &lt; _x_. If an interval's upper bound is described as inclusive, its upper bound criteria is _x_ &le; _b_; otherwise it is _x_ &lt; _b_. An <dfn id="inclusive-interval">inclusive interval</dfn> is an interval where both bounds are inclusive. It is an editorial error if a lower bound or upper bound is not described as either inclusive or exclusive. For the purpose of defining intervals, *-0*<sub>𝔽</sub> &lt; <!-- this comment here to avoid ecmarkup complaining about comparison to 0 --> *+0*<sub>𝔽</sub>. *NaN* is never included in an interval.</p>
```
bakkot, line 1027: Re: wording, I find this harder to follow than my original suggestion (mostly because it's much more verbose). I won't die on this hill, but I would prefer my shorter, IMO clearer suggestion.

If we do keep this wording, it would benefit from an example or two - e.g. "For example, 'the real interval from 1 (inclusive) to 5 (exclusive)' consists of all real numbers between 1 and 5, including 1 and not including 5."

---

I would `em`, not `dfn`, the "criteria" terms. They're not used outside this paragraph.

Also the singular of "criteria" is "criterion".

michaelficarra: Added two examples and changed the wording for you. I hate singular "criterion", so I changed it to "condition".

bakkot: This works for me.

spec.html
```diff
@@ -3901,7 +3901,7 @@ <h1>Well-Known Intrinsic Objects</h1>
     <h1>ECMAScript Specification Types</h1>
     <p>A specification type corresponds to meta-values that are used within algorithms to describe the semantics of ECMAScript language constructs and ECMAScript language types. The specification types include Reference, List, Completion Record, Property Descriptor, Environment Record, Abstract Closure, and Data Block. Specification type values are specification artefacts that do not necessarily correspond to any specific entity within an ECMAScript implementation. Specification type values may be used to describe intermediate results of ECMAScript expression evaluation but such values cannot be stored as properties of objects or values of ECMAScript language variables.</p>
 
-    <emu-clause id="sec-list-and-record-specification-type">
+    <emu-clause id="sec-list-and-record-specification-type"
```
ljharb, line 3904:
```suggestion
    <emu-clause id="sec-list-and-record-specification-type">
```

michaelficarra: Yeah that's already been fixed. See https://github.com/tc39/ecma262/pull/2848/commits/81129217591513ed4c4b6dbd0274e356249bafa0.

spec.html
```diff
@@ -1024,6 +1024,7 @@ <h1>Mathematical Operations</h1>
...
..._a_ to upper bound _b_ is a possibly-infinite, possibly-empty set of numeric values of the same numeric type. A numeric value _x_ is included in the interval if it satisfies both its <em>lower bound condition</em> and its <em>upper bound condition</em>. If an interval's lower bound is described as inclusive, its lower bound condition is _a_ &le; _x_; otherwise it is _a_ &lt; _x_. If an interval's upper bound is described as inclusive, its upper bound condition is _x_ &le; _b_; otherwise it is _x_ &lt; _b_. For example, the interval from 1 (inclusive) to 2 (exclusive) consists of all mathematical values between 1 and 2, including 1 and not including 2. An <dfn id="inclusive-interval">inclusive interval</dfn> is an interval where both bounds are inclusive. It is an editorial error if any bound of an interval is not described as either inclusive or exclusive. For the purpose of defining intervals, *-0*<sub>𝔽</sub> &lt; <!-- this comment here to avoid ecmarkup complaining about comparison to 0 --> *+0*<sub>𝔽</sub>, so, for example, an inclusive interval with a lower bound of *+0*<sub>𝔽</sub> includes *+0*<sub>𝔽</sub> but not *-0*<sub>𝔽</sub>. *NaN* is never included in an interval.</p>
```
bakkot, line 1027: > `If an interval's lower bound is described as inclusive, its lower bound condition is _a_ &le; _x_; otherwise [...]`

I would replace "otherwise" with "if it is described as exclusive". Seems weird not to mention "exclusive" in this part of the definition.

spec.html
```diff
@@ -1094,7 +1095,7 @@ <h1>
           1. Let _len_ be the length of _string_.
           1. If _searchValue_ is the empty String and _fromIndex_ &le; _len_, return _fromIndex_.
           1. Let _searchLen_ be the length of _searchValue_.
-          1. For each integer _i_ starting with _fromIndex_ such that _i_ &le; _len_ - _searchLen_, in ascending order, do
+          1. For each integer _i_ in the inclusive interval from _fromIndex_ to _len_ - _searchLen_, in ascending order, do
```
bakkot, line 1098: This doesn't seem like an improvement but I guess I'm fine with it.

spec.html
```diff
@@ -1760,17 +1761,17 @@ <h1>The Number Type</h1>
         <div class="math-display">
           _s_ &times; _m_ &times; 2<sup>_e_</sup>
         </div>
-        <p>where _s_ is 1 or -1, _m_ is an integer such that 2<sup>52</sup> &le; _m_ &lt; 2<sup>53</sup>, and _e_ is an integer such that -1074 &le; _e_ &le; 971.</p>
+        <p>where _s_ is 1 or -1, _m_ is an integer in the interval from 2<sup>52</sup> (inclusive) to 2<sup>53</sup> (exclusive), and _e_ is an integer in the inclusive interval from -1074 to 971.</p>
```
bakkot, line 1764: This one really doesn't seem like an improvement.

spec.html
```diff
@@ -35302,7 +35303,7 @@ <h1>
               1. If _max_ is +&infin;, let _max2_ be +&infin;; otherwise let _max2_ be _max_ - 1.
               1. Return RepeatMatcher(_m_, _min2_, _max2_, _greedy_, _y_, _c_, _parenIndex_, _parenCount_).
             1. Let _cap_ be a copy of _x_'s _captures_ List.
-            1. [id="step-repeatmatcher-clear-captures"] For each integer _k_ such that _parenIndex_ &lt; _k_ and _k_ &le; _parenIndex_ + _parenCount_, set _cap_[_k_] to *undefined*.
+            1. [id="step-repeatmatcher-clear-captures"] For each integer _k_ in the interval from _parenIndex_ (exclusive) to _parenIndex_ + _parenCount_ (inclusive), set _cap_[_k_] to *undefined*.
```
bakkot, line 35306: Ugh, gross. I realize we've defined things in a way which lets you write this but half-open intervals where the closed half is the _upper_ bound are unnatural. If we must use an explicit interval here, let it be `the inclusive interval from _parenIndex_ + 1 to _parenIndex_ + _parenCount_`.

spec.html
```diff
@@ -35833,7 +35834,7 @@ <h1>
           1. Return the union of _D_ and _C_.
         </emu-alg>
         <emu-note>
-          <p>|ClassRanges| can expand into a single |ClassAtom| and/or ranges of two |ClassAtom| separated by dashes. In the latter case the |ClassRanges| includes all characters between the first |ClassAtom| and the second |ClassAtom|, inclusive; an error occurs if either |ClassAtom| does not represent a single character (for example, if one is \w) or if the first |ClassAtom|'s character value is greater than the second |ClassAtom|'s character value.</p>
+          <p>|ClassRanges| can expand into a single |ClassAtom| and/or ranges of two |ClassAtom| separated by dashes. In the latter case the |ClassRanges| includes all characters in the inclusive interval from the first |ClassAtom| to the second |ClassAtom|; an error occurs if either |ClassAtom| does not represent a single character (for example, if one is \w) or if the first |ClassAtom|'s character value is greater than the second |ClassAtom|'s character value.</p>
```
bakkot, line 35837: Feels a bit iffy given that we defined intervals to be over numeric types, though I suppose in this part of the spec we did say "character" means "code unit" or "code point" depending on `u` mode.

jmdyck: The decision on issue #2181 suggests (to me, at least) that 'code unit' and 'code point' (and thus 'character') should not be treated as numeric types.

spec.html
michaelficarra: It's just the same interval that was already being described. But if you want it changed, I'll change it.

bakkot: The problem isn't the set of numbers being described, it's the way of describing it.

spec.html
michaelficarra: I'd prefer to be consistent in using intervals with these "for each" loops unless there's good reason not to.

spec.html
michaelficarra: You mean the notation, right? Because they're basically the same to me, each is a different way to describe this particular interval.

spec.html
michaelficarra: Yes this is not a good way to describe that set of characters, and uses implicit conversion that we've decided against in #2181. However, I don't think this change makes a difference with that existing problem. I can drop the change until we address #2181, but we'll probably want to use an interval here at that point anyway, so I would keep this change.

spec.html
bakkot: Yes, "way of describing" means the notation used, and "this particular interval" means the set of numbers. My problem is not with the set of numbers being described, it's the way of describing it.

spec.html
bakkot: I think this change makes the problem a bit worse, because the existing notation is not explicitly defined, and the new notation is. Using an imprecisely defined term is better than using a precisely defined term in a place it doesn't apply.

spec.html
michaelficarra: How is this? https://github.com/tc39/ecma262/pull/2848/commits/19bd4f621f27b83e73efd92e5598c0ea45a9bb06

bakkot: I still would much prefer the wording I suggested. As mentioned previously, I very specifically do not want to give a definition in the case that one of the bounds is not described as being either inclusive or exclusive.

spec.html
michaelficarra: "way of describing" could mean either the bounds used or the notation (prose vs comparisons). I didn't understand because you were asking for a change in the bounds but the change I've made in this PR is only to the notation. So I guess those bounds are acceptable to you using the previous notation but not the new notation? To me, changing the bounds seems like an unrelated change that could warrant its own small PR.

spec.html
michaelficarra: @bakkot I think you've misread it. Can you look again?

bakkot: > If an interval's lower bound is described as inclusive, its lower bound condition is _a_ &le; _x_; otherwise, it's lower bound is exclusive and its lower bound condition is _a_ &lt; _x_.

This says to me that if the bound is not described as being inclusive then it is assumed to be exclusive. That is the specific thing I want to avoid doing.

ljharb: I would certainly prefer to avoid implicit boundaries, and to have inclusiveness and exclusiveness explicitly indicated.

spec.html
michaelficarra: Okay, to me those aspects were independent. I will change the bounds.

spec.html
michaelficarra: @ljharb It already always is explicit. This is about whether the notation would even be defined for such an editorial error.

@bakkot We already explicitly describe that scenario as an editorial error. I don't think we also need to go out of our way to avoid the (true!) implication that the two are mutually exclusive. We're going to sacrifice clarity going through these gymnastics.

spec.html
michaelficarra: Okay I buy that. I'll drop it for now.

spec.html
michaelficarra: I still don't like your suggested wording. Saying "if inclusive" and then "if exclusive" feels like it's intentionally trying to permit a case where a bound is both inclusive and exclusive.

I really don't think it's necessary to be trying to prevent this error by construction. How about we just add a lint rule for the usage?

bakkot: I strongly prefer that we avoid defining it entirely. I am open to another wording which accomplishes this if you have one to suggest, but I am against any wording which gives picks one of the two to be the default when neither "inclusive" nor "exclusive" is used, even if we scrupulously avoid ever being in that scenario.

bakkot: If you want to be more explicit that an interval can't have one bound be both inclusive and exclusive - which really doesn't feel like something we need to tell our readers, but whatever - you can add a sentence before this one which says "Either condition may be either inclusive or exclusive", or something to that effect.

michaelficarra: @bakkot This wording doesn't have a default. I feel like we're looking at different things.

bakkot: > If an interval's lower bound is described as inclusive, its lower bound condition is a ≤ x; otherwise, it's lower bound is exclusive

How is that not a default? There is "described as inclusive" and there is "otherwise".

I want the cases to be "described as inclusive" and "described as exclusive".

michaelficarra: Otherwise, it is described as exclusive. The "otherwise" is there to make the two cases mutually exclusive, not to provide some default in the absence of a descriptor. We literally say

> It is an editorial error if any bound of an interval is not described as either inclusive or exclusive.

to make that impossible! I just can't fathom how that's not sufficient for you.

bakkot: To say "it is an editorial error to write X" is different from _declining to give a definition to X_, and the second thing is the thing I want us to do.

spec.html
```diff
@@ -1024,6 +1024,7 @@ <h1>Mathematical Operations</h1>
...
...ric value _x_ is included in the interval if it satisfies both its <em>lower bound condition</em> and its <em>upper bound condition</em>. If an interval's lower bound is described as inclusive, its lower bound condition is _a_ &le; _x_; otherwise, it's lower bound is exclusive and its lower bound condition is _a_ &lt; _x_. If an interval's upper bound is described as inclusive, its upper bound condition is _x_ &le; _b_; otherwise, it's upper bound is exclusive and its upper bound condition is _x_ &lt; _b_. For example, the interval from 1 (inclusive) to 2 (exclusive) consists of all mathematical values between 1 and 2, including 1 and not including 2. An <dfn id="inclusive-interval">inclusive interval</dfn> is an interval where both bounds are inclusive. It is an editorial error if any bound of an interval is not described as either inclusive or exclusive. For the purpose of defining intervals, *-0*<sub>𝔽</sub> &lt; <!-- this comment here to avoid ecmarkup complaining about comparison to 0 --> *+0*<sub>𝔽</sub>, so, for example, an inclusive interval with a lower bound of *+0*<sub>𝔽</sub> includes *+0*<sub>𝔽</sub> but not *-0*<sub>𝔽</sub>. *NaN* is never included in an interval.</p>
```
bakkot, line 1027:
```suggestion
      <p>An <dfn id="interval">interval</dfn> from lower bound _a_ to upper bound _b_ is a possibly-infinite, possibly-empty set of numeric values of the same numeric type. A numeric value _x_ is included in the interval if it satisfies both its <em>lower bound condition</em> and its <em>upper bound condition</em>. Each bound will always be described as either inclusive or exclusive. If an interval's lower bound is described as inclusive, its lower bound condition is _a_ &le; _x_; if its lower bound is described as exclusive, its lower bound condition is _a_ &lt; _x_. If an interval's upper bound is described as inclusive, its upper bound condition is _x_ &le; _b_; if its upper bound is described as exclusive, its upper bound condition is _x_ &lt; _b_. For example, the interval from 1 (inclusive) to 2 (exclusive) consists of all mathematical values between 1 and 2, including 1 and not including 2. An <dfn id="inclusive-interval">inclusive interval</dfn> is an interval where both bounds are inclusive. It is an editorial error if any bound of an interval is not described as either inclusive or exclusive. For the purpose of defining intervals, *-0*<sub>𝔽</sub> &lt; <!-- this comment here to avoid ecmarkup complaining about comparison to 0 --> *+0*<sub>𝔽</sub>, so, for example, an inclusive interval with a lower bound of *+0*<sub>𝔽</sub> includes *+0*<sub>𝔽</sub> but not *-0*<sub>𝔽</sub>. *NaN* is never included in an interval.</p>
```

bakkot (changes requested):

spec.html
michaelficarra: Adopted with tweaks.

bakkot: Works for me, thanks.

spec.html
```diff
@@ -1024,6 +1024,7 @@ <h1>Mathematical Operations</h1>
...
...val if it satisfies both its <em>lower bound condition</em> and its <em>upper bound condition</em>. Each bound will be described as either inclusive or exclusive, but not both. If an interval's lower bound is described as inclusive, its lower bound condition is _a_ &le; _x_. If an interval's lower bound is instead described as exclusive, its lower bound condition is _a_ &lt; _x_. If an interval's upper bound is described as inclusive, its upper bound condition is _x_ &le; _b_. If an interval's upper bound is instead described as exclusive, its upper bound condition is _x_ &lt; _b_. For example, the interval from 1 (inclusive) to 2 (exclusive) consists of all mathematical values between 1 and 2, including 1 and not including 2. An <dfn id="inclusive-interval">inclusive interval</dfn> is an interval with an inclusive lower bound and an inclusive upper bound. For the purpose of defining intervals, *-0*<sub>𝔽</sub> &lt; <!-- this comment here to avoid ecmarkup complaining about comparison to 0 --> *+0*<sub>𝔽</sub>, so, for example, an inclusive interval with a lower bound of *+0*<sub>𝔽</sub> includes *+0*<sub>𝔽</sub> but not *-0*<sub>𝔽</sub>. *NaN* is never included in an interval.</p>
```
syg, line 1027: The first few sentences read overly verbose to me. How about

An interval from lower bound _a_ to upper bound _b_ is a possibly-infinite, possibly-empty set of numeric values of the same numeric type. A bound is either inclusive or exclusive, but not both. There are four kinds of intervals as follows.

- An interval from _a_ (inclusive) to _b_ (inclusive) includes all numeric values _x_ such that _a_ &le; _x_ &le; _b_
- An interval from _a_ (inclusive) to _b_ (exclusive) includes all numeric values _x_ such that _a_ &le; _x_ &lt; _b_
- An interval from _a_ (exclusive) to _b_ (inclusive) includes all numeric values _x_ such that _a_ &lt; _x_ &le; _b_
- An interval from _a_ (exclusive) to _b_ (exclusive) includes all numeric values _x_ such that _a_ &lt; _x_ &lt; _b_

An inclusive interval is [...]

bakkot: That also WFM. I might collapse the last sentence in:

---

- An interval from _a_ (inclusive) to _b_ (inclusive), also called an `<dfn>inclusive interval</dfn>` includes all numeric values _x_ such that _a_ &le; _x_ &le; _b_
- An interval from _a_ (inclusive) to _b_ (exclusive) includes all numeric values _x_ such that _a_ &le; _x_ &lt; _b_
- An interval from _a_ (exclusive) to _b_ (inclusive) includes all numeric values _x_ such that _a_ &lt; _x_ &le; _b_
- An interval from _a_ (exclusive) to _b_ (exclusive) includes all numeric values _x_ such that _a_ &lt; _x_ &lt; _b_

spec.html
```diff
@@ -46541,7 +46542,7 @@ <h1>Valid Chosen Reads</h1>
           1. Let _readLen_ be the number of elements of _readValue_.
           1. If _chosenLen_ &ne; _readLen_, then
             1. Return *false*.
-          1. If _chosenValue_[_i_] &ne; _readValue_[_i_] for any integer value _i_ in the range 0 through _chosenLen_, exclusive, then
+          1. If _chosenValue_[_i_] &ne; _readValue_[_i_] for some integer _i_ in the interval from 0 (inclusive) to _chosenLen_ (exclusive), then
```
ljharb, line 46545: why some instead of any?

bakkot: "any" is ambiguous; it can be read to mean "every". (As in "for any integer n, n can be written as a unique product of primes" - that's a claim about _every_ integer.) So [pedantic grammarians](https://faculty.math.illinois.edu/~west/grammar.html) [see point 19] suggest using "some" in contexts like this.

syg: Yeah natural language quantification is wild.

spec.html
michaelficarra: Updated.

spec.html
michaelficarra: Changed all the loops over intervals to use chained comparison instead.

michaelficarra:
I noticed this weird ecmarkup rendering issue, so we may want to hold off merging until it is fixed.

image

Ping @bakkot if you want to open an ecmarkup issue with a repro case. I'm guessing it has to do with the HTML comment?

spec.html
```diff
@@ -33759,7 +33767,7 @@ <h1>String.prototype.lastIndexOf ( _searchString_ [ , _position_ ] )</h1>
           1. Let _start_ be the result of clamping _pos_ between 0 and _len_.
           1. If _searchStr_ is the empty String, return 𝔽(_start_).
           1. Let _searchLen_ be the length of _searchStr_.
-          1. For each non-negative integer _i_ starting with _start_ such that _i_ &le; _len_ - _searchLen_, in descending order, do
+          1. For each non-negative integer _i_ such that _start_ &le; _i_ &le; _len_ - _searchLen_, in descending order, do
```
bakkot, line 33770: This is wrong - because this is in _descending_ order, `_start_` is (despite appearances) an _upper_ bound; the lower bound is 0, given by the "non-negative" part.

I think instead where we set `_start_` to the result of clamping `_pos_` below `_len_`, we should instead clamp it below `_len_ - _searchLen_`. Then this can become

```suggestion
          1. For each integer _i_ such that 0 &le; _i_ &le; _start_, in descending order, do
```
which is more reasonable.

spec.html
```diff
@@ -35920,7 +35928,7 @@ <h1>
         <!-- CharacterClassEscape -->
         <emu-grammar>CharacterClassEscape :: `d`</emu-grammar>
         <emu-alg>
-          1. Return the ten-element CharSet containing the characters `0` through `9` inclusive.
+          1. Return the ten-element CharSet containing the characters in the inclusive interval from `0` to `9`.
```
bakkot, line 35931: Thought we weren't going to do ranges over characters, per https://github.com/tc39/ecma262/pull/2848#discussion_r946283258?

spec.html
michaelficarra: Accepted your suggestion.

spec.html
michaelficarra: True. Fixed by mirroring the steps for `\s` but with `DecimalDigit`. See https://github.com/tc39/ecma262/commit/4bc999dfba270cc8b08b278cff77dd65ebff56a9.

bakkot: I would prefer to keep what was originally there; this change adds a layer of indirection for no benefit.

michaelficarra: I think it's an improvement over the hand-waving that was there previously. The other way I could see fixing this is listing all the digits inline. But this seems fine, since it's just like what we do already for the other character class escapes.

bakkot: I don't think it's an improvement over the hand-waving that was previously there, which was at no risk of causing confusion and did not require chasing the definition of a nonterminal to parse. Either way, that change doesn't belong in this PR; you can back it out and make a separate PR which we can discuss separately, if you'd like, but I don't want it to go in as part of this PR.

bakkot: This PR changes all current definitions of _numeric_ intervals/ranges, which this isn't.

michaelficarra:
@bakkot done

spec.html
michaelficarra: #2876

spec.html
```diff
@@ -1760,17 +1768,17 @@ <h1>The Number Type</h1>
         <div class="math-display">
           _s_ &times; _m_ &times; 2<sup>_e_</sup>
         </div>
-        <p>where _s_ is 1 or -1, _m_ is an integer such that 2<sup>52</sup> &le; _m_ &lt; 2<sup>53</sup>, and _e_ is an integer such that -1074 &le; _e_ &le; 971.</p>
+        <p>where _s_ is 1 or -1, _m_ is an integer in the interval from 2<sup>52</sup> (inclusive) to 2<sup>53</sup> (exclusive), and _e_ is an integer in the inclusive interval from -1074 to 971.</p>
```
syg, line 1771: Fairly neutral on using intervals here, what do you think is gained from the current way @michaelficarra?

syg (approved):

spec.html
michaelficarra: I'm pretty neutral on this one, too, but I would err on the side of consistent phrasing. That means using intervals wherever it doesn't detract from readability.

syg: Fair enough.

PR #2853, opened 2022-08-10
Editorial: properly define and `<dfn>` spec enums

michaelficarra:
Split from #2821.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1029,7 +1029,6 @@
     <emu-clause id="sec-value-notation">
       <h1>Value Notation</h1>
       <p>In this specification, ECMAScript language values are displayed in *bold*. Examples include *null*, *true*, or *"hello"*. These are distinguished from ECMAScript source text such as `Function.prototype.apply` or `let n = 42;`.</p>
-      <p>Values which are internal to the specification and not directly observable from ECMAScript code are indicated with a ~sans-serif~ typeface. For instance, a Completion Record's [[Type]] field takes on values like ~normal~, ~return~, or ~throw~.</p>
     </emu-clause>
   </emu-clause>
 </emu-clause>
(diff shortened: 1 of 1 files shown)


bakkot (approved):

spec.html
```diff
@@ -3900,6 +3899,11 @@ <h1>Well-Known Intrinsic Objects</h1>
     <h1>ECMAScript Specification Types</h1>
     <p>A specification type corresponds to meta-values that are used within algorithms to describe the semantics of ECMAScript language constructs and ECMAScript language types. The specification types include Reference, List, Completion Record, Property Descriptor, Environment Record, Abstract Closure, and Data Block. Specification type values are specification artefacts that do not necessarily correspond to any specific entity within an ECMAScript implementation. Specification type values may be used to describe intermediate results of ECMAScript expression evaluation but such values cannot be stored as properties of objects or values of ECMAScript language variables.</p>
 
+    <emu-clause id="sec-enum-specification-type">
+      <h1>The Enum Specification Type</h1>
```
jmdyck, line 3903: `Enum` is a type insofar as you can define it to be a collection of values. But the spec never uses it as a type (e.g., to specify the type of a parameter or a record field), and probably never will, because doing so would be less informative than listing the specific values allowed.

spec.html
```diff
@@ -3900,6 +3899,11 @@ <h1>Well-Known Intrinsic Objects</h1>
...
+      <p><dfn variants="enum,enums">Enums</dfn> are values which are internal to the specification and not directly observable from ECMAScript code. Enums are denoted using a ~sans-serif~ typeface. For instance, a Completion Record's [[Type]] field takes on values like ~normal~, ~return~, or ~throw~. The content of an enum serves no purpose other than to distinguish it from other enums, and implies nothing about its usage or meaning in context.</p>
```
jmdyck, line 3904: > Enums are values which are internal to the specification and not directly observable from ECMAScript code.

While this is true, it's not very good as a definition, because it's true of *every* specification type value.

> The content of an enum ...

Enums have content? I'd be inclined to say the *name* of an enum.

spec.html
michaelficarra: Agreed. Still, the type is useful within this section to talk about how all enum values behave. Also we need a descriptive section header.

spec.html
michaelficarra: > While this is true, it's not very good as a definition, because it's true of _every_ specification type value.

I was mostly just taking it from the existing wording. Added "Enums have no characteristics other than their name".

> Enums have content? I'd be inclined to say the _name_ of an enum.

Sure. Done.

PR #2854, opened 2022-08-13
Editorial: clarify string representation of numbers with radix ≠ 10

### LINKED ISSUE #2847: How “implementation-defined” is %Number.prototype.toString% meant to be?
The final step given for [Number.prototype.toString](https://tc39.es/ecma262/#sec-number.prototype.tostring) says:

> 6. Return the String representation of this [Number value](https://tc39.es/ecma262/#number-value) using the radix specified by radixMV. Letters a-z are used for digits with values 10 through 35. The precise algorithm is [implementation-defined](https://tc39.es/ecma262/#implementation-defined), however the algorithm should be a generalization of that specified in [6.1.6.1.20](https://tc39.es/ecma262/#sec-numeric-types-number-tostring).

The “generalization of...” bit is only qualified with “should” (and I’m not certain how much “generalization of” conveys anyway). It surprised me to learn that this has been left pretty open-ended mainly because I wasn’t aware of engine differences that might have motivated it. BigInt inherited the same pattern.

Could/should the exact behavior be specified? If not, what prevented it?

michaelficarra:
Fixes #2847 by adding a `radix` parameter to `Number::toString`.

/cc @bathos

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2184,42 +2184,55 @@
           <h1>
             Number::toString (
               _x_: a Number,
+              _radix_: an integer in the inclusive interval from 2 to 36,
             ): a String
           </h1>
           <dl class="header">
             <dt>description</dt>
-            <dd>It converts _x_ to String format.</dd>
+            <dd>It represents _x_ as a String using a positional numeral system with radix _radix_. The digits used in the representation of a number using radix _r_ are taken from the first _r_ code units of *"0123456789abcdefghijklmnopqrstuvwxyz"* in order. The representation of numbers with magnitude greater than or equal to *1*<sub>𝔽</sub> never includes leading zeroes.</dd>
           </dl>
           <emu-alg>
             1. If _x_ is *NaN*, return the String *"NaN"*.
             1. If _x_ is *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, return the String *"0"*.
-            1. If _x_ &lt; *-0*<sub>𝔽</sub>, return the string-concatenation of *"-"* and Number::toString(-_x_).
+            1. If _x_ &lt; *-0*<sub>𝔽</sub>, return the string-concatenation of *"-"* and Number::toString(-_x_, _radix_).
             1. If _x_ is *+&infin;*<sub>𝔽</sub>, return the String *"Infinity"*.
-            1. [id="step-number-tostring-intermediate-values"] Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, 10<sup>_k_ - 1</sup> &le; _s_ &lt; 10<sup>_k_</sup>, 𝔽(_s_ &times; 10<sup>_n_ - _k_</sup>) is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the decimal representation of _s_, that _s_ is not divisible by 10, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
-            1. If _k_ &le; _n_ &le; 21, return the string-concatenation of:
-              * the code units of the _k_ digits of the decimal representation of _s_ (in order, with no leading zeroes)
-              * _n_ - _k_ occurrences of the code unit 0x0030 (DIGIT ZERO)
-            1. If 0 &lt; _n_ &le; 21, return the string-concatenation of:
-              * the code units of the most significant _n_ digits of the decimal representation of _s_
-              * the code unit 0x002E (FULL STOP)
-              * the code units of the remaining _k_ - _n_ digits of the decimal representation of _s_
-            1. If -6 &lt; _n_ &le; 0, return the string-concatenation of:
-              * the code unit 0x0030 (DIGIT ZERO)
-              * the code unit 0x002E (FULL STOP)
-              * -_n_ occurrences of the code unit 0x0030 (DIGIT ZERO)
-              * the code units of the _k_ digits of the decimal representation of _s_
-            1. Otherwise, if _k_ = 1, return the string-concatenation of:
-              * the code unit of the single digit of _s_
-              * the code unit 0x0065 (LATIN SMALL LETTER E)
-              * the code unit 0x002B (PLUS SIGN) or the code unit 0x002D (HYPHEN-MINUS) according to whether _n_ - 1 is positive or negative
-              * the code units of the decimal representation of the integer abs(_n_ - 1) (with no leading zeroes)
+            1. [id="step-number-tostring-intermediate-values"] Let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, _radix_<sup>_k_ - 1</sup> &le; _s_ &lt; _radix_<sup>_k_</sup>, 𝔽(_s_ &times; _radix_<sup>_n_ - _k_</sup>) is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the representation of _s_ using radix _radix_, that _s_ is not divisible by _radix_, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
+            1. If _radix_ &ne; 10 or _n_ is in the inclusive interval from -5 to 21, then
+              1. If _n_ &ge; _k_, then
+                1. Return the string-concatenation of:
+                  * the code units of the _k_ digits of the representation of _s_ using radix _radix_
+                  * _n_ - _k_ occurrences of the code unit 0x0030 (DIGIT ZERO)
+              1. Else if _n_ &gt; 0, then
+                1. Return the string-concatenation of:
+                  * the code units of the most significant _n_ digits of the representation of _s_ using radix _radix_
+                  * the code unit 0x002E (FULL STOP)
+                  * the code units of the remaining _k_ - _n_ digits of the representation of _s_ using radix _radix_
+              1. Else,
+                1. Assert: _n_ &le; 0.
+                1. Return the string-concatenation of:
+                  * the code unit 0x0030 (DIGIT ZERO)
+                  * the code unit 0x002E (FULL STOP)
+                  * -_n_ occurrences of the code unit 0x0030 (DIGIT ZERO)
+                  * the code units of the _k_ digits of the representation of _s_ using radix _radix_
+            1. NOTE: In this case, the input will be represented using scientific E notation, such as `1.2e+3`.
+            1. Assert: _radix_ is 10.
+            1. If _n_ &lt; 0, then
+              1. Let _exponentSign_ be the code unit 0x002D (HYPHEN-MINUS).
+            1. Else,
+              1. Let _exponentSign_ be the code unit 0x002B (PLUS SIGN).
+            1. If _k_ is 1, then
+              1. Return the string-concatenation of:
+                * the code unit of the single digit of _s_
+                * the code unit 0x0065 (LATIN SMALL LETTER E)
+                * _exponentSign_
+                * the code units of the decimal representation of abs(_n_ - 1)
             1. Return the string-concatenation of:
-              * the code units of the most significant digit of the decimal representation of _s_
+              * the code unit of the most significant digit of the decimal representation of _s_
               * the code unit 0x002E (FULL STOP)
               * the code units of the remaining _k_ - 1 digits of the decimal representation of _s_
               * the code unit 0x0065 (LATIN SMALL LETTER E)
-              * the code unit 0x002B (PLUS SIGN) or the code unit 0x002D (HYPHEN-MINUS) according to whether _n_ - 1 is positive or negative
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -2185,42 +2185,51 @@ <h1>
...
-              * the code units of the most significant _n_ digits of the decimal representation of _s_
-              * the code unit 0x002E (FULL STOP)
-              * the code units of the remaining _k_ - _n_ digits of the decimal representation of _s_
-            1. If -6 &lt; _n_ &le; 0, return the string-concatenation of:
-              * the code unit 0x0030 (DIGIT ZERO)
-              * the code unit 0x002E (FULL STOP)
-              * -_n_ occurrences of the code unit 0x0030 (DIGIT ZERO)
-              * the code units of the _k_ digits of the decimal representation of _s_
-            1. Otherwise, if _k_ = 1, return the string-concatenation of:
+            1. [id="step-number-tostring-intermediate-values"] Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, _radix_<sup>_k_ - 1</sup> &le; _s_ &lt; _radix_<sup>_k_</sup>, 𝔽(_s_ &times; _radix_<sup>_n_ - _k_</sup>) is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the representation of _s_ using using radix _radix_, that _s_ is not divisible by _radix_, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
```
bakkot, line 2200: You should update the NOTE which refers to this step also, since it still assumes base 10.

bakkot (approved):
LGTM other than comment but I want to wait for @syg's approval as well.

spec.html
michaelficarra: Done.

bathos:
@michaelficarra [%Bigint.prototype.toString%](https://tc39.es/ecma262/#sec-bigint.prototype.tostring) uses the same “should be a generalization” text and points at [BigInt::toString](https://tc39.es/ecma262/#sec-numeric-types-bigint-tostring). As far as I can tell the algorithm-in-practice aligns with Number-when-not-base-10. Would make sense (and would it be desirable) for them to point to a common algorithm that’s agnostic to the specific numeric language type?

(This PR already stands on its own with plenty of value, so I’m not suggesting it should address anything related to %BigInt.prototype.toString% — just curious about how that could end up looking.)

spec.html
```diff
@@ -2184,42 +2184,51 @@ <h1>
...
+              1. If _radix_ ≠ 10 or _n_ &gt; -6, return the string-concatenation of:
+                * the code unit 0x0030 (DIGIT ZERO)
+                * the code unit 0x002E (FULL STOP)
+                * -_n_ occurrences of the code unit 0x0030 (DIGIT ZERO)
+                * the code units of the _k_ digits of the representation of _s_ using radix _radix_
+            1. Assert: _radix_ is 10.
+            1. If _n_ &lt; 0, then
+              1. Let _sign_ be the code unit 0x002D (HYPHEN-MINUS).
+            1. Else,
+              1. Let _sign_ be the code unit 0x002B (PLUS SIGN).
```
gibson042, line 2219: This alias could be more clearly named.
```suggestion
            1. If _n_ &lt; 0, then
              1. Let _exponentSign_ be the code unit 0x002D (HYPHEN-MINUS).
            1. Else,
              1. Let _exponentSign_ be the code unit 0x002B (PLUS SIGN).
```

gibson042 (approved):
I like the improved generality and precision, and this does a decent job of containing the decimal-specific scale weirdness. That said, I do wonder if the latter could be achieved even better, e.g.
```
1. If _radix_ &ne; 10 or _n_ is in the inclusive interval from -5 to 21, then
  1. If _n_ &ge; _k_, return the string-concatenation of…
  1. If _n_ &gt; 0, return the string-concatenation of…
  1. Else,
    1. Assert: _n_ &le; 0.
    1. Return the string-concatenation of…
1. Assert: _radix_ is 10.
1. …
```

spec.html
```diff
@@ -2184,42 +2184,51 @@ <h1>
...
-              * the code units of the remaining _k_ - _n_ digits of the decimal representation of _s_
-            1. If -6 &lt; _n_ &le; 0, return the string-concatenation of:
-              * the code unit 0x0030 (DIGIT ZERO)
-              * the code unit 0x002E (FULL STOP)
-              * -_n_ occurrences of the code unit 0x0030 (DIGIT ZERO)
-              * the code units of the _k_ digits of the decimal representation of _s_
-            1. Otherwise, if _k_ = 1, return the string-concatenation of:
+            1. [id="step-number-tostring-intermediate-values"] Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, _radix_<sup>_k_ - 1</sup> &le; _s_ &lt; _radix_<sup>_k_</sup>, 𝔽(_s_ &times; _radix_<sup>_n_ - _k_</sup>) is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the representation of _s_ using using radix _radix_, that _s_ is not divisible by _radix_, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
+            1. If _k_ &le; _n_, then
+              1. If _radix_ ≠ 10 or _n_ &le; 21, return the string-concatenation of:
```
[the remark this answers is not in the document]
michaelficarra: I don't know if we ever actually decided that, but we are at least consistent at the moment (aside from ℝ, ℤ, 𝔽). 

See https://github.com/tc39/ecmarkup/issues/173#issuecomment-658447485

michaelficarra:
@gibson042 Thanks for the suggestions. I took all of them. Please re-review to check my work.

@bathos Agree, we could do something similar about `%Bigint.prototype.toString%`. Can you open an issue to track it? I don't think they should share an algorithm, but we can at least make a similar clarification.

spec.html
```diff
@@ -2184,42 +2184,54 @@ <h1>
...
-              * the code unit 0x002E (FULL STOP)
-              * -_n_ occurrences of the code unit 0x0030 (DIGIT ZERO)
-              * the code units of the _k_ digits of the decimal representation of _s_
-            1. Otherwise, if _k_ = 1, return the string-concatenation of:
-              * the code unit of the single digit of _s_
-              * the code unit 0x0065 (LATIN SMALL LETTER E)
-              * the code unit 0x002B (PLUS SIGN) or the code unit 0x002D (HYPHEN-MINUS) according to whether _n_ - 1 is positive or negative
-              * the code units of the decimal representation of the integer abs(_n_ - 1) (with no leading zeroes)
+            1. [id="step-number-tostring-intermediate-values"] Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, _radix_<sup>_k_ - 1</sup> &le; _s_ &lt; _radix_<sup>_k_</sup>, 𝔽(_s_ &times; _radix_<sup>_n_ - _k_</sup>) is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the representation of _s_ using using radix _radix_, that _s_ is not divisible by _radix_, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
```
jmdyck, line 2199:
```suggestion
            1. [id="step-number-tostring-intermediate-values"] Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, _radix_<sup>_k_ - 1</sup> &le; _s_ &lt; _radix_<sup>_k_</sup>, 𝔽(_s_ &times; _radix_<sup>_n_ - _k_</sup>) is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the representation of _s_ using radix _radix_, that _s_ is not divisible by _radix_, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
```

spec.html
```diff
@@ -31256,9 +31268,8 @@ <h1>Number.prototype.toString ( [ _radix_ ] )</h1>
           1. Let _x_ be ? thisNumberValue(*this* value).
           1. If _radix_ is *undefined*, let _radixMV_ be 10.
           1. Else, let _radixMV_ be ? ToIntegerOrInfinity(_radix_).
-          1. If _radixMV_ &lt; 2 or _radixMV_ &gt; 36, throw a *RangeError* exception.
-          1. If _radixMV_ = 10, return ! ToString(_x_).
-          1. Return the String representation of this Number value using the radix specified by _radixMV_. Letters `a`-`z` are used for digits with values 10 through 35. The precise algorithm is implementation-defined, however the algorithm should be a generalization of that specified in <emu-xref href="#sec-numeric-types-number-tostring"></emu-xref>.
+          1. If _radixMV_ is not in the inclusive interval from 2 to 36, throw a *RangeError* exception.
+          1. Return Number::toString(_x_, _radixMv_).
```
jmdyck, line 31272:
```suggestion
          1. Return Number::toString(_x_, _radixMV_).
```

michaelficarra:
Thanks @jmdyck. Done.

spec.html
```diff
@@ -2184,42 +2184,54 @@ <h1>
...
-              * the code unit 0x002E (FULL STOP)
-              * -_n_ occurrences of the code unit 0x0030 (DIGIT ZERO)
-              * the code units of the _k_ digits of the decimal representation of _s_
-            1. Otherwise, if _k_ = 1, return the string-concatenation of:
-              * the code unit of the single digit of _s_
-              * the code unit 0x0065 (LATIN SMALL LETTER E)
-              * the code unit 0x002B (PLUS SIGN) or the code unit 0x002D (HYPHEN-MINUS) according to whether _n_ - 1 is positive or negative
-              * the code units of the decimal representation of the integer abs(_n_ - 1) (with no leading zeroes)
+            1. [id="step-number-tostring-intermediate-values"] Otherwise, let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, _radix_<sup>_k_ - 1</sup> &le; _s_ &lt; _radix_<sup>_k_</sup>, 𝔽(_s_ &times; _radix_<sup>_n_ - _k_</sup>) is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the representation of _s_ using radix _radix_, that _s_ is not divisible by _radix_, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
```
gibson042, line 2199: Unnecessary "otherwise".
```suggestion
            1. [id="step-number-tostring-intermediate-values"] Let _n_, _k_, and _s_ be integers such that _k_ &ge; 1, _radix_<sup>_k_ - 1</sup> &le; _s_ &lt; _radix_<sup>_k_</sup>, 𝔽(_s_ &times; _radix_<sup>_n_ - _k_</sup>) is _x_, and _k_ is as small as possible. Note that _k_ is the number of digits in the representation of _s_ using radix _radix_, that _s_ is not divisible by _radix_, and that the least significant digit of _s_ is not necessarily uniquely determined by these criteria.
```

gibson042 (approved):

syg (approved):

spec.html
```diff
@@ -2184,42 +2184,54 @@ <h1>
...
+                  * the code unit 0x002E (FULL STOP)
+                  * the code units of the remaining _k_ - _n_ digits of the representation of _s_ using radix _radix_
+              1. Else,
+                1. Assert: _n_ &le; 0.
+                1. Return the string-concatenation of:
+                  * the code unit 0x0030 (DIGIT ZERO)
+                  * the code unit 0x002E (FULL STOP)
+                  * -_n_ occurrences of the code unit 0x0030 (DIGIT ZERO)
+                  * the code units of the _k_ digits of the representation of _s_ using radix _radix_
+            1. Assert: _radix_ is 10.
```
bakkot, line 2217:
```suggestion
            1. NOTE: In this case, the input will be represented using `1.2e+3`-style scientific E notation.
            1. Assert: _radix_ is 10.
```

michaelficarra: Slightly tweaked the wording to not imply that this is the one and only format.

bakkot: Works for me.

spec.html
```diff
@@ -2184,42 +2184,55 @@ <h1>
           <h1>
             Number::toString (
               _x_: a Number,
+              _radix_: an integer in the inclusive interval from 2 to 36,
             ): a String
           </h1>
           <dl class="header">
             <dt>description</dt>
-            <dd>It converts _x_ to String format.</dd>
+            <dd>It represents _x_ as a String using a positional numeral system with radix _radix_. The digits used in the representation of a number using radix _r_ are taken from the first _r_ code units of *"0123456789abcdefghijklmnopqrstuvwxyz"* in order. The representation of numbers with magnitude greater than or equal to *1*<sub>𝔽</sub> never includes leading zeros.</dd>
```
michaelficarra, line 2192: FYI this spelling of "zeros" is different from what we had before but I'm unable to determine what the OED says, if anything, about the preferred spelling. I will change it back to "zeroes" for now.

PR #2857, opened 2022-08-15
Editorial: remove misleading sentence about outdated use of "empty"

### LINKED ISSUE #2856: A CompletionRecord [[Value]] can be empty
The [Evaluation](https://tc39.es/ecma262/#sec-evaluation) of empty statements, empty statement lists, empty blocks (and a few other kinds of statements) does "Return `empty`", which according to the [implicit normal completion convention](https://tc39.es/ecma262/#sec-implicit-normal-completion) means to return a normal Completion record with `empty` as the [[Value]].

Sometimes the spec appears to treat `empty` like a special (specification type) value, however [`empty` is described as](https://tc39.es/ecma262/#sec-ecmascript-data-types-and-values) "*When the term “empty” is used as if it was naming a value, it is equivalent to saying “no value of any type”.*"

In particular, the [Completion Record Specification Type](https://tc39.es/ecma262/#sec-completion-record-specification-type) table of fields currently says that *[[Value]]* can be "*any value except a Completion Record*" meaning "*The value that was produced.*". I don't take "no value" to be included in "any value", so imo the table should instead say that *[[Value]]* can be "*any value except a Completion Record, or `empty`*", with the meaning of "*The value that was produced, if any*".

michaelficarra:
Fixes #2856.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1036,7 +1036,7 @@
 <emu-clause id="sec-ecmascript-data-types-and-values" aoid="Type">
   <h1>ECMAScript Data Types and Values</h1>
   <p>Algorithms within this specification manipulate values each of which has an associated type. The possible value types are exactly those defined in this clause. Types are further subclassified into ECMAScript language types and specification types.</p>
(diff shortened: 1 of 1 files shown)


syg (approved):
This is an improvement, lgtm.

bakkot (approved):

PR #2859, opened 2022-08-16
Editorial: add stars around null values that are missing them

### LINKED ISSUE #2565: Editorial: sometimes we don't put stars around null
We should probably be consistent.

E.g. [here](https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#table-exportentry-records) we leave them out.

michaelficarra:
Fixes #2565. The original issue mentioned in that issue had already been fixed. This was the rest of what I could find with some grepping.


spec.html
```diff
@@ -320,7 +320,7 @@ <h1>built-in object</h1>
     </emu-clause>
 
     <emu-clause id="sec-undefined-value">
-      <h1>undefined value</h1>
+      <h1>*undefined* value</h1>
```
ljharb, line 323: this kind of feels circular - `*undefined*` is the "undefined value", and vice versa.

(same below with "null value")

michaelficarra: Doesn't matter much to me. I was just including everything I could find. Happy to drop it if that's not what we want. Actual presentation is identical, anyway (though markup is different).

bakkot: I would very weakly lean towards leaving these without stars, I guess, given that it doesn't affect the market.

bakkot (approved):
LGTM other than the `<h1>`s.

spec.html

syg:
Happy to defer.

PR #2862, opened 2022-08-18
Editorial: replace implicit `~[empty]~` in ForBodyEvaluation with explicit `~empty~`

michaelficarra:
Previously we were acting as though the value of a missing parse node in an SDO definition was `~[empty]~` but that's not how it works as far as I'm aware. Now I explicitly test for their presence and pass the (more typical) enum `~empty~`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -21757,12 +21757,16 @@
           1. If the first |Expression| is present, then
             1. Let _exprRef_ be ? Evaluation of the first |Expression|.
             1. Perform ? GetValue(_exprRef_).
-          1. Return ? ForBodyEvaluation(the second |Expression|, the third |Expression|, |Statement|, &laquo; &raquo;, _labelSet_).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -21750,12 +21750,20 @@ <h1>
           1. If the first |Expression| is present, then
             1. Let _exprRef_ be ? Evaluation of the first |Expression|.
             1. Perform ? GetValue(_exprRef_).
-          1. Return ? ForBodyEvaluation(the second |Expression|, the third |Expression|, |Statement|, &laquo; &raquo;, _labelSet_).
+          1. Let _test_ be ~empty~.
+          1. If the second |Expression| is present, set _test_ to the second |Expression|.
```
bakkot, line 21754:
```suggestion
          1. If the second |Expression| is present, let _test_ be the second |Expression|; else let _test_ be ~empty~.
```
maybe?

bakkot (approved):
I had the same thought; this is definitely an improvement.

spec.html

PR #2865, opened 2022-08-19
Editorial: simplify IsNoTearConfiguration

michaelficarra:
The `open` parameter of `IsNoTearConfiguration` can only be `~SeqCst~`, `~Unordered~`, or `~Init~`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -41020,7 +41020,7 @@ THH:mm:ss.sss
         </dl>
         <emu-alg>
           1. If IsUnclampedIntegerElementType(_type_) is *true*, return *true*.
-          1. If IsBigIntElementType(_type_) is *true* and _order_ is not ~Init~ or ~Unordered~, return *true*.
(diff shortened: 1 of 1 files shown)


syg (commented):
I prefer leaving it as it is for future-proofing reasons. It's likely that WebAssembly will be adding release/acquire atomics soon, in which case we'd need to do the same in JS. And release/acquire atomic accesses ought to also be tear-free.

I admit this rationale is pretty weak. Happy to be convinced.

michaelficarra:
@syg Wouldn't we prefer the step to read something like `... and _order_ is ~SeqCst~, ~release~, or ~acquire~` at that point? Are you just afraid that we'll forget to update `IsNoTearConfiguration`?

michaelficarra:
> Are you just afraid that we'll forget to update `IsNoTearConfiguration`?

The answer was yes, and this is reasonable.

PR #2866, opened 2022-08-20
Editorial: clarify BigInt.prototype.toString for non-decimal radices

### LINKED ISSUE #2860: %Bigint.prototype.toString% clarification
Initially the [How “implementation-defined” is %Number.prototype.toString% meant to be?](https://github.com/tc39/ecma262/issues/2847) issue concerned both `%Number.prototype.toString%` and `%BigInt.prototype.toString%`, which followed a common pattern that appeared to permit a degree of behavior divergence among implementations that seemed like it might not be necessary/desirable/intentional.

@michaelficarra brought the question to an editor call which (I’m inferring) resolved to make de facto / implied behavior requirements explicit for %Number.prototype.toString%. This is being achieved by giving `Number::toString` a new `radix` parameter and updated algorithmic steps that explain the derivation of non-decimal representations, as can be seen in [his corresponding PR](https://github.com/tc39/ecma262/pull/2854) (not yet merged at the time of writing this).

As that PR and the related discussion concerned `%Number.prototype.toString%` / `Number::toString` only, the “How implementation-defined...” issue should now be regarded as only concerning the `Number` case too while this new issue captures the related question of whether/how `%BigInt.prototype.toString%` / `BigInt::toString` might also be clarified/narrowed. [Per @michaelficarra](https://github.com/tc39/ecma262/pull/2854#issuecomment-1217098540):

> Agree, we could do something similar about %Bigint.prototype.toString%. [...] I don't think they should share an algorithm, but we can at least make a similar clarification.

michaelficarra:
Fixes #2860. Similar to what we did for `Number.prototype.toString` in #2854.

/cc @bathos

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2631,15 +2631,16 @@
           <h1>
             BigInt::toString (
               _x_: a BigInt,
+              _radix_: an integer in the inclusive interval from 2 to 36,
             ): a String
           </h1>
           <dl class="header">
             <dt>description</dt>
-            <dd>It converts _x_ to String format.</dd>
+            <dd>It represents _x_ as a String using a positional numeral system with radix _radix_. The digits used in the representation of a BigInt using radix _r_ are taken from the first _r_ code units of *"0123456789abcdefghijklmnopqrstuvwxyz"* in order. The representation of BigInts other than *0*<sub>ℤ</sub> never includes leading zeroes.</dd>
           </dl>
           <emu-alg>
-            1. If _x_ &lt; *0*<sub>ℤ</sub>, return the string-concatenation of the String *"-"* and BigInt::toString(-_x_).
-            1. Return the String value consisting of the code units of the digits of the decimal representation of _x_.
+            1. If _x_ &lt; *0*<sub>ℤ</sub>, return the string-concatenation of the String *"-"* and BigInt::toString(-_x_, _radix_).
+            1. Return the String value consisting of the representation of _x_ using radix _radix_.
           </emu-alg>
         </emu-clause>
       </emu-clause>
@@ -5622,7 +5623,7 @@
               BigInt
             </td>
             <td>
-              Return ! BigInt::toString(_argument_).
+              Return BigInt::toString(_argument_, 10).
             </td>
           </tr>
           <tr>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -31407,9 +31408,8 @@ <h1>BigInt.prototype.toString ( [ _radix_ ] )</h1>
           1. Let _x_ be ? thisBigIntValue(*this* value).
           1. If _radix_ is *undefined*, let _radixMV_ be 10.
           1. Else, let _radixMV_ be ? ToIntegerOrInfinity(_radix_).
-          1. If _radixMV_ &lt; 2 or _radixMV_ &gt; 36, throw a *RangeError* exception.
-          1. If _radixMV_ = 10, return ! ToString(_x_).
-          1. Return the String representation of this Number value using the radix specified by _radixMV_. Letters `a`-`z` are used for digits with values 10 through 35. The precise algorithm is implementation-defined, however the algorithm should be a generalization of that specified in <emu-xref href="#sec-numeric-types-bigint-tostring"></emu-xref>.
+          1. If _radixMV_ is not in the inclusive interval from 2 to 36, throw a *RangeError* exception.
+          1. Return BigInt::toString(_x_, _radixMV_).
```
bathos, line 31412: Should this be `!`’d? It seemingly can’t throw and the other `BigInt::toString()` does have an explicit `!`.

michaelficarra: No, ever since we changed how completion records are used, this is unnecessary for AOs that don't return completion records. The other call site's use of `!` should be removed though, so thanks for the catch.

michaelficarra: By the way, ecmarkup normally would catch these cases for us, but I'm guessing it isn't tracking these AOs because of their strange names. @bakkot?

bakkot: It's not the name. Rather, linting rules like that one only apply within algorithm steps, which this isn't, as far as ecmarkup is concerned.

bakkot (approved):

spec.html
ljharb: Shouldn't algorithm steps inside methods also be considered algorithm steps?

bakkot: Sorry, I'm referring specifically to the [table under ToString](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-tostring), which does not have anything to identify that it contains algorithm steps.

ljharb: ah right, good point

PR #2867, opened 2022-08-22
Editorial: get rid of BigInt::sameValue and BigInt::sameValueZero

michaelficarra:
These were only added because we were doing dispatch on Number and BigInt which meant they needed matching AOs. We no longer do that.

We could also get rid of `BigInt::equal` by inlining it into the single call site you see here.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1627,21 +1627,16 @@
             <td>
               Number::sameValue
             </td>
-            <td rowspan="2">
+            <td>
               `Object.is(x, y)`
             </td>
-            <td rowspan="2">
+            <td>
               Object internal methods,
               via <emu-xref href="#sec-samevalue" title></emu-xref>,
               to test exact value equality
             </td>
-            <td rowspan="2">
-              Boolean
-            </td>
-          </tr>
-          <tr>
             <td>
-              BigInt::sameValue
+              Boolean
             </td>
           </tr>
 
@@ -1649,21 +1644,16 @@
             <td>
               Number::sameValueZero
             </td>
-            <td rowspan="2">
+            <td>
               `[x].includes(y)`
             </td>
-            <td rowspan="2">
+            <td>
               Array, Map, and Set methods,
               via <emu-xref href="#sec-samevaluezero" title></emu-xref>,
               to test value equality, ignoring the difference between *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub>
             </td>
-            <td rowspan="2">
-              Boolean
-            </td>
-          </tr>
-          <tr>
             <td>
-              BigInt::sameValueZero
+              Boolean
             </td>
           </tr>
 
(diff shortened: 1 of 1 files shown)


ljharb:
It seems kind of confusing to have some of the Number/BigInt things be hardcoded algorithms and some be dispatched methods, rather than them all being the same thing?

michaelficarra:
Added a commit to inline `BigInt::equal`.

michaelficarra:
@ljharb we no longer do that dynamic dispatch anywhere. Any of these AOs that are only called in a single place should probably be inlined, since deduplication would be their only value at this point.

spec.html
```diff
@@ -6069,18 +6010,16 @@ <h1>
...
         <p>SameValueZero differs from SameValue only in that it treats *+0*<sub>𝔽</sub> and *-0*<sub>𝔽</sub> as equivalent.</p>
       </emu-note>
     </emu-clause>
 
-    <emu-clause id="sec-samevaluenonnumeric" type="abstract operation" oldids="sec-samevaluenonnumber">
+    <emu-clause id="sec-samevaluenonnumber" type="abstract operation" oldids="sec-samevaluenonnumeric">
       <h1>
-        SameValueNonNumeric (
+        SameValueNonNumber (
           _x_: an ECMAScript language value, but not a Number or a BigInt,
```
bakkot, line 6023: Gotta fix the type.

bakkot (approved):
I am neutral on this change but it looks correct, apart from my comment.

ljharb:
@michaelficarra i understand we have explicit dispatch now, which is obviously superior - but i meant, it's nice to have the API-ish consistency of "every X number type has one of the expected operation in the X:: set".

spec.html
michaelficarra: Thanks. Fixed.

michaelficarra:
@ljharb I could see that argument being made for most of the other AOs like this, but these equality comparison ones are only really going to be giving different results for Number.

spec.html
```diff
@@ -2441,20 +2436,6 @@ <h1>
           </emu-alg>
         </emu-clause>
 
-        <emu-clause id="sec-numeric-types-bigint-equal" type="numeric method">
```
ljharb, line 2444: this needs an oldID somewhere

spec.html
```diff
@@ -2465,34 +2455,6 @@ <h1>
...
-            ): a Boolean
-          </h1>
-          <dl class="header">
-          </dl>
-          <emu-alg>
-            1. Return BigInt::equal(_x_, _y_).
-          </emu-alg>
-        </emu-clause>
-
-        <emu-clause id="sec-numeric-types-bigint-sameValueZero" type="numeric method">
```
ljharb, line 2482: these IDs both need to be oldIDs somewhere

michaelficarra: Good catch. Done.

PR #2869, opened 2022-08-23
Editorial: define thenable

ikravets:
a quick fix for #2868

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -43635,6 +43635,7 @@ THH:mm:ss.sss
     </ul>
     <p>A promise is said to be <em>settled</em> if it is not pending, i.e. if it is either fulfilled or rejected.</p>
     <p>A promise is <em>resolved</em> if it is settled or if it has been &ldquo;locked in&rdquo; to match the state of another promise. Attempting to resolve or reject a resolved promise has no effect. A promise is <em>unresolved</em> if it is not resolved. An unresolved promise is always in the pending state. A resolved promise may be pending, fulfilled or rejected.</p>
+    <p>A <em>thenable</em> is an object with a callable `then` property.</p>
 
     <emu-clause id="sec-promise-abstract-operations">
       <h1>Promise Abstract Operations</h1>
--- a/spec.html
+++ b/spec.html
@@ -50582,6 +50582,7 @@ THH:mm:ss.sss
     </ul>
     <p>A promise is said to be <em>settled</em> if it is not pending, i.e. if it is either fulfilled or rejected.</p>
     <p>A promise is <em>resolved</em> if it is settled or if it has been “locked in” to match the state of another promise. Attempting to resolve or reject a resolved promise has no effect. A promise is <em>unresolved</em> if it is not resolved. An unresolved promise is always in the pending state. A resolved promise may be pending, fulfilled or rejected.</p>
+    <p>A <dfn variants="thenables">thenable</dfn> is an object _obj_ for which Get(_obj_, *"then"*) returns a function object.</p>
 
     <emu-clause id="sec-promise-abstract-operations">
       <h1>Promise Abstract Operations</h1>


spec.html
```diff
@@ -43635,6 +43635,7 @@ <h1>Promise Objects</h1>
     </ul>
     <p>A promise is said to be <em>settled</em> if it is not pending, i.e. if it is either fulfilled or rejected.</p>
     <p>A promise is <em>resolved</em> if it is settled or if it has been &ldquo;locked in&rdquo; to match the state of another promise. Attempting to resolve or reject a resolved promise has no effect. A promise is <em>unresolved</em> if it is not resolved. An unresolved promise is always in the pending state. A resolved promise may be pending, fulfilled or rejected.</p>
+    <p>A <em>thenable</em> is an object `o` with `o.then(f, r)` method defined. Any Promise is thenable.</p>
```
zloirock, line 43638: > Any Promise is thenable.

```js
const p = Promise.resolve();
p.__proto__ = null;
```
Is `p` thenable? No. Is `p` promise? For the spec - yes.

nicolo-ribaudo:
```suggestion
    <p>A <em>thenable</em> is an object with a callable `then` property.</p>
```

ikravets: @zloirock 
Interesting example! I'd like to discuss your seemingly simple question "Is p promise?" a little bit more in a separate comment below and not to bury it in a "resolved conversation backyard".

@nicolo-ribaudo 
For consistency, my wording for the definition is modeled upon the wording just couple of paragraphs above it:

> A promise `p` is fulfilled if `p.then(f, r)` will immediately enqueue a Job to call the function `f`.

However, I'm fine with your wording as well. I'm not sure who is considered a maintainer of this repo to give an official approval to either wording and merge.

ikravets:
I'd like to discuss a seemingly simple question "Is `p` promise?" a little bit more.

First, it leads me to a more general question: what constitute an [*is-a*](https://en.wikipedia.org/wiki/Is-a) relation  in JavaScript? I'm not sure the spec addresses this question explicitly. In general OOP sense, *is-a* determines which behavior is expected from an object. `[[Prototype]]` assignment changes the behavior and can be considered a change of an *is-a* relation. For example, `instanceof` can be easily "broken":

```javascript
const p = Promise.resolve();
console.log(p instanceof Promise); // --> true
Object.setPrototypeOf(p, null); 
console.log(p instanceof Promise); // --> false
```

When we "break" the expected behavior of an object is an *is-a* relation preserved or broken as well?

In case of `Promise` we indeed have `IsPromise()` in the spec, but is it the sole source of truth for an *is-a* relation? How *is-a* determined for other types which do not have `Is%Typename%()` operation?

Back to the [Promise Objects](https://tc39.es/ecma262/#sec-promise-objects) section. There are two definitions literally a dozen lines above the proposed patch:

> - A promise `p` is fulfilled if `p.then(f, r)` will immediately enqueue a Job to call the function `f`.
> - A promise `p` is rejected if `p.then(f, r)` will immediately enqueue a Job to call the function `r`.

When we break the `[[Prototype]]` these definitions no longer work. I guess, a non-negligible part of the spec breaks as well when we allow `[[Prototype]]` setting in arbitrary places. And of course there are also infinite number of possibilities to adjust the `[[Prototype]]` and thus the behavior to different extents which may constitute a different degrees of "behavior breakage".

Now back to your question "Is `p` promise?". I think that until *is-a* is formally defined[^1] my answer would be: it depends on what do you mean by that question. If you mean `IsPromise(p)` the answer is 'yes', if you mean `p instanceof Promise` the answer is 'no'. 

Summarizing the more general part:
- is there an official definition for *is-a* relation in the JavaScript spec[^1]?
- should one be added?
- should I open a new discussion/issue regarding this subject?

What do you think?

I'm by no means a JavaScript expert, quite the opposite. A feedback from more experienced people is much appreciated.

[^1]: and please point me to such a definition if it exists!

spec.html
ikravets: BTW, the wording

> Any Promise is thenable.

can be adjusted to

> A promise created by the Promise Constructor is *thenable* upon creation.

What do you think?

zloirock:
```js
delete `Promise.prototype.then`
```

bakkot:
>is there an official definition for is-a relation in the JavaScript spec?

No.

> what about subtyping in the general OOP sense?

No.

> should one be added?

In my opinion, no. None of the normative requirements in the spec require us to give precise definitions to any of those, and they are used in somewhat vague ways in general, so it's best we stay out of trying to give precise definitions.

spec.html
ikravets: I understand your approach. I think that breaking the `[[Prototype]]` and overriding built-in properties breaks the spec for any type, not just `Promise`. You can do the same with `Number`. 

We could add "in unmodified standard environment" footnote for every second sentence in the spec to make it true even under the breaking runtime changes. I think that this is implied, but I don't see a point to continue this discussion here. It's better to merge something than nothing. I like @nicolo-ribaudo's suggestion.

zloirock (commented):
👍

ljharb:
This would be nice to rebase, properly dfn, and land.

spec.html
nicolo-ribaudo:
```suggestion
    <p>A <dfn>thenable</dfn> is an object _obj_ for which Get(_obj_, *"then"*) returns a function object.</p>
```

michaelficarra: Also include a plural variant on the `dfn`.

nicolo-ribaudo (approved):

linusg (approved):

michaelficarra (approved):

ljharb:
@ikravets would you mind [registering as a contributor](https://tc39.es/agreements/contributor/)?

nicolo-ribaudo:
Fwiw the text has been completely reworded by me in 8b21e0c61f3604a0c0ebc51ea7ac1a6677eb847a, so we might also just ignore this specific commit in the IPR check.

ljharb:
fair point. i'll push up an exclusion and land the rest of the queue later today.

PR #2872, opened 2022-08-25
Meta: properly fix IPR script (#2872)

ljharb:
This uses the commits list API instead of the compare API, and actually covers every author. It also adds exceptions, and some legacy TODOs we need to address over time

### DIFF
--- a/scripts/check-form.js
+++ b/scripts/check-form.js
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
 
+const { execSync } = require('child_process');
+
 // web URL: `https://docs.google.com/spreadsheets/d/${sheetID}/edit`
 const sheetID = '1if5bU0aV5MJ27GGKnRzyAozeKP-ILXYl5r3dzvkGFmg';
 
@@ -25,6 +27,8 @@ if (!slug || !branch) {
 if (typeof all !== 'undefined' && all !== '--all') {
 	throw '`all` arg, if provided, must be `--all`'
 }
+
+const sha = String(execSync(`git rev-parse --short ${branch}`)).trim();
 
 const request = async (url, method = 'GET', postData) => {
 	// adapted from https://medium.com/@gevorggalstyan/how-to-promisify-node-js-http-https-requests-76a5a58ed90c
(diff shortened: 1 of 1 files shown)


scripts/check-form.js
```diff
@@ -70,11 +74,18 @@ const request = async (url, method = 'GET', postData) => {
...
 
-const branchURL = `https://api.github.com/repos/${slug}/compare/${all ? initial : 'HEAD'}...${branch}?anon=1`;
+const perPage = 100;
+
+async function getAllCommits(page = 1) {
+	const commitsURL = `https://api.github.com/repos/${slug}/commits?anon=1&per_page=${perPage}&page=${page}&sha=${sha}`;
+	const commits = await request(commitsURL).then((json) => JSON.parse(json));
+	return [...new Set([].concat(
+		commits.flatMap(x => x?.author?.login || []),
+		(true && commits.length < perPage) ? [] : await getAllCommits(page + 1),
```
bakkot, line 84: `true &&` seems stray?

ljharb: oh whoops, thanks, i was testing

scripts/check-form.js
```diff
@@ -128,10 +146,79 @@ const usernames = request(sheetData).then((json) => JSON.parse(json)).then(data
...
+				legacyCount -= 1;
+				console.warn(`TODO: remove ${a} from legacy list`);
+			}
+			return false;
+		}
+		return !signed;
+	});
 	if (missing.length > 0) {
-		throw `Missing authors: ${missing}`;
+		throw `Missing ${missing.length} authors: ${missing}${legacyCount > 0 ? ` \n(legacy missing: ${legacyCount})` : ''}`;
```
bakkot, line 221: I'm confused by the "legacy missing" part here. Members of the `legacy` set don't contribute to `missing.length` whether or not they've signed, so I don't know why they belong in the error message about `missing.length > 0`.

bakkot (approved):
Didn't run it but LGTM at a glance other than the error message I've commented on.

scripts/check-form.js
ljharb: this is a way for me to see, in the case where there is someone missing, how many people *would* have been missing if not for the legacy list.

i could certainly console log it outside the "if", though ¯\\\_(ツ)_/¯

ljharb: k, i've fixed that

Issue #2870, opened 2022-08-23
make Type(...) macro an AO and return a spec enum


michaelficarra:
The usage of `Type(...)` in the spec is quite strange and doesn't follow our usual editorial conventions. I would prefer if we just made it a regular AO and had it return spec enums, which can be more comfortably passed around and compared than the built-in constructor or whatever it is that is returned today. Also, we won't have to think about realm weirdness.

Jack-Works:
And the link to the definition of `Type` (it links to https://tc39.es/ecma262/#sec-ecmascript-data-types-and-values) doesn't explain what it will return very clearly. (e.g. I don't know if `Type(null)` is `null` or `object` at the first glance)

bathos:
I believe there was at least one usage where the `x` of  `Type(x)` wasn’t guaranteed to be a language value. I’m using past tense because I can’t find it now and it may have been removed. OTOH, it was pretty well-hidden because the _comparisons_ were always to language types (e.g. `if Type(x) is Object`), so I could just be missing it now. The current definition of `Type()` permits this (vaguely), so I don’t think it was a bug, but I’m guessing a clean up would want to eliminate non-language-type usage if it is till lurking out there somewhere...

ljharb:
@Jack-Works it returns "Null" for that, unambiguously.

@bathos that seems like it'd be an editorial error for a Type AO.

jmdyck:
Note that most uses of `Type()` could be easily eliminated. E.g., replace:
> `If Type(_v_) is String`

with:
> `If _v_ is a String`

However, there are a few places where it wouldn't be as easy, e.g.
- `If Type(_next_) is not an element of _elementTypes_`
- `If Type(_x_) is different from Type(_y_)`
- `If Type(_nx_) is the same as Type(_ny_)`

@bathos:
> I believe there was at least one usage where the `x` of `Type(x)` wasn’t guaranteed to be a language value.

In `SetFunctionName`, there's a `Type(_name_)`, where `_name_` could be a Private Name, which is not a language value. (But that case disappears if you perform the above replacement.)

There are also some cases like `Type(_completionRecord_.[[Value]])` where the argument might be `~empty~`, although it's possible that you could prove (in those particular cases) that it can't be.

Whether it's an editorial error to pass a spec value to a Type AO would depend on whether the AO is written to accept language values or all values. I agree that restricting it to just language values would be preferable, but ensuring that a spec value is never passed to it might take some diligence.

[...]

PR #2874, opened 2022-08-30
Editorial: avoid using Type macro for simple type tests

michaelficarra:
This is a partial resolution for #2870, as [suggested](https://github.com/tc39/ecma262/issues/2870#issuecomment-1224640184) by @jmdyck.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1065,12 +1065,12 @@
 
     <emu-clause id="sec-ecmascript-language-types-boolean-type">
       <h1>The Boolean Type</h1>
-      <p>The Boolean type represents a logical entity having two values, called *true* and *false*.</p>
+      <p>The <dfn variants="is a Boolean,is not a Boolean">Boolean type</dfn> represents a logical entity having two values, called *true* and *false*.</p>
     </emu-clause>
 
     <emu-clause id="sec-ecmascript-language-types-string-type">
       <h1>The String Type</h1>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -4417,7 +4417,7 @@ <h1>
         <dl class="header">
         </dl>
         <emu-alg>
-          1. If Type(_Obj_) is not Object, throw a *TypeError* exception.
+          1. If _Obj_ is not an Object, throw a *TypeError* exception.
```
ljharb, line 4420: i kind of think that this reads much less clearly for Object comparisons - with Type, i can look and confirm whether functions are Objects or not, but with this definition i'm left wondering.

jmdyck: To clarify, are you saying this?:
- In the rendered spec, the use of "Type" here is linked to section 6 "ECMAScript Data Types and Values" which (eventually) answers the question of whether a function is an Object.
- Whereas in the rendered spec, "is not an Object" doesn't link to anything.

ljharb: Yes, exactly.

jmdyck: In that case, we could make "is not an Object" link to something. E.g., somewhere add a sentence with `<dfn variants="is not an Object">is an Object</dfn>`. In fact, this would be an improvement over the status quo, because the fact that a function is an object could be stated right there at the `<dfn>`, whereas currently that information is rather a long way from the definition of `Type()`.

michaelficarra: Added `<dfn>`s. I think it's actually pretty nice.

bakkot (approved):
Didn't review for completeness, but what's here LGTM.

I think it would be worth putting something in [the "Object Type" section](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-object-type) which specifically calls out that some members of the type are callable. The current "is logically a collection of properties" definition really obscures that. Not as part of this PR, though.

PR #2875, opened 2022-08-30
Editorial: use "is not finite" in more places where appropriate and add dfn for "finite" so they link

michaelficarra:
We already have 17 occurrences of "is not finite" and 7 occurrences of "is finite", so I've identified the remaining locations where "is not finite" can be used and changed those. I also added a `<dfn>` for "finite" so it links.


bakkot (approved):

syg (approved):
lgtm

PR #2876, opened 2022-08-30
Editorial: define `\d` semantics in a similar way to `\s`, avoiding character range

michaelficarra:
This avoids the ill-defined character range.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -35853,7 +35853,7 @@ THH:mm:ss.sss
         <!-- CharacterClassEscape -->
         <emu-grammar>CharacterClassEscape :: `d`</emu-grammar>
         <emu-alg>
-          1. Return the ten-element CharSet containing the characters `0` through `9` inclusive.
+          1. Return the ten-element CharSet containing the characters `0`, `1`, `2`, `3`, `4`, `5`, `6`, `7`, `8`, and `9`.
         </emu-alg>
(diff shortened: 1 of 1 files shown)


gibson042 (approved):

bakkot (approved):
I am OK with this I guess.

syg (commented):
I find the new version less clear. Is there an alternative wording that keeps the 0-9?

syg:
> @syg We can just explicitly list the decimal digits, I guess.

Yeah that's fine by me.

spec.html
```diff
@@ -35844,7 +35844,7 @@ <h1>
         <!-- CharacterClassEscape -->
         <emu-grammar>CharacterClassEscape :: `d`</emu-grammar>
         <emu-alg>
-          1. Return the ten-element CharSet containing the characters `0` through `9` inclusive.
+          1. Return the ten-element CharSet containing the characters whose mathematical values are 0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, and 0x39.
```
bakkot, line 35847: Listing the code point values is also less clear than what was here originally. If you want to explicitly list the decimal digits, say "the characters `0`, `1`, `2`, `3`, `4`, `5`, `6`, `7`, `8`, `9`", not this.

syg (approved):

PR #2877, opened 2022-08-30
various editorial changes for comparisons

### LINKED ISSUE #2794: use consistent phrasing for comparing string values

michaelficarra:
Extracted from #2821. Fixes #2794.

### Editorial choices included in this PR
* use `=`, <code>&ne;</code>, `<`, <code>&le;</code>, `>`, and <code>&ge;</code> for mathematical value and bigint comparisons (except in asserts); prefer `=` over <code>&ne;</code>
* use `is`, `is not`, `<`, <code>&le;</code>, `>`, and <code>&ge;</code> for number comparisons
* use `SameValue(..., ...) is *true*` for equality comparison of symbols (except well-known) / objects / unknown ECMAScript language values
* use `is` and `is not` for equality comparison of all other values, including booleans, strings, well-known symbols, null, undefined, enums, etc; avoid "is different from" or "is the same as"
* prefer "if a is either b or c" over "if a is b or a is c"; use "any of" in place of "either" when more than 2 options
* prefer "if a is c and b is c" over "if a and b are both c"
* presence in a List is tested with `_list_ contains _element_` or `_list_ does not contain _element_`, not `_element_ is an element of _list_` or `_element_ is in _list_`
* don't say `the value of _a_`; just use `_a_`
* prefer "if a or b" over "if a or if b"; prefer "if a and b" over "if a and if b"
* prefer "whose _ is _" over "whose _ is that of _"

### Other editorial choices (still to do, not as part of this PR)
* use `a`/`an`/`a new` for creating and referencing values with identity
* use `the` for referencing values without identity

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -1160,11 +1160,11 @@
           1. Let _searchLen_ be the length of _searchValue_.
           1. For each integer _i_ such that _fromIndex_ ≤ _i_ ≤ _len_ - _searchLen_, in ascending order, do
             1. Let _candidate_ be the substring of _string_ from _i_ to _i_ + _searchLen_.
-            1. If _candidate_ is the same sequence of code units as _searchValue_, return _i_.
+            1. If _candidate_ is _searchValue_, return _i_.
           1. Return -1.
         </emu-alg>
         <emu-note>
-          <p>If _searchValue_ is the empty String and _fromIndex_ is less than or equal to the length of _string_, this algorithm returns _fromIndex_. The empty String is effectively found at every position within a string, including after the last code unit.</p>
+          <p>If _searchValue_ is the empty String and _fromIndex_ ≤ the length of _string_, this algorithm returns _fromIndex_. The empty String is effectively found at every position within a string, including after the last code unit.</p>
         </emu-note>
         <emu-note>
           <p>This algorithm always returns -1 if _fromIndex_ + the length of _searchValue_ > the length of _string_.</p>
@@ -1867,7 +1867,7 @@
           </dl>
           <emu-alg>
             1. If _exponent_ is *NaN*, return *NaN*.
-            1. If _exponent_ is *+0*<sub>𝔽</sub> or _exponent_ is *-0*<sub>𝔽</sub>, return *1*<sub>𝔽</sub>.
+            1. If _exponent_ is either *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, return *1*<sub>𝔽</sub>.
             1. If _base_ is *NaN*, return *NaN*.
             1. If _base_ is *+∞*<sub>𝔽</sub>, then
               1. If _exponent_ > *+0*<sub>𝔽</sub>, return *+∞*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
@@ -1886,11 +1886,11 @@
             1. Assert: _base_ is finite and is neither *+0*<sub>𝔽</sub> nor *-0*<sub>𝔽</sub>.
             1. If _exponent_ is *+∞*<sub>𝔽</sub>, then
               1. If abs(ℝ(_base_)) > 1, return *+∞*<sub>𝔽</sub>.
-              1. If abs(ℝ(_base_)) is 1, return *NaN*.
+              1. If abs(ℝ(_base_)) = 1, return *NaN*.
               1. If abs(ℝ(_base_)) &lt; 1, return *+0*<sub>𝔽</sub>.
             1. If _exponent_ is *-∞*<sub>𝔽</sub>, then
               1. If abs(ℝ(_base_)) > 1, return *+0*<sub>𝔽</sub>.
-              1. If abs(ℝ(_base_)) is 1, return *NaN*.
+              1. If abs(ℝ(_base_)) = 1, return *NaN*.
               1. If abs(ℝ(_base_)) &lt; 1, return *+∞*<sub>𝔽</sub>.
             1. Assert: _exponent_ is finite and is neither *+0*<sub>𝔽</sub> nor *-0*<sub>𝔽</sub>.
             1. If _base_ &lt; *-0*<sub>𝔽</sub> and _exponent_ is not an integral Number, return *NaN*.
@@ -1914,12 +1914,12 @@
           </dl>
           <emu-alg>
             1. If _x_ is *NaN* or _y_ is *NaN*, return *NaN*.
-            1. If _x_ is *+∞*<sub>𝔽</sub> or _x_ is *-∞*<sub>𝔽</sub>, then
-              1. If _y_ is *+0*<sub>𝔽</sub> or _y_ is *-0*<sub>𝔽</sub>, return *NaN*.
+            1. If _x_ is either *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, then
+              1. If _y_ is either *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, return *NaN*.
               1. If _y_ > *+0*<sub>𝔽</sub>, return _x_.
               1. Return -_x_.
-            1. If _y_ is *+∞*<sub>𝔽</sub> or _y_ is *-∞*<sub>𝔽</sub>, then
-              1. If _x_ is *+0*<sub>𝔽</sub> or _x_ is *-0*<sub>𝔽</sub>, return *NaN*.
+            1. If _y_ is either *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, then
+              1. If _x_ is either *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, return *NaN*.
               1. If _x_ > *+0*<sub>𝔽</sub>, return _y_.
               1. Return -_y_.
             1. If _x_ is *-0*<sub>𝔽</sub>, then
@@ -1948,16 +1948,16 @@
           </dl>
           <emu-alg>
             1. If _x_ is *NaN* or _y_ is *NaN*, return *NaN*.
-            1. If _x_ is *+∞*<sub>𝔽</sub> or _x_ is *-∞*<sub>𝔽</sub>, then
-              1. If _y_ is *+∞*<sub>𝔽</sub> or _y_ is *-∞*<sub>𝔽</sub>, return *NaN*.
+            1. If _x_ is either *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, then
+              1. If _y_ is either *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, return *NaN*.
               1. If _y_ is *+0*<sub>𝔽</sub> or _y_ > *+0*<sub>𝔽</sub>, return _x_.
               1. Return -_x_.
             1. If _y_ is *+∞*<sub>𝔽</sub>, then
               1. If _x_ is *+0*<sub>𝔽</sub> or _x_ > *+0*<sub>𝔽</sub>, return *+0*<sub>𝔽</sub>. Otherwise, return *-0*<sub>𝔽</sub>.
             1. If _y_ is *-∞*<sub>𝔽</sub>, then
               1. If _x_ is *+0*<sub>𝔽</sub> or _x_ > *+0*<sub>𝔽</sub>, return *-0*<sub>𝔽</sub>. Otherwise, return *+0*<sub>𝔽</sub>.
-            1. If _x_ is *+0*<sub>𝔽</sub> or _x_ is *-0*<sub>𝔽</sub>, then
-              1. If _y_ is *+0*<sub>𝔽</sub> or _y_ is *-0*<sub>𝔽</sub>, return *NaN*.
+            1. If _x_ is either *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, then
+              1. If _y_ is either *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, return *NaN*.
               1. If _y_ > *+0*<sub>𝔽</sub>, return _x_.
               1. Return -_x_.
             1. If _y_ is *+0*<sub>𝔽</sub>, then
@@ -1981,15 +1981,15 @@
           </dl>
           <emu-alg>
             1. If _n_ is *NaN* or _d_ is *NaN*, return *NaN*.
-            1. If _n_ is *+∞*<sub>𝔽</sub> or _n_ is *-∞*<sub>𝔽</sub>, return *NaN*.
-            1. If _d_ is *+∞*<sub>𝔽</sub> or _d_ is *-∞*<sub>𝔽</sub>, return _n_.
-            1. If _d_ is *+0*<sub>𝔽</sub> or _d_ is *-0*<sub>𝔽</sub>, return *NaN*.
-            1. If _n_ is *+0*<sub>𝔽</sub> or _n_ is *-0*<sub>𝔽</sub>, return _n_.
+            1. If _n_ is either *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, return *NaN*.
+            1. If _d_ is either *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, return _n_.
+            1. If _d_ is either *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, return *NaN*.
+            1. If _n_ is either *+0*<sub>𝔽</sub> or *-0*<sub>𝔽</sub>, return _n_.
             1. Assert: _n_ and _d_ are finite and non-zero.
             1. Let _quotient_ be ℝ(_n_) / ℝ(_d_).
             1. Let _q_ be truncate(_quotient_).
             1. Let _r_ be ℝ(_n_) - (ℝ(_d_) × _q_).
-            1. If _r_ is 0 and _n_ &lt; *-0*<sub>𝔽</sub>, return *-0*<sub>𝔽</sub>.
+            1. If _r_ = 0 and _n_ &lt; *-0*<sub>𝔽</sub>, return *-0*<sub>𝔽</sub>.
             1. Return 𝔽(_r_).
           </emu-alg>
           <emu-note>
@@ -2013,8 +2013,8 @@
             1. If _x_ is *NaN* or _y_ is *NaN*, return *NaN*.
             1. If _x_ is *+∞*<sub>𝔽</sub> and _y_ is *-∞*<sub>𝔽</sub>, return *NaN*.
             1. If _x_ is *-∞*<sub>𝔽</sub> and _y_ is *+∞*<sub>𝔽</sub>, return *NaN*.
-            1. If _x_ is *+∞*<sub>𝔽</sub> or _x_ is *-∞*<sub>𝔽</sub>, return _x_.
-            1. If _y_ is *+∞*<sub>𝔽</sub> or _y_ is *-∞*<sub>𝔽</sub>, return _y_.
+            1. If _x_ is either *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, return _x_.
+            1. If _y_ is either *+∞*<sub>𝔽</sub> or *-∞*<sub>𝔽</sub>, return _y_.
             1. Assert: _x_ and _y_ are both finite.
             1. If _x_ is *-0*<sub>𝔽</sub> and _y_ is *-0*<sub>𝔽</sub>, return *-0*<sub>𝔽</sub>.
             1. Return 𝔽(ℝ(_x_) + ℝ(_y_)).
@@ -2106,7 +2106,7 @@
           <emu-alg>
             1. If _x_ is *NaN*, return *undefined*.
             1. If _y_ is *NaN*, return *undefined*.
-            1. If _x_ and _y_ are the same Number value, return *false*.
+            1. If _x_ is _y_, return *false*.
             1. If _x_ is *+0*<sub>𝔽</sub> and _y_ is *-0*<sub>𝔽</sub>, return *false*.
             1. If _x_ is *-0*<sub>𝔽</sub> and _y_ is *+0*<sub>𝔽</sub>, return *false*.
             1. If _x_ is *+∞*<sub>𝔽</sub>, return *false*.
@@ -2130,7 +2130,7 @@
           <emu-alg>
             1. If _x_ is *NaN*, return *false*.
             1. If _y_ is *NaN*, return *false*.
-            1. If _x_ is the same Number value as _y_, return *true*.
+            1. If _x_ is _y_, return *true*.
             1. If _x_ is *+0*<sub>𝔽</sub> and _y_ is *-0*<sub>𝔽</sub>, return *true*.
             1. If _x_ is *-0*<sub>𝔽</sub> and _y_ is *+0*<sub>𝔽</sub>, return *true*.
             1. Return *false*.
@@ -2150,7 +2150,7 @@
             1. If _x_ is *NaN* and _y_ is *NaN*, return *true*.
             1. If _x_ is *+0*<sub>𝔽</sub> and _y_ is *-0*<sub>𝔽</sub>, return *false*.
             1. If _x_ is *-0*<sub>𝔽</sub> and _y_ is *+0*<sub>𝔽</sub>, return *false*.
-            1. If _x_ is the same Number value as _y_, return *true*.
+            1. If _x_ is _y_, return *true*.
             1. Return *false*.
           </emu-alg>
         </emu-clause>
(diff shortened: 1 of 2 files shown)


spec.html
```diff
@@ -15939,7 +15939,7 @@ <h1>
         1. Let _size_ be the length of _string_.
         1. Assert: _position_ &ge; 0 and _position_ &lt; _size_.
         1. Let _first_ be the code unit at index _position_ within _string_.
-        1. Let _cp_ be the code point whose numeric value is that of _first_.
+        1. Let _cp_ be the code point whose numeric value is _first_.
```
jmdyck, line 15942: The intent is to compare two numeric values: the numeric value of `_cp_` and the numeric value of `_first_`. You're proposing to compare the numeric value of `_cp_` with `_first_`, which is a code unit.

If you don't like the "is that of" construction, you could expand it to:
```suggestion
        1. Let _cp_ be the code point whose numeric value is the numeric value of _first_.
```

And similarly for the other cases.

michaelficarra: Ah, yes, I didn't check these and assumed they were all comparing against numeric values. I will accept your suggestion.

spec.html
```diff
@@ -36466,8 +36479,7 @@ <h1>
           <dl class="header">
           </dl>
           <emu-alg>
-            1. Assert: _match_.[[StartIndex]] is a non-negative integer less than or equal to the length of _S_.
-            1. Assert: _match_.[[EndIndex]] is an integer between _match_.[[StartIndex]] and the length of _S_, inclusive.
+            1. Assert: _match_.[[StartIndex]] &le; _match_.[[EndIndex]] &le; the length of _S_.
```
jmdyck, line 36482: That loses the assertion that `_match_.[[StartIndex]]` is a non-negative integer. So:
```suggestion
            1. Assert: 0 &le; _match_.[[StartIndex]] &le; _match_.[[EndIndex]] &le; the length of _S_.
```

And likewise for the next diff.

michaelficarra: It also no longer says anything about it being integral. That was intentional. `[[StartIndex]]` is [declared to be a non-negative integer](https://tc39.es/ecma262/#table-match-record).

michaelficarra:
FYI I'm done with my review of the spec for this PR. All changes should now be included and rebased on `main`. Next, I'll go through all the changes and pull out anything not identity/comparison related into separate PRs, then I'll mark this PR as ready for review.

spec.html
```diff
@@ -17942,7 +17946,7 @@ <h1>Static Semantics: Early Errors</h1>
           It is a Syntax Error if the goal symbol of the syntactic grammar is |Module| and the StringValue of |IdentifierName| is *"await"*.
         </li>
         <li>
-          It is a Syntax Error if StringValue of |IdentifierName| is the same String value as the StringValue of any |ReservedWord| except for `yield` or `await`.
+          It is a Syntax Error if ther StringValue of |IdentifierName| is the StringValue of any |ReservedWord| except for `yield` or `await`.
```
michaelficarra, line 17949: Reminder to self

```suggestion
          It is a Syntax Error if the StringValue of |IdentifierName| is the StringValue of any |ReservedWord| except for `yield` or `await`.
```

spec.html
```diff
@@ -30581,7 +30587,7 @@ <h1>Symbol.prototype [ @@toPrimitive ] ( _hint_ )</h1>
 
       <emu-clause id="sec-symbol.prototype-@@tostringtag">
         <h1>Symbol.prototype [ @@toStringTag ]</h1>
-        <p>The initial value of the @@toStringTag property is the String value *"Symbol"*.</p>
+        <p>The initial value of the @@toStringTag property is *"Symbol"*.</p>
```
michaelficarra, line 30590: Not to self: put this back.

```suggestion
        <p>The initial value of the @@toStringTag property is the String value *"Symbol"*.</p>
```

spec.html
```diff
@@ -34662,7 +34668,7 @@ <h1>Static Semantics: Early Errors</h1>
         <emu-grammar>QuantifierPrefix :: `{` DecimalDigits `,` DecimalDigits `}`</emu-grammar>
         <ul>
           <li>
-            It is a Syntax Error if the MV of the first |DecimalDigits| is larger than the MV of the second |DecimalDigits|.
+            It is a Syntax Error if the MV of the first |DecimalDigits| is greater than the MV of the second |DecimalDigits|.
```
michaelficarra, line 34671: Note to self: I think I missed some of these.

spec.html
```diff
@@ -39386,19 +39384,11 @@ <h1>%TypedArray%.prototype.slice ( _start_, _end_ )</h1>
...
-            1. If _srcType_ is different from _targetType_, then
-              1. Let _n_ be 0.
-              1. Repeat, while _k_ &lt; _final_,
-                1. Let _Pk_ be ! ToString(𝔽(_k_)).
-                1. Let _kValue_ be ! Get(_O_, _Pk_).
-                1. Perform ! Set(_A_, ! ToString(𝔽(_n_)), _kValue_, *true*).
-                1. Set _k_ to _k_ + 1.
-                1. Set _n_ to _n_ + 1.
-            1. Else,
+            1. If _srcType_ is _targetType_, then
```
michaelficarra, line 39387: Note to reviewers: I'm just swapping the positive/negative branches here.

spec.html
```diff
@@ -39577,7 +39575,7 @@ <h1>
         <emu-alg>
           1. Let _newTypedArray_ be ? Construct(_constructor_, _argumentList_).
           1. Perform ? ValidateTypedArray(_newTypedArray_).
-          1. If _argumentList_ is a List of a single Number, then
+          1. If the number of elements in _argumentList_ = 1 and Type(_argumentList_[0]) is Number, then
```
michaelficarra, line 39578:
```suggestion
          1. If the number of elements in _argumentList_ = 1 and _argumentList_[0] is a Number, then
```

michaelficarra:
Okay that was actually easier than I expected. This is ready for review.

michaelficarra:
resolved conflicts

spec.html
```diff
@@ -35357,7 +35363,7 @@ <h1>
             1. Assert: _c_ is a Continuation.
             1. Let _Input_ be _x_'s _input_.
             1. Let _e_ be _x_'s _endIndex_.
-            1. If _e_ = 0, or if _rer_.[[Multiline]] is *true* and the character _Input_[_e_ - 1] is one of |LineTerminator|, then
+            1. If _e_ = 0 or _rer_.[[Multiline]] is *true* and the character _Input_[_e_ - 1] is matched by |LineTerminator|, then
```
bakkot, line 35366:
```suggestion
            1. If _e_ = 0, or if _rer_.[[Multiline]] is *true* and the character _Input_[_e_ - 1] is matched by |LineTerminator|, then
```
In this case "or if" isn't just easier to read, it's also less ambiguous (English does not have precedence rules for "or" vs "and", so "if x or y and z" is always ambiguous).

Ditto several places below.

bakkot (commented):
Some comments. Big thing is that I think changing "if x, or if y, then" to "if x or y, then" is a hit to readability in a number of cases; I'd prefer we continue to use both forms as seems appropriate rather than trying to be entirely consistent.

spec.html
```diff
@@ -18504,7 +18508,7 @@ <h1>
           <dd>It determines if its argument is a valid regular expression literal.</dd>
         </dl>
         <emu-alg>
-          1. If FlagText of _literal_ contains any code points other than `d`, `g`, `i`, `m`, `s`, `u`, or `y`, or if it contains the same code point more than once, return *false*.
+          1. If FlagText of _literal_ contains any code points other than `d`, `g`, `i`, `m`, `s`, `u`, or `y`, or _literal_ contains any code point more than once, return *false*.
```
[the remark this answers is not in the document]
gibson042: I agree, and the proposed change doesn't even appear to be correct right now because the part after "or" references "_literal_" directly rather than "FlagText of _literal_".
```suggestion
          1. Let _flags_ be FlagText of _literal_.
          1. If _flags_ contains any code point other than `d`, `g`, `i`, `m`, `s`, `u`, or `y`, or if _flags_ contains any duplicate code points, return *false*.
```

jmdyck:
Even after merging my suggestions (thanks!), there's still some inconsistency in the forms that this PR addresses. Here are some counts I get when parsing pseudocode.

----

Composing a condition from sub-conditions...

... when there are only 2 sub-conditions:
```
  229  {COND} and {COND}
   80  {COND} or {COND}
    4  {COND} unless {COND}
vs
    3  {COND}, or {COND}
    1  {COND}, unless {COND}
vs
    1  either {COND} or {COND}
```

... when there are 3+ sub-conditions:
```
    1  {COND}, {COND}, or {COND}
    8  {COND}, {COND}, and {COND}
    6  {COND}, {COND}, {COND}, and {COND}
vs
    5  {COND} and {COND} and {COND}
    1  {COND} or {COND} or {COND}
    1  {COND} or {COND} or {COND} or {COND}
```

(And while I'm here, conditions that mix operators without parens:
```
    1  {COND} and {COND} or {COND} and {COND}
    6  {COND} and {COND}, or {COND} and {COND}
    4  {COND} or {COND} and {COND}
    1  {COND} unless {COND} and {COND}
```
)

----

Testing whether or not a value matches some list of descriptions (e.g., `_x_ is *undefined* or a String`)...

... when there are only 2 descriptions:
```
  189  {D} or {D}
vs
  373  either {D} or {D}
vs
   25  either {D}, or {D}
```

... when there are 3+ descriptions:
```
   11  any of {D}, {D}, or {D}
    7  any of {D}, {D}, {D}, or {D}
    4  any of {D}, {D}, {D}, {D}, or {D}
    1  any of {D}, {D}, {D}, {D}, {D}, or {D}
    1  any of {D}, {D}, {D}, {D}, {D}, {D}, {D}, or {D}
vs
    6  either {D}, {D}, or {D}
    4  either {D}, {D}, {D}, or {D}
vs
    3  one of {D}, {D}, or {D}
    2  one of {D}, {D}, {D}, or {D}
vs
   31  {D}, {D}, or {D}
    6  {D}, {D}, {D}, or {D}
    2  {D}, {D}, {D}, {D}, or {D}
    2  {D}, {D}, {D}, {D}, {D}, or {D}
    1  {D}, {D}, {D}, {D}, {D}, {D}, {D}, {D}, {D}, {D}, {D}, or {D}
```

spec.html
```diff
@@ -30304,7 +30310,7 @@ <h1>Symbol ( [ _description_ ] )</h1>
           1. If NewTarget is not *undefined*, throw a *TypeError* exception.
           1. If _description_ is *undefined*, let _descString_ be *undefined*.
           1. Else, let _descString_ be ? ToString(_description_).
-          1. Return a new unique Symbol value whose [[Description]] value is _descString_.
+          1. Return a new Symbol whose [[Description]] is _descString_.
```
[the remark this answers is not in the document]
michaelficarra: I think "unique" was harmful here.

spec.html
```diff
@@ -36042,7 +36048,7 @@ <h1>
           1. Else, let _P_ be ? ToString(_pattern_).
           1. If _flags_ is *undefined*, let _F_ be the empty String.
           1. Else, let _F_ be ? ToString(_flags_).
-          1. If _F_ contains any code unit other than *"d"*, *"g"*, *"i"*, *"m"*, *"s"*, *"u"*, or *"y"* or if it contains the same code unit more than once, throw a *SyntaxError* exception.
+          1. If _F_ contains any code unit other than *"d"*, *"g"*, *"i"*, *"m"*, *"s"*, *"u"*, or *"y"*, or _F_ contains the same code unit more than once, throw a *SyntaxError* exception.
```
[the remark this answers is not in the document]
michaelficarra: I don't think this reads better and I don't understand why you think that. Is it because `F` is used in both cases?

spec.html
```diff
@@ -39546,7 +39544,7 @@ <h1>
         <emu-alg>
           1. Let _newTypedArray_ be ? Construct(_constructor_, _argumentList_).
           1. Perform ? ValidateTypedArray(_newTypedArray_).
-          1. If _argumentList_ is a List of a single Number, then
+          1. If the number of elements in _argumentList_ = 1 and _argumentList_[0] is a Number, then
```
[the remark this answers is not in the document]
michaelficarra: We've decided to make an exception to the `is`/`=` guidance for this specific phrasing.

spec.html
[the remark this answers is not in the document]
bakkot: It's an easier-to-identify separator between long subclauses. To know that "or" is starting a new subclause you have to do nonlocal mental parsing (e.g. the "or" two words prior did _not_ start a new subclause), whereas with "or if" it's unambiguous.

spec.html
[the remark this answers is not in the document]
bakkot: Why?

michaelficarra:
@bakkot Comments addressed.

spec.html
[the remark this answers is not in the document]
michaelficarra: Because symbols are *all* unique: they have identity. Using "unique" here implies that there could be a symbol that is *not* unique.

spec.html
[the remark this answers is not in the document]
michaelficarra: Okay, I see that now.

spec.html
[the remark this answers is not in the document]
ljharb: fwiw i definitely don't agree it's harmful nor has that implication - however, i think the word "new" makes "unique" redundant, so it's fine to remove it.

spec.html
```diff
@@ -31668,7 +31673,7 @@ <h1>Overview of Date Objects and Definitions of Abstract Operations</h1>
...
         <p>Time values do not account for UTC leap seconds—there are no time values representing instants within positive leap seconds, and there are time values representing instants removed from the UTC timeline by negative leap seconds. However, the definition of time values nonetheless yields piecewise alignment with UTC, with discontinuities only at leap second boundaries and zero difference outside of leap seconds.</p>
         <p>A Number can exactly represent all integers from -9,007,199,254,740,992 to 9,007,199,254,740,992 (<emu-xref href="#sec-number.min_safe_integer"></emu-xref> and <emu-xref href="#sec-number.max_safe_integer"></emu-xref>). A time value supports a slightly smaller range of -8,640,000,000,000,000 to 8,640,000,000,000,000 milliseconds. This yields a supported time value range of exactly -100,000,000 days to 100,000,000 days relative to midnight at the beginning of 1 January 1970 UTC.</p>
         <p>The exact moment of midnight at the beginning of 1 January 1970 UTC is represented by the time value *+0*<sub>𝔽</sub>.</p>
```
gibson042, line 31679: Assuming «_use `is`, `is not`, `<`, `≤`, `>`, and `≥` for number comparisons_» is still a goal of this PR, it looks like some adjustment is necessary in InLeapYear and DateFromTime.
```patch
@@ -31706,8 +31706,8 @@
         <p>The leap-year function is *1*<sub>𝔽</sub> for a time within a leap year and otherwise is
 *+0*<sub>𝔽</sub>:</p>
         <emu-eqn id="eqn-InLeapYear" aoid="InLeapYear">
           InLeapYear(_t_)
-            = *+0*<sub>𝔽</sub> if DaysInYear(YearFromTime(_t_)) = *365*<sub>𝔽</sub>
-            = *1*<sub>𝔽</sub> if DaysInYear(YearFromTime(_t_)) = *366*<sub>𝔽</sub>
+            = *+0*<sub>𝔽</sub> if DaysInYear(YearFromTime(_t_)) is *365*<sub>𝔽</sub>
+            = *1*<sub>𝔽</sub> if DaysInYear(YearFromTime(_t_)) is *366*<sub>𝔽</sub>
         </emu-eqn>
       </emu-clause>
 
```
```patch
@@ -31739,18 +31739,18 @@
         <p>A date number is identified by an integral Number in the inclusive interval from *1*<sub
>𝔽</sub> to *31*<sub>𝔽</sub>. The mapping DateFromTime(_t_) from a time value _t_ to a date number i
s defined by:</p>
         <emu-eqn aoid="DateFromTime">
           DateFromTime(_t_)
-            = DayWithinYear(_t_) + *1*<sub>𝔽</sub> if MonthFromTime(_t_) = *+0*<sub>𝔽</sub>
-            = DayWithinYear(_t_) - *30*<sub>𝔽</sub> if MonthFromTime(_t_) = *1*<sub>𝔽</sub>
-            = DayWithinYear(_t_) - *58*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *2*<s
ub>𝔽</sub>
-            = DayWithinYear(_t_) - *89*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *3*<s
ub>𝔽</sub>
-            = DayWithinYear(_t_) - *119*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *4*<
sub>𝔽</sub>
-            = DayWithinYear(_t_) - *150*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *5*<
sub>𝔽</sub>
-            = DayWithinYear(_t_) - *180*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *6*<
sub>𝔽</sub>
-            = DayWithinYear(_t_) - *211*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *7*<
sub>𝔽</sub>
-            = DayWithinYear(_t_) - *242*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *8*<sub>𝔽</sub>
-            = DayWithinYear(_t_) - *272*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *9*<sub>𝔽</sub>
-            = DayWithinYear(_t_) - *303*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *10*<sub>𝔽</sub>
-            = DayWithinYear(_t_) - *333*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) = *11*<sub>𝔽</sub>
+            = DayWithinYear(_t_) + *1*<sub>𝔽</sub> if MonthFromTime(_t_) is *+0*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *30*<sub>𝔽</sub> if MonthFromTime(_t_) is *1*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *58*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *2*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *89*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *3*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *119*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *4*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *150*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *5*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *180*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *6*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *211*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *7*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *242*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *8*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *272*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *9*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *303*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *10*<sub>𝔽</sub>
+            = DayWithinYear(_t_) - *333*<sub>𝔽</sub> - InLeapYear(_t_) if MonthFromTime(_t_) is *11*<sub>𝔽</sub>
         </emu-eqn>
       </emu-clause>
 
```

gibson042 (approved):
Are there plans to document and/or enforce these conventions?

spec.html
```diff
@@ -18286,7 +18289,7 @@ <h1>
           <dd>It determines if its argument is a valid regular expression literal.</dd>
         </dl>
         <emu-alg>
-          1. If FlagText of _literal_ contains any code points other than `d`, `g`, `i`, `m`, `s`, `u`, or `y`, or if it contains the same code point more than once, return *false*.
+          1. If FlagText of _literal_ contains any code points other than `d`, `g`, `i`, `m`, `s`, `u`, or `y`, or FlagText of _literal_ contains any code point more than once, return *false*.
           1. Let _patternText_ be BodyText of _literal_.
           1. If FlagText of _literal_ contains `u`, let _u_ be *true*; else let _u_ be *false*.
```
gibson042, line 18294: Readability suggestion:
```suggestion
          1. Let _flags_ be FlagText of _literal_.
          1. If _flags_ contains any code points other than `d`, `g`, `i`, `m`, `s`, `u`, or `y`, or _flags_ contains any code point more than once, return *false*.
          1. If _flags_ contains `u`, let _u_ be *true*; else let _u_ be *false*.
          1. Let _patternText_ be BodyText of _literal_.
```

spec.html
michaelficarra: Good catch.

bakkot (approved):

bakkot:
@syg may not care to review again, since we've talked about all the decisions, but we should check before landing.

jmdyck:
Re-doing my analysis from Oct 17...

----

Composing a condition from sub-conditions...

... when there are only 2 sub-conditions:
```
  229  {COND} and {COND}
   80  {COND} or {COND}
    4  {COND} unless {COND}
vs
    2  {COND}, or if {COND}
    1  {COND}, or {COND}
    1  {COND}, unless {COND}
vs
    1  either {COND} or {COND}
```

... when there are 3+ sub-conditions:
```
    1  {COND}, {COND}, or {COND}
    8  {COND}, {COND}, and {COND}
    6  {COND}, {COND}, {COND}, and {COND}
vs
    5  {COND} and {COND} and {COND}
    1  {COND} or {COND} or {COND}
    1  {COND} or {COND} or {COND} or {COND}
```

(And while I'm here, conditions that mix operators without parens:
```
    1  {COND} and {COND} or {COND} and {COND}
    3  {COND} and {COND}, or if {COND} and {COND}
    2  {COND} and {COND}, or {COND} and {COND}
    2  {COND} or {COND} <ins>and {COND}</ins>
    1  {COND} unless {COND} and {COND}
    2  {COND}, or if {COND} and {COND}
```
)

----

Testing whether or not a value matches some list of descriptions (e.g., `_x_ is *undefined* or a String`)...

... when there are only 2 descriptions:
```
  189  {D} or {D}
vs
  376  either {D} or {D}
vs
   26  either {D}, or {D}
```

... when there are 3+ descriptions:
```
    8  any of {D}, {D}, or {D}
    6  any of {D}, {D}, {D}, or {D}
    4  any of {D}, {D}, {D}, {D}, or {D}
    1  any of {D}, {D}, {D}, {D}, {D}, or {D}
    1  any of {D}, {D}, {D}, {D}, {D}, {D}, or {D}
    1  any of {D}, {D}, {D}, {D}, {D}, {D}, {D}, or {D}
vs
    6  either {D}, {D}, or {D}
    5  either {D}, {D}, {D}, or {D}
vs
    3  one of {D}, {D}, or {D}
    1  one of {D}, {D}, {D}, or {D}
vs
   33  {D}, {D}, or {D}
    7  {D}, {D}, {D}, or {D}
    2  {D}, {D}, {D}, {D}, or {D}
    2  {D}, {D}, {D}, {D}, {D}, or {D}
    1  {D}, {D}, {D}, {D}, {D}, {D}, {D}, {D}, {D}, {D}, {D}, or {D}
```

jmdyck (commented):
Some things I didn't put into PR #2970.

spec.html
```diff
@@ -11398,13 +11402,12 @@ <h1>
...
-          1. If _names_ contains a Private Name whose [[Description]] is _identifier_, then
-            1. Let _name_ be that Private Name.
-            1. Return _name_.
-          1. Else,
-            1. Let _outerPrivEnv_ be _privEnv_.[[OuterPrivateEnvironment]].
-            1. Assert: _outerPrivEnv_ is not *null*.
-            1. Return ResolvePrivateIdentifier(_outerPrivEnv_, _identifier_).
+          1. For each Private Name _pn_ of _names_, do
+            1. If _pn_.[[Description]] is _identifier_, then
+              1. Return _pn_.
```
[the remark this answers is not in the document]

bakkot:
I don't love the `ToIntegerOrInfinity` -> `ToUint32` change, because `ToUint32` has wrapping behavior. It happens that the wrapping behavior doesn't come up in this case because array indices are bounded below the modulus, but I don't like making the reader think that hard.

Can I suggest `! ToNumber(_P_) ≥ _newLen_` instead? I would also be fine with `StringToNumber(_P_) ≥ _newLen_`, or with `! ToIntegerOrInfinity(_P_) ≥ ℝ(_newLen_)`.

syg (approved):

PR #2878, opened 2022-08-31
Editorial: refactor String.fromCharCode to build a String directly

michaelficarra:
I think this is simpler, more idiomatic, and reads better.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -33456,11 +33456,11 @@ THH:mm:ss.sss
         <p>This function may be called with any number of arguments which form the rest parameter _codeUnits_.</p>
         <p>It performs the following steps when called:</p>
         <emu-alg>
-          1. Let _elements_ be a new empty List.
+          1. Let _result_ be the empty String.
           1. For each element _next_ of _codeUnits_, do
-            1. Let _nextCU_ be ℝ(? ToUint16(_next_)).
-            1. Append _nextCU_ to _elements_.
-          1. Return the String value whose code units are the elements in the List _elements_. If _codeUnits_ is empty, the empty String is returned.
+            1. Let _nextCU_ be the code unit whose numeric value is ℝ(? ToUint16(_next_)).
+            1. Set _result_ to the string-concatenation of _result_ and _nextCU_.
+          1. Return _result_.
         </emu-alg>
         <p>The *"length"* property of this function is *1*<sub>𝔽</sub>.</p>
       </emu-clause>


spec.html
```diff
@@ -33492,11 +33492,11 @@ <h1>String.fromCharCode ( ..._codeUnits_ )</h1>
...
         <p>It performs the following steps when called:</p>
         <emu-alg>
-          1. Let _elements_ be a new empty List.
+          1. Let _result_ be the empty String.
           1. For each element _next_ of _codeUnits_, do
-            1. Let _nextCU_ be ℝ(? ToUint16(_next_)).
-            1. Append _nextCU_ to _elements_.
-          1. Return the String value whose code units are the elements in the List _elements_. If _codeUnits_ is empty, the empty String is returned.
+            1. Let _nextCU_ be the code unit whose numeric value is ℝ(? ToUint16(_next_)).
+            1. Append _nextCU_ to _result_.
```
bakkot, line 33498: I'm a little suspicious of this usage of "append". It's supposed to update _the alias_ `result`, rather than mutating the non-mutable String value which the alias refers to. But we use exactly the same syntax - "append v to list" - to _mutate_ lists.

ljharb: I agree; append means mutate, not “concatenate and reassign”

michaelficarra: Oops, you're right.

spec.html
```diff
@@ -33492,11 +33492,11 @@ <h1>String.fromCharCode ( ..._codeUnits_ )</h1>
...
         <p>It performs the following steps when called:</p>
         <emu-alg>
-          1. Let _elements_ be a new empty List.
+          1. Let _result_ be the empty String.
           1. For each element _next_ of _codeUnits_, do
-            1. Let _nextCU_ be ℝ(? ToUint16(_next_)).
-            1. Append _nextCU_ to _elements_.
-          1. Return the String value whose code units are the elements in the List _elements_. If _codeUnits_ is empty, the empty String is returned.
+            1. Let _nextCU_ be the code unit whose numeric value is ℝ(? ToUint16(_next_)).
+            1. Set _result_ to the result of appending _nextCU_ to _result_.
```
bakkot, line 33498: [`string-concatenation`](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#string-concatenation) actually allows you to concatenate individual code units, which is a thing we do elsewhere. Seems like that would be best?

```suggestion
            1. Set _result_ to the result of the string-concatenation of _result_ and _nextCU_.
```

gibson042 (approved):

spec.html
michaelficarra: I did not realise string-concatenation was defined for non-sequences. I guess we should use it here too? https://github.com/tc39/proposal-is-usv-string/pull/20/files#diff-181371b08d71216599b0acccbaabd03c306da6de142ea6275c2135810999805aR39

ljharb: https://github.com/tc39/proposal-is-usv-string/pull/20/files#r956319959 seems to contradict  https://github.com/tc39/ecma262/pull/2878#discussion_r959111556

bakkot (approved):

PR #2879, opened 2022-08-31
Editorial: remove some unnecessary complexity due to the SourceText AO

michaelficarra:
This conversion to a List wasn't helping anything. It's still not perfect, but I think it's much better.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -833,7 +833,7 @@
 
     <emu-clause id="sec-algorithm-conventions-syntax-directed-operations">
       <h1>Syntax-Directed Operations</h1>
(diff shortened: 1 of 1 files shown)


jmdyck:
Yeah, the SourceText SDO is unnecessary.

If you want to preserve `sec-static-semantics-sourcetext` as an oldid, it might make sense to attach it to [11.1 Source Text](https://tc39.es/ecma262/#sec-source-text).

michaelficarra:
I was hoping we could get away with not preserving that old ID.

bakkot (approved):

PR #2880, opened 2022-08-31
Editorial: dfn the Pattern data structures

michaelficarra:
I also updated the Notation section's ID since it was *way* under-specific.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -35071,27 +35071,27 @@ THH:mm:ss.sss
         <p>An implementation may not actually perform such translations to or from UTF-16, but the semantics of this specification requires that the result of pattern matching be as if such translations were performed.</p>
       </emu-note>
 
-      <emu-clause id="sec-notation">
+      <emu-clause id="sec-pattern-notation" oldids="sec-notation">
         <h1>Notation</h1>
         <p>The descriptions below use the following internal data structures:</p>
         <ul>
           <li>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -35151,27 +35151,27 @@ <h1>Pattern Semantics</h1>
...
+            A <dfn id="pattern-state" variants="States">State</dfn> is an ordered triple (_input_, _endIndex_, _captures_) where _input_ is a List of characters representing the String being matched, _endIndex_ is an integer, and _captures_ is a List of values, one for each left-capturing parenthesis in the pattern. States are used to represent partial match states in the regular expression matching algorithms. The _endIndex_ is one plus the index of the last input character matched so far by the pattern, while _captures_ holds the results of capturing parentheses. The _n_<sup>th</sup> element of _captures_ is either a Range representing the range of characters captured by the _n_<sup>th</sup> set of capturing parentheses, or *undefined* if the _n_<sup>th</sup> set of capturing parentheses hasn't been reached yet. Due to backtracking, many States may be in use at any time during the matching process.
```
bakkot, line 35165: Can we rename this to `MatchState` or something, instead of needing to emu-not-ref the uses of the word "State"?

bakkot:
cf #2486.

spec.html
michaelficarra: @bakkot Done.

bakkot (approved):

syg (commented):
I like the MatchState change. I have some concern over Range and Continuation being too generic to be dfn'd here. CaptureRange and MatcherContinuation maybe?

michaelficarra:
@syg Renamed.

syg (approved):

PR #2882, opened 2022-08-31
Editorial: turn "no action is required" step in Array reverse into note

michaelficarra:
Alternatively we could make it a note, but I don't think it's needed given the assert.


bakkot:
Eh, I kind of want a note; it's not totally obvious why the assertion implies no action is required. It's easy enough to figure that out, but comments are free.

michaelficarra:
@bakkot Alright, done.

bakkot (approved):

PR #2883, opened 2022-09-01
Editorial: in NumericToRawBytes, consolidate byte ordering decision

michaelficarra:
This mirrors what we do in the complementary AO, `RawBytesToNumeric`.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -41030,17 +41030,18 @@ THH:mm:ss.sss
         </dl>
         <emu-alg>
           1. If _type_ is ~Float32~, then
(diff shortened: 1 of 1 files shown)


bakkot (approved):
I am OK with this but honestly I'd prefer to go the other way, and un-consolidate the ordering decision in `RawBytesToNumeric`. Either way I'd defer to @syg on this.

syg (approved):
I do prefer doing the order reversing once instead of multiple times, so this PR lgtm.

PR #2884, opened 2022-09-01
Editorial: Add note that roundtripping ToIntegerOrInfinity never produces -0

syg:
A V8 engineer was confused about this today, figured it doesn't hurt to add a note.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -5160,6 +5160,9 @@
         1. If _number_ &lt; *-0*<sub>𝔽</sub>, set _integer_ to -_integer_.
         1. Return _integer_.
       </emu-alg>
+      <emu-note>
+        𝔽(ToIntegerOrInfinity(_x_)) never returns *-0*<sub>𝔽</sub> for any value of _x_. The truncation of the fractional part is performed after converting _x_ to a mathematical value.
+      </emu-note>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -5160,6 +5160,9 @@ <h1>
         1. If _number_ &lt; *-0*<sub>𝔽</sub>, set _integer_ to -_integer_.
         1. Return _integer_.
       </emu-alg>
+      <emu-note>
+        𝔽(ToIntegerOrInfinity(_x_)) never returns *-0*<sub>𝔽</sub> for all values of _x_. The truncation of the fractional part is performed after converting _x_ to a mathematical value.
```
bakkot, line 5164:
```suggestion
        𝔽(ToIntegerOrInfinity(_x_)) never returns *-0*<sub>𝔽</sub> for any value of _x_. The truncation of the fractional part is performed after converting _x_ to a mathematical value.
```

rkirsling: (Good ol' "negative polarity items"...)

syg: Done.

michaelficarra (approved):

bakkot (approved):

PR #2905, opened 2022-09-15
Layering: Add HostLoadImportedModule hook

nicolo-ribaudo:
This PR also removes the HostResolveImportedModule and HostImportModuleDynamically hooks.

A few comments/questions:
1. ContinueDynamicImport is kinda messy, with all the closures and PerformPromiseThen. Is it ok to extend AsyncBlockStart like in https://tc39.es/proposal-array-from-async/#sec-asyncblockstart so that I can use the Await macro?
1. ResolveExport is described as returning *null* or something, but it can also return a throw completion (if it's called on a module that re-exports from a dependency that hasn't been already loaded - **NOTE:** this can only happen if a host calls ResolveExport() even if LoadRequestedModules() doesn't succeed, and they don't have any good reason to do so). This was the behavior even before this PR.
   - Is it ok?
   - Should we make it return null?
   - Should we require that it's called only after having loaded the dependencies?
1. I named the new Module Records method LoadRequestedModules() to match the name of the [[RequestedModules]] slot, but maybe LoadImportedModules() or LoadDependencies() are better?
1. I find the _hostDefined_ parameter of LoadRequestedModules() to be incredibly ugly, but it's needed by HTML to pass some context only to modules imported in "the current loading process" and not in "future loading processes" (the `.destination` flag it sets on the `Request` object it creates to fetch modules). It's possible that one day HTML will align requests of static and dynamic imports, but today is not that day yet.
1. I tracked the loading state across the ecma262-host-ecma262 round trip using a `ModuleLoadState` record, which is overloaded for two usages that are distinguished using an [[Action]] enum field. Is it ok, or should I split it into two different record types (`GraphLoadingState` and `DynamicImportState`)? (related: https://github.com/tc39/ecma262/pull/2905#discussion_r988214326)

The HTML PR is https://github.com/whatwg/html/pull/8253. Today's consensus was based on the fact that this refactor is purely editorial for the ECMA-262+(insert known host here) pair, so I would like to get a full review on the HTML side to check if we need any tweaks before moving on with this PR. **EDIT:** Got a :+1: from the HTML side.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -11474,6 +11474,20 @@
         </tr>
         <tr>
           <td>
+            [[LoadedModules]]
+          </td>
+          <td>
+            a List of Records with fields [[Specifier]] (a String) and [[Module]] (a Module Record)
+          </td>
+          <td>
+            <p>A map from the specifier strings imported by this realm to the resolved Module Record. The list does not contain two different Records with the same [[Specifier]].</p>
+            <emu-note>
+              As mentioned in HostLoadImportedModule (<emu-xref href="#note-HostLoadImportedModule-referrer-Realm-Record"></emu-xref>), [[LoadedModules]] in Realm Records is only used when running an `import()` expression in a context where there is no active script or module.
+            </emu-note>
+          </td>
+        </tr>
+        <tr>
+          <td>
             [[HostDefined]]
           </td>
           <td>
@@ -14756,12 +14770,12 @@
           1. Let _exports_ be _O_.[[Exports]].
           1. If _P_ is not an element of _exports_, return *undefined*.
           1. Let _m_ be _O_.[[Module]].
-          1. Let _binding_ be ! _m_.ResolveExport(_P_).
+          1. Let _binding_ be _m_.ResolveExport(_P_).
           1. Assert: _binding_ is a ResolvedBinding Record.
           1. Let _targetModule_ be _binding_.[[Module]].
           1. Assert: _targetModule_ is not *undefined*.
           1. If _binding_.[[BindingName]] is ~namespace~, then
-            1. Return ? GetModuleNamespace(_targetModule_).
+            1. Return GetModuleNamespace(_targetModule_).
           1. Let _targetEnv_ be _targetModule_.[[Environment]].
           1. If _targetEnv_ is ~empty~, throw a *ReferenceError* exception.
           1. Return ? _targetEnv_.GetBindingValue(_binding_.[[BindingName]], *true*).
@@ -19201,15 +19215,56 @@
 
         <emu-grammar>ImportCall : `import` `(` AssignmentExpression `)`</emu-grammar>
         <emu-alg>
-          1. Let _referencingScriptOrModule_ be GetActiveScriptOrModule().
+          1. Let _referrer_ be GetActiveScriptOrModule().
+          1. If _referrer_ is *null*, set _referrer_ to the current Realm Record.
           1. Let _argRef_ be ? Evaluation of |AssignmentExpression|.
           1. Let _specifier_ be ? GetValue(_argRef_).
           1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
           1. Let _specifierString_ be Completion(ToString(_specifier_)).
           1. IfAbruptRejectPromise(_specifierString_, _promiseCapability_).
-          1. Perform HostImportModuleDynamically(_referencingScriptOrModule_, _specifierString_, _promiseCapability_).
+          1. Perform HostLoadImportedModule(_referrer_, _specifierString_, ~empty~, _promiseCapability_).
           1. Return _promiseCapability_.[[Promise]].
         </emu-alg>
+
+        <emu-clause id="sec-ContinueDynamicImport" type="abstract operation">
+          <h1>
+            ContinueDynamicImport (
+              _promiseCapability_: a PromiseCapability Record,
+              _moduleCompletion_: either a normal completion containing a Module Record or a throw completion,
+            ): ~unused~
+          </h1>
+          <dl class="header">
+            <dt>description</dt>
+            <dd>It completes the process of a dynamic import originally started by an <emu-xref href="#sec-import-calls">`import()`</emu-xref> call, resolving or rejecting the promise returned by that call as appropriate.</dd>
+          </dl>
+          <emu-alg>
+            1. If _moduleCompletion_ is an abrupt completion, then
+              1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, « _moduleCompletion_.[[Value]] »).
+              1. Return ~unused~.
+            1. Let _module_ be _moduleCompletion_.[[Value]].
+            1. Let _loadPromise_ be _module_.LoadRequestedModules().
+            1. Let _rejectedClosure_ be a new Abstract Closure with parameters (_reason_) that captures _promiseCapability_ and performs the following steps when called:
+              1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, « _reason_ »).
+              1. Return ~unused~.
+            1. Let _onRejected_ be CreateBuiltinFunction(_rejectedClosure_, 1, *""*, « »).
+            1. Let _linkAndEvaluateClosure_ be a new Abstract Closure with no parameters that captures _module_, _promiseCapability_, and _onRejected_ and performs the following steps when called:
+              1. Let _link_ be Completion(_module_.Link()).
+              1. If _link_ is an abrupt completion, then
+                1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, « _link_.[[Value]] »).
+                1. Return ~unused~.
+              1. Let _evaluatePromise_ be _module_.Evaluate().
+              1. Let _fulfilledClosure_ be a new Abstract Closure with no parameters that captures _module_ and _promiseCapability_ and performs the following steps when called:
+                1. Let _namespace_ be GetModuleNamespace(_module_).
+                1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, « _namespace_ »).
+                1. Return ~unused~.
+              1. Let _onFulfilled_ be CreateBuiltinFunction(_fulfilledClosure_, 0, *""*, « »).
+              1. Perform PerformPromiseThen(_evaluatePromise_, _onFulfilled_, _onRejected_).
+              1. Return ~unused~.
+            1. Let _linkAndEvaluate_ be CreateBuiltinFunction(_linkAndEvaluateClosure_, 0, *""*, « »).
+            1. Perform PerformPromiseThen(_loadPromise_, _linkAndEvaluate_, _onRejected_).
+            1. Return ~unused~.
+          </emu-alg>
+        </emu-clause>
       </emu-clause>
     </emu-clause>
 
@@ -25487,6 +25542,17 @@
           </tr>
           <tr>
             <td>
+              [[LoadedModules]]
+            </td>
+            <td>
+              a List of Records with fields [[Specifier]] (a String) and [[Module]] (a Module Record)
+            </td>
+            <td>
+              A map from the specifier strings imported by this script to the resolved Module Record. The list does not contain two different Records with the same [[Specifier]].
+            </td>
+          </tr>
+          <tr>
+            <td>
               [[HostDefined]]
             </td>
             <td>
@@ -25516,7 +25582,7 @@
       <emu-alg>
         1. Let _script_ be ParseText(_sourceText_, |Script|).
         1. If _script_ is a List of errors, return _script_.
-        1. Return Script Record { [[Realm]]: _realm_, [[ECMAScriptCode]]: _script_, [[HostDefined]]: _hostDefined_ }.
+        1. Return Script Record { [[Realm]]: _realm_, [[ECMAScriptCode]]: _script_, [[LoadedModules]]: « », [[HostDefined]]: _hostDefined_ }.
       </emu-alg>
       <emu-note>
         <p>An implementation may parse script source text and analyse it for Early Error conditions prior to evaluation of ParseScript for that script source text. However, the reporting of any errors must be deferred until the point where this specification actually performs ParseScript upon that source text.</p>
@@ -25852,10 +25918,19 @@
             </tr>
             <tr>
               <td>
+                LoadRequestedModules( [ _hostDefined_ ] )
+              </td>
+              <td>
+                <p>Prepares the module for linking by recursively loading all its dependencies, and returns a promise.</p>
+              </td>
+            </tr>
+            <tr>
+              <td>
                 GetExportedNames([_exportStarSet_])
               </td>
               <td>
-                Return a list of all names that are either directly or indirectly exported from this module.
+                <p>Return a list of all names that are either directly or indirectly exported from this module.</p>
+                <p>LoadRequestedModules must have completed successfully prior to invoking this method.</p>
               </td>
             </tr>
             <tr>
@@ -25864,7 +25939,8 @@
               </td>
               <td>
                 <p>Return the binding of a name exported by this module. Bindings are represented by a <dfn id="resolvedbinding-record" variants="ResolvedBinding Records">ResolvedBinding Record</dfn>, of the form { [[Module]]: Module Record, [[BindingName]]: String | ~namespace~ }. If the export is a Module Namespace Object without a direct binding in any module, [[BindingName]] will be set to ~namespace~. Return *null* if the name cannot be resolved, or ~ambiguous~ if multiple bindings were found.</p>
-                <p>Each time this operation is called with a specific _exportName_, _resolveSet_ pair as arguments it must return the same result if it completes normally.</p>
+                <p>Each time this operation is called with a specific _exportName_, _resolveSet_ pair as arguments it must return the same result.</p>
+                <p>LoadRequestedModules must have completed successfully prior to invoking this method.</p>
               </td>
             </tr>
             <tr>
@@ -25873,6 +25949,7 @@
               </td>
               <td>
                 <p>Prepare the module for evaluation by transitively resolving all module dependencies and creating a Module Environment Record.</p>
+                <p>LoadRequestedModules must have completed successfully prior to invoking this method.</p>
               </td>
             </tr>
             <tr>
@@ -25910,10 +25987,10 @@
                 [[Status]]
               </td>
               <td>
-                ~unlinked~, ~linking~, ~linked~, ~evaluating~, ~evaluating-async~, or ~evaluated~
+                ~new~, ~unlinked~, ~linking~, ~linked~, ~evaluating~, ~evaluating-async~, or ~evaluated~
               </td>
               <td>
-                Initially ~unlinked~. Transitions to ~linking~, ~linked~, ~evaluating~, possibly ~evaluating-async~, ~evaluated~ (in that order) as the module progresses throughout its lifecycle. ~evaluating-async~ indicates this module is queued to execute on completion of its asynchronous dependencies or it is a module whose [[HasTLA]] field is *true* that has been executed and is pending top-level completion.
+                Initially ~new~. Transitions to ~unlinked~, ~linking~, ~linked~, ~evaluating~, possibly ~evaluating-async~, ~evaluated~ (in that order) as the module progresses throughout its lifecycle. ~evaluating-async~ indicates this module is queued to execute on completion of its asynchronous dependencies or it is a module whose [[HasTLA]] field is *true* that has been executed and is pending top-level completion.
               </td>
             </tr>
             <tr>
@@ -25958,6 +26035,17 @@
               </td>
               <td>
                 A List of all the |ModuleSpecifier| strings used by the module represented by this record to request the importation of a module. The List is in source text occurrence order.
+              </td>
+            </tr>
+            <tr>
+              <td>
+                [[LoadedModules]]
+              </td>
+              <td>
+                a List of Records with fields [[Specifier]] (a String) and [[Module]] (a Module Record)
+              </td>
+              <td>
+                A map from the specifier strings used by the module represented by this record to request the importation of a module to the resolved Module Record. The list does not contain two different Records with the same [[Specifier]].
               </td>
             </tr>
             <tr>
@@ -26058,6 +26146,166 @@
           </table>
         </emu-table>
 
+        <p>A <dfn id="graphloadingstate-record" variants="GraphLoadingState Records">GraphLoadingState Record</dfn> is a Record that contains information about the loading process of a module graph. It's used to continue loading after a call to HostLoadImportedModule. Each GraphLoadingState Record has the fields defined in <emu-xref href="#table-graphloadingstate-record-fields"></emu-xref>:</p>
+        <emu-table id="table-graphloadingstate-record-fields" caption="GraphLoadingState Record Fields">
+          <table>
+            <tr>
+              <th>
+                Field Name
+              </th>
+              <th>
+                Value Type
+              </th>
+              <th>
+                Meaning
+              </th>
+            </tr>
+            <tr>
+              <td>
+                [[PromiseCapability]]
+              </td>
+              <td>
+                a PromiseCapability Record
+              </td>
+              <td>
+                The promise to resolve when the loading process finishes.
+              </td>
+            </tr>
+            <tr>
+              <td>
+                [[IsLoading]]
+              </td>
+              <td>
+                a Boolean
+              </td>
+              <td>
+                It is true if the loading process has not finished yet, neither successfully nor with an error.
+              </td>
+            </tr>
+            <tr>
+              <td>
+                [[PendingModulesCount]]
+              </td>
+              <td>
+                a non-negative integer
+              </td>
+              <td>
+                It tracks the number of pending HostLoadImportedModule calls.
+              </td>
+            </tr>
+            <tr>
+              <td>
+                [[Visited]]
+              </td>
+              <td>
+                a List of Cyclic Module Records
+              </td>
+              <td>
+                It is a list of the Cyclic Module Records that have been already loaded by the current loading process, to avoid infinite loops with circular dependencies.
+              </td>
+            </tr>
+            <tr>
+              <td>
+                [[HostDefined]]
+              </td>
+              <td>
+                anything (default value is ~empty~)
+              </td>
+              <td>
+                It contains host-defined data to pass from the LoadRequestedModules caller to HostLoadImportedModule.
+              </td>
+            </tr>
+          </table>
+        </emu-table>
+
+        <emu-clause id="sec-LoadRequestedModules" type="concrete method">
+          <h1>
+            LoadRequestedModules (
+              optional _hostDefined_: anything,
+            ): a Promise
+          </h1>
+          <dl class="header">
+            <dt>for</dt>
+            <dd>a Cyclic Module Record _module_</dd>
+
+            <dt>description</dt>
+            <dd>It populates the [[LoadedModules]] of all the Module Records in the dependency graph of _module_ (most of the work is done by the auxiliary function InnerModuleLoading). It takes an optional _hostDefined_ parameter that is passed to the HostLoadImportedModule hook.</dd>
+          </dl>
+
+          <emu-alg>
+            1. If _hostDefined_ is not present, let _hostDefined_ be ~empty~.
+            1. Let _pc_ be ! NewPromiseCapability(%Promise%).
+            1. Let _state_ be the GraphLoadingState Record { [[IsLoading]]: *true*, [[PendingModulesCount]]: 1, [[Visited]]: « », [[PromiseCapability]]: _pc_, [[HostDefined]]: _hostDefined_ }.
+            1. Perform InnerModuleLoading(_state_, _module_).
+            1. Return _pc_.[[Promise]].
+          </emu-alg>
+
+          <emu-note>
+            The _hostDefined_ parameter can be used to pass additional information necessary to fetch the imported modules. It is used, for example, by HTML to set the correct fetch destination for <code>&lt;link rel="preload" as="..."&gt;</code> tags.
+            <code>import()</code> expressions never set the _hostDefined_ parameter.
+          </emu-note>
+
+          <emu-clause id="sec-InnerModuleLoading" type="abstract operation">
+            <h1>
+              InnerModuleLoading (
+                _state_: a GraphLoadingState Record,
+                _module_: a Module Record,
+              ): ~unused~
+            </h1>
+            <dl class="header">
+              <dt>description</dt>
+              <dd>It is used by LoadRequestedModules to recursively perform the actual loading process for _module_'s dependency graph.</dd>
+            </dl>
+
+            <emu-alg>
+              1. Assert: _state_.[[IsLoading]] is *true*.
+              1. If _module_ is a Cyclic Module Record, _module_.[[Status]] is ~new~, and _state_.[[Visited]] does not contain _module_, then
+                1. Append _module_ to _state_.[[Visited]].
+                1. Let _requestedModulesCount_ be the number of elements in _module_.[[RequestedModules]].
+                1. Set _state_.[[PendingModulesCount]] to _state_.[[PendingModulesCount]] + _requestedModulesCount_.
+                1. For each String _required_ of _module_.[[RequestedModules]], do
+                  1. If _module_.[[LoadedModules]] contains a Record whose [[Specifier]] is _required_, then
+                    1. Let _record_ be that Record.
+                    1. Perform InnerModuleLoading(_state_, _record_.[[Module]]).
+                  1. Else,
+                    1. Perform HostLoadImportedModule(_module_, _required_, _state_.[[HostDefined]], _state_).
+                    1. NOTE: HostLoadImportedModule will call FinishLoadingImportedModule, which re-enters the graph loading process through ContinueModuleLoading.
+                  1. If _state_.[[IsLoading]] is *false*, return ~unused~.
+              1. Assert: _state_.[[PendingModulesCount]] ≥ 1.
+              1. Set _state_.[[PendingModulesCount]] to _state_.[[PendingModulesCount]] - 1.
+              1. If _state_.[[PendingModulesCount]] = 0, then
+                1. Set _state_.[[IsLoading]] to *false*.
+                1. For each Cyclic Module Record _loaded_ of _state_.[[Visited]], do
+                  1. If _loaded_.[[Status]] is ~new~, set _loaded_.[[Status]] to ~unlinked~.
+                1. Perform ! Call(_state_.[[PromiseCapability]].[[Resolve]], *undefined*, « *undefined* »).
+              1. Return ~unused~.
+            </emu-alg>
+          </emu-clause>
+
+          <emu-clause id="sec-ContinueModuleLoading" type="abstract operation">
+            <h1>
+              ContinueModuleLoading (
+                _state_: a GraphLoadingState Record,
+                _moduleCompletion_: either a normal completion containing a Module Record or a throw completion,
+              ): ~unused~
+            </h1>
+            <dl class="header">
+              <dt>description</dt>
+              <dd>It is used to re-enter the loading process after a call to HostLoadImportedModule.</dd>
+            </dl>
+
+            <emu-alg>
+              1. If _state_.[[IsLoading]] is *false*, return ~unused~.
+              1. If _moduleCompletion_ is a normal completion, then
+                1. Perform InnerModuleLoading(_state_, _moduleCompletion_.[[Value]]).
+              1. Else,
+                1. Set _state_.[[IsLoading]] to *false*.
+                1. Perform ! Call(_state_.[[PromiseCapability]].[[Reject]], *undefined*, « _moduleCompletion_.[[Value]] »).
+              1. Return ~unused~.
+            </emu-alg>
+          </emu-clause>
+        </emu-clause>
+
         <emu-clause id="sec-moduledeclarationlinking" type="concrete method" oldids="sec-moduledeclarationinstantiation">
           <h1>Link ( ): either a normal completion containing ~unused~ or a throw completion</h1>
           <dl class="header">
@@ -26069,7 +26317,7 @@
           </dl>
 
           <emu-alg>
-            1. Assert: _module_.[[Status]] is not ~linking~ or ~evaluating~.
+            1. Assert: _module_.[[Status]] is ~unlinked~, ~linked~, ~evaluating-async~, or ~evaluated~.
             1. Let _stack_ be a new empty List.
             1. Let _result_ be Completion(InnerModuleLinking(_module_, _stack_, 0)).
             1. If _result_ is an abrupt completion, then
@@ -26109,7 +26357,7 @@
               1. Set _index_ to _index_ + 1.
               1. Append _module_ to _stack_.
               1. For each String _required_ of _module_.[[RequestedModules]], do
-                1. Let _requiredModule_ be ? HostResolveImportedModule(_module_, _required_).
+                1. Let _requiredModule_ be GetImportedModule(_module_, _required_).
                 1. Set _index_ to ? InnerModuleLinking(_requiredModule_, _stack_, _index_).
                 1. If _requiredModule_ is a Cyclic Module Record, then
                   1. Assert: _requiredModule_.[[Status]] is either ~linking~, ~linked~, ~evaluating-async~, or ~evaluated~.
@@ -26202,8 +26450,7 @@
               1. Set _index_ to _index_ + 1.
               1. Append _module_ to _stack_.
               1. For each String _required_ of _module_.[[RequestedModules]], do
-                1. Let _requiredModule_ be ! HostResolveImportedModule(_module_, _required_).
-                1. NOTE: Link must be completed successfully prior to invoking this method, so every requested module is guaranteed to resolve successfully.
+                1. Let _requiredModule_ be GetImportedModule(_module_, _required_).
                 1. Set _index_ to ? InnerModuleEvaluation(_requiredModule_, _stack_, _index_).
                 1. If _requiredModule_ is a Cyclic Module Record, then
                   1. Assert: _requiredModule_.[[Status]] is either ~evaluating~, ~evaluating-async~, or ~evaluated~.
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -27616,7 +27879,8 @@ <h1>
               1. Assert: _module_ imports a specific binding for this export.
               1. Append _e_.[[ExportName]] to _exportedNames_.
             1. For each ExportEntry Record _e_ of _module_.[[StarExportEntries]], do
-              1. Let _requestedModule_ be ? HostResolveImportedModule(_module_, _e_.[[ModuleRequest]]).
+              1. Let _requiredModule_ be GetImportedModule(_module_, _e_.[[ModuleRequest]]).
+              1. If _requiredModule_ is ~empty~, throw a *TypeError* exception.
```
Jack-Works, line 27883: Should this be an assertion or a TypeError?

spec.html
```diff
@@ -27792,123 +28058,85 @@ <h1>
...
-          1. Perform PerformPromiseThen(_innerPromise_, _onFulfilled_, _onRejected_).
-          1. Return ~unused~.
+          1. If _result_ is a normal completion, then
+            1. If _referrer_.[[LoadedModules]] contains a Record _record_ such that _record_.[[Specifier]] is _specifier_, then
+              1. Assert: _record_.[[Module]] is _result_.[[Value]].
+            1. Else, add Record { [[Specifier]]: _specifier_, [[Module]]: _result_.[[Value]] } to _referrer_.[[LoadedModules]].
+          1. If _state_.[[Action]] is ~graph-loading~, then
+            1. Perform ContinueModuleLoading(_state_, _result_).
+          1. Else,
+            1. Perform ContinueDynamicImport(_state_, _result_).
```
Jack-Works, line 28139:
```suggestion
          1. If _payload_.[[Action]] is ~graph-loading~, then
            1. Perform ContinueModuleLoading(_payload_, _result_).
          1. Else,
            1. Perform ContinueDynamicImport(_payload_, _result_).
```

spec.html
nicolo-ribaudo: The current spec uses ? so it's an error, but I think an assertion would be better (it's the same as point 2 of the PR description)

nicolo-ribaudo:
The HTML side of this refactor is starting to get reviewed, I would love to receive a review for this PR too!

spec.html
```diff
@@ -26403,6 +26494,169 @@ <h1>Cyclic Module Records</h1>
...
+
+            <emu-alg>
+              1. Assert: _state_.[[IsLoading]] is *true*.
+              1. If _module_ is a Cyclic Module Record, _module_.[[Status]] is ~new~, and _state_.[[Visited]] does not contain _module_, then
+                1. Add _module_ to _state_.[[Visited]].
+                1. Let _requestedModulesCount_ be the length of _module_.[[RequestedModules]].
+                1. Set _state_.[[PendingModules]] to _state_.[[PendingModules]] + _requestedModulesCount_.
+                1. For each String _required_ of _module_.[[RequestedModules]], do
+                  1. If _module_.[[LoadedModules]] contains a Record _record_ such that _record_.[[Specifier]] is _required_, then
+                    1. Perform ContinueModuleLoading(_state_, NormalCompletion(_record_.[[Module]])).
```
guybedford, line 26623: Does `ContinueModuleLoading` really need to be a separate function? The logic seems short enough to me that it might actually be worth inlining. Even though it causes a little duplication, I believe it would also lead to some simplifications and clearer reading.

spec.html
```diff
@@ -26403,6 +26494,169 @@ <h1>Cyclic Module Records</h1>
...
+                  1. Else,
+                    1. Perform HostLoadImportedModule(_module_, _required_, _state_.[[HostDefined]], _state_).
+                    1. NOTE: HostLoadImportedModule will call ContinueModuleLoading.
```
guybedford, line 26626: If inlining `ContinueModuleLoading` perhaps note that `HostLoadImportedModule` will recursively call `InnerModuleLoading` when it has loaded the next module.

spec.html
```diff
@@ -27849,123 +28115,85 @@ <h1>
...
-              <dd>
-                <ul>
-                  <li>At some future time, the host environment must perform FinishDynamicImport(_referencingScriptOrModule_, _specifier_, _promiseCapability_, _promise_), where _promise_ is a Promise rejected with an error representing the cause of failure.</li>
-                </ul>
-              </dd>
-            </dl>
+        <p>An implementation of HostLoadImportedModule must conform to the following requirements:</p>
+        <ul>
+          <li>
+            The host environment must perform FinishLoadingImportedModule(_referrer_, _specifier_, _payload_, _result_), where _result_ is either a normal completion containing the loaded Module Record or a throw completion, either synchronously or asynchronously.
```
guybedford, line 28162: In the case where a module has already been loaded, is it up to the host whether it resolves synchronously or asynchronously? I wonder if we should mandate already-loaded modules return synchronously to avoid unnecessary divergence?

spec.html
```diff
@@ -26403,6 +26494,169 @@ <h1>Cyclic Module Records</h1>
...
+            <tr>
+              <td>
+                [[IsLoading]]
+              </td>
+              <td>
+                a boolean
+              </td>
+              <td>
+                This field is only used if [[Action]] is ~graph-loading~. It is true if the loading process has not finished yet, neither successfully nor with an error.
+              </td>
```
guybedford, line 26542: Do we really need a separate `[[IsLoading]]` state if it is equivalent to checking `[[PendingModules]] > 0`?

spec.html
```diff
@@ -19462,15 +19476,61 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+          1. Let _state_ be a new ModuleLoadState Record { [[Action]]: ~dynamic-import~, [[PromiseCapability]]: _promiseCapability_ }.
+          1. Perform HostLoadImportedModule(_referrer_, _specifierString_, *undefined*, _state_).
           1. Return _promiseCapability_.[[Promise]].
         </emu-alg>
+
+        <emu-clause id="sec-ContinueDynamicImport" type="abstract operation">
+          <h1>
+            ContinueDynamicImport (
+              _state_: a ModuleLoadState Record whose [[Action]] is ~dynamic-import~,
+              _module_: either a normal completion containing a Module Record or a throw completion,
```
guybedford, line 19495: Perhaps use `_moduleCompletion_` to avoid overloading the `_module_` variable type?

spec.html
```diff
@@ -27849,123 +28115,85 @@ <h1>
...
-          1. Perform PerformPromiseThen(_innerPromise_, _onFulfilled_, _onRejected_).
-          1. Return ~unused~.
+          1. If _result_ is a normal completion, then
+            1. If _referrer_.[[LoadedModules]] contains a Record _record_ such that _record_.[[Specifier]] is _specifier_, then
+              1. Assert: _record_.[[Module]] is _result_.[[Value]].
+            1. Else, add Record { [[Specifier]]: _specifier_, [[Module]]: _result_.[[Value]] } to _referrer_.[[LoadedModules]].
+          1. If _state_.[[Action]] is ~graph-loading~, then
+            1. Perform ContinueModuleLoading(_state_, _result_).
+          1. Else,
+            1. Perform ContinueDynamicImport(_state_, _result_).
```
guybedford, line 28196: Why do dynamic imports not go through recursive `LoadRequestedModules` calls the same as static imports? Would such work be possible here or as a follow-up, or does it involve non-trivial integration changes to the dynamic import hooks?

spec.html
```diff
@@ -19462,15 +19476,61 @@ <h1>Runtime Semantics: Evaluation</h1>
...
-          1. Let _referencingScriptOrModule_ be GetActiveScriptOrModule().
+          1. Let _referrer_ be GetActiveScriptOrModule().
+          1. If _referrer_ is *null*, set _referrer_ to the current Realm Record.
           1. Let _argRef_ be ? Evaluation of |AssignmentExpression|.
           1. Let _specifier_ be ? GetValue(_argRef_).
           1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
           1. Let _specifierString_ be Completion(ToString(_specifier_)).
           1. IfAbruptRejectPromise(_specifierString_, _promiseCapability_).
-          1. Perform HostImportModuleDynamically(_referencingScriptOrModule_, _specifierString_, _promiseCapability_).
+          1. Let _state_ be a new ModuleLoadState Record { [[Action]]: ~dynamic-import~, [[PromiseCapability]]: _promiseCapability_ }.
```
guybedford, line 19486: We should probably still set the other fields to empty?

guybedford (commented):
> It's possible that one day HTML will align requests of static and dynamic imports, but today is not that day yet.

Would be interested to hear more!

> ResolveExport is described as returning null or something, but it can also return a throw completion (if it's called on a module that re-exports from a dependency that hasn't been already loaded - NOTE: this can only happen if a host calls ResolveExport() even if LoadRequestedModules() doesn't succeed, and they don't have any good reason to do so):

I don't think there is any real-world expectation that has ever relied upon `ResolveExport` throwing, so my preference would be turning this into an internal assertion and enforcing the behaviour that it is only ever called on loaded modules. Is that possible to do?

spec.html
nicolo-ribaudo: I like it as a separate function because it allows keeping the loading logic clearly separated in three sections:
1- AOs for static imports: `LoadRequestedModules`, `InnerModuleLoading`, and `ContinueModuleLoading`
2. AOs for dynamic imports: Runtime Semantics of `import ( AssignmentExpression )` and `ContinueDynamicImport`
3. AOs used by both to interact with the host: `HostLoadImportedModule` and `FinishLoadingImportedModule`, which just caches and dispatches back to 1 or 2

spec.html
nicolo-ribaudo: Until last week HTML was ambiguous allowed loading of already-loaded modules to be asynchronous. I updated HTML to make it synchronous, but without first discussing with implementers, and browsers don't currently agree.

I'm working on it on https://github.com/web-platform-tests/wpt/pull/35949, but I prefer to keep it out of this PR (since the consensus was based on the fact that this was not observably normative in common hosts).

spec.html
nicolo-ribaudo: They do, it's at step 8 of `ContinueDynamicImport`.

spec.html
nicolo-ribaudo: It's not equivalent if the host calls `FinishLoadingImportedModule` with a throw completion: there might still be pending modules, but we stop the loading process.

spec.html
guybedford: Glad to hear you're working on it!

spec.html
```diff
@@ -19462,15 +19476,61 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+                1. Let _namespace_ be Completion(GetModuleNamespace(_module_)).
+                1. If _namespace_ is an abrupt completion, then
+                  1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _namespace_.[[Value]] &raquo;).
+                1. Else,
+                  1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, &laquo; _namespace_.[[Value]] &raquo;).
+                1. Return ~unused~.
+              1. Let _onFulfilled_ be CreateBuiltinFunction(_fulfilledClosure_, *""*, 0, &laquo; &raquo;).
+              1. Perform PerformPromiseThen(_evaluatePromise_, _onFulfilled_, _onRejected_).
+            1. Let _linkAndEvaluate_ be CreateBuiltinFunction(_linkAndEvaluateClosure_, *""*, 0, &laquo; &raquo;).
+            1. Let _loadPromise_ be _module_.LoadRequestedModules().
```
guybedford, line 19529: Maybe move this up, since it doesn't explicitly depend on `_linkAndEvaluate_` being defined? I missed it because I scanned link and evaluate and assumed that it was already moving to that stage of processing.

spec.html
guybedford: That's amazing you could work the full unification here, thanks.

spec.html
guybedford: When reading the `ContinueModuleLoading` call in `LoadRequestedModules` one has to jump to `ContinueModuleLoading`, which then already passes the "isLoaded" check and also the normal completion check so that it's effectively just a one-liner. Perhaps just inlinine it here only then? It's easier to see that `LoadRequestedModules` is a recursive function that way.

nicolo-ribaudo: The [[IsLoading]] check is still needed, because a call to HostLoadImportedModule done previously in the loop might have set it to false. However, you made me realize that we need that check in LoadReqeustedModules anyway, to avoid calling HostLoadImportedModule when [[IsLoading]] becomes false.

spec.html
```diff
@@ -26403,6 +26493,170 @@ <h1>Cyclic Module Records</h1>
...
+              <td>
+                The promise to resolve when the loading process finishes.
+              </td>
+            </tr>
+            <tr>
+              <td>
+                [[IsLoading]]
+              </td>
+              <td>
+                a boolean
```
jmdyck, line 26537:
```suggestion
                a Boolean
```

spec.html
```diff
@@ -26403,6 +26493,170 @@ <h1>Cyclic Module Records</h1>
...
+              </td>
+              <td>
+                It contains host-defined data to pass from the LoadRequestedModules caller to HostLoadImportedModule.
+              </td>
+            </tr>
+          </table>
+        </emu-table>
+
+        <emu-clause id="sec-LoadRequestedModules" type="concrete method">
+          <h1>LoadRequestedModules ( [ _hostDefined_ ] ): a Promise object</h1>
```
jmdyck, line 26580:
```suggestion
          <h1>
            LoadRequestedModules (
              optional _hostDefined_: anything,
            ): a Promise
          </h1>
```

spec.html
```diff
@@ -26403,6 +26493,170 @@ <h1>Cyclic Module Records</h1>
...
+            <dd>a Cyclic Module Record _module_</dd>
+
+            <dt>description</dt>
+            <dd>It populates the [[LoadedModules]] of all the Modue Records in the dependency graph of _module_ (most of the work is done by the auxiliary function InnerModuleLoading). It takes an optional _hostDefined_ parameter, that is passed to the HostLoadImportedModule hook.</dd>
+          </dl>
+
+          <emu-alg>
+            1. If _hostDefined_ is not present, let _hostDefined_ be *undefined*.
+            1. Let _pc_ be ! NewPromiseCapability(%Promise%).
+            1. Let _state_ be a new ModuleLoadState Record { [[Action]]: ~graph-loading~, [[IsLoading]]: *true*, [[PendingModules]]: 1, [[Visited]]: a new empty List, [[PromiseCapability]]: _pc_, [[HostDefined]]: _hostDefined_ }.
```
jmdyck, line 26592:
```suggestion
            1. Let _state_ be the ModuleLoadState Record { [[Action]]: ~graph-loading~, [[IsLoading]]: *true*, [[PendingModules]]: 1, [[Visited]]: &laquo; &raquo;, [[PromiseCapability]]: _pc_, [[HostDefined]]: _hostDefined_ }.
```

spec.html
```diff
@@ -26403,6 +26493,170 @@ <h1>Cyclic Module Records</h1>
...
+            </h1>
+            <dl class="header">
+              <dt>description</dt>
+              <dd>It is used by LoadRequestedModules to recursively perform the actual loading process for _module_'s dependency graph.</dd>
+            </dl>
+
+            <emu-alg>
+              1. Assert: _state_.[[IsLoading]] is *true*.
+              1. If _module_ is a Cyclic Module Record, _module_.[[Status]] is ~new~, and _state_.[[Visited]] does not contain _module_, then
+                1. Add _module_ to _state_.[[Visited]].
```
jmdyck, line 26617:
```suggestion
                1. Append _module_ to _state_.[[Visited]].
```

spec.html
```diff
@@ -26403,6 +26493,170 @@ <h1>Cyclic Module Records</h1>
...
+            </dl>
+
+            <emu-alg>
+              1. If _state_.[[IsLoading]] is *false*, return.
+              1. If _moduleCompletion_ is a normal completion, then
+                1. Perform InnerModuleLoading(_state_, _moduleCompletion_.[[Value]]).
+              1. Else,
+                1. Set _state_.[[IsLoading]] to *false*.
+                1. Perform ! Call(_state_.[[PromiseCapability]].[[Reject]], *undefined*, &laquo; _moduleCompletion_.[[Value]] &raquo;).
+            </emu-alg>
```
jmdyck, line 26656:
```suggestion
              1. Return ~unused~.
            </emu-alg>
```

spec.html
```diff
@@ -27849,123 +28115,85 @@ <h1>
...
-          1. Let _rejectedClosure_ be a new Abstract Closure with parameters (_error_) that captures _promiseCapability_ and performs the following steps when called:
-            1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _error_ &raquo;).
-            1. Return ~unused~.
-          1. Let _onRejected_ be CreateBuiltinFunction(_rejectedClosure_, 0, *""*, &laquo; &raquo;).
-          1. Perform PerformPromiseThen(_innerPromise_, _onFulfilled_, _onRejected_).
-          1. Return ~unused~.
+          1. If _result_ is a normal completion, then
+            1. If _referrer_.[[LoadedModules]] contains a Record _record_ such that _record_.[[Specifier]] is _specifier_, then
+              1. Assert: _record_.[[Module]] is _result_.[[Value]].
+            1. Else, add Record { [[Specifier]]: _specifier_, [[Module]]: _result_.[[Value]] } to _referrer_.[[LoadedModules]].
```
jmdyck, line 28192:
```suggestion
            1. Else, append the Record { [[Specifier]]: _specifier_, [[Module]]: _result_.[[Value]] } to _referrer_.[[LoadedModules]].
```

jmdyck (changes requested):

jmdyck:
Oh, I forgot to say: This PR undefines the ids:
- sec-hostresolveimportedmodule
- sec-hostimportmoduledynamically
- sec-finishdynamicimport

which should maybe become oldids.

nicolo-ribaudo:
Thanks for the review @jmdyck! I added `sec-hostresolveimportedmodule`/`sec-hostimportmoduledynamically` as oldids of HostLoadImportedModule, and `sec-finishdynamicimport` as oldids fo FinishLoadingImportedModule. However, I'm not sure if it's correct because those aren't just renames but completely new AOs.

spec.html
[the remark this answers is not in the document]
nicolo-ribaudo: LoadRequestedModules is part of the "public API", so I prefer to keep it this way. I don't think that this spec optimization matters in practice.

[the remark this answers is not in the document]
guybedford: What do you mean when you say "public API" here?

I do think it is quite a wart to ignore fields, then recreate the same record twice in spec text.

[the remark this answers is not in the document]
nicolo-ribaudo: It's expected to be called by hosts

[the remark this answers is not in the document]
guybedford: Right, but within an individual spec load operation, we shouldn't need to overload the internal state while also having two separate records representing that - we should pick one or the other approach.

If you don't want to add an optional record to the "public API", we could instead separate:

* `LoadRequestedModules` -> `1. Let _state_ be a new record with fields... 2. LoadRequestedModulesInternal(_state_)
* Dynamic import would then directly call `LoadRequestedModulesInternal(_state_).

So I don't think the public / private argument causes any issue here?

When you import a module reflection, that's a branch in the graph loader on this edge property of the node, just like it is for dynamic import, and which we will have to write, just like we will have to write branching in dynamic import.

The issue seems to me more to do with the fact that dynamic import has to go through an _initial phase_, before it can start the graph loading algorithm, where that initial phase is defined by the host hook which needs to call back with some state tracking. An alternative might be to use another type of state record to represent this instead of overloading?

Please let me know if I'm missing any further congruences here though.

caridy (approved):

spec.html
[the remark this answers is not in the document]
nicolo-ribaudo: I implemented your suggestion at https://github.com/nicolo-ribaudo/ecma262/pull/3, however I find it way less clear than the current status. It feels like an optimizations that engines might do to avoid instantiating two records, but which isn't necessary at the spec level.

I'll leave the commit there, and defer to the editors the decision whether to do it or not.

spec.html
```diff
@@ -19462,15 +19476,60 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+            <dt>description</dt>
+            <dd>It completes the process of a dynamic import originally started by an <emu-xref href="#sec-import-calls">`import()`</emu-xref> call, resolving or rejecting the promise returned by that call as appropriate.</dd>
+          </dl>
+          <emu-alg>
+            1. Let _promiseCapability_ be _state_.[[PromiseCapability]].
+            1. If _moduleCompletion_ is an abrupt completion, then
+              1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _moduleCompletion_.[[Value]] &raquo;).
+              1. Return ~unused~.
+            1. Let _module_ be _moduleCompletion_.[[Value]].
+            1. Let _loadPromise_ be _module_.LoadRequestedModules().
```
Jack-Works, line 19508:
```suggestion
            1. Let _loadPromise_ be _module_.LoadRequestedModules(state.[[HostDefined]]).
```

nicolo-ribaudo: `state.[[HostDefined]]` is always `undefined` here, because the import has not been triggered by the host. Maybe I should add an assertion to make it more clear?

nicolo-ribaudo:
@tc39/ecma262-editors is there anything I could do to help moving this PR forward? There are a bunch of proposals that needs to be updated on top of this, so having it reviewed&merged would be nice.

There are some editorial questions I have ([in the issue description](https://github.com/tc39/ecma262/pull/2905#issue-1374634403)), and you can probably give your opinion on them just reading the new spec text without first doing a deep review, so that I can start resolving them :)

spec.html
```diff
@@ -26403,6 +26493,176 @@ <h1>Cyclic Module Records</h1>
...
+              <td>
+                a Boolean
+              </td>
+              <td>
+                This field is only used if [[Action]] is ~graph-loading~. It is true if the loading process has not finished yet, neither successfully nor with an error.
+              </td>
+            </tr>
+            <tr>
+              <td>
+                [[PendingModules]]
```
syg, line 26545: Nit: PendingModuleCount instead of PendingModules, which sounds more like a list.

spec.html
```diff
@@ -26403,6 +26493,176 @@ <h1>Cyclic Module Records</h1>
...
+          <dl class="header">
+            <dt>for</dt>
+            <dd>a Cyclic Module Record _module_</dd>
+
+            <dt>description</dt>
+            <dd>It populates the [[LoadedModules]] of all the Modue Records in the dependency graph of _module_ (most of the work is done by the auxiliary function InnerModuleLoading). It takes an optional _hostDefined_ parameter, that is passed to the HostLoadImportedModule hook.</dd>
+          </dl>
+
+          <emu-alg>
+            1. If _hostDefined_ is not present, let _hostDefined_ be *undefined*.
```
syg, line 26594: `~empty~` would be more appropriate as the default for _hostDefined_.

spec.html
```diff
@@ -26403,6 +26493,176 @@ <h1>Cyclic Module Records</h1>
...
+
+            <emu-alg>
+              1. Assert: _state_.[[IsLoading]] is *true*.
+              1. If _module_ is a Cyclic Module Record, _module_.[[Status]] is ~new~, and _state_.[[Visited]] does not contain _module_, then
+                1. Append _module_ to _state_.[[Visited]].
+                1. Let _requestedModulesCount_ be the length of _module_.[[RequestedModules]].
+                1. Set _state_.[[PendingModules]] to _state_.[[PendingModules]] + _requestedModulesCount_.
+                1. For each String _required_ of _module_.[[RequestedModules]], do
+                  1. If _state_.[[IsLoading]] is *true*, then
+                    1. If _module_.[[LoadedModules]] contains a Record _record_ such that _record_.[[Specifier]] is _required_, then
```
syg, line 26626:
```suggestion
                    1. If _module_.[[LoadedModules]] contains a Record whose [[Specifier]] is _required_, then
```

spec.html
```diff
@@ -27849,123 +28121,85 @@ <h1>
...
+            The operation must treat _payload_ as an opaque value to be passed through to FinishLoadingImportedModule.
           </li>
         </ul>
 
-        <p>The actual process performed is host-defined, but typically consists of performing whatever I/O operations are necessary to allow HostResolveImportedModule to synchronously retrieve the appropriate Module Record, and then calling its Evaluate concrete method. This might require performing similar normalization as HostResolveImportedModule does.</p>
+        <p>The actual process performed is host-defined, but typically consists of performing whatever I/O operations are necessary to load the appropriate Module Record. Multiple different (_referrer_, _specifier_) pairs may map to the same Module Record instance. The actual mapping semantic is host-defined but typically a normalization process is applied to _specifier_ as part of the mapping process. A typical normalization process would include actions such as expansion of relative and abbreviated path specifiers.</p>
```
syg, line 28178:
```suggestion
        <p>The actual process performed is host-defined, but typically consists of performing whatever I/O operations are necessary to load the appropriate Module Record. Multiple different (_referrer_, _specifier_) pairs may map to the same Module Record instance. The actual mapping semantics is host-defined but typically a normalization process is applied to _specifier_ as part of the mapping process. A typical normalization process would include actions such as expansion of relative and abbreviated path specifiers.</p>
```

spec.html
```diff
@@ -27849,123 +28121,85 @@ <h1>
...
-            1. Return ~unused~.
-          1. Let _onFulfilled_ be CreateBuiltinFunction(_fulfilledClosure_, 0, *""*, &laquo; &raquo;).
-          1. Let _rejectedClosure_ be a new Abstract Closure with parameters (_error_) that captures _promiseCapability_ and performs the following steps when called:
-            1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, &laquo; _error_ &raquo;).
-            1. Return ~unused~.
-          1. Let _onRejected_ be CreateBuiltinFunction(_rejectedClosure_, 0, *""*, &laquo; &raquo;).
-          1. Perform PerformPromiseThen(_innerPromise_, _onFulfilled_, _onRejected_).
-          1. Return ~unused~.
+          1. If _result_ is a normal completion, then
+            1. If _referrer_.[[LoadedModules]] contains a Record _record_ such that _record_.[[Specifier]] is _specifier_, then
```
syg, line 28196:
```suggestion
            1. If _referrer_.[[LoadedModules]] contains a Record whose [Specifier]] is _specifier_, then
```

spec.html
```diff
@@ -27849,123 +28121,85 @@ <h1>
...
-          <li>
-            If a Module Record corresponding to the pair _referencingScriptOrModule_, _specifier_ does not exist or cannot be created, an exception must be thrown.
-          </li>
-          <li>
-            Each time this operation is called, if it completes normally and the (_referencingScriptOrModule_, _specifier_, current Realm Record) triple is the same, it must return the same Module Record instance.
-          </li>
-        </ul>
-        <p>Multiple different _referencingScriptOrModule_, _specifier_ pairs may map to the same Module Record instance. The actual mapping semantic is host-defined but typically a normalization process is applied to _specifier_ as part of the mapping process. A typical normalization process would include actions such as alphabetic case folding and expansion of relative and abbreviated path specifiers.</p>
+        <emu-alg>
+          1. If _referrer_.[[LoadedModules]] contains a Record _record_ such that _record_.[[Specifier]] is _specifier_, then
```
syg, line 28137:
```suggestion
          1. If _referrer_.[[LoadedModules]] contains a Record _record_ whose [[Specifier]] is _specifier_, then
```

syg:
@nicolo-ribaudo We discussed some of your questions on the editor call, our preferences are:

1. We have plans eventually to basically have asynchronous Abstract Closures. For now, since you already wrote the manual promise-handler version, just keep it as is, we'll simplify this when we introduce the ability to have async ACs.

2. We don't quite grok the scenario you're describing. Are you saying the only reason this method would throw today is if a host calls it in a weird way that no host actually does? If so, please help us confirm if that's the case, and once confirmed, we should add assertions and make the method infallible.

3. Name sounds good.

5. Please name and split them into two separate Record types. Get rid of [[Action]] and just query for if the record is one of the specific record types.

spec.html
nicolo-ribaudo: This is more clear now that I don't use a `ModuleLoadState Record` for both actions.

nicolo-ribaudo:
I added a commit to do (3), which probably is the most sensible option. I can revert it if you prefer (1) or (2).

guybedford (approved):
Really happy with how this has turned out! I would still slightly prefer if `ContinueModuleLoading` could be somehow inlined / folded into idempotency of `InnerModuleLoading`, but that's nothing major at all.

spec.html
```diff
@@ -26111,6 +26198,166 @@ <h1>Cyclic Module Records</h1>
...
+              </td>
+            </tr>
+          </table>
+        </emu-table>
+
+        <emu-clause id="sec-LoadRequestedModules" type="concrete method">
+          <h1>
+            LoadRequestedModules (
+              optional _hostDefined_: anything,
+            ): a Promise object
```
jmdyck, line 26277:
```suggestion
            ): a Promise
```

spec.html
```diff
@@ -26111,6 +26198,166 @@ <h1>Cyclic Module Records</h1>
...
+            <dd>a Cyclic Module Record _module_</dd>
+
+            <dt>description</dt>
+            <dd>It populates the [[LoadedModules]] of all the Modue Records in the dependency graph of _module_ (most of the work is done by the auxiliary function InnerModuleLoading). It takes an optional _hostDefined_ parameter, that is passed to the HostLoadImportedModule hook.</dd>
+          </dl>
+
+          <emu-alg>
+            1. If _hostDefined_ is not present, let _hostDefined_ be ~empty~.
+            1. Let _pc_ be ! NewPromiseCapability(%Promise%).
+            1. Let _state_ be the GraphLoadingState Record { [[IsLoading]]: *true*, [[PendingModulesCount]]: 1, [[Visited]]: a new empty List, [[PromiseCapability]]: _pc_, [[HostDefined]]: _hostDefined_ }.
```
jmdyck, line 26290:
```suggestion
            1. Let _state_ be the GraphLoadingState Record { [[IsLoading]]: *true*, [[PendingModulesCount]]: 1, [[Visited]]: « », [[PromiseCapability]]: _pc_, [[HostDefined]]: _hostDefined_ }.
```

spec.html
```diff
@@ -26111,6 +26198,166 @@ <h1>Cyclic Module Records</h1>
...
+                    1. Perform InnerModuleLoading(_state_, _record_.[[Module]]).
+                  1. Else,
+                    1. Perform HostLoadImportedModule(_module_, _required_, _state_.[[HostDefined]], _state_).
+                    1. NOTE: HostLoadImportedModule will call FinishLoadingImportedModule, which re-enters the graph loading process through ContinueModuleLoading.
+                  1. If _state_.[[IsLoading]] is *false*, return ~unused~.
+              1. Assert: _state_.[[PendingModulesCount]] ≥ 1.
+              1. Set _state_.[[PendingModulesCount]] to _state_.[[PendingModulesCount]] - 1.
+              1. If _state_.[[PendingModulesCount]] = 0, then
+                1. Set _state_.[[IsLoading]] to *false*.
+                1. For each Cyclic Module Record _loaded_ in _state_.[[Visited]], do
```
jmdyck, line 26330:
```suggestion
                1. For each Cyclic Module Record _loaded_ of _state_.[[Visited]], do
```
See PR #2152.

spec.html
```diff
@@ -27350,7 +27604,7 @@ <h1>
...
+            1. Return Source Text Module Record { [[Realm]]: _realm_, [[Environment]]: ~empty~, [[Namespace]]: ~empty~, [[CycleRoot]]: ~empty~, [[HasTLA]]: _async_, [[AsyncEvaluation]]: *false*, [[TopLevelCapability]]: ~empty~, [[AsyncParentModules]]: « », [[PendingAsyncDependencies]]: ~empty~, [[Status]]: ~new~, [[EvaluationError]]: ~empty~, [[HostDefined]]: _hostDefined_, [[ECMAScriptCode]]: _body_, [[Context]]: ~empty~, [[ImportMeta]]: ~empty~, [[RequestedModules]]: _requestedModules_, [[LoadedModules]]: a new empty List, [[ImportEntries]]: _importEntries_, [[LocalExportEntries]]: _localExportEntries_, [[IndirectExportEntries]]: _indirectExportEntries_, [[StarExportEntries]]: _starExportEntries_, [[DFSIndex]]: ~empty~, [[DFSAncestorIndex]]: ~empty~ }.
```
jmdyck, line 27607:
```suggestion
            1. Return Source Text Module Record { [[Realm]]: _realm_, [[Environment]]: ~empty~, [[Namespace]]: ~empty~, [[CycleRoot]]: ~empty~, [[HasTLA]]: _async_, [[AsyncEvaluation]]: *false*, [[TopLevelCapability]]: ~empty~, [[AsyncParentModules]]: « », [[PendingAsyncDependencies]]: ~empty~, [[Status]]: ~new~, [[EvaluationError]]: ~empty~, [[HostDefined]]: _hostDefined_, [[ECMAScriptCode]]: _body_, [[Context]]: ~empty~, [[ImportMeta]]: ~empty~, [[RequestedModules]]: _requestedModules_, [[LoadedModules]]: « », [[ImportEntries]]: _importEntries_, [[LocalExportEntries]]: _localExportEntries_, [[IndirectExportEntries]]: _indirectExportEntries_, [[StarExportEntries]]: _starExportEntries_, [[DFSIndex]]: ~empty~, [[DFSAncestorIndex]]: ~empty~ }.
```

spec.html
```diff
@@ -27557,122 +27812,85 @@ <h1>
...
-            If a Module Record corresponding to the pair _referencingScriptOrModule_, _specifier_ does not exist or cannot be created, an exception must be thrown.
-          </li>
-          <li>
-            Each time this operation is called, if it completes normally and the (_referencingScriptOrModule_, _specifier_, current Realm Record) triple is the same, it must return the same Module Record instance.
-          </li>
-        </ul>
-        <p>Multiple different _referencingScriptOrModule_, _specifier_ pairs may map to the same Module Record instance. The actual mapping semantic is host-defined but typically a normalization process is applied to _specifier_ as part of the mapping process. A typical normalization process would include actions such as alphabetic case folding and expansion of relative and abbreviated path specifiers.</p>
+        <emu-alg>
+          1. Let _record_ be the sole element of _referrer_.[[LoadedModules]] whose [[Specifier]] is _specifier_.
+          1. Assert: _record_ exists, because LoadRequestedModules has completed successfully on _referrer_ prior to invoking this abstract operation.
```
jmdyck, line 27829: The spec doesn't currently use this form:
```
1. Let _alias_ be [something that might or might not exist].
2. Assert: _alias_ exists. // If _alias_ exists ...
```
but maybe the editors are okay with it.

jmdyck (changes requested):

spec.html
nicolo-ribaudo: I could use one of these:
- ```
  1. Let _record_ be ~empty~.
  2. If _referrer_.[[LoadedModules]] contains a Record whose [[Specifier]] is _specifier_, set _record_ to that Record.
  3. Assert: _record_ is not ~empty~.
  ```
  which reuses the "If ... contains X such that ... / Let _x_ be that X" wording from https://tc39.es/ecma262/#sec-privateelementfind
- ```
  1. Let _record_ be ~empty~.
  2. For each Record _r_ in _referrer_.[[LoadedModules]], do
     1. If _r_.[[Specifier]] is _specifier, then
        1. Assert: _record_ is ~empty~.
        2. Set _record_ to _r_.
  5. Assert: _record_ is not ~empty~.
  ```

bakkot: https://github.com/tc39/ecma262/pull/2877 changes PrivateElementFind to be a more explicit loop:

```
1. For each Private Name _pn_ of _names_, do
  1. If _pn_.[[Description]] is _identifier_, then
    1. Return _pn_.
```
But in that case you don't know it's _exactly_ one, because you might have to traverse the environment chain.

A more analogous case is [in the evaluation semantics for private names as class elements](https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#sec-class-definitions-runtime-semantics-evaluation):

```
4. Assert: Exactly one element of _names_ is a Private Name whose [[Description]] is _privateIdentifier_.
5. Let _privateName_ be the Private Name in _names_ whose [[Description]] is _privateIdentifier_.
```

That's what I'd do here:

```suggestion
          1. Assert: Exactly one element of _referrer_.[[LoadedModules]] is a Record whose [[Specifier]] is _specifier_, since LoadRequestedModules has completed successfully on _referrer_ prior to invoking this abstract operation.
          1. Let _record_ be the Record in _referrer_.[[LoadedModules]] whose [[Specifier]] is _specifier_.
```

spec.html
```diff
@@ -27557,122 +27812,85 @@ <h1>
...
-              <dd>
-                <ul>
-                  <li>At some future time, the host environment must perform FinishDynamicImport(_referencingScriptOrModule_, _specifier_, _promiseCapability_, _promise_), where _promise_ is a Promise resolved with *undefined*.</li>
+        <emu-note>
+          <p>An example of when _referrer_ can be a Realm Record is in a web browser host. There, if a user clicks on a control given by</p>
 
-                  <li>Any subsequent call to HostResolveImportedModule after FinishDynamicImport has completed, given the arguments _referencingScriptOrModule_ and _specifier_, must return a normal completion containing a module which has already been evaluated, i.e. whose Evaluate concrete method has already been called and returned a normal completion.</li>
-                </ul>
-              </dd>
+          <pre><code class="html">&lt;button type="button" onclick="import('./foo.mjs')"&gt;Click me&lt;/button&gt;</code></pre>
```
bakkot, line 27851: Frankly it's pretty weird that this has `null` active script. (For reference, it's in step 9 [here](https://html.spec.whatwg.org/#internal-raw-uncompiled-handler).)

Not something to worry about here, though.

bakkot (approved):
Some nits, but I _think_ I understand what's going on and it seems to hang together.

spec.html
```diff
@@ -11525,6 +11525,20 @@ <h1>Realms</h1>
...
+          <td>
+            [[LoadedModules]]
+          </td>
+          <td>
+            a List of Records with fields [[Specifier]] (a String) and [[Module]] (a Module Record)
+          </td>
+          <td>
+            A map from the specifier strings imported by this realm to the resolved Module Record. The list does not contain two different Records with the same [[Specifier]].
+            <emu-note>
+              As mentioned in HostLoadImportedModule, [[LoadedModules]] in Realm Records is only used when running an `import()` expression in a context where there is no active script or module.
```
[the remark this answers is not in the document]
nicolo-ribaudo: HostLoadImportedModule gets auto-linked to the AO, and right after the AO header there is a big note explaining it.

If it's not enough, can I `<emu-xref>` an `<emu-note>` (I tried just giving the note an id, but the xref doesn't pick it up)?

spec.html
nicolo-ribaudo: Maybe we can eventually get HTML to create a script for inline handlers :shrug:

spec.html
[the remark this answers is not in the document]
nicolo-ribaudo: I don't know what I tried 9 days ago, but it works now. Done!

spec.html
```diff
@@ -11525,6 +11525,20 @@ <h1>Realms</h1>
...
+          <td>
+            [[LoadedModules]]
+          </td>
+          <td>
+            a List of Records with fields [[Specifier]] (a String) and [[Module]] (a Module Record)
+          </td>
+          <td>
+            A map from the specifier strings imported by this realm to the resolved Module Record. The list does not contain two different Records with the same [[Specifier]].
+            <emu-note>
+              As mentioned in HostLoadImportedModule (<emu-xref href="#node-HostLoadImportedModule-referrer-Realm-Record"></emu-xref>), [[LoadedModules]] in Realm Records is only used when running an `import()` expression in a context where there is no active script or module.
```
jmdyck, line 11538:
```suggestion
              As mentioned in HostLoadImportedModule (<emu-xref href="#note-HostLoadImportedModule-referrer-Realm-Record"></emu-xref>), [[LoadedModules]] in Realm Records is only used when running an `import()` expression in a context where there is no active script or module.
```

spec.html
```diff
@@ -19254,15 +19268,55 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+              1. Let _link_ be Completion(_module_.Link()).
+              1. If _link_ is an abrupt completion, then
+                1. Perform ! Call(_promiseCapability_.[[Reject]], *undefined*, « _link_.[[Value]] »).
+                1. Return ~unused~.
+              1. Let _evaluatePromise_ be _module_.Evaluate().
+              1. Let _fulfilledClosure_ be a new Abstract Closure with no parameters that captures _module_ and _promiseCapability_ and performs the following steps when called:
+                1. Let _namespace_ be GetModuleNamespace(_module_).
+                1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, « _namespace_ »).
+                1. Return ~unused~.
+              1. Let _onFulfilled_ be CreateBuiltinFunction(_fulfilledClosure_, *""*, 0, « »).
```
jmdyck, line 19313:
```suggestion
              1. Let _onFulfilled_ be CreateBuiltinFunction(_fulfilledClosure_, 0, *""*, « »).
```

spec.html
```diff
@@ -19254,15 +19268,55 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+              1. Perform PerformPromiseThen(_evaluatePromise_, _onFulfilled_, _onRejected_).
```
jmdyck, line 19314: Insert `Return ~unused~.` step after this one.

spec.html
```diff
@@ -26111,6 +26198,166 @@ <h1>Cyclic Module Records</h1>
...
+            <dl class="header">
+              <dt>description</dt>
+              <dd>It is used by LoadRequestedModules to recursively perform the actual loading process for _module_'s dependency graph.</dd>
+            </dl>
+
+            <emu-alg>
+              1. Assert: _state_.[[IsLoading]] is *true*.
+              1. If _module_ is a Cyclic Module Record, _module_.[[Status]] is ~new~, and _state_.[[Visited]] does not contain _module_, then
+                1. Append _module_ to _state_.[[Visited]].
+                1. Let _requestedModulesCount_ be the length of _module_.[[RequestedModules]].
```
jmdyck, line 26316:
```suggestion
                1. Let _requestedModulesCount_ be the number of elements in _module_.[[RequestedModules]].
```
(Currently, we only use "the length of" for String values.)

spec.html
```diff
@@ -27557,122 +27812,85 @@ <h1>
...
-          </li>
-          <li>
-            The host environment must conform to one of the two following sets of requirements:
-            <dl>
-              <dt>Success path</dt>
 
-              <dd>
-                <ul>
-                  <li>At some future time, the host environment must perform FinishDynamicImport(_referencingScriptOrModule_, _specifier_, _promiseCapability_, _promise_), where _promise_ is a Promise resolved with *undefined*.</li>
+        <emu-note id="node-HostLoadImportedModule-referrer-Realm-Record">
```
jmdyck, line 27848:
```suggestion
        <emu-note id="note-HostLoadImportedModule-referrer-Realm-Record">
```

jmdyck (changes requested):

spec.html
nicolo-ribaudo:
```suggestion
              1. Perform PerformPromiseThen(_evaluatePromise_, _onFulfilled_, _onRejected_).
              1. Return ~unused~.
```

nicolo-ribaudo:
@jmdyck Done, thanks!

nicolo-ribaudo:
Regarding the IPR failure: currently the IPR signatures list only contains my personal email, but I have committed with my Igalia email. I'll fix it.

michaelficarra:
@nicolo-ribaudo Can you try rebasing on `main`?

nicolo-ribaudo:
Still failing, maybe because Tobie left google (or at least, it's not linked anymore in his GitHub profile)?

ljharb:
@nicolo-ribaudo ill take a look at it

spec.html
```diff
@@ -19201,15 +19215,56 @@ <h1>Runtime Semantics: Evaluation</h1>
...
+                1. Return ~unused~.
+              1. Let _evaluatePromise_ be _module_.Evaluate().
+              1. Let _fulfilledClosure_ be a new Abstract Closure with no parameters that captures _module_ and _promiseCapability_ and performs the following steps when called:
+                1. Let _namespace_ be GetModuleNamespace(_module_).
+                1. Perform ! Call(_promiseCapability_.[[Resolve]], *undefined*, « _namespace_ »).
+                1. Return ~unused~.
+              1. Let _onFulfilled_ be CreateBuiltinFunction(_fulfilledClosure_, 0, *""*, « »).
+              1. Perform PerformPromiseThen(_evaluatePromise_, _onFulfilled_, _onRejected_).
+              1. Return ~unused~.
+            1. Let _linkAndEvaluate_ be CreateBuiltinFunction(_linkAndEvaluateClosure_, *""*, 0, « »).
```
jmdyck, line 19263:
```suggestion
            1. Let _linkAndEvaluate_ be CreateBuiltinFunction(_linkAndEvaluateClosure_, 0, *""*, « »).
```
(I saw this yesterday, but somehow it didn't get into my review.)

spec.html
```diff
@@ -11472,6 +11472,20 @@ <h1>Realms</h1>
...
         </tr>
+        <tr>
+          <td>
+            [[LoadedModules]]
+          </td>
+          <td>
+            a List of Records with fields [[Specifier]] (a String) and [[Module]] (a Module Record)
+          </td>
+          <td>
+            A map from the specifier strings imported by this realm to the resolved Module Record. The list does not contain two different Records with the same [[Specifier]].
```
jmdyck, line 11483:
```suggestion
            <p>A map from the specifier strings imported by this realm to the resolved Module Record. The list does not contain two different Records with the same [[Specifier]].</p>
```
When you've got a block-level element (`<emu-note>`), I think it makes more sense for its 'sibling' content to be block-level as well (as in the cell above for `[[TemplateMap]]`).

ljharb:
It's very strange; the passing tests found 146 authors, in the last day or two it's found 147 including tobie. I'll fix it by adding them to the Emeriti team, since they were employed at Google at the time.

jmdyck (dismissed):

nicolo-ribaudo:
Thanks @ljharb :)

@jmdyck done!

syg (approved):

PR #2915, opened 2022-09-27
Editorial: replace steps embedded in lookup tables with more typical algorithms

michaelficarra:
This will help us catch editorial errors (it already caught 1) and IMO reads better by allowing for other groupings (such as by output in `ToBoolean`).

I'd recommend reviewing before/after rendering instead of reading the changeset.

There's probably more opportunities to replace tables, but this is a good start.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4847,7 +4847,7 @@
       </emu-clause>
     </emu-clause>
 
-    <emu-clause id="sec-toboolean" type="abstract operation">
+    <emu-clause id="sec-toboolean" oldids="table-toboolean-conversions" type="abstract operation">
       <h1>
         ToBoolean (
           _argument_: an ECMAScript language value,
@@ -4855,87 +4855,14 @@
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It converts _argument_ to a value of type Boolean according to <emu-xref href="#table-toboolean-conversions"></emu-xref>:</dd>
+        <dd>It converts _argument_ to a value of type Boolean.</dd>
       </dl>
-      <emu-table id="table-toboolean-conversions" caption="ToBoolean Conversions" oldids="table-10">
-        <table>
-          <tr>
-            <th>
-              Argument Type
-            </th>
-            <th>
-              Result
-            </th>
-          </tr>
-          <tr>
-            <td>
-              Undefined
-            </td>
-            <td>
-              Return *false*.
-            </td>
-          </tr>
-          <tr>
-            <td>
-              Null
-            </td>
-            <td>
-              Return *false*.
-            </td>
-          </tr>
-          <tr>
-            <td>
-              Boolean
-            </td>
-            <td>
-              Return _argument_.
-            </td>
-          </tr>
-          <tr>
-            <td>
-              Number
-            </td>
-            <td>
-              If _argument_ is *+0*<sub>𝔽</sub>, *-0*<sub>𝔽</sub>, or *NaN*, return *false*; otherwise return *true*.
-            </td>
-          </tr>
-          <tr>
-            <td>
-              String
-            </td>
-            <td>
-              If _argument_ is the empty String (its length is 0), return *false*; otherwise return *true*.
-            </td>
-          </tr>
-          <tr>
-            <td>
-              Symbol
-            </td>
-            <td>
-              Return *true*.
-            </td>
-          </tr>
-          <tr>
-            <td>
-              BigInt
-            </td>
-            <td>
-              If _argument_ is *0*<sub>ℤ</sub>, return *false*; otherwise return *true*.
-            </td>
-          </tr>
-          <tr>
-            <td>
-              Object
-            </td>
-            <td>
-              Return *true*.
-              <emu-note>
-                <p>An alternate algorithm related to the [[IsHTMLDDA]] internal slot is mandated in section <emu-xref href="#sec-IsHTMLDDA-internal-slot-to-boolean"></emu-xref>.</p>
-              </emu-note>
-            </td>
-          </tr>
-        </table>
-      </emu-table>
+      <emu-alg>
+        1. If _argument_ is a Boolean, return _argument_.
+        1. If _argument_ is any of *undefined*, *null*, *+0*<sub>𝔽</sub>, *-0*<sub>𝔽</sub>, *NaN*, *0*<sub>ℤ</sub>, or the empty String, return *false*.
+        1. [id="step-to-boolean-web-compat-insertion-point"] NOTE: This step is replaced in section <emu-xref href="#sec-IsHTMLDDA-internal-slot-to-boolean"></emu-xref>.
+        1. Return *true*.
+      </emu-alg>
     </emu-clause>
 
     <emu-clause id="sec-tonumeric" type="abstract operation">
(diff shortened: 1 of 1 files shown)


ljharb:
so many claps; this is awesome

bakkot:
This seems to me to be worse; I'd mildly prefer to leave as-is.

ljharb:
Worse how? Algorithm steps are so much clearer imo than a table.

bakkot:
I find the table easier to read /shrug

michaelficarra:
I'm actually really happy with this change. It exceeded my expectations.

jmdyck:
The ids of the former `<emu-table>` elements should maybe become oldids on the respective `<emu-clause>` elements.

michaelficarra:
@jmdyck Good point. Done.

syg:
I think the algorithmic steps here help clarity over tables, especially when reviewing implementations. It'd also be nice to have this be precedent setting as editorial style to not loop over tables in algorithmic steps. To wit I'd like to tell 404 proposal authors to stop doing stuff like [step 17 here](https://tc39.es/proposal-intl-duration-format/#sec-Intl.DurationFormat) in the Intl.DurationFormat proposal.

bakkot:
I'm fine with this approach, then.

michaelficarra:
@syg @bakkot Can I get a review for correctness then?

spec.html
```diff
@@ -48101,16 +47815,15 @@ <h1>The [[IsHTMLDDA]] Internal Slot</h1>
 
       <emu-annex id="sec-IsHTMLDDA-internal-slot-to-boolean">
         <h1>Changes to ToBoolean</h1>
-        <p>The result column in <emu-xref href="#table-toboolean-conversions"></emu-xref> for an argument type of Object is replaced with the following algorithm:</p>
-        <emu-alg>
+        <p>The following step replaces step <emu-xref href="#step-to-boolean-web-compat-insertion-point"></emu-xref> of ToBoolean:</p>
+        <emu-alg replaces-step="step-to-boolean-web-compat-insertion-point">
           1. If _argument_ has an [[IsHTMLDDA]] internal slot, return *false*.
```
bakkot, line 47820:
```suggestion
          1. If _argument_ is an Object and _argument_ has an [[IsHTMLDDA]] internal slot, return *false*.
```

spec.html
```diff
@@ -19751,95 +19543,17 @@ <h1>Runtime Semantics: Evaluation</h1>
             1. If IsUnresolvableReference(_val_) is *true*, return *"undefined"*.
           1. Set _val_ to ? GetValue(_val_).
           1. [id="step-typeof-web-compat-insertion-point"] NOTE: This step is replaced in section <emu-xref href="#sec-IsHTMLDDA-internal-slot-typeof"></emu-xref>.
```
bakkot, line 19545: Might as well move this down below the `Assert: _val_ is an Object` step below, and remove the `If _val_ is an Object` from the replacement step.

spec.html
```diff
@@ -19751,95 +19543,17 @@ <h1>Runtime Semantics: Evaluation</h1>
...
-          1. Return a String according to <emu-xref href="#table-typeof-operator-results"></emu-xref>.
+          1. If _val_ is *undefined*, return *"undefined"*.
+          1. If _val_ is a Boolean, return *"boolean"*.
+          1. If _val_ is a Number, return *"number"*.
+          1. If _val_ is a String, return *"string"*.
+          1. If _val_ is a Symbol, return *"symbol"*.
+          1. If _val_ is a BigInt, return *"bigint"*.
+          1. If _val_ is *null*, return *"object"*.
```
bakkot, line 19552: Seems like it would be nice for the order of this operation and ToString to match. I don't really care what the order is (though I'd prefer string be next to symbol, number next to bigint, undefined next to null, and object last).

bakkot (commented):
LGTM other than nits.

michaelficarra:
@bakkot Addressed comments.

spec.html
```diff
@@ -48101,16 +47815,15 @@ <h1>The [[IsHTMLDDA]] Internal Slot</h1>
 
       <emu-annex id="sec-IsHTMLDDA-internal-slot-to-boolean">
         <h1>Changes to ToBoolean</h1>
-        <p>The result column in <emu-xref href="#table-toboolean-conversions"></emu-xref> for an argument type of Object is replaced with the following algorithm:</p>
-        <emu-alg>
-          1. If _argument_ has an [[IsHTMLDDA]] internal slot, return *false*.
-          1. Return *true*.
+        <p>The following step replaces step <emu-xref href="#step-to-boolean-web-compat-insertion-point"></emu-xref> of ToBoolean:</p>
+        <emu-alg replaces-step="step-to-boolean-web-compat-insertion-point">
+          1. If _argument_ is an Obejct and _argument_ has an [[IsHTMLDDA]] internal slot, return *false*.
```
bakkot, line 47820:
```suggestion
          1. If _argument_ is an Object and _argument_ has an [[IsHTMLDDA]] internal slot, return *false*.
```

spec.html
```diff
@@ -4911,96 +4838,28 @@ <h1>
...
-              </emu-alg>
-            </td>
-          </tr>
-        </table>
-      </emu-table>
+      <emu-alg>
+        1. If _argument_ is a Number, return _argument_.
+        1. If _argument_ is either a Symbol or a BigInt, throw a *TypeError* exception.
+        1. If _argument_ is *undefined*, return *NaN*.
+        1. If _argument_ is *null* or *false*, return *+0*<sub>𝔽</sub>.
```
syg, line 4855: Since this line doesn't have "either" and L4853 do, I presume it's intentional and your preference to have "either" for types but not for values? If unintentional, let's make them the same.

syg (approved):
lgtm

spec.html
michaelficarra: Thanks for the catch. It was unintentional.

spec.html
```diff
@@ -4847,95 +4847,22 @@ <h1>
...
-              <emu-note>
-                <p>An alternate algorithm related to the [[IsHTMLDDA]] internal slot is mandated in section <emu-xref href="#sec-IsHTMLDDA-internal-slot-to-boolean"></emu-xref>.</p>
-              </emu-note>
-            </td>
-          </tr>
-        </table>
-      </emu-table>
+      <emu-alg>
+        1. If _argument_ is a Boolean, return _argument_.
+        1. If _argument_ is any of *undefined*, *null*, *+0*<sub>𝔽</sub>, *-0*<sub>𝔽</sub>, *NaN*, *0*<sub>ℤ</sub>, or the empty String, return *false*.
```
jmdyck, line 4862: PR #2877 proposes "any of" used this way, but the current spec doesn't use it. Instead, the current spec would say:
- `_argument_ is *undefined*, ...`
- `_argument_ is either *undefined*, ...`
- `_argument_ is one of *undefined*, ...`

michaelficarra: I'm okay with this. It's informal language.

jmdyck: > It's informal language.

It is, but that's true of most of the spec. Every "consistent phrasing" PR changes informal language. So it seems like "informal language" can't be a reason to not aim for consistency. So why bring it up here?

Now, if you were to say "Yeah, it's an anomaly, but #2877 will take care of it soon, so it's not worth fixing here," I'd find that less puzzling.

michaelficarra: @jmdyck I am aiming for consistency, but am okay with more inconsistency for the time being (in informal language) if it's heading in the direction we want to go, as here.

PR #2917, opened 2022-09-27
Editorial: Make better use of clamping

gibson042:
* Reduce single-use aliases.
* Introduce `ToClampedIndex` for relative-to-start-or-end index values.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -5586,19 +5586,15 @@
         ToClampedIndex (
           _value_: an ECMAScript language value,
           _length_: a non-negative integer,
-          optional _lowerBound_: an integer,
-          optional _upperBound_: an integer,
         ): either a normal completion containing an integer or a throw completion
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It converts _value_ to an integer (interpreting negative values relative to _length_) and clamps the result to the inclusive interval from _lowerBound_ to _upperBound_ (respectively defaulting to 0 and _length_).</dd>
+        <dd>It converts _value_ to an integer (interpreting negative values relative to _length_) and clamps the result to the inclusive interval from 0 to _length_.</dd>
       </dl>
       <emu-alg>
-        1. If _lowerBound_ is not present, set _lowerBound_ to 0.
-        1. If _upperBound_ is not present, set _upperBound_ to _length_.
-        1. Assert: _lowerBound_ ≤ _upperBound_.
-        1. Return the result of clamping ? ToAbsoluteIndex(_value_, _length_) between _lowerBound_ and _upperBound_.
+        1. Let _index_ be ? ToAbsoluteIndex(_value_, _length_).
+        1. Return the result of clamping _index_ between 0 and _length_.
       </emu-alg>
     </emu-clause>
   </emu-clause>
@@ -37872,7 +37868,8 @@ THH:mm:ss.sss
           1. Let _O_ be ? ToObject(*this* value).
           1. Let _len_ be ? LengthOfArrayLike(_O_).
           1. If _len_ is 0, return *-1*<sub>𝔽</sub>.
-          1. If _fromIndex_ is not present, let _k_ be _len_ - 1; else let _k_ be ? ToClampedIndex(_fromIndex_, _len_, -1, _len_ - 1).
+          1. If _fromIndex_ is not present, let _k_ be _len_ - 1.
+          1. Else, let _k_ be min(? ToAbsoluteIndex(_fromIndex_, _len_), _len_ - 1).
           1. Repeat, while _k_ ≥ 0,
             1. Let _kPresent_ be ? HasProperty(_O_, ! ToString(𝔽(_k_))).
             1. If _kPresent_ is *true*, then
@@ -39281,7 +39278,8 @@ THH:mm:ss.sss
           1. Perform ? ValidateTypedArray(_O_).
           1. Let _len_ be _O_.[[ArrayLength]].
           1. If _len_ is 0, return *-1*<sub>𝔽</sub>.
-          1. If _fromIndex_ is not present, let _k_ be _len_ - 1; else let _k_ be ? ToClampedIndex(_fromIndex_, _len_, -1, _len_ - 1).
+          1. If _fromIndex_ is not present, let _k_ be _len_ - 1.
+          1. Else, let _k_ be min(? ToAbsoluteIndex(_fromIndex_, _len_), _len_ - 1).
           1. Repeat, while _k_ ≥ 0,
             1. Let _kPresent_ be ! HasProperty(_O_, ! ToString(𝔽(_k_))).
             1. If _kPresent_ is *true*, then
--- a/spec.html
+++ b/spec.html
@@ -30014,11 +30014,12 @@
           1. If _targetHasLength_ is *true*, then
             1. Let _targetLen_ be ? Get(_Target_, *"length"*).
             1. If _targetLen_ is a Number, then
-              1. Set _L_ to ! ToIntegerOrInfinity(_targetLen_).
-              1. If _L_ is -∞, set _L_ to 0.
-              1. Else if _L_ is not +∞, then
+              1. Let _targetLenAsInt_ be ! ToIntegerOrInfinity(_targetLen_).
+              1. If _targetLenAsInt_ is +∞, set _L_ to +∞.
+              1. Else if _targetLenAsInt_ is -∞, set _L_ to 0.
+              1. Else,
                 1. Let _argCount_ be the number of elements in _args_.
-                1. Set _L_ to max(_L_ - _argCount_, 0).
+                1. Set _L_ to max(_targetLenAsInt_ - _argCount_, 0).
           1. Perform SetFunctionLength(_F_, _L_).
           1. Let _targetName_ be ? Get(_Target_, *"name"*).
           1. If _targetName_ is not a String, set _targetName_ to the empty String.
--- a/spec.html
+++ b/spec.html
@@ -5562,6 +5562,29 @@
           1. Return _integer_.
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-toclampedindex" type="abstract operation">
+      <h1>
+        ToClampedIndex (
+          _value_: an ECMAScript language value,
+          _length_: a non-negative integer,
+          optional _lowerBound_: an integer,
+          optional _upperBound_: an integer,
+        ): either a normal completion containing an integer or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It converts _value_ to an integer (interpreting negative values relative to _length_) and clamps the result to the inclusive interval from _lowerBound_ to _upperBound_ (respectively defaulting to 0 and _length_).</dd>
+      </dl>
+      <emu-alg>
+        1. If _lowerBound_ is not present, set _lowerBound_ to 0.
+        1. If _upperBound_ is not present, set _upperBound_ to _length_.
+        1. Assert: _lowerBound_ ≤ _upperBound_.
+        1. Let _integer_ be ? ToIntegerOrInfinity(_value_).
+        1. If _integer_ is finite and _integer_ &lt; 0, set _integer_ to _length_ + _integer_.
+        1. Return the result of clamping _integer_ between _lowerBound_ and _upperBound_.
+      </emu-alg>
+    </emu-clause>
   </emu-clause>
 
   <emu-clause id="sec-testing-and-comparison-operations">
@@ -33737,11 +33760,7 @@ THH:mm:ss.sss
           1. Let _O_ be ? RequireObjectCoercible(*this* value).
           1. Let _S_ be ? ToString(_O_).
           1. Let _len_ be the length of _S_.
-          1. Let _relativeIndex_ be ? ToIntegerOrInfinity(_index_).
-          1. If _relativeIndex_ ≥ 0, then
-            1. Let _k_ be _relativeIndex_.
-          1. Else,
-            1. Let _k_ be _len_ + _relativeIndex_.
+          1. Let _k_ be ? ToClampedIndex(_index_, _len_, -1).
           1. If _k_ &lt; 0 or _k_ ≥ _len_, return *undefined*.
           1. Return the substring of _S_ from _k_ to _k_ + 1.
         </emu-alg>
@@ -34304,14 +34323,8 @@ THH:mm:ss.sss
           1. Let _O_ be ? RequireObjectCoercible(*this* value).
           1. Let _S_ be ? ToString(_O_).
           1. Let _len_ be the length of _S_.
-          1. Let _intStart_ be ? ToIntegerOrInfinity(_start_).
-          1. If _intStart_ = -∞, let _from_ be 0.
-          1. Else if _intStart_ &lt; 0, let _from_ be max(_len_ + _intStart_, 0).
-          1. Else, let _from_ be min(_intStart_, _len_).
-          1. If _end_ is *undefined*, let _intEnd_ be _len_; else let _intEnd_ be ? ToIntegerOrInfinity(_end_).
-          1. If _intEnd_ = -∞, let _to_ be 0.
-          1. Else if _intEnd_ &lt; 0, let _to_ be max(_len_ + _intEnd_, 0).
-          1. Else, let _to_ be min(_intEnd_, _len_).
+          1. Let _from_ be ? ToClampedIndex(_start_, _len_).
+          1. If _end_ is *undefined*, let _to_ be _len_; else let _to_ be ? ToClampedIndex(_end_, _len_).
           1. If _from_ ≥ _to_, return the empty String.
           1. Return the substring of _S_ from _from_ to _to_.
         </emu-alg>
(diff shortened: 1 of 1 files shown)


ljharb:
Nice, this looks a lot cleaner to me

jmdyck:
This has a conflict with main in `String.prototype.lastIndexOf`.

spec.html
```diff
@@ -5730,8 +5730,8 @@ <h1>
       </dl>
       <emu-alg>
         1. Let _len_ be ? ToIntegerOrInfinity(_argument_).
-        1. If _len_ &le; 0, return *+0*<sub>𝔽</sub>.
-        1. Return 𝔽(min(_len_, 2<sup>53</sup> - 1)).
+        1. Let _actualLen_ be the result of clamping _len_ between 0 and 2<sup>53</sup> - 1.
+        1. Return 𝔽(_len_).
```
jmdyck, line 5734:
```suggestion
        1. Return 𝔽(_actualLen_).
```

spec.html
```diff
@@ -38752,18 +38724,9 @@ <h1>%TypedArray%.prototype.copyWithin ( _target_, _start_ [ , _end_ ] )</h1>
...
-          1. If _relativeStart_ is -&infin;, let _from_ be 0.
-          1. Else if _relativeStart_ &lt; 0, let _from_ be max(_len_ + _relativeStart_, 0).
-          1. Else, let _from_ be min(_relativeStart_, _len_).
-          1. If _end_ is *undefined*, let _relativeEnd_ be _len_; else let _relativeEnd_ be ? ToIntegerOrInfinity(_end_).
-          1. If _relativeEnd_ is -&infin;, let _final_ be 0.
-          1. Else if _relativeEnd_ &lt; 0, let _final_ be max(_len_ + _relativeEnd_, 0).
-          1. Else, let _final_ be min(_relativeEnd_, _len_).
+          1. Let _to_ be ? ToClampedIndex(_target_, _len_).
+          1. Let _from_ be ? ToClampedIndex(_start_, _len_).
+          1. If _end_ is *undefined*, let _final_ be _len_; else let _final_ ? ToClampedIndex(_end_, _len_).
```
jmdyck, line 38729:
```suggestion
          1. If _end_ is *undefined*, let _final_ be _len_; else let _final_ be ? ToClampedIndex(_end_, _len_).
```

jmdyck (changes requested):

gibson042:
@jmdyck Thanks, fixed all three issues.

jmdyck (commented):
As far as I can tell, this PR is semantics-preserving (or, in one case, semantics-clarifying), and could be merged without calamity.

My comments are stylistic points that you and/or the editors might want to consider.

spec.html
```diff
@@ -5783,6 +5783,29 @@ <h1>
...
+        <dt>description</dt>
+        <dd>It converts _value_ to an integer (interpreting negative values relative to _length_) and clamps the result to the inclusive interval from _lowerBound_ to _upperBound_ (respectively defaulting to 0 and _length_).</dd>
+      </dl>
+      <emu-alg>
+        1. If _lowerBound_ is not present, set _lowerBound_ to 0.
+        1. If _upperBound_ is not present, set _upperBound_ to _length_.
+        1. Assert: _lowerBound_ &le; _upperBound_.
+        1. Let _integer_ be ? ToIntegerOrInfinity(_value_).
+        1. If _integer_ is finite and _integer_ &lt; 0, set _integer_ to _length_ + _integer_.
+        1. Return the result of clamping _integer_ between _lowerBound_ and _upperBound_.
```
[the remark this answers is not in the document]
gibson042: > E.g., a caller that needs all those transforms might say:
> 
> ```
> Let _index_ be ? ToIntegerOrInfinity(_arg_).
> Set _index_ to HandleNegativeIndex(_index_, _len_).
> Set _index_ to ClampBelow(_index_, 0).
> Set _index_ to ClampAbove(_index_, _len_ - 1).
> ```

My goal in introducing ToClampedIndex is to eliminate that kind of noisy boilerplate so the relevant aspects of operations can be more visible, and I think it becomes a sufficiently beneficial pattern to justify use in `at` etc. even though they don't need clamping. This is definitely a matter of taste, though.

spec.html
```diff
@@ -30059,13 +30082,11 @@ <h1>Function.prototype.bind ( _thisArg_, ..._args_ )</h1>
...
-              1. Else if _targetLen_ is *-&infin;*<sub>𝔽</sub>, set _L_ to 0.
-              1. Else,
-                1. Let _targetLenAsInt_ be ! ToIntegerOrInfinity(_targetLen_).
-                1. Assert: _targetLenAsInt_ is finite.
+              1. Set _L_ to ! ToIntegerOrInfinity(_targetLen_).
+              1. If _L_ is -&infin;, set _L_ to 0.
+              1. Else if _L_ is not +&infin;, then
                 1. Let _argCount_ be the number of elements in _args_.
-                1. Set _L_ to max(_targetLenAsInt_ - _argCount_, 0).
+                1. Set _L_ to max(_L_ - _argCount_, 0).
```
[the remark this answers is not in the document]
gibson042: > I don't think these changes fit in a commit whose title is "Make better use of clamping without introducing temporary aliases". There's better use of ToIntegerOrInfinity maybe, but not clamping.

Yeah, it looked a little different in an earlier iteration. I thought about removing these changes from the PR completely, and still could be persuaded to do so.

> Also, reusing the `_L_` alias for the value of `ToIntegerOrInfinity(_targetLen_)` seems a bit odd. It lets you 'skip' the case where it's positive-infinity, but I think that actually makes the algorithm a bit harder to grasp.

I'd also be open to restoring the other alias, e.g.
```suggestion
              1. Set _targetLenAsInt_ to ! ToIntegerOrInfinity(_targetLen_).
              1. If _targetLenAsInt_ is +&infin;, set _L_ to +&infin.
              1. Else if _targetLenAsInt_ is -&infin;, set _L_ to 0.
              1. Else,
                1. Let _argCount_ be the number of elements in _args_.
                1. Set _L_ to max(_targetLenAsInt_ - _argCount_, 0).
```

spec.html
```diff
@@ -33519,11 +33540,7 @@ <h1>String.prototype.at ( _index_ )</h1>
           1. Let _O_ be ? RequireObjectCoercible(*this* value).
           1. Let _S_ be ? ToString(_O_).
           1. Let _len_ be the length of _S_.
-          1. Let _relativeIndex_ be ? ToIntegerOrInfinity(_index_).
-          1. If _relativeIndex_ &ge; 0, then
-            1. Let _k_ be _relativeIndex_.
-          1. Else,
-            1. Let _k_ be _len_ + _relativeIndex_.
+          1. Let _k_ be ? ToClampedIndex(_index_, _len_, -1).
```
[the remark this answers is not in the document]
gibson042: I tried to find a name that conveyed enough information to justify use in both \*indexOfs and ats, but ultimately decided that including "clamp" was important to understand the arguments (ClampRelativeIndex was a possibility, but it seemed a little weird to refer to nonnegative indices as "relative").

spec.html
bakkot, line 30089: I find the original easier to follow than this refactoring. @gibson042 can you say more about why you are in favor of it?

spec.html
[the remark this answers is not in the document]
bakkot: I second @jmdyck's comment about this doing too many things. There's probably a factoring which can reduce some of the boilerplate, but I don't think it's a good idea to try to combine all this stuff into a single operation that can be used everywhere. Boilerplate is much better than having operations which are hard to follow.

I think the HandleNegativeIndex suggestion might be viable. Taking the example of `Array.prototype.at`, for example, right now it's

```
1. Let _relativeIndex_ be ? ToIntegerOrInfinity(_index_).
1. If _relativeIndex_ &ge; 0, then
  1. Let _k_ be _relativeIndex_.
1. Else,
  1. Let _k_ be _len_ + _relativeIndex_.
1. If _k_ &lt; 0 or _k_ &ge; _len_, return *undefined*.
```

With the current PR that becomes
```
1. Let _k_ be ? ToClampedIndex(_index_, _len_, -1).
1. If _k_ &lt; 0 or _k_ &ge; _len_, return *undefined*.
```
and you have to read quite carefully to understand that the range for the "clamped index" actually starts at _negative one_, and therefore will hit the `k < 0` case in the second line. With something like HandleNegativeIndex it could instead be
```
1. Let _relativeIndex_ be ? ToIntegerOrInfinity(_index_).
1. Let _k_ be HandleNegativeIndex(_relativeIndex_, _len_). 
1. If _k_ &lt; 0 or _k_ &ge; _len_, return *undefined*.
```
which is, I think, much easier to follow than trying to figure out what's going on with ToClampedIndex.

I could see a case for also having an operation like `ToClampedIndex` which does not take the final two optional parameters, for use in cases where it's actually doing the thing its name implies; that's a large majority of the callsites.

spec.html
gibson042: Basically covered at https://github.com/tc39/ecma262/pull/2917#discussion_r989509978 , but two reviewers raising the same issue has persuaded me to undo it.

spec.html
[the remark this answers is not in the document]
gibson042: Sure, I can see adding that for use in `at`s. I do like the extended-parameter ToClampedIndex for `lastIndexOf`s (where it supports "let _k_ be ? ToClampedIndex(_fromIndex_, _len_, -1, _len_ - 1)" rather than e.g. "let _k_ be min(ToAbsoluteIndex(_fromIndex_, _len_), _len_ - 1)"), but that's only two sites. What are your thoughts on those?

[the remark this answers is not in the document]
bakkot: Since it's only two callsites I'm inclined not to both trying to factor that logic out, and letting ToClampedIndex be simpler.

spec.html
```diff
@@ -5738,8 +5738,8 @@ <h1>
       </dl>
       <emu-alg>
         1. Let _len_ be ? ToIntegerOrInfinity(_argument_).
-        1. If _len_ &le; 0, return *+0*<sub>𝔽</sub>.
-        1. Return 𝔽(min(_len_, 2<sup>53</sup> - 1)).
+        1. Let _clampedLen_ be the result of clamping _len_ between 0 and 2<sup>53</sup> - 1.
```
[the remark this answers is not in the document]
gibson042: 0 and 2<sup>53</sup> here are not defining an [_interval_](https://tc39.es/ecma262/multipage/notational-conventions.html#interval), but rather arguments for ["the result of _clamping_ x between _lower_ and _upper_"](https://tc39.es/ecma262/multipage/notational-conventions.html#clamping).

spec.html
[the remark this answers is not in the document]
gibson042: Okay, done.

spec.html
```diff
@@ -37650,12 +37653,8 @@ <h1>Array.prototype.lastIndexOf ( _searchElement_ [ , _fromIndex_ ] )</h1>
...
           1. Let _len_ be ? LengthOfArrayLike(_O_).
           1. If _len_ is 0, return *-1*<sub>𝔽</sub>.
-          1. If _fromIndex_ is present, let _n_ be ? ToIntegerOrInfinity(_fromIndex_); else let _n_ be _len_ - 1.
-          1. If _n_ is -&infin;, return *-1*<sub>𝔽</sub>.
-          1. If _n_ &ge; 0, then
-            1. Let _k_ be min(_n_, _len_ - 1).
-          1. Else,
-            1. Let _k_ be _len_ + _n_.
+          1. If _fromIndex_ is not present, let _k_ be _len_ - 1.
+          1. Else, let _k_ be min(? ToAbsoluteIndex(_fromIndex_, _len_), _len_ - 1).
```
michaelficarra, line 37657: Any reason not to make this a positive test?

```suggestion
          1. If _fromIndex_ is present, let _k_ be min(? ToAbsoluteIndex(_fromIndex_, _len_), _len_ - 1).
          1. Else, let _k_ be _len_ - 1.
```

michaelficarra (approved):

spec.html
gibson042: Not particularly, but "_If \<input> is \<negative condition>, let \<position> be \<maximum>… else let \<position> be \<treat input numerically>_" is the overwhelmingly dominant pattern in such steps of the algorithms affected by this PR (although `{Array,%TypedArray%}.prototype.lastIndexOf` are unique in their condition distinguishing absence from undefined[^1]).

How strongly do you feel about this suggestion?

[^1]: …and `String.prototype.lastIndexOf` is unique in conflating non-numeric _end-position_ input with absent/undefined, as opposed to e.g. `slice` and `endsWith` and `subarray` :upside_down_face:

spec.html
```diff
@@ -33703,15 +33718,15 @@ <h1>String.prototype.lastIndexOf ( _searchString_ [ , _position_ ] )</h1>
...
           1. Assert: If _position_ is *undefined*, then _numPos_ is *NaN*.
-          1. If _numPos_ is *NaN*, let _pos_ be +&infin;; otherwise, let _pos_ be ! ToIntegerOrInfinity(_numPos_).
-          1. Let _len_ be the length of _S_.
-          1. Let _searchLen_ be the length of _searchStr_.
-          1. Let _start_ be the result of clamping _pos_ between 0 and _len_ - _searchLen_.
+          1. If _numPos_ is *NaN*, let _start_ be _len_.
+          1. Else, let _start_ be the result of clamping ! ToIntegerOrInfinity(_numPos_) between 0 and _len_.
           1. If _searchStr_ is the empty String, return 𝔽(_start_).
-          1. For each integer _i_ such that 0 &le; _i_ &le; _start_, in descending order, do
+          1. For each integer _i_ such that 0 &le; _i_ &le; min(_start_, _len_ - _searchLen_), in descending order, do
```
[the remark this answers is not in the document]
gibson042: Subsequently addressed independently by #3660.

spec.html
```diff
@@ -36290,8 +36320,7 @@ <h1>String.prototype.endsWith ( _searchString_ [ , _endPosition_ ] )</h1>
           1. If _isRegexp_ is *true*, throw a *TypeError* exception.
           1. Set _searchString_ to ? ToString(_searchString_).
           1. Let _length_ be the length of _string_.
-          1. If _endPosition_ is *undefined*, let _position_ be _length_; else let _position_ be ? ToIntegerOrInfinity(_endPosition_).
-          1. Let _end_ be the result of clamping _position_ between 0 and _length_.
+          1. If _endPosition_ is *undefined*, let _end_ be _length_; else let _end_ be the result of clamping ? ToIntegerOrInfinity(_endPosition_) between 0 and _length_.
```
nicolo-ribaudo, line 36323: These lines start to be quite complex, with multiple steps collapsed together:
- the if condition
- the "then" assignment
- the "else" assignment
- the clamping
- the ToIntegerOrInfinity call

We don't have a limit, but I feel like this is reaching the maximum of what I am comfortable with.

spec.html
```diff
@@ -54508,13 +54448,8 @@ <h1>String.prototype.substr ( _start_, _length_ )</h1>
...
           1. Perform ? RequireObjectCoercible(_obj_).
           1. Let _string_ be ? ToString(_obj_).
-          1. Let _size_ be the length of _string_.
-          1. Let _intStart_ be ? ToIntegerOrInfinity(_start_).
-          1. If _intStart_ = -∞, set _intStart_ to 0.
-          1. Else if _intStart_ &lt; 0, set _intStart_ to max(_size_ + _intStart_, 0).
-          1. Else, set _intStart_ to min(_intStart_, _size_).
-          1. If _length_ is *undefined*, let _intLength_ be _size_; else let _intLength_ be ? ToIntegerOrInfinity(_length_).
-          1. Set _intLength_ to the result of clamping _intLength_ between 0 and _size_.
+          1. Let _intStart_ be ? ToClampedIndex(_start_, _size_).
```
nicolo-ribaudo, line 54414: `_size_` is not defined

spec.html
```diff
@@ -5765,6 +5765,41 @@ <h1>
...
+    <emu-clause id="sec-toabsoluteindex" type="abstract operation">
+      <h1>
+        ToAbsoluteIndex (
+          _value_: an ECMAScript language value,
+          _length_: a non-negative integer,
+        ): either a normal completion containing either an integer, +&infin;, or -&infin;, or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It converts _value_ to an integer, interpreting negative values relative to _length_.</dd>
```
nicolo-ribaudo: Nit: it converts `_value_` to _an integer, +inf, or -inf_ (the `ToIntegerOrInfinity` description, which is basically the same as this function except for the negative handling, mentions all of them).

nicolo-ribaudo (commented):
This PR is old enough that it predates change array by copy, we should also update those methods :)

spec.html
```diff
@@ -31884,13 +31919,12 @@ <h1>Function.prototype.bind ( _thisArg_, ..._args_ )</h1>
           1. If _targetHasLength_ is *true*, then
             1. Let _targetLength_ be ? Get(_target_, *"length"*).
             1. If _targetLength_ is a Number, then
-              1. If _targetLength_ is *+∞*<sub>𝔽</sub>, then
+              1. Let _targetLengthAsInt_ be ! ToIntegerOrInfinity(_targetLength_).
+              1. If _targetLengthAsInt_ is +∞, then
```
```suggestion
              1. If _targetLengthAsInt_ = +∞, then
```

spec.html
```diff
@@ -31884,13 +31919,12 @@ <h1>Function.prototype.bind ( _thisArg_, ..._args_ )</h1>
...
                 1. Set _length_ to +∞.
-              1. Else if _targetLength_ is *-∞*<sub>𝔽</sub>, then
+              1. Else if _targetLengthAsInt_ is -∞, then
```
```suggestion
              1. Else if _targetLengthAsInt_ = -∞, then
```

spec.html
gibson042: Fixed, and also opened https://github.com/tc39/ecmarkup/issues/715 .

spec.html
gibson042: Agreed, this is also about the limit for me unless we pursue #2919.

jmdyck (commented):
No complaints from ecmaspeak, but I haven't looked that closely at the changes.

nicolo-ribaudo (approved):

PR #2918, opened 2022-09-27
Editorial: Add a missing check in String.prototype.substring

gibson042:
Not necessary, but (IMO) helpful.


anba:
I don't think this is necessary, the [substring](https://tc39.es/ecma262/#substring) operation already handles this case:
> ... (which is the empty String when inclusiveStart = exclusiveEnd).

ljharb:
You're right, but imo it's still much clearer when boundary/edge cases are handled explicitly.

michaelficarra:
I'm not a fan.

gibson042:
Upon reconsideration, I think the text in [substring](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#substring) is sufficient.

PR #2920, opened 2022-09-29
Fixed bugs in return type of CatchClauseEvaluation algorithm

jhnaldo:
We found the return type bug in the syntax-directed operation: [14.15.2 Runtime semantics: CatchClauseEvaluation](https://tc39.es/ecma262/#sec-runtime-semantics-catchclauseevaluation) via [ESMeta](https://github.com/es-meta/esmeta) type analyzer.

```
       Runtime Semantics: CatchClauseEvaluation (
         _thrownValue_: an ECMAScript language value,
```

In step 7 for ``Catch : `catch` `(` CatchParameter `)` Block`` case, the variable `B` stores the result of `Evaluation of |Block|`. According to [14.2.2 Runtime Semantics: Evaluation](https://tc39.es/ecma262/#sec-block-runtime-semantics-evaluation), the variable `B` might be a normal completion containing `~empty~`. Then, step 9 returns the value stored in the variable `B`.

Similarly, in step 1 for ``Catch : `catch` Block`` case, it directly returns the result of `Evaluation of |Block|`.

```diff
         Runtime Semantics: CatchClauseEvaluation (
           _thrownValue_: an ECMAScript language value,
-        ): either a normal completion containing an ECMAScript language value or an abrupt completion
+        ): either a normal completion containing either an ECMAScript language value or ~empty~, or an abrupt completion
```

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -22912,7 +22912,7 @@
       <h1>
         Runtime Semantics: CatchClauseEvaluation (
           _thrownValue_: an ECMAScript language value,
-        ): either a normal completion containing an ECMAScript language value or an abrupt completion
+        ): either a normal completion containing either an ECMAScript language value or ~empty~, or an abrupt completion
       </h1>
       <dl class="header">
       </dl>


jhnaldo:
We will merge this PR to #2920.

jhnaldo:
I will request a new PR.

ljharb:
What was wrong with this one?

jhnaldo:
Since our tool detected other similar bugs, I wanted to update this PR. But, I accidentally removed my branch. So, I reopened another PR with other detected bugs: #2924.

PR #2922, opened 2022-09-29
Fixed bugs in return type of EvaluateClassStaticBlockBody algorithm

jhnaldo:
We found the return type bug in the syntax-directed operation: [15.7.12 Runtime Semantics: EvaluateClassStaticBlockBody](https://tc39.es/ecma262/#sec-runtime-semantics-evaluateclassstaticblockbody) via [ESMeta](https://github.com/es-meta/esmeta) type analyzer.

```
Runtime Semantics: EvaluateClassStaticBlockBody (
  _functionObject_: a function object,
```

In step 2 for, it returns the result of `Evaluation of |ClassStaticBlockStatementList|`. Because of the implicit definition of operations for chain productions (`ClassStaticBlockStatementList` -> `StatementList` -> `StatementListItem` -> `Statement` -> `BlockStatement` -> `Block`), the `Evaluation of |Block|` might be invoked, and it returns a normal completion containing `~empty~`. Thus, `CatchClauseEvaluation` algorithm could return `a normal completion containing ~empty~`.

```diff
         Runtime Semantics: EvaluateClassStaticBlockBody (
           _functionObject_: a function object,
-        ): either a normal completion containing an ECMAScript language value or an abrupt completion
+        ): either a normal completion containing either an ECMAScript language value or ~empty~, or an abrupt completion
```

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -24772,7 +24772,7 @@
       <h1>
         Runtime Semantics: EvaluateClassStaticBlockBody (
           _functionObject_: a function object,
-        ): either a normal completion containing an ECMAScript language value or an abrupt completion
+        ): either a normal completion containing an either ECMAScript language value or ~empty~, or an abrupt completion
       </h1>
       <dl class="header">
       </dl>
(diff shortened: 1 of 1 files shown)


jhnaldo:
We will merge this PR to #2920.

PR #2936, opened 2022-10-16
Editorial: Improve integer index operations

gibson042:
* Rename StringGetOwnProperty to StringGetIndexPropertyDescriptor
* Rename IsValidIntegerIndex to IntegerIndexedElementExists
* Introduce IsIntegerIndexInRange to support both

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -14508,7 +14508,8 @@
           1. If _P_ is a String, then
             1. Let _numericIndex_ be CanonicalNumericIndexString(_P_).
             1. If _numericIndex_ is not *undefined*, then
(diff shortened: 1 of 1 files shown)


ljharb:
What’s the improvement here?

gibson042:
The name of "StringGetOwnProperty" suggests that it gets any own property, when in reality it is limited to index properties (and returns a descriptor rather than just a value). Similarly, "IsValidIntegerIndex" suggests a static test of arbitrary integer index properties, when in reality it is limited to those of a specific typed array instance. Both of those are renamed for clarity, and "IsIntegerIndexInRange" is introduced as a generic operation backing them both (basically, what the "IsValidIntegerIndex" name suggests).

PR #2942, opened 2022-10-24
Editorial: support built-in async functions

michaelficarra:
This adds support for built-in async functions, as needed by the [iterator helpers](https://github.com/tc39/proposal-iterator-helpers) and [`Array.fromAsync`](https://github.com/tc39/proposal-array-from-async) proposals.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13866,7 +13866,7 @@
     <p>The behaviour specified for each built-in function via algorithm steps or other means is the specification of the function body behaviour for both [[Call]] and [[Construct]] invocations of the function. However, [[Construct]] invocation is not supported by all built-in functions. For each built-in function, when invoked with [[Call]], the [[Call]] _thisArgument_ provides the *this* value, the [[Call]] _argumentsList_ provides the named parameters, and the NewTarget value is *undefined*. When invoked with [[Construct]], the *this* value is uninitialized, the [[Construct]] _argumentsList_ provides the named parameters, and the [[Construct]] _newTarget_ parameter provides the NewTarget value. If the built-in function is implemented as an ECMAScript function object then this specified behaviour must be implemented by the ECMAScript code that is the body of the function. Built-in functions that are ECMAScript function objects must be strict functions. If a built-in constructor has any [[Call]] behaviour other than throwing a *TypeError* exception, an ECMAScript implementation of the function must be done in a manner that does not cause the function's [[IsClassConstructor]] internal slot to have the value *true*.</p>
     <p>Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function. When a built-in constructor is called as part of a `new` expression the _argumentsList_ parameter of the invoked [[Construct]] internal method provides the values for the built-in constructor's named parameters.</p>
     <p>Built-in functions that are not constructors do not have a *"prototype"* property unless otherwise specified in the description of a particular function.</p>
-    <p>If a built-in function object is not implemented as an ECMAScript function it must provide [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
+    <p>Unless otherwise specified, a built-in function object must provide [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
 
     <emu-clause id="sec-built-in-function-objects-call-thisargument-argumentslist" type="internal method">
       <h1>
@@ -13938,7 +13938,10 @@
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
         1. Let _internalSlotsList_ be a List containing the names of all the internal slots that <emu-xref href="#sec-built-in-function-objects"></emu-xref> requires for the built-in function object that is about to be created.
         1. Append to _internalSlotsList_ the elements of _additionalInternalSlotsList_.
-        1. Let _func_ be a new built-in function object that, when called, performs the action described by _behaviour_ using the provided arguments as the values of the corresponding parameters specified by _behaviour_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
+        1. If _behaviour_ is described as async, then
+          1. Let _func_ be a new built-in async function object that, when called, performs the action described by _behaviour_ using the provided arguments as the values of the corresponding parameters specified by _behaviour_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
+        1. Else,
+          1. Let _func_ be a new built-in function object that, when called, performs the action described by _behaviour_ using the provided arguments as the values of the corresponding parameters specified by _behaviour_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
         1. Set _func_.[[Prototype]] to _prototype_.
         1. Set _func_.[[Extensible]] to *true*.
         1. Set _func_.[[Realm]] to _realm_.
@@ -13951,6 +13954,57 @@
         1. Return _func_.
       </emu-alg>
       <p>Each built-in function defined in this specification is created by calling the CreateBuiltinFunction abstract operation.</p>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-built-in-async-function-objects">
+    <h1>Built-in Async Function Objects</h1>
+    <p><dfn variants="built-in async function object,built-in async function objects">Built-in async function objects</dfn> are built-in function objects that provide alternative [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
+
+    <emu-clause id="sec-built-in-async-function-objects-call" type="internal method">
+      <h1>
+        [[Call]] (
+          _thisArgument_: an ECMAScript language value,
+          _argumentsList_: a List of ECMAScript language values,
+        ): a normal completion containing an ECMAScript language value
+      </h1>
+      <dl class="header">
+        <dt>for</dt>
+        <dd>a built-in async function object _F_</dd>
+      </dl>
+      <emu-alg>
+        1. Let _callerContext_ be the running execution context.
+        1. If _callerContext_ is not already suspended, suspend _callerContext_.
+        1. Let _calleeContext_ be a new execution context.
+        1. Set the Function of _calleeContext_ to _F_.
+        1. Let _calleeRealm_ be _F_.[[Realm]].
+        1. Set the Realm of _calleeContext_ to _calleeRealm_.
+        1. Set the ScriptOrModule of _calleeContext_ to *null*.
+        1. Perform any necessary implementation-defined initialization of _calleeContext_.
+        1. Push _calleeContext_ onto the execution context stack; _calleeContext_ is now the running execution context.
+        1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+        1. Let _resultsClosure_ be a new Abstract Closure that captures _F_, _thisArgument_, and _argumentsList_ and performs the following steps when called:
+          1. Return the result of evaluating _F_ in a manner that conforms to the specification of _F_. _thisArgument_ is the *this* value, _argumentsList_ provides the named parameters, and the NewTarget value is *undefined*.
+        1. Perform AsyncFunctionStart(_promiseCapability_, _resultsClosure_).
+        1. Remove _calleeContext_ from the execution context stack and restore _callerContext_ as the running execution context.
+        1. Return _promiseCapability_.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-built-in-async-function-objects-construct" type="internal method">
+      <h1>
+        [[Construct]] (
+          _argumentsList_: a List of ECMAScript language values,
+          _newTarget_: a constructor,
+        ): either a normal completion containing an Object or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>for</dt>
+        <dd>a built-in async function object _F_</dd>
+      </dl>
+      <emu-alg>
+        1. Throw a *TypeError* exception.
+      </emu-alg>
     </emu-clause>
   </emu-clause>
 
@@ -46053,7 +46107,7 @@ THH:mm:ss.sss
         <h1>
           AsyncFunctionStart (
             _promiseCapability_: a PromiseCapability Record,
-            _asyncFunctionBody_: a |FunctionBody| Parse Node or an |ExpressionBody| Parse Node,
+            _asyncFunctionBody_: a |FunctionBody| Parse Node, an |ExpressionBody| Parse Node, or an Abstract Closure with no parameters,
           ): ~unused~
         </h1>
         <dl class="header">
@@ -46071,7 +46125,7 @@ THH:mm:ss.sss
         <h1>
           AsyncBlockStart (
             _promiseCapability_: a PromiseCapability Record,
-            _asyncBody_: a Parse Node,
+            _asyncBody_: a Parse Node or an Abstract Closure with no parameters,
             _asyncContext_: an execution context,
           ): ~unused~
         </h1>
@@ -46081,7 +46135,13 @@ THH:mm:ss.sss
           1. Assert: _promiseCapability_ is a PromiseCapability Record.
           1. Let _runningContext_ be the running execution context.
           1. [fence-effects="user-code"] Set the code evaluation state of _asyncContext_ such that when evaluation is resumed for that execution context the following steps will be performed:
-            1. Let _result_ be Completion(Evaluation of _asyncBody_).
+            1. If _asyncBody_ is a Parse Node, then
+              1. Let _result_ be the result of evaluating _asyncBody_.
+            1. Else,
+              1. Assert: _asyncBody_ is an Abstract Closure with no parameters.
+              1. Let _result_ be _asyncBody_().
+              1. If _result_ is a normal completion, then
+                1. Set _result_ to Completion Record { [[Type]]: ~return~, [[Value]]: _result_.[[Value]], [[Target]]: ~empty~ }.
             1. Assert: If we return here, the async function either threw an exception or performed an implicit or explicit return; all awaiting is done.
             1. Remove _asyncContext_ from the execution context stack and restore the execution context that is at the top of the execution context stack as the running execution context.
             1. If _result_.[[Type]] is ~normal~, then


bakkot (approved):

spec.html
```diff
@@ -13866,7 +13866,7 @@ <h1>Built-in Function Objects</h1>
...
     <p>Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function. When a built-in constructor is called as part of a `new` expression the _argumentsList_ parameter of the invoked [[Construct]] internal method provides the values for the built-in constructor's named parameters.</p>
     <p>Built-in functions that are not constructors do not have a *"prototype"* property unless otherwise specified in the description of a particular function.</p>
-    <p>If a built-in function object is not implemented as an ECMAScript function it must provide [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
+    <p>Unless otherwise specified, a built-in function object must provide [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
```
jmdyck, line 13869: This says that a built-in function is exotic (and synchronous) unless otherwise specified. So then a built-in would be ordinary (i.e., an ECMAScript function object) only if *specified* to be so. But, of course, none of them are, because the spec explicitly says that built-ins can be ordinary or exotic.

bakkot: @jmdyck I don't understand this comment. I think I'm missing some assumption. Can you rephrase?

spec.html
```diff
@@ -13954,6 +13957,57 @@ <h1>
     </emu-clause>
   </emu-clause>
 
+  <emu-clause id="sec-built-in-async-function-objects">
+    <h1>Built-in Async Function Objects</h1>
+    <p><dfn variants="built-in async function object,built-in async function objects">Built-in async function objects</dfn> are built-in function objects that provide alternative [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
```
syg, line 13962:
```suggestion
    <p><dfn variants="built-in async function object,built-in async function objects">Built-in async function objects</dfn> are built-in function objects that provide [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
```

Nit: I don't feel like "alternative" is needed.

spec.html
```diff
@@ -13954,6 +13957,57 @@ <h1>
...
+        1. Set the Realm of _calleeContext_ to _calleeRealm_.
+        1. Set the ScriptOrModule of _calleeContext_ to *null*.
+        1. Perform any necessary implementation-defined initialization of _calleeContext_.
+        1. Push _calleeContext_ onto the execution context stack; _calleeContext_ is now the running execution context.
+        1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+        1. Let _resultsClosure_ be a new Abstract Closure that captures _F_, _thisArgument_, and _argumentsList_ and performs the following steps when called:
+          1. Return the result of evaluating _F_ in a manner that conforms to the specification of _F_. _thisArgument_ is the *this* value, _argumentsList_ provides the named parameters, and the NewTarget value is *undefined*.
+        1. Perform AsyncFunctionStart(_promiseCapability_, _resultsClosure_).
+        1. Remove _calleeContext_ from the execution context stack and restore _callerContext_ as the running execution context.
+        1. Return _promiseCapability_.
```
syg, line 13990:
```suggestion
        1. Return _promiseCapability_.[[Promise]].
```

spec.html
```diff
@@ -13938,7 +13938,10 @@ <h1>
         1. If _prototype_ is not present, set _prototype_ to _realm_.[[Intrinsics]].[[%Function.prototype%]].
         1. Let _internalSlotsList_ be a List containing the names of all the internal slots that <emu-xref href="#sec-built-in-function-objects"></emu-xref> requires for the built-in function object that is about to be created.
         1. Append to _internalSlotsList_ the elements of _additionalInternalSlotsList_.
-        1. Let _func_ be a new built-in function object that, when called, performs the action described by _behaviour_ using the provided arguments as the values of the corresponding parameters specified by _behaviour_. The new function object has internal slots whose names are the elements of _internalSlotsList_, and an [[InitialName]] internal slot.
+        1. If _behaviour_ is described as async, then
```
syg, line 13941: What's the verbiage for describing an async AC? "Let _c_ be a new async Abstract Closure"? Or "Async Abstract Closure" with capital A?

syg (approved):
lgtm % question and nits

spec.html
bakkot: I would say "async Abstract Closure", I think; the idea is what we will say "async function" or "async method" for the built-in function case (as in [proposal-iterator-helpers](https://tc39.es/proposal-iterator-helpers/#sec-asynciteratorprototype.every)).


We don't actually have any async ACs at the moment so it's not established yet.

spec.html
```diff
@@ -13954,6 +13957,57 @@ <h1>
...
+  <emu-clause id="sec-built-in-async-function-objects">
+    <h1>Built-in Async Function Objects</h1>
+    <p><dfn variants="built-in async function object,built-in async function objects">Built-in async function objects</dfn> are built-in function objects that provide alternative [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
+
+    <emu-clause id="sec-built-in-async-function-objects-call" type="internal method">
+      <h1>
+        [[Call]] (
+          _thisArgument_: an ECMAScript language value,
+          _argumentsList_: a List of ECMAScript language values,
+        ): a normal completion containing an ECMAScript language value
```
michaelficarra, line 13969: What do you think about this?

```suggestion
        ): a normal completion containing a Promise
```

spec.html
jmdyck: - 10.3 "Built-in Function Objects" says that built-in functions may be implemented as either ECMAScript function objects or as exotic function objects.

- 10.2 "ECMAScript Function Objects" says that an ECMAScript function object has the `[[Call]]` internal method defined in 10.2.1.

- 10.3 "Built-in Function Objects" says (in the status quo) that a built-in function that isn't implemented as an ECMAScript function  object must have the `[[Call]]` internal method defined in 10.3.1.

So, in the status quo, a built-in's `[[Call]]` can be either 10.2.1 or 10.3.1, and the spec leaves the choice to the implementation.

- This PR replaces the third point with a sentence that says a built-in function object must use 10.3.1 "unless otherwise specified".

- But for all of the existing built-in functions, the spec never specifies whether to use 10.3.1 or 10.2.1 (because it's leaving that choice to the implementation), so it is never "otherwise specified", so the "unless otherwise specified" clause never holds.

So this new sentence implies that all existing built-in functions must use 10.3.1.

jmdyck: (And I know that, at some point, the idea was floated to get rid of the 10.2.1 possibility for built-ins, but that's clearly not what this PR is trying to accomplish. That would presumably be a separate PR.)

michaelficarra: This is not an observable difference. I am okay with it. Besides, we wanted to eliminate all the implications (and explicit statements) that "built-ins can be ECMAScript code, too" because it doesn't matter.

jmdyck: > Besides, we wanted to eliminate all the implications (and explicit statements) that "built-ins can be ECMAScript code, too" because it doesn't matter.

But this doesn't accomplish that, it just creates a contradiction between two statements in the spec.

If you want to eliminate the idea that "built-ins can be ECMAScript code, too", that's fine, but this isn't the way to do it.

michaelficarra: I'm really not very concerned about having this contradiction in the meantime. It won't lead to misimplementation and it's not observable by user code. It will get cleaned up on its own when we remove the notion of ECMAScript code built-ins.

Jack-Works: Is there any example of built-in functions that are exotic? They all like normal functions

spec.html
syg: I personally like it, but I can see how some people would prefer all overloads of concrete methods to have the same signature.

spec.html
jmdyck: > Is there any example of built-in functions that are exotic?

In the spec? No, because the spec says that's up to the implementation.

If you're asking for an example of an implementation in which built-ins are exotic, there's v8. E.g. [here](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/src/builtins/string-endswith.tq) is its implementation of `String.p.endswith`. It looks like most/all of its builtins are implemented in the Torque programming language. Not in ECMAScript, anyhow, so they're not 'ECMAScript function objects', i.e. not ordinary functions, so they're exotic functions.

In [SpiderMonkey](https://searchfox.org/mozilla-central/source/js/src/builtin/String.cpp#2445), it's implemented in C++, so also exotic.

My guess is that built-ins are exotic in most high-performance implementations, because they're easier to optimize that way.

spec.html
jmdyck: Environment Records have lots of examples of concrete methods with different (but compatible) signatures.

E.g., the five implementations of `HasThisBinding` have 3 different return types: `*false*`, `*true*`, and `a Boolean`.

bakkot: "a Promise" SGTM

spec.html
```diff
@@ -13954,6 +13957,57 @@ <h1>
...
+        1. If _callerContext_ is not already suspended, suspend _callerContext_.
+        1. Let _calleeContext_ be a new execution context.
+        1. Set the Function of _calleeContext_ to _F_.
+        1. Let _calleeRealm_ be _F_.[[Realm]].
+        1. Set the Realm of _calleeContext_ to _calleeRealm_.
+        1. Set the ScriptOrModule of _calleeContext_ to *null*.
+        1. Perform any necessary implementation-defined initialization of _calleeContext_.
+        1. Push _calleeContext_ onto the execution context stack; _calleeContext_ is now the running execution context.
+        1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+        1. Let _resultsClosure_ be a new Abstract Closure that captures _F_, _thisArgument_, and _argumentsList_ and performs the following steps when called:
```
jmdyck, line 13986:
```suggestion
        1. Let _resultsClosure_ be a new Abstract Closure with no parameters that captures _F_, _thisArgument_, and _argumentsList_ and performs the following steps when called:
```

jmdyck (changes requested):

ptomato:
Should this include an update to [10.3, "Built-in Function Objects"](https://tc39.es/ecma262/#sec-built-in-function-objects) about the prototype of built-in async functions?

> Unless otherwise specified every built-in [function object](https://tc39.es/ecma262/#function-object) has the [%Function.prototype%](https://tc39.es/ecma262/#sec-properties-of-the-function-prototype-object) object as the initial value of its [[Prototype]] internal slot.

But async functions constructed in userland have %AsyncFunction.prototype% as their [[Prototype]], so it seems inconsistent if built-in async functions don't.

spec.html
michaelficarra: Resolution from editor call is to change the definition of ordinary object to explicitly allow these `[[Call]]` and `[[Construct]]` methods so they're all considered ordinary for now. We'll do this in a separate PR.

bakkot:
Per plenary, built-in async functions will continue to have Function.prototype as their prototype.

spec.html
michaelficarra: Resolved in #2969.

jmdyck: #2969 expands the definition of "ordinary function" to include all built-in functions (however implemented). But I don't think that resolves this, because my concern doesn't hinge on the definition of "ordinary function". If you look at my response to bakkot above, it all still holds in a post-2969 spec (except that "exotic" in the first point now becomes "non-ECMAScript").

(Yes, it was during discussion of this concern that the editors call decided to expand the definition of "ordinary", but maybe that was just to deal with the discovery that built-ins could be exotic?)

jmdyck:
Roughly speaking, the status quo says that a function is ordinary if its [[Call]] conforms to either 10.2.1 or 10.3.1. This PR defines a distinct [[Call]] for built-in async function objects, which implies that these objects are not ordinary. Is that intentional, or do you need to tweak the definition of ordinary to include them?

spec.html
jmdyck: I think PR #3007 would be a basis for resolving my concern.

syg:
Recording my thoughts here for the record. While being able to specify built-ins as async functions in spec text significantly aids spec authors and many readers, I've found that it simultaneously really hinders understanding by implementers that work in a non-self hosting implementation, like V8.

For non-self hosting implementations, the implementer would need to do a manual transform to some kind of state machine. This manual transformation process increases the likelihood of bugs and subtle non-interop (like getting the number of Promise ticks wrong or something).

There are possible paths forward. The long-term solution is for V8 to be build something so that async functions can be written as straightline code as well. The spec can also include the transformed version and offer an alternate formulation of async builtins.

js-choi:
Once tc39/proposal-array-from-async#50’s reviews are finished and it is merged, Array.fromAsync will be ready to make a pull request against this pull request for formal Stage 3 reviews, as per https://github.com/tc39/proposal-array-from-async/issues/14#issuecomment-1489520146.

However, this pull request now seems to have merge conflicts with the main ecma262 branch.

Should I write a pull request against this pull request now, or should I wait until its conflicts are resolved?

michaelficarra:
@js-choi Rebased.

spec.html
```diff
@@ -14097,6 +14100,57 @@ <h1>
     </emu-clause>
   </emu-clause>
 
+  <emu-clause id="sec-built-in-async-function-objects">
+    <h1>Built-in Async Function Objects</h1>
+    <p><dfn variants="built-in async function object,built-in async function objects">Built-in async function objects</dfn> are built-in function objects that provide [[Call]] and [[Construct]] internal methods that conform to the following definitions:</p>
```
ExE-Boss, line 14105:
```suggestion
    <p><dfn variants="built-in async function object,built-in async function objects">Built-in async function objects</dfn> are built-in function objects that provide [[Call]] internal method that conforms to the following definition:</p>
```

spec.html
```diff
@@ -14097,6 +14100,57 @@ <h1>
...
+      </h1>
+      <dl class="header">
+        <dt>for</dt>
+        <dd>a built-in async function object _F_</dd>
+      </dl>
+      <emu-alg>
+        1. Throw a *TypeError* exception.
+      </emu-alg>
+    </emu-clause>
+  </emu-clause>
```
ExE-Boss, line 14152:
```suggestion
```

ExE-Boss (commented):
This shouldn’t be present, as the presence/absence of `[[Construct]]` is used to differentiate constructors from non‑constructors and it’s observable, and all existing and future async‑like built‑in functions lack it.

spec.html
```diff
@@ -14097,6 +14100,57 @@ <h1>
...
+        1. Let _calleeContext_ be a new execution context.
+        1. Set the Function of _calleeContext_ to _F_.
+        1. Let _calleeRealm_ be _F_.[[Realm]].
+        1. Set the Realm of _calleeContext_ to _calleeRealm_.
+        1. Set the ScriptOrModule of _calleeContext_ to *null*.
+        1. Perform any necessary implementation-defined initialization of _calleeContext_.
+        1. Push _calleeContext_ onto the execution context stack; _calleeContext_ is now the running execution context.
+        1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+        1. Let _resultsClosure_ be a new Abstract Closure with no parameters that captures _F_, _thisArgument_, and _argumentsList_ and performs the following steps when called:
+          1. Return the Completion Record that is the result of evaluating _F_ in a manner that conforms to the specification of _F_. _thisArgument_ is the *this* value, _argumentsList_ provides the named parameters, and the NewTarget value is *undefined*.
```
jmdyck, line 14130:
```suggestion
          1. Return the Completion Record that is the result of evaluating _F_ in a manner that conforms to the specification of _F_. _thisArgument_ provides the *this* value, _argumentsList_ provides the named parameters, and the NewTarget value is *undefined*.
```

bakkot (approved):

jmdyck (changes requested):
I think it's fairly clear that "built-in async functions" are a subset of "built-in functions". However, there are a few things that make it less clear:

- `CreateBuiltinFunction` has a dichotomy between creating "a new built-in async function object" vs "a new built-in function object". The latter could instead be "a new built-in sync function object" (or whatever you want to call it, I don't think there's a precedent).

- `10.3.1 [[Call]]` and `10.3.2 [[Construct]]` say they are for "a built-in function object _F_". Similarly, `10.3.3 BuiltinCallOrConstruct` says that its parameter `_F_` is "a built-in function object". But presumably these are only for synchronous built-ins.

- The new section `10.4 Built-in Async Function Objects` is a sibling of `10.3 Built-in Function Objects`. You could instead make it a subsection. At the same time, it might be helpful to factor out the sync-specific parts of 10.3 and put them in a complementary subsection. That is, something like this:
```
10.3 Built-in Function Objects
     10.3.1 CreateBuiltinFunction
     10.3.2 Built-in Sync Function Objects
            10.3.2.1 [[Call]]
            10.3.2.2 [[Construct]]
            10.3.2.3 BuiltinCallOrConstruct
     10.3.3 Built-in Async Function Objects
            10.3.3.1 [[Call]]
```

michaelficarra:
Addressed comments from @jmdyck except for the relationship between the terms "async built-in function object" and "built-in function object". We'll have to figure that one out together in editor call.

syg:
Editor call: for 10.3.1 let's go with branching in the algorithm steps on whether the built-in is async or not, instead of defining a new kind of built-in function and having to answer all the other questions that entails.

michaelficarra:
Okay this should be updated as requested.

spec.html
```diff
@@ -13874,7 +13874,24 @@ <h1>
         <dd>a built-in function object _F_</dd>
       </dl>
       <emu-alg>
-        1. Return ? BuiltinCallOrConstruct(_F_, _thisArgument_, _argumentsList_, *undefined*).
+        1. If _F_ is described as “async”, then
```
jmdyck: It seems to me that this check, and the resulting substeps, would make more sense in BuiltinCallOrConstruct. That's where all the corresponding sync steps exist, so it would be easier to compare and contrast between the two. (Also, you could factor out a bunch of common steps.)

michaelficarra: Good call! Done.

spec.html
```diff
@@ -13917,10 +13917,20 @@ <h1>
...
+            1. NOTE: If _F_ is defined in this document, “the specification of _F_” is the behaviour specified for it via algorithm steps or other means.
+            1. Return ? _result_.
+          1. Perform AsyncFunctionStart(_promiseCapability_, _resultsClosure_).
+          1. Remove _calleeContext_ from the execution context stack and restore _callerContext_ as the running execution context.
+          1. Return _promiseCapability_.[[Promise]].
+        1. Else,
+          1. [id="step-call-builtin-function-result"] Let _result_ be the Completion Record that is <emu-meta effects="user-code">the result of evaluating</emu-meta> _F_ in a manner that conforms to the specification of _F_. If _thisArgument_ is ~uninitialized~, the *this* value is uninitialized; otherwise _thisArgument_ provides the *this* value. _argumentsList_ provides the named parameters. _newTarget_ provides the NewTarget value.
```
michaelficarra, line 13931: Side note: this ID isn't referenced anywhere in 262. Is there another spec that references it? I hope not. Maybe we can get rid of it.

jmdyck: Formerly, `[[Construct]]` for built-in functions was defined as "the same as `[[Call]]` [for built-in functions], except that [one step] is replaced by [a different step]", and it used  `href="#step-call-builtin-function-result"` to refer to the one step in `[[Call]]`. That cross-reference disappeared when PR #2637 factored out `BuiltinCallOrConstruct`.

But there might be another spec (or web-page, or bookmark) that references it.

spec.html
```diff
@@ -13917,10 +13917,20 @@ <h1>
...
-        1. Remove _calleeContext_ from the execution context stack and restore _callerContext_ as the running execution context.
-        1. Return ? _result_.
+        1. If _F_ is described as “async”, then
+          1. Let _promiseCapability_ be ! NewPromiseCapability(%Promise%).
+          1. Let _resultsClosure_ be a new Abstract Closure with no parameters that captures _F_, _thisArgument_, _argumentsList_, and _newTarget_ and performs the following steps when called:
+            1. Let _result_ be the Completion Record that is <emu-meta effects="user-code">the result of evaluating</emu-meta> _F_ in a manner that conforms to the specification of _F_. If _thisArgument_ is ~uninitialized~, the *this* value is uninitialized; otherwise _thisArgument_ provides the *this* value. _argumentsList_ provides the named parameters. _newTarget_ provides the NewTarget value.
+            1. NOTE: If _F_ is defined in this document, “the specification of _F_” is the behaviour specified for it via algorithm steps or other means.
+            1. Return ? _result_.
```
```suggestion
            1. Return Completion(_result_).
```
... because usually, an abstract closure that returns a completion record is explicit about it (because the implicit rules don't apply, because ACs don't declare their return type). See PR #3601.

michaelficarra: Agreed. Done.

jmdyck (commented):
It slightly bugs me that the "Push" step is factored out, but the "Remove" step isn't. You could factor it out by ending the async branch with
```
1. Let _result_ be NormalCompletion(_promiseCapability_.[[Promise]]).
```
but then there's the question of whether the AC's `_result_` shadows that outer one.

But that's mostly a matter of style. I think the PR is okay to go as is.

michaelficarra:
@jmdyck Yeah we try not to shadow. We could use different aliases if we really cared, but I think it's nice and clear how it is.

spec.html
```diff
@@ -13917,10 +13917,20 @@ <h1>
         1. Set the ScriptOrModule of _calleeContext_ to *null*.
         1. Perform any necessary implementation-defined initialization of _calleeContext_.
         1. Push _calleeContext_ onto the execution context stack; _calleeContext_ is now the running execution context.
-        1. [id="step-call-builtin-function-result"] Let _result_ be the Completion Record that is <emu-meta effects="user-code">the result of evaluating</emu-meta> _F_ in a manner that conforms to the specification of _F_. If _thisArgument_ is ~uninitialized~, the *this* value is uninitialized; otherwise _thisArgument_ provides the *this* value. _argumentsList_ provides the named parameters. _newTarget_ provides the NewTarget value.
-        1. NOTE: If _F_ is defined in this document, “the specification of _F_” is the behaviour specified for it via algorithm steps or other means.
-        1. Remove _calleeContext_ from the execution context stack and restore _callerContext_ as the running execution context.
-        1. Return ? _result_.
+        1. If _F_ is described as “async”, then
```
bakkot: On re-reading I don't love this specific wording, because I'm worried about DOM or other specs having "async functions" which aren't intended to take this branch; possibly "as an async function or async method"?

bakkot: This doesn't obviously generalize to _anonymous_ built-in functions (e.g. created from Abstract Closures), but the current wording doesn't really either - if you do `Let _F_ be CreateBuiltInFunction(...)` there's not really a place to "describe [it] as 'async'".

michaelficarra: `Let _F_ be CreateBuiltInFunction(...). _F_ is async.`

bakkot: I don't love that, but I guess if it were factored out to `CreateBuiltInAsyncFunction` or something it could be ok. At which point we could say in the description for `CreateBuiltInAsyncFunction` that it "creates a built-in async function" (and maybe have it link to this step), and then it would satisfy the alternative wording I proposed.

jmdyck: Alternatively,
- `CreateBuiltinFunction` would take another parameter, indicating whether the function is sync or async,
- the value of that param would be put in a new internal slot of the function object, and
- `BuiltinCallOrConstruct` here would check the value of that internal slot.

michaelficarra: I'd be fine with that, but you'd still have to have some sort of phrasing for [`CreateIntrinsics`](https://tc39.es/ecma262/#sec-createintrinsics) step 2. And probably more detail in the paragraph following `CreateBuiltinFunction`.

bakkot: We'd just say `, _async_ is *true* if the function is described as async and *false* otherwise`.

I mildly prefer the internal slot over the current approach.

michaelficarra: I don't see how saying "is described as async" there is better than saying it here.

bakkot: The important difference is that CreateIntrinsics only applies to things defined in our spec, not in hosts. It's feasible for us to ensure that _our_ spec only describes things as "async" if they're intended to take this path, but it's not feasible to ensure that for every host.

jmdyck: It's better in `CreateIntrinsics` because that's what `CreateIntrinsics` does: it looks at the function descriptions in sections 19 through 28 and creates function objects that reflect those descriptions, after which the object shouldn't need to know anything about its description.

To say "is described as async" here, you have to imagine that the function object `_F_` somehow remembers how it was described.

bakkot (approved):

PR #2945, opened 2022-10-25
Editorial: Remove redundant table of assignment operators

gibson042:
The relevant grammar production is already linked to by |AssignmentOperator|.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -20505,25 +20505,9 @@
         1. Let _rref_ be ? Evaluation of |AssignmentExpression|.
         1. Let _rval_ be ? GetValue(_rref_).
         1. Let _assignmentOpText_ be the source text matched by |AssignmentOperator|.
-        1. Let _opText_ be the sequence of Unicode code points associated with _assignmentOpText_ in the following table:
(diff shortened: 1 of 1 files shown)


jmdyck:
I did something like this in my first draft of PR #1961, but got suggestions to provide the table instead. See https://github.com/tc39/ecma262/pull/1961#discussion_r411822697

michaelficarra:
Yeah I still think I prefer the table lookup.

gibson042:
Having the redundant table invites gratuitous divergence, such as `**` being first in [Evaluation](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-assignment-operators-runtime-semantics-evaluation) but last in [|AssignmentOperator|](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#prod-AssignmentOperator).

bakkot:
Thanks for the PR, but the current editors all prefer the table. If you have a different suggestion feel free to re-open (or ping us).

If you'd like to fix the divergence in a separate PR that would be welcome.

PR #2946, opened 2022-10-25
Editorial: Clean up the regular expressions section

gibson042:
* Move notes for regular expression pattern semantics to their respective sections.
* Express regular expression grammar production code point sequences [contiguously](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-terminal-symbols).
* Improve note for Canonicalize.


jmdyck:
(Those separated terminals in the Pattern grammar go back to the introduction of regular expressions in ES3. Which is a little odd, because the rule that says they can be contiguous goes back even farther, to ES1.)

michaelficarra (approved):

bakkot:
I am a little worried people are going to see the `\b` and think they're somehow supposed to interpret that as an escape sequence _in the meta-language_, rather than as a literal backslash followed by a literal `b`. But maybe that's not worth worrying about?

gibson042:
I don't think I share that concern, but if it is considered substantial then I think the most conventional fix would be to define and reference a new nonterminal with productions for `b` and `B`, aligning with the many other ``` `\` nonterminal ``` grammar productions (cf. [RegularExpressionBackslashSequence](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-RegularExpressionBackslashSequence) and [TemplateCharacter](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-TemplateCharacter) and [Atom](https://tc39.es/ecma262/multipage/text-processing.html#prod-Atom)).

bakkot (approved):
I've been convinced not to care about the `\b` thing, so this LGTM as is.

syg (approved):

PR #2947, opened 2022-10-26
Editorial: For clarity, manipulate _record_.[[Field]] rather than its alias

gibson042:
Even though aliases are "reference-like" per [Algorithm Conventions](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-algorithm-conventions), it is not always clear in context that steps like "append \<value> to _listAlias_" and "remove \<value> from _listAlias_" affect the Record whose field is referenced by _listAlias_—and to be honest, the presence of many such aliases doesn't even contribute very much in my opinion. But this PR is still split into meaningful commits in case the editors disagree with a subset of its changes.

Another alternative option would be to keep the aliases but make their behavior more explicit, as is done like "Let _S_ be **a reference to** the list of waiters in _WL_." in [RemoveWaiters](https://tc39.es/ecma262/multipage/structured-data.html#sec-removewaiters).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -4502,7 +4502,8 @@
         <emu-alg>
           1. Let _db_ be a new Shared Data Block value consisting of _size_ bytes. If it is impossible to create such a Shared Data Block, throw a *RangeError* exception.
           1. Let _execution_ be the [[CandidateExecution]] field of the surrounding agent's Agent Record.
-          1. Let _eventList_ be the [[EventList]] field of the element of _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
+          1. Let _eventsRecord_ be the Agent Events Record of _execution_.[[EventsRecords]] whose [[AgentSignifier]] is AgentSignifier().
+          1. Let _eventList_ be _eventsRecord_.[[EventList]].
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -40423,7 +40418,7 @@ <h1>Map.prototype.set ( _key_, _value_ )</h1>
               1. Return _M_.
           1. If _key_ is *-0*<sub>𝔽</sub>, set _key_ to *+0*<sub>𝔽</sub>.
           1. Let _p_ be the Record { [[Key]]: _key_, [[Value]]: _value_ }.
-          1. Append _p_ to _entries_.
+          1. Append _p_ to _M_.[[MapData]].
```
bakkot, line 40421: Might as well get rid of the `_entries_` aliases entirely at this point; it seems weirder to refer to the same value in two different ways in the same algorithm.

bakkot (approved):
LGTM other than comment

michaelficarra (approved):
LGTM other than https://github.com/tc39/ecma262/pull/2947/files#r1065284972

syg (approved):

spec.html

PR #2948, opened 2022-10-26
Editorial: Use consistent algorithm steps to consume the head of a List

gibson042:
cf. [AsyncGeneratorCompleteStep](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgeneratorcompletestep) and [%ForInIteratorPrototype%.next](https://tc39.es/ecma262/multipage/ecmascript-language-statements-and-declarations.html#sec-%foriniteratorprototype%.next)
> 1. Let _element_ be the first element of _list_.
> 2. Remove the first element from _list_.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -42020,13 +42020,11 @@ THH:mm:ss.sss
         </dl>
         <emu-alg>
           1. Assert: The surrounding agent is in the critical section for _WL_.
-          1. Let _L_ be a new empty List.
           1. Let _S_ be a reference to the list of waiters in _WL_.
-          1. Repeat, while _c_ > 0 and _S_ is not empty,
-            1. Let _W_ be the first waiter in _S_.
-            1. Append _W_ to _L_.
-            1. Remove _W_ from _S_.
-            1. If _c_ is finite, set _c_ to _c_ - 1.
+          1. Let _len_ be the number of elements in _S_.
+          1. Let _n_ be min(_c_, _len_).
+          1. Let _L_ be a List whose elements are the first _n_ elements of _S_.
+          1. Remove the first _n_ elements of _S_.
           1. Return _L_.
         </emu-alg>
       </emu-clause>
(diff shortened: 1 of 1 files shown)


ljharb:
Is this common enough to warrant a prose "unshift the first element of $list" macro?

michaelficarra (approved):

gibson042:
> Is this common enough to warrant a prose "unshift the first element of $list" macro?

I don't think so; it looks like only four algorithm steps do this.
```sh
$ git grep -A1 -i '^ *1[.].*first element' spec.html | grep -i remove
spec.html-                  1. Remove the first element from _remaining_.
spec.html:              1. Let _kValue_ be the first element of _values_ and remove that element from _values_.
spec.html:              1. Let _kValue_ be the first element of _values_ and remove that element from _values_.
spec.html-          1. Remove the first element from _queue_.
```

jmdyck:
[Atomics.notify](https://tc39.es/ecma262/#sec-atomics.notify) has:
```
11.a. Let _W_ be the first agent in _S_.
11.b. Remove _W_ from the front of _S_.
```

[RemoveWaiters](https://tc39.es/ecma262/#sec-removewaiters) has:
```
4.a. Let _W_ be the first waiter in _S_.
4.c. Remove _W_ from _S_.
```
(Between them there's also `4.b. Append _W_ to _L_.` but I *think* it'd be okay to move that to after an "unshift", because `_L_` is 'local' to the AO, so can't be observed by another agent.)

gibson042:
Good points. I've updated both, and `Atomics.notify` didn't even need List manipulation at all.

bakkot (approved):
LGTM. See previously https://github.com/tc39/ecma262/pull/2137/. But I'd like to wait for review from @syg on the WaiterList parts before landing.

syg (approved):

PR #2955, opened 2022-11-10
Editorial: Clarify proxy operations with a named helper operation

gibson042:
Consolidates the exception throwing associated with trying to interact with a revoked proxy.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -5667,7 +5667,8 @@
         1. If _argument_ is not an Object, return *false*.
         1. If _argument_ is an Array exotic object, return *true*.
         1. If _argument_ is a Proxy exotic object, then
-          1. Let _proxyTarget_ be ? GetNonRevokedProxyTarget(_argument_).
+          1. Perform ? ValidateNonRevokedProxy(_argument_).
+          1. Let _proxyTarget_ be _argument_.[[ProxyTarget]].
           1. Return ? IsArray(_proxyTarget_).
         1. Return *false*.
       </emu-alg>
@@ -6522,7 +6523,8 @@
           1. Let _boundTargetFunction_ be _obj_.[[BoundTargetFunction]].
           1. Return ? GetFunctionRealm(_boundTargetFunction_).
         1. If _obj_ is a Proxy exotic object, then
-          1. Let _proxyTarget_ be ? GetNonRevokedProxyTarget(_obj_).
+          1. Perform ? ValidateNonRevokedProxy(_obj_).
+          1. Let _proxyTarget_ be _obj_.[[ProxyTarget]].
           1. Return ? GetFunctionRealm(_proxyTarget_).
         1. [id="step-getfunctionrealm-default-return"] Return the current Realm Record.
       </emu-alg>
@@ -15033,7 +15035,8 @@
         <dd>a Proxy exotic object _O_</dd>
       </dl>
       <emu-alg>
-        1. Let _target_ be ? GetNonRevokedProxyTarget(_O_).
+        1. Perform ? ValidateNonRevokedProxy(_O_).
+        1. Let _target_ be _O_.[[ProxyTarget]].
         1. Let _handler_ be _O_.[[ProxyHandler]].
         1. Assert: _handler_ is an Object.
         1. Let _trap_ be ? GetMethod(_handler_, *"getPrototypeOf"*).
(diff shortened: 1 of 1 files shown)


bakkot:
I have no idea what's up with the CI failure. It's not the actual formatting that's wrong; rather the command itself is broken:

> /home/runner/work/_temp/56612d39-a02c-4d11-a6ba-19bb33eb9618.sh: line 1: Unknown: command not found

Don't worry about that failure; we'll deal with it.

ljharb:
@bakkot the issue is that npm 9 removed `npm bin` entirely, and was promoted to `latest` today. I'll put up a PR to fix it.

spec.html
```diff
@@ -5724,9 +5724,8 @@ <h1>
         1. If _argument_ is not an Object, return *false*.
         1. If _argument_ is an Array exotic object, return *true*.
         1. If _argument_ is a Proxy exotic object, then
-          1. If _argument_.[[ProxyHandler]] is *null*, throw a *TypeError* exception.
-          1. Let _target_ be _argument_.[[ProxyTarget]].
-          1. Return ? IsArray(_target_).
+          1. Let _proxyTarget_ be ? GetNonRevokedProxyTarget(_argument_).
```
caridy, line 5727: everywhere else you call it `_target_`, except here.

spec.html
```diff
@@ -15634,6 +15619,23 @@ <h1>
...
+        GetNonRevokedProxyTarget (
+          _proxy_: a Proxy exotic object,
+        ): either a normal completion containing an Object or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It returns _proxy_.[[ProxyTarget]] unless _proxy_ has been revoked.</dd>
+      </dl>
+      <emu-alg>
+        1. Let _target_ be _proxy_.[[ProxyTarget]].
```
caridy, line 15633: should we assert `_proxy_` argument to be a proxy exotic object here?

caridy (approved):
LGTM

spec.html
gibson042: Only in Proxy-specific algorithms. IsArray and GetFunctionRealm are not specific to Proxy objects, so I prefixed the alias name in them.

michaelficarra (commented):
It's kind of strange that, in almost all the call sites, we want both the `[[ProxyTarget]]` and the `[[ProxyHandler]]` but the AO just gets the one and we rely on the post-condition of it not being revoked to get the other in the next step. Maybe it would be better for the AO to just make the assertion, like what we do with `ValidateTypedArray`.

gibson042:
I think introducing some kind of operation with "revoked" in its name rather than having that information be implicit in null [[ProxyHandler]] and/or [[ProxyTarget]] is a significant improvement provided by these changes. I had that new operation return [[ProxyTarget]] out of call-site convenience (saving a step in each algorithm), but would alternatively be willing to have it instead return nothing/boolean (e.g., `InsistNonRevokedProxy`/`IsRevokedProxy`) or a Record(-like) (e.g., `GetNonRevokedProxy`/`GetNonRevokedProxyData`).

michaelficarra:
@gibson042 Result from editor call is that we should have the AO simply throw for a revoked proxy (a la `ValidateTypedArray`), and the field accesses should be done at the call sites after the call. It's also probably a good idea to put an assert in an AO step that describes the postcondition that both slots are present and non-null.

As a follow-up, it would be nice to replace steps like

> If obj.[[ProxyHandler]] is null, throw a TypeError exception.

or 

> If obj.[[ProxyTarget]] is null, throw a TypeError exception.

with the AO, if any such steps exist.

gibson042:
Thanks @michaelficarra; done.

bakkot (approved):
Thanks, this is nice.

spec.html
```diff
@@ -15076,10 +15076,10 @@ <h1>[[GetPrototypeOf]] ( ): either a normal completion containing either an Obje
         <dd>a Proxy exotic object _O_</dd>
       </dl>
       <emu-alg>
+        1. Perform ? ValidateNonRevokedProxy(_O_).
+        1. Let _target_ be _O_.[[ProxyTarget]].
         1. Let _handler_ be _O_.[[ProxyHandler]].
-        1. If _handler_ is *null*, throw a *TypeError* exception.
         1. Assert: _handler_ is an Object.
```
michaelficarra, line 15082: These asserts seem a little redundant now.

michaelficarra (approved):

spec.html
gibson042: They're certainly not necessary, but still arguably a helpful primer for the subsequent GetObject. I have no strong feelings about them.

syg (approved):

PR #2956, opened 2022-11-10
Meta: avoid removed `npm bin`; use `npx` instead

ljharb:


bakkot (approved):
It's gonna take me a while to unlearn that.

ljharb:
No need, the default in npm 8+ is not to install, i believe.

PR #2957, opened 2022-11-12
Normative: Spec liveness constraint for template objects

syg:
The [Lit framework](https://lit.dev/) depends on the identity of template arrays being stable (and not being recreated) so long as the site might still get evaluated in the future.

This PR removes the incorrect note about observability of template array objects and tries to spec normative liveness constraints. Firstly, the collection of template objects is most definitely now observable because of WeakRefs. The intention IIRC was that _so long_ as the template site could still be evaluated, we can't GC the template object, but if the template site could no longer be evaluated (what I understand to be meant by "the Parse Node is unreachable), then we _can_ GC it.

Instead of depending on this misleading note about observability, this PR specs the liveness of Parse Nodes, and directly specs the removal from the template registry if the Parse Nodes are no longer live.

cc @leszekswirski and @justinfagnani

Also see [v8:13190](https://bugs.chromium.org/p/v8/issues/detail?id=13190)

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -12195,7 +12195,7 @@
     <emu-clause id="sec-liveness">
       <h1>Liveness</h1>
 
-      <p>For some set of objects and Parse Nodes _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref of a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
+      <p>For some set of objects _S_, a <dfn>hypothetical WeakRef-oblivious</dfn> execution with respect to _S_ is an execution whereby the abstract operation WeakRefDeref of a WeakRef whose referent is an element of _S_ always returns *undefined*.</p>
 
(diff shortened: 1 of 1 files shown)


syg:
Would appreciate reviews from Kevin and Waldemar, who've thought deeply about our liveness definition in the past.

waldemarhorwat:
What does it mean to "observe the identity" of a Parse Node in any valid Future Execution? If this means "same Parse Node" comparisons, then *GetTemplateObject* always observes the identities of *all* Parse Nodes in a Realm, so how can any cease to be live?

syg:
@waldemarhorwat I had intended "observe the identity" of a Parse Node to mean the "same Parse Node" comparison. But that indeed is buggy as written, because the template registry is specced as an associative list, this means GetTemplateObject observes the identity of _every_ Parse Node in the [[TemplateMap]].

My intention here was to only count the Parse Node passed as the argument to GetTemplateObject as "observed". I am not sure what the best way to capture that intention is currently.

PR #2958, opened 2022-11-17
Editorial: Evaluation is user code by definition

bakkot:
Ecmarkup currently special-cases the phrase "the result of evaluating" as invoking user code, but as of https://github.com/tc39/ecma262/pull/2744 we barely use that terminology and instead invoke the Evaluation SDO. But we forgot to mark that SDO as invoking user code, despite that being definitionally what it is.

I recall discussing that we'd need to do this, but either I dreamt that or we forgot. Either way, whoops.


michaelficarra (approved):

syg (approved):
Oops!

PR #2960, opened 2022-11-20
Editorial: Make more use of sign-preserving truncate

gibson042:


jmdyck:
I think you can also use it in BigInt::divide.

michaelficarra (approved):

syg (approved):

PR #2969, opened 2022-12-06
Editorial: consider built-in functions to be ordinary objects

michaelficarra:
This should implement the resolution to @jmdyck's concern in https://github.com/tc39/ecma262/pull/2942#discussion_r1014498722.


spec.html
```diff
@@ -2813,10 +2813,10 @@ <h1>Object Internal Methods and Internal Slots</h1>
             For the internal methods listed in <emu-xref href="#table-essential-internal-methods"></emu-xref>, the object uses those defined in <emu-xref href="#sec-ordinary-object-internal-methods-and-internal-slots"></emu-xref>.
           </li>
           <li>
-            If the object has a [[Call]] internal method, it uses the one defined in <emu-xref href="#sec-ecmascript-function-objects-call-thisargument-argumentslist"></emu-xref>.
+            If the object has a [[Call]] internal method, it uses the one defined in either <emu-xref href="#sec-ecmascript-function-objects-call-thisargument-argumentslist"></emu-xref> or <emu-xref href="#sec-built-in-function-objects-call-thisargument-argumentslist"></emu-xref>.
```
bakkot, line 2816: This grammar is a bit weird. "It uses the one defined in either X or Y" sounds like there is one thing which is defined in one of two places, as opposed to there being two things where either choice is acceptable.

```suggestion
            If the object has a [[Call]] internal method, it uses either the one defined in <emu-xref href="#sec-ecmascript-function-objects-call-thisargument-argumentslist"></emu-xref> or the one defined in <emu-xref href="#sec-built-in-function-objects-call-thisargument-argumentslist"></emu-xref>.
```

jmdyck (changes requested):
Given this change, 10.3 "Built-in Function Objects" is now incorrect. It says that a built-in function may be implemented as either:
- an ECMAScript function object (using 10.2.1 [[Call]]), or
- a function **exotic** object, using 10.3.1 [[Call]].

But in this PR, an object using 10.3.1 [[Call]] is no longer exotic.

10.3 has 3 occurrences of "exotic". I suggest the following fixes:
- Delete the first.
- Change "is implemented as an exotic object" to "is not implemented as an ECMAScript function object".
- Delete the third.

(Interestingly, the resulting wording is also compatible with the status quo's definition of ordinary function.)

----

In the status quo, "ordinary function" means the same as "ECMAScript function",
so "Ordinary" is used in the names of various operations that are specific to ECMAScript functions:
- PrepareForOrdinaryCall
- OrdinaryCallBindThis
- OrdinaryCallEvaluateBody
- OrdinaryFunctionCreate

But in this PR, ECMAScript functions become a subset of ordinary functions, so using "Ordinary" in these contexts is misleading, so the operations should maybe be renamed. Possibly:
- "OrdinaryCall" -> "ECMAScriptFunctionCall"
- OrdinaryFunction" -> "ECMAScriptFunction"

bakkot (approved):

jmdyck:
> This should implement the resolution to @jmdyck's concern in [#2942 (comment)](https://github.com/tc39/ecma262/pull/2942#discussion_r1014498722).

As I point out there, I don't think it resolves my concern.

(Which might affect whether you want to merge it, but might not.)

PR #2965, opened 2022-11-29
Normative: Treat the Atomics.wait timeout as an extended mathematical lower bound

gibson042:
Fixes #2914

PR #2973, opened 2022-12-15
Normative: Allow implementations to pad timeouts in SuspendAgent

### LINKED ISSUE #2914: Atomics.wait and SuspendAgent disagree about the domain of valid timeouts
**Description:**
[`Atomics.wait`](https://tc39.es/ecma262/multipage/structured-data.html#sec-atomics.wait) sets _t_ to either +∞ or a nonnegative mathematical value that may have a fractional component, and if [SuspendAgent](https://tc39.es/ecma262/multipage/structured-data.html#sec-suspendagent) is invoked it receives that value as its _timeout_. But _timeout_ is specified to be a non-negative [integer](https://tc39.es/ecma262/multipage/notational-conventions.html#integer), which excludes +∞ and non-integer mathematical values, making the invocation inconsistent.

**eshost Output:**
Of the implementations that support `Atomics.wait`, most appear to respect a fractional component but GraalJS appears to truncate it.
```sh
$ eshost -sx '
  const { now } = Date;
  const arr = new Int32Array(new SharedArrayBuffer(1024));
  for (let timeout of [99.999, 100, 100.300, 100.499]) {
    const N = 30;
    const t0 = now();
    for(i = 0; i < N; i++) Atomics.wait(arr, 0, 0, timeout);
    const duration = now() - t0;
    const totalDelay = duration - timeout * N;
    const meanDelay = totalDelay / N;
    print(`mean delay ${Math.round(meanDelay * 5) / 5} ms over ${timeout}`);
  }
'
#### engine262

ReferenceError: 'SharedArrayBuffer' is not defined

#### GraalJS
mean delay -0.8 ms over 99.999
mean delay 0.2 ms over 100
mean delay 0 ms over 100.3
mean delay -0.2 ms over 100.499

#### JavaScriptCore, SpiderMonkey, V8
mean delay 0.2 ms over 99.999
mean delay 0.2 ms over 100
mean delay 0.2 ms over 100.3
mean delay 0.2 ms over 100.499

#### Moddable XS

TypeError: Atomics.wait: main thread cannot wait
```

syg:
Closes #2914.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -42214,7 +42214,7 @@ THH:mm:ss.sss
           SuspendAgent (
             _WL_: a WaiterList,
             _W_: an agent signifier,
-            _timeout_: a non-negative integer,
+            _minimumTimeout_: a non-negative extended mathematical value,
           ): a Boolean
         </h1>
         <dl class="header">
@@ -42224,11 +42224,17 @@ THH:mm:ss.sss
           1. Assert: _W_ is equivalent to AgentSignifier().
           1. Assert: _W_ is on the list of waiters in _WL_.
           1. Assert: AgentCanSuspend() is *true*.
-          1. Perform LeaveCriticalSection(_WL_) and suspend _W_ for up to _timeout_ milliseconds, performing the combined operation in such a way that a notification that arrives after the critical section is exited but before the suspension takes effect is not lost. _W_ can notify either because the timeout expired or because it was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, and not for any other reasons at all.
+          1. Let _additionalTimeout_ be an implementation-defined non-negative mathematical value.
+          1. Let _timeout_ be _minimumTimeout_ + _additionalTimeout_.
+          1. NOTE: When _minimumTimeout_ is +∞, _timeout_ is also +∞ and the following step can terminate only by another agent calling NotifyWaiter.
+          1. Perform LeaveCriticalSection(_WL_) and suspend _W_ for up to _timeout_ milliseconds, performing the combined operation in such a way that a notification that arrives after the critical section is exited but before the suspension takes effect is not lost. _W_ can wake from suspension either because the timeout expired or because it was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, and not for any other reasons at all.
           1. Perform EnterCriticalSection(_WL_).
           1. If _W_ was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, return *true*.
           1. Return *false*.
         </emu-alg>
+        <emu-note>
+          <p>_additionalTimeout_ allows implementations to pad timeouts as necessary, such as for reducing power consumption or coarsening timer resolution to mitigate timing attacks. This value may differ from call to call of SuspendAgent.</p>
+        </emu-note>
       </emu-clause>
 
       <emu-clause id="sec-notifywaiter" type="abstract operation">


spec.html
```diff
@@ -42207,11 +42207,19 @@ <h1>
           1. Assert: _W_ is equivalent to AgentSignifier().
           1. Assert: _W_ is on the list of waiters in _WL_.
           1. Assert: AgentCanSuspend() is *true*.
-          1. Perform LeaveCriticalSection(_WL_) and suspend _W_ for up to _timeout_ milliseconds, performing the combined operation in such a way that a notification that arrives after the critical section is exited but before the suspension takes effect is not lost. _W_ can notify either because the timeout expired or because it was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, and not for any other reasons at all.
+          1. If _minimumTimeout_ is +∞, then
+            1. Perform LeaveCriticalSection(_WL_) and suspend _W_, performing the combined operation in such a way that a notification that arrives after the critical section is exited but before the suspension takes effect is not lost. _W_ can notify either because it was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, and not for any other reasons at all.
```
michaelficarra, line 42211:
```suggestion
            1. Perform LeaveCriticalSection(_WL_) and suspend _W_, performing the combined operation in such a way that a notification that arrives after the critical section is exited but before the suspension takes effect is not lost. _W_ can notify because it was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, and not for any other reasons at all.
```

michaelficarra (approved):
LGTM otherwise.

spec.html
syg: Done.

spec.html
```diff
@@ -42207,11 +42207,19 @@ <h1>
...
+          1. If _minimumTimeout_ is +∞, then
+            1. Perform LeaveCriticalSection(_WL_) and suspend _W_, performing the combined operation in such a way that a notification that arrives after the critical section is exited but before the suspension takes effect is not lost. _W_ can wake from suspension because it was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, and not for any other reasons at all.
+          1. Else,
+            1. Let _additionalTimeout_ be an implementation-defined non-negative mathematical value.
+            1. Let _timeout_ be _minimumTimeout_ + _additionalTimeout_.
+            1. Perform LeaveCriticalSection(_WL_) and suspend _W_ for up to _timeout_ milliseconds, performing the combined operation in such a way that a notification that arrives after the critical section is exited but before the suspension takes effect is not lost. _W_ can wake from suspension either because the timeout expired or because it was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, and not for any other reasons at all.
```
gibson042, line 42215: I think the core step is intricate enough that it shouldn't be repeated.
```suggestion
          1. Let _additionalTimeout_ be an implementation-defined non-negative mathematical value.
          1. Let _timeout_ be _minimumTimeout_ + _additionalTimeout_.
          1. NOTE: If _minimumTimeout_ is +∞, then _timeout_ is also +∞ and the following step can terminate only by another agent calling NotifyWaiter.
          1. Perform LeaveCriticalSection(_WL_) and suspend _W_ for up to _timeout_ milliseconds, performing the combined operation in such a way that a notification arriving after the critical section is exited but before the suspension takes effect is not lost. _W_ can wake from suspension either because the timeout expired or because it was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, and not for any other reasons at all.
```

spec.html
```diff
@@ -42207,11 +42207,19 @@ <h1>
...
           1. Perform EnterCriticalSection(_WL_).
           1. If _W_ was notified explicitly by another agent calling NotifyWaiter with arguments _WL_ and _W_, return *true*.
           1. Return *false*.
         </emu-alg>
+        <emu-note>
+          <p>_additionalTimeout_ is intended to allow implementations to pad timeouts as necessary, such as for power consumption or coarsening timer resolution for side-channel attacks. This value may differ from call to call of SuspendAgent.</p>
```
gibson042, line 42221:
```suggestion
          <p>_additionalTimeout_ allows implementations to pad timeouts, which they can use to coarsen resolution for mitigating timing attacks and/or reducing power consumption. This value may differ from call to call of SuspendAgent.</p>
```

gibson042 (commented):
I closed #2965 in favor of the fix happening here.

spec.html
syg: I adopted this but didn't take it verbatim.

spec.html
syg: Fair enough. We chatted about this in the editor call and @bakkot agrees. I've re-merged the 2 branches.

syg:
> I closed https://github.com/tc39/ecma262/pull/2965 in favor of the fix happening here.

Sorry! I didn't realize you already had a PR open before I uploaded this.

bakkot (approved):

PR #2707, opened 2022-03-23
Editorial: Improve text for "locale-sensitive" operations

gibson042:
* Standardize the spelling of "locale-sensitive"
* Align String.prototype.localeCompare with ECMA-402
* Improve the description of String.prototype.localeCompare
* Add some examples of Unicode canonical equivalence
* Capitalize "Unicode Standard"

PR #2974, opened 2022-12-15
Editorial: Increase consistency and clarity of string operations

gibson042:
There was a lot of gratuitous divergence between Encode, Decode, escape, and unescape, and also between those operations and other string operations.

Non-comprehensive improvements:
* Make better use of ParseText to get Parse Nodes or errors rather than testing syntax first.
* Align on `code {unit,point} whose numeric value is…` phrasing rather than `code {unit,point} whose value is…`.
* Increase the majority of `Let _len_ be the length of…` over alternate aliases for `_len_`.
* Increase the majority of `Repeat, while _k_ &lt; _len_` iteration over alternatives like `while _k_ ≠ _len_` or embedded `If _k_ = _len_, return…`.
* Make better use of string-concatenation accepting both strings and code units.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -28953,12 +28953,12 @@
           <dd>It performs URI encoding and escaping, interpreting _string_ as a sequence of UTF-16 encoded code points as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>. If a character is identified as unreserved in RFC 2396 or appears in _extraUnescaped_, it is not escaped.</dd>
         </dl>
         <emu-alg>
-          1. Let _strLen_ be the length of _string_.
+          1. Let _len_ be the length of _string_.
           1. Let _R_ be the empty String.
           1. Let _alwaysUnescaped_ be the string-concatenation of the ASCII word characters and *"-.!~\*'()"*.
           1. Let _unescapedSet_ be the string-concatenation of _alwaysUnescaped_ and _extraUnescaped_.
           1. Let _k_ be 0.
-          1. Repeat, while _k_ &lt; _strLen_,
+          1. Repeat, while _k_ &lt; _len_,
             1. Let _C_ be the code unit at index _k_ within _string_.
             1. If _unescapedSet_ contains _C_, then
               1. Set _k_ to _k_ + 1.
@@ -28990,16 +28990,16 @@
           <dd>It performs URI unescaping and decoding, preserving any escape sequences that correspond to Basic Latin characters in _preserveEscapeSet_.</dd>
         </dl>
         <emu-alg>
-          1. Let _strLen_ be the length of _string_.
+          1. Let _len_ be the length of _string_.
           1. Let _R_ be the empty String.
           1. Let _k_ be 0.
-          1. Repeat, while _k_ &lt; _strLen_,
+          1. Repeat, while _k_ &lt; _len_,
             1. Let _C_ be the code unit at index _k_ within _string_.
             1. If _C_ is not the code unit 0x0025 (PERCENT SIGN), then
               1. Let _S_ be the String value containing only the code unit _C_.
             1. Else,
               1. Let _start_ be _k_.
-              1. If _k_ + 2 ≥ _strLen_, throw a *URIError* exception.
+              1. If _k_ + 2 ≥ _len_, throw a *URIError* exception.
               1. Let _hexDigits_ be the substring of _string_ from _k_ + 1 to _k_ + 3.
               1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
               1. If _parseResult_ is not a Parse Node, throw a *URIError* exception.
@@ -29014,7 +29014,7 @@
                   1. Let _S_ be the substring of _string_ from _start_ to _k_ + 1.
               1. Else,
                 1. If _n_ = 1 or _n_ > 4, throw a *URIError* exception.
-                1. If _k_ + (3 × (_n_ - 1)) ≥ _strLen_, throw a *URIError* exception.
+                1. If _k_ + (3 × (_n_ - 1)) ≥ _len_, throw a *URIError* exception.
                 1. Let _Octets_ be « _B_ ».
                 1. Let _j_ be 1.
                 1. Repeat, while _j_ &lt; _n_,
@@ -47955,11 +47955,11 @@ THH:mm:ss.sss
         <p>It performs the following steps when called:</p>
         <emu-alg>
           1. Set _string_ to ? ToString(_string_).
-          1. Let _strLen_ be the length of _string_.
+          1. Let _len_ be the length of _string_.
           1. Let _R_ be the empty String.
           1. Let _unescapedSet_ be the string-concatenation of the ASCII word characters and *"@\*+-./"*.
           1. Let _k_ be 0.
-          1. Repeat, while _k_ &lt; _strLen_,
+          1. Repeat, while _k_ &lt; _len_,
             1. Let _C_ be the code unit at index _k_ within _string_.
             1. If _unescapedSet_ contains _C_, then
               1. Let _S_ be _C_.
@@ -47987,18 +47987,18 @@ THH:mm:ss.sss
         <p>It performs the following steps when called:</p>
         <emu-alg>
           1. Set _string_ to ? ToString(_string_).
-          1. Let _strLen_ be the length of _string_.
+          1. Let _len_ be the length of _string_.
           1. Let _R_ be the empty String.
           1. Let _k_ be 0.
-          1. Repeat, while _k_ &lt; _strLen_,
+          1. Repeat, while _k_ &lt; _len_,
             1. Let _C_ be the code unit at index _k_ within _string_.
             1. If _C_ is the code unit 0x0025 (PERCENT SIGN), then
               1. Let _hexDigits_ be the empty String.
               1. Let _optionalAdvance_ be 0.
-              1. If _k_ + 5 &lt; _strLen_ and the code unit at index _k_ + 1 within _string_ is the code unit 0x0075 (LATIN SMALL LETTER U), then
+              1. If _k_ + 5 &lt; _len_ and the code unit at index _k_ + 1 within _string_ is the code unit 0x0075 (LATIN SMALL LETTER U), then
                 1. Set _hexDigits_ to the substring of _string_ from _k_ + 2 to _k_ + 6.
                 1. Set _optionalAdvance_ to 5.
-              1. Else if _k_ + 2 &lt; _strLen_, then
+              1. Else if _k_ + 2 &lt; _len_, then
                 1. Set _hexDigits_ to the substring of _string_ from _k_ + 1 to _k_ + 3.
                 1. Set _optionalAdvance_ to 2.
               1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
--- a/spec.html
+++ b/spec.html
@@ -48000,7 +48000,7 @@ THH:mm:ss.sss
               1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
               1. If _parseResult_ is a Parse Node, then
                 1. Let _n_ be the MV of _parseResult_.
-                1. Set _C_ to the code unit whose numeric value is ℝ(_n_).
+                1. Set _C_ to the code unit whose numeric value is _n_.
                 1. Set _k_ to _k_ + _optionalAdvance_.
             1. Set _R_ to the string-concatenation of _R_ and _C_.
             1. Set _k_ to _k_ + 1.
--- a/spec.html
+++ b/spec.html
@@ -15685,9 +15685,9 @@
       </dl>
       <emu-alg>
         1. Assert: 0 ≤ _cp_ ≤ 0x10FFFF.
-        1. If _cp_ ≤ 0xFFFF, return the String value consisting of the code unit whose value is _cp_.
-        1. Let _cu1_ be the code unit whose value is floor((_cp_ - 0x10000) / 0x400) + 0xD800.
-        1. Let _cu2_ be the code unit whose value is ((_cp_ - 0x10000) modulo 0x400) + 0xDC00.
+        1. If _cp_ ≤ 0xFFFF, return the String value consisting of the code unit whose numeric value is _cp_.
+        1. Let _cu1_ be the code unit whose numeric value is floor((_cp_ - 0x10000) / 0x400) + 0xD800.
+        1. Let _cu2_ be the code unit whose numeric value is ((_cp_ - 0x10000) modulo 0x400) + 0xDC00.
         1. Return the string-concatenation of _cu1_ and _cu2_.
       </emu-alg>
     </emu-clause>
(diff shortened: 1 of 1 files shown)


michaelficarra (approved):

spec.html
```diff
@@ -28955,46 +28955,46 @@ <h1>
...
                   1. Set _k_ to _k_ + 1.
+                  1. If _k_ + 2 ≥ _len_, throw a *URIError* exception.
                   1. If the code unit at index _k_ within _string_ is not the code unit 0x0025 (PERCENT SIGN), throw a *URIError* exception.
-                  1. If the code units at index (_k_ + 1) and (_k_ + 2) within _string_ do not represent hexadecimal digits, throw a *URIError* exception.
-                  1. Let _B_ be the 8-bit value represented by the two hexadecimal digits at index (_k_ + 1) and (_k_ + 2).
-                  1. Set _k_ to _k_ + 2.
+                  1. Let _hexDigits_ be the substring of _string_ from _k_ + 1 to _k_ + 3.
+                  1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
+                  1. If _parseResult_ is not a Parse Node, throw a *URIError* exception.
+                  1. Let _B_ be the unsigned 8-bit value corresponding with the MV of _parseResult_.
```
jmdyck, line 28987: These re-use the aliases `_hexDigits_`, `_parseResult_`, and `_B_`, so should use `Set-to` syntax rather than `Let-be`.

spec.html
```diff
@@ -47724,26 +47724,26 @@ <h1>unescape ( _string_ )</h1>
...
+              1. If _k_ + 5 &lt; _len_ and the code unit at index _k_ + 1 within _string_ is the code unit 0x0075 (LATIN SMALL LETTER U), then
+                1. Set _hexDigits_ to the substring of _string_ from _k_ + 2 to _k_ + 6.
+                1. Set _optionalAdvance_ to 5.
+              1. Else if _k_ + 2 &lt; _len_, then
+                1. Set _hexDigits_ to the substring of _string_ from _k_ + 1 to _k_ + 3.
+                1. Set _optionalAdvance_ to 2.
+              1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
+              1. If _parseResult_ is a Parse Node, then
+                1. Let _n_ be the MV of _parseResult_.
+                1. Set _C_ to the code unit whose numeric value is ℝ(_n_).
```
jmdyck, line 47744:
```suggestion
                1. Set _C_ to the code unit whose numeric value is _n_.
```
(`_n_` is already a mathematical value)

jmdyck (changes requested):

spec.html
gibson042: I consider these to be separate aliases in a deeper scope that shadow those in the outer scope, which has the joint benefits of a) matching likely implementation where shadowing is available, and b) allowing reuse of the pattern without gratuitous variation.

jmdyck: > I consider these to be separate aliases in a deeper scope that shadow those in the outer scope,

That's reasonable from a programming language point of view, but I don't think there's anything in the spec that supports it. E.g., 5.2 "Algorithm Conventions" doesn't say anything about scopes or shadowing. It *could*, but my guess is, that would be more bother than it's worth.

Anyway, I'll leave it to the editors.

bakkot: I agree with @jmdyck; we don't currently define any scoping rules except for Abstract Closures, and it's not worth the complexity of doing so vs just picking different aliases or (my preference here) using `Set`.

Also, nested scoping would conflict with the current pattern of
```
1. If condition,
  1. Let _x_ be 1.
1. Else,
  1. Let _x_ be 2.
1. Use _x_.
```

gibson042: Sorry, I meant "deeper scope" from a _logical_ perspective rather than algorithmic. AFAICT from [Algorithm Conventions](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-algorithm-conventions), `Let _x_ be …` after a preceding `Let _x_ be …` is not an error and has the effect of declaring a new alias that shadows the previous one temporally rather than lexically (i.e., "_may be referenced in any subsequent steps_"), which in the case of this algorithm are indistinguishable anyway because these declarations are in the same loop body (itself also supporting the "temporal shadowing" interpretation because each iteration encounters the same `Let` steps) and there are no references at decreasing indentation.

As noted above, I think that switching to `Set` would result in gratuitous variation for identical logic:
```diff
               1. Let _hexDigits_ be the substring of _string_ from _k_ + 1 to _k_ + 3.
               1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
               1. If _parseResult_ is not a Parse Node, throw a *URIError* exception.
               1. Let _B_ be the unsigned 8-bit value corresponding with the MV of _parseResult_.
               1. …
               1. If _n_ = 0, then
                 1. …
               1. Else,
                 1. …
                 1. Repeat, while _j_ &lt; _n_,
                   1. …
-                  1. Let _hexDigits_ be the substring of _string_ from _k_ + 1 to _k_ + 3.
-                  1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
+                  1. Set _hexDigits_ to the substring of _string_ from _k_ + 1 to _k_ + 3.
+                  1. Set _parseResult_ to ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
                   1. If _parseResult_ is not a Parse Node, throw a *URIError* exception.
-                  1. Let _B_ be the unsigned 8-bit value corresponding with the MV of _parseResult_.
+                  1. Set _B_ to the unsigned 8-bit value corresponding with the MV of _parseResult_.
```

If that is indeed the editorial preference, I think I would instead encapsulate it in a micro-operation, e.g.:
```diff
-              1. Let _hexDigits_ be the substring of _string_ from _k_ + 1 to _k_ + 3.
-              1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
-              1. If _parseResult_ is not a Parse Node, throw a *URIError* exception.
-              1. Let _B_ be the unsigned 8-bit value corresponding with the MV of _parseResult_.
+              1. Let _B_ be ReadHexOctet(_string_, _k_ + 1).
+              1. If _B_ is not an integer, throw a *URIError* exception.
               1. …
               1. If _n_ = 0, then
                 1. …
               1. Else,
                 1. …
                 1. Repeat, while _j_ &lt; _n_,
                   1. …
-                  1. Let _hexDigits_ be the substring of _string_ from _k_ + 1 to _k_ + 3.
-                  1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
-                  1. If _parseResult_ is not a Parse Node, throw a *URIError* exception.
-                  1. Let _B_ be the unsigned 8-bit value corresponding with the MV of _parseResult_.
+                  1. Let _continuationByte_ be ReadHexOctet(_string_, _k_ + 1).
+                  1. If _continuationByte_ is not an integer, throw a *URIError* exception.
```

bakkot: From the call today, editors are in agreement that we don't want to do this kind of shadowing. If you want to introduce a new AO here that's fine.

spec.html
```diff
@@ -28955,51 +28955,71 @@ <h1>
...
+              1. If _B_ is not an integer, throw a *URIError* exception.
               1. Set _k_ to _k_ + 2.
               1. Let _n_ be the number of leading 1 bits in _B_.
               1. If _n_ = 0, then
-                1. Let _C_ be the code unit whose value is _B_.
-                1. If _C_ is not in _preserveEscapeSet_, then
-                  1. Let _S_ be the String value containing only the code unit _C_.
-                1. Else,
-                  1. Let _S_ be the substring of _string_ from _start_ to _k_ + 1.
+                1. Let _c_ be the code unit whose numeric value is _B_.
```
jmdyck, line 28972: There's already a `_C_`, so `_c_` is maybe not a good choice of name.

spec.html
```diff
@@ -28955,51 +28955,71 @@ <h1>
...
+      <emu-clause id="sec-parsehexoctet" type="abstract operation">
+        <h1>
+          ParseHexOctet (
+            _string_: a String,
+            _position_: a non-negative integer,
+          ): either a non-negative integer or a non-empty List of *SyntaxError* objects
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It parses a sequence of two hexadecimal characters at the specified _position_ in _string_ into an unsigned 8-bit integer.</dd>
```
jmdyck, line 29009:
```suggestion
          <dd>It parses a sequence of two hexadecimal code units at the specified _position_ in _string_ into an unsigned 8-bit integer.</dd>
```
We mostly avoid using the word "character" outside of RegExps.

spec.html
```diff
@@ -28955,51 +28955,71 @@ <h1>
...
+        </dl>
+        <emu-alg>
+          1. Let _len_ be the length of _string_.
+          1. If _position_ + 2 > _len_, then
+            1. Let _error_ be a newly created *SyntaxError* object.
+            1. Return « _error_ ».
+          1. Let _hexDigits_ be the substring of _string_ from _position_ to _position_ + 2.
+          1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
+          1. If _parseResult_ is not a Parse Node, return _parseResult_.
+          1. Let _n_ be the unsigned 8-bit value corresponding with the MV of _parseResult_.
```
jmdyck, line 29019:
```suggestion
          1. Let _n_ be the MV of _parseResult_.
```
because MV returns an integer (in this case), and this AO is declared to return an integer.

jmdyck (changes requested):

spec.html
gibson042: I think it's fine, this alias does relate to the other one, and anyway is only used once and it's in the immediately following step. A later PR can improve _all_ the alias names in these operations.

spec.html
```diff
@@ -28974,51 +28974,71 @@ <h1>
...
+          ): either a non-negative integer or a non-empty List of *SyntaxError* objects
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It parses a sequence of two hexadecimal characters at the specified _position_ in _string_ into an unsigned 8-bit integer.</dd>
+        </dl>
+        <emu-alg>
+          1. Let _len_ be the length of _string_.
+          1. If _position_ + 2 > _len_, then
+            1. Let _error_ be a newly created *SyntaxError* object.
```
bakkot, line 29033: This is never actually reached, is it? If callers guard before calling this, it can be an assert instead.

spec.html
```diff
@@ -28974,51 +28974,71 @@ <h1>
...
+            1. Return « _error_ ».
+          1. Let _hexDigits_ be the substring of _string_ from _position_ to _position_ + 2.
+          1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
+          1. If _parseResult_ is not a Parse Node, return _parseResult_.
+          1. Let _n_ be the unsigned 8-bit value corresponding with the MV of _parseResult_.
```
bakkot, line 29038: Nit:
```suggestion
          1. Let _n_ be the unsigned 8-bit value corresponding to the MV of _parseResult_.
```
Both are technically grammatical, but I see "corresponding with" much less often, and it's more easily misunderstood.

(I see there's 9 existing uses of the "corresponding with" already in the spec, introduced in #2707 and #2781; I'm inclined to change them as well, though not in this PR.)

bakkot (approved):
LGTM other than nits.

spec.html
gibson042: Sure, done.

spec.html
gibson042: Done.

spec.html
bakkot: I also don't love this choice of names, but I agree the status quo is already bad.

spec.html
```diff
@@ -28974,51 +28974,69 @@ <h1>
...
               1. Set _k_ to _k_ + 2.
               1. Let _n_ be the number of leading 1 bits in _B_.
               1. If _n_ = 0, then
-                1. Let _C_ be the code unit whose value is _B_.
-                1. If _C_ is not in _preserveEscapeSet_, then
-                  1. Let _S_ be the String value containing only the code unit _C_.
-                1. Else,
-                  1. Let _S_ be the substring of _string_ from _start_ to _k_ + 1.
+                1. Let _c_ be the code unit whose numeric value is _B_.
+                1. If _c_ is in _preserveEscapeSet_, set _S_ to _escape_. Otherwise, set _S_ to _c_.
```
jmdyck, line 28992:
```suggestion
                1. If _preserveEscapeSet_ contains _c_, set _S_ to _escape_. Otherwise, set _S_ to _c_.
```
(PR #2877 eliminated most uses of "is in".)

spec.html
```diff
@@ -28974,51 +28974,69 @@ <h1>
...
+          <dt>description</dt>
+          <dd>It parses a sequence of two hexadecimal characters at the specified _position_ in _string_ into an unsigned 8-bit integer.</dd>
+        </dl>
+        <emu-alg>
+          1. Let _len_ be the length of _string_.
+          1. Assert: _position_ + 2 ≤ _len_.
+          1. Let _hexDigits_ be the substring of _string_ from _position_ to _position_ + 2.
+          1. Let _parseResult_ be ParseText(StringToCodePoints(_hexDigits_), |HexDigits[~Sep]|).
+          1. If _parseResult_ is not a Parse Node, return _parseResult_.
+          1. Let _n_ be the unsigned 8-bit value corresponding to the MV of _parseResult_.
```
jmdyck, line 29036:
```suggestion
          1. Let _n_ be the MV of _parseResult_.
```
The MV of a `|HexDigits|` Parse Node is always a non-negative integer, and here it's guaranteed to be in the range [0, 255], so "the unsigned 8-bit value corresponding to" doesn't do anything. (I suggested this before; I think maybe it got lost when this AO got factored out.)

It might be worth inserting `Assert: _n_ is in the inclusive interval from 0 to 255.`

jmdyck (changes requested):

gibson042:
Thanks @jmdyck; good suggestions.

jmdyck (commented):
Looks okay to me.

spec.html
```diff
@@ -28974,51 +28974,70 @@ <h1>
...
+              1. If _B_ is not an integer, throw a *URIError* exception.
               1. Set _k_ to _k_ + 2.
               1. Let _n_ be the number of leading 1 bits in _B_.
               1. If _n_ = 0, then
-                1. Let _C_ be the code unit whose value is _B_.
-                1. If _C_ is not in _preserveEscapeSet_, then
-                  1. Let _S_ be the String value containing only the code unit _C_.
-                1. Else,
-                  1. Let _S_ be the substring of _string_ from _start_ to _k_ + 1.
+                1. Let _c_ be the code unit whose numeric value is _B_.
```
syg, line 28991: Please choose a different alias than _c_ so we don't have both big _C_ and small _c_.

syg (approved):
lgtm % small comments

spec.html
gibson042: Done.

spec.html
```diff
@@ -5788,9 +5788,9 @@ <h1>
         <dd>It interprets _string_ as a sequence of UTF-16 encoded code points, as described in <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, and determines whether it is a <a href="http://www.unicode.org/glossary/#well_formed_code_unit_sequence">well formed</a> UTF-16 sequence.</dd>
       </dl>
       <emu-alg>
-        1. Let _strLen_ be the length of _string_.
+        1. Let _len_ be the length of _string_.
```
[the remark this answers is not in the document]
syg: Ah okay, thanks for the pointer.

PR #2975, opened 2022-12-15
Editorial: Improve the explanation of split methods

gibson042:
* Move note content describing special cases into the introductory \<p>.
* Rename _limit_ to _substringLimit_ to indicate that it does not apply to the count of splits (as in some other languages).
* Eliminate horizontal scrolling in a code sample.


spec.html
```diff
@@ -34306,17 +34306,17 @@ <h1>String.prototype.slice ( _start_, _end_ )</h1>
       </emu-clause>
 
       <emu-clause id="sec-string.prototype.split">
-        <h1>String.prototype.split ( _separator_, _limit_ )</h1>
-        <p>This method returns an Array into which substrings of the result of converting this object to a String have been stored. The substrings are determined by searching from left to right for occurrences of _separator_; these occurrences are not part of any String in the returned array, but serve to divide up the String value. The value of _separator_ may be a String of any length or it may be an object, such as a RegExp, that has a @@split method.</p>
+        <h1>String.prototype.split ( _separator_, _substringLimit_ )</h1>
```
ljharb, line 34309: why is "limit" any less clear than "substringLimit" in a "string split" method?

gibson042: Some languages and libraries limit the count of _splits_ while others limit the count of _substrings_. This rename provides a hint that ECMAScript is in the latter category, as opposed to e.g. [Python](https://docs.python.org/3/library/stdtypes.html#str.split) (`len("foo bar baz".split(" ", 1)) == 2`) and [JMESPath](https://jmespath.site/main/#func-split) (``` length(split('foo bar baz', ' ', `1`)) == `2` ```) and [Swift](https://developer.apple.com/documentation/swift/int/words-swift.struct/split(separator:maxsplits:omittingemptysubsequences:)) (`"foo bar baz".split(separator: " ", maxSplits: 1).count == 2`).

ljharb: I understand the difference; I'm skeptical that a) the parameter name in the specification is the source of that confusion, and that fixing it will have any impact, and b) that the majority of newcomers to JS will necessarily be familiar with languages in the former category.

gibson042: I'm not claiming that renaming a parameter in the specification will fix whatever confusion exists in the ecosystem at large, but doing so will at least prevent creating unnecessary additional confusion when reading the spec.

spec.html
```diff
@@ -34306,17 +34306,17 @@ <h1>String.prototype.slice ( _start_, _end_ )</h1>
...
+        <p>This method finds substrings separated by _separator_ in its *this* value and returns them in an Array, truncated as necessary to satisfy _substringLimit_ (in which case later contents are absent rather than collected into an element that includes _separator_). Behaviour is completely delegated to the @@split method of _separator_ if present (e.g., because _separator_ is a RegExp instance). Otherwise, a _substringLimit_ coerced to zero produces an empty Array, an *undefined* _separator_ produces a single-element Array containing only the string-coerced *this* value, any _separator_ coerced to the empty String splits the *this* value into individual code units, and all other normal completion results have at least one element (which will be the entire string-coerced *this* value if it _separator_ is not present).</p>
```
michaelficarra, line 34310:
```suggestion
        <p>This method finds substrings separated by _separator_ in its *this* value and returns them in an Array, truncated as necessary to satisfy _substringLimit_ (in which case later contents are absent rather than collected into an element that includes _separator_). Behaviour is completely delegated to the @@split method of _separator_ if present (e.g., because _separator_ is a RegExp instance). Otherwise, a _substringLimit_ coerced to zero produces an empty Array, an *undefined* _separator_ produces a single-element Array containing only the string-coerced *this* value, any _separator_ coerced to the empty String splits the *this* value into individual code units, and all other normal completion results have at least one element (which will be the entire string-coerced *this* value if _separator_ is not present).</p>
```

spec.html
```diff
@@ -36662,16 +36657,16 @@ <h1>
       </emu-clause>
 
       <emu-clause id="sec-regexp.prototype-@@split">
-        <h1>RegExp.prototype [ @@split ] ( _string_, _limit_ )</h1>
+        <h1>RegExp.prototype [ @@split ] ( _string_, _substringLimit_ )</h1>
+        <p>This method finds substrings separated by matches of the regular expression *this* value in _string_ (ignoring its *"global"*, *"lastIndex"*, and *"sticky"* properties) and returns them in an Array with any match captures interleaved, truncated as necessary to satisfy _substringLimit_ (in which case later contents are absent rather than collected into an element that includes _separator_). A _substringLimit_ coerced to zero produces an empty Array, a *this* value that can match an empty String splits _string_ into individual code units (but will still also split on longer matches), and all other normal completion results have at least one element (which will be _string_ itself if no match is found).</p>
```
michaelficarra, line 36661: > which will be _string_ itself if no match is found

Nit, but it will be the result of coercing `string` to a string. We probably don't need to capture that though.

spec.html
```diff
@@ -39925,16 +39920,16 @@ <h1>
       </emu-clause>
 
       <emu-clause oldids="sec-regexp.prototype-@@split" id="sec-regexp.prototype-%symbol.split%" type="built-in function">
-        <h1>RegExp.prototype [ %Symbol.split% ] ( _string_, _limit_ )</h1>
+        <h1>RegExp.prototype [ %Symbol.split% ] ( _string_, _substringLimit_ )</h1>
+        <p>This method finds substrings separated by matches of the regular expression *this* value in _string_ (ignoring its *"global"*, *"lastIndex"*, and *"sticky"* properties) and returns them in an Array with any match captures interleaved, truncated as necessary to satisfy _substringLimit_ (in which case later contents are absent rather than collected into an element that includes _separator_). A _substringLimit_ coerced to zero produces an empty Array, a *this* value that can match an empty String splits _string_ into individual code units (but will still also split on longer matches), and all other normal completion results have at least one element (which will be the entire string-coerced _string_ itself if no match is found).</p>
```
jmdyck: This para references `_separator_`, but that alias isn't defined here.

Ditto the (pre-existing) reference in the later emu-note.

PR #2976, opened 2022-12-16
Editorial: Highlight diff in B.3.4 compared to the original semantics

nicolo-ribaudo:
This is also done in [B.3.1](https://tc39.es/ecma262/#sec-labelled-function-declarations), [B.3.2.4](https://tc39.es/ecma262/#sec-block-duplicates-allowed-static-semantics) and [B.3.2.5](https://tc39.es/ecma262/#sec-switch-duplicates-allowed-static-semantics), and it makes it easier to understand how Annex B is modifying the original semantics.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -48246,7 +48246,7 @@ THH:mm:ss.sss
 
     <emu-annex id="sec-variablestatements-in-catch-blocks">
       <h1>VariableStatements in Catch Blocks</h1>
-      <p>The content of subclause <emu-xref href="#sec-try-statement-static-semantics-early-errors"></emu-xref> is replaced with the following:</p>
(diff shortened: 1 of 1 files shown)


michaelficarra:
@jmdyck Could we incorporate this into #2952 using the flagging approach from that PR?

jmdyck:
[PR #2952](https://github.com/tc39/ecma262/pull/2952/files#diff-181371b08d71216599b0acccbaabd03c306da6de142ea6275c2135810999805aR22630) already incorporates it.

michaelficarra:
@jmdyck Oh, I looked but must have missed it because of the big difference in line numbers.

Okay, then I am going to close this in favour of #2952, since we already decided we will take that approach.

PR #2982, opened 2022-12-18
Editorial: Use linkable terms for coercion to BigInt

gibson042:
Followup to #2960.

"**BigInt value for**" is a term, but "**BigInt value that represents**" is not (and most uses of the latter can be replaced with "ℤ" anyway). Even the ℝ-coercions are technically unnecessary because of [Mathematical Operations](https://tc39.es/ecma262/multipage/notational-conventions.html#sec-mathematical-operations) ("_When applied to BigInts, [operators such as +, ×, =, and ≥] refer to the usual mathematical operations applied to the mathematical value of the BigInt_"), but I've kept them here to avoid making the BigInt operations trivial (e.g., [BigInt::unaryMinus](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-numeric-types-bigint-unaryMinus) would be `<emu-alg>1. Return -_x_.</emu-alg>`).

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -2352,7 +2352,7 @@
           </dl>
           <emu-alg>
             1. If _x_ is *0*<sub>ℤ</sub>, return *0*<sub>ℤ</sub>.
-            1. Return the BigInt value that represents the negation of ℝ(_x_).
+            1. Return -_x_.
           </emu-alg>
         </emu-clause>
 
@@ -2383,7 +2383,7 @@
           <emu-alg>
             1. If _exponent_ &lt; *0*<sub>ℤ</sub>, throw a *RangeError* exception.
             1. If _base_ is *0*<sub>ℤ</sub> and _exponent_ is *0*<sub>ℤ</sub>, return *1*<sub>ℤ</sub>.
-            1. Return the BigInt value that represents ℝ(_base_) raised to the power ℝ(_exponent_).
+            1. Return _base_ raised to the power _exponent_.
           </emu-alg>
         </emu-clause>
 
@@ -2397,7 +2397,7 @@
           <dl class="header">
           </dl>
           <emu-alg>
-            1. Return the BigInt value that represents the product of _x_ and _y_.
+            1. Return _x_ × _y_.
           </emu-alg>
           <emu-note>Even if the result has a much larger bit width than the input, the exact mathematical answer is given.</emu-note>
         </emu-clause>
@@ -2447,7 +2447,7 @@
           <dl class="header">
           </dl>
           <emu-alg>
-            1. Return the BigInt value that represents the sum of _x_ and _y_.
+            1. Return _x_ + _y_.
           </emu-alg>
         </emu-clause>
 
@@ -2461,7 +2461,7 @@
           <dl class="header">
           </dl>
           <emu-alg>
-            1. Return the BigInt value that represents the difference _x_ minus _y_.
+            1. Return _x_ - _y_.
           </emu-alg>
         </emu-clause>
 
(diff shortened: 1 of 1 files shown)


bakkot (approved):

spec.html
```diff
@@ -2365,7 +2365,7 @@ <h1>
           <emu-alg>
             1. If _exponent_ &lt; *0*<sub>ℤ</sub>, throw a *RangeError* exception.
             1. If _base_ is *0*<sub>ℤ</sub> and _exponent_ is *0*<sub>ℤ</sub>, return *1*<sub>ℤ</sub>.
-            1. Return the BigInt value that represents ℝ(_base_) raised to the power ℝ(_exponent_).
+            1. Return the BigInt value for ℝ(_base_) raised to the power ℝ(_exponent_).
```
michaelficarra, line 2368: This will help with association of "the BigInt value for":

```suggestion
            1. Return ℤ(ℝ(_base_) raised to the power ℝ(_exponent_)).
```

Alternatively, we could make it two steps.

syg (commented):
I think we should remove the superfluous casts where not needed. I don't see these operations appearing too trivial as an issue, if that's the main concern.

spec.html
```diff
@@ -2458,8 +2458,8 @@ <h1>
           </dl>
           <emu-alg>
             1. If _y_ &lt; *0*<sub>ℤ</sub>, then
-              1. Return the BigInt value that represents ℝ(_x_) / 2<sup>-ℝ(_y_)</sup>, rounding down to the nearest integer, including for negative numbers.
-            1. Return the BigInt value that represents ℝ(_x_) × 2<sup>ℝ(_y_)</sup>.
+              1. Return ℤ(floor(ℝ(_x_) / 2<sup>-ℝ(_y_)</sup>)).
```
syg, line 2461: I think this is the only line where we need ℝ casts inside a ℤ cast.

tran09 (commented):
Nice

spec.html
tran09: > This will help with association of "the BigInt value for":
> 
> 
> 
> ```suggestion
> 
>             1. Return ℤ(ℝ(_base_) raised to the power ℝ(_exponent_)).
> 
> ```
> 
> 
> 
> Alternatively, we could make it two steps.

tran09: Help me

michaelficarra:
@gibson042 Were you planning on addressing this feedback or should we take this PR over?

gibson042:
Oh, go ahead. I'm not sure exactly what you want for the couple of open points, so please resolve them to your satisfaction.

michaelficarra:
Alright I've update the PR to not do unnecessary casts to reals.

spec.html
```diff
@@ -31356,7 +31356,7 @@ <h1>BigInt.asUintN ( _bits_, _bigint_ )</h1>
         <emu-alg>
           1. Set _bits_ to ? ToIndex(_bits_).
           1. Set _bigint_ to ? ToBigInt(_bigint_).
-          1. Return the BigInt value that represents ℝ(_bigint_) modulo 2<sup>_bits_</sup>.
+          1. Return _bigint_ modulo *2*<sub>ℤ</sub><sup>_bits_</sup>.
```
bakkot, line 31359:
```suggestion
          1. Return ℤ(ℝ(_bigint_) modulo 2<sup>_bits_</sup>).
```

michaelficarra: Accepted.

bakkot (approved):

PR #2991, opened 2023-01-03
Editorial: Tweak FunctionDeclarationInstantiation

gibson042:
* Clarify explanatory notes.
* Extract shared steps from slightly-different branches.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -13526,8 +13526,8 @@
         1. Else if _hasParameterExpressions_ is *false*, then
           1. If _functionNames_ contains *"arguments"* or _lexicalNames_ contains *"arguments"*, then
             1. Set _argumentsObjectNeeded_ to *false*.
-        1. NOTE: If direct eval calls in the formal parameter list can create bindings which are visible outside of the `eval`, then a separate Environment Record is needed to ensure that such bindings do not affect the environment where parameters are declared.
         1. If _strict_ is *false* and _hasParameterExpressions_ is *true*, then
+          1. NOTE: Direct eval calls in the formal parameter list of a non-strict function can create bindings which are visible outside of the `eval`, so a separate Environment Record is needed to ensure that such bindings do not affect the environment where parameters are declared.
           1. Let _calleeEnv_ be the LexicalEnvironment of _calleeContext_.
           1. Assert: The VariableEnvironment of _calleeContext_ is _calleeEnv_.
           1. Set the LexicalEnvironment of _calleeContext_ to NewDeclarativeEnvironment(_calleeEnv_).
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -13524,70 +13526,58 @@ <h1>
...
-              1. Perform ! _varEnv_.InitializeBinding(_n_, _initialValue_).
-              1. NOTE: A var with the same name as a formal parameter initially has the same value as the corresponding initialized parameter.
+        1. Else,
+          1. Let _varEnv_ be _env_.
+          1. Let _instantiatedVarNames_ be a copy of the List _parameterBindings_.
+        1. For each element _n_ of _varNames_, do
+          1. If _n_ is not an element of _instantiatedVarNames_, then
+            1. Append _n_ to _instantiatedVarNames_.
+            1. Perform ! _varEnv_.CreateMutableBinding(_n_, *false*).
+            1. NOTE: A var with the same name as a formal parameter initially has the same value as the corresponding initialized parameter.
```
bakkot, line 13575: This refactoring does preserve observable behavior, from what I can tell, but it also kind of obscures what's going on here. I think this note should explain what was previously implicit in the structure. Something like

```suggestion
            1. NOTE: When there are expressions in the parameter list, it is possible to observe a difference between a variable in the parameter list and a variable in the body with the same name. The variable in the body initially has the same value as the corresponding initialized parameter.
```

Though, I do think eventually we will want to drop the whole `hasParameterExpressions` thing as mentioned in https://github.com/tc39/ecma262/issues/623, and instead have a NOTE pointing out the possibility of optimizing out the distinction between the two variables when that distinction is not observable. This refactoring makes it easier to get there, so :+1:

spec.html
```diff
@@ -13524,70 +13526,58 @@ <h1>
...
         1. [id="step-functiondeclarationinstantiation-web-compat-insertion-point"] NOTE: Annex <emu-xref href="#sec-web-compat-functiondeclarationinstantiation"></emu-xref> adds additional steps at this point.
-        1. If _strict_ is *false*, then
-          1. Let _lexEnv_ be NewDeclarativeEnvironment(_varEnv_).
-          1. NOTE: Non-strict functions use a separate Environment Record for top-level lexical declarations so that a direct eval can determine whether any var scoped declarations introduced by the eval code conflict with pre-existing top-level lexically scoped declarations. This is not needed for strict functions because a strict direct eval always places all declarations into a new Environment Record.
-        1. Else, let _lexEnv_ be _varEnv_.
+        1. NOTE: Non-strict functions use a separate Environment Record for top-level lexical declarations so that a direct eval can determine whether any var scoped declarations introduced by the eval code conflict with pre-existing top-level lexically scoped declarations. This is not needed for strict functions, in which direct eval declarations never affect the outer environment.
```
bakkot, line 13579: "direct eval declarations never affect the outer environment" isn't true - it can't create new bindings, but it can alter the values of existing ones. The old wording seems fine to me.

(Sidebar: it's quite hard to notice when you move code and simultaneously tweak it. For reviewability purposes I'd appreciate it if in the future you could call such things out, or split them into distinct commits.)

spec.html
```diff
@@ -13524,70 +13526,58 @@ <h1>
         1. Else if _hasParameterExpressions_ is *false*, then
           1. If *"arguments"* is an element of _functionNames_ or if *"arguments"* is an element of _lexicalNames_, then
             1. Set _argumentsObjectNeeded_ to *false*.
-        1. If _strict_ is *true* or if _hasParameterExpressions_ is *false*, then
-          1. NOTE: Only a single Environment Record is needed for the parameters, since calls to `eval` in strict mode code cannot create new bindings which are visible outside of the `eval`.
-          1. Let _env_ be the LexicalEnvironment of _calleeContext_.
-        1. Else,
-          1. NOTE: A separate Environment Record is needed to ensure that bindings created by direct eval calls in the formal parameter list are outside the environment where parameters are declared.
+        1. NOTE: If direct eval calls in the formal parameter list can create bindings which are visible outside of the `eval`, then a separate Environment Record is needed to ensure that such bindings do not affect the environment where parameters are declared.
```
bakkot, line 13529: The previous note made it clear why strictness was relevant; this new wording requires the reader to know that strict calls to `eval` place bindings in a new inner scope. I would prefer to keep the previous notes, or to reword this so it does not require the reader to know that.

spec.html
```diff
@@ -13524,70 +13526,58 @@ <h1>
...
-            1. NOTE: A mapped argument object is only provided for non-strict functions that don't have a rest parameter, any parameter default value initializers, or any destructured parameters.
+          1. NOTE: A mapped argument object is only provided for non-strict functions with <emu-xref href="#sec-static-semantics-issimpleparameterlist">simple parameter lists</emu-xref> (which don't contain any parameter initializers, destructured parameters, or rest parameters).
+          1. If _strict_ is *false* and _simpleParameterList_ is *true*, then
             1. Let _ao_ be CreateMappedArgumentsObject(_func_, _formals_, _argumentsList_, _env_).
+          1. Else,
+            1. Let _ao_ be CreateUnmappedArgumentsObject(_argumentsList_).
           1. If _strict_ is *true*, then
             1. Perform ! _env_.CreateImmutableBinding(*"arguments"*, *false*).
-            1. NOTE: In strict mode code early errors prevent attempting to assign to this binding, so its mutability is not observable.
+            1. NOTE: In strict mode code, the immutability of this binding is not observable because early errors prevent attempting to assign to it.
```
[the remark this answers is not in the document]
gibson042: Introducing the prepositional comma improves flow IMO, but more importantly the use of "immutability" rather than "mutability" clarifies the note (because the binding is in fact immutable, making it strange—even if correct in some senses—to talk about its [lack of] mutability).

spec.html
gibson042:
```suggestion
        1. NOTE: Non-strict functions use a separate Environment Record for top-level lexical declarations so that a direct eval can determine whether any var scoped declarations introduced by the eval code conflict with pre-existing top-level lexically scoped declarations. This is not needed for strict functions, in which direct eval declarations never add bindings to the outer environment.
```

spec.html
gibson042: I thought the following condition explained that, but will clarify by moving the note into the subsequent block.

spec.html
```diff
@@ -13524,70 +13526,58 @@ <h1>
         1. Else if _hasParameterExpressions_ is *false*, then
           1. If *"arguments"* is an element of _functionNames_ or if *"arguments"* is an element of _lexicalNames_, then
             1. Set _argumentsObjectNeeded_ to *false*.
-        1. If _strict_ is *true* or if _hasParameterExpressions_ is *false*, then
-          1. NOTE: Only a single Environment Record is needed for the parameters, since calls to `eval` in strict mode code cannot create new bindings which are visible outside of the `eval`.
-          1. Let _env_ be the LexicalEnvironment of _calleeContext_.
-        1. Else,
-          1. NOTE: A separate Environment Record is needed to ensure that bindings created by direct eval calls in the formal parameter list are outside the environment where parameters are declared.
+        1. NOTE: If direct eval calls in the formal parameter list can create bindings which are visible outside of the `eval`, then a separate Environment Record is needed to ensure that such bindings do not affect the environment where parameters are declared.
+        1. If _strict_ is *false* and _hasParameterExpressions_ is *true*, then
```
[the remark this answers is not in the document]
bakkot: This still doesn't actually make it clear why strictness is relevant. It is easier to infer this way around, but I would prefer to just mention strictness in the note.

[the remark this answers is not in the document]

PR #2997, opened 2023-01-19
Normative: Adds change-array-by-copy

acutmore:
Stage 4 spec for https://github.com/tc39/proposal-change-array-by-copy

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -38188,24 +38188,21 @@ THH:mm:ss.sss
           1. Let _obj_ be ? ToObject(*this* value).
           1. [id="step-array-sort-len"] Let _len_ be ? LengthOfArrayLike(_obj_).
           1. Let _SortCompare_ be a new Abstract Closure with parameters (_x_, _y_) that captures _comparefn_ and performs the following steps when called:
-            1. If _x_ and _y_ are both *undefined*, return *+0*<sub>𝔽</sub>.
-            1. If _x_ is *undefined*, return *1*<sub>𝔽</sub>.
-            1. If _y_ is *undefined*, return *-1*<sub>𝔽</sub>.
-            1. If _comparefn_ is not *undefined*, then
-              1. Let _v_ be ? ToNumber(? Call(_comparefn_, *undefined*, « _x_, _y_ »)).
-              1. If _v_ is *NaN*, return *+0*<sub>𝔽</sub>.
-              1. Return _v_.
-            1. [id="step-sortcompare-tostring-x"] Let _xString_ be ? ToString(_x_).
-            1. [id="step-sortcompare-tostring-y"] Let _yString_ be ? ToString(_y_).
-            1. Let _xSmaller_ be ! IsLessThan(_xString_, _yString_, *true*).
-            1. If _xSmaller_ is *true*, return *-1*<sub>𝔽</sub>.
-            1. Let _ySmaller_ be ! IsLessThan(_yString_, _xString_, *true*).
-            1. If _ySmaller_ is *true*, return *1*<sub>𝔽</sub>.
-            1. Return *+0*<sub>𝔽</sub>.
-          1. Return ? SortIndexedProperties(_obj_, _len_, _SortCompare_).
+            1. Return ? CompareArrayElements(_x_, _y_, _comparefn_).
+          1. [id="step-array-sortindexedproperties"] Let _sortedList_ be ? SortIndexedProperties(_obj_, _len_, _SortCompare_, ~skip-holes~).
+          1. Let _itemCount_ be the number of elements in _sortedList_.
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _itemCount_,
+            1. Perform ? Set(_obj_, ! ToString(𝔽(_j_)), _sortedList_[_j_], *true*).
+            1. Set _j_ to _j_ + 1.
+          1. NOTE: The call to SortIndexedProperties in step <emu-xref href="#step-array-sortindexedproperties"></emu-xref> uses ~skip-holes~. The remaining indices are deleted to preserve the number of holes that were detected and excluded from the sort.
+          1. Repeat, while _j_ &lt; _len_,
+            1. Perform ? DeletePropertyOrThrow(_obj_, ! ToString(𝔽(_j_))).
+            1. Set _j_ to _j_ + 1.
+          1. Return _obj_.
         </emu-alg>
         <emu-note>
-          <p>Because non-existent property values always compare greater than *undefined* property values, and *undefined* always compares greater than any other value, *undefined* property values always sort to the end of the result, followed by non-existent property values.</p>
+          <p>Because non-existent property values always compare greater than *undefined* property values, and *undefined* always compares greater than any other value (see CompareArrayElements), *undefined* property values always sort to the end of the result, followed by non-existent property values.</p>
         </emu-note>
         <emu-note>
           <p>Method calls performed by the ToString abstract operations in steps <emu-xref href="#step-sortcompare-tostring-x"></emu-xref> and <emu-xref href="#step-sortcompare-tostring-y"></emu-xref> have the potential to cause _SortCompare_ to not behave as a consistent comparator.</p>
@@ -38220,7 +38217,8 @@ THH:mm:ss.sss
               _obj_: an Object,
               _len_: a non-negative integer,
               _SortCompare_: an Abstract Closure with two parameters,
-            ): either a normal completion containing an Object or an abrupt completion
+              _holes_: ~skip-holes~ or ~read-through-holes~,
+            ): either a normal completion containing a List of ECMAScript language values or a throw completion
           </h1>
           <dl class="header">
           </dl>
@@ -38229,21 +38227,17 @@ THH:mm:ss.sss
             1. Let _k_ be 0.
             1. Repeat, while _k_ &lt; _len_,
               1. Let _Pk_ be ! ToString(𝔽(_k_)).
-              1. Let _kPresent_ be ? HasProperty(_obj_, _Pk_).
-              1. If _kPresent_ is *true*, then
+              1. If _holes_ is ~skip-holes~, then
+                1. Let _kRead_ be ? HasProperty(_obj_, _Pk_).
+              1. Else,
+                1. Assert: _holes_ is ~read-through-holes~.
+                1. Let _kRead_ be *true*.
+              1. If _kRead_ is *true*, then
                 1. Let _kValue_ be ? Get(_obj_, _Pk_).
                 1. Append _kValue_ to _items_.
               1. Set _k_ to _k_ + 1.
-            1. Let _itemCount_ be the number of elements in _items_.
             1. [id="step-array-sort"] Sort _items_ using an implementation-defined sequence of <emu-meta effects="user-code">calls to _SortCompare_</emu-meta>. If any such call returns an abrupt completion, stop before performing any further calls to _SortCompare_ and return that Completion Record.
-            1. Let _j_ be 0.
-            1. Repeat, while _j_ &lt; _itemCount_,
-              1. Perform ? Set(_obj_, ! ToString(𝔽(_j_)), _items_[_j_], *true*).
-              1. Set _j_ to _j_ + 1.
-            1. Repeat, while _j_ &lt; _len_,
-              1. Perform ? DeletePropertyOrThrow(_obj_, ! ToString(𝔽(_j_))).
-              1. Set _j_ to _j_ + 1.
-            1. Return _obj_.
+            1. Return _items_.
           </emu-alg>
           <p>The <dfn id="sort-order">sort order</dfn> is the ordering of _items_ after completion of step <emu-xref href="#step-array-sort"></emu-xref> of the algorithm above. The sort order is implementation-defined if _SortCompare_ is not a consistent comparator for the elements of _items_. When SortIndexedProperties is invoked by <emu-xref href="#sec-array.prototype.sort">Array.prototype.sort</emu-xref>, the sort order is also implementation-defined if _comparefn_ is *undefined*, and all applications of ToString, to any specific value passed as an argument to _SortCompare_, do not produce the same result.</p>
           <p>Unless the sort order is specified to be implementation-defined, it must satisfy all of the following conditions:</p>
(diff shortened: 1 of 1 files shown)


jmdyck:
Rather than leaving 'suggested change' comments here, I've submitted a PR against this PR's branch: https://github.com/acutmore/ecma262/pull/2

spec.html
```diff
@@ -39767,6 +39966,27 @@ <h1>
         </emu-alg>
       </emu-clause>
 
+      <emu-clause id="typedarray-create-same-type" type="abstract operation">
+        <h1>
+          TypedArrayCreateSameType (
+            _exemplar_: a TypedArray,
```
ljharb, line 39972:
```suggestion
            _exemplar_: a Typed Array,
```

spec.html
```diff
@@ -39767,6 +39966,27 @@ <h1>
...
+            _argumentList_: unknown,
+          )
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It is used to specify the creation of a new TypedArray using a constructor function that is derived from _exemplar_. Unlike TypedArraySpeciesCreate, which can construct custom TypedArray subclasses through the use of @@species, this operation always uses one of the built-in TypedArray constructors.</dd>
+        </dl>
+        <emu-alg>
+          1. Assert: _exemplar_ is an Object that has [[TypedArrayName]] and [[ContentType]] internal slots.
```
ljharb, line 39981: is this needed, given the assertion in the header?

spec.html
```diff
@@ -39811,6 +40031,36 @@ <h1>
...
+          1. If _v_ is *NaN*, return *+0*<sub>𝔽</sub>.
+          1. Return _v_.
+        1. If _x_ and _y_ are both *NaN*, return *+0*<sub>𝔽</sub>.
+        1. If _x_ is *NaN*, return *1*<sub>𝔽</sub>.
+        1. If _y_ is *NaN*, return *-1*<sub>𝔽</sub>.
+        1. If _x_ &lt; _y_, return *-1*<sub>𝔽</sub>.
+        1. If _x_ > _y_, return *1*<sub>𝔽</sub>.
+        1. If _x_ is *-0*<sub>𝔽</sub> and _y_ is *+0*<sub>𝔽</sub>, return *-1*<sub>𝔽</sub>.
+        1. If _x_ is *+0*<sub>𝔽</sub> and _y_ is *-0*<sub>𝔽</sub>, return *1*<sub>𝔽</sub>.
+        1. Return *+0*<sub>𝔽</sub>.
```
ljharb, line 40057: is there a reason this can't use `IsLessThan`, like `CompareArrayElements` does? (with a numeric rather than a string comparison)

spec.html
jmdyck: Hm? The status quo has 20 occurrences of "a TypedArray", none of "a Typed Array".

ljharb: hm, ok, i didn't check but it looked wrong

spec.html
ljharb: actually i see how the different NaN and -0/+0 behavior would make that no clearer, so i'll resolve this.

spec.html
```diff
@@ -39809,6 +40015,36 @@ <h1>
...
+          CompareTypedArrayElements (
+            _x_: a Number or a BigInt,
+            _y_: a Number or a BigInt,
+            _comparefn_: a function object or *undefined*,
+          ): either a normal completion containing a Number or an abrupt completion
+        </h1>
+        <dl class="header">
+        </dl>
+        <emu-alg>
+          1. Assert: _x_ is a Number and _y_ is a Number, or _x_ is a BigInt and _y_ is a BigInt.
```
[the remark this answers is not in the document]
michaelficarra: I'd like to reduce the use of the `Type` macro if we can. The current text isn't all that bad.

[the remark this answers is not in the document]
ljharb: It’s workable, but it reads clearer to me to have some way of saying “the types are the same” without enumerating the possibilities.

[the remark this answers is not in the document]
michaelficarra: There's at least one other place where we do that. Maybe we should come up with a phrase to define/use.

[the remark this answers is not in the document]
ljharb: `x and y are the same type` as a macro for `Type(x) is Type(y)`?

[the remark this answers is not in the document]
michaelficarra: Yeah something like that. Either way, I'd prefer the current text for now until we introduce something like that.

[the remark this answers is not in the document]
ljharb: Why? We already have existing uses of the Type macro for this. Shouldn't we keep the status quo until a holistic change is made?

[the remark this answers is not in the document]
michaelficarra: Other uses like this are harder to avoid.

spec.html
```diff
@@ -38227,7 +38224,8 @@ <h1>
               _obj_: an Object,
               _len_: a non-negative integer,
               _SortCompare_: an Abstract Closure with two parameters,
-            ): either a normal completion containing an Object or an abrupt completion
+              _skipHoles_: a Boolean,
```
syg, line 38227: Just like in code, I personally prefer named enums instead of booleans so callsites are easier to read. Something like `_skipHoles_: ~skip-holes~ or ~do-not-skip-holes~`? The spec doesn't really do this in general but I'd like to start. @michaelficarra @bakkot WDYT?

spec.html
```diff
@@ -38397,6 +38418,86 @@ <h1>Array.prototype.toLocaleString ( [ _reserved1_ [ , _reserved2_ ] ] )</h1>
...
+        <p>This method performs the following steps when called:</p>
+        <emu-alg>
+          1. Let _O_ be ? ToObject(*this* value).
+          1. Let _len_ be ? LengthOfArrayLike(_O_).
+          1. Let _relativeStart_ be ? ToIntegerOrInfinity(_start_).
+          1. If _relativeStart_ is -∞, let _actualStart_ be 0.
+          1. Else if _relativeStart_ &lt; 0, let _actualStart_ be max(_len_ + _relativeStart_, 0).
+          1. Else, let _actualStart_ be min(_relativeStart_, _len_).
+          1. Let _insertCount_ be the number of elements in _items_.
+          1. If _start_ is not present, then
```
syg, line 38469: Man I had to remind myself that this (and `splice`) is *actually* "is not present", i.e. checking argument length, instead of checking for **undefined**. I'm sure I did the same thing while implementing and forgot promptly. Wack semantics bro.

spec.html
```diff
@@ -39767,6 +39952,27 @@ <h1>
...
+            _exemplar_: a TypedArray,
+            _argumentList_: a List of ECMAScript language values,
+          ): either a normal completion containing a TypedArray or a throw completion
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It is used to specify the creation of a new TypedArray using a constructor function that is derived from _exemplar_. Unlike TypedArraySpeciesCreate, which can construct custom TypedArray subclasses through the use of @@species, this operation always uses one of the built-in TypedArray constructors.</dd>
+        </dl>
+        <emu-alg>
+          1. Assert: _exemplar_ is an Object that has [[TypedArrayName]] and [[ContentType]] internal slots.
```
syg, line 39967: This assert isn't needed because _exemplar_ is given the type "a TypedArray" already in the header.

spec.html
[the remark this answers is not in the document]
syg: I think the current version of enumerating both types is pretty good. At first glance `Type(_x_) is Type(_y_)` may suggest some generality beyond the two types that's not there.

syg (approved):
lgtm

spec.html
michaelficarra: I would also like to start, and now is as good a time as any.

spec.html
ljharb: https://github.com/tc39/ecma262/pull/2997#discussion_r1082149982

spec.html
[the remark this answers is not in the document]
ljharb: Fair enough; i think "are the same type" would still be an improvement over both, just not as part of this proposal obv.

spec.html
ljharb: maybe `_holes_: ~skip~ or ~treat-as-undefined~`?

acutmore: Hard agree on switching to enum over boolean, thanks for suggesting.

`~treat-as-undefined~` isn't quite right for the `false` case, because we unconditionally do a `Get`, so we can get values other than `undefined` from the prototype.
It also felt odd to write `~do-not-skip-holes~` in the steps for `TypedArray` methods, as by definition have no holes.
So I've gone with `_holes_: ~skip-holes~ or ~do-not-detect-holes~`. Thoughts?

spec.html
acutmore: The good news is @nicolo-ribaudo covered that subtle difference in a test https://github.com/tc39/test262/blob/main/test/built-ins/Array/prototype/toSpliced/start-and-deleteCount-undefineds.js

spec.html
ljharb: maybe `~skip-holes~` and `~get-holes~`?

syg: I like `~skip-holes~` and `~do-not-detect-holes~` out of all the options so far. `~get-holes~` makes it sound like a hole is some special JS value that you can "get", which is misleading.

acutmore:
This proposal is now Stage 4: https://github.com/tc39/proposals/commit/ad4df8435f27f39eda26db3b940ae151980c8015
Commits squashed and rebased.

bakkot (approved):
LGTM other than typo.

spec.html
```diff
@@ -38195,21 +38195,18 @@ <h1>Array.prototype.sort ( _comparefn_ )</h1>
...
-            1. Return *+0*<sub>𝔽</sub>.
-          1. Return ? SortIndexedProperties(_obj_, _len_, _SortCompare_).
+            1. Return ? CompareArrayElements(_x_, _y_, _comparefn_).
+          1. [id="step-array-sortindexedproperties"] Let _sortedList_ be ? SortIndexedProperties(_obj_, _len_, _SortCompare_, ~skip-holes~).
+          1. Let _itemCount_ be the number of elements in _sortedList_.
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _itemCount_,
+            1. Perform ? Set(_obj_, ! ToString(𝔽(_j_)), _sortedList_[_j_], *true*).
+            1. Set _j_ to _j_ + 1.
+          1. NOTE: The call to SortIndexedProperties in step <emu-xref href="#step-array-sortindexedproperties"></emu-xref> has the *_holes_* parameter set to ~skip-holes~. The remaining indices are deleted to preserve the number of holes that were detected and excluded from the sort.
```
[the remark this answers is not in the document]
acutmore: > This diff has recently been updated. Refresh and try again.

My kingdom for this to work.

spec.html
michaelficarra: I'm late to the party, but how about `~skip-holes~` and `~read-through-holes~`? Reading the diff today, I initially didn't know what `~do-not-detect-holes~` meant and had to infer it from the algorithm steps.

syg: I'm also happy with @michaelficarra's suggestion.

acutmore: Switched to `~do-not-detect-holes~`

spec.html
```diff
@@ -38390,6 +38412,86 @@ <h1>Array.prototype.toLocaleString ( [ _reserved1_ [ , _reserved2_ ] ] )</h1>
...
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _len_,
+            1. Perform ! CreateDataPropertyOrThrow(_A_, ! ToString(𝔽(_j_)), _sortedList_[_j_]).
+            1. Set _j_ to _j_ + 1.
+          1. Return _A_.
+        </emu-alg>
+      </emu-clause>
+
+      <emu-clause id="sec-array.prototype.tospliced">
+        <h1>Array.prototype.toSpliced ( _start_, _deleteCount_, ..._items_ )</h1>
```
michaelficarra, line 38453: Nit, but the use of `delete` in these aliases implies mutation. Maybe it would be better to use `skip` instead.

acutmore: have updated to use 'skip'

spec.html
```diff
@@ -39728,6 +39914,26 @@ <h1>
...
+            _exemplar_: a TypedArray,
+            _argumentList_: a List of ECMAScript language values,
+          ): either a normal completion containing a TypedArray or a throw completion
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It is used to specify the creation of a new TypedArray using a constructor function that is derived from _exemplar_. Unlike TypedArraySpeciesCreate, which can construct custom TypedArray subclasses through the use of @@species, this operation always uses one of the built-in TypedArray constructors.</dd>
+        </dl>
+        <emu-alg>
+          1. Let _constructor_ be the intrinsic object listed in column one of <emu-xref href="#table-the-typedarray-constructors"></emu-xref> for _exemplar_.[[TypedArrayName]].
```
michaelficarra, line 39929: We don't typically number the column names like this. Although verbose, we would typically say

```suggestion
          1. Let _constructor_ be the intrinsic object associated with the constructor name _exemplar_.[[TypedArrayName]] in <emu-xref href="#table-the-typedarray-constructors"></emu-xref>.
```

Separately, we should probably split that column into two at some point. I have no idea why they are combined. But even if we do, that wording should be fine.

michaelficarra (approved):
LGTM. There are some opportunities I noticed to use our new interval notation, but I'm not sure what our editorial stance is on that at the moment, so this is fine as-is.

spec.html
```diff
@@ -39663,6 +39824,31 @@ <h1>%TypedArray%.prototype.values ( )</h1>
...
+        <p>This method performs the following steps when called:</p>
+        <emu-alg>
+          1. Let _O_ be the *this* value.
+          1. Perform ? ValidateTypedArray(_O_).
+          1. Let _len_ be _O_.[[ArrayLength]].
+          1. Let _relativeIndex_ be ? ToIntegerOrInfinity(_index_).
+          1. If _relativeIndex_ ≥ 0, let _actualIndex_ be _relativeIndex_.
+          1. Else, let _actualIndex_ be _len_ + _relativeIndex_.
+          1. If _O_.[[ContentType]] is ~BigInt~, set _value_ to ? ToBigInt(_value_).
+          1. Else, set _value_ to ? ToNumber(_value_).
```
jmdyck, line 39838: These steps overwrite the `_value_` parameter, which is somewhat discouraged. 
```suggestion
          1. If _O_.[[ContentType]] is ~BigInt~, let _numericValue_ be ? ToBigInt(_value_).
          1. Else, let _numericValue_ be ? ToNumber(_value_).
```
and then tweak the subsequent reference to `_value_`.

acutmore: Done. Much nicer.

spec.html
```diff
@@ -38188,21 +38188,18 @@ <h1>Array.prototype.sort ( _comparefn_ )</h1>
...
+            1. Perform ? Set(_obj_, ! ToString(𝔽(_j_)), _sortedList_[_j_], *true*).
+            1. Set _j_ to _j_ + 1.
+          1. NOTE: The call to SortIndexedProperties in step <emu-xref href="#step-array-sortindexedproperties"></emu-xref> uses ~skip-holes~. The remaining indices are deleted to preserve the number of holes that were detected and excluded from the sort.
+          1. Repeat, while _j_ &lt; _len_,
+            1. Perform ? DeletePropertyOrThrow(_obj_, ! ToString(𝔽(_j_))).
+            1. Set _j_ to _j_ + 1.
+          1. Return _obj_.
         </emu-alg>
         <emu-note>
           <p>Because non-existent property values always compare greater than *undefined* property values, and *undefined* always compares greater than any other value, *undefined* property values always sort to the end of the result, followed by non-existent property values.</p>
```
jmdyck, line 38205: In the status quo, this is talking about the consequences of steps that appear immediately above, but this PR moves those steps elsewhere. So maybe:
```suggestion
          <p>Because non-existent property values always compare greater than *undefined* property values, and *undefined* always compares greater than any other value (see CompareArrayElements), *undefined* property values always sort to the end of the result, followed by non-existent property values.</p>
```

spec.html
```diff
@@ -39591,20 +39719,13 @@ <h1>%TypedArray%.prototype.sort ( _comparefn_ )</h1>
...
+            1. Return ? CompareTypedArrayElements(_x_, _y_, _comparefn_).
+          1. Let _sortedList_ be ? SortIndexedProperties(_obj_, _len_, _SortCompare_, ~read-through-holes~).
+          1. Let _j_ be 0.
+          1. Repeat, while _j_ &lt; _len_,
+            1. Perform ! Set(_obj_, ! ToString(𝔽(_j_)), _sortedList_[_j_], *true*).
+            1. Set _j_ to _j_ + 1.
+          1. Return _obj_.
         </emu-alg>
         <emu-note>
           <p>Because *NaN* always compares greater than any other value, *NaN* property values always sort to the end of the result when _comparefn_ is not provided.</p>
```
jmdyck, line 39731: Similarly, this PR moves the steps that make `*NaN*` always compare greater than any other value. So maybe:
```suggestion
          <p>Because *NaN* always compares greater than any other value (see CompareTypedArrayElements), *NaN* property values always sort to the end of the result when _comparefn_ is not provided.</p>
```

Issue #2944, opened 2022-10-25
Table interactions should be driven by operations rather than prose


gibson042:
ECMA-262 uses a wide variety of patterns when referencing tabular data—implicitly, by column ordinal position, by [quoted or unquoted] column name, etc. Some examples:
* [ToBigInt](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-tobigint): "Return the value that _prim_ corresponds to in Table 14."
* [CreateIntrinsics](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-createintrinsics): "Set fields of _realmRec_.[[Intrinsics]] with the values listed in Table 6. The field names are the names listed in column one of the table."
* [OrdinaryFunctionCreate](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-ordinaryfunctioncreate) et al.: "Let _internalSlotsList_ be [_the list-concatenation of « … » and_] the internal slots listed in Table \<number>."
* [Evaluation of `typeof`](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-typeof-operator-runtime-semantics-evaluation): "Return a String according to Table 41."
* [ApplyStringOrNumericBinaryOperator](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-applystringornumericbinaryoperator): "Let _operation_ be the abstract operation associated with _opText_ and Type(_lnum_) in the following table"
* [UnicodeMatchProperty](https://tc39.es/ecma262/multipage/text-processing.html#sec-runtime-semantics-unicodematchproperty-p): "Assert: _p_ is a Unicode property name or property alias listed in the “Property name and aliases” column of Table 68 or Table 69… Let _c_ be the canonical property name of _p_ as given in the “Canonical property name” column of the corresponding row."
* [TypedArray](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-typedarray): "Let _constructorName_ be the String value of the Constructor Name value specified in Table 71 for this TypedArray constructor."
* [NumericToRawBytes](https://tc39.es/ecma262/multipage/structured-data.html#sec-numerictorawbytes): "Let _n_ be the Element Size value specified in Table 71 for Element Type _type_.… Let _convOp_ be the abstract operation named in the Conversion Operation column in Table 71 for Element Type _type_."

[...]

PR #3002, opened 2023-01-25
Editorial: Replace the operator tables with explicit steps

gibson042:
The tables in [Assignment Operators Evaluation](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-assignment-operators-runtime-semantics-evaluation) and [ApplyStringOrNumericBinaryOperator](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-applystringornumericbinaryoperator) are somewhat redundant with each other, and the latter in particular is more complex than https://github.com/tc39/ecma262/issues/2944#issuecomment-1292722630 suggests to me is desirable (especially considering the resulting `_operation_(_lnum_, _rnum_)` indirection), and incidentally are also the only examples of block-level elements inside an \<emu-alg> step. I think they can be clarified and simplified by replacement with explicit algorithm steps.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -20236,25 +20236,9 @@
         1. Let _rref_ be ? Evaluation of |AssignmentExpression|.
         1. Let _rval_ be ? GetValue(_rref_).
         1. Let _assignmentOpText_ be the source text matched by |AssignmentOperator|.
-        1. Let _opText_ be the sequence of Unicode code points associated with _assignmentOpText_ in the following table:
-          <figure>
-            <!-- emu-format ignore -->
-            <table class="lightweight-table">
-              <tr><th> _assignmentOpText_ </th><th> _opText_       </th></tr>
-              <tr><td> `**=`              </td><td> `**`           </td></tr>
-              <tr><td> `*=`               </td><td> `*`            </td></tr>
-              <tr><td> `/=`               </td><td> `/`            </td></tr>
-              <tr><td> `%=`               </td><td> `%`            </td></tr>
-              <tr><td> `+=`               </td><td> `+`            </td></tr>
-              <tr><td> `-=`               </td><td> `-`            </td></tr>
-              <tr><td> `&lt;&lt;=`        </td><td> `&lt;&lt;`     </td></tr>
-              <tr><td> `&gt;&gt;=`        </td><td> `&gt;&gt;`     </td></tr>
-              <tr><td> `&gt;&gt;&gt;=`    </td><td> `&gt;&gt;&gt;` </td></tr>
-              <tr><td> `&amp;=`           </td><td> `&amp;`        </td></tr>
-              <tr><td> `^=`               </td><td> `^`            </td></tr>
-              <tr><td> `|=`               </td><td> `|`            </td></tr>
-            </table>
-          </figure>
+        1. Let _len_ be the number of code points in _assignmentOpText_.
+        1. Assert: _len_ ≥ 2 and the last code point in _assignmentOpText_ is `=`.
+        1. Let _opText_ be the sequence of the first _len_ - 1 code points of _assignmentOpText_.
         1. Let _r_ be ? ApplyStringOrNumericBinaryOperator(_lval_, _opText_, _rval_).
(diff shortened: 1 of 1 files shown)


ljharb (approved):
i like this

spec.html
```diff
@@ -20329,39 +20313,34 @@ <h1>
         1. Let _lnum_ be ? ToNumeric(_lval_).
         1. Let _rnum_ be ? ToNumeric(_rval_).
         1. If Type(_lnum_) is different from Type(_rnum_), throw a *TypeError* exception.
+        1. NOTE: The following steps can only throw an exception if _lnum_ is a BigInt and _opText_ is `**`, `/`, `%`, or `>>>`.
```
michaelficarra, line 20316:
```suggestion
        1. NOTE: The following steps can only throw an exception if _lnum_ is a BigInt and _opText_ is one of `**`, `/`, `%`, or `>>>`.
```

michaelficarra (approved):
LGTM

jmdyck:
Note that this reverses some of the directions of PR #1961:
- [using lookup tables rather than 'if' steps](https://github.com/tc39/ecma262/pull/1961#discussion_r412300323)
- [not doing 'string manipulation' on the text of |AssignmentOperator|](https://github.com/tc39/ecma262/pull/1961#discussion_r411914953)

See also PR #2945, which concluded with "the current editors all prefer the table" (in the semantics for |AssignmentOperator|).

Not objecting, just making historical connections.

michaelficarra:
Yeah I'm neutral on the table for compound assignment. But the change to the Number/BigInt operator dispatch is very nice!

spec.html
gibson042: Done.

gibson042:
> Note that this reverses some of the directions of PR #1961:
> 
> * [using lookup tables rather than 'if' steps](https://github.com/tc39/ecma262/pull/1961#discussion_r412300323)
> * [not doing 'string manipulation' on the text of |AssignmentOperator|](https://github.com/tc39/ecma262/pull/1961#discussion_r411914953)
> 
> See also PR #2945, which concluded with "the current editors all prefer the table" (in the semantics for |AssignmentOperator|).
> 
> Not objecting, just making historical connections.

@jmdyck Much appreciated. This is indeed not even my first crack at this, let alone the first historical attempt. I did link to #2944 (whose rejection prompted the same fate for #2945) in the description, and both of them reference #1961. But even so, I feel like this form of the operations has legs.

> Yeah I'm neutral on the table for compound assignment. But the change to the Number/BigInt operator dispatch is very nice!

@michaelficarra Thank you. The compound assignment table does have a simpler single-lookup-key structure, and is therefore less objectionable—my biggest issue is not its existence (I actually like tables, cf. #2944) but rather its placement inside an \<emu-alg> step (cf. https://github.com/tc39/ecmarkup/pull/513 ). If consensus is still to prefer it over text manipulation, then I would propose promoting it into a full \<emu-table> at e.g. [Assignment Operators](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-assignment-operators) or a new operation. Or yet another possibility would be introducing a new syntax-directed operation for |AssignmentOperator|. But the bottom line for me is that having it _inside_ an algorithm step is unnecessarily awkward.

bakkot:
@gibson042 we talked about this at the editor call today and couldn't come to an agreement. Personally I think the inline tables are the clearest option of those discussed, so I prefer the status quo. I am slightly negative on the algorithm steps which do string manipulation (since I think they require more thought to follow, plus the assert), and strongly negative on the algorithm steps which determine which AO to call (since I think they obscure what's going on, particularly since four of the 24 are different from the rest). But @michaelficarra was very positive about the second change.

We'll talk about it a bit more next call, but unless we come to agreement we'll probably end up sticking with the status quo.

bakkot:
We talked about this again and are happier with the status quo. We did discuss the option of factoring out the tables into their own steps, or introducing SDOs to do the parsing, but those both seem less readable. Thanks for the suggestion, though.

PR #3006, opened 2023-01-29
Meta: Limit IPR check job to the tc39 repository

gibson042:
This aligns it with the other privileged jobs.

### DIFF
--- a/.github/workflows/ipr.yml
+++ b/.github/workflows/ipr.yml
@@ -12,6 +12,7 @@ jobs:
   build:
     name: 'check IPR form'
     runs-on: ubuntu-latest
+    if: ${{ secrets.GH_IPR_TOKEN != '' && secrets.GOOGLE_API_KEY != '' }}
 
     steps:
       - uses: actions/checkout@v2


bakkot (approved):

ljharb:
I had it this way so I could test it on my fork; maybe instead it could check not the repo name, but the presence of the secrets?

gibson042:
I'm not sure I understand; are you saying that this job _should_ be subject to different conditions than the other privileged [ecma262 workflows](https://github.com/tc39/ecma262/tree/main/.github/workflows) such as [deploy](https://github.com/tc39/ecma262/blob/main/.github/workflows/deploy.yml) and [upload preview](https://github.com/tc39/ecma262/blob/main/.github/workflows/preview.yml)?

ljharb:
Yes - in general, I think an action should be checking for secret presence instead of hardcoding the repo name, but "upload preview" can't ever work for another repo (on the begin side), so that check is correct.

gibson042:
I think that's answering a different question than I asked. I agree that hardcoding the repo name isn't great, but I see more value in consistency across checks than in updating just one workflow to use a better approach. Regardless, I don't feel very strongly about this so if you make a concrete ````` ```suggestion ````` I'll accept it (or you can just apply a change directly).

ljharb:
I don't think consistency has much value here; each job is doing its own bespoke thing. I'll add a suggestion, though.

.github/workflows/ipr.yml
```diff
@@ -12,6 +12,7 @@ jobs:
   build:
     name: 'check IPR form'
     runs-on: ubuntu-latest
+    if: ${{ github.repository == 'tc39/ecma262' }}
```
ljharb, line 15:
```suggestion
    if: ${{ secrets.GH_IPR_TOKEN != '' && secrets.GOOGLE_API_KEY != '' }}
```
i think this might do it?

gibson042: Doesn't this increase the risk of silent failure if e.g. a secret is accidentally removed or renamed?

ljharb: yes, but that's not a meaningful risk because they're required branch protections - we'll discover it pretty quickly.

gibson042: If you say so. Accepted.

ljharb (approved):

michaelficarra (approved):

ljharb:
I ended up having to revert this; apparently secrets aren't available inside `if` conditionals. We'd have to figure out something more complex to make this job conditional on the secrets, and I prefer not to hardcode it to the specific repo.

If you'd like to try an alternative approach please open up a new PR.

gibson042:
@ljharb What about the original approach in 919b36d92f925bbf9d76972da617ea2730816cc1 that is aligned with the other privileged [ecma262 workflows](https://github.com/tc39/ecma262/tree/main/.github/workflows) such as [deploy](https://github.com/tc39/ecma262/blob/main/.github/workflows/deploy.yml) and [upload preview](https://github.com/tc39/ecma262/blob/main/.github/workflows/preview.yml)?

ljharb:
That approach would work, but it would limit testing - the reason those specific deploys have that check is because the begin.com service they talk to also has that check.

gibson042:
Right now this check is generating spurious failures in my own fork repository, and presumably in other forks as well. Where do you think it needs to happen other than tc39/ecma262, and why does the same not apply to other privileged workflows?

ljharb:
I test changes to the workflow in ljharb/ecma262 so that I know they pass before landing them here. I failed to do that very testing on this PR, unfortunately.

The other privileged workflows are *unique* in that they contact a third party service that explicitly rejects, serverside, anything that's not tc39/ecma262.

I agree that spurious failures in a fork are undesirable, I just don't want to prevent testability of the workflow in the process if it can be avoided.

PR #3014, opened 2023-02-15
Editorial: add emu-not-ref for "date-time values"

justingrant:
Found this while investigating a different issue here: https://github.com/tc39/ecmarkup/issues/518#issuecomment-1430542065


bakkot (approved):
Thanks!

PR #3019, opened 2023-02-20
Editorial: Replace biblio term "safe" with the more specific "safe integer"

gibson042:
This preserves the existing "safe-integer" id, but fixes autolinking issues necessitating things like https://github.com/tc39/proposal-temporal/pull/2502 .

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -31015,7 +31015,7 @@
       <emu-clause id="sec-number.issafeinteger">
         <h1>Number.isSafeInteger ( _number_ )</h1>
         <emu-note>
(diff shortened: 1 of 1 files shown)


bakkot:
This should update the existing auto-references in `Number.MAX_SAFE_INTEGER` and `Number.MIN_SAFE_INTEGER ` to explicitly link, I think.

gibson042:
Those references already use `<emu-xref href="#safe-integer">safe</emu-xref>`, which are not affected by this change.

bakkot (approved):

PR #3021, opened 2023-03-01
Editorial: Clarify GetIterator

syg:
I find GetIterator's optional parameters confusing because the hint is ignored when the method is passed. There's also a recursive call that serves no real purpose.

This PR gets rid of the optional parameters and does some refactoring.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -6797,31 +6797,43 @@
       </emu-table>
     </emu-clause>
 
+    <emu-clause id="sec-getiteratorfrommethod" type="abstract operation">
+      <h1>
+        GetIteratorFromMethod (
+          _obj_: an ECMAScript language value,
+          _method_: a function object,
+        ): either a normal completion containing an Iterator Record or a throw completion
+      </h1>
+      <dl class="header">
+      </dl>
+      <emu-alg>
+        1. Let _iterator_ be ? Call(_method_, _obj_).
+        1. If _iterator_ is not an Object, throw a *TypeError* exception.
+        1. Let _nextMethod_ be ? GetV(_iterator_, *"next"*).
+        1. Let _iteratorRecord_ be the Iterator Record { [[Iterator]]: _iterator_, [[NextMethod]]: _nextMethod_, [[Done]]: *false* }.
+        1. Return _iteratorRecord_.
+      </emu-alg>
+    </emu-clause>
+
     <emu-clause id="sec-getiterator" type="abstract operation">
       <h1>
         GetIterator (
           _obj_: an ECMAScript language value,
           optional _hint_: ~sync~ or ~async~,
-          optional _method_: a function object,
         ): either a normal completion containing an Iterator Record or a throw completion
       </h1>
       <dl class="header">
       </dl>
       <emu-alg>
         1. If _hint_ is not present, set _hint_ to ~sync~.
-        1. If _method_ is not present, then
-          1. If _hint_ is ~async~, then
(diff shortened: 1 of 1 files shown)


michaelficarra (approved):
This is an improvement and doesn't impact the iterator helpers proposal.

ljharb (approved):
I implemented all the AO changes in es-abstract, and it looks good to me. Additionally, iterator helpers' `GetIteratorDirect` AO is basically the last 4 steps of GetIteratorFromMethod, so it seems pretty compatible there.

spec.html
```diff
@@ -6996,20 +7007,15 @@ <h1>
...
     </emu-clause>
 
-    <emu-clause id="sec-iterabletolist" type="abstract operation">
+    <emu-clause id="sec-iteratortolist" oldids="sec-iterabletolist" type="abstract operation">
       <h1>
-        IterableToList (
-          _items_: an ECMAScript language value,
-          optional _method_: a function object,
+        IteratorToList (
+          _iteratorRecord_: an Iterator Record
```
jmdyck, line 7013:
```suggestion
          _iteratorRecord_: an Iterator Record,
```

spec.html
```diff
@@ -6824,6 +6814,27 @@ <h1>
...
+      <dl class="header">
+      </dl>
+      <emu-alg>
+        1. If _kind_ is ~async~, then
+          1. Let _method_ be ? GetMethod(_obj_, @@asyncIterator).
+          1. If _method_ is *undefined*, then
+            1. Let _syncMethod_ be ? GetMethod(_obj_, @@iterator).
+            1. Let _syncIteratorRecord_ be ? GetIteratorFromMethod(_obj_, _syncMethod_).
+            1. Return CreateAsyncFromSyncIterator(_syncIteratorRecord_).
+        1. Otherwise, let _method_ be ? GetMethod(_obj_, @@iterator).
```
jmdyck, line 6833: I'm pretty sure that, for both occurrences of `GetMethod(_obj_, @@iterator)`, the result might be `*undefined*`, which isn't being handled, it's just being passed to the second parameter of `GetIteratorFromMethod`, which is expecting a function object.

(This is a pre-existing problem, but you might want to fix it in this PR.)

jmdyck (changes requested):

spec.html
```diff
@@ -6996,20 +7007,15 @@ <h1>
...
 
-    <emu-clause id="sec-iterabletolist" type="abstract operation">
+    <emu-clause id="sec-iteratortolist" oldids="sec-iterabletolist" type="abstract operation">
       <h1>
-        IterableToList (
-          _items_: an ECMAScript language value,
-          optional _method_: a function object,
+        IteratorToList (
+          _iteratorRecord_: an Iterator Record
         ): either a normal completion containing a List of ECMAScript language values or a throw completion
```
gibson042, line 7014: @ptomato Assuming this lands, [Temporal IterableToListOfType](https://tc39.es/proposal-temporal/#sec-iterabletolistoftype) should probably change accordingly.

gibson042 (approved):
I think you missed the downstream consumers check, but I've opened https://github.com/tc39/ecma402/pull/756 to be ready.

spec.html
syg: Done.

spec.html
syg: Thank you.

bakkot (approved):
Much nicer.

PR #3023, opened 2023-03-02
Normative: Avoid initializing fields twice for instances of base classes without explicit constructors

### LINKED ISSUE #3020: InitializeInstanceElements is invoked twice for base classes with no explicit constructor
**Description:** 
According to section `15.7.14 RS: ClassDefinitionEvaluation`, step `14.a.vi`, when a class is defined without an explicit `constructor` method, the abstract closure that is used as the constructor function for this class invokes `InitializeInstanceElements`, regardless of whether it is a base class or a derived class. This is necessary for derived classes, because, as there is no explicit `constructor` method, there is no explicit `super` call, which would call `InitializeInstanceElements` for the derived class. However, for base classes, according to `10.2.2`, step `6.b`, `[[Construct]]` will call `InitializeInstanceElements` for base classes. This means that, per the spec, an instance of a base class without an explicit constructor should have its fields initialized twice, once due to the `new` operator calling `[[Construct]]` and then due to the call to the implicit constructor.

However, this is not the case in major implementations of the language. For the following example:
```javascript
let obj = {
    get prop() {
        print('Initializer');
        return 1;
    }
}

class A {
    field = obj.prop
}

let a = new A();
```` 

 **eshost output**:
 ```bash
 #### engine262, GraalJS, JavaScriptCore, Moddable XS, QuickJS, SpiderMonkey, V8
Initializer
````

I believe this is either just a specification error as it would be pointless initializing the fields twice, as all major implementations have figured, but there's also a chance that I misunderstood something in the specification. In the case of the former, the solution would simply be to move step `14.a.vi` in `ClassDefinitionEvaluation` to `14.a.iv.5`

StefanOctavian:
Closes #3020 
Since both [[Construct]] and the constructor function that the class is (as described in Runtime Semantics: ClassDefinitionEvaluation) initialize the instance fields for base classes with implicit constructor method, and since [[Construct]] does not initialize them for instances of derived classes, but instead super calls do, the step that initializes them in the implicit constructor should only be present for derived classes. This PR fixes this, as discussed in the issue refferenced.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -24657,10 +24657,10 @@
               1. Let _func_ be ! _F_.[[GetPrototypeOf]]().
               1. If IsConstructor(_func_) is *false*, throw a *TypeError* exception.
               1. Let _result_ be ? Construct(_func_, _args_, NewTarget).
+              1. Perform ? InitializeInstanceElements(_result_, _F_).
             1. Else,
               1. NOTE: This branch behaves similarly to `constructor() {}`.
               1. Let _result_ be ? OrdinaryCreateFromConstructor(NewTarget, *"%Object.prototype%"*).
-            1. Perform ? InitializeInstanceElements(_result_, _F_).
             1. Return _result_.
           1. Let _F_ be CreateBuiltinFunction(_defaultConstructor_, 0, _className_, « [[ConstructorKind]], [[SourceText]] », the current Realm Record, _constructorParent_).
         1. Else,


michaelficarra (approved):

StefanOctavian:
Checks appeared to be frozen by the Github Actions outage yesterday, so I added an empty commit to retrigger them

ljharb:
@StefanOctavian that's not the case; it's just that first-time contributors need to have their CI runs manually approved.

StefanOctavian:
Oh, sorry, didn't know that

syg (approved):

StefanOctavian:
@bakkot  OOH, you're right. I'm sorry for not realising that, but I'm also happy you made me understand the fault in my logic. Thank you so much and with that I will close this PR and issue.

PR #3027, opened 2023-03-09
Editorial: Clarify identity as from perspective of spec vs language

syg:
This 2-level stratification of identity is needed rigor for Symbols as WeakMap keys (#2777) to capture the reason for why Symbol.for symbols cannot be used as weak targets.


spec.html
```diff
@@ -1085,10 +1085,10 @@ <h1>Value Notation</h1>
...
+      <p>From the perspective of this specification, the word “is” is used to compare two values for equality, as in “If _bool_ is *true*, then ...”, and the word “contains” is used to search for a value inside lists using equality comparisons, as in "If _list_ contains a Record _r_ such that _r_.[[Foo]] is *true*, then ...". From the perspective of the ECMAScript language, language values are compared for quality using the SameValue abstract operation and the abstract operations it transitively calls.</p>
```
bakkot, line 1089:
```suggestion
      <p>From the perspective of this specification, the word “is” is used to compare two values for equality, as in “If _bool_ is *true*, then ...”, and the word “contains” is used to search for a value inside lists using equality comparisons, as in "If _list_ contains a Record _r_ such that _r_.[[Foo]] is *true*, then ...". From the perspective of the ECMAScript language, language values are compared for equality using the SameValue abstract operation and the abstract operations it transitively calls.</p>
```

spec.html
```diff
@@ -1085,10 +1085,10 @@ <h1>Value Notation</h1>
...
+      <p>ECMAScript language values without identity are <emu-xref href="#sec-ecmascript-language-types-undefined-type">*undefined*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-null-type">*null*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-boolean-type">Booleans</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-string-type">Strings</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-number-type">Numbers</emu-xref>, and <emu-xref href="#sec-ecmascript-language-types-bigint-type">BigInts</emu-xref>. ECMAScript language values with identity are <emu-xref href="#sec-ecmascript-language-types-symbol-type">Symbols</emu-xref> and <emu-xref href="#sec-object-type">Objects</emu-xref>. Note that Symbols produced by <emu-xref href="#sec-symbol.for">Symbol.for</emu-xref> are a special case; they have identity from the perspective of the specification but do not have identity from the perspective of the ECMAScript language. These lists are exhaustive.</p>
```
bakkot, line 1091:
```suggestion
      <p>The ECMAScript language values without identity are <emu-xref href="#sec-ecmascript-language-types-undefined-type">*undefined*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-null-type">*null*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-boolean-type">Booleans</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-string-type">Strings</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-number-type">Numbers</emu-xref>, and <emu-xref href="#sec-ecmascript-language-types-bigint-type">BigInts</emu-xref>. The ECMAScript language values with identity are <emu-xref href="#sec-ecmascript-language-types-symbol-type">Symbols</emu-xref> and <emu-xref href="#sec-object-type">Objects</emu-xref>. Note that Symbols produced by <emu-xref href="#sec-symbol.for">Symbol.for</emu-xref> are a special case; they have identity from the perspective of the specification but do not have identity from the perspective of the ECMAScript language.</p>
```
maybe? Could also keep the last sentence.

bakkot (approved):

spec.html
syg: I like your version better, will apply.

spec.html
```diff
@@ -1085,10 +1085,10 @@ <h1>Value Notation</h1>
...
+      <p>The ECMAScript language values without identity are <emu-xref href="#sec-ecmascript-language-types-undefined-type">*undefined*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-null-type">*null*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-boolean-type">Booleans</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-string-type">Strings</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-number-type">Numbers</emu-xref>, and <emu-xref href="#sec-ecmascript-language-types-bigint-type">BigInts</emu-xref>. The ECMAScript language values with identity are <emu-xref href="#sec-ecmascript-language-types-symbol-type">Symbols</emu-xref> and <emu-xref href="#sec-object-type">Objects</emu-xref>. Note that Symbols produced by <emu-xref href="#sec-symbol.for">Symbol.for</emu-xref> are a special case; they have identity from the perspective of this specification but do not have identity from the perspective of the ECMAScript language.</p>
```
michaelficarra, line 1091:
```suggestion
      <p>The ECMAScript language values without identity are <emu-xref href="#sec-ecmascript-language-types-undefined-type">*undefined*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-null-type">*null*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-boolean-type">Booleans</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-string-type">Strings</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-number-type">Numbers</emu-xref>, and <emu-xref href="#sec-ecmascript-language-types-bigint-type">BigInts</emu-xref>. The ECMAScript language values with identity are <emu-xref href="#sec-ecmascript-language-types-symbol-type">Symbols</emu-xref> and <emu-xref href="#sec-object-type">Objects</emu-xref>. Note that Symbols produced by <emu-xref href="#sec-symbol.for">Symbol.for</emu-xref> are a special case: from the perspective of the ECMAScript language, they are compared for equality as if they did not have identity.</p>
```

spec.html
```diff
@@ -1085,10 +1085,10 @@ <h1>Value Notation</h1>
...
...uality, values fall into one of two categories. <dfn variants="values without identity,value without identity">Values without identity</dfn> are equal to other values without identity if all of their innate characteristics are the same — characteristics such as the magnitude of an integer or the length of a sequence. Values without identity may be manifest without prior reference by fully describing their characteristics. In contrast, each <dfn variants="values with identity">value with identity</dfn> is unique and therefore only equal to itself. Values with identity are like values without identity but with an additional unguessable, unchangeable, universally-unique characteristic called <em>identity</em>. References to existing values with identity cannot be manifest simply by describing them, as the identity itself is indescribable; instead, references to these values must be explicitly passed from one place to another. Some values with identity are mutable and therefore can have their characteristics (except their identity) changed in-place, causing all holders of the value to observe the new characteristics. A value without identity is never equal to a value with identity.</p>
```
michaelficarra, line 1088:
```suggestion
      <p>In this specification, both specification values and ECMAScript language values are compared for equality. When comparing for equality, values fall into one of two categories. <dfn variants="values without identity,value without identity">Values without identity</dfn> are equal to other values without identity if all of their innate characteristics are the same — characteristics such as the magnitude of an integer or the length of a sequence. Values without identity may be manifest without prior reference by fully describing their characteristics. In contrast, each <dfn variants="values with identity">value with identity</dfn> is unique and therefore only equal to itself. Values with identity are like values without identity but with an additional unguessable, unchangeable, universally-unique characteristic called <em>identity</em>. References to existing values with identity cannot be manifest simply by describing them, as the identity itself is indescribable; instead, references to these values must be explicitly passed from one place to another. Some values with identity are mutable and therefore can have their characteristics (except their identity) changed in-place, causing all holders of the value to observe the new characteristics. A value without identity is never equal to a value with identity.</p>
```

The explicit "to each other" feels like it's implying that we compare across these kinds of values, which we never do. Without it, I don't infer that.

spec.html
syg: We discussed this and I'm keeping it as is, but noted it's under protest.

spec.html
syg: Done.

spec.html
syg: Marking resolved as outdated, see new commit.

spec.html
```diff
@@ -1085,10 +1085,11 @@ <h1>Value Notation</h1>
...
+      <p>From the perspective of this specification, the word “is” is used to compare two values for equality, as in “If _bool_ is *true*, then ...”, and the word “contains” is used to search for a value inside lists using equality comparisons, as in "If _list_ contains a Record _r_ such that _r_.[[Foo]] is *true*, then ...". The <em>specification identity</em> of values determines the result of these comparisons and is axiomatic in this specification.</p>
+      <p>From the perspective of the ECMAScript language, language values are compared for equality using the SameValue abstract operation and the abstract operations it transitively calls. The algorithms of these comparison abstract operations determine <em>language identity</em> of ECMAScript language values.</p>
```
ljharb, line 1090: i'd strongly prefer this to be dfn-d, but obv landing this is more important than addressing that even if editors preferred to, so it can be addressed later if needed.

syg:
Please squash when merging, @ljharb, thanks.

spec.html
```diff
@@ -1085,10 +1085,11 @@ <h1>Value Notation</h1>
...
+      <p>For specification values, examples of values without specification identity include, but are not limited to: mathematical values and extended mathematical values; ECMAScript source text, surrogate pairs, Directive Prologues, etc; UTF-16 code units; Unicode code points; enums; abstract operations, including syntax-directed operations, host hooks, etc; and ordered pairs. Examples of specification values with specification identity include, but are not limited to: any kind of Records, including Property Descriptors, PrivateElements, etc; Parse Nodes; Lists; <emu-xref href="#sec-set-and-relation-specification-type">Sets</emu-xref> and Relations; Abstract Closures; Data Blocks; Private Names; execution contexts and execution context stacks; agent signifiers; and WaiterLists. These lists are not exhaustive.</p>
```
bakkot, line 1091:
```suggestion
      <p>For specification values, examples of values without specification identity include, but are not limited to: mathematical values and extended mathematical values; ECMAScript source text, surrogate pairs, Directive Prologues, etc; UTF-16 code units; Unicode code points; enums; abstract operations, including syntax-directed operations, host hooks, etc; and ordered pairs. Examples of specification values with specification identity include, but are not limited to: any kind of Records, including Property Descriptors, PrivateElements, etc; Parse Nodes; Lists; <emu-xref href="#sec-set-and-relation-specification-type">Sets</emu-xref> and Relations; Abstract Closures; Data Blocks; Private Names; execution contexts and execution context stacks; agent signifiers; and WaiterLists.</p>
```

Comment is redundant with "these lists are not exhaustive". I don't care which gets dropped, but probably one of them should be.

spec.html
```diff
@@ -1085,10 +1085,11 @@ <h1>Value Notation</h1>
...
+      <p>The ECMAScript language values without specification identity and without language identity are <emu-xref href="#sec-ecmascript-language-types-undefined-type">*undefined*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-null-type">*null*</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-boolean-type">Booleans</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-string-type">Strings</emu-xref>, <emu-xref href="#sec-ecmascript-language-types-number-type">Numbers</emu-xref>, and <emu-xref href="#sec-ecmascript-language-types-bigint-type">BigInts</emu-xref>. The ECMAScript language values with specification identity and language identity are the <emu-xref href="#sec-object-type">Objects</emu-xref>. <emu-xref href="#sec-ecmascript-language-types-symbol-type">Symbols</emu-xref> are a special case. All Symbol values have specification identity. Symbol values produced by <emu-xref href="#sec-symbol.for">Symbol.for</emu-xref> do not have language identity, while Symbol values not produced by <emu-xref href="#sec-symbol.for">Symbol.for</emu-xref> have language identity.</p>
```
bakkot, line 1092: I am OK with this, but I think is a little awkward to read. I think I'd do it a little differently:

> Specification identity agrees with language identity for all ECMAScript language values except Symbol values produced by `Symbol.for`. The ECMAScript language values without specification or language identity are (list). The ECMAScript language values with specification and language identity are (list). Symbol values produced by `Symbol.for` have language identity, but not specification identity.

I guess the main thing I want is a sentence up front saying that they agree in all cases except Symbol.for symbols; the other wording tweaks are less important.

spec.html
syg: Dropped "these lists are not exhaustive"

spec.html
michaelficarra: WFM

PR #3029, opened 2023-03-15
Editorial: Remove incorrect equivalence claim for RegExp constructor

### LINKED ISSUE #3028: Claimed equivalence between `RegExp(…)` and `new RegExp(…)` doesn’t hold
According to [§22.2.4 The RegExp Constructor](https://tc39.es/ecma262/multipage/text-processing.html#sec-regexp-constructor):

> Thus the function call `RegExp(…)` is equivalent to the object creation expression `new RegExp(…)` with the same arguments.

But empirically (in at least JavaScriptCore, SpiderMonkey, and V8), these calls are not equivalent:

```js
> r = /a/;
/a/
> RegExp(r) === r
true
> new RegExp(r) === r
false
```

and looking at the steps in §22.2.4.1, it should only be possible for the `RegExp` constructor to return its argument unchanged (2.b.ii) when NewTarget is **undefined**, i.e. it was called as `RegExp(…)` rather than `new RegExp(…)`.

Does the equivalence claim need to be corrected? It seems to have motivated an incorrect optimization in the SWC minifier:

- swc-project/swc#6941

[Previously posted [to the ES Discourse](https://es.discourse.group/t/is-regexp-really-equivalent-to-new-regexp/1637).]

andersk:
`RegExp(…)` is not equivalent to `new RegExp(…)`, because when the argument is an existing `RegExp` object, the former passes through the same object and the latter creates a new one (with e.g. `lastIndex` reset).

Fixes #3028.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -36389,7 +36389,7 @@ THH:mm:ss.sss
       <ul>
         <li>is <dfn>%RegExp%</dfn>.</li>
         <li>is the initial value of the *"RegExp"* property of the global object.</li>
-        <li>creates and initializes a new RegExp object when called as a function rather than as a constructor. Thus the function call `RegExp(…)` is equivalent to the object creation expression `new RegExp(…)` with the same arguments.</li>
+        <li>when called as a function, returns either a new RegExp object, or the argument itself if the only argument is a RegExp object.</li>
         <li>may be used as the value of an `extends` clause of a class definition. Subclass constructors that intend to inherit the specified RegExp behaviour must include a `super` call to the RegExp constructor to create and initialize subclass instances with the necessary internal slots.</li>
       </ul>
 


bakkot (approved):

spec.html
```diff
@@ -36368,7 +36368,7 @@ <h1>The RegExp Constructor</h1>
       <ul>
         <li>is <dfn>%RegExp%</dfn>.</li>
         <li>is the initial value of the *"RegExp"* property of the global object.</li>
-        <li>creates and initializes a new RegExp object when called as a function rather than as a constructor. Thus the function call `RegExp(…)` is equivalent to the object creation expression `new RegExp(…)` with the same arguments.</li>
+        <li>creates and initializes a new RegExp object when called as a function rather than as a constructor (unless the only argument is already a RegExp object).</li>
```
ljharb, line 36371:
```suggestion
        <li>when called as a function, returns either a new RegExp object, or the argument itself if the argument is a RegExp.</li>
```
?

syg: I like this suggestion.

andersk: Updated (but I kept the “only argument” wording to acknowledge the distinction between `RegExp(r) === r` and `RegExp(r, "") !== r`).

michaelficarra (approved):

syg (approved):

ljharb:
@andersk would you please register as a [contributor](https://tc39.es/agreements/contributor/)?

andersk:
Done.

PR #3030, opened 2023-03-16
Editorial: Fix `[[DFSIndex]]` in cyclic module record async evaluation example

nicolo-ribaudo:
As the textual description of the example suggests, _D_ is visited before _C_ and thus its `[[DFSIndex]]` is lower.

You can test the indexes for various graphs at https://nicolo-ribaudo.github.io/es-module-evaluation/#s=IEEKQiBDCiBEIEUK&c=QSAtPiBCCkEgLT4gQwpCIC0%2BIEQKQyAtPiBECkQgLT4gQQpDIC0%2BIEUK&a=RSBEIEMgQiBB&f=

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -26972,7 +26972,7 @@
               </tr>
               <tr>
                 <th>_C_</th>
-                <td>2</td>
+                <td>3</td>
                 <td>0</td>
                 <td>~evaluating-async~</td>
                 <td>*true*</td>
@@ -26981,7 +26981,7 @@
               </tr>
               <tr>
                 <th>_D_</th>
-                <td>3</td>
+                <td>2</td>
                 <td>0</td>
                 <td>~evaluating-async~</td>
                 <td>*true*</td>
@@ -27017,7 +27017,7 @@
               </thead>
               <tr>
                 <th>_C_</th>
-                <td>2</td>
+                <td>3</td>
                 <td>0</td>
                 <td>~evaluating-async~</td>
                 <td>*true*</td>
@@ -27062,7 +27062,7 @@
               </tr>
               <tr>
                 <th>_C_</th>
-                <td>2</td>
+                <td>3</td>
                 <td>0</td>
                 <td>~evaluating-async~</td>
                 <td>*true*</td>
@@ -27071,7 +27071,7 @@
               </tr>
               <tr>
                 <th>_D_</th>
-                <td>3</td>
+                <td>2</td>
                 <td>0</td>
                 <td>~evaluated~</td>
                 <td>*true*</td>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -26820,7 +26820,7 @@ <h1>Example Cyclic Module Record Graphs</h1>
               </tr>
               <tr>
                 <th>_D_</th>
-                <td>3</td>
+                <td>2</td>
```
guybedford, line 26823: We also need to correct the C entry just before this one.

guybedford (approved):

spec.html
nicolo-ribaudo: This PR does it :P

guybedford: Not on line 26814...

nicolo-ribaudo:
Updated, this should be ready now :)

spec.html
```diff
@@ -26957,8 +26957,8 @@ <h1>Example Cyclic Module Record Graphs</h1>
               </tr>
               <tr>
                 <th>_C_</th>
-                <td>2</td>
-                <td>1</td>
+                <td>3</td>
+                <td>0</td>
```
syg, line 26961: Could you explain the [[DFSAncestorIndex]] change? Can't follow the algorithm very well.

syg (commented):
I think I convinced myself that C and D should have their [[DFSIndex]] reversed since D is visited first in DFS.

I don't remember why [[DFSIndex]] is set twice: once in Link and once in Evaluate. How does that work?

guybedford:
> I don't remember why [[DFSIndex]] is set twice: once in Link and once in Evaluate. How does that work?

The graph is iterated twice. These fields are basically reused without contention, as from Allen's original implementation. I'm not sure if it is possible for the orderings to differ or if they could be reused, but that would depend on carefully ensuring invariants of the callers maintaining Link and Evaluate call orderings I think.

spec.html
guybedford: I believe this is just a typo since it is `0` in all previous cases. The mutable state changes are invariant of the global numbering of DFSIndicies, so it should at least be consistent with the others since it's the same graph as the other tables.

syg:
> The graph is iterated twice. These fields are basically reused without contention, as from Allen's original implementation. I'm not sure if it is possible for the orderings to differ or if they could be reused, but that would depend on carefully ensuring invariants of the callers maintaining Link and Evaluate call orderings I think.

But we also don't reset the indices. In today's world, if the orderings aren't the same, wouldn't we be getting bogus behavior?

guybedford:
> But we also don't reset the indices. In today's world, if the orderings aren't the same, wouldn't we be getting bogus behavior?

The invariants have never changed - the numberings are "locked" while a linking or evaluation operation assigns them synchronously. And during a top-level Link or Evaluation call they are all assigned uniquely without reading any existing values.

syg (approved):

bakkot (approved):

PR #3038, opened 2023-04-05
Editorial: Fix notes about explicit setting of "length" in certain Array methods

### LINKED ISSUE #3036: Explicit setting "length" in splice/slice is always necessary
[`Array.p.slice()`](https://tc39.es/ecma262/#sec-array.prototype.slice) and [`Array.p.splice()`](https://tc39.es/ecma262/#sec-array.prototype.splice) has a note that reads:

> Setting "length" became unnecessary starting in ES2015

This is still necessary, because the constructor used in `ArraySpeciesCreate` may not actually use the parameter to set its `length` property.

```js
const arr = [1, 2, 3];
arr.constructor = { [Symbol.species]: function () {} };

print(Array.prototype.slice.call(arr, 0, 3).length);
```

```
#### engine262, GraalJS, JavaScriptCore, SpiderMonkey, V8
3
```

If the step that sets the length is removed, the `Array.prototype.slice.call(arr, 0, 3)` step would not have the `length` property.

bakkot:
Fixes #3036. Closes #1277.

There's other places where such a note could be warranted (e.g. in `push`) but I'm not adding new notes in this PR.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -37435,7 +37435,7 @@ THH:mm:ss.sss
         </emu-alg>
         <p>The *"length"* property of this method is *1*<sub>𝔽</sub>.</p>
         <emu-note>
-          <p>The explicit setting of the *"length"* property in step <emu-xref href="#step-array-proto-concat-set-length"></emu-xref> is necessary to ensure that its value is correct in situations where the trailing elements of the result Array are not present.</p>
+          <p>The explicit setting of the *"length"* property in step <emu-xref href="#step-array-proto-concat-set-length"></emu-xref> is intended to ensure the length is correct when the final non-empty element of _items_ has trailing holes or when _A_ is not a built-in Array.</p>
         </emu-note>
         <emu-note>
           <p>This method is intentionally generic; it does not require that its *this* value be an Array. Therefore it can be transferred to other kinds of objects for use as a method.</p>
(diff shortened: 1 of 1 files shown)


spec.html
```diff
@@ -37435,8 +37435,9 @@ <h1>Array.prototype.concat ( ..._items_ )</h1>
         </emu-alg>
         <p>The *"length"* property of this method is *1*<sub>𝔽</sub>.</p>
         <emu-note>
-          <p>The explicit setting of the *"length"* property in step <emu-xref href="#step-array-proto-concat-set-length"></emu-xref> is necessary to ensure that its value is correct in situations where the trailing elements of the result Array are not present.</p>
+          <p>The explicit setting of the *"length"* property in step <emu-xref href="#step-array-proto-slice-set-length"></emu-xref> is intended to ensure the length is correct when the final elements of _items_ have trailing holes or when _A_ is not a built-in Array.</p>
```
michaelficarra, line 37438:
```suggestion
          <p>The explicit setting of the *"length"* property in step <emu-xref href="#step-array-proto-concat-set-length"></emu-xref> is intended to ensure the length is correct when the final elements of _items_ have trailing holes or when _A_ is not a built-in Array.</p>
```

michaelficarra, line 37438:
```suggestion
          <p>The explicit setting of the *"length"* property in step <emu-xref href="#step-array-proto-slice-set-length"></emu-xref> is intended to ensure the length is correct when the final element of _items_ has trailing holes or when _A_ is not a built-in Array.</p>
```

michaelficarra (approved):

syg (approved):

PR #3039, opened 2023-04-11
Normative: integrate well-formed Unicode strings proposal

michaelficarra:
See https://github.com/tc39/proposal-is-usv-string.


syg (approved):
lgtm. I agree with keeping it in alphabetical order.

bakkot (approved):

PR #3045, opened 2023-04-17
Editorial: Simplify EnumerableOwnProperties

gibson042:
Extends #3043 to simplify CreateArrayIterator by combining Append steps.


michaelficarra (commented):
I'm neutral on this change. I don't mind there being multiple places where an append happens.

gibson042:
I agree; the biggest benefit here IMO is clearly conveying at a glance that step 3.a.ii always results in an append.

michaelficarra:
@gibson042 Can you rebase this?

michaelficarra:
@gibson042 After discussion in editor call today, we've actually decided that we do not consider this to be an improvement. Thanks anyway for the suggestion.

PR #3046, opened 2023-04-17
Editorial: Simplify Date(…) and Date.UTC(…)

gibson042:
Extends #3043 to simplify `Date(…)` and `Date.UTC(…)` by combining common steps for handling years between 0 and 99 into a new MakeFullYear operation.

### DIFF
--- a/spec.html
+++ b/spec.html
@@ -32403,6 +32403,7 @@
           </dd>
         </dl>
         <emu-alg>
+          1. If _t_ is not finite, return *NaN*.
           1. Let _systemTimeZoneIdentifier_ be SystemTimeZoneIdentifier().
           1. If IsTimeZoneOffsetString(_systemTimeZoneIdentifier_) is *true*, then
             1. Let _offsetNs_ be ParseTimeZoneOffsetString(_systemTimeZoneIdentifier_).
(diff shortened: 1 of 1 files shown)


michaelficarra (approved):

spec.html
```diff
@@ -32409,6 +32409,24 @@ <h1>
...
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>It returns the full year associated with the integer part of _year_, interpreting any value in the inclusive interval from 0 to 99 as a count of years since the start of 1900. For alignment with the proleptic Gregorian calendar, "full year" is defined as the signed count of complete years since the start of year 0 (1 B.C.).</dd>
+        </dl>
+        <emu-alg>
+          1. If _year_ is *NaN*, return *NaN*.
+          1. Let _truncated_ be ! ToIntegerOrInfinity(_year_).
+          1. If _truncated_ is in the inclusive interval from 0 to 99, return *1900*<sub>𝔽</sub> + 𝔽(_truncated_).
+          1. Return 𝔽(_truncated_).
```
jmdyck, line 32426: Returning the truncated value here is different from what the current steps do (i.e., they say `let _yr_ be _y_`, not `let _yr_ be _yi_`), so the commit message should maybe acknowledge that and say why it's okay.

jmdyck:
It seems to me you could use `MakeFullYear` in `Date.prototype.setYear` too.

spec.html
gibson042: Fair enough; done.

gibson042:
> It seems to me you could use `MakeFullYear` in `Date.prototype.setYear` too.

Good suggestion. Done.

bakkot (approved):

